PR #244, opened 2015-01-11
add rc/grench.kak

alexherbo2:
How about pressing _Enter_ to eval selection with Grench when editing Clojure code ?


jjthrash:
I like the idea of mapping <ret> to grinch. I'd have to use it before I really figured it out, but it's easy enough to make manual again.

PR #251, opened 2015-02-20
add Relative line numbers

ekie:


mawww:
Looking at the changes, I wonder if it would not be better to remove LineNumber as a builtin face, to take the face in the command line `:addhl number_line <face spec>` and support a '-relative' switch so to enable relative mode (eventually with parameter for the current line face).

What do you think ?

ekie:
i did not remove the builtin faces yet. i thought about it and think it would be nice to keep builtin faces for the builtin highlighters, for example if u master a color scheme ur consistent with the commands defining it and moreover one could leverage matching faces when the face constructor would take the alias. it could be still possible to take a face/facespec over the commandline.

mawww:
Hello,

Yeah, for the face I agree there is an advantage to have a builtin one, we can let it like that for now.

On the implementation side, I would extract the lambdas as a single template function (with a bool relative template parameter) that does both type of highlighting, it seems there is a lot of duplicated code between the two.

Something like that:
```
template<bool relative>
void show_line_numbers(...) { ... }

HighlighterAndId number_line_factory(...)
{
    return (params[0] == "-relative") ? {"number_lines_relative", show_line_numbers<true> }
                                      : { "number_lines", show_line_numbers<false> };
}
```

mawww:
(And sorry for  the delay, I have been quite busy lately)

src/highlighters.cc
```diff
@@ -594,26 +594,58 @@ void show_whitespaces(const Context& context, HighlightFlags flags, DisplayBuffe
...
+                snprintf(buffer, 16, format, current_line - main_selection);
+                DisplayAtom atom{buffer};
+                atom.face = face;
+                line.insert(line.begin(), std::move(atom));
+            }
+        }
+        else {
+            snprintf(buffer, 16, format, current_line);
+            DisplayAtom atom{buffer};
+            (main_selection == current_line) ? atom.face = face_absolute
```
ekie: for consistency i have chosen to highlight the main selection for the default and the relative line numbers

ekie:
thank u for the hint, should be ready now.

mawww:
Merged, I followed with a small refactoring and adding a `-hlcursor` switch to chose if we want to highlight the cursor line differently.

Issue #249, opened 2015-02-10
Advertise Kakoune


mawww:
Hello Kakoune users,

I feel Kakoune is now quite mature, with no major feature missing. It seems the interaction interface is good enough for most of the use cases I thrown at it, with a good example being `clang.kak` providing code completion and syntastic like diagnostics in ~120 lines.

We have basic syntax highlighting for most common programming language (perl being the main lacking one), reasonable make and grep interaction, even a proof of concept YouCompleteMe support script (https://github.com/mawww/kak-ycmd) .

All that to say, I think it is now time to try to get more users, and hence more use cases.

This issue is there to list the different things that would help Kakoune get noticed, and hopefully attract users.

Off the top of my head:
- [x] A website
- [x] An Icon/Logo
- [x] A new demonstration video/gif demonstrating the most impressive features
- [ ] An equivalent to vimtutor to help newcomers understand how Kakoune works
- [x] Posting about Kakoune on different news site
- [ ] getting Kakoune reviewed by a Vim news site, http://usevim.com talked about Kakoune a while ago, but nothing in depth (@alexyoung Hello !)
- [x] https://github.com/neovim/neovim/issues/211 it seems there is some demand for a multiselection supporting vim-like. Maybe advertise Kakoune to the people asking for multiple carrets there (@Ivanca, @FelikZ greetings !)
- [x] Get some feedback on what Kakoune is missing, what are the pain points.

Any additional idea/comments ?

Cheers,

Maxime.

dumblob:
Just for reference, I made this info public in (sorry, it's not in English - let the Google translate be your companion :smiley:):
http://www.root.cz/zpravicky/kakoune-novy-textovy-editor-inspirovany-vimem/
http://www.abclinuxu.cz/zpravicky/kakoune-novy-textovy-editor-inspirovany-vimem

Regarding additional ideas - I have actually a feature-request list already cca half a year old as I was afraid to propose it too early. Here you are:

1. show detected file type
1. set paste (using `\` before entering insert mode will disable hooks during the whole insert session?)
1. set wrap
1. vsplit (without tmux?) + easy copying, switching etc. between the windows
1. jump to function decl
1. jump to matching parenthesis/brace/bracket/...
1. auto-enclosure of parenthesis/braces/brackets/...
1. highlight matching parenthesis/brace/bracket/...
1. loop over last 3 significant cursor positions (in Vim it's double-apostrophe `''` which loops over only 2 such positions)
[...]

PR #253, opened 2015-03-07
Logo proposal.

p0nce:
This is in reply to #249, it happens I fancy doing logos :)

### DIFF
--- a/doc/kakoune_logo.svg
+++ b/doc/kakoune_logo.svg
@@ -1,0 +1,98 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="240" height="240">
+  <defs>
+    <linearGradient id="gtl" x1="0%" y1="0%" x2="100%" y2="100%">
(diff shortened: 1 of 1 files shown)


mawww:
Looks nice, it seems inspired by the Vim logo, was that an influence ?

p0nce:
Absolutely, also I saw that neovim also got inspired by the V of vim.

mawww:
Could you make those tweaks: merge the two branch of the V, and tweak the background diamond so that the light appears to come from a specific direction ?

Not sure if that would look better, but I'd like to take a look.

p0nce:
Why not, give me some days.

p0nce:
Made some updates.

mawww:
Looks nice ! could you merge that commits in a single one and add a copyright disclaimer one before (see the CONTRIBUTING file)

p0nce:
Squashed and rebased (Gaaah `git rebase` has eaten my empty commit 3 times, very annoying)

PR #264, opened 2015-04-09
Add nextbuffer command

jjthrash:
Wraps around at the end. If only 1 buffer loaded, does nothing.

This isn't actually done. @mawww Could you please review the code (it's been years since I've done C++)? I tried a prevbuffer command, but got stuck with C++ details.

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -375,6 +375,54 @@ const CommandDesc buffer_cmd = {
     }
 };
 
+Buffer& next_buffer(Context& context) {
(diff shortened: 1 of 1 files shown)


jjthrash:
For what it's worth, here's a diff to get to my current working copy: https://gist.github.com/3b29efbc83ad1e612ebb

My goal was to get that working, then refactor next_buffer(context) into next_buffer(begin, end, context), and have next_buffer be able to use a reverse_iterator or something. Not sure if that's feasible.

PR #270, opened 2015-05-09
Markup color

benoit-intrw:
This is a proposal to used named color for markup languages like it is done for other languages.


mawww:
Thanks !

PR #274, opened 2015-06-02
Fix Makefile for Fedora packaging system

jkonecny12:
There was a problem with absolute path for symlink.
I think most of the packaging systems will be affected by this problem.

This happens because packaging build to special folder and after that it will create package from that folder. This means that the package is created with bad absolute symlink.


jkonecny12:
Only for information your last commit https://github.com/mawww/kakoune/commit/1ff0fbb4e2c761f263133be8824b333c3b4cfbdb unfortunately doesn't fix the issue.

You can look [here](https://copr-be.cloud.fedoraproject.org/results/jkonecny/kakoune/fedora-20-i386/kakoune-0-28.20150602git1ff0fbb4.fc22/build.log.gz).

PR #280, opened 2015-06-10
Add syntax highlighter for kickstart files

jkonecny12:
This is my first highlighter so I will be grateful for any comments.


jkonecny12:
I did the changes on which you pointed me and also I did small change on the behavior.

------
I think that the ``%pre``,``%post``, ``%end`` etc... are better on their own lines so I changed the behavior that only ``\h*`` can be from beginning of the line.
If I found this behavior as incorrect I will change this in the future.

PR #289, opened 2015-06-23
Fix homebrew url

danielma:


danielma:
Whoops, this is a duplicate of #288 

Closing

PR #292, opened 2015-06-25
Adding an argument text object

elegios:
I made an argument text object, as it's a feature I find really, really useful in vim. It works for all cases I tested it with, but I haven't yet looked into how testing is done in this project (though I plan to), thus this request is mostly to get someone else to look at it and get comments.


elegios:
The object now detects mismatched braces and gives the previous selection if they don't match up. Still thinking about what would be best to use as a key for it/as a replacement for the angle text object.

elegios:
As far as I can tell there are essentially two options for the key:

- Keep vim compatibility by using ```a``` for the argument object and either dropping the extra key for angles or using a new key for it.
- Keep compatibility with what has been done so far in kakoune and give the argument object a new key. One idea here would be ```u```, which is the first key in "argument" that has not yet been used.

Personally I would much prefer the former alternative, but I don't really think a random first-time committer should be allowed to make that decision. Ideal here would probably be to allow remapping of object keys, but I'm not really sure how much work that is.

Either way, if I get a decision on how to proceed with the key I'll implement it, fix tests and add new ones, after which I hope this thing can be merged.

mawww:
I think we can go with u for now, adding a few tests would be nice as well.

mawww:
Could you add a copyright waiver commit as well (see the CONTRIBUTING file) ?

elegios:
There, added tests, it uses ```u```, and there is a copyright waiver. Hope it's good enough :)

mawww:
Thank you.

PR #300, opened 2015-06-27
Ui related fixes

ekie:


mawww:
Anything remaing here you'd like merged ?

mawww:
I think it is better to keep the previous behavior of falling back to default when the option is not found, rather than keeping the previous value. The reason is that `ui_options` can be set by buffer/window, so when not specified, you will get the value from the previous buffer/window displayed, which is some kind of confusing statefulness. So I'd rather keep the current fallback to default when not specified behaviour.

mawww:
cherry picked 959c28a and added support for disabling mouse a little differently.

PR #305, opened 2015-07-01
Remap objects

elegios:
Added support for remapping text object keys.

I'm impressed with how simple this change was. Nevertheless, let me know if I've missed something.


mawww:
Thanks.

PR #306, opened 2015-07-02
Add a scrolloffalways option

elegios:
Gives the option of always having ```scrolloff``` work, even when near the end of the buffer.

Also fixes a crash bug when the cursor is near the end of the window, the end of the buffer and the beginning of the buffer and ```scrolloff``` is set to something greater than 0.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -186,6 +186,9 @@ void register_options()
     reg.declare_option("scrolloff",
                        "number of lines and columns to keep visible main cursor when scrolling",
                        CharCoord{0,0});
+    reg.declare_option("scrolloffalways",
+                       "always keep lines visible, even when at the end of the buffer",
+                       false);
     reg.declare_option("eolformat", "end of line format: 'crlf' or 'lf'", "lf"_str);
     reg.declare_option("BOM", "insert a byte order mark when writing buffer",
                        "no"_str);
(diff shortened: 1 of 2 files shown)


mawww:
Hello,

I am not really convinced having an option to control that is worth it, why do you want the window to continue scrolling past the end of the buffer ?

I would more be in favor of displaying top/bot somewhere in the status line.

what advantages do you see in this new behaviour ?

elegios:
It's essentially the feel of the thing. I personally prefer to always have the current line use the scrolloff margin, as I find it uncomfortable to edit things at the very bottom of the window. I always want to see what I'm about to "write into", even if it's just empty space. 

So a subjective thing that I'd personally like to have I guess. Don't know to what degree others want it, but it kinda bothers me when it's absent. 

Showing top/bot would be solving a different problem.

elegios:
Scrolloffbot/-tom might be a better name though

mawww:
Hum, I do not use scroll offset a lot myself, so I think I would prefer to just change the default behaviour. What do other editor do for that ? Any opinions from other users of scrolloff ?

d3m1gd:
if i understand correctly, the editing is always happening in the middle of the screen?
if so, i am all for it! and have a vim settings for it for ages already: 

    set scrolloff=999    " scrolling with cursor always
    set sidescrolloff=15 " staying in the middle
    set sidescroll=1     " of the screen

elegios:
I believe that vim has the current kakoune behavior with regards to the end of file while emacs scrolls quite differently, but does show lines past the end of the buffer. Sublime text has an option for whether it is allowed to scroll past the end of the buffer or not, but it doesn't appear to have a scrolloff equivalent without packages.

elegios:
Thank you :)

PR #339, opened 2015-07-25
Use CPPFLAGS to specify flags for C++ preprocessor

MakeNowJust:
GNU make document says:

> <dl>
>   <dt><code>CXXFLAGS</code></dt>
>   <dd>Extra flags to give to the C++ compiler.</dd>
>  <dl>
>   <dt><code>CPPFLAGS</code></dt>
>   <dd>Extra flags to give to the C preprocessor and programs that use it (the C and Fortran compilers).</dd>
> </dl>

And, `CPPFLAGS` are also used by C++ preprocessing in GNU make's implicit rules.

In other case, homebrew set `CPPFLAGS` to specify C preprocessor flags (e.g. include path, macro definition).  So, please use `CPPFLAGS` in kakoune Makefile.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -19,13 +19,13 @@ else ifneq (,$(findstring CYGWIN,$(os)))
     LIBS += -lncursesw -lboost_regex -ldbghelp
 else
     LIBS += -lncursesw -lboost_regex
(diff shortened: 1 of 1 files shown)


mawww:
That makes sense, thanks for the patch.

PR #340, opened 2015-07-26
exheres.kak: highlighting for Exherbo's exheres and exlibs 

Somasis:
Exherbo uses shell scripts which have exheres-0 and exlib as the extensions.


Somasis:
@mawww It has been changed to just be it's own separate file which sets the filetype to sh when the file suffix matches.

Somasis:
I am going to close this and submit another request when I get stuff merged correctly.

PR #341, opened 2015-07-27
Makefile: don't hardcode ncursesw path in CPPFLAGS

Somasis:
The hardcoded path seems pointless since it should be found by CPP through searching the system's include directory. In addition this would probably help with some cross compilation issues.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -19,7 +19,7 @@ else ifneq (,$(findstring CYGWIN,$(os)))
     LIBS += -lncursesw -lboost_regex -ldbghelp
 else
     LIBS += -lncursesw -lboost_regex
-    CPPFLAGS += -I/usr/include/ncursesw
+    CPPFLAGS += -Incursesw
     LDFLAGS += -rdynamic
 endif
 


Somasis:
On a properly configured system with a properly installed ncursesw library, it should.

```
$ x86_64-pc-linux-gnu-gcc -print-search-dirs | cut -d' ' -f2- | tr : '\n' | cut -d= -f2- | uniq | while read line;do readlink -m "$line";done
/usr/x86_64-pc-linux-gnu/lib/gcc/x86_64-pc-linux-gnu/5.2.0
/usr/x86_64-pc-linux-gnu/libexec/gcc/x86_64-pc-linux-gnu/5.2.0
/usr/x86_64-pc-linux-gnu/libexec/gcc/x86_64-pc-linux-gnu
/usr/x86_64-pc-linux-gnu/lib/gcc/x86_64-pc-linux-gnu/5.2.0
/usr/x86_64-pc-linux-gnu/lib/gcc/x86_64-pc-linux-gnu
/usr/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/bin/x86_64-pc-linux-gnu/5.2.0
/usr/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/bin
/usr/x86_64-pc-linux-gnu/lib/gcc/x86_64-pc-linux-gnu/5.2.0
/usr/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/lib/x86_64-pc-linux-gnu/5.2.0
/usr/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/lib
/usr/x86_64-pc-linux-gnu/lib/x86_64-pc-linux-gnu/5.2.0
/usr/x86_64-pc-linux-gnu/lib
/usr/x86_64-pc-linux-gnu/lib/x86_64-pc-linux-gnu/5.2.0
/usr/x86_64-pc-linux-gnu/lib
```

mawww:
I was talking about the include directory, it seems to me that -Incursesw (thats a capital i) will add the ncursesw directory relative to gcc invocation in the search path, not /usr/include/ncursesw.

Somasis:
Oh. https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html seems to say that there's no need for specifying /usr/include/ncursesw, though? It should be able to find it with those paths it searches by default listed on that page, shouldn't it?

mawww:
Not sure I got you, the gcc doc does not say anywhere that `-Incursesw` will add /usr/include/ncursesw to the include search path. That is needed for Debian, that does not install ncurses header in /usr/include directly (and I'd like to avoid adding a dependency on pkg-config if possible). Do you have a case where that `-I/usr/include/ncursesw` is problematic ?

PR #342, opened 2015-07-27
glep42: highlighting for Gentoo/Exherbo news items

Somasis:
This adds a glep42.kak file which does syntax highlighting for GLEP-42 news items on Gentoo and Exherbo Linux. The format is close to the format of mail files, so I used that as a base.


Somasis:
I am going to close this and submit another request when I get stuff merged correctly.

PR #351, opened 2015-08-02
Prevent the hook from wrapping the next line when the cursor is at th…

lenormf:
…e end of the current line


lenormf:
The command now uses `<a-x>` instead of `<a-h><a-l>`.

jkonecny12:
Good job I was trying to fix this yesterday

PR #356, opened 2015-08-03
Allow different type of pre-processing guards to be included in C headers.

lenormf:


lenormf:
Is that feature OK ? I'd like to be able to chose whether kak populates empty C headers.

mawww:
That feature is fine, however I'd rather have a single option "c_include_guard_style" which could be either ifdef, pragma, or none.

lenormf:
Done.

PR #358, opened 2015-08-03
Document the commands that didn't have any docstring

lenormf:
I also modified ctags' `funcinfo`'s name (added a "ctags" prefix, since it's a publicly visible command), and grep's `next` and `jump` (one of which isn't supposed to be used directly, as far as I can tell)

### DIFF
--- a/rc/autorestore.kak
+++ b/rc/autorestore.kak
@@ -36,7 +36,7 @@ def -hidden _autorestore-restore-buffer %{
 }
 
 ## Remove all the backups that have been created for the current buffer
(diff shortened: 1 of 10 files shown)


lenormf:
I rebased a couple commits, according to the comments that were made on the original ones:
* The `try %{}` scope now also includes the `%sh{}` statement of the `ctags`funcinfo` command
* The `jump` wasn't supposed to be hidden in the first place, so it's now publicly visible again, and it has been renamed to `grep-jump`. The `prev` command has also been renamed to `grep-prev`

PR #360, opened 2015-08-04
Add a script that fetches remote files into a new buffer

lenormf:
When a new buffer is created with a URL as its name, the autodownload
script will fetch the file and insert its content into the buffer.
Several options are included to automatically hide the download progress
when the file has been successfully downloaded, or to pick a particular
downloading program.

### DIFF
--- a/rc/autodownload.kak
+++ b/rc/autodownload.kak
@@ -6,8 +6,8 @@ decl bool autodownload_keep_log no
 decl str autodownload_format
 
 ## Pre-defined formats for different popular download tools
-decl str autodownload_format_wget "wget -o '{fifo}' -O '{buffer}' '{url}'"
-decl str autodownload_format_aria2 "aria2c -o $(basename '{buffer}') -d $(dirname '{buffer}') '{url}' > '{fifo}'"
+decl str autodownload_format_wget "wget -o '{progress}' -O '{output}' '{url}'"
+decl str autodownload_format_aria2 "aria2c -o $(basename '{output}') -d $(dirname '{output}') '{url}' > '{progress}'"
 
 ## Set the default downloader to be wget
 set global autodownload_format %opt{autodownload_format_wget}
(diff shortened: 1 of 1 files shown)


lenormf:
Quick note: this should have its own entry on the tracker, but since the issue only arises when using this script, I'll just post a comment on the PR.

When a URL is passed directly to kak (`kak http://site.com/file`), everything works fine, but when I try to edit a new buffer with the URL as a name, I get a failed assertion:
```
assert failed "m_count == 0" at safe_ptr.hh:30
[Debug Infos]
pid: 12427
callstack:
kak(_ZN7Kakoune9BacktraceC1Ev+0x18) [0x5002f8]
kak(_ZN7Kakoune16on_assert_failedEPKc+0x2b) [0x4feb9b]
kak(_ZN7Kakoune6BufferD2Ev+0x4c8) [0x4e6668]
kak(_ZN7Kakoune6BufferD0Ev+0x9) [0x4e6a09]
kak(_ZN7Kakoune13BufferManager18clear_buffer_trashEv+0x26) [0x492536]
kak(_Z10run_serverN7Kakoune10StringViewES0_bbbNS_9ArrayViewIKS0_EE+0xa20) [0x4d29b0]
kak(main+0xd51) [0x456711]
/usr/lib/libc.so.6(__libc_start_main+0xf0) [0x7efd93f26790]
kak(_start+0x29) [0x459d69]
```
Maybe I'm doing something unproperly in my script ?

mawww:
The bug is due to kakoune not properly supporting a buffer being deleted in the Buf{New,Create,Open} hooks. I'll see if that can be fixed.

lenormf:
So in theory, it is possible to issue an `edit -fifo` command on a buffer that has already been open without a named pipe ? That'd be convenient.

lenormf:
Since the script has been added to the community scripts and I can't reproduce the bug anymore (it has probably been fixed), I'm closing this issue.

PR #363, opened 2015-08-05
Fix an unecessary files listing when restoring buffers

lenormf:


lenormf:
I tossed in a couple commits that fix the depth issue with the purge command of script, and the useless check for the yes value of booleans.

mawww:
merged (through a rebase), thanks

PR #366, opened 2015-08-05
Add autowrap to git-commit

jkonecny12:
Git message is recommended to be wrapped to 72 characters.
[Source](http://chris.beams.io/posts/git-commit/#wrap-72)

This is a kind of RFC I want to ask everyone if we want this as opt-in or opt-out solution.

And @lenormf thank you for fixing the ``autorwrap-enable`` function.

### DIFF
--- a/rc/git.kak
+++ b/rc/git.kak
@@ -8,6 +8,9 @@ hook global WinSetOption filetype=git-commit %{
     addhl -group git-commit-highlight regex "\`[^\n]*\n\h*(?!#)([^\n]*)\n?" 1:default,red
     addhl -group git-commit-highlight regex "^\h*#[^\n]*\n" 0:cyan,default
     addhl -group git-commit-highlight regex "\<(?:(modified)|(deleted)|(new file)|(renamed)):([^\n]*)\n" 1:yellow 2:red 3:green 4:blue 5:magenta
+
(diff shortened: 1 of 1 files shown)


lenormf:
You're welcome.

However, I think it'd be best to only highlight the parts of the message that within the boundaries of the "git message guidelines", just like it's done with the excerpt of the commit:
`addhl -group git-commit-highlight regex "\`[^\n]{1,50}" 0:yellow`

Maybe you could make it so that the first 72 characters of the lines that follow the excerpt are highlighted as well ?

mawww:
My problem with that is the dependency from git.kak to autowrap.kak, I am not convinced it is worth it to introduce it, Keeping that as opt-in (its an easy hook to add in the user kakrc) has my preference.

Somasis:
Opt-in sounds better to me. I usually write more than 72 characters and think that guideline is dumb. :)

jkonecny12:
@lenormf I don't think that highlighting is a proper way. You can always look on the lower left corner where you will see on which column you are now.

So I think opt-in is the option and if we don't want to introduce dependency from kak to kak file here. I think I will add it to kakoune tips instead of this PR.

I'm closing this PR. Thank you all for the responses.

PR #367, opened 2015-08-06
Add ini files highlighter

jkonecny12:
Now it's used for .repo, .socket, .target, .service and .ini files


jkonecny12:
Finally updated, sorry it takes so long.

PR #368, opened 2015-08-08
rc/exherbo: initial support rc

Somasis:
Requesting feedback.


lenormf:
You might want to escape the dot before the extension in your regexp.

PR #391, opened 2015-08-29
Removed useless softlink creation that blocked reinstallation

Maroniemann:
I don't know why that recursive softlink was created, but it only works when Kakoune isn't already installed.


Maroniemann:
Seems like I misunderstood that line. Should I put the fix on a new commit or amend the previous one?

mawww:
Amending the previous one would be best. Thanks for reporting that.

Maroniemann:
Updated

mawww:
Thanks a lot !

Maroniemann:
You're welcome

PR #394, opened 2015-09-04
Add D support, with syntax highlighting and auto-formatting using dfmt

lenormf:


lenormf:
The regex I came up with to highlight numbers should be reviewed and possibly backported to other scripts, as it handles several cases other highlighters don't (floating points, prefixes etc).

PR #399, opened 2015-09-07
Allow custom format string for line numbers

### LINKED ISSUE #295: Allow customization of the number_lines format
Hi,

I'd like to be able to customize how the line numbers are displayed at the beginning of each line, and thus have control over the ```format``` used in ```highlighters.cc:693```. This feature would be convenient to print a space after the line count (and not a pipe), or change the character that follows it.

Thanks.

ekie:
Closes #295


mawww:
Took another approach with 3262a6902c021a0e2b1b978012132e61bf697940 as I'd like to avoid adding built in options to control behaviour of highlighters. Use `addhl number_lines -separator '...'` to control the separator.

PR #400, opened 2015-09-10
rc/sh: support single brackets

Somasis:
This adds support for things like `if [ -e "$file" ]` whereas without this, only `if [[ -e "$fille" ]]` would get highlighted.


Somasis:
Okay, I changed it to just use a `.`.

mawww:
I still dont get it, why do we want to match anything after the closing `]` ? Why  is not `\s\]` sufficient ?

Somasis:
The idea was to support both single and double brackets, I figured out a better way to do it now

mawww:
Ok, however double brackets are already handled by the existing `[\[\]\(\)&|]{2}` regex, they miss the spaces though, so it would be nice to remove the brackets from there.

mawww:
Fixed by 0df72bd

PR #402, opened 2015-09-11
Add basic support for tup's Tupfile format

lenormf:


pickfire:
@lenormf Sorry for bumping this but do you find `tup` faster than `make`? Also, `tup` seemed to be broken right now, should we remove it?

lenormf:
Thanks for reminding me, I will make a PR to update the `Tupfile`.

Tup is indeed noticeably faster at initiating the compilation process (i.e. running `g++`).

pickfire:
Nah, I even used `tup monitor` but still Makefile is faster. `tup` seemed to only be faster when no files are changed.

PR #431, opened 2015-10-11
C-family: Add support for astyle code-formatting

lenormf:


mawww:
Hello, I am not sure that should be in c-family.kak, but rather in a new astyle.kak file, I'd say c-family.kak should stay independent of non-posix tools.

lenormf:
I'm just trying to be consistent with the other language files I maintain (`dlang.kak` + `golang.kak`), they all have their own `format-code` alias which use the most popular formatting tool. I thought it'd be nice to have a `format-code` alias declared for C/C++/ObjC as well, since `astyle` handles those three languages and is arguably the best formatting tool available.

lenormf:
Ignore the changes I made in the last commit related `comment-line`, they'll go away in the future.

PR #444, opened 2015-10-22
Introduce `formatter.kak` to handle formatting.

lenormf:
The modified files used to handle formatting code with the same code, and a
different command, the process is now centralized into a single file. Language
support files should set the `formatcmd` string to whatever formatter is the
most popular in their respective community, eventually with default options.


lenormf:
I stripped the commenting functions from the commit.

PR #446, opened 2015-10-25
Introduce the `commenting.kak` file

lenormf:
Centralize commenting functions into a single file, and declare
variables used by those functions in language support files. The
functions are `comment-selection` and `comment-line`, users should bind
those functions to the keys of their liking for easy access.


lenormf:
Quick note: there's still a massic bug in that commenting function, when dealing with commenting characters that contain '<' or '>' signs (which HTML uses…). The exec calls would rather have '<lt>' or '<gt>' sequences instead of the raw characters, but I haven't come up with a solution that works for all types of commenting characters yet.

PR #447, opened 2015-10-26
Use braille characters to mark whitespaces

FranzSkuffka:
Replaced the dots to mark whitespace with appropriate braille characters.
Now we can finally distinguish dots and whitespaces.

### DIFF
--- a/src/highlighters.cc
+++ b/src/highlighters.cc
@@ -683,7 +683,7 @@ void show_whitespaces(const Context& context, HighlightFlags flags, DisplayBuffe
                         atom_it->replace("→" + String(' ', count-1));
                     }
                     else if (c == ' ')
-                        atom_it->replace("·");
+                        atom_it->replace("⠐");
                     else if (c == '\n')
                         atom_it->replace("¬");
                     break;


FranzSkuffka:
Jan Wirth Copyright Waiver

  I dedicate any and all copyright interest in this software to the
  public domain.  I make this dedication for the benefit of the public at
  large and to the detriment of my heirs and successors.  I intend this
  dedication to be an overt act of relinquishment in perpetuity of all
  present and future rights to this software under copyright law.

lenormf:
I have several questions:
* why use those characters instead ? The `¬` character is widely known for being a "whitespace character" already
* why `source` another `kakrc` file in an arbitrary directory ?

As for the waiver, it should be added to the project as an actual commit (and usually is required on arguably big contributions).

FranzSkuffka:
I tried to avoid collision/ similarity with characters being used in the code. the whitespace characters that were used before looked like letters and had the exact same colors as 90% of the source code at some. It was just incredibly hard to read because they look like non-whitespaces. I got used to it too, but this weekend I had to code with other people and they could barely read my spaces-indented files. `¬` is debatable, but it has some visual similarity with `|` or `-`. The character I chose now looks like `:` though, wich is even worse.

image

I did not mean to add the source command to the pull request. It was an accident. But I did this to extend my my `kakrc` through my dotfiles directory so that I do not have to modify the source code of kak to customize it. Or is there a different way to do this (like the `.vimrc`)?

lenormf:
I think the main issue here is that there's no way to differenciate what characters are actual parts of the buffer, and which ones are whitespaces: why don't you create a particular face for those characters (to display them in inverse video for example), instead of changing the characters themselves ? This maybe raises a point: should users be able to customize those whitespace characters ?

As for the `kakrc` thing, I still don't get it, why not just do everything from `${XDG_CONFIG_HOME}/kak/kakrc` ?

FranzSkuffka:
Indeed, this is just a quick fix. We could use a more advanced whitespace highlighter. So we could work with background colors for example instead of inverse video.

vim currently displays whitespace like this for me:
image


Is the `${XDG_CONFIG_HOME}/kak/kakrc` feature documented somewhere I could read it up?

lenormf:
You might be able to get away with it by setting the space char to a unicode box (`▮`) with a particular foreground color.

FranzSkuffka:
now only if i knew how to set the foreground color :grin:

FranzSkuffka:
Am I correct assuming that to add user-configured whitespace I would have to add it as an option to `main.cc` and getting the option value inside the `highlighters.cc` ?

mawww:
yep, what I wonder is if we want a configurable whitespace replacement highlighter, or a more general codepoint replacement (display a set of arbitrary codepoints using other codepoints) or even a general regex based one (display all matches of a regex with a different text).

FranzSkuffka:
regex always sounds great. unfortunately I cannot contribute then (weak C++Fu), but i would greatly appreciate it. I would dive into creating a smart regular expression for whitespace though to get my regex-fu up and running.

FranzSkuffka:
I changed the font now and it's better. A superior syntax highlighter would be great though.

PR #457, opened 2015-11-02
Patch to build on FreeBSD

alpha123:
Added a condition in file.cc `get_kak_binary_path()` for FreeBSD. Uses sysctl, which as far as I can tell is the standard way to get the binary's path on FreeBSD. Might not be applicable to other BSD systems.

The Makefile was a little tricky to get right and was mostly trial-and-error. Building it with gcc requires rebuilding boost with gcc and libstdc++ (which I didn't feel like doing), but clang seems to require libgcc_s to build kakoune, thus I included `-L/usr/local/lib/gcc48`, which of course won't exist if the user has a different gcc version. Nevertheless, I did manage to build it on my system (FreeBSD 10.1, clang 3.4.1) with this.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -28,6 +28,10 @@ os := $(shell uname)
 
 ifeq ($(os),Darwin)
     LIBS += -lncurses -lboost_regex-mt
+else ifeq ($(os),FreeBSD)
+    LIBS += -ltinfow -lncursesw -lboost_regex
+    CPPFLAGS += -I/usr/local/include
+    LDFLAGS += -L/usr/local/lib
 else ifeq ($(os),Haiku)
     LIBS += -lncursesw -lboost_regex -lnetwork -lbe
 else ifneq (,$(findstring CYGWIN,$(os)))
--- a/src/file.cc
+++ b/src/file.cc
@@ -12,6 +12,10 @@
 #include <unistd.h>
 #include <dirent.h>
 
+#if defined(__FreeBSD__)
+#include <sys/sysctl.h>
+#endif
+
 #if defined(__APPLE__)
 #include <mach-o/dyld.h>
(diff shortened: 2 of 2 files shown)


mawww:
Hello, thanks for the pull request, could you squash the two commits together, and maybe see what we can do about that libgcc_s thing, its really strange it is needed, do you know what library requires it ? Which symbols are needed ?

alpha123:
Squashed.

Something or another requires `_Unwind_Resume@@GCC_3.0`, and ld helpfully suggests "try adding -lgcc_s", even if I attempt to build it with libunwind instead. (Why it gets mangled with @@GCC_3.0 I have no clue.)

Of course it's entirely possible my setup is screwy (I have at least 1 clang and 2 gcc versions). Can anyone else on FreeBSD verify whether or not we need libgcc_s?

mawww:
I was wondering if backtrace.cc was requiring that (the backtrace call is glibc specific IIRC), but it should not be compiled on FreeBSD.

That `@@GCC_3.0` sure is strange, it seems _Unwind_Resume is part of the exception handling runtime. Why clang does not link to it automatically is curious as well.

somasis:
I don't use FreeBSD, but I'm pretty sure that's just that clang uses libgcc_s (which is GCC's runtime library) if it exists.

mawww:
Hi,

I am thinking we could merge that change without the gcc48 lib path and require users to set LD_LIBRARY_PATH if they do not have libgcc_s directly available. I'd like to avoid working around a system problem in the Makefile.

What do you think ?

alpha123:
I think that would be fine. Most of the time I think clang should be able to find it (it's one of the last gcc pieces that FreeBSD still relies on, because libunwind is still rather primitive). I don't know why it couldn't in this case.

PR #459, opened 2015-11-03
Use c-family's next line indentation for Java files

dermesser:
Make editing Java less annoying with this simple change!

### DIFF
--- a/rc/c-family.kak
+++ b/rc/c-family.kak
@@ -112,7 +112,7 @@ addhl -group /objc/code regex "\<(const|auto|inline|static|volatile|struct|enum|
 addhl -group /objc/code regex "@(property|synthesize|interface|implementation|protocol|end|selector|autoreleasepool|try|catch|class|synchronized)\>" 0:attribute
 addhl -group /objc/code regex "\<(IBAction|IBOutlet)\>" 0:attribute
 
-hook global WinSetOption filetype=(c|cpp|objc) %[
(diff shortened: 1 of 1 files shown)


lenormf:
I don't think partly handling java buffers in the `c-family` script is a good solution, you'll have to find a way to keep hooks where they belong (i.e. modify the java file).

mawww:
Hello,

We have different options here:

* We add java indentation directly to c-family.kak, as you proposed, which is not very satisfying because we have java support in both c-family.kak and java.kak.
* We copy the indentation code from c-family.kak to java.kak, which duplicates code, but allow us to fix some incompatibilities (we dont want to add a semicolon after class closing brace in java).
* We reference the `_c-family-indent-*` commands from java.kak, in the existing filetype detection hook, which adds a dependency between java.kak and c-family.kak

None is totallly ideal, I guess the simple change you propose could be merged and we could eventually fix that more cleanly later.

Any additional feedback ?

lenormf:
I'd rather go for #3, creating a dependency seems more sane than sprinkling keywords in an unrelated file.

alpha123:
+1 to @lenormf, a dependency with c-family seems reasonable since Java is, after all, a C-syntax language.

mawww:
Ok, option 3 seems fine, can you change the pull request to implement that ?

mawww:
This is pretty out of date, closing.

PR #460, opened 2015-11-03
Add some regex to parse c-family numbers

angt:
Better but still not perfect.

### DIFF
--- a/rc/c-family.kak
+++ b/rc/c-family.kak
@@ -93,12 +93,24 @@ def -hidden _c-family-indent-on-closing-curly-brace %[
 }
 
 # c specific
+addhl -group /c/code regex %{\<(0[xX][0-9a-fA-F]+|0[0-7]*|[1-9][0-9]*)[uU]?[lL]{0,2}\>} 0:value
+addhl -group /c/code regex %{\<[0-9]+\.[0-9]*([eE][+-]?[0-9]+)?([fF]|[lL])?\>} 0:value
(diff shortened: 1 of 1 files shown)


jedisct1:
I agree that the current regex is way too permissive.

Highlighting is really useful in order to spot typos and other syntax errors, and the current regex will let them slip through.

This pull request seems to work fine for me. It would be nice to consider its inclusion.

lenormf:
I wrote a regex to highlight numbers in the [D support script](https://github.com/mawww/kakoune/blob/master/rc/dlang.kak#L33), and I faced the same dilemma: do we want to keep the regex as simple as possible but potentially have mismatches ? Or do we make the regex engine work more but have a stricter rule on what values are actually numbers ?

Like @mawww, I think there's very little benefit from lining up the most accurate regular expressions, as what we want to match are *just* numbers. Simpler -although more error prone- regular expressions will keep highlighting coherent and the load on the regex matcher lighter.

mawww:
I do not oppose more strict number matching, but I think 5 new expressions is too much, we should compact that, and the current commit keeps the existing regex which defeat the purpose.

jedisct1:
They can be condensed into a single expression, but it will greatly impact readability.

mawww:
Does not need to be a single one, I'd be ok with one integer regex and one floating point regex

mawww:
This is pretty out of date, closing.

PR #466, opened 2015-11-04
Gas syntax

danlrobertson:
I added this as I personally need syntax highlighting for gas assembly files for viewing disassembled C/C++/Rust code. As a result, I definitely will not be offended if you do not merge this pull request, especially since there are so many different assemblers. Comments and suggestions are welcome. I havn't come even close to covering the whole of gas assembly with this, but I plan to keep working on this, as I personally need this feature. I will need to write a kak file for LLVM IR in the near future as well, so let me know if you guys will want that once I get the ball rolling there.

I only just built and started using kakoune a few hours ago, so I'm a super noobie :smile:

Any suggestions on places in the code to look/explore as a noobie contributor are very very welcome! As a tmux + vim user/addict I dig this project!

### DIFF
--- a/rc/gas.kak
+++ b/rc/gas.kak
@@ -1,0 +1,80 @@
+# Detection
+# ---------
+hook global BufCreate .*\.(s|S|asm)$ %{
+    set buffer filetype gas
+}
+
+
+addhl -group / regions -default code gas \
+    string         '"' (?<!\\)(\\\\)*"        '' \
+    commentMulti   /\*       \*/              '' \
+    commentSingle1 '#'       '$'              '' \
+    commentSingle2 ';'       '$'              ''
+
+addhl -group /gas/string fill string
+addhl -group /gas/commentMulti fill comment
+addhl -group /gas/commentSingle1 fill comment
+addhl -group /gas/commentSingle2 fill comment
+
+# Constant
+addhl -group /gas/code regex (0[xX][0-9]+|\b[0-9]+)\b 0:value
+# Label
+addhl -group /gas/code regex ^\h*([A-Za-z0-9_-]+)\h*: 0:operator
+# ARM Directives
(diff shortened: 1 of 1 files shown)


rc/gas.kak
```diff
@@ -0,0 +1,61 @@
+# Detection
+# ---------
+hook global BufSetOption mimetype=text/x-python %{
+    set buffer filetype gas
```
lenormf: This is going to conflict with regular python files I reckon, not good.

rc/gas.kak
```diff
@@ -0,0 +1,61 @@
...
+}
+
+hook global BufCreate .*\.(s|S|asm)$ %{
+    set buffer filetype gas
```
lenormf: This is could be a solution, except not all `.s` files use follow the GAS syntax. Some trickery might be needed here, maybe a quick grep into the buffer to find unique keywords that specific to this syntax, to be sure of the type of file you're dealing with ?

lenormf:
On a side note: in my opinion the `rc` directory of the project should only contain files that support common languages, support for a particular flavor of ASM should go onto the [community scripts page](http://kakoune.org/scripts.html) (PRs on the `gh-pages` branch).

mawww:
I agree we should avoid too exotic languages in rc/ however gas  is basically AT&T assembly syntax, which is the defacto standard assembly syntax on Linux, so I would be interested in merging that. Mime type detection should be fixed though, I dont know other filetypes that uses s/S as a common extension, so for the time being, if the file shell command does not give meaningful mimetype, the mimetype hook could just be removed.

danlrobertson:
Sorry for the pythin mixup. I must have forgot to remove that after I copied it from the python file. My bad...

I'll remove that and I'll remove the mime type detection.

Since in intel assembly the registers are not prefixed with a % that could be an easy solution using grep. Thanks for the comments. I'll push some edits in a little bit.

danlrobertson:
Removed the mime type detection and should have all of the general purpose instructions for x86 att/gas. I'll be adding floating point instructions sometime soon. Let me know if you see anything else that should be edited.

lenormf:
@mawww OK then. How about a cleanup of the current list of scripts that's automatically loaded ? We can move some of the ones that are -arguably- not used much at all to the community list. A couple ones off the top of my head:
* coffeescript
* cucumber
* exherbo
* kickstart
* …

PR #471, opened 2015-11-06
Add missing keywords and attributes to c-family syntax

danlrobertson:
Add a few more common types and C++ extensions to c-family
syntax file


lenormf:
I don't think it's a good idea to start adding to the regex all the types that are defined in every header of `/usr/include` ever. It's better to stick to the keywords that are actually defined as part of the language but not too specific, until we have a generic way of detetcing types (including user types).

danlrobertson:
No problem, removed the various types from ctypes.h but left the missing attributes and keywords.

PR #472, opened 2015-11-06
Move tmux support to its own file, and improve it

lenormf:


lenormf:
Branch updated to reflect the changes suggested.

mawww:
Looking a little further, seems the client.kak now basically became x11.kak, so maybe we should rename focus-default as focus-x11 and have new-x11/new-tmux commands with new being an alias.

The remaining problem would be to cleanly handle priorities, the use ok KakBegin hook in tmux.kak to make sure we run after x11/client.kak is quite hackish. I guess it'll do for now until we find a better way.

PR #499, opened 2015-12-01
Makefile: only add the -g flag to the CXXFLAGS in debug mode

lenormf:


ff2000:
What about using "-ggdb" instead of simply "-g"? That would make backtraces more informative.

lenormf:
That can be arranged, I'll add it to this PR if @mawww doesn't mind :)

mawww:
Why remove the -g on non debug builds ? it is still nice to have the debug symbols available, and I dont see any disadvantages to them (apart from binary size, but you can always strip the binary afterwards)

lenormf:
Since debug flags are added to the binary even in non-debug builds, it seemed like a reasonable decision. I found out about this behavior while building kakoune on an old computer, so I'd rather have the correct version built at once without having to spawn another process after compilation.

Issue #494, opened 2015-11-25
Include the page number in the completion suggestions of man pages


lenormf:

lenormf:
Just throwing an idea: what about buffer metadata that could be displayed after the buffer names when completed upon ?

For instance, when switching buffers with `:b man`, the completion engine could print `*man* (3)`, or something. Other buffers could have the cursor's position set as metadata or something useful.

PR #500, opened 2015-12-01
Man: add page numbers

lenormf:
Issue #494


lenormf:
Dropped the commit that limited the amount of suggestions displayed when using the completion.

PR #502, opened 2015-12-01
Possibly use __attribute__((constructor/destructor)) to remove some startup/cleanup code from main

danlrobertson:
This is more of a thought than a full fledged pull-request, but I'm pretty sure most compilers should support __attribute__((constructor/destructor)) by now. I'm sure clang and gcc do, but I'm definitely not sure about Windows. As a result, it could be possible to reduce the amount of "startup" code by using __attribute__((constructor)) and "cleanup" with __attribute__((destructor)). If I'm honest there are quite a few reasons to be weary of a pull request like this, as it has the potential to decrease portability and increase complexity. However, if handled appropriately it could actually reduce complexity and remove some boilerplate code. In addition, constructor/destructor priorities are a nice feature that could be used in the future as more code is needed. Again, this is just a thought that in some ways I'd suggest rejecting :smile:

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -643,8 +643,7 @@ int run_pipe(StringView session)
     return 0;
 }
 
-int main(int argc, char* argv[])
-{
+__attribute__((constructor(100))) void startup() {
     setlocale(LC_ALL, "");
 
     signal(SIGSEGV, signal_handler);
@@ -654,7 +653,11 @@ int main(int argc, char* argv[])
     signal(SIGPIPE, SIG_IGN);
     signal(SIGINT, [](int){});
     signal(SIGCHLD, [](int){});
+}
 
+
+int main(int argc, char* argv[])
+{
     Vector<String> params;
     for (size_t i = 1; i < argc; ++i)
         params.push_back(argv[i]);


lenormf:
AFAIK the constructors and destructors should not contain code that is directly related to what the program is supposed to do. For example, if you wanted to patch a few addresses in the offsets table of the program, you could certainly do that in the "constructor", because this operation doesn't really have anything to do with the program, and running it between the compiler's `_main` and your code's `main` doesn't really matter.

However using them to split the code appart seems like an unecessary -and potentially non portable- trick, I don't see how it could be used to "reduce complexity" or "remove boilerplate code". If you have a particular use case in mind that would make your points, feel free to add commits to this PR !

danlrobertson:
> seems like an unecessary -and potentially non portable- trick, I don't see how it could be used to "reduce complexity" or "remove boilerplate code".

Agreed. Definitely has the strong potential to be so. As I stated, and should be `s/potential/strong potential/`

> it has the potential to decrease portability and increase complexity.

and

> this is just a thought that in some ways I'd suggest rejecting :smile:

more or less wanted to bring it up and discuss.

mawww:
Hello,

I dont see much interest in that change, it seems simpler and cleaner to have these calls at the beginning of main.

This being C++, we have a portable way to do that already.

```
struct Init
{
    Init() { ... } // put whatever code we want here
} init; // global object whose constructor will get called before getting into main 
```

PR #507, opened 2015-12-02
Suppress warnings when compiling with clang

danlrobertson:
Super small nit pick. clang complains as per function optimizations have not been implemented yet. Added simple macro check to suppress the warning.

There has been some talk about adding the feature [here](http://clang-developers.42468.n3.nabble.com/PROPOSAL-per-function-optimization-level-control-td4031670.html).

### DIFF
--- a/src/string.hh
+++ b/src/string.hh
@@ -168,7 +168,9 @@ public:
     };
     ZeroTerminatedString zstr() const { return {begin(), end()}; }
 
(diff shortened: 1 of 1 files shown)


mawww:
I am not really keen on that, if you really want to get rid of the warning, I'd suggest to suppress it in the compiler options.

PR #511, opened 2015-12-05
Add an user.mk file to store custom Makefile settings

doppioandante:
I think this is a more general solution than the one in #507

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.o
 .*.d
 *.pyc
+src/user.mk
 src/kak
 tags
 GPATH
--- a/contrib/user.mk
+++ b/contrib/user.mk
@@ -1,0 +1,17 @@
+# This file will be included by the Makefile
+
+# change to no to compile in release mode
+debug ?= yes
+
+XDG_CONFIG_HOME ?= $(HOME)/.config
+
+# kakoune will be installed in $(DESTDIR)$(PREFIX)
+DESTDIR ?= # root dir
+PREFIX ?= /usr/local
+
+# path of ncursesw headers
+NCURSESW_INCLUDE ?= /usr/include/ncursesw
+
+#uncomment the following lines to compile with clang
+#CXX=clang++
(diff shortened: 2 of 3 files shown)


lenormf:
What are the benefits of creating this file and splitting the options into several files ?

doppioandante:
@lenormf Well the options are not split, they're duplicated. This may sound bad, but simply extracting the options in another file would defeat the purpose of this change.
This is useful because you can specify your install destination/compiler/etc just once, put in the user.mk file, and then make will "remember" these options.
I think that this is better than having to mantain a custom Makefile or a custom out-of-tree script to compile (they add noise when working with git).
The worse that can happen is that Makefile and user.mk get out of sync, but this is not a big problem because make would still work ok.
The usermk rule that I have added in the Makefile is not that necessary, but I thought it is good for a user to know that it is "the way" to customize the make process.

lenormf:
Why not just `-include user.mk` then ? Also customizing the build with make is already easy to do, all you have to do is pass additional flags to the command line, so including a file shouldn't be needed (unless you have a specific problematic case in mind ?).

doppioandante:
@lenormf if you make clean/ make frequently using the flags on the shell quickly become inconvenient. In my case it would be
```CXX=clang++ CXXFLAGS=-Wno-unknown-attributes -PREFIX=/home/enrico/long stuff make```
And could potentially become worse as the project evolves.

My PR is just basically an `-include user.mk` (user.mk is inside contrib/ as an example). The only questionable part is the added rule to the Makefile, but I don't think it's problematic at all.

mawww:
So this is basically providing the equivalent to a `configure` script, I really tried to avoid that, but if we go this way, I'd rather have a config.mk file that gets generated by a `configure.sh` script and included (if present) in the Makefile. I think that would be a more familiar to most users than doing a `make usermk`.

doppioandante:
Woah woah, I would even prefer cmake to a configure script of any kind.
At this point I think I'll just mantain a separate branch for myself with a modified Makefile.

PR #512, opened 2015-12-06
Run tests loading kakrc

doppioandante:
The test/run script can optionally omit the -n flag setting the env
variable TEST_LOAD_KAKRC. This is useful to check that the default kak scripts
are actually correct and don't break kakoune's behaviour.

### DIFF
--- a/test/run
+++ b/test/run
@@ -1,8 +1,12 @@
 #!/bin/sh
 
 # Main ├────────────────────────────────────────────────────────────────────────
-
 main() {
+  if [ "$1" = "-l" ]; then
+    load_kakrc=true
+    shift
+  fi
+
   number_tests=0
   number_failures=0
   dirs="${@:-.}"
@@ -27,7 +31,13 @@ main() {
     else
       number_tests=$(($number_tests + 1))
       touch in; cp in out
+      if [ -z $load_kakrc ]; then
+        load_kakrc_command=""
+      else
+        load_kakrc_command="source $test/../share/kak/kakrc"
+      fi
       kak_commands="
+        $load_kakrc_command
         set global autoreload yes
         set global autoinfo ''
         set global autoshowcompl false
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,4 +50,4 @@ addons:
     build_command: "make -j4"
     branch_pattern: coverity-scan
 
-script: cd src && make && make test
+script: export POSIXLY_CORRECT=1 && cd src && make && make test && make test-kakrc
--- a/src/Makefile
+++ b/src/Makefile
@@ -84,6 +84,9 @@ kak : $(objects)
 
 test:
 	cd ../test && ./run
+	
+test-kakrc:
+	cd ../test && ./run -l
 tags:
 	ctags -R
 man: ../doc/kak.1.gz
--- a/test/README.asciidoc
+++ b/test/README.asciidoc
@@ -26,3 +26,5 @@ Usage
 To test, just type +run [test]+ in the +test+ directory.
 It will print each passing test.  If a test fails, a {unified-context-diff}[unified context diff]
 is printed showing the test’s expected output and the actual output.
+
+Pass the -l switch as the first argument to load the kakrc for the unit tests


doppioandante:
I'm not really sure on the approach here. Maybe there's a better way?

If I didn't mess up the travis file, this PR should fail until #513 is merged.

doppioandante:
... of course it passes, but it shouldn't. I am puzzled.

test/run
```diff
@@ -44,7 +44,13 @@ main() {
         write out
         quit!
       "
-      ${test}/../src/kak out -n -u -e "$kak_commands"
+      if [ -z ${TEST_LOAD_KAKRC+x} ]; then
```
lenormf: >${TEST_LOAD_KAKRC+x}

Is the `+x` inside the braces a syntax I don't know about, or a typo ?

I'm not too sure about the way the option is detected too, using an environment variable as a flag seems wonky, especially when you have a makefile available.

doppioandante: Yeah it is wonky I agree. That condition is true only if TEST_LOAD_KAKRC is effectively not set, but fails if TEST_LOAD_KAKRC is set and empty. +x is just a random thing, could have been +anything (http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_02).

The problem is that I don't know what variable would be sensible to use in this case. Using the actual -d switch as its value sounds ugly.
The biggest problem though is that it doesn't even seem to work on travis as it is now.

lenormf:
After reading the actual code, I'm not sure why `.travis.yml` has to be modified at all: the `run` script is in charge of testing kak, so only that file should be modified to test that default scripts are ok.

Moreover, I'm assuming the `-n` flag was originally passed because the point of this test suite is to test `kak` itself, not ensure that the script it loads don't contain any error, so if that feature had to be added, I think the sumplest would just be to remove the `-n` flag from the command.

doppioandante:
@lenormf: the problem is that testing with -n and testing without -n could potentially have different results and we are interested in both (but I may be wrong here, I still don't know the possibilities of the scripting language that well).
Although rare, It could be that without -n the test passes, but they don't *with* -n.

mawww:
I would prefer passing an argument to `test/run` than using that environment var

doppioandante:
Now the script checks for the -l flags (**l**oad kakrc). I didn't go with getops it seemed impractical.
Unfortunately the check fails only on the apple build, while passes on debian.
I have no idea why.

mawww:
Argh, `sed -r` is not posix, and not supported by osx sed... and its used in many places... I guess most of these are candidates for replacing sed with a `kak -f`...

lenormf:
Those changes seem incredibly complicated for the simple feature they're supposed to implement. Also that "buffer change" test does not seem to be testing much, and is bound hardly on the message that is displayed in the debug buffer (in the end it doesn't check that switching buffer works in general, just that there's a message in the *debug* buffer).

doppioandante:
Just to be clear, Travis should **not** pass, in fact I'm not finished yet.
I agree that the test is hardly useful and badly named, I was just trying to replicate the failure.

Regarding the complexity, I think that less than this can't be done:
 * An option in the Shell script is needed.
 * I've given up tweaking the cmd options (-n) and tried sourcing the kakrc manually. In addition to be more explicit, it is the only way to load it: the alternative would be to sudo make install on Travis so that /usr/share/kak/kakrc is loaded, or even adding an option to kak to specify the runtime path used.
* the main kakrc assumes the scripts to be in $XDG_CONFIG_HOME, so those must b e copied for the test to have any meaning.

Unfortunately I still can't get this to fail properly, maybe there's something I'm missing about the Travis environment.

test/run
```diff
@@ -1,7 +1,11 @@
 #!/bin/sh
 
-# Main ├────────────────────────────────────────────────────────────────────────
+if [ "$1" = "-l" ]; then
```
lenormf: Handling arguments should be done in main() btw.

doppioandante:
Hi, I've revamped this PR a bit, do you feel like it could be still useful?

mawww:
I am still undecided on that, at least in its current form, I fear it will cause more problems than it will solve, as the kakrc script will load the user configuration, which could break the test in many ways (hooks, changed options...).

It would be more useful if it just loaded the bundled files, which would allow us to check they dont break tests, but that does not seems very important either.

mawww:
I wont merge this in the current state, what would be nice is actual tests for the rc files.

PR #514, opened 2015-12-06
Add editorconfig script

doppioandante:
[link to the gist](https://gist.githubusercontent.com/doppioandante/daaccd116834c3900024/raw/66854d529958fa2b47f7e5aeb5758b059537e28d/editorconfig.kak)

### DIFF
--- a/scripts.html
+++ b/scripts.html
@@ -112,6 +112,15 @@
                                 <td>04/11/2015</td>
                                 <td></td>
                             </tr>
+                            <tr>
+                                <td><a href="https://gist.githubusercontent.com/doppioandante/daaccd116834c3900024/raw/66854d529958fa2b47f7e5aeb5758b059537e28d/editorconfig.kak">EditorConfig</a></td>
+                                <td>Support for <a href="http://editorconfig.org/">EditorConfig</a></td>
+                                <td><a href="https://gist.github.com/doppioandante"></a>Enrico Lumetti</td>
+                                <td>1.0</td>
+                                <td>MIT</td>
+                                <td>6/12/2015</td>
+                                <td></td>
+                            </tr>
                         </tbody>
                     </table>
                 </div>
--- a/scripts.html
+++ b/scripts.html
@@ -113,12 +113,12 @@
                                 <td></td>
                             </tr>
                             <tr>
-                                <td><a href="https://gist.githubusercontent.com/doppioandante/daaccd116834c3900024/raw/66854d529958fa2b47f7e5aeb5758b059537e28d/editorconfig.kak">EditorConfig</a></td>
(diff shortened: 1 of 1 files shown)


lenormf:
It would be more convenient to have an actual implementation in the PR rather than a link to a function, simply pulling a branch is more convenient than downloading/creating a file locally.

Other than that, it might not be a good idea to completely wrap the code logic into `awk` just yet, at least not until we can have custom interpreters (in which case you would be able to just use %awk{}).

doppioandante:
@lenormf what's the alternative? I did like that because I've seen it in a script somewhere (maybe it was clang.kak?)

lenormf:
The alternative is to use mainly the language that is being interpreted, as opposed to wrapping the entire function in a program which uses the current interpreter as a process launcher.

The `clang.kak` script uses `awk` because it would be complicated to have the same equivalent code written in pure bash. On the other hand a few `if/else if` statements with a few regular expressions should be fairly simple to have done in shell script.

doppioandante:
I see, thank you very much for the input, I will see what I come up with.

mawww:
I dont really mind this is a full awk script, the argument against direct custom interpreter is that its really easy just to do `%sh{ myinterpretter "..." }` so I wont complain if a script uses that.

So, as you prefer, I can merge that version, or if you want to change it to use sh directly, you can as well, I wont merge it if you use non posix features in sh however.

Also, BOM option syntax changed, its `none` and `utf8` now instead of `no` and `utf-8`

mawww:
Hi, would you mind changing that pull request so that the script gets included in Kakoune ? out of the box editorconfig support would be nice.

doppioandante:
As soon as I have time I will!

lenormf:
Whenever you do though, remember to modify the values assigned to the variables whose type changed some time ago (i.e. `BOM` is a flag now).

rennis250:
Is there any reason to not include this now?

doppioandante:
Sorry I forgot to mention that I updated it.
Also, it sets the indentation options to the buffer level, is this a good idea?

FranzSkuffka:
How do I use this?

mawww:
You can put the following line in you kakrc:

```
hook global BufOpen .* %{editorconfig-load}
```

That will set options from editor config for every file you open in Kakoune.

FranzSkuffka:
I have trouble getting this to work - I'm on OSX 10.12.5.
The command does execute, I have tested this. But the editorconfig is not applied.

lenormf:
@FranzSkuffka create a new issue if you have trouble using a module, otherwise this comment will most likely be forgotten as it's made on a closed PR.

alyssais:
For @FranzSkuffka and anybody else stumbling across this thread because editorconfig doesn't seem to be working, you need to install [editorconfig-core-c](https://github.com/editorconfig/editorconfig-core-c), eg with `brew install editorconfig`.

PR #522, opened 2015-12-10
Make autoload scripts more robust

doppioandante:
I had moved my kakoune installation folder and the links in ~/.config/kak were still pointing to the old scripts.
This pull request updates those links on 'make userconfig' and adds a little debug message in case one cannot be loaded.

### DIFF
--- a/share/kak/kakrc
+++ b/share/kak/kakrc
@@ -36,7 +36,7 @@ def -params 1 -docstring "colorscheme <name>: enable named colorscheme" \
     autoload() {
         dir=$1
         for rcfile in ${dir}/*.kak; do
-            echo "try %{ source '${rcfile}' } catch %{ }";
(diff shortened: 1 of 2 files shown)


lenormf:
:+1:

mawww:
I like the debug information thing, however the userconfig rule should rather be removed, by default kakoune will autoload the distribution autoload folder if no user autoload folder is present, and if the user wants to control its autoload, he can either link every files himself, or just add a link to the distribution autoload folder.

doppioandante:
What's the point of the userconfig rule then? I thought it was a nice shortcut to link all the scripts at once (therefore the -f options would have made sense).

mawww:
The userconfig should have been removed long ago, thats mostly laziness from my part, its from when kakoune would only autoload from the user folder, we changed that a while ago so that `make userconfig` was not requiered to get access to provided scripts, as that was not convenients for user installing through a package manager.

doppioandante:
Thanks for the explanation. I've pushed the changes.

PR #525, opened 2015-12-10
Add a lucius theme

benoit-intrw:


benoit-intrw:
Strange, I have tested it before sending the pull request and seen no error. 

And after checking it again, it seems that when the missing color (bad name) is for a "bg_color" there is no error. Maybe because "bg_color" is optionnal ?

Anyway i'll fix it.

lenormf:
Can you repost what your originally faulty line was please ? After reading the implementation, it seems that the color parsing function does throw an error when a color wasn't declared/detected.

lenormf:
Bug fix is in #526

benoit-intrw:
Yes it is about  `lucius_blue_2` ; when I put an unknow variable into a `fg_color` I got error :

```
'face' wrong argument count
```

But if this is a `bg_color` which is unknow, there is no error.

It's certainly good that the *facespec* parsing complain about these error, but why it is possible to use an undefined variable in the `%sh` block ?

lenormf:
The shell replaces undeclared variables with an empty "" string when the variables are expanded, which means that your face description expands from:
```
face Name default,${undeclared}
```
to:
```
face Name default,
```

Currently, kak will not complain about the later syntax and just assign a default color to the background, but my patch changes this.

mawww:
Looks good, however I dont think I have a copyright waiver commit from you yet, and this is a rather sizeable contribution, so could you add one ?

lenormf:
Btw the issue that occurred originally with undeclared variables was fixed, by 49a5bbf3ca441e71cdeea376a2809b289efc5b33.

benoit-intrw:
@mawww Yes I have e1790bc891d2f78f733111354f8f23a3b42ac8e1

benoit-intrw:
@lenormf Great, and thanks  for the explanation, I forgotten for a moment that `%sh` block are interpreted by `/bin/sh`.

mawww:
@benoit-intrw Ohh right, I did not catch it as its with a different email address, thanks, merging.

benoit-intrw:
@mawww hum, you are right, thanks for noticing it !

PR #526, opened 2015-12-11
Ensure that at least one character follows a ',' or a '+' sign in a face

lenormf:
description (respectively a background color and attributes).


doppioandante:
Is there a way to write an unit test that should fail? It would be handy for this.

lenormf:
@doppioandante I wrote a couple tests for this case, but it needs the code in #527

PR #527, opened 2015-12-12
Allow unit tests to test for errors.

lenormf:
I would have made it so uncapitalized unit tests names test for failures and the rest behave like regular tests, but I don't think that would fare well with case insensitive filesystems (ntfs etc) so I went for the "-fail" suffix.

### DIFF
--- a/test/run
+++ b/test/run
@@ -15,11 +15,8 @@ main() {
     test_files=$(ls out selections state 2>/dev/null)
     cd $work/$dir;
     indent="$(echo "${dir}/" | sed -e 's|[^/]*/\+|  |g')"
-    name=$(echo ${PWD##*/} | sed 's/-fail^//')
-    should_fail=0
-    if [ ${PWD##*-} = fail ]; then
-      should_fail=1
-    fi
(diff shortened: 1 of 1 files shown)


test/run
```diff
@@ -15,7 +15,11 @@ main() {
     test_files=$(ls out selections state 2>/dev/null)
     cd $work/$dir;
     indent="$(echo "${dir}/" | sed -e 's|[^/]*/\+|  |g')"
-    name=$(basename $PWD)
+    name=$(echo ${PWD##*/} | sed 's/-fail^//')
```
lenormf:
```
name=$(sed 's,-fail,,i' <<< ${PWD##*/})
```

Had to change this because travis' shell couldn't handle it…

mawww:
I would prefer an additional file in the directory that would give the expected error. If not present, we consider the test should succeed, if present we consider it should fail (and maybe we can check we get the expected error, maybe by comparing stderr and the file content)

lenormf:
I'm running into an issue, cf #529, so this last commit won't work with failure tests.

PR #537, opened 2015-12-20
Use a wrapper for sigaction instead of signal

danlrobertson:
Use a wrapper based on the function in Figure 10.18 of [Advanced Programming in the Unix Environment](http://poincare.matf.bg.ac.rs/~ivana//courses/ps/sistemi_knjige/pomocno/apue.pdf) pg. 391. The benefits of which are explained in the text prior to the figure. I didn't see an instance of SIGALRM and didn't implement an error handler if `sigaction` failed.

### DIFF
--- a/src/event_manager.cc
+++ b/src/event_manager.cc
@@ -133,4 +133,14 @@ void EventManager::force_signal(int fd)
     FD_SET(fd, &m_forced_fd);
 }
 
+SignalHandler set_signal_wrapper(int signum, SignalHandler handler)
+{
+    struct sigaction new_action, old_action;
+
+    sigemptyset(&new_action.sa_mask);
+    new_action.sa_handler = handler;
+    new_action.sa_flags = SA_RESTART;
+    sigaction(signum, &new_action, &old_action);
+    return old_action.sa_handler;
+}
 }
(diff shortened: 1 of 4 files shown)


lenormf:
There's chances that `signal()` is being used because it's compatible with a window environment, which would make an eventual port easier.

mawww:
Hi, what is the purpose of that change, is that SA_RESTART flag important ?

danlrobertson:
> There's chances that `signal()` is being used because it's compatible with a window
> environment, which would make an eventual port easier.

Valid, and definitely something to consider.

> what is the purpose of that change, is that `SA_RESTART` flag important?

In this case the `SA_RESTART` flag is really the only big difference between `signal` and the `safe_signal` function in this patch. It ensures that when slower system calls like `read`, `wait`, etc are interrupted, they resume. Some older *nix based systems won't do this with the plain `signal`. They'll return without resuming and set `errno` to `EINTR`. Whether or not this is worth possibly making it harder to make an eventual port as @lenormf mentioned, I don't know :smile:

lenormf:
Since there's no windows port in sight, I suppose it wouldn't cost much to test/merge a PR.

PR #540, opened 2015-12-26
Find rightmost match in case of overlapping regex matches during reverse search.

occivink:
The kind of issue this fixes is especially apparent when typing a reverse-search of `.....` with incremental search enabled.


occivink:
Mmh looks like I didn't think this through quite enough

PR #542, opened 2015-12-28
Fix reverse search in case of overlapping matches.

occivink:
Hi,

There are some problems with the current reverse search strategy, mainly that it may miss potentially better matches that are overlapping. The added tests highlight some of these cases.
This probably introduces a performance hit in the case of numerous overlapping matches (e.g. `....`), but the most common case of literal search shouldn't be too affected. 

I wonder if it is possible to reverse a given regex and perform the match using reverse iterators? Performance would definitely be better.


mawww:
using reverse iterators would be ideal, but that would require a way to reverse regexes, which does not exists AFAIK.

The performance hit should not be big, reverse searching is already very innefficient.

Could you add a copyright waiver commit as well (see CONTRIBUTING) ? It does not look like you already have pushed one.

occivink:
>reverse searching is already very innefficient
Yes, it's unfortunate. Maybe the result from the search could be cached to avoid performing a search every time the previous match is selected.

PR #555, opened 2016-01-12
Add 'cfg' extension to 'ini-file' filetype

benoit-intrw:
'cfg' is another common file extension for ini like files. My use case is about
http://www.buildout.org/ config files.

### DIFF
--- a/rc/ini.kak
+++ b/rc/ini.kak
@@ -1,4 +1,4 @@
-hook global BufCreate .*\.(repo|service|target|socket|ini) %{
+hook global BufCreate .*\.(repo|service|target|socket|ini|cfg) %{
     set buffer filetype ini-file
 }
 


lenormf:
This extension is a pretty generic one, and doesn't necessarily implies that the content of the file will be following the INI format. Maybe other users are used to their `.cfg` files containing JSON, and will be inconvenienced by an arbitrary decision as to how kak highlights their files ?

benoit-intrw:
I agree that it is a common extension.

On vim side, it's quite complicated, you can find many cfg mentions, and not all are associated configuration files (see this file https://github.com/vim/vim/blob/master/runtime/filetype.vim)

The syntax files are here :
https://github.com/vim/vim/blob/master/runtime/syntax/dosini.vim
https://github.com/vim/vim/blob/master/runtime/syntax/cfg.vim

But I'm not the only one which associate cfg and ini, as a well know example there's the pygments syntax highlighter : http://pygments.org/docs/lexers/#pygments.lexers.configs.IniLexer

Anyway, it seems enough to create a cfg.kak file with something like : 
hook global BufCreate .*\.(cfg) %{
    set buffer filetype ini-file
}

That let me think `.gitconfig` and `.hgrc` are not highlighted and are also config like files.

mawww:
I think cfg as ini is a good default, better than any alternative I can think of at least.

Issue #556, opened 2016-01-15
Allow scripts to declare 'enum' or 'flags' variables


lenormf:

mawww:
Not sure its worth it, the current flags/enum implementation works well for c++ types, as it uses to type system directly. That would require quite a few changes for a small improvements.

lenormf:
It would allow custom scripts (such as the one in #557) not to declare a simple `str` typed variable, and enforce combinations of values from a given set. I can work around that, but it feels wrong to have such a lose type when we're only interested in a couple words pipe-separated.

PR #557, opened 2016-01-15
Python: default formatter, customizable static completion

lenormf:
Hi,

The first commit of this pull request simply sets `pythontidy` as the default formatter for python files.

The second commit is kind of a proof of concept, as I don't think the changes that it contains would be suitable for widespread adoption across language-support scripts. What it does is:
* Add a "Grammar" section to allow splitting the highlighted words into different variables (→ categories, i.e. keywords, types etc)
* Add a /sortof generic/ way of adding highlighters for those categories (`addhl` calls)
* Add a variable that acts like a `flags` type var (see #556) and whose role is to allow the user to chose what categories of keywords will be added to the `static_words` list
* Add a loop that generically adds the language's keywords to the `static_words` list, according to the contents of the variable mentioned above

The idea behind that second commit is that I don't necessarily want to have the entire grammar added to the statically completed words, but implementing that implies that the keywords are stored by category in the language's support script (and that snowballed into changing how those keywords are highlighted).

Cheers.


benoit-intrw:
pythontidy ? Never eared about it, do you mean this one https://pypi.python.org/pypi/PythonTidy/ ?

I personally know autopep8 (https://pypi.python.org/pypi/autopep8) and yapf (https://pypi.python.org/pypi/yapf). Both are maintained and autopep8 is available in Debian.

lenormf:
I'll be glad to change the default to whatever the python community uses the most.

lenormf:
Something I haven't mentioned: this PR also removes the hook that gets rid of trailing whitespace. I feel like this is less and less relevant with the introduction of code formatting, and it was also plain annoying to work around when trying to append a yanked selection to another line.

Example:
* we have a line written as such in a python buffer: `self.a =`
* we want to append something to this line, but also add a space after the `=` sign
* yank some data somewhere else in the file, i.e. `[1, 2, 3]`
* place the cursor on the `=` sign, hit `a` to insert a space
* hit the escape key to be able to hit the `p` key, and eventually insert the content of the "clipboard" into the current line
* the space immediately gets removed

Even there are workarounds around this issue, it was still annoying so I just suggest we get rid of this hook (which is also used in several other language support files).

jkonecny12:
Thank you for creating this PR. This will be really helpful to me in my work.

lenormf:
:+1:

mawww:
I would prefer if we avoid storing all that in options until we have a proper use case for it.

lenormf:
It's done. The code is much uglier (I wonder how we can eventually port that to other languages properly), and the callbacks execute more code but the options are not polluted. Also the static words are refreshed automatically whenever the `python_static_complete` variable is modified.

lenormf:
I managed to contain the code that handles static completion into a single file. It generates a kak function dynamically, which will be called from hooks to populate the `static_words` option.

lenormf:
As per the discussion on IRC, I removed all dynamic handling of static completion and added all the language's keywords into the `static_words` variable.

mawww:
Hi, could you squash all those changes in a single commit ?

lenormf:
Sure, assuming we don't need to keep the above comments anymore.

Issue #516, opened 2015-12-07
Decide on what scripts should be auto-loaded by default


lenormf:
Hi,

Let's trim the scripts currently loaded by default when no `autoload` directory has been created in the user's home directory. For that we need to decide what scripts should be taken out, and possibly come up with a checklist that would help decide if future scripts should be loaded by default.

lenormf:
Here is my list of scripts I think should be moved out:
* `asciidoc.kak`
* `cabal.kak`
* `coffee.kak`
* `cucumber.kak`
* `dockerfile.kak`
* `exherbo.kak`
* `fish.kak`
* `grench.kak`
* `haml.kak`
* `jedi.kak`
* `julia.kak`
* `kickstart.kak`
* `ragel.kak`
* `sass.kak`
* `scss.kak`
* `tupfile.kak`
* `yaml.kak`
* `scala.kak`

This would leave the following scripts in the default installation:
* `autorestore.kak`
* `autowrap.kak`
* `c-family.kak`
* `clang.kak`
* `clojure.kak`
* `commenting.kak`
* `css.kak`
* `ctags.kak`
* `diff.kak`
* `dlang.kak`
* `etc.kak`
* `formatter.kak`
* `gas.kak`
* `git.kak`
* `git-tools.kak`
* `golang.kak`
* `grep.kak`
* `haskell.kak`
* `html.kak`
* `ini.kak`
* `java.kak`
* `javascript.kak`
* `json.kak`
* `kakrc.kak`
* `lisp.kak`
* `mail.kak`
* `makefile.kak`
* `make.kak`
* `man.kak`
* `markdown.kak`
* `mime.kak`
* `perl.kak`
* `python.kak`
* `ruby.kak`
* `rust.kak`
* `sh.kak`
* `tmux.kak`
* `x11.kak`

I'm going to assume that the former list is going to be moved to a `kakoune-extra` repository, although @mawww hasn't decided anything yet AFAIK (I'm all about that idea btw, as a replacement for `scripts.html` on `kakoune.org`).

Now about how to decide which scripts to integrate into the main repo: if the file implements a functionality of the editor (`grep.kak`, `make.kak` etc) or if the file adds support for something (language, tool etc) that is objectively used by a lot of people that don't all belong to the same "field" (i.e. `sass.kak` is used mostly by web devs vs `etc.kak` used by anybody who runs a UNIX), then it should probably be kept.

doppioandante:
[...]

PR #561, opened 2016-01-21
Move some scripts to an `extra` repository

lenormf:
Hi,

Kakoune has gained a lot of traction lately, and more and more people contribute to the codebase on a regular basis. That's a great thing, however I feel like the `rc` directory (that contains scripts that implement features, highlight files/languages etc) has become bloated: everyone has a different use for their favorite editor, would like some things to be supported, and their contributions have been made available to everyone through that `rc` directory.

Kakoune is a minimalist editor and does not aim at becoming a full blown integrated development environment (the decision to let user make symlinks in their `autoload` nudged me into that direction anyway), and as such I propose that some scripts that provide support for arguably less popular/more esoteric /things/ be moved to an entirely seperate repository. This would have several advantages:
* prevent the distribution packages to ship lots of scripts that might not be wanted
* decentralize maintainance of 2nd category scripts (unless @mawww is fine with handling both)
* allow users that have custom scripts that are not elligible for a merge with the official repository to easily deploy them (who would be merged in the `extra` repository)

This commit moves a few files from the `rc` directory to another repository that would be available through the `rc-extra` submodule, which means that it would be as easy as updating said submodule and add symlinks to the wanted scripts in the `autoload` directory to have scripts autoloaded. Or even copy the scripts to the `autoload` directory. As for distribution packages, I suppose they could just copy all the scripts to `/usr/share/kak/rc`, and seamlessly allow users to symlink those files to their `autoload` directory.

Right now the `rc-extra` submodule points at a repository on my own :octocat:github profile, but the ownership of said repo can be transfered to anybody who wants to maintain it (I can do it, unless someone/@mawww wants it more).

See #516.


mawww:
Hello,

First thanks for working on that, however I am not sure about a few scripts.

* asciidoc should really be in base IMHO as we use it to edit the README, 
* jedi.kak and clang.kak both provide the same thing (one for c family, the other for python), so I'd expect to have them in the same repo as well (could be extra or base).
* scala is a quite popular language nowadays so I'd keep it in base.
* git-tools.kak could move in extra (git highlighting is in git.kak, git-tools provide deeper git integration)
* not sure what to do with autorestore.kak and autowrap.kak, they could move to extra IMHO.

Regarding where to put the extra repo, I think it would make sense to keep it in my github so that it keeps some kind of "official" stance, I'll be happy to give you push access to it.

What do you think ?

mawww:
@jkonecny12 how would that impact the packaging ? Can we easily have the scripts from extra distrubuted as well ?

lenormf:
Hi,

I am fine with those changes to the list I made. I will make the changes to the "extra" repository, and transfer it over to your account (I can help maintain it too).

As for package, I'll probably have my debian contrib script generate a "kakoune-extra" package as well, which I was thinking could just install all the files over in `/usr/share/kak/rc`.

gokcehan:
Just a few words for myself, I'm fine either way but I have some doubts especially regarding the exclusion of highlighting scripts. I think it would be a bummer for a user to see their favorite language needs some extra labor to work, especially for newcomers. Is there a performance reason for doing this?

I think it's good have concerns about bloating but I feel like kakoune is still far from it. Just for comparison these are some numbers from my machine:

    $ find /usr/share/kak/ | wc -l
    68
    $ find /usr/share/vim/vim74/ | wc -l
    1683
    $ find /usr/share/vim/vim74/syntax/ | wc -l
    587
    $ find /usr/share/vim/vim74/ftplugin/ | wc -l
    204
    $ ls -l /usr/share/vim/vim74/macros/  # what % of users make use of these?
    total 116
    -rw-r--r-- 1 root root  2443 Jan 13 22:08 dvorak
    -rw-r--r-- 1 root root  3706 Jan 13 22:08 editexisting.vim
    drwxr-xr-x 2 root root  4096 Jan 18 13:00 hanoi
    -rw-r--r-- 1 root root  9160 Jan 13 22:08 justify.vim
    -rw-r--r-- 1 root root   288 Jan 13 22:08 less.bat
    -rwxr-xr-x 1 root root   588 Jan 13 22:08 less.sh
    -rw-r--r-- 1 root root  5322 Jan 13 22:08 less.vim
    drwxr-xr-x 2 root root  4096 Jan 18 13:00 life
    -rw-r--r-- 1 root root 19303 Jan 13 22:08 matchit.txt
    -rw-r--r-- 1 root root 30734 Jan 13 22:08 matchit.vim
    drwxr-xr-x 2 root root  4096 Jan 18 13:00 maze
    -rw-r--r-- 1 root root  1100 Jan 13 22:08 README.txt
    -rw-r--r-- 1 root root  3922 Jan 13 22:08 shellmenu.vim
    -rw-r--r-- 1 root root   856 Jan 13 22:08 swapmous.vim
    drwxr-xr-x 2 root root  4096 Jan 18 13:00 urm

I for one certainly not use 200+ languages but it is still nice to have them when I need. Even with quite a few package managers out there they don't feel like separating these and I feel like this is a rightly call since there are still many vim users who prefer not to have any package managers at all for every piece of software they use. Do not underestimate the power of out-of-the-box experience :)

lenormf:
From my perspective, kakoune is a minimalistic code editor, it's not being promoted as being a full blown development environment, so basicall no-one will have an entirely compatible setup straight "out of the box". New users who understand the concept of minimalism will surely not be surprised by the fact that their `grunt` config file or brainfuck code is not being highlighted, and regardless one will only have to install an additional package. Although there is objectively a lot of support scripts missing for several things, the `extra` repo is bound to grow and over time will provide -ideally- as many things editors like vim provide while keeping minimalistic users happy.

Btw, by default, `find` will go down all the sub-directories of the given path, and will also list directories, so your numbers might be a bit off (although I understand your point, even though just because vim ships 10 million files by default shouldn't necessarily make it a project to look up to).

gokcehan:
Ok @lenormf , maybe it was a mistake giving vim as an example. I certainly don't regard it as a good model and if I wasn't stiving for minimalism as well I wouldn't be here trying kakoune anyways.

But minimalism is subjective. Consider this, there are people out there who argue against syntax highlighting, so kakoune should ship without any syntax highlighting at all? You would certainly do not respect their choice in that regard. What happens in reality is that these people usually can live with the fact that their editor is shipped with syntax highlighting. And this should be true for all. No matter how you divide these files there will be some extras people don't want to use from the original repo. Considering my setups with different programs, even in my "entirely compatible setup" I have some extras that I don't use often which constantly keep changing. In that regard, I don't think there is such a thing as "entirely compatible setup".

With my experience so far, my perspective for kakoune's minimalism is this. It does not reinvent the wheel for functionality that could be provided by external tools instead it provides a good way to interact with other programs and not provide a custom scripting language at all. This seems to me like the genuine "unix philosophy" and the most important thing that drives me here. I can live with some extra wrappers for tools even if I don't use them. Take `commenting.kak` for example. I find its behavior a little confusing so I'm not using it but it doesn't bother me at all since it does not assign a key to it by default. I think this is the important bit here, not to change the default behavior for people not using the functionality.

Let me materialize this discussion with some specific things that concerns me:

1. I fear that having a different repository for extras could easily become a place for people to submit their bloat. If I rely on something from extra, say `julia.kak` for julia editing, I'm gonna left with a choice either to live without syntax highlight or install lots of bloat. Which takes me to the second point..

2. By having extras, ultimately we're gonna need to come up with package managers like vim which I think is just an abomination. Text editor is just one of the softwares I use while coding. If I were to start managing packages for every piece of software I use on every platform I work on, I would only be doing configurations all day long.

3. Who's gonna decide which language or feature to go on to main or extra? What will be the criteria?

I think it's nice we both strive for minimalism but these are my concerns I would like to be addressed.

jkonecny12:
About the packaging to Fedora it should be ok. It could be created as special package or I can only add new Source and it would be same as now.

About the splitting I think we have 3 options now: 
1) To have all the scripts here as now - I don't think it's good idea many scripts will be doing magic outside of kakoune (I would like to create python script for better utilization of jedi which will be used from kakoune) and there will be more PRs changing these scripts then PRs for kakoune which seems somehow wrong to me.
2) Separate folder in repository and users can easily enable and disable modules as they want. This doesn't solve many problems from 1.
3) Separate repository. I fear this solution will be more confusing for people and also it will be really hard for packagers. I can create only the kakoune + main scripts package or create one package for all extra or for some extras or every package for every extra script? That is what I'm thinking now.

This leads me to interesting problem. Don't we want to have separate folder for these scripts. If they are extra and the others are main you don't want to mix them thought. Also linking the scripts from the /usr/share/kak/ now seems as something what users don't want to do. He want to create folder and give there his own scripts or so but linking files from the system directory... I don't know.

lenormf:
>I fear that having a different repository for extras could easily become a place for people to submit their bloat. If I rely on something from extra, say julia.kak for julia editing, I'm gonna left with a choice either to live without syntax highlight or install lots of bloat. Which takes me to the second point..
>By having extras, ultimately we're gonna need to come up with package managers like vim which I think is just an abomination. Text editor is just one of the softwares I use while coding. If I were to start managing packages for every piece of software I use on every platform I work on, I would only be doing configurations all day long.

We neither need nor want any vundle/pathogen clone, if a user just wants a couple files, they can just grab them off github and drop them in their `autoload` directory. If they don't want to be bothered, they can install the `extra` package (or clone the repo somewhere on disk), and make a symlink in their `autoload` directory. It's not a big change compared with what we have now.

>Who's gonna decide which language or feature to go on to main or extra? What will be the criteria?

This is purely subjective, and relies mostly on the objectiveness of contributors who want their script merged into the `extra` repository.

If the maintainer and the contributor can agree that the script (whether it implements a feature or highlights a particular format) could be objectively useful to more than just a handful of people, I suppose it could be added straight to the main repository. @mawww mentioned once that all scripts could be first added to the `extra` repository, and then upgraded to core if it turned out that a lot of users used it in the end (some sort of "proving ground") - but this needs to be discussed too.

> Separate repository. I fear this solution will be more confusing for people and also it will be really hard for packagers. I can create only the kakoune + main scripts package or create one package for all extra or for some extras or every package for every extra script? That is what I'm thinking now.
>This leads me to interesting problem. Don't we want to have separate folder for these scripts. If they are extra and the others are main you don't want to mix them thought.

The main idea behind having an `extra` repository is merely to prevent kakoune from shipping with a dozen scripts that will potentially not be used by most users. So from a packager's perspective, a "kakoune-extra" package would merely add some scripts back into the project, we don't need anything fancier.

>Also linking the scripts from the /usr/share/kak/ now seems as something what users don't want to do. He want to create folder and give there his own scripts or so but linking files from the system directory... I don't know.

Users already create symbolic links from `/usr/share/kak/rc` into their autoload directory, nothing has changed. If users want to have their own "autoload" folder somewhere on the system, they can implement it with a simple `%sh{}` scope in their `kakrc` file (I do it myself).

mawww:
So, lots of good points in all these comments, which got me thinking.

I am less convinced an extra repository is a good idea right now, one reason being that submodules are cumbersome to use, as we will need to update the commit they point to.

However there is one thing that could be done: organize rc/ into subdirectories, like rc/base and rc/extra (for example). That means we keep all the scripts autoloaded by default, but make it simpler for minimalist users to opt in for only the core scripts (just add a link to rc/base in your autotload directory).

I think organizing the rc/ scripts would make sense, and would document their status, there could be a rc/staging/ directory as well.

I think keeping most scripts loaded by default is the best way to go, then letting users opt-in to granular control of what scripts they want loaded makes sense.

I hope I make sense, jet lag is kicking in quite heavily right now.

jkonecny12:
@lenormf thank you for the nice explanation.
@mawww Good solution I like it. You have my :+1:

lenormf:
The point of this pull request is specifically to prevent all the scripts that are made by the community to be included in the repository, and there is already the `autoload` directory for users that want to selectively load scripts when kakoune starts.

I'm going to reiterate because this thread is headed into a spaghettis Q/A about issues that possibly don't have much to do with the PR.

Right now, kakoune ships with several scripts that are downloaded by all users. This is not an issue yet, because there's a fairly low amount of scripts and loading time is insignificant, but we can reasonably assume that the amount of scripts is only going to grow (as contributions come in), which is going -in a fairly distant future- to impact loading time.

Now users can **already** pick what scripts will be loaded, by creating symbolic links in their `autoload` directory, but that does not prevent users from all downloading things they might possibly not want, and that goes against the arguably minimalistic nature of this project.

So my solution is to allow users who don't want to bother with all of this to just install an extra package (and **nothing will change for them compared to how things are working now**), e.g. a package that installs al the scripts back into the `/usr/share/kak/rc` directory. Everything will be just the same as things are now, and looking back at this thread, I think most users fall into this category.

Users who do want to keep a minimal amount of scripts installed on their system will only have to install the main `kakoune` package, but will still be able to pick the scripts they want from the `extra` repository by e.g. downloading them off of github.

Having an `extra` repo will also allow contributors (like me) who do kakoune scripts for a lot of things but don't necessarily submit them (because they would only be useful to a couple of people and are really only related to a "niche technology") to have a way to still share those scripts with the rest of the community without forcing everyone to download/install it when they install kakoune. Example: I recently made a highlight script for `riotjs`, and I did not ask for it to be merged upstream because I'm sure no-one (in terms of community percentage) uses it:
* with an `extra` repository, someone who would need such a script for that "niche tech" would have a chance at not reimplementing it all over again, because there would actually be a place to check for that script first, *and* it would allow those not-so-popular scripts to be version-ed in a VCS
* without it, either everyone shares every script that was ever made for everything, or people have to reimplement things that they need even though it might have been made in the past

The one relevant question that was raised in the thread was about the criterias that will dictate which repository (core/extra) a script goes to, and that still **will not change compared to how things are working now** because @mawww is in charge of the `extra` repository. If you trusted him to take the right decision regarding whether or not to merge scripts into the main repository (that's where we're at now), then there's no reason not to trust him to do the right choice for the core/extra split.

gokcehan:
@lenormf #516 is closed with the note that all changes mentioned are added to core although `julia.kak` is still in extra if I'm not mistaken. Already two people mention julia in that thread so I don't think I'm being selfish for asking its inclusion in the core as it is not less popular than those languages included in the core.

+1 for the @mawww solution by the way.

mawww:
Seems Julia is getting some traction, so having it in the base would make sense.

@lenormf although I agree being minimalistic by default and opt-in bloat would be cleaner, I think on a more pragmatic note autoloading everything by default and making it easy to switch to a leaner base would be better.

by splitting rc/ in subdirectories, we keep the newcomers friendly "everything is autoloaded" while providing the advanced users a convenient way to select the minimalistic "base" set (just link rc/base in your autoload). We could eventually be even more granular (if it makes sense), with rc/core (tmux, x11, and very popular languages like C/C++/Java/Python/shell), rc/base (autorestore/, autowrap, common languages/tools, git support, ctags), rc/extra for the less ubiquitous languages/tools.

We could eventually have an extra git repository for in progress/really obscure things, but I don't think Julia/jedi/clang should be in that, things like riotjs/cucumber/grench could go there, although I'd say we should do that only once we get some reports that loading time is slow on the default install.

In other words, I think its better to ask the advanced users to setup their config (by autoloading whatever they want) than to ask newcomers to get a second repository if they want some extra bells and whistles.

lenormf:
>In other words, I think its better to ask the advanced users to setup their config (by autoloading whatever they want) than to ask newcomers to get a second repository if they want some extra bells and whistles.

Even though I agree with this, creating subdirectories in the `rc` directory is unrelated to the issue that this pull request is supposed to fix -an ever growing amount of scripts that will be shipped and installed on users' system regardless of what their use for kakoune is, and preventing users from recoding something that might have been made already.

Installing an additional package is the simplest possible way of doing this (and a consensus adopted by several other projects to split up core plugins/components/etc with the rest), we're going to end up a hundred scripts installed by default.

PR #562, opened 2016-01-21
Remove the scripts page, replaced by the `extra` repository

lenormf:
See #516 and #561.


benoit-intrw:
Replacing the gist link for a dedicated repo is a good idea.

But keeping a script page which explain how to contribute scripts is still useful.

PR #565, opened 2016-01-24
basic repl support for tmux and x11

gokcehan:
This makes it possible to send text to an interpreter from within kakoune (a.k.a. poor man's [SLIME](https://en.wikipedia.org/wiki/SLIME)). It sort of works but it's a little bit hackish so I guess it could be cleaned up by someone with more tmux and x11 knowledge than myself.

### DIFF
--- a/rc/tmux-repl.kak
+++ b/rc/tmux-repl.kak
@@ -1,0 +1,51 @@
+# http://tmux.github.io/
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+hook global KakBegin .* %{
+    %sh{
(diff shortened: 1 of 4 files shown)


lenormf:
Nice ☺

mawww:
Wow thats a nice feature, I think it would be nicer to have them in separate files though, as tmux-repl.kak and x11-repl.kak. Or is there a strong reason to keep them in the base files (with the rc splitting that @lenormf proposed, I expect the repl to go in extra, but x11.kak and tmux.kak to stay in base).

gokcehan:
Sounds alright @mawww. Shall I add some commits to separate the files and wait for @lenormf to merge?

lenormf:
Yes, ignore the split to `extra`.

PR #566, opened 2016-01-25
Add commands to select an HTML tag or its content

lenormf:
Hi,

This has been asked for, and I've often ran into the same need, so here it is. I would have made it so that the commands automatically select the parent tag, but it would require to parse the html in order not to falsely consider child tags as being the parent tag, i.e. `<tag><foo></foo>{cursor around here}</tag>`.

Usage: `:html-select-tag-outer tag` or `:html-select-tag-inner tag`

### DIFF
--- a/rc/html.kak
+++ b/rc/html.kak
@@ -60,6 +60,52 @@ def -hidden _html_indent_on_new_line %{
     }
 }
 
+def -hidden -params 1 _html_select_tag %{
+    ## Select everything till the closest opened bracket followed by the name of the tag
+    ## Select everything till the closed bracket followed by a slash sign and the name of the tag
+    ## Extend the selection to the next closed bracket (closing tag)
+    exec "
+        <a-?><lt>%arg{1}\><ret>
+        ; ?<lt>/%arg{1}\><ret>
+        ?><ret>
+    "
+}
+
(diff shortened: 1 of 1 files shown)


mawww:
I think this is superseeded by the recently pushed support for user defined text object, we now can use the whole set of object selection commands ([, ], <a-i>, <a-a>) with user given open/close strings, and respecting nesting correctly.

for example to select the containing div block, one could use: `<a-a>:<div><ret></div><ret>`

Is there some other features in that PR that are not supported ?

lenormf:
Indeed.

pickfire:
Should we have a section to do some commonly done but non-trivial tasks in the wiki for different languages family?

I was thinking of how to do:
```html
  <html>
    <p><em>Hello</em> world</p>
  </html>
```

How does one does these kind of manipulation?

- change `<p>` to `<h1>`
- add `<strong>` around `world`
- reformat `<p>` to its own line, leaving `<em>Hello</em> world` in one line indented
- remove `<p>`

The way I current can do is to do both `i` and `a` twice, typing the same thing again. I think I could try multi-selection but that does not handle the closing tag (`/`).

One know one of the trick above for vim which I found it on the internet, like `yitvitp` but kakoune does not have the selection for tags or the idea of extending selection, start and end tag does not count as a match.

Issue #256, opened 2015-03-24
Modeline parsing script


mawww:
A bundled script able to parse vim/emacs modelines and convert the options to kakoune ones would be nice.

PR #567, opened 2016-01-26
Add a modeline support script (vim + kakoune options)

lenormf:
Hi,

This patch addresses #256. As of now, the script only supports vim format modelines and detects vim/kakoune options (whitelist-oriented, options have to be added to the script to extend the list), as emacs formated modelines are a different format and wouldn't mingle well with the current parsing code.

Cheers.

EDIT: that would probably be an "extra" script, but I'm submitting it to the core in the meantime.

### DIFF
--- a/rc/modeline.kak
+++ b/rc/modeline.kak
@@ -7,8 +7,6 @@
 ## Only a few options are supported, in order to prevent the
 ## buffers from poking around the configuration too much
 
-# Set to 'true' to parse modelines in buffers
-decl bool modeline "true"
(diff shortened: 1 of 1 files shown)


mawww:
Hi,

finally took the time to take a look, instead of an option I'd rather let the user add the hook (only provide a `modeline-parse` command). The rest looks fine, especially using opt in for settable options.

Cheers.

lenormf:
Boom. It will need a `git mv` though, to place the script in the appropriate subfolder in `rc`.

PR #570, opened 2016-02-02
Get rid of tabs in the man pages to simplify highlighting, improve an…

lenormf:
…d document highlighting, get rid of the arbitrary binding

### DIFF
--- a/rc/man.kak
+++ b/rc/man.kak
@@ -2,11 +2,14 @@ decl str docsclient
 
 hook global WinSetOption filetype=man %{
     addhl group man-highlight
+    # Sections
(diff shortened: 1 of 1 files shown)


greduan:
:+1: from me, although please `--amend` that commit message. lol

Remember commit messages should not be over 50 chars long. http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

lenormf:
Actually the recommended width is 72 columns, and I don't follow those rules when the commit packs a few small changes that are not worth detailing.

greduan:
The commit message, the main commit message line, that's 50. Then the following lines are 72. :)

PR #571, opened 2016-02-02
remove Grench

alexherbo2:
I think this is superseded by [X11][rc/x11-repl.kak] and [tmux][rc/tmux-repl.kak] REPLs.

[rc/x11-repl.kak]:  https://github.com/mawww/kakoune/blob/master/rc/x11-repl.kak
[rc/tmux-repl.kak]: https://github.com/mawww/kakoune/blob/master/rc/tmux-repl.kak


greduan:
Leiningen is a Clojure(Script) lang REPL. Has nothing to do with the X11 and tmux REPLs.

Issue #503, opened 2015-12-02
Have a man page for the project


lenormf:
Hi,

Some time ago I had started writing a man page for the project (and had it installed by default, in the Makefile), but several recent changes would have me go over everything so I'm putting this on the back burner for now.

How about we generate the man page automatically from the `README` of the repository ? It could be done with `rst2man.py` or directly from `asciidoc`, after touching up the file a bit.

Cheers.

jkonecny12:
If I understand you right you want to convert the mega big README in git to man page? I don't think this is good idea. IMHO manpage should be only about what is the program, how to start it and description of console parameters which is maybe 5% of the README and I think in bad format.

jkonecny12:
Maybe it would be best give here the converted manpage we can then look on it and we will see.

lenormf:
It would take at least an hour of work to move everything from the README to a man page, the only sane way to do this is generate it.

jkonecny12:
Yeah I meant generated manpage of course I didn't want you to write it. 

Truth is it's better to have any manpage than nothing and I think we don't find somebody who have time to create written documentation and will synchronize it with newest features. 
So you have my vote for the generated manpage.

It will be really nice to have it in the Fedora package.

lenormf:
Done!

Issue #106, opened 2014-05-02
Add :doc command


alexherbo2:
```
:doc [section]…
```

to display documentation in the `*doc*` buffer, with `section` being either a file name (without extension) in doc/ or self documented command.  You can fill sections to access a subsection.

Examples

```
:doc
:doc map
:doc map global
:doc map global normal
```

lenormf:
This issue can probably be closed.

PR #572, opened 2016-02-02
Man page, `doc` command

lenormf:
Hi,

Addresses #503 and #106.

This pull request adds a man page to the `doc` directory of the project, the `Makefile` has been modified to automatically install it in the appropriate system directory. The man page does not list all the features of the editor, it sticks to command line arguments and files (the rest of the documentation is split in files handled by the `doc` command described below).

I also added a `doc` command which will display a manpage-formatted `*doc*` buffer about different "themes", mostly based off the section names of the project's `README`. The first argument of the `doc` command is complete-able, and the second one is an optional pattern that will be used to select lines that you're interested in (i.e. `doc commands write` will open the 'commands' documentation page and select all the lines containing the word "write").

I put most of the `README` file into those documentation pages, I edited some lines and shuffled some things around but it's still pretty much a WIP. It took me some time to format everything, but I think on the long run it is better to have formatted pager-friendly content rather than a single asciidoc document (on the long run, maybe we should consider getting rid of raw documentation in the `README` file, and have everything in the builtin documentation).

Some bits will have to be rewritten, I don't like the hardcoded path in `doc.kak`, the main man page in `doc/kakoune.1` needs to be complemented, and the documentation pages could use some more highlighting.

Cheers.


lenormf:
Imported commit ecfd22a.

lenormf:
All the documentation pages have been re-rewritten into the `asciidoc` format, from which man pages are generated and compressed (before being used by the `doc` command).

PR #575, opened 2016-02-03
Adapt 'make install' to rc/ reorganization

occivink:
It's a little ugly to have such repetition, but I don't think we can rely on brace expansion without specifying the shell.


lenormf:
Why wouldn't the shell complete the directory names ? Do you have an example of a case in which those lines you modifed won't install the files properly ?

occivink:
Sorry, what do you mean? install doesn't recurse through directories, it only copies file. At least that's what's happening on my system.

lenormf:
My bad.

PR #580, opened 2016-02-06
Add new assistants (owl, gnu, dilbert, tux)

lenormf:
For fun.


mawww:
Hi,

nice ones, two things: assistants are a ncurses ui "easter egg" (clippy used to be hidden), I dont think they would make sense in other UI (non terminal ones) so I'd prefer to keep them in the ncurses_ui.hh file, or alternatively in a ncurses_ui_assistant.hh.

Another thing is copyrights, can the various asciiart added by released under public domain ?

lenormf:
The `assistant` namespace was moved to `ncurses_ui.hh`.

About the ascii art, I all found them on [this gallery](http://www.chris.com/ascii/). I haven't seen any license or copyright that would blanket the entire site, so I just kept the site's address and the author's name when specified as a comment in the code.

gokcehan:
Dilbert looks quite nice :) but I don't think it is a good idea to add these to the codebase. It seems like a rather personal choice so I don't think there will ever be enough of these. Maybe a configuration option could be added for this purpose but I don't think it is a good idea either.

I think the clippy is goofy enough the way it is (I once read some guy thought kakoune was an April's fool joke because s/he saw the clippy in the screenshot :p).  Even the cat option seems redundant to me but I guess it could be nice to have another option for people who wants to get rid of the Microsoft reference.

mawww:
I cannot merge that due to licensing problems, I am afraid we dont know the license for those ascii arts so we cannot put them in the public domain (and if they are not in the public domain, we cannot put them there by our selves anyway).

PR #581, opened 2016-02-07
Update rust.kak to post-1.0 rust

occivink:
Hi,
The rust syntax highlighting was quite outdated and predated the 1.0 release so I've cleaned it up a bit.
I've considered adding highlighters for basic types from the standard library (Option, Result, Vec...)
but it got a bit too colorful, and it seems like you don't usually highlight things related to standard library so I've left it out.
The number literals regex has one false negative that I know of, which are the float ending in '.' (so '1.' or '01_.') but I suppose it's acceptable. The word boundary matching makes this awkward to handle.
I've removed a *lot* from the 'identifier' group, but it contained a lot of unrelated things (which were also not identifiers) and it didn't provide much value in my opinion.
Since @alexherbo2 introduced this and @jkonecny12 also committed to it, I would like to have their input.

### DIFF
--- a/rc/base/rust.kak
+++ b/rc/base/rust.kak
@@ -19,19 +19,22 @@ hook global BufCreate .*[.](rust|rs) %{
 addhl -group / regions -default code rust \
     string  '"' (?<!\\)(\\\\)*"        '' \
     comment //   $                     '' \
-    comment /\* \*/                   /\* \
-    macro  ^\h*?\K# (?<!\\)\n          ''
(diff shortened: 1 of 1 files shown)


FreeFull:
I haven't actually seen any Rust code that makes use of `1.` instead of `1.0`. I actually thought it wasn't even valid syntax until I tested it in the playpen just now. Judging by this, the limitation of the regex shouldn't prove to be much trouble in practice.

PR #582, opened 2016-02-08
Makefile: Add a hook that inserts tab characters in the implementations of targets

lenormf:
Hi,

As mentioned by @somasis on IRC, it can be annoying to have to insert tabs in the right places when you indent with spaces.

This commit automatically replaces `tabstop` amount of spaces with a tab character in Makefile rules.

HTH.

### DIFF
--- a/rc/base/makefile.kak
+++ b/rc/base/makefile.kak
@@ -28,12 +28,8 @@ addhl -group /makefile/content regex [+?:]= 0:operator
 
 def -hidden _makefile_fix_whitespaces %{
     try %{
-        exec -draft "
-            \%
-s^[^\s][^:\n]+:[^\n]*\n(^\h+[^\n]+\n?)+<ret>
-s {%opt{tabstop}}<ret>
-c<tab><esc>
-        "
+        exec -draft <esc>K <a-k>^\S[^:\v]*:[^:=]<ret>
+        exec <tab>
     }
 }
 
@@ -43,7 +39,7 @@ c<tab><esc>
 hook global WinSetOption filetype=makefile %{
     addhl ref makefile
 
-    hook buffer BufWritePre .* -group makefile-indent _makefile_fix_whitespaces
+    hook window InsertChar \n -group makefile-indent _makefile_fix_whitespaces
 
     set window comment_selection_chars ""
     set window comment_line_chars "#"
@@ -51,5 +47,5 @@ hook global WinSetOption filetype=makefile %{
 
 hook global WinSetOption filetype=(?!makefile).* %{
     rmhl makefile
-    rmhooks buffer makefile-indent
+    rmhooks window makefile-indent
 }
--- a/rc/base/makefile.kak
+++ b/rc/base/makefile.kak
@@ -23,13 +23,33 @@ addhl -group /makefile/content regex ^[\w.%]+\h*:\s 0:identifier
 addhl -group /makefile/content regex \b(ifeq|ifneq|else|endif)\b 0:keyword
 addhl -group /makefile/content regex [+?:]= 0:operator
 
+# Commands
+# ‾‾‾‾‾‾‾‾
+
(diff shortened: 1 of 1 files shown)


lenormf:
Removed the hook that modifies the buffer when it's saved, rebased against `master` too.

gokcehan:
What do you mean by 'when you indent with spaces'?

greduan:
@gokcehan You use spaces instead of tabs to indent.

gokcehan:
Sorry @greduan, I should have been more clear. What I meant to ask is how to insert spaces instead of tabs. If I'm not mistaken there is no equivalent setting of vim's 'expandtab' option in kakoune, so the only way I can think of is to press the space key multiple times. If that's the case do we really want to fix it?

greduan:
@gokcehan there is a setting, I seriously don't remember which one it was though, let me check... The following are in my `kakrc`:

```
set global indentwidth 2
set global aligntab false
```

gokcehan:
@greduan I tried those settings and a tab still inserts a tab character. I think aligntab option stands for something else.

lenormf:
```
## width of a tab character
set global tabstop 4

## width (in spaces) used for indentation, 0 means a tab character
set global indentwidth 4
```

Indenting with the `>` key will insert spaces with the above configuration options, and that's an issue when "implementing" a Makefile target because those expect a tab character for indentation.

gokcehan:
@lenormf that sounds like a rare case, and also I don't think this PR targets that use case either. How about adding a hook that sets the indentwidth to 0 for makefiles with something like:

    hook global WinSetOption filetype=makefile %{
        set buffer indentwidth 0
    }

And I'm not sure if this belong here, people can use this in their configurations if they like. In any case I think it's programmer's responsibility to know makefiles only works with tabs but not spaces, so I don't think it's a good idea to sneakily do their jobs behind them. Better not be too smart.

lenormf:
Some users map the tab key to `>` to handle indenting more easily than when using `>` directly, which makes it difficult to insert a literal tab character without another mapping.

Spaces are allowed in other parts of a Makefile, so setting `indentwidth` to 0 just for the sake of not inserting tabs under targets is overkill. You can also disable indenting hooks if you don't want the editor to insert the right characters at the right places for you.

gokcehan:
Again @lenormf;

- It sounds like a rare case,
- Your patch does not solve the issue you mention,
- People can add this to their configurations if they like such a custom behavior. I don't think it is in the best interest of the majority to do this by default.

In addition;

- Those users who map tab key to `>` can still use tabs in insert mode or I just don't understand what you mean. Are you one of these users?
- I tried your additions and it does not work as expected since it assumes all rules have only a single body line. Even in the kakoune's own makefile there are rules with multiple lines. This is just too specific to be ever useful.

I say we better leave it as is.

lenormf:
Indenting with spaces is not convenient when declaring targets in Makefiles, so the hook in this PR inserts tabs where they should be. Just like the `C` or `javscript` support scripts automatically indent after hitting the return key.

gokcehan:
This fails miserably @lenormf, try this:

	all:<1>      # hit enter
		<2>cc..  # hit enter
		<3>cc..  # hit enter and then bs
	<4>          # hit enter
		<5>      # and I'm here

Even if you fix this it will still not work for many other cases (consider if's and for's in rules). If you're going to insist on this idea at least participate in the discussion. You're not responding to any of the points mentioned above and to me just trying to fix a non-existent issue.

lenormf:
Thanks for your exceptionally thorough tests of the PR, keep it up.

gokcehan:
Seriously @lenormf? It literally took me 5 seconds to find it. Have you
even test it yourself once?

And I certainly won't 'keep it up' because I'm still not convinced why this
is necessary. It is very difficult to stay rational with you when you're
trying to bully your ideas around. Good luck with what you do..

lenormf:
Thanks, you too!

PR #585, opened 2016-02-09
man.kak: Remove ascii codes from man output

somasis:
Some implementations of man do not output pages stripped of color
codes and such, even if the stdout isn't a tty.


lenormf:
What implementation of `man` doesn't strip color codes ? Also `doc.kak` will need some modifying, as it's using `man` too.

somasis:
I can confirm that busybox's man doesn't strip color codes.

lenormf:
Shouldn't you fix the bug in busybox's `man` then, since this is clearly a bug on their side ?

mawww:
I agree with @lenormf on this one, it really looks like a busybox bug, or maybe there is a way to force it not to generate the ansi sequences.

somasis:
Okay, I'll take a look at fixing busybox instead.

PR #591, opened 2016-02-14
Cache reverse search results

occivink:
The idea is to restart the search at the position of the best known match. The forward search is not cached, because it may lead to subtly wrong results.
This doesn't help the first reverse search, but greatly speeds up subsequent reverse search of the same pattern.

### DIFF
--- a/src/buffer.cc
+++ b/src/buffer.cc
@@ -636,43 +636,6 @@ String Buffer::debug_description() const
                   content_size, additional_size);
 }
 
-void Buffer::MatchesCache::insert(ByteCoord end, ByteCoord start)
-{
-    matches.insert(std::make_pair(end, start));
-}
-
-Optional<ByteCoord> Buffer::MatchesCache::preceding_match_start(ByteCoord before) const
-{
-    auto it = matches.lower_bound(before);
-    if (it != matches.begin())
-        return Optional<ByteCoord>((--it)->second);
-    return Optional<ByteCoord>();
-}
-
-Buffer::MatchesCache& Buffer::get_regex_cache(StringView regex) const
-{
-    size_t stamp = timestamp();
-    if (StringView(m_cache.regex) != regex)
-    {
-        m_cache.matches.clear();
-        m_cache.regex = regex.str();
-        m_cache.valid_at = stamp;
-    }
-    else if (m_cache.valid_at < stamp)
-    {
-        auto changes = changes_since(m_cache.valid_at);
-        auto min_change = std::min_element(changes.begin(), changes.end(),
-                                    [](const Change& lhs, const Change& rhs)
-                                    {
-                                        return lhs.begin < rhs.begin;
-                                    });
-        auto& map = m_cache.matches;
-        map.erase(map.lower_bound(min_change->begin), map.end());
-        m_cache.valid_at = stamp;
-    }
-    return m_cache;
-}
-
 UnitTest test_buffer{[]()
 {
     Buffer empty_buffer("empty", Buffer::Flags::None, {});
--- a/src/buffer.cc
+++ b/src/buffer.cc
@@ -636,6 +636,43 @@ String Buffer::debug_description() const
                   content_size, additional_size);
 }
 
+void Buffer::MatchesCache::insert(ByteCoord end, ByteCoord start)
+{
+    matches.insert(std::make_pair(end, start));
+}
+
+Optional<ByteCoord> Buffer::MatchesCache::preceding_match_start(ByteCoord before) const
+{
+    auto it = matches.lower_bound(before);
+    if (it != matches.begin())
+        return Optional<ByteCoord>((--it)->second);
+    return Optional<ByteCoord>();
+}
+
+Buffer::MatchesCache& Buffer::get_regex_cache(StringView regex) const
+{
+    size_t stamp = timestamp();
+    if (StringView(m_cache.regex) != regex)
+    {
+        m_cache.matches.clear();
+        m_cache.regex = regex.str();
+        m_cache.valid_at = stamp;
+    }
+    else if (m_cache.valid_at < stamp)
+    {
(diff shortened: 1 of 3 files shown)


mawww:
Thats a good idea, instead of adding the cache to the buffer, can you use the buffer values (see the get_word_db function for a usage example, its a way to attach arbitrary data to a buffer). That way the cache struct can be declared directly in selectors.hh.

Also, avoid std::map, use either a unordered map or a sorted vector (seems a sorted vector is what you need here).

occivink:
Should be good now. Why no std::map though? Is it a performance concern?

mawww:
Could you move the BufferRange struct from display_buffer.hh into buffer.hh and use that instead of Match ? also, do we need to reexec the regex_search ? cant we rely directly on the cached matches ?

occivink:
>also, do we need to reexec the regex_search ? cant we rely directly on the cached matches ?
Since find_next_match() needs the captures from the results, we could either store the entire MatchResults in the cache and pick the last valid, or only the main range and reexec the regex_search().
I'm starting to realize that the second option (which also means the current state of the PR) does not work if the regex contains lookbehind or '\k' assertions, since the search restarts right after the lookbehind part.
The first option involves storing `BufferIterator` throughout changes to the Buffer, which I don't think is valid. If not, we could invalidate all cached results as soon as any change happen. I think that would be the best solution given the constraints.

mawww:
invalidating the matches whenever the timestamp changes seems like a good option, the use case we want to optmize is cycling through matches in reverse direction, which does not typicaly happen in parallel with buffer changes.

occivink:
Indeed, I'll give it a shot.

occivink:
I think this should do it. I couldn't really help moving the cache logic into the function.
Lookbehind now work properly as they did before. A `MatchResults<BufferIterator>` is much bigger than only its boundaries at around ~300 bytes but this should only be noticeable with _a lot_ of matches.
The cache is now entirely invalidated whenever the buffer is changed. I kind of liked the idea of only invalidating partially as needed, but that wasn't really the point.

lenormf:
Have you tested your code with a non-debug build on a massive file (i.e. a 100MB+  log file), and made a multiple selection with several matches ? Maybe you can time some operations to highlight the improvements that this PR propose ?

occivink:
Good point. On a file of 3,795,570 lines and with a search that has 394,620 results, this brings the time that a reverse search (after the first) takes near the bottom from ~7s to instantaneous (and uses an additional 140mb of memory).

I'll bring up some more detailed numbers later.

occivink:
Also right now the cache can grow arbitrarily large (I've had it use 40GB of memory with 72M matches...) but it could easily be limited to an arbitrary size too.

lenormf:
I think they are a lot of corner cases in which `kakoune` will go through all the available RAM, given the right regex and input data size, so don't worry about those extreme cases.

However I'd be nice to have an idea of how far we can go with this caching thing, and make sure it won't be too much of a burden when dealing with large files and reasonably complex regex.

mawww:
Hum, first we could replace MatchResult<BufferIterator> with a Vector<BufferRange> which should be lighter (at least we get rid of all the backpointers to the buffer). We could even just store the capture count, and hence only have a single Vector that stores all the matches (match 0 is element 0 .. element num-capture -1, match 1 is element num-capture .. 2*num-capture-1 etc...). That should make the memory usage lower.

To measure properly the memory useage, you could add a new memory domain in memory.hh (ReverseMatchCache ?) and use that for the cache data (Vector second template argument is the memory domain). you can then get memory statistics written in the debug buffer with the `:debug memory` command.

occivink:
How about a `Vector<BufferRange>` to hold the match and submatches, instead of the MatchResults? 
By the way, what are the submatches used for?

lenormf:
>By the way, what are the submatches used for?

I would think for placing the cursor on the submatch whose index is passed through the selection param.

E.g. given a buffer containing `abc 123`, executing the following `2` `s(\w+) (\d+)<ret>` would place the cursor on the `1` of `123` (the parameter passed to the selection primitive is the number of the submatch).

occivink:
I don't think that's the case, the parameter passed to search or reverse search is how many times the search is executed i.e. (parameter-1) results get skipped.

lenormf:
Not sure how much sense that makes, and how it would be useful to execute a search multiple times.

occivink:
>We could even just store the capture count, and hence only have a single Vector that stores all the matches

I'm kind of stumped with that approach due to the lack of stride support in STL's iterators. Would it be reasonable to have a reimplementation of std::lower_bound (and upper_bound but that's just a matter of using <= instead of <) with a stride equal to the capture count just for that purpose? Or is there some other solution I'm not seeing?

occivink:
Much to my sadness, I've realized that some regexes are subtly broken by caching (or at least all forms I've thought of) so I'll be closing this. Sorry.

mawww:
Out of curiosity, what kind of regexes gets broken by caching ?

occivink:
Basically all variable-length patterns that can end before or after the current selection don't play nicely
For example take the input
`aaa
b`
and the regex `a+`
if you start the search from the end, the cache will get these potential matches
(0,0)->(0,2)
(0,1)->(0,2)
(0,2)->(0,2)
and select the first one. If you now put the cursor on the last 'a' at (0,2) and reverse search again, you would expect (0,0)->(0,1) to be selected, but the cache gives you (0,0)->(0,2) as the best match.
You cannot reexecute the search as I initially implemented because of lookarounds and `\k` assertions.

In the end I think the only way to get efficient reverse search would be to switch to a regex library that is implemented in a reversible way (with NFA for example). But this would also mean losing lookarounds and backreferences which inherently make regexes non-regular, and are used by some .kak files. But since this hypothetical library also has to take iterators as input and ideally supports UTF8, I don't think it exists, so it's kind of a moot point.

PR #593, opened 2016-02-17
Buffer padding

lenormf:
Hi,

The following commits add options that control how buffers are padded in a window (the dead space between the last line of a file and the bottom side of the window, currently filled with blue `~` signs).

The user can now chose to add a single string of his choosing after the last line of the buffers, fill the dead space with that same string, or do nothing.

There's also a `BufferPadding` face that comes with it, applied on the padding string (as of now, it's the same blue on black `~` sign, and that doesn't fare so well with certain themes).

I didn't think much about the names, maybe someone will think of something more appropriate.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -920,6 +920,7 @@ There are some builtins faces used by internal Kakoune functionalities:
  * `Prompt`: face used prompt displayed on the status line
  * `MatchingChar`: face used by the show_matching highlighter
  * `Search`: face used to highlight search results
+ * `BufferPadding`: face applied on the characters that follow the last line of a buffer
 
 Advanced topics
 ---------------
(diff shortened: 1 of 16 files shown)


mawww:
I am not keen on passing another face to draw for just that use case. And I am not sure its really important to be able to configure that, beside the fact that any (hypotetical, I know) additional ui will need to add its own options to control that.

lenormf:
Well it's impossible to know how the characters should be displayed because that depends on the theme, chosing an arbitrary face (currently, `blue` on `default`) does not go well at all on some default themes (`solarized` for example). At the same time we can't use the default face because those characters are not part of the contents of the file.

As for controlling the behaviour, I don't necessarily want to fill my entire window with characters, particularly when I don't have the line numbers highlighter enabled, so having just one (or none) is nice.

PR #594, opened 2016-02-17
Make smart search ignore case when the pattern doesn't contain capita…

lenormf:
…lized characters

### DIFF
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -667,6 +667,12 @@ void search_next(Context& context, NormalParams params)
 template<bool smart>
 void use_selection_as_search_pattern(Context& context, NormalParams params)
 {
+    auto has_upper =  [](const String &s) -> bool {
+        for (CharCount i = 0; i < s.char_length(); i++)
+            if (std::iswupper(s[i]))
+                return true;
+        return false;
+    };
     Vector<String> patterns;
     auto& sels = context.selections();
     const auto& buffer = context.buffer();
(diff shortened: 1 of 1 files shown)


mawww:
I still wonder if we want this behaviour, I do not want an option for that, and I am not sure it should be the what `*` does, but seems many people want it. Any additional data points / opinions ?

lenormf:
The documentation does mention "smart searches", so "smart casing" seems to be in order.

I can't think of a case where I'd like word boundaries with case-sensibility on a lowercase word, but if that happened I could still modify the pattern in the `/` register manually, so we might as well make the smart search as smart as possible.

By the way having a flags option that controls this behaviour could be nice, why not do just that ? The more granularity on a software's "complex" behaviours, the better imo.

mawww:
most languages I use are case sensitive, so when I search for the identifier under the cursor (which is by far my most common use case), I do want case sensitive search, and word boundaries.

I am generally against any option that control behaviour of normal mode commands except very specific cases (tabstop) as it means scripts cannot expect a certain behaviour anymore.

lenormf:
The implementation of `*` is also a particular case too, though. It's more of an interactive thing than a feature that would be used by scripts imo, considering those can use plain selects and apply the exact regex they need. I understand the position you're in, but I think an option is the best compromise between the two sides.

Users who need this could also bind the smart case search to `<user>*`, it's not very difficult to do with an `exec`.

PR #631, opened 2016-03-14
Misc: Add static completion suggestions to several languages

lenormf:
This commit also sets `autopep8` as the default python formatter.

I haven't modified the files that support languages I don't use, it was tedious enough to modify the ones in `core` and `base`.


lenormf:
Turns out here-strings are not posix either, so I replaced them with `printf` calls and a pipe to `tr`. Should be good to go now.

PR #632, opened 2016-03-15
POSIX changes

lenormf:
A work in progress branch that aims at making shell scripts POSIX-compliant.

Some references:
http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html
http://www.etalabs.net/sh_tricks.html

### DIFF
--- a/rc/base/ctags.kak
+++ b/rc/base/ctags.kak
@@ -7,7 +7,7 @@ decl str-list ctagsfiles 'tags'
 
 def -params 0..1 \
     -shell-completion '
-        ( for tags in $(printf %s "${kak_opt_ctagsfiles}" | tr \':\' \'\n\');
+        ( for tags in $(printf %s\\n "${kak_opt_ctagsfiles}" | tr \':\' \'\n\');
               do readtags -t "${tags}" -p "$1"
           done ) | cut -f 1 | sort | uniq' \
     -docstring 'Jump to tag definition' \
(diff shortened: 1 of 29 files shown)


rennis250:
This raises a recent question that I have had. Is the "-r" option of sed a POSIX standard? This option is depended upon in a number of kak scripts. However, the default sed binary on OS X does not recognize that option and fails.

mawww:
According to http://pubs.opengroup.org/onlinepubs/9699919799/utilities/sed.html sed does not support the -r option, so we should get rid of that.

lenormf:
Yes, raw strings that don't contain any backslash or flags should be fine to use with `echo`, I just replaced them all out of lazyness.

lenormf:
@rennis250 GNU tools don't fare well with their BSD equivalents, in some cases. Speaking about `sed` specifically, I think we should be able to use kak instead, that way the scripts would be compatible.

mawww:
Whats the status of this PR ?

lenormf:
Still wondering whether it's good to replace all calls to `echo` with `printf %s\\n`, or actually think about what's passed to `echo` and leave it be. I replaced everything in the original commits, but maybe I've traded readbility for consistency where it could have been avoided.

I'm going to rebase this whole thing against master and remove the `WIP` tag in the title once it's ready to be merged.

mawww:
I'd be in favor of keeping echo fo clearly valid cases.

lenormf:
OK I think that will be all for this PR, I've rebased on master and I'm ready to squash all the commits.

Fun fact: for my tests I changed the default shell invoked by kak with `posh` (https://packages.debian.org/sid/posh), which helped me find a mistake in the `colorscheme` function. Maybe there should be an option that holds the path to the shell that would execute shell scripts, so that we don't need to recompile to check that scripts are POSIX-compliant ?

PR #636, opened 2016-03-16
update rc/ruby.kak

alexherbo2:


lenormf:
Would it be possible to get rid of or declare regex regions as a simple `regex` ? The highlighting gets in the way whenever using the division operator, and messes big chunks of the files.

PR #639, opened 2016-03-23
Regex: add support for unicode regular expressions in boost

lenormf:
Hi,

This PR modifies the `Regex` abstration to make it inherit from unicode compatible boost regular expressions. Keeping everything somewhat compatible with `std::regex` was difficult, I did what I could to factorize the code while keeping it clear. I forwarded all the calls to `regex_match` and `regex_search` in the original code to intermediate functions (with the same names) in the `Kakoune` namespace, who then call the std/boost equivalent accordingly.

I also had a weird compilation error on a call to `read`, which happened when I tried to compile with `KAK_USE_STDREGEX`, even though it was compiling just fine without that macro defined.

I haven't tested things very thoroughly, I just checked that simple search/select patterns were working properly with simple and unicode strings.

HTH.


lenormf:
Although this branch compiles on a regular mac system, I'm waiting on travis [1] to have an `icu` ready version of `boost` installed by default, so that the code can be linked properly.

[1] https://github.com/travis-ci/travis-ci/issues/5840

lenormf:
Fixing the compilation issues on travis have proven to be a time sink (even though this branch works perfectly fine on a regular mac, it won't compile because of missing dependencies on travis), and we're more and more headed towards an all around boost-less version of kakoune. I'm closing this for now, maybe somebody will want to have a boost-icu-ready version despite the move to std ?

PR #647, opened 2016-04-08
Expose cursor_byte_offset env variable

pauldub:
Expose a new environment variable `cursor_byte_offset` which is the position of the cursor in bytes from the start of the buffer.

It can be used to integrate third party tools like `godef`, for example:

```
def -allow-override godef %{
	%sh{
		output=$(mktemp -d -t kak-temp-XXXXXXXX)/fifo
		mkfifo ${output}
		(godef -t -f ${kak_buffile} -o ${kak_cursor_byte_offset} > ${output}) > /dev/null 2>&1 < /dev/null &
		result=$(cat $output | grep -v '^/')
		echo "info -anchor ${kak_cursor_line}.${kak_cursor_column} -placement above %{ $result }"
        }
}
```

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -115,6 +115,11 @@ void register_env_vars()
             { auto coord = context.selections().main().cursor();
               return to_string(context.buffer()[coord.line].char_count_to(coord.column) + 1); }
         }, {
+            "cursor_byte_offset", false,
+            [](StringView name, const Context& context) -> String
+            { auto cursor = context.selections().main().cursor();
+              return to_string(context.buffer().distance(context.buffer().begin().coord(), cursor)); }
+        }, {
             "selection_desc", false,
             [](StringView name, const Context& context)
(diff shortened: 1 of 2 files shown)


src/main.cc
```diff
@@ -115,6 +115,11 @@ void register_env_vars()
             { auto coord = context.selections().main().cursor();
               return to_string(context.buffer()[coord.line].char_count_to(coord.column) + 1); }
         }, {
+            "cursor_byte_offset", false,
+            [](StringView name, const Context& context) -> String
+            { auto cursor = context.selections().main().cursor();
+              return to_string(context.buffer().distance(context.buffer().begin().coord(), cursor)); }
```
mawww: buffer begin coord is always going to be {0,0}

pauldub: So I can use `ByteCoord(0, 0)` ?

mawww: Yeah, even just `{0,0}`

PR #650, opened 2016-04-09
Add dash and underscore text objects

Delapouite:
Hi

Those text objects are quite useful when dealing with:
- `kebab-case-variable-name` as it's usually the case in LISP languages
- `snake_case_variable_name` as found in Ruby or Markdown for example

The `d` letter stands for **d**ash, the letter u was already taken, so `l` stands for **l**ow dash


mawww:
That has been superseeded by the general punctuation handling (any punctuation character thats not recognized as a special pair will be treated as an object starting/ending by that punctuation char)

PR #652, opened 2016-04-12
Change ruby-insert group to ruby-indent to be consistent with convention

jjthrash:


mawww:
In the end, I think we are going to split the other indent between basic indent (whitespaces changes) in the`<lang>-indent` hook group, and more advanced autoedit (auto close parens, auto insert various stuffs) in `<lang>-insert` so that users can more easily disable one or the other.

jjthrash:
Works for me.
 
On Wed, Apr 13, 2016, at 08:44 AM, Maxime Coste wrote:
> In the end, I think we are going to split the other indent between
> basic indent (whitespaces changes) in the<lang>-indent hook group,
> and more advanced autoedit (auto close parens, auto insert various
> stuffs) in <lang>-insert so that users can more easily disable one or
> the other.
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub[1]
>
 

Links:

  1. https://github.com/mawww/kakoune/pull/652#issuecomment-209412049

PR #653, opened 2016-04-12
Flake8 diagnostics support

Kareeeeem:
This follows the clang diagnostics example and does the same thing. It
just uses the flake8 program and the regex's are a bit different.


mawww:
Hi, seems you have some tabs in that, which we dont use in kak scripts. Can you as well add a copyright waiver commit as this is your first contribution (see the CONTRIBUTING file)

Kareeeeem:
Done.

PR #671, opened 2016-05-11
[WIP] Reload the buffer on terminal resize

lenormf:
Hi,

This PR is supposed to do what the title says, except for now it's crashing.

I'll be waiting for a fix to continue working on that, and eventually add "auto reload" to the `:doc` command as well.

### DIFF
--- a/rc/core/man.kak
+++ b/rc/core/man.kak
@@ -1,5 +1,8 @@
 decl str docsclient
 
+decl -hidden str _man_page
+decl -hidden str _man_subject
+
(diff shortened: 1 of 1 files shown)


lenormf:
The original issue has been fixed, unfortunately the code of this PR has a weird side effect that I don't really know where it comes from: the buffer is emptied out completely when the window is resized.

lenormf:
We need this reopened, as it might be implenting a feature we want (parsing of the page name/number seperately and calls to the high level `man` function), and it's also possibly a testcase for a bug.

PR #684, opened 2016-05-25
Fix the `tag` function error handling and tag lookup

lenormf:
This should work, although I still want to remove the whole thing.

### DIFF
--- a/rc/base/ctags.kak
+++ b/rc/base/ctags.kak
@@ -16,21 +16,24 @@ def -params 0..1 \
         export tagname=${1:-${kak_selection}}
         tag_found=0
         for tags in $(printf %s\\n "${kak_opt_ctagsfiles}" | tr ':' '\n'); do
-            export tagroot="$(readlink -f $(dirname "$tags"))/"
             tagdata=$(readtags -t "${tags}" "${tagname}" 2>/dev/null)
             if [ $? -ne 0 ] || [ -z "${tagdata}" ]; then
                 continue
             fi
+            export tagroot=$(readlink -e $(dirname "$tags"))'/'
             printf %s\\n "${tagdata}" | awk -F '\t|\n' -e '
                 /[^\t]+\t[^\t]+\t\/\^.*\$?\// {
                     re=$0;
                     sub(".*\t/\\^", "", re); sub("\\$?/$", "", re); gsub("(\\{|\\}|\\\\E).*$", "", re);
                     keys=re; gsub(/</, "<lt>", keys); gsub(/\t/, "<c-v><c-i>", keys);
-                    out = out " %{" $2 " {MenuInfo}" re "} %{eval -collapse-jumps %{ try %{ edit %{" ENVIRON["tagroot"] $2 "}; exec %{/\\Q" keys "<ret>vc} } catch %{ echo -color Error unable to find tag } } }"
+                    if(length(re) > 0){
+                        out = out " %{" $2 " {MenuInfo}" re "} %{eval -collapse-jumps %{ try %{ edit %{" ENVIRON["tagroot"] $2 "}; exec %{/\\Q" keys "<ret>vc} } catch %{ echo -color Error unable to find tag } } }"
+                    }
                 }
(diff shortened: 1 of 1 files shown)


lenormf:
Made it so we stop searching the remaining tag files in `ctagsfile` once we have a symbol definition, I also had a weird case where only a filename would be added to the menu, additionaly to the correct symbol entries.

lenormf:
For reference, here's the hook I use to autoload a tag file's path:
```
## Detect tag files in parent directories
hook global BufCreate .* %{ %sh{
        MAX_DEPTH=10
        basedir=$(readlink -e $(dirname "${kak_opt_buffile}"))
        paths=''
        while [ $MAX_DEPTH -gt 0 ]; do
            if [ -s "${basedir}/tags" ]; then
                paths="${paths}:${basedir}/tags"
            fi

            if [ "${basedir}" = '/' ]; then
                break
            fi

            MAX_DEPTH=$((MAX_DEPTH - 1))
            basedir=$(readlink -e "${basedir}/..")
        done

        printf %s\\n "set buffer ctagsfiles '${paths}'"
    } %sh{
        ordered_paths=$(printf %s\\n "${kak_opt_ctagsfiles}" | tr ':' '\n' | sort -u | grep -v '^$' | tr '\n' ':')
        printf %s\\n "set buffer ctagsfiles '${ordered_paths}'"
} }
```

It seeks tag files whenever a buffer is opened, which results in lots of paths added automatically to the `ctagsfiles` variable and a "harder" stress test of the tag parsing function. I don't want to remove the `awk` call anymore.

mawww:
Hi, I finally took time to take a look at that, I am wondering if we want to merge the found tags along all files (the original behaviour before my buggy change), or just take the found tags from the first file that has one. It seems your approach does the latter, are you confident thats the way to go ?

lenormf:
I was thinking that there are not many cases where a symbol will be listed in two different tags files but refer to two different implementations. However, it's much more common to have several ctags files (a big one in the root directory, a smaller one in a local directory) accessible for parsing, so I thought it would be a waste to parse all the tags files in the file tree.
E.g. you want to code a driver for the linux kernel, and have a tags file in your work directory (say, `linux/src/drivers/mydriver/tags`), and a main one that you use to lookup symbols that are implemented all over the kernel (`linux/tags`). If I'm just looking to jump to the definition of a function declared in my driver code, I don't want the `tags` function to parse the main tags file, which contains thousands of symbols, so I make it stop as soon as a match is found.

My configuration automatically adds all the tags files that are in the current and parent directories (up to 10 levels), so my use of the `tags` command might be a corner case that others are not interested in.

mawww:
Hi,

I thought about that a bit more, and it seems to me we really should merge results from all the tag files: if you want the tag from the first tag file, having the other results is a minor inconvenience, however if you want a tag that is not in the first tag file, having to remove that file from the option is a major inconvenience, so I'd say we really want to merge all the results.

In C it is not a big problem as there is no overloading, so its very likely you dont have many different tags sharing the same name, in other languages (C++ in particular), with overloading, you can endup with the same function name being used in many places, specially if you use namespaces but lookup tags unqualified.

lenormf:
Parsing all the tag files can massively slow down all the tag lookups (several seconds in the case of a big project) no matter what tag is selected, whereas it's still possible to lookup a symbol that has several definitions (which doesn't happen too often over all the symbols that I look up when coding) when it's fully qualified.

mawww:
Thats strange, here performances are "acceptable" with an ~300MB tag file.

Also, besides requiring fully qualified names being very inconvenient (cannot easily jump to selected name tag), overloading is still problematic.

It seems the only real problem with the full tag approach is the performance one, do you have an idea where the performance problem comes from ? what is the size of the tag files in your use case ?

lenormf:
My "root" tags file is 1.8GB, and I'm working over NFS so the two combine well into a massive time sink if I'm not careful.

But even without talking about big tags files, I thought it made sense not to spawn as many `readline` or `awk` processes as there are tags files just to cover a corner case that it's possible to work around of, when in reality the first result returned by `readtag` is often the one we want.

Then again, I might have a different use for the `tag` function then most other users, so idk (I don't even use object oriented programming languages that often).

lenormf:
Closing this for now, as the file itself has changed since I first opened this PR. A fix singled out in #782 still survives the PR, I'll see what to do about the lookup algorithm later.

PR #687, opened 2016-05-31
Update README.asciidoc with debian installation instructions.

matthewjberger:


lenormf:
This section is more of a list of the different packages that exist for kakoune in several distributions, and adding commands that are generic across all UNIX systems doesn't make much sense (the dependency installation phase, followed by `git clone … && cd src && make install`).

However, there is a script ([`contrib/make_deb.bash`](https://github.com/matthewjberger/kakoune/blob/master/contrib/make_deb.bash)) that generates a .deb package for debian-based systems that users of those systems might miss, so I think it'd be more relevant to describe quickly how to use it as a tip. If an official package ever makes his way to the upstream repositories, we'll update that section with its name (`sudo aptitude install kakoune` or something).

matthewjberger:
Would something like this be more appropriate? I wasn't aware of this script until I read your comment, and the script worked properly for me when I ran it with the appropriate flags set.

lenormf:
Yea I think that's more relevant, thanks.

mawww:
Hi, Can you use your name for the copyright waiver commit ? and maybe squash the various changes together in a single commit ?

matthewjberger:
I cleaned up the commits as you requested. Thank you for your consideration!

mawww:
Thank you !

PR #696, opened 2016-06-11
[POC] Add `m4` support to the string expansions

lenormf:
Hi,

Despite what the numerous changes to the code seem to suggest, this PR is less about adding arbitrary scopes in a generic manner than it is about a not very well known tool, `m4`. The following changeset adds the `%m4{}` scope to the already supported string expansions, and since I was more interested into the results than the changes themselves, I just duplicated code that worked for the shell and patched it (hence the `POC` tag in the title).

**About `m4`**

It's a GNU pre-processor that's included by default into most (all?) distributions, and allows simple macro replacements, functions, execution of shell commands, and much more.

**Why we're interested in it**

There are some cases where, when we develop a kak script, the only feature we're interested in is variable replacement, such as with color schemes. And the current way of implementing those scripts is to spawn a shell, so we can use the variables syntax, which is quite heavy/unecessary.

Using `m4` is a good replacement in those cases, as it's way faster than an entire shell, doesn't leave big memory footprints, and makes the scripts more readable (check out [my modified base16 theme](https://github.com/lenormf/kakoune/blob/be3ae4d468c863ca5c4db9d9b9f5e4d531755326/colors/base16.kak)). And it might even be possible to use `m4` in cases where we need more advanced features (a couple functions, mathematical evaluations, file inclusions, shell commands, …).

**Gains**

I haven't made any benchmarking or anything, but I wouldn't be surprised if we gained a good second during startup by simply replacing the `%sh` scopes with `%m4` ones in the color schemes. And as stated before, the features of this pre-processor go beyond simple symbol substitution, so changing a few shell scopes here and there in default kak scripts as well could improve things even further.

The `m4` syntax is based on defines, like `cpp`, and even though it's not much different from the regular shell script syntax, using it removes some clutter such as dollar signs and braces. Not everything has to be enclosed in double quotes either, which makes syntax highlighting slightly better.

This tool is also installed by default on all distributions I could try (it has a cygwin package as well, but I haven't checked whether it's installed by default), which means it's a very good trade if we use it.

**Documentation**

For more info about `m4`'s features, refer to the [documentation](https://www.gnu.org/software/m4/manual/m4.html).

### DIFF
--- a/colors/base16.kak
+++ b/colors/base16.kak
@@ -2,68 +2,62 @@
 ## base16.kak by lenormf
 ##
 
-%sh{
-    black_lighterer='rgb:383838'
-    black_lighter='rgb:2D2D2D'
-    black_light='rgb:1C1C1C'
-    cyan_light='rgb:7CB0FF'
-    green_dark='rgb:A1B56C'
-    grey_dark='rgb:585858'
-    grey_light='rgb:D8D8D8'
-    magenta_dark='rgb:AB4642'
-    magenta_light='rgb:AB4434'
-    orange_dark='rgb:DC9656'
-    orange_light='rgb:F7CA88'
-    purple_dark='rgb:BA8BAF'
+%m4{
+    define(`black_lighterer', `rgb:383838')dnl
+    define(`black_lighter', `rgb:2D2D2D')dnl
+    define(`black_light', `rgb:1C1C1C')dnl
+    define(`cyan_light', `rgb:7CB0FF')dnl
+    define(`green_dark', `rgb:A1B56C')dnl
+    define(`grey_dark', `rgb:585858')dnl
+    define(`grey_light', `rgb:D8D8D8')dnl
+    define(`magenta_dark', `rgb:AB4642')dnl
+    define(`magenta_light', `rgb:AB4434')dnl
+    define(`orange_dark', `rgb:DC9656')dnl
+    define(`orange_light', `rgb:F7CA88')dnl
+    define(`purple_dark', `rgb:BA8BAF')dnl
 
     ## code
-    echo "
-        face value ${orange_dark}+b
-        face type ${orange_light}
-        face identifier ${magenta_dark}
-        face string ${green_dark}
-        face error ${grey_light},${magenta_light}
-        face keyword ${purple_dark}+b
-        face operator ${grey_light}
-        face attribute ${orange_dark}
-        face comment ${grey_dark}
-        face meta ${orange_light}
-    "
+    face value              orange_dark+b
+    face type               orange_light
+    face identifier         magenta_dark
+    face string             green_dark
+    face error              grey_light,magenta_light
+    face keyword            purple_dark+b
+    face operator           grey_light
+    face attribute          orange_dark
+    face comment            grey_dark
+    face meta               orange_light
 
     ## markup
-    echo "
-        face title blue
-        face header ${cyan_light}
-        face bold ${orange_light}
-        face italic ${orange_dark}
-        face mono ${green_dark}
-        face block ${orange_dark}
-        face link blue
-        face bullet ${magenta_light}
-        face list ${magenta_dark}
-    "
+    face title              blue
+    face header             cyan_light
+    face bold               orange_light
+    face italic             orange_dark
+    face mono               green_dark
+    face block              orange_dark
+    face link               blue
+    face bullet             magenta_light
+    face list               magenta_dark
 
     ## builtin
-    echo "
-        face Default ${grey_light},${black_lighter}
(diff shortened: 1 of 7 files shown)


mawww:
Hello,

While I agree the color scheme gets more readable, I dont think that by iself warrants adding support for m4, and without measurement, I am not convinced this will actually be faster than using the shell.

On my laptop, the `dash` shell is ~800k (static binary), whereas `m4` is ~150k but is dynamicaly linked to the libc.so (1.7M), so I'd be surprised if m4 does indeed start much faster.

On my laptop, if I add `set global debug profile` on top of the share/kak/kakrc file, and aggregate the shell evaluation timings, a kakoune startup spends 87ms in the shell. which can probably be improved, but is arguably fast enough.

Another approach could be to *replace* %sh with %m4 altogether, but I'd argue that, although m4 syntax is clearly simpler and cleaner, people are much more familiar with sh syntax.

lenormf:
I did a quick test, and made a kak script that loads 1000 times `base16.kak` (and `base16.m4` the second time), loaded that up into kakoune and got the following results:
```
## source base16.kak
> time ./kak -n -e 'source /tmp/test.kak; quit'
./kak -n -e 'source /tmp/test.kak; quit'  0.71s user 0.19s system 25% cpu 3.517 total

## source base16.m4
 > time ./kak -n -e 'source /tmp/test.kak; quit'  
./kak -n -e 'source /tmp/test.kak; quit'  0.55s user 0.15s system 53% cpu 1.297 total
```

Probably not the most relevant test, since there are less than 100 files in the `rc` directory, but it still shows an improvement.

A shell spends a ton of time reading files when starting up, whereas `m4` is ready to go as soon as the binary is mapped into memory. I don't think replacing `%sh` with it would be good because they are both useful in certain situations, but it definitely feels cleaner to use `%m4` when all you want is a search&replace. Projecting myself into a future where there are 50+ colorschemes available by default, I would assume we would not want kakoune to spawn 50 shells when starting up (also there are some parts in the other scripts -not colorschemes- that actually only use a shell to have variables, making those parts use a smaller pre-processor would help the load as well).

lenormf:
Also the shell that is pointed to by `/bin/sh` (when it's actually a symlink) depends on the host system, so `m4` might compare much differently from one system to another (archlinux points `/bin/sh` to `/bin/bash` for instance).

mawww:
Regarding colorschemes, they are not loaded on startup, so they dont impact load time (`:colorscheme` is just a wrapper around `:source`).

I dont think we have any performance problem with %sh that would be solved by %m4, if we end up hitting a bottleneck on script file loading, I expect we will need a more complicated solution than just adding another interpreter.

Additionally, m4 does not seem to support environment variable access, which means it would rely on shell evaluation to access all kakoune values, which is not very satisfying either. Or we would need to add built-in support for m4 directly into kakoune so that values are passed to the m4 interpreter via command line argument.

So yeah, to recap, I am not very favorable to this direction, which looks like feature creep with not enough benefits to me.

alexherbo2:
I like the syntax but I’m not convinced too.

lenormf:
>we would need to add built-in support for m4 directly into kakoune so that values are passed to the m4 interpreter via command line argument.

That was the idea, it's a trivial change and support for the scope would have to be tailored to `m4` anyway if generic code was implemented to allow arbitrary scopes to be added easily (e.g. some interpreters require a `-` argument, some a `-f -` arguments pair etc).

Something like `m4 -DKAK_SESSION=1234 -D… -` would work.

What if no scope other than `%sh` was supported explicitely in the editor, but there would still be a way to add your own arbitrary scope to it (with the possibility to more or less configure the command implementing the scope) ?

lenormf:
@mawww I'm closing, what about [my last message](https://github.com/mawww/kakoune/pull/696#issuecomment-225412620) though ?

mawww:
Hi,

Supporting arbitrary interpretter was on the TODO for some time, but I ultimately decided not to go for it, because it means any kak script trying to use them depends on the user options (if a script uses %m4, and the user does not have the m4 interpretter set up, it will fail), and also because its really easy to just do that in a %sh{...} (just run m4 in your %sh).

Additionally, the interpretter configuration system would need to be very expressive:
* How to pass the kakoune variables (env, through arguments, which format ?)
* How to pass the script itself (in stdin ? in the command arguments ? in a temp file ?)

It seems much simpler to let all that be handled in a `%sh` directly. The main downside is the potential performance problems, but its never been a problem in practice outside of cygwin, where forks are extremely expensive (but we now have Windows subsystem for Linux as a peformant alternative).

Cheers,

Maxime.

lenormf:
ACK, thanks for the reply.

PR #711, opened 2016-06-26
Add a shebang detection script

lenormf:
This script reads the first line of the buffer and does a stupid&simple
association between the last word (excluding interpreter flags) and the
filetype to assign to the buffer.

### DIFF
--- a/rc/base/shebang.kak
+++ b/rc/base/shebang.kak
@@ -13,9 +13,9 @@ def shebang-detect -docstring "Detect the shebang at the top of the buffer and a
         %sh{
             filetype=''
             case \"\${kak_reg_m}\" in
-                zsh|bash|csh|ksh|sh) filetype=sh;;
-                python|python2|python3) filetype=python;;
-                lua|perl|ruby) filetype=\"\${kak_reg_m}\";;
+                zsh|bash|csh|ksh) filetype=sh;;
+                python2|python3) filetype=python;;
+                *) filetype=\"\${kak_reg_m}\";;
             esac
             if [ -n \"\${filetype}\" ]; then
                 printf %s \"set buffer filetype \${filetype}\"
--- a/rc/base/shebang.kak
+++ b/rc/base/shebang.kak
@@ -1,0 +1,28 @@
+##
+## shebang.kak by lenormf
+## Detect a shebang in a buffer and assign the proper filetype accordingly
+##
+
+def shebang-detect -docstring "Detect the shebang at the top of the buffer and assign the proper filetype accordingly" %{ %sh{
+    if [ -n "${kak_opt_filetype}" ]; then
+        exit
+    fi
+
+    printf %s "eval -draft -save-regs m %{ try %{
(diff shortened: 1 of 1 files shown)


lenormf:
Any comments on this ?

rc/base/shebang.kak
```diff
@@ -0,0 +1,28 @@
...
+    if [ -n "${kak_opt_filetype}" ]; then
+        exit
+    fi
+
+    printf %s "eval -draft -save-regs m %{ try %{
+        exec -save-regs '/' gg <a-x> s^#!\s*[^\n-]+<ret> \;be\"my
+        %sh{
+            filetype=''
+            case \"\${kak_reg_m}\" in
+                zsh|bash|csh|ksh) filetype=sh;;
```
casimir: What about `*sh)`? Builtins are almost always the same in shells.

casimir:
I really tried hard to find something but out of unrealistic edge case LGTM.

About the edge case, it doesn't work for a shebang of this form (rarely saw it):
```
#!/usr/bin/env bash -e
```

casimir:
Nevermind I already had a filetype for the buffer I tested in. It works too in this case.

lenormf:
I wouldn't be surprised if someone found a way to break this, it can be tricky sometimes to detect what executable is the interpreter. I hope that it covers at least the most common cases though.

casimir:
@lenormf I was still hoping for a merge. Even more now than the mime-type thing as been statued.

maximbaz:
I was about to start implementing this feature when I found this PR. @lenormf why did you close this? Sounds like a good feature to have, when all other detection mechanisms have failed.

lenormf:
I closed this PR because it's redundant with the `file` functionality. Arguably you could have used it if you don't have `file` installed, but it's probably not worth the time.

maximbaz:
Ah, it seems `file.kak` can recognize variations like `#/bin/bash`, `#!/usr/bin/zsh` and `#!/usr/bin/env bash`, but sadly not `#!/usr/bin/env sh` or `#!/usr/bin/env zsh`. Does anyone know if we can "teach" `file` more shebangs?

lenormf:
Use the `magic` file.

codesoap:
@lenormf could you provide a link to the `magic` file? I couldn't find a`magic.kak` file in `/usr/local/share/kak/rc` or on github.

alexherbo2:
@codesoap https://github.com/mawww/kakoune/issues/2858#issuecomment-482899359

alexherbo2:
@maximbaz Add a `magic` file to your dotfiles.  If you want to take example, you can find mine [here][magic].

[magic]: https://github.com/alexherbo2/configuration/blob/master/home/magic

jmpohl:
@codesoap, were you able to figure out the nuances for OpenBSD? I am on and getting similar behavior to what you described and am not having much success adding `#!/usr/bin/env zsh` to my magic file.

jmpohl:
Disregard previous message. I got my solution and documented it [here](https://github.com/mawww/kakoune/issues/3982#issue-773969548).

PR #713, opened 2016-06-27
Fix the EOL format detection

lenormf:
Only switch the end of line format to `CRLF` if all the lines in a
buffer end with a CRLF, as opposed to require only one.

### DIFF
--- a/src/buffer.cc
+++ b/src/buffer.cc
@@ -34,6 +34,7 @@ static ParsedLines parse_lines(StringView data)
         pos = data.begin() + 3;
     }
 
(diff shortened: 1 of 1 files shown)


mawww:
Unfortunately that change is not enough:
* It will default to Crlf on empty files
* It wont preserve the `CR` on lines that have it when we have a file with mixed line endings.

What we need is to remove the `CR` on all lines only when we are in `CRLF` mode.

lenormf:
Not the easy fix I thought it would be, I'll just close this until I'm done with the other things I'm working on, chances are you'll come up with a better fix before then.

PR #721, opened 2016-07-03
Add CMakeLists.txt to contrib

tony:
Carried on from https://github.com/mawww/kakoune/pull/693#issuecomment-226737860.

Tested against e262dc1 on OS X 10.11 and FreeBSD 11-CURRENT

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ tags
 GPATH
 GRTAGS
 GTAGS
+/build/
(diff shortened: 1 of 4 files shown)


lenormf:
This is going to need some squashing :p

tony:
Better?

doppioandante:
Looks good,just a small note on the msvc code path: as far as I know kakoune doesn't compile under it, but leaving it there shouldn't be a big deal.

tony:
@doppioandante good catch, removed the mscv / conditions around that

mawww:
Can you prepend a copyright waiver commit as well ? (see CONTRIBUTING)

PR #731, opened 2016-07-11
Readonly mode

lenormf:


src/file.cc
```diff
@@ -222,65 +222,74 @@ void write(int fd, StringView data)
...
-        eoldata = "\r\n";
-    else
-        eoldata = "\n";
-
-    if (buffer.options()["BOM"].get<ByteOrderMark>() == ByteOrderMark::Utf8)
-        if (::write(fd, "\xEF\xBB\xBF", 3) < 0)
-            throw runtime_error(format("unable to write data to the buffer (fd: {}; errno: {})", fd, ::strerror(errno)));
-
-    for (LineCount i = 0; i < buffer.line_count(); ++i)
+    if ((buffer.flags() & Buffer::Flags::ReadOnly) == false)
```
mawww: Do not move the whole function logic inside an if block, just early return.

src/file.cc
```diff
@@ -222,65 +222,74 @@ void write(int fd, StringView data)
...
+            write(fd, eoldata);
+        }
     }
 }
 
 void write_buffer_to_file(Buffer& buffer, StringView filename)
 {
-    buffer.run_hook_in_own_context("BufWritePre", buffer.name());
-
+    if ((buffer.flags() & Buffer::Flags::ReadOnly) == false)
```
mawww: same as previous

src/file.cc
```diff
@@ -222,65 +222,74 @@ void write(int fd, StringView data)
...
-
-    char pattern[PATH_MAX];
-    if (dir.empty())
-        format_to(pattern, ".{}.kak.XXXXXX", file);
-    else
-        format_to(pattern, "{}/.{}.kak.XXXXXX", dir, file);
-
-    int fd = mkstemp(pattern);
-    if (fd >= 0)
+    if ((buffer.flags() & Buffer::Flags::ReadOnly) == false)
```
mawww: idem

src/main.cc
```diff
@@ -610,7 +615,7 @@ int run_server(StringView session, StringView init_command,
     return 0;
 }
 
-int run_filter(StringView keystr, StringView commands, ConstArrayView<StringView> files, bool quiet)
+int run_filter(StringView keystr, StringView commands, ConstArrayView<StringView> files, bool quiet, bool readonly)
```
mawww: I am not sure a read only filter makes sense

src/normal.cc
```diff
@@ -88,11 +88,17 @@ template<InsertMode mode>
 void enter_insert_mode(Context& context, NormalParams)
 {
     context.input_handler().insert(mode);
+
```
mawww: I dont like all this repetition in every function, not sure what the correct solution would be (maybe Buffer should throw a runtime_error when we try to insert in a read only buffer), but this change is too intrusive.

src/file.cc
lenormf: I don't like using `return` if there's nothing to return. I would have returned a boolean, but it seemed more trouble for not much value.

src/main.cc
lenormf: That can be arranged, I modified this function for the sake of consistence with the `run_server` function, since they both declared options.

mawww:
I think one thing remaining to be defined is the desired behaviour when the user tries to modify a read only buffer.

To properly define that, what are the actual use cases for read only buffers that are not already handled by filesystem permissions ? Is it important to prevent modification of the buffer in Kakoune, or only the saving of the buffer to its corresponding file ? Should we be able to write a read only buffer to another file ?

lenormf:
My main usecase for the readonly mode is to be able to use kakoune as a mail editor, I want to be able to pass an existing mail (from my inbox) and be sure that I won't modify it by mistake.

I chose to print a warning when modifying a buffer in readonly mode because that's what vim does©, and it made sense at the time the decision was taken. Maybe there's a better option (c.f. the comment above about polluting the insertion primitives), but I haven't really thought about it.

Saving a readonly buffer to another file is still possible afaik, I'd like to keep it that way since I don't see why it wouldn't be possible if all you want is protect your current buffer (regardless of whether the buffer is a file or not).

Thanks for the comments.

src/file.cc
mawww: I prefer early returns as they avoid a needless indentation, and I dont see a big problem with `return;` in void functions, we are not returning a value, but we are returning (early) from the function.

Thats a matter of taste, however at least for consistency could you convert that to just an early return ?

src/normal.cc
mawww: I am not really keen on any solution that requires us to add code/wrap every single function that might end up trying to modify the buffer. I'd prefer a solution which will, by design, work. I think throwing a runtime error on tentative to modify a read only buffer could make sense.

src/file.cc
lenormf: Sure!

src/normal.cc
lenormf: The thing is that we don't want to prevent the user from modifying the buffer, just let him know that he won't be able to save it under the same name (`:w somewhere_else` should work). That way you could open a file that contains text you're interested in -e.g. a mail-, transform it using kakoune primitives and put it in your clipboard for external use, all of that with the guarantee that the original buffer won't be modified.

lenormf:
I've fixed the issues with the failing tests, and I also removed the warnings when modifying the buffer in readonly mode. I figured that users would know if they had started kakoune in readonly mode, and they can also use a modeline widget (or read the `readonly` variable directly) if a reminder is needed.

mawww:
We should probably complain that -ro and filter mode are not compatible instead of silently ignoring -ro. Also, could you clean up the history so that we dont have 2 full rewrites of the file manipulation functions (one moving code inside the if, then another one moving the code outside).

lenormf:
I squashed a few commits together, improved some commit messages, added warnings when needed (`-f`, `-c`, `-p`).

src/file.cc
```diff
@@ -245,6 +248,9 @@ void write_buffer_to_fd(Buffer& buffer, int fd)
 
 void write_buffer_to_file(Buffer& buffer, StringView filename)
 {
+    if (buffer.flags() & Buffer::Flags::ReadOnly)
```
mawww: This is called by `write_buffer`, so this means we wont be able to write buffer even if we specify the filename explicitely.

src/commands.cc
```diff
@@ -239,6 +239,9 @@ void write_buffer(const ParametersParser& parser, Context& context, const ShellC
     if (parser.positional_count() == 0 and !(buffer.flags() & Buffer::Flags::File))
         throw runtime_error("cannot write a non file buffer without a filename");
 
+    if ((context.buffer().flags() & Buffer::Flags::ReadOnly) and parser.positional_count() == 0)
+        throw runtime_error("cannot overwrite the buffer when in readonly mode");
+
```
mawww: Do we want to allow writing explicitely to the the buffer existing name ? or do we need to check we are referencing another file ?

src/file.cc
```diff
@@ -266,6 +272,9 @@ void write_buffer_to_file(Buffer& buffer, StringView filename)
 
 void write_buffer_to_backup_file(Buffer& buffer)
 {
+    if (buffer.flags() & Buffer::Flags::ReadOnly)
+        return;
+
```
mawww: Are we sure we want to disable writing backup files for read only buffers ?

lenormf: Changes on readonly buffers are supposed to be temporary, so I disabled backups. It's arguable, maybe it's not the best way, but it was the simplest.

src/file.cc
lenormf: I'll have to think about this one.

lenormf:
Sorry for the rebasing madness, I want to keep the history clean.

I moved the checks from the writing primitives to higher level functions (`file.cc` → `commands.cc`), and also added a check to make sure we can't overwrite a buffer indirectly using a symlink.

src/buffer_manager.cc
```diff
@@ -78,7 +78,8 @@ void BufferManager::backup_modified_buffers()
 {
     for (auto& buf : m_buffers)
     {
-        if ((buf->flags() & Buffer::Flags::File) and buf->is_modified())
+        if ((buf->flags() & Buffer::Flags::File) and buf->is_modified()
+            and !(buf->flags() & Buffer::Flags::ReadOnly))
```
mawww: can you replace the `!` there with `not`, in Kakoune source code we use the alternative, word based logical operators.

src/commands.cc
```diff
@@ -239,6 +239,12 @@ void write_buffer(const ParametersParser& parser, Context& context, const ShellC
     if (parser.positional_count() == 0 and !(buffer.flags() & Buffer::Flags::File))
         throw runtime_error("cannot write a non file buffer without a filename");
 
+    // if the buffer is in read-only mode and we try to save it directly
+    // or we try to write to it indirectly using e.g. a symlink, throw an error
+    if ((context.buffer().flags() & Buffer::Flags::ReadOnly)
+        && (parser.positional_count() == 0 || real_path(parser[0]) == buffer.name()))
+        throw runtime_error("cannot overwrite the buffer when in readonly mode");
```
mawww: replace `&&` with `and` and `||` with `or` please.

src/buffer_manager.cc
lenormf: Some statements still use the regular C operators though, is there a reason for that ?

mawww: involuntary deviation from the coding standard, we should fix them when we see them.

lenormf:
If no other modifications are needed, this PR should be good to merge!

PR #732, opened 2016-07-12
[WIP] C-family hooks split and improved

lenormf:
Hi,

I haven't really extensively used this patch, hence the WIP tag. The PR implements several features that were discussed on IRC, and fixes a few issues that I kept stumbling upon regularly.

* it's now possible to only use indenting hooks and disactivate the ones that insert things in the buffer (e.g. semi-colons)
* simplify a few patterns, such as indenting after certain keywords
* don't remove all the trailing whitespace in the buffer, only remove those that were added by the indenting hook
* proper comment supports: automatically close a multiline comment when the return key is hit twice, copy the commenting prefix used on the previous line etc

Like I said, I haven't tested it much, but if someone wants to give it a try until I do, please do.

Cheers.

### DIFF
--- a/rc/core/c-family.kak
+++ b/rc/core/c-family.kak
@@ -37,7 +37,7 @@ hook global BufSetOption mimetype=text/x-objc %{
 
 def -hidden _c-family-trim-autoindent %[ eval -draft -itersel %[
     ## remove the line if it's empty when leaving the insert mode
-    try %[ exec <a-x> <a-k>^\h+$<ret> d ]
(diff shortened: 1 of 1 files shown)


doppioandante:
One nice thing to have would be support for comment-line and comment-selection, if I have time I'll send a pull request to you

lenormf:
It's supported by default, the characters are already set in `commenting.kak`.

lenormf:
There are still issues with this that I haven't fixed !

mawww:
Fixed a few locally, will push when i get access to a connection.

PR #740, opened 2016-07-20
Change session name

doppioandante:
Hi, I frequently find myself wanting to give the session a more descriptive name (to ease kak -c).
This is what I came up with, any opinions?

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -1895,6 +1895,25 @@ const CommandDesc change_working_directory_cmd = {
     }
 };
 
+const CommandDesc change_session_name = {
+    "session",
+    nullptr,
+    "session <name>: change remote session name",
+    ParameterDesc{{}, ParameterDesc::Flags::None, 1, 1},
+    CommandFlags::None,
+    CommandHelper{},
+    CommandCompleter{},
+    [](const ParametersParser& parser, Context&, const ShellContext&)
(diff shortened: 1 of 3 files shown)


mawww:
Hello,

That wont work, we would need as well to open a new session socket and close the previous one. We need to check as well if that would impact already connected clients (I think it will not, but we need to test that).

On the feature itself, I think it makes sense to be able to change the session name after the fact, so I am in favor of adding that feature.

doppioandante:
Hi, from my tests it seems that simply renaming the socket (link+unlink) does the trick. I don't know if this is the case everywhere, I will check tomorrow.
The connected clients don't seem to care.

I also need to check if the session name is already in use.

doppioandante:
I couldn't find much about moving unix domain socket handles. I'm not very familiar with unix programming but from what I could understand they contain just some metadata required for the actual socket connection, so it should be OK.
I would be more confident if someone tried on cygwin or bsd though.

What name would you like for the command? It currently is :session <new_session_name>

mawww:
renamed the command to `:namesession` to be consistent with `:nameclient` and `:namebuf` however I think we really want to changes all of these to something less ugly.

lenormf:
Something like `:name {session,client,buffer} <str>` ?

PR #744, opened 2016-07-25
iTerm2 support

casimir:
This is a proposition to support window management with [iTerm2](www.iterm2.com) (OSX only). This is a shameless ripoff of [tmux.kak](https://github.com/mawww/kakoune/blob/master/rc/core/tmux.kak) using the AppleScript API of the terminal.

Some details:
- you can create splits, tab and window with `iterm-new-*` functions
- it override `new` and `focus` if we are in iTerm and not in a tux session
- I could find a way to identify each iTerm sessions but not how to focus to it, for now `focus` just print a message

### DIFF
--- a/rc/core/iterm.kak
+++ b/rc/core/iterm.kak
@@ -1,0 +1,75 @@
+# https://www.iterm2.com
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+## The default behaviour for the `new` command is to open an vertical pane in
+## an iTerm session if not in a tmux session.
+hook global KakBegin .* %{
+    %sh{
+        if [ "$TERM_PROGRAM" = "iTerm.app" -a -z "$TMUX" ]; then
+            echo "
+                alias global new iterm-new-vertical
+                alias global focus iterm-focus
+            "
+        fi
+    }
+}
+
+def -hidden -params 1.. iterm-new-split-impl %{
+    %sh{
+        direction="$1"
+        shift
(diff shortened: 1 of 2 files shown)


mawww:
Wow, thats very nice.

I think iterm2.kak should be in `rc/extra/` instead of `rc/core/` (the real question here would be if `x11.kak` and `tmux.kak` should really be in `rc/core/`).

Also, could you use you name in place of "John Doe" in the copyright waiver ?

Thanks,

Maxime.

lenormf:
Hi,

Good one, this will prove convenient to iTerm2 users.

However, I think it might be better to have the script added to `extra` instead of `core`. You also have to replace "John Doe" with your actual name in the copyright waiver.

casimir:
Well that was a quick (double) reply! I amend the waiver and moved the file to the right directory.
I couldn't find out of fef02779981acb615339697 so I just put it with the tmux one.

casimir:
Looking through the sources I feel like window management scripts should be in `rc/base` rather than `rc/extra`. What do you think?

mawww:
the current logic for organization between directory is not per functionality, but per popularity/criticality. `core` should be for critical support for a confortable edition of kakoune itself, `base` for general, widely available tools, `extra` for more specific tools.

In that sense, I think `x11.kak` and `tmux.kak` should probably move to `base`, as they are not really critical (not in the same sense c++ highlighting, make and grep support are). `iterm2.kak` being specific to a terminal emulator restricted to a single platform (although pretty popular), would be between `base` and `extra`, my current preference is `extra`.

casimir:
I see your point I can only agree with that.

PR #768, opened 2016-08-23
Add a basic FAQ with unanswered questions

lenormf:
Hi,

The same questions come back over and over on IRC and the bug tracker, so I thought having a FAQ that we could point new users to would be helpful.

For now I haven't written any answer to the questions in the `FAQ.asciidoc` file, it seemed more important to decide what questions should be included in the file first.

HTH.

### DIFF
--- a/doc/manpages/faq.asciidoc
+++ b/doc/manpages/faq.asciidoc
@@ -1,13 +1,16 @@
-Frequently Asked Questions
---------------------------
+KAKOUNE(1)
+==========
+
+NAME
+----
+faq - a
 
 How to pronounce the name of the project and what does it mean ?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The name of the project is pronounced "Kak-oon", and is a word taken from
-a New Caledonia dialect based on french. It means a hard blow, usually a
-punch, but generally refers to a blow in which all of one's strength went
-into doing.
+The name of the project is pronounced "Kak-oon", and is a word taken from a
+New Caledonian dialect based on french. It means a hard blow, usually a punch,
+but generally refers to a blow into which which all of one's strength went.
 
 Is there going to be a Windows port of Kakoune ?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


doppioandante:
I think it's quite complete. We could add <space> to reduce to a single selection which also comes often.

lenormf:
I did think of that one, but it boils down to reading the documentation so I didn't add it to the list.

doppioandante:
Maybe we should add a little explanation about the scripting language in the FAQ? I think there's something in the design document, but it's frequently asked.

lenormf:
If answers take more than a couple sentences to explain, then they turn into documentation which isn't the point of this file.

Questions about kakoune scripts are indeed common, but understanding them boils down to reading how the current scripts are implemented, and I don't see myself adding a question in the FAQ whose answer is "just look at how we do it".

Or maybe I'm wrong, I'll go for whatever is most convenient.

mawww:
I think we will need something regarding insert complation (how  can I enable clang/jedi based intelligent completion).

Also, I am still wondering if this needs to be a file in the source tree, instead of a page in the wiki. I recon a file in the source tree would be easier to find for people who compile Kakoune themselves, but as we are pretty well supported in the ditribution package managers, that might be not a lot of users. Installing the FAQ in /usr/share/doc/kakoune is arguably less discoverable than the wiki on github.

An alternative could be a manpage (accessible via `:doc faq`).

What do you think ?

lenormf:
I'd rather we make this file accessible offline, a `doc` page sounds good to me. I'll start writing replies and move everything to `faq.asciidoc`.

lenormf:
Moved the documentation to a `doc` page, replied to most questions (still waiting on PR #783), propose edits.

doc/manpages/faq.asciidoc
```diff
@@ -0,0 +1,92 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.
+
+Are there any non-console based frontends available ?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+No graphical frontend is currently officially maintained, you can however try experimental community-developed ones.
+
+How do I automatically indent code, as Vim does with `==` ?
```
doppioandante: I think it is just `=`

doc/manpages/faq.asciidoc
```diff
@@ -0,0 +1,92 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+As `Kakoune` doesn't parse the contents of the buffers, there is no
+builtin equivalent for this Vim feature. Use a formatter/prettifier
+dedicated to the language you're using (e.g. `indent` for C).
```
doppioandante: Maybe it would be worth to mention the pipe command here

doc/manpages/faq.asciidoc
```diff
@@ -0,0 +1,92 @@
...
+Why aren't widely known command line shortcuts such as <c-w> or <c-u> available in Kakoune ?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Despite their widespread availability in multiple tools, those
+shortcuts do not fit the paradigm that Kakoune implements, which is
+based on selections first.
+
+However, you can easily declare key mappings in your configuration
+file to be able to use those control-based shortcuts in insert mode
+(c.f. the "map" command in the "commands" documentation page).
```
doppioandante: I would put the mapping code directly here because people are lazy

doc/manpages/faq.asciidoc
lenormf: To indent a single line, you have to hit it twice as well, I guess everyone will get it if I only use one `=` sign ?

doc/manpages/faq.asciidoc
lenormf: I would have as well, except the question is more about how to implement mappings that are already available elsewhere rather than how to implement the two given as example. I also tried to avoid giving any code at all in the file, as I want to give generic guidance without having to maintain the code I write there (which could change in the future).

doc/manpages/faq.asciidoc
lenormf: Good point, thanks.

doc/manpages/faq.asciidoc
casimir: Actually like a lot of Vim command it's `=` but if you double the key it applies to the current line `==` ⇔ `V=`.

lenormf:
More.

doc/manpages/faq.asciidoc
```diff
@@ -0,0 +1,123 @@
...
+"..." }`), which has a negligible impact on performance.
+
+Can I disable auto-indentation completely ?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+All the indentation hooks are conventionally named `<lang>-indent`, which
+allows us to use the `disabled_hooks` variable to disable indentation globally
+with the following command: `set global disabled_hooks '.+-indent'`
+
+Can I disable syntax highligting completely ?
```
doppioandante: typo in highlighting

lenormf: And thanks to completion, this typo has been inserted two more times afterwards :(

Thanks for noticing.

lenormf:
Rebased on `master`, formatted the paragraphs to 80 columns, fixed compiling of the page, added more info.

mawww:
Where are we at with this ? Is there some other changes pending ?

lenormf:
We're not guaranteed `kak` will work as a pager, c.f. #783, so the corresponding question is still unanswered.

lenormf:
I think the document is pretty much ready at this point, unless there's other questions to add.

lenormf:
Added another question, fixed a couple mistakes.

PR #780, opened 2016-08-30
Allow instant completion of `doc` pages

### LINKED ISSUE #750: :doc command should show autocompletion immediately when command is typed
~~I think it would be nice to autocomplete ``:doc`` command. If I understand it correctly it's only showing content of files in some folder so it shouldn't be so hard.~~

EDIT: Doc command is showing autocompletion only after TAB key is pressed

lenormf:
Closes #750


jkonecny12:
Thank you for this PR :+1:

PR #783, opened 2016-09-01
Add a wrapper to allow using `kak` as a pager

lenormf:
I've been working on other things to make kakoune an actual pager (key bindings etc), but I think people could use this script until I finish the whole PR.

### DIFF
--- a/contrib/kakpager.bash
+++ b/contrib/kakpager.bash
@@ -1,0 +1,23 @@
+#! /usr/bin/env bash
+##
+## kakpager.bash by lenormf
+## Instead of using `kak` directly in a shell pipeline to use the editor as a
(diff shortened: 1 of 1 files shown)


lenormf:
This fixes cases in which `kak` doesn't work when used as a pager, .e.g `man -P kak man`.

mawww:
I'd rather get a proper repro case for what prevents Kakoune from being used as a pager just with 'command | kak'. Once we fix that, this script will be unnecessary.

mawww:
I wont be merging that script, as it works excatly as simply piping into kakoune should. It seems we have a bug lurking around for that, but using a script to work around it is not a solution I want to sanction officially.

PR #791, opened 2016-09-07
Make `modeline.kak` POSIX-compatible

casimir:
This PR does 2 things, one by commit:
- change the case syntax to the POSIX one (OSX still has Bash 3 😒)
- use `sed -E` instead of `sed -r`

NB: Just like the BSD one the GNU `sed` acts the same for the 2 flags, one is the alias of the other. The OSX implementation only knows the `-E` form. I tested this without problem on OSX, Archlinux and Ubuntu.

### DIFF
--- a/rc/extra/modeline.kak
+++ b/rc/extra/modeline.kak
@@ -20,26 +20,19 @@ def -hidden _modeline-parse %{
             local tr=""
 
             case "${key}" in
-                so) ;&
-                scrolloff) tr="scrolloff ${value},${kak_opt_scrolloff##*,}";;
-                siso) ;&
-                sidescrolloff) tr="scrolloff ${kak_opt_scrolloff%%,*},${value}";;
-                ts) ;&
-                tabstop) tr="tabstop ${value}";;
-                sw) ;&
-                shiftwidth) tr="indentwidth ${value}";;
-                tw) ;&
-                textwidth) tr="autowrap_column ${value}";;
-                ff) ;&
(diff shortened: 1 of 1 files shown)


lenormf:
I made this script before we had the whole POSIX EVERYTHING overhaul, sorry about that.

However your changes only suit BSD `sed`, and will not work with GNU `sed`. We should try to stick to the options defined by the [POSIX sed utility page](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/sed.html), and there are not many of them.

I don't have an proper alternative to those patterns in BRCE syntax to propose spontaneously, I think this will need some actual testing work.

casimir:
This was really the only way I found to make it work on both Linux and OSX (I use this version every day).

But I agree with the POSIX EVERYTHING. Don't close the PR yet I'll finish the job!

lenormf:
The `-E` flag will **not** work on `GNU/sed`, the regular expression has to be changed and the flag dropped.

casimir:
I got your point. I was just saying I use it as a workaround since it works on Archlinux (GNU sed 4.2.2). Meanwhile I'm working on the regex overhaul. I'll ping when it's done.

casimir:
Ok so no more OSes just some POSIX. I squashed the three commits so it's easier to review.

I also made some very basic tests to check on several `sed` implementations https://gist.github.com/casimir/23c4e708d2b11d480c2c8b39555e430f.

PR #800, opened 2016-09-10
Implement some `readline` primitives in the prompt input handler

lenormf:
Hi,

I keep forgetting what the shortcuts in prompt mode are, and several people have asked about `readline` type bindings so I gave it a shot.

It might not be pretty/efficient/stable, but that's something I can work on (there might be bugs that need reporting btw).

I implemented the following shortcuts, usable in the prompt mode (only):
- `c-b`: move back a character
- `c-f`: move forward a character
- `c-a`: move to the beginning of the line
- `c-e`: move to the end of the line
- `c-d`: delete one char
- `a-f`: move forward a word
- `a-b`: move backward a word
- `c-t`: exchange the char before cursor with the character at cursor
- `a-t`: exchange the word before cursor with the word at cursor
- `a-u`: uppercase the current or following word
- `a-l`: lowercase the current or following word
- `a-c`: capitalize the current or following word
- `c-k`: remove text to the end of line
- `c-u`: remove text to the beginning of the line
- `a-d`: remove text to the end of the word
- `c-w`: remove text to the beginning of the word

### DIFF
--- a/src/input_handler.cc
+++ b/src/input_handler.cc
@@ -427,9 +427,8 @@ public:
                 if (m_cursor_pos > m_line.char_length() - 2)
                     m_cursor_pos = m_line.char_length() - 2;
 
-                insert(m_line[m_cursor_pos + 1]);
-                m_line = m_line.substr(0, m_cursor_pos + 1)
-                         + m_line.substr(m_cursor_pos + 2);
+                std::swap(m_line[m_line.byte_count_to(m_cursor_pos)],
+                            m_line[m_line.byte_count_to(m_cursor_pos + 1)]);
                 m_cursor_pos = old_cursor_pos;
             }
         }
--- a/src/input_handler.cc
+++ b/src/input_handler.cc
@@ -498,6 +498,8 @@ public:
             if (m_cursor_pos < len)
                 m_line[m_line.byte_count_to(m_cursor_pos)]
                     = to_upper(m_line[m_cursor_pos]);
+
+            to_next_word_end<Word>(m_cursor_pos, m_line);
         }
         else if (key == ctrl('k'))
         {
--- a/src/input_handler.cc
+++ b/src/input_handler.cc
@@ -385,21 +385,21 @@ class LineEditor
 public:
     void handle_key(Key key)
     {
-        if (key == Key::Left or key == alt('h'))
+        if (key == Key::Left or key == ctrl('b'))
         {
             if (m_cursor_pos > 0)
                 --m_cursor_pos;
         }
-        else if (key == Key::Right or key == alt('l'))
+        else if (key == Key::Right or key == ctrl('f'))
         {
             if (m_cursor_pos < m_line.char_length())
                 ++m_cursor_pos;
         }
-        else if (key == Key::Home)
+        else if (key == Key::Home or key == ctrl('a'))
             m_cursor_pos = 0;
-        else if (key == Key::End)
+        else if (key == Key::End or key == ctrl('e'))
             m_cursor_pos = m_line.char_length();
-        else if (key == Key::Backspace or key == alt('x'))
+        else if (key == Key::Backspace)
         {
             if (m_cursor_pos != 0)
             {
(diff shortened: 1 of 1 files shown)


lenormf:
How are we doing on this?

mawww:
Took a look, thats a big amount of commands, and frankly I am not sure thats the direction we want to take to make command line editing easier. I am still undecided on this.

lenormf:
The bindings that are implemented already are unsufficient to properly edit a command, and are counter-intuitive because they use the same keys used by `readline` but don't do the same things. So even if in the long term this PR is not what you want (#801 would be nice in my opinion), it will at least be useful to all of us (or maybe just me) who are used to the readline bindings.

lenormf:
If we're not going to have readline shortcuts, what functionalities do we want available in the prompt?

I'm pretty much interested in everything I implemented, but I guess it would help to have at least:

* move to the beginning/end of the line
* move forward/backward a word
* remove text to the beginning/end of the word

dlee:
This is so very needed. It is currently very painful to edit text in prompt and search modes.

I would even argue that these basic movement and deletion shortcuts are needed in insert mode for insertion-time edits that don't requirement losing your currently held selections.

KoFish:
IMHO this editor is useless without readline in insert/prompt. Could learn a new editing language but line editing when writing is just too useful.

lenormf:
I can understand why one wouldn't want any level of `readline` compatibility, so I think the best we can do now is make a list of the most needed operations and what kakoune-compliant shortcuts they would have.

dlee:
But whatever you do, please make them consistent across insert/prompt/search.

lenormf:
@dlee insert has no bindings to move around (that wouldn't fit kakoune's paradigm), and search uses regular expressions, what did you mean?

dlee:
@lenormf Sorry if I wasn't clear. I should have said "search prompt" rather than "search mode". I'm speaking of the various places where the user can input text by typing. Kakoune allows you to move around character-by-character with arrow keys and delete previous character with backspace while 1) typing into the buffer, 2) typing into the search prompt, 3) typing into the command prompt (there are probably other places as well).

I'm saying that while I'm inserting text, or typing search into the search prompt, or typing something into the command prompt; I want to move around word-wise and delete previous word and not just be restricted to moving character-wise and deleting previous character.

mawww:
@dlee no worries about prompts, its exactly the same code that handle line editing in all prompts, regardless of if its a search, command, object, user or whatever prompt.

For insert mode, I dont want to introduce another small editing language there, normal mode is better for that, and from insert mode, `<a-;>` allows you to enter a single normal mode command without quitting it.

The main reason I did not merge this CL is that I hope we can design a better way to edit the prompts, that would fit better with the rest of Kakoune, but we did not find a nice way yet.

lenormf:
Prompts as one-line-high always-in-insert-mode buffers!

doppioandante:
> Prompts as one-line-high always-in-insert-mode buffers!
That's actually something we should really think about IMO

dlee:
> For insert mode, I dont want to introduce another small editing language there,

There's already a small editing language in there (arrow keys, backspace, home, end). Although they're redundant and not necessary for editing text in kakoune, there's a reason those are there (presumably: easy, familiar, quick, universal). I think expanding that language to include word-wise editing (easy, familiar, quick, universal) would bring joy to users. They wouldn't interfere with other aspects of the editor, and having them would smooth out the learning curve.

> normal mode is better for that

I think the biggest case for insert-time editing shortcuts is for quick fixes of typos. Normal mode is better for navigation, large scale editing, or one-off changes, but for editing typos in insert mode, nothing's quicker and easier than insert-time editing. I would rather press backspace than `<esc>-d-i`. It's the same when I want to rub out the last word.

Maybe it would be more palatable if you considered it insertion-refinement rather than yet-another-text-editing language. The biggest use case for insert mode shortcuts is to fix the *insertion* not general editing of the buffer.

lenormf:
@doppioandante for having discussed this with @mawww already, I don't think this is coming any time soon as it would require a big overhaul of the code. But it would totally be the best if it happened.

alexherbo2:
How about implementing command-line mappings?

Consider the command-line like a buffer, starting in normal mode and returning to insert mode, single selection and reduce selection to its cursor.

``` kak
map global command-line <c-w> bd
```

``` kak
map global command-line <c-u> Ghd
```

And `<a-;>` command in command-line mode.

I noticed – unlike [Vim i_CTRL-O][] – it’s not possible to execute multiple commands.

``` kak
<a-;>:exec bd
```

``` kak
<a-;>:exec Ghd
```

[Vim i_CTRL-O]: http://vimdoc.sourceforge.net/htmldoc/insert.html#ins-special-special

martanne:
> Prompts as one-line-high always-in-insert-mode buffers!

That is similar to how I implemented it in [vis](https://github.com/martanne/vis/blob/master/vis-prompt.c). Except there is no need to restrict the prompt buffer to insert mode, it just starts in it when switching to the prompt. If a selection is active <kbd>Enter</kbd> treats it as the prompt input (e.g. a command to execute) otherwise the current line is used. These few special key bindings are the only difference between the prompt buffer and a "normal" one.

There is absolutely no need for an editor within an editor, you already have one so use it to edit the prompt content. This also gives you searchable command history for free.

lenormf:
The issue is that (apparently) the current codebase doesn't allow for an easy implementation of that feature.

danr:
@lenormf please tell me if you update and rebase this patch to current HEAD since I would like to use it in the meantime!

lenormf:
@danr I rebased the PR, but only tested a couple shortcuts to verify that things are working.

danr:
Thanks a lot, works fine!

net:
I think this feature is *expected* given how well Kakoune tries to integrate with Unix-like environments. The inconsistency between Kakoune's command prompt and other tools—including various shells—is frustrating.

At the very least, I think the missing movement and editing primitives should be introduced in such a way as to allow users to define readline/Emacs-like keybinds themselves.

Delapouite:
I think this is the current situation. Don't hesitate to tell me in case of mistake or if something's missing.

| action          |  readline   |kakoune   | map                              |
|-----------------|-------------|----------|----------------------------------|
| **char**        |             |          |                                  |
| ←char           |  `<c-b>`    |`<a-h>`   | `map global prompt <c-b> <a-h>`  |
| →char           |  `<c-f>`    |`<a-l>`   | `map global prompt <c-f> <a-l>`  |
| del char        |  `<c-d>`    |`<a-d>`   | `map global prompt <c-d> <a-d>`  |
| exchange char   |  `<c-t>`    |          |                                  |
| **word**        |             |          |                                  |
| ←word           |  `<a-b>`    |`<c-b>`   | `map global prompt <a-b> <c-b>`  |
| →word           |  `<a-f>`    |`<c-w>`   | `map global prompt <a-f> <c-w>`  |
| exchange word   |  `<a-t>`    |          |                                  |
| upper word      |  `<a-u>`    |          |                                  |
| lower word      |  `<a-l>`    |          |                                  |
| capitalize word |  `<a-c>`    |          |                                  |
| del end of word |  `<a-d>`    |          |                                  |
| del beg of word |  `<c-w>`    |          |                                  |
| **line**        |             |          |                                  |
| ←line           |  `<c-a>`    |`<home>`  | `map global prompt <c-a> <home>` |
| →line           |  `<c-e>`    |`<end>`   | `map global prompt <c-e> <end>`  |
| del end of line |  `<c-k>`    |          |                                  |
| del beg of line |  `<c-u>`    |          |                                  |

vbauerster:
IMO, cmd prompt should start in insert mode and hitting <kbd>esc</kbd>, should switch to normal mode, instead of quitting. Actually such behaviour is already implemented in zsh:

[](https://postimg.cc/image/cpbzklb4r/)
Whenever I press <kbd>esc</kbd>, zsh's `vicmd` mode is activated (note prompt arrow change), it is not full featured vi mode, but sufficient for basic edits, and if I feel like I need to do more complex edit, I can press `^x^e^` to launch visual editor.

PR #802, opened 2016-09-13
Fix for puppet tags

kurkale6ka:
Use an
`if/else` vs
`.. ? .. : ..`
because puppet tags for instance are most of the time classes with colons in their names

### DIFF
--- a/rc/base/ctags.kak
+++ b/rc/base/ctags.kak
@@ -16,8 +16,9 @@ def -params 0..1 \
         for tags in "$@"
         do
             [ -f "$tags" ] || continue
-            namecache="$(dirname "$tags")"/.kak."$(basename "$tags")".namecache
-            if [ "$namecache" -ot "$tags" ]; then
+            namecache="${tags%/*}"/.kak."${tags##*/}".namecache
+            if [ -z "$(find "$namecache" -prune -newer "$tags")" ]
+            then
                 cut -f1 "$tags" | grep -v \'^!_\' | uniq > "$namecache"
             fi
             cat "$namecache"
--- a/rc/base/ctags.kak
+++ b/rc/base/ctags.kak
@@ -7,13 +7,22 @@ decl str-list ctagsfiles 'tags'
 
 def -params 0..1 \
     -shell-candidates '
-        ( for tags in $(printf %s\\n "${kak_opt_ctagsfiles}" | tr \':\' \'\n\'); do
-              namecache=$(dirname ${tags})/.kak.$(basename ${tags}).namecache
-              if [ -z "$(find ${namecache} -prune -newer ${tags})" ]; then
-                  cat ${tags} | cut -f 1 | uniq > ${namecache}
-              fi
-              cat ${namecache}
-          done )' \
+        {
+        oldIFS="$IFS"
+        IFS=:
+        # $1=tag1, $2=tag2 ...
+        set -- $kak_opt_ctagsfiles
+        IFS="$oldIFS"
+        for tags in "$@"
+        do
+            [ -f "$tags" ] || continue
+            namecache="$(dirname "$tags")"/.kak."$(basename "$tags")".namecache
+            if [ "$namecache" -ot "$tags" ]; then
+                cut -f1 "$tags" | grep -v \'^!_\' | uniq > "$namecache"
+            fi
+            cat "$namecache"
+        done
+        }' \
     -docstring 'Jump to tag definition' \
     tag \
     %{ %sh{
@@ -21,7 +30,6 @@ def -params 0..1 \
 
         oldIFS="$IFS"
         IFS=:
-        # $1=tag1, $2=tag2 ...
         set -- $kak_opt_ctagsfiles
         IFS="$oldIFS"
 
@@ -35,7 +43,7 @@ def -params 0..1 \
                 then
                     [ "$tags" != tags ] && continue
                 fi
-                echo -n "$tagroot/	"; readtags -t "$tags" "$tagname"
+                printf %s "$tagroot/	"; readtags -t "$tags" "$tagname"
             fi
         done | awk -F '\t|\n' '
         /[^\t]+\t[^\t]+\t\/\^.*\$?\// {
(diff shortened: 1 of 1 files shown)


lenormf:
This whole chunk actually has to go because it will print an error if a tag couldn't be found in a single file, even when there are other tag files to parse afterwards.

kurkale6ka:
Not sure I follow. I haven't changed the logic. I've just used a different yet equivalent construct.

lenormf:
I'm not talking about your changes specifically, but the bit in that `awk` script that prints an error. If there are several tag files to be parsed, an error will be displayed for all the files that do not contain the tag being looked up (even if the tag might indeed be in one of the files).

kurkale6ka:
yes, you are right

mawww:
Hello,

I dont really get what that change fixes, it seems stricly equivalent, what am I missing ?

kurkale6ka:
So, the problem is I am dealing with puppet tags such as:
```
% readtags -t tags servers::crontabs                                                   
servers::crontabs       modules/servers/manifests/crontabs.pp   /^class servers::crontabs {$/
```
Because of the `::` in the class name awk's ternary operator is getting confused.

Strangely the behaviour has changed. Whereas before `:tag servers::crontabs` wouldn't bring me at all to the tag now it does but it also prints the error message:
screenshot_2016-09-21_09-42-52

So either the class name need quoting I guess or I thought it as simple to just use an equivalent `if/else` construct to avoid the `:` issue

kurkale6ka:
So, I found at least why I get this message ^ (no such tag...). This is exactly what @lenormf pointed in his very first comment.

My setting goes like this: `set global ctagsfiles 'tags:/path/to/second/tags:/path/to/third/tags'` and that tag ^ (`servers::crontabs`) is only present in the second tags file, not the third one. There.

Still not sure why I can't reproduce the initial awk error I was getting and that was the reason for me submitting this PR. Guess it must've been something to do with my environment

kurkale6ka:
The above commit fixes 3 issues:
- no error for tags with white spaces in their names (including path). that's the `IFS/"$@"` hack
- stop erroneous reporting of non existing tags
- show menu when tag exists in several locations

Note: other portions of the code still raise errors like:
```
find: ‘tags’: No such file or directory
cat: tags: No such file or directory
```
That's because we want the `ctagsfiles` option to start with `'tags:...:...'` but obviously that file isn't always existing.

rc/base/ctags.kak
```diff
@@ -17,19 +17,36 @@ def -params 0..1 \
...
+        do
+            if [ -f "$tags" ]
+            then
+                tagroot="$(readlink -f "$(dirname "$tags")")"
+                # Avoid reading the same tag twice
+                if [ "$tagroot" = "$PWD" ] || [ "$tagroot" = "$(pwd -P)" ]
+                then
+                    [ "$tags" != tags ] && continue
+                fi
+                echo -n "$tagroot/	"; readtags -t "$tags" "$tagname"
```
lenormf: We need a `printf %s` here.

rc/base/ctags.kak
```diff
@@ -17,19 +17,36 @@ def -params 0..1 \
...
+            fi
+        done | awk -F '\t|\n' '
+        /[^\t]+\t[^\t]+\t\/\^.*\$?\// {
+            tagroot = $1
+            re=$0;
+            sub(".*\t/\\^", "", re); sub("\\$?/$", "", re); gsub("(\\{|\\}|\\\\E).*$", "", re);
+            keys=re; gsub(/</, "<lt>", keys); gsub(/\t/, "<c-v><c-i>", keys);
+            out = out " %{" re " {MenuInfo}" tagroot $3 "} %{eval -collapse-jumps %{ try %{ edit %{" tagroot $3 "}; exec %{/\\Q" keys "<ret>vc} } catch %{ echo %{unable to find tag} } } }"
```
lenormf: Can you add the `-color Error` flag to the last `echo` call please ?

lenormf:
>Note: other portions of the code still raise errors

In that case could you fix those errors (add a simple test to check that the path/file exists) ? Since you're modifying the file, might as well make the whole thing work.

rc/base/ctags.kak
```diff
@@ -7,29 +7,54 @@ decl str-list ctagsfiles 'tags'
...
-        ( for tags in $(printf %s\\n "${kak_opt_ctagsfiles}" | tr \':\' \'\n\'); do
-              namecache=$(dirname ${tags})/.kak.$(basename ${tags}).namecache
-              if [ -z "$(find ${namecache} -prune -newer ${tags})" ]; then
-                  cat ${tags} | cut -f 1 | uniq > ${namecache}
-              fi
-              cat ${namecache}
-          done )' \
+        {
+        oldIFS="$IFS"
+        IFS=:
```
kurkale6ka: You've got a problem when your `ctagsfiles` has white spaces in its path (ex: `/path/to/folder with spaces/tags`). It's ugly but hey, we are dealing with `sh` here! :). Plus we restore IFS so it's fine. I don't think we have many alternatives.

rc/base/ctags.kak
```diff
@@ -7,29 +7,54 @@ decl str-list ctagsfiles 'tags'
...
+        # $1=tag1, $2=tag2 ...
+        set -- $kak_opt_ctagsfiles
+        IFS="$oldIFS"
+        for tags in "$@"
+        do
+            [ -f "$tags" ] || continue
+            namecache="$(dirname "$tags")"/.kak."$(basename "$tags")".namecache
+            if [ "$namecache" -ot "$tags" ]; then
```
kurkale6ka: Yes, that's a pity. I realised after

rc/base/ctags.kak
```diff
@@ -16,8 +16,9 @@ def -params 0..1 \
         for tags in "$@"
         do
             [ -f "$tags" ] || continue
-            namecache="$(dirname "$tags")"/.kak."$(basename "$tags")".namecache
-            if [ "$namecache" -ot "$tags" ]; then
+            namecache="${tags%/*}"/.kak."${tags##*/}".namecache
```
lenormf: Actually there are some cases that `dirname` handles that are not replaceable with a shell expansion (like returning `.` when the argument is a simple filename), maybe this first subshell should have stayed?

rc/base/ctags.kak
mawww: In which case, why do we need oldIFS for ? we know IFS is supposed to be the default one right ? so we can just unset IFS to restore it.

mawww:
Hello,

I went another way, those IFS were bothering me too much, as well as the tags deduplication code that was only handling a single, hard coded case (the plain 'tags' file in the list). I hope I did not miss some other features.

Thanks a lot for this PR.

Maxime.

PR #819, opened 2016-09-24
autowrap: Adapt to recent changes to `exec`'s behavior, remove non POSIX double…

lenormf:
… brackets, change the scope of the hooks

### DIFF
--- a/rc/base/autowrap.kak
+++ b/rc/base/autowrap.kak
@@ -16,18 +16,20 @@ def -hidden autowrap-cursor %{ eval -save-regs '/"|^@m' %{
             reg m "%val{selections_desc}"
 
             ## if we're adding characters past the limit, just wrap them around
-            exec -draft "<a-h><a-k>.{%opt{autowrap_column}}\h*[^\s]*<ret>1s(\h+)[^\h]*\'<ret>c<ret><esc>"
+            exec -draft "<a-h><a-k>.{%opt{autowrap_column}}\h*[^\s]*<ret>1s(\h+)[^\h]*\'<ret>c<ret>"
         } catch %{
             ## if we're adding characters in the middle of a sentence, use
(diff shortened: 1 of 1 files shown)


rc/base/autowrap.kak
```diff
@@ -37,9 +39,9 @@ def -hidden autowrap-cursor %{ eval -save-regs '/"|^@m' %{
 } }
 
 def autowrap-enable -docstring "Wrap the lines in which characters are inserted" %{
-    hook -group autowrap window InsertChar [^\n] autowrap-cursor
+    hook -group autowrap buffer InsertChar [^\n] autowrap-cursor
```
mawww: Why buffer ? seems window is more adapted there.

lenormf: I thought that hooks that modify the buffer would be better declared in the `buffer` scope, was I wrong ? Tbh I was never really clear on the difference between the `buffer` and `windows` scope (even if I get roughly what they refer to because of their name), maybe the scopes should have their own documentation page ?

mawww: well, if we put the hook in window scope, it means the user can setup some window to autowrap, with others not autowrapping, if we put the hook in the buffer, the windows are forced to follow the same behaviour.

Its more flexible to put that in window scope.

lenormf: That's the thing, when I thought about it, I it looked like I would typically want a file to have it's contents wrapped, not having two windows open on the same file and have only one that has wrapping activated.

mawww: I wonder if we even need that autowrap-enable command, we could just let users put that hook themselves, on the scope they prefer.

But if we provide a command, I'd rather have it put hook on the current window scope, rather than on the buffer, as I'd rather leave the user choose (he can just always do for all windows if he wants to).

PR #820, opened 2016-09-24
spell: Fix non POSIX instructions, print errors when they occur

lenormf:


rc/base/spell.kak
```diff
@@ -9,6 +9,10 @@ def -params ..1 spell -docstring "Check spelling of the current buffer with aspe
         printf %s\\n "set buffer spell_tmp_file ${file}"
     }
     %sh{
+        _echo() {
```
mawww: Not a big fan on introducing this function, I think I prefer seeing printf direct.y

lenormf: Calls to `printf` tend to clutter the lines unecessarily, so whenever there's more than a couple calls to it, I create an `_echo` function. Users know what the function does, it's 7 byte less long than its `printf` counterpart, and it simplifies the code.

Why don't you want this function in ? Not enough calls to `printf` to replace with it ?

somasis: It just seems like a pointless abstraction to me... additionally, don't most of these printf calls not even need a `\n` anyway?

mawww: In all the cases where you use _echo, we can just use normal echo safely, as we always know the single argument we pass is not starting with '-', I'd prefer if you removed that _echo function and just used built-in echo.

lenormf: Expanding a variable with arbitrary content is unsafe with `echo`, it's not just about minus signs (backslashes matter as well), and since I don't want to have to think about whether a variable can contain those characters constantly, I use `printf`. Just like when I have to use a function with a hardcoded format string more than a couple times, I abstract it in `_echo` so that I don't litter the code with a parameter that is irrelevant (we only want to print a stupid string).

mawww: I still dont like that _echo, can we just use printf there, please ?

lenormf:
Ready.

PR #822, opened 2016-09-24
flake8: Rewrite some non POSIX conditionals

lenormf:


lenormf:
@Kareeeeem could you test that PR to verify that it's not breaking anything please ?

PR #826, opened 2016-09-25
Add some standard GNU targets to the Makefile

lenormf:


src/Makefile
```diff
@@ -82,35 +83,51 @@ kak : $(objects)
 	gzip -f $(@:.gz=.1)
 	mv -f $(@:.gz=.1.gz) $@
 
+check: test
 test:
```
mawww: I dont like having these two targets just for the sake of having the GNU one as well. We could either keep test, or rename it to check. I'd favor test, which seems more explicit, but I dont feel very strongly about that.

In general, following the GNU conventions just to follow them does not seems very important, they are nice to follow to stay consistent with other software, but only when they make sense.

lenormf: Actually I was about to completely get rid of the original target names and only stick to the GNU ones, but then I thought it would break package maintainers' scripts out of nowhere for no real reason, so I just let those targets point to each other.

I would rather have standard target names, because it's always a pain to deal with software that make their own rules when people try to include them in e.g. automated builds.

I deal with entire distribution builds on a regular basis, and there's nothing worse than having to patch your scripts (or the software's code directly) when you want to include a program in your distribution but the project doesn't follow any standard regarding Makefile targets, autoconf files etc. So, on the scale of a single user, those targets are fine and following the GNU convention isn't important to them, but it is for people who expect the hundreds of projects that they want to chain-build to have at least the most important targets implemented (`all`, `check`, `install`, `distclean`).

Issue #806, opened 2016-09-15
Buffer Delete vs Delete Buffer


bhajneet:
There is only one command that begins with "del", instead if someone were trying to find commands related to ":buffer" they would not see the "delbuf" or "db" command. I would personally think bd, bufdel, bufferdelete makes more sense.

lenormf:
Commands all have semanticaly the same form (except for `buffernext` and `bufferprev` which I think were named that way because they are standard commands across all editors): first a verb indicating what action is made, and then the subject that action applies to.

Examples: `cd` (change directory), `nameclient`, `namesession`, `rmhooks`, `addhl` etc.

That makes discovering commands intuitively harder in theory, but the fuzzy matching and the `:doc commands` command make up for that.

bhajneet:
I see that verb-subject pattern in a lot of them, that makes sense too.

What about those that are grouped up like autowrap-enable / disable? There are others too under clang, jedi, tmux but I figure those are because they are external perhaps.

Also double-verbs like grep-next, prev, jump (Or would grep be considered a subject?). Buffer can be used as a verb too, idk.

Regardless, if you want delbuf, you should probably add deletebuffer then, so that when someone types "buffer" they see the relevant commands, regardless of naming scheme. Some might not type "buf" only, or might not know if there are other commands starting with "buf". Since the commands are few, it's not a big deal, just wanted to point it out.

lenormf:
Non-builtin commands have the name of the "module" they were declared in as prefix, and then a verb usually follows.

Feel free to make a pull request if you want to take it up with @mawww, I personally only use the aliases to builtin commands so name changes won't impact me.

mawww:
Hi, got a quick moment with an internet connection to add my 2 cents,

So, frankly, all that naming did not follow much logic, it was whatever felt ok whenever I named the commands. I actually think it would be nice to do a proper pass on all the commands, try to get more consistency. I'd go even further than commands, options naming would also benefit a bit from more consistency.

Regarding the commands, now that we have nice auto completion, I'd favor longer and more expressive command names, using `-` as a word separator, for example, `cd` would be an alias to `change-directory`, `addhl` would be an alias to add-highlighter (or maybe change the alias to `ah` ?...)

Thats my current idea (that is, this evening, after a long day hiking and swimming), I'd need to think a bit more about it (as often, the existing commands/behaviours are good enough, and have the advantage of being the known ones, so that tends not to be very high priority).

Cheers,

Maxime.

[...]

PR #827, opened 2016-09-25
Rename commands to fit a more consistent naming scheme

lenormf:
Hi,

As discussed in #806, the naming scheme of the commands can feel a bit inconsistent, so I decided to make my proposal for new names.

This PR does not modify anything other than the hardcoded names in `commands.cc`, as I think we should all agree (and by all I mean @mawww) on what names to use before changing the scripts.

Discuss.

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -339,7 +339,7 @@ void write_all_buffers()
 }
 
 const CommandDesc writeall_cmd = {
-    "writeall",
+    "write-all",
     "wa",
     "write all buffers that are associated to a file",
     no_params,
@@ -372,8 +372,8 @@ static void ensure_all_buffers_are_saved()
 }
 
 const CommandDesc kill_cmd = {
-    "kill",
-    nullptr,
+    "kill-session",
+    "ks",
     "kill current session, quit all clients and server",
     no_params,
     CommandFlags::None,
@@ -387,8 +387,8 @@ const CommandDesc kill_cmd = {
 
 
 const CommandDesc force_kill_cmd = {
-    "kill!",
-    nullptr,
+    "kill-session!",
+    "ks!",
     "kill current session, quit all clients and server, do not check for unsaved buffers",
     no_params,
     CommandFlags::None,
@@ -440,8 +440,8 @@ void write_quit(const ParametersParser& parser, Context& context,
 }
 
 const CommandDesc write_quit_cmd = {
+    "write-quit",
     "wq",
-    nullptr,
     "write current buffer and quit current client",
     no_params,
     CommandFlags::None,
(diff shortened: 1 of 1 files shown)


src/commands.cc
```diff
@@ -982,9 +982,9 @@ void define_command(const ParametersParser& parser, Context& context, const Shel
 }
 
 const CommandDesc define_command_cmd = {
+    "define",
```
mawww: shall we use a more explicit name ? `command` ?

src/commands.cc
```diff
@@ -1251,9 +1251,9 @@ const CommandDesc unset_option_cmd = {
 };
 
 const CommandDesc declare_option_cmd = {
+    "declare",
```
mawww: A more explicit name ? `option` ?

src/commands.cc
```diff
@@ -667,9 +667,9 @@ Highlighter& get_highlighter(const Context& context, StringView path)
 }
 
 const CommandDesc add_highlighter_cmd = {
+    "add-highlighter",
```
mawww: or just `highlight` ?

lenormf:
>I wonder the kind of commands we need, change-buffer vs buffer, hook or add-hook...

That's the issue I hit when changing the names, and I've decided not to go completely for one naming format, even if that made the changeset inconsistent.

On one hand you have commands whose name is just a verb (`alias`, `map` etc), those are easy to remember and the name of their removal counter part simply starts with "un" (`unalias`, `unmap` etc). I could have made it so they all followed the `action-object` name format, but it seemed rather unnecessary and cumbersome (`add-alias`, `add-map` etc).

But on the other hand there are commands whose removal counter part would not sound so great if they followed the `verb` only convention: `unhook`, `unhighlight`… So since the name for the removing commands had to be prefixed with `remove-`, I made the regular command start with `add-`. That's why we don't have `highlight`/`unhighlight` or `hook`/`unhook`, but `add-highlighter`/`remove-highlighter` and `add-hook`/`remove-hooks` (the latter should be renamed to `remove-hook` by the way, even though the current command name is `rmhooks`).

The corner case of all this is `buffer`, I hesitated before changing its name because the entire world is so used to that command. But since this PR is based on the remark that it's not very complicated to intuitively find the correct command name with fuzzy matching, I think it's the right decision to steer away from the `vim` model.

src/commands.cc
lenormf: I think it's better if we keep this command name a verb, in the name of consistency.

src/commands.cc
lenormf: C.f. [my answer](https://github.com/mawww/kakoune/pull/827#issuecomment-251585573).

Delapouite:
I also prefer a consistent naming scheme over vim old names.

lenormf:
Should this be closed now that new names have been merged? Should we move on to option names?

mawww:
Names have been changed through some other commits, thanks for the inspiration 👍

PR #840, opened 2016-10-02
Add a `colorscheme` option, POSIX changes

lenormf:
Hi,

I need to be able to select what faces I use in my configuration depending on the colorscheme that's being used, so I added a `colorscheme` option.

I also replaced some calls to `echo` with their `printf` counterpart.

HTH.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -924,6 +924,9 @@ Some options are built in Kakoune, and can be used to control it's behaviour:
  * `modelinefmt` _string_: A format string used to generate the mode line, that
    string is first expanded as a command line would be (expanding `%...{...}`
    strings), then markup tags are applied (see <<Markup strings>>).
+ * `colorscheme` _string_:
+   Name of the colorscheme that was passed to the 'colorscheme' command,
+   without any '.kak' extension
  * `ui_options`: colon separated list of key=value pairs that are forwarded to
    the user interface implementation. The NCurses UI support the following options:
    - `ncurses_set_title`: if `yes` or `true`, the terminal emulator title will
--- a/doc/manpages/options.asciidoc
+++ b/doc/manpages/options.asciidoc
@@ -110,7 +110,7 @@ Builtin options
 		provides completion based on local filesystem
 
 	*option=<opt-name>*:::
-		where *opt-name* is a 'completions' option.
+		where *opt-name* is a 'completions' option
 
 *static_words* 'str-list'::
 	list of words that are always added to completion candidates
@@ -118,7 +118,7 @@ Builtin options
 
 *completions_extra_word_chars* 'str'::
 	a string containing all additional character that should be considered
-	as word character for the purpose of insert mode completion.
+	as word character for the purpose of insert mode completion
 
 *autoreload* 'enum(yes|no|ask)'::
 	auto reload the buffers when an external modification is detected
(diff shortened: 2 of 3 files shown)


mawww:
I am not fond of that colorscheme option, dont you already know what colorscheme you are using ? (its in your kakrc, right ?)

lenormf:
I want to have a generic way of picking the faces used in `modelinefmt`, and for that I need a variable that tells me what theme I'm using.

mawww:
Cant `modelinefmt` use standard faces that will get set by the colorscheme ?

lenormf:
I want to allow people who use my script to pick an arbitrary color, and not only have a choice between the faces used by the current theme.

lenormf:
Am I closing this PR? Or are you still undecided?

lenormf:
Rebased on `master`, also added documentation for the option.

mawww:
I still dont like that option, colorschemes are self contained, having that option means external scripts know the content of certain colorschemes and react accordingly. That seems wrong to me, Scripts should use colors provided by colorscheme, not add colors themselves.

lenormf:
What if my scripts wants to use colors that are not parts of any face? Or even if that was the case, what if a particular face has attributes that the script doesn't want to use?

mawww:
Your script is going to need a big switch with a set of colorschemes it knows about. I dont want to encourage that, I think scripts should not care about the actual colorscheme, and use one of the defined faces.

It might be that we need to add a few generic faces to colorschemes for other uses. Popular scripts could also introduce faces that are specific to them, with colorscheme deciding to add, or not, special support for that script (the `GitBlame` face from `git-tools.kak` comes to mind).

lenormf:
>Your script is going to need a big switch with a set of colorschemes it knows about.

That's exactly what my script does, as it's the only way to allow users to specifiy arbitrary colors.

>It might be that we need to add a few generic faces to colorschemes for other uses. Popular scripts could also introduce faces that are specific to them, with colorscheme deciding to add, or not, special support for that script (the GitBlame face from git-tools.kak comes to mind).

This centralizes everything and is not very generic/scalable, if scripts have a way of knowing what colorscheme is being used then there's no dependency on anything (as opposed to relying on faces declared in colorscheme scripts).

mawww:
> >   Your script is going to need a big switch with a set of colorschemes it knows about.

> That's exactly what my script does, as it's the only way to allow users to specifiy arbitrary colors.

Yep, and thats what I dont like at all, it means scripts end up knowing about various colorschemes, trying to add colors that match them, missing any new colorscheme. I think thats a wrong approach. case I think a few generic faces provided by all colorscheme would cater for most scripting uses. Actually, why cant you use one of the generally available faces in your use case ? What are you trying to highlight ?

> This centralizes everything and is not very generic/scalable ...

Thats true, and its the same problem with the colorscheme option, either each colorscheme knows about N scripts, or each scripts knows about N colorschemes. Which is why I'd favor using generic faces, if the existing ones are not enough we could add a few. That way we could decouple colorscheme and scripts.

lenormf:
>Actually, why cant you use one of the generally available faces in your use case ? What are you trying to highlight ?

I would like to use the `value` face to highlight things in my modeline, but use another face when using `base16` because the `value` face used in `base16` doesn't look good in the modeline. Also the `value` face is bold by default in `zenburn`, and I don't want the text to highlight to be bold. So I've declared faces in my script, whose colors are set in a switch over the name of the colorscheme.

mawww:
For the status line, we already have 4 faces: StatusLine, StatusLineMode, StatusLineInfo and StatusLineValue, I am surprised you have a use case that needs a 5th one.

Its the colorscheme job to put together a set of colors that work well, trying to outsmart it seems wrong. What you describe ought to either be fixed in the color scheme (if colors objectively dont work well together, for example are hard to read), or if its just a matter of taste, in your kakrc/personal scripts.

In these two cases, it does not justify a general colorscheme option.

lenormf:
There's no face that would highlight correctly some atoms in the modeline across all colorschemes, and even if that was the case sometimes those faces have attributes I don't want.

mawww:
I still dont get it: StatusLine, StatusLineMode, StatusLineInfo and StatusLineValue are 4 faces that *have* to work well in the modeline, they are meant for that, if that is not the case, the colorscheme should be fixed.

If there are attributes you dont like in the colorscheme, and you want to patch them, you are free to do it, you can even introduce your own option, and wrap colorscheme with another command + alias that will set it, but that does not justify adding a colorscheme option to upstream.

lenormf:
I want the atoms to stand out, so I'm using a face that's not part of the regular UI. The colorschemes are all fine, it's just that some of them have default attributes on faces that I would use if I didn't have the switch.

leemachin:
This is interesting.

> Popular scripts could also introduce faces that are specific to them, with colorscheme deciding to add, or not, special support for that script (the GitBlame face from git-tools.kak comes to mind).

This is how emacs does it and I think it works fairly well, given that you cannot know everything in advance about what people want to customise. You would set a custom face in your script, and many themes set colours for the various faces other scripts define. This means that your colourscheme is still 100% responsible for styling the editor and your script is merely setting a sensible default for itself. The people who care about the theme can maintain it if popular packages declare new faces.

PR #848, opened 2016-10-08
Remove "-lang" suffix from D and Go

casimir:
Since the language name is Go, see [golang.org](http://golang.org) for reference. This is really just a proposition for consistency.
Impacts #851.

Edit: Apply the same to D as suggested by https://github.com/mawww/kakoune/pull/848#issuecomment-252464610

### DIFF
--- a/rc/base/d.kak
+++ b/rc/base/d.kak
@@ -6,13 +6,13 @@
 
 hook global BufCreate .*\.di? %{
     set buffer mimetype ""
-    set buffer filetype dlang
+    set buffer filetype d
 }
 
 # Highlighters
 # ‾‾‾‾‾‾‾‾‾‾‾‾
 
-addhl -group / regions -default code dlang \
+addhl -group / regions -default code d \
     string '"' (?<!\\)(\\\\)*" '' \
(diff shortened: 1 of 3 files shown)


lenormf:
You might as well rename `dlang` to `d` as well then.

casimir:
A quick google search for D blog post shows this is a good idea. But like I know Go Inreally don't know D. Maybe a user could validate this?

Either way I'll add a second commit for this today.

lenormf:
I'm a dlang user.

mawww:
I dont mind much between the two, can you get to an agreement ?

casimir:
Since there's no ambiguity in this context I'm for the merge. It's also what you find in a lot of editors.
@lenormf any thought?

greduan:
In this case, since there is no ambiguity in this context and I would personally guess that a filetype is named `go` not `golang`, I'm OK with this merge.

lenormf:
Filetypes for C/C++/ObjC are c/cpp/objc, so I guess it makes sense to apply the same logic to  Go and D. I don't mind either way.

PR #851, opened 2016-10-09
Extended Go support

casimir:
This is my attempt to add Go tooling support to kakoune. This is still a work in progress but I'll gladly accept any feedback.

## In progress
- [x] autocompletion with [gocode](https://github.com/nsf/gocode)
  - [x] `go-(enable|disable)-autocomplete`
- [x] format on save
  - [x] working with goimports
  - [x] allow to choose `gofmt` or `goimports` (command arguments are different)
- [x] documentation display
  - [x] symbol documentation in info box with [gogetdoc](github.com/zmb3/gogetdoc)
  - [x] echo the signature of the symbol under the cursor
- [x] jump to symbol
- [x] share the current selection with the Go Playground

## Note and future work
- gometalinter since it doesn't really fit with kakoune's idea, you can look at [golint-kak](https://github.com/casimir/golint-kak) for a replacement.
- IDE features will be implemented in a future PR (e.g.: guru)
- Big documentation (mostly package doc) take too much space and won't be rendered in an info box.

### DIFF
--- a/rc/extra/go-tools.kak
+++ b/rc/extra/go-tools.kak
@@ -1,0 +1,182 @@
+# Provides integration of the following tools:
+# - gocode for code completion (github.com/nsf/gocode)
+# - goimports for code formatting on save 
+# - gogetdoc for documentation display and source jump (needs jq) (github.com/zmb3/gogetdoc)
+# Needs the following tools in the path:
+# - jq for json deserializaton
+
+# Auto-completion
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+decl -hidden str go_complete_tmp_dir
+decl -hidden completions gocode_completions
+
+def go-complete -docstring "Complete the current selection with gocode" %{
+    %sh{
+        dir=$(mktemp -d -t kak-go.XXXXXXXX)
+        printf %s\\n "set buffer go_complete_tmp_dir ${dir}"
+        printf %s\\n "eval -no-hooks write ${dir}/buf"
+    }
+    %sh{
+        dir=${kak_opt_go_complete_tmp_dir}
+        (
+            gocode_data=$(gocode -f=godit --in=${dir}/buf autocomplete ${kak_cursor_byte_offset})
+            rm -r ${dir}
+            column_offset=$(echo "${gocode_data}" | head -n1 | cut -d, -f1)
+
+            header="${kak_cursor_line}.$((${kak_cursor_column} - $column_offset))@${kak_timestamp}"
+            compl=$(echo "${gocode_data}" | sed 1d | awk -F ",," '{print $2 "||" $1}' | paste -s -d: -)
+            printf %s\\n "eval -client '${kak_client}' %{
+                set buffer=${kak_bufname} gocode_completions '${header}:${compl}'
+            }" | kak -p ${kak_session}
+        ) > /dev/null 2>&1 < /dev/null &
+    }
+}
+
+def go-enable-autocomplete -docstring "Add gocode completion candidates to the completer" %{
+    set window completers "option=gocode_completions:%opt{completers}"
+    hook window -group go-autocomplete InsertIdle .* %{ try %{
+        exec -draft <a-h><a-k>[\w\.].\z<ret>
+        go-complete
+    } }
+    alias window complete go-complete
+}
+
+def go-disable-autocomplete -docstring "Disable gocode completion" %{
+    set window completers %sh{ printf %s\\n "'${kak_opt_completers}'" | sed 's/option=gocode_completions://g' }
+    rmhooks window go-autocomplete
+    unalias window complete go-complete
+}
+
+# Auto-format
+# ‾‾‾‾‾‾‾‾‾‾‾
+
+decl -hidden str go_format_tmp_dir
+
+def -params ..1 go-format \
+    -docstring "go-format [goimports]: custom formatter for go files" %{
+    %sh{
+        dir=$(mktemp -d -t kak-go.XXXXXXXX)
+        printf %s\\n "set buffer go_format_tmp_dir ${dir}"
+        printf %s\\n "eval -no-hooks write ${dir}/buf"
+    }
+    %sh{
+        dir=${kak_opt_go_format_tmp_dir}
+        if [ "$1" = "1" ]; then
+            fmt_cmd="goimports -srcdir '${kak_buffile}'"
+        else
+            fmt_cmd="gofmt -s"
+        fi
+        eval "${fmt_cmd} -e -w ${dir}/buf 2> ${dir}/stderr"
+        if [ $? ]; then
+            cp ${dir}/buf "${kak_buffile}"
+        else
+            # we should report error if linting isn't active
+            printf %s\\n "echo -debug '$(cat ${dir}/stderr)'"
+        fi
+        rm -r ${dir}
+    }
+    edit!
+}
+
+def go-enable-format-onsave -docstring "Enable formatting on save for go files" %{
+    hook buffer -group go-format-onsave BufWritePre .+\.go %{ go-format }
+}
+
+def go-enable-format-imports-onsave -docstring "Enable formatting (with goimports) on save for go files" %{
+    hook buffer -group go-format-onsave BufWritePre .+\.go %{ go-format 1 }
+}
+
+def go-disable-format-onsave -docstring "Disable formatting on save for go files" %{
+    rmhooks buffer go-format-onsave
+}
+
+# Documentation
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+decl -hidden str go_doc_tmp_dir
+
+# FIXME text escaping
+def -hidden -params 1..2 _gogetdoc-cmd %{
+   %sh{
+        dir=$(mktemp -d -t kak-go.XXXXXXXX)
+        printf %s\\n "set buffer go_doc_tmp_dir ${dir}"
+        printf %s\\n "eval -no-hooks write ${dir}/buf"
+    }
+    %sh{
+        dir=${kak_opt_go_doc_tmp_dir}
+        (
+            printf %s\\n "${kak_buffile}" > ${dir}/modified
+            cat ${dir}/buf | wc -c >> ${dir}/modified
+            cat ${dir}/buf >> ${dir}/modified
+
+            if [ "$2" = "1" ]; then
+                args="-json"
+            fi
+            output=$(cat ${dir}/modified                                                    \
+		| gogetdoc $args -pos "${kak_buffile}:#${kak_cursor_byte_offset}" -modified \
+		| sed 's/%/%%/g')
+            rm -r ${dir}
+            printf %s "${output}" | grep -v -q "^gogetdoc: "
+            status=$?
+
+            case "$1" in
+                "info")
+                    if [ ${status} -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


lenormf:
There's a `format` command that you can customize to set the default formatter for go buffers.

casimir:
I know about this command but I struggled a lot more that needed with it and couldn't make it work within a `BufWritePre` hook. This was before https://github.com/mawww/kakoune/commit/d50087eabee0fa0067a5f44462782332230bdf3e so maybe I could try again.

casimir:
@lenormf it is working now, nice catch.

rc/extra/go-tools.kak
```diff
@@ -45,17 +45,83 @@ def go-disable-autocomplete -docstring "Disable gocode completion" %{
...
+    %sh{
+        dir=$(mktemp -d -t kak-go.XXXXXXXX)
+        printf %s\\n "set buffer go_format_tmp_dir ${dir}"
+        printf %s\\n "eval -no-hooks write ${dir}/buf"
+    }
+    %sh{
+        dir=${kak_opt_go_format_tmp_dir}
+        goimports -e -srcdir ${kak_buffile} -w ${dir}/buf 2> ${dir}/stderr
+        if [ $? ]; then
+            cp ${dir}/buf ${kak_buffile}
```
lenormf: Don't forget to wrap variables with double/simple quotes whenever possible.

rc/extra/go-tools.kak
```diff
@@ -45,17 +45,83 @@ def go-disable-autocomplete -docstring "Disable gocode completion" %{
...
+}
+
 # FIXME works only for functions outside of stdlib
 # FIXME text escaping
 def go-doc-info -docstring "Show the documention of the symbol under the cursor" %{ %sh{ (
-    doc=$(gogetdoc -pos "${kak_buffile}:#${kak_cursor_byte_offset}" | sed "s/%/%%/g;s/'/\\'/g")
+    doc=$(gogetdoc -json -pos "${kak_buffile}:#${kak_cursor_byte_offset}" | sed "s/%/%%/g;s/'/\\'/g")
     printf %s\\n "eval -client '${kak_client}' %{
-        info -anchor ${kak_cursor_line}.${kak_cursor_column} '${doc}'
+        info -anchor ${kak_cursor_line}.${kak_cursor_column} %@${doc}@
```
lenormf: Why change this line ?

rc/extra/go-tools.kak
casimir: Is `$kak_buffile` allowed to contains whitespaces?

rc/extra/go-tools.kak
casimir: `$doc` can contains `'`, this is the least used char I found as delimiter.

casimir:
@lenormf thanks for the feedback! I answered inline.

rc/extra/go-tools.kak
lenormf: It's a path, so it can contain anything. Usually we assume that it won't contain double quotes (otherwise we would end up escaping things everywhere).

rc/extra/go-tools.kak
```diff
@@ -45,17 +45,83 @@ def go-disable-autocomplete -docstring "Disable gocode completion" %{
...
+            set buffer filetype make
+            set buffer _make_current_error_line 0
+            hook -group fifo buffer BufCloseFifo .* %{
+                nop %sh{ rm -r ${dir} }
+                rmhooks buffer fifo
+            }
+        }"
+        (
+            gometalinter --sort=severity --fast --tests $(dirname ${kak_buffile}) > /${dir}/stderr
+            flags=$(cat /${dir}/stderr | sed -nE                                 \
```
lenormf: Relying on undocumented features is a bad idea.

casimir:
Whereas it's a nice tool `gometalinter` doesn't really fit with the way kakoune handles files, therefore I drop the idea of supporting it in this script. cc @lenormf

pauldub:
This is nice! I use a custom integration of `godef` to show type hints in the statusbar like this http://i.imgur.com/CqlA2Wh.png using the following snippet.

```kak
def -allow-override godef %{
	%sh{
		output=$(mktemp -d -t kak-temp-XXXXXXXX)/fifo
		mkfifo ${output}
		(godef -t -f ${kak_buffile} -o ${kak_cursor_byte_offset} > ${output}) > /dev/null 2>&1 < /dev/null &
		result=$(cat $output | grep -v '^/' | awk '{$1=$1};1')
		if [ result != '' ]; then
    		echo "echo $result"
		fi
    }
}

hook global WinSetOption filetype=go %{
	hook window NormalIdle .* -group go-hooks %{ try %{ godef } }
}
```

It looks like el-doc from emacs, which maybe interesting to have?

lenormf:
@pauldub This seems rather specific, usually tools that are supported builtin are either officially supported ones, or those that are used by an overwhelming amount of people in the community, so that we don't end up having code that's used by a fraction of kakoune users.

@casimir You could provide support for this tool in an `extra/gometalinter.kak` script, it will be useful.

casimir:
@pauldub This is possible with a lot of different tools. Do you want to have every definition always? Isn't it overkill? I use a binding to show doc and signature when I need it, I find it sufficent.
Still, I know some people like the feature of vim-go does that. If needed I can add a command that activate the same behaviour.

@lenormf I lost the will to support it, instead I made a tool specific to kakoune that wraps the 2 official linters: https://github.com/casimir/golint-kak

pauldub:
@casimir @lenormf I'm only used to always have type definitions in the status bar and I only look at it when I need to, it is a habit I have from emacs' go-eldoc :) If there is demand for it I would be happy to contribute it.

lenormf:
Quick question: do you have tools in Go that allow to have more information about a symbol that just a prototype (arguments, types, declaration file)? If that's possible, maybe you could create a special buffer in which all this info is written, and have that buffer open in a separate client (but within the same session), IDE style?

pauldub:
@lenormf There is go oracle which  answers questions such as:

> What is the type of this expression?  What are its methods?
> What’s the value of this constant expression?
> Where is the definition of this identifier?
> What are the exported members of this imported package?
> What are the free variables of the selected block of code?
> What interfaces does this type satisfy?
> Which concrete types in the package implement this interface?
> And:
> Where are the references to this identifier?
> What are the possible concrete types of this interface value?
> What are the possible callees of this dynamic call?
> What are the possible callers of this function?
> What objects might this pointer point to?
> Where will a value sent on this channel be received?
> Which statements could update this field/local/global/map/array/etc?
> Which functions might be called indirectly from this one?

casimir:
As @pauldub said there's one:
https://docs.google.com/document/d/1_Y9xCEMj5S-7rv2ooHpZNH15JgRT5iM742gJkw5LtmQ/mobilebasic
It has a pretty could support in some editors, I think we can borrow the
way emacs implements it (see gif in the doc).

That being said guru doesn't handle documentation fetching. Also it does a
lot so the integration need some work. For these reason I planned to make
another PR for this when this one is merged.

After that I'd say the Go support is complete. May later refactoring tools
or more exotic things could be added.

*tl;dr* there is one that has its place in `go-tools.kak` but this is
planned for the next PR.

lenormf:
Very nice, if something comes out of the idea I posted earlier, maybe an API could be extracted and the concept adapted to other languages in a semi-generic way.

rc/extra/go-tools.kak
```diff
@@ -80,9 +80,15 @@ def go-disable-format-onsave -docstring "Disable formatting on save for go files
 # FIXME use the -modified flag
 def go-doc-info -docstring "Show the documention of the symbol under the cursor" %{ %sh{ (
     doc=$(gogetdoc -pos "${kak_buffile}:#${kak_cursor_byte_offset}" | sed "s/%/%%/g;s/'/\\'/g")
-    printf %s\\n "eval -client '${kak_client}' %{
-        info -anchor ${kak_cursor_line}.${kak_cursor_column} %@${doc}@
-    }" | kak -p ${kak_session}
+    if [ $(printf %s ${doc} | grep -c import) = 1 ]; then
```
lenormf: I think this could have been written as:
```
if printf %s "${doc}" | grep -q import; then
```

casimir:
I think I've done everything I wanted in this PR. ping @mawww

rc/extra/go-tools.kak
```diff
@@ -0,0 +1,182 @@
...
+            # we should report error if linting isn't active
+            printf %s\\n "echo -debug '$(cat ${dir}/stderr)'"
+        fi
+        rm -r ${dir}
+    }
+    edit!
+}
+
+def go-enable-format-onsave -docstring "Enable formatting on save for go files" %{
+    hook buffer -group go-format-onsave BufWritePre .+\.go %{ go-format }
```
mawww: I dont think these two commands (`go-enable-format-...` are worth it, they are single lines, and the user will use a hook to call them anyway, I'd say let the user put a hook that calls go-format himself.

rc/extra/go-tools.kak
```diff
@@ -0,0 +1,182 @@
...
+                    ;;
+        	"jump")
+                    if [ ${status} -eq 0 ]; then
+                        pos=$(printf %s "${output}" | jq -r .pos)
+                        file=$(printf %s "${pos}" | cut -d: -f1)
+                        line=$(printf %s "${pos}" | cut -d: -f2)
+                        col=$(printf %s "${pos}" | cut -d: -f3)
+                        printf %s\\n "eval -client '${kak_client}' %{
+                            eval -try-client '${kak_opt_jumpclient}' edit -existing ${file} ${line} ${col}
+                            try %{ focus '${kak_opt_jumpclient}' }
```
mawww: Big trailing whitespaces here, there are a few others lying around.

rc/extra/go-tools.kak
casimir: Like for completion I like the expressiveness of calling a function with an expressive name in my `kakrc`. It easy to see what feature is enabled for a filetype. I tried to ease the choice of which program to use for formatting too.
This is my preference but I'll follow your decision on this. Should I still remove it?

casimir: I could also provide 2 formatting functions (`go-format`, `go-format-imports`) and let the user write the hook. Is it a better solution?

mawww: Yeah, I'd rather let user write hooks because that make them discover Kakoune, hooks are a bit tricky but they allow for a good understanding of how Kakoune is designed.

Here for example, the user still needs to write a hook anyway, to call `go-enable-format-onsave` on the buffer (so maybe on BufSetOption). 

regarding `go-format-imports` you could just name use `-only-imports` or `-imports` in place of `1`.

casimir: I'm convince. I made a new proposition for the format command and removed the others.

vbauerster:
@casimir Was the `go-guru` tool integrated? If not, are there any plans? I also like the way it works in emacs, so looking forward for some IDE style.

casimir:
@vbauerster guru wasn’t integrated. About the IDE-style features do you have a feature in mind? I found the current implementation for my need so I didn’t carry on enhancing this script.

vbauerster:
@casimir Yes, I'd like to see auto symbol highlighter, similar to vim-go (go_auto_sameids) and emacs (go-guru-hl-identifier-mode). Guru describe is must have, possibly showing its output in dedicated tmux pane or popup. Maybe it could be easy to integrate, but I'm lacking expertise at the moment.

PR #856, opened 2016-10-10
Only create a scratch buffer when no data is written on stdin

### LINKED ISSUE #852: *scratch* buffer is not necessary when running Kakoune with redirected stdin
How about to not create the `*scratch*` buffer in that case?, since the `*stdin*` buffer is there already.

lenormf:
Closes #852


lenormf:
Why were so many changes necessary by the way?

PR #862, opened 2016-10-11
[RDY] Add generic linter

kurkale6ka:


rc/core/sh-lint.kak
```diff
@@ -0,0 +1,88 @@
+# ex: '-e SC2087'
+decl str sh_options
+
+decl -hidden line-flags sh_flags
+decl -hidden str sh_errors
+
+def sh-parse -params 0..1 -docstring 'Parse the contents of the current buffer with shellcheck' %{
```
lenormf: It doesn't look like this function takes any parameter, or does it?

rc/core/sh-lint.kak
```diff
@@ -0,0 +1,88 @@
...
+        first_line=-1
+        while read -r line_content; do
+            candidate="${line_content%%,*}"
+            if [ -n "$candidate" ]
+            then
+                first_line=$(( first_line == -1 ? candidate : first_line ))
+                line=$((candidate > kak_cursor_line && (candidate < line || line == -1) ? candidate : line ))
+            fi
+        done
+        line=$((line == -1 ? first_line : line))
```
lenormf: This part is a bit ugly, I fixed it in [the original file](https://github.com/mawww/kakoune/blob/master/rc/extra/flake8.kak#L80).

kurkale6ka: cool, I will amend tomorrow

mawww:
I think you should generally rename all that from sh to shellcheck, and rename the file shellcheck.kak, as what this file does is integrating shellcheck with kakoune.

rc/core/sh-lint.kak
```diff
@@ -0,0 +1,88 @@
...
+        { # do the parsing in the background and when ready send to the session
+
+        shellcheck -fgcc ${kak_opt_sh_options} "$dir"/buf > "$dir"/stderr
+        printf '%s\n' "eval -client $kak_client echo 'shellcheck parsing done'" | kak -p "$kak_session"
+
+        # Update the gutter with errors and warnings: line3|{red}:line11|{yellow}
+        flags=$(awk -F: '
+                    /:[0-9]+:([0-9]+:)? (fatal )?error/ { print $2"|{red}█" }
+                    /:[0-9]+:([0-9]+:)? (warning|note)/ { print $2"|{yellow}█" }
+                ' "$dir"/stderr | paste -s -d ':')
```
mawww: maybe we can get rid of paste and do that logic directly in the awk script ?

rc/core/sh-lint.kak
```diff
@@ -0,0 +1,88 @@
...
+
+        # 1,err1
+        # 2,err2
+        # ...
+        errors=$(awk -F: '
+                    /:[0-9]+:([0-9]+:)? ((fatal )?error|warning|note)/ { print $2","substr($4,2)":"$5 }
+                 ' "$dir"/stderr | sort -n)
```
mawww: similarly, maybe we can join these two awk scripts to build two different variables, (and maybe even the two option set commands) in one go. (Not sure its worth it, maybe it will clean up the code, maybe not, so its really up to you).

rc/core/sh-lint.kak
kurkale6ka: Hey, btw, the `then` on that line is missing :)

kurkale6ka:
Actually not sure about the renaming. What if next there is a better tool than `shellcheck`. I wanna use it. All I have to do is adapt the line where I invoke `shellcheck`

lenormf:
inb4 let's make a generic `lint` command.

kurkale6ka:
Note: this script works well with `puppet-lint --log-format "%{filename}:%{line}:%{column}: %{kind}: %{message} [%{check}]"`

casimir:
@kurkale6ka are you sure about the generic part?

kurkale6ka:
@casimir it works for at least 2 checkers. Check this (from my kakrc):
```
hook global WinSetOption filetype=sh %{                                                                         
    set buffer linter 'shellcheck -fgcc'                                                                        
    lint-enable                                                                                                 
}                                                                                                               
                                                                                                                
hook global WinSetOption filetype=puppet %{                                                                     
    set buffer linter 'puppet-lint --log-format "%{filename}:%{line}:%{column}: %{kind}: %{message} [%{check}]"'
    lint-enable                                                                                                 
}                                                                                                               
                                                                                             
map global user n ':lint-next<ret>'                                              
```

casimir:
What is your linter doesn't support output formatting? And if it takes the path of file's directory instead of the file's one as argument?

casimir:
More specifically I would like to now if #851 can be based on this PR.

kurkale6ka:
yeah, it will only work if your linter can output the right way. but a few can so that's already an improvement. for all the others a custom linter kak file will need creating

casimir:
Vim plugins like syntastic and neomake take the output regex in the linter config (something like `linter_ouput`). Would it be possible to add this while keeping the current regex as default?

Tell me if I'm wrong but if I want to add a linter with a non supported output I just have to redefine the `lint` command?

kurkale6ka:
> Tell me if I'm wrong but if I want to add a linter with a non supported output I just have to redefine the lint command?

that's right. you would only have to change this line:

`eval "$kak_opt_linter $kak_opt_linter_options '$dir'/buf" > "$dir"/stderr`

and the whole `awk` block

lenormf:
In theory you could declare an overridable function whose only job is to do the eval bit, so that other scripts can override it and format the output of their linter with an `awk` script.

kurkale6ka:
I experimented with a `lint-output` function following @lenormf's suggestion but couldn't make it work:

`printf '%s\n' "eval -client $kak_client lint-output $dir/buf $dir/stderr" | kak -p "$kak_session"`
instead of the current
`eval "$kak_opt_linter $kak_opt_linter_options '$dir'/buf" > "$dir"/stderr` and
```
hook global WinSetOption filetype=sh %{

    lint-enable
    def -hidden -allow-override -params 2 lint-output %{ %sh{
        shellcheck -fgcc "$1" > "$2"
    }}
}
```
in my kakrc

casimir:
@kurkale6ka last minute feature request. Since lot of linters provide not only the line but also the column, could you add it in the jump command? Something like `<line>g <col>l`.

rc/core/lint.kak
```diff
@@ -80,18 +80,23 @@ def lint-disable -docstring "Disable automatic diagnostics of the code" %{
...
             fi
         done
         if [ "$candidate" -gt "$kak_cursor_line" ]
         then
-            printf '%s\n' "exec $candidate g"
+            col="${coords#*,}"
+            printf '%s\n' "exec $candidate g\; $((col-1)) l"
         else
-            printf '%s\n' "exec ${kak_opt_lint_errors%%,*} g"
+            col=`printf '%s\n' "$kak_opt_lint_errors" | head -n1 | cut -d, -f2`
```
casimir: For coherence with the rest of the script change backquotes to `$(cmd)`

rc/core/lint.kak
```diff
@@ -1,6 +1,6 @@
 # your linter should output in this format:
 # {filename}:{line}:{column}: {kind}: {message}
-decl str linter
+decl str lintercmd
```
lenormf: This option would also contain flags and arguments, so `linter_options` can be removed now.

mawww:
Looks good now, I am not sure we need to preserve the whole history, this could be squashed in a single commit.

One last change, this should not be in `rc/core` but in `rc/base` or `rc/extra`.

I did not find a copyright waiver from you in the current git log, could you prepend one (see the CONTRIBUTING file)

lenormf:
So, no default linters for the languages supported? I always hear people who use `vim` say that "Kakoune doesn't even support syntastic", and that syntastic "just works out of the box". How far can we go to invalidate those two statements?

lenormf:
Can you remove `linter_options` and rename `lintercmd` to `lintcmd` please?

lenormf:
When I use it, I don't see any marker in the sidebar, on the lines where there's an error. There's also no small window popping up with the error message, I have to look into the output buffer to know what the error is (that's all after using `lint-next`).

Also it would be nice to sort the error messages by line, to guarantee that we won't jump up and down when using `lint-next`.

Thanks.

lenormf:
It looks like I mispoke, the `lint-enable` function does everything I mentioned, but maybe it should be renamed? I was under the impression (just by reading the docstring) that it would automatize the linting process, not additionally highlight parts of the buffer.

PR #865, opened 2016-10-13
Document the non inclusiveness of the end column in range faces options

### LINKED ISSUE #864: Incoherence with the `range-faces` descriptor end column index
Hi,

As pointed out by @occivink, it doesn't make sense that the end column index isn't inclusive in the descriptor for a `range-faces` options:
```
*range-faces*::
	a `:` separated list of a pairs of a buffer range
	(`<begin line>.<begin column>,<end line>.<end column>`) and  a face
	(separated by `|`), except for the first element which is just the
	timestamp of the buffer.
```

Since the end line parameter is inclusive, it'd be better to make the end column one inclusive as well. Validating test that should highlight the first character of the first line:
```
0:1.1,1.1|Error
```

Thanks.

lenormf:
Changing the behavior of the option in the code would imply changes in scripts, so I just documented this behavior.

Maybe it will change later on, but that's not too important for now as this behavior exists because of the translation between data and C++ iterators (it's not a random mistake).

Closes #864

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -824,7 +824,8 @@ Options are typed, their type can be
  * `range-faces`: a `:` separated list of a pairs of a buffer range
    (`<begin line>.<begin column>,<end line>.<end column>`) and  a face
    (separated by `|`), except for the first element which is just the
-   timestamp of the buffer.
+   timestamp of the buffer. Note that the end column is not inclusive,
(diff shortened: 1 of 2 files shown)


lenormf:
This could be merged temporarily, until a decision is taken in #864.

PR #867, opened 2016-10-13
Add a dedicated documentation page about scopes

lenormf:
Hi,

I feel like the documentation could use a dedicated page about scopes, so here it is.

I also want to specify what scopes some variables/hooks are ignored in certain contexts, but it's going to need more work.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -871,12 +871,15 @@ Some options are built in Kakoune, and can be used to control it's behaviour:
  * `scrolloff` _coord_: number of lines,columns to keep visible around
    the cursor when scrolling.
  * `eolformat` _enum(lf|crlf)_: the format of end of lines when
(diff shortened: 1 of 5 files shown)


Delapouite:
Thanks for this page.
It clarifies things for me. I never paid pay too much attention to scopes yet, but intuitively I thought that the buffer scope had priority over the window one, but it appears it's the other way around.

mawww:
Looks good to me, any other changes you want to make before merging ?

lenormf:
Yes I wanted to talk about the reach of the hooks, but I think a simple line who be enough given their names (e.g. the `BufWrite` buffers don't consider the `window` scope because they are buffer related actions).

lenormf:
I think that's all, should be ready for review/merge.

PR #870, opened 2016-10-14
Remove `formatcmd` and comment options from default scripts, fix `commenting.kak`

lenormf:
Hi,

The details are in the commits. Now the next big question is: do we declare default formatters in `formatting.kak`, as we do for commenting sequences in `commenting.kak`? The two cases are similar despite this major difference: commenting sequences in languages do not change, but users have a choice between several formatters, so do we take the liberty of giving a default value?

### DIFF
--- a/rc/core/commenting.kak
+++ b/rc/core/commenting.kak
@@ -1,21 +1,48 @@
-## Characters that will be used to surround a selection with
-decl str-list comment_selection_chars "/*:*/"
-
 ## Characters that will be inserted at the beginning of a line to comment
-decl str comment_line_chars "//"
+decl str comment_line_chars "#"
+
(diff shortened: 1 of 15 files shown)


rc/core/commenting.kak
```diff
@@ -4,9 +4,53 @@ decl str-list comment_selection_chars "/*:*/"
...
+    set buffer comment_line_chars '--'
+    set buffer comment_selection_chars '--[[:]]'
+}
+
+hook global BufSetOption filetype=moon %{
+    set buffer comment_line_chars '--'
+    set buffer comment_selection_chars ''
+}
+
+hook global BufSetOption filetype=(perl|makefile) %{
```
kurkale6ka: can we add kak and sh to this list

PR #871, opened 2016-10-15
[WIP] Introduce individual unit tests for default scripts

lenormf:
Hi,

There's no much to this single-commit PR, I just wanted to have something to show to prompt a discussion on how to proceed forward. I've had in mind to create unit tests for all the function provided by the project by default, and there's several obstacles to that that will need addressing:

* some scripts required specific applications to have been installed to be run, how do we implement dependency checks in the `run` script
* do the tests that don't have the executables they need to run to fail without being marked as "FAILED" (another status could be created)
* should the format of the `in` file in the tests be changed to simplify calls to commands, in order not to have to do things like `:<command><ret>`

HTH.


mawww:
To filter out test, we could have the run script check for a 'skip' file appearing, and the rc file could do something like `%sh{ which jq || touch skip }`

You can as well call the commands directly in the 'rc' script, instead of using `:command` in the `cmd` file

doppioandante:
`command -v` is the POSIX for `which`.

lenormf:
>To filter out test, we could have the run script check for a 'skip' file appearing, and the rc file could do something like %sh{ which jq || touch skip }

In the end I decided to make a `dependencies` script that will allow each test to check themselves the dependencies they need. The `skip` approach required modifying the template of kakoune commands passed to `kak`, which didn't really look good.

>You can as well call the commands directly in the 'rc' script, instead of using :command in the cmd file

I wanted to keep commands to be tested in the `cmd` file, so I dedicated the `rc` file only to contextual things e.g. dependency sources or variable initializations

lenormf:
Terminology wise, are builtin scripts to be refered to as "scripts", or is it OK to start talking about "modules"? Sometimes "script" is a bit vague.

lenormf:
I'm still hitting this issue with asynchronous commands whose result is not taken into account, is there really no way to stall the editor for long enough for a background shell command to complete?

mawww:
We cannot easily guess if there exists a process that might end up connecting to the control socket and wait for that to eventually happen.

We could however rely on a switch/an option value to tell us that we should not quit immediately.

Not sure exactly how to put that in place though.

PR #878, opened 2016-10-18
Empty the `mimetype` variable of buffers with a static extension

lenormf:
Hi,

Got tired of setting `filetype` manually when opening some files, so here you go. Details in the commit.

HTH.

### DIFF
--- a/rc/base/git.kak
+++ b/rc/base/git.kak
@@ -1,4 +1,5 @@
 hook global BufCreate .*COMMIT_EDITMSG %{
+    set buffer mimetype ""
     set buffer filetype git-commit
 }
 
--- a/rc/base/mail.kak
+++ b/rc/base/mail.kak
@@ -1,5 +1,9 @@
 hook global BufSetOption mimetype=message/rfc822 %{ set buffer filetype mail }
 
+hook global BufCreate .+\.eml %{
+    set buffer mimetype "message/rfc822"
+}
+
 addhl -group / group mail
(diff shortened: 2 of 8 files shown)


lenormf:
How we doing in here?

lenormf:
RFC!

mawww:
I'd like to do a general cleanup of filetype detection, between `file` command, extensions based detection, shebang detection...

What we have currently is pretty messy, hard to predict, and we are using hacks to tie it all together (like setting mimetype to an empty value).

We should have a well defined detection sequence, thats predictible and clean, this PR goes further into the hacks I'd like to get rid of.

lenormf:
I understand, what do you plan on replacing all that with?

However, in the mean time this PR is needed for files that contain code from different languages, so users still don't have proper filetype detection by default without it.

casimir:
I don't know if it's the same root cause but I just had a 300 loc python file with a filetype set as `cpp`.

lenormf:
That's `file` shitting the bed. And since the extension doesn't have the priority over the mimetype (which this PR fixes), if `file` is wrong, then you end up with a wrong filetype.

mawww:
I wonder in what cases the `file` command is really useful now, I think we are at a point where it does more harm than good, and we could just get rid of `mimetype` simplifying all filetype scripts. At least, we should change the logic so that `file` is just a fallback, and that should not require dirty hacks like setting mimetype to empty.

lenormf:
Mimetypes are very handy when opening files that don't have extensions, they are a good compliment to hooks that check file extensions, e.g. when downloading and editing a pastebin (you just get a raw file).

lenormf:
Superseeded by #949.

PR #883, opened 2016-10-20
Add a spell replace function

kurkale6ka:
Proposed map: `map global user = '<a-i>w:spell-replace<ret>'`

### DIFF
--- a/rc/base/spell.kak
+++ b/rc/base/spell.kak
@@ -55,7 +55,7 @@ Formats of language supported:
     }
 }
 
-def spell_replace %{%sh{
+def spell-replace %{%sh{
     suggestions=$(echo "$kak_selection" | aspell -a | grep '^&' | cut -d: -f2)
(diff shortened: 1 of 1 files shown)


kurkale6ka:
It'd be good to have words red underlined btw (vs in red background)

mawww:
We cant have underlining with another color from fg on terminals unfortunately.

casimir:
What about using bold as emphasis rather than underline? Would it allow another color from fg?

lenormf:
Aren't those things customizable already (faces) ?

rc/base/spell.kak
```diff
@@ -54,3 +54,14 @@ Formats of language supported:
         } </dev/null >/dev/null 2>&1 &
     }
 }
+
+def spell_replace %{%sh{
```
mawww: currently, commands use `-` as a word separator, not `_`

PR #892, opened 2016-10-22
Add missing Go keyword

casimir:
Add `var` to the keywords, per https://golang.org/ref/spec#Keywords.


lenormf:
`var` isn't an attribute though, maybe it would make more sense to put it in the `keywords` list?

casimir:
Right but neither is `const`. Both are a keywords for declaration just like `func`.

lenormf:
In theory yes, but having all the keywords highlighted as keywords clutters the buffers visually, that's why  I declared an attribute category even though it doesn't exist according to the documentation.

casimir:
I see, should have read the rest of the script :roll_eyes:. Still `const` and `var` are really close in concept, immutable vs mutable variable. They should be highlighted the same way IMO.

PR #902, opened 2016-10-30
Add a gallery, remove the scripts page, translate in French

lenormf:


Delapouite:
Nice, thanks. The images in the gallery are a bit blurry in their current size, it's hard to see what's going on in the screenshot. It should be fixable by adding a link to open them fullscreen.

lenormf:
I noticed this as well, it was implemented in the original version but the `a` tag got lost during a rebase. I'll fix this soon.

PR #907, opened 2016-11-02
Change the faces used to highlight certain shell atoms

lenormf:
This commit replaces the `identifier` face with others, in order to
diversify the amount of colors used to highlight shell scripts.

### DIFF
--- a/rc/core/sh.kak
+++ b/rc/core/sh.kak
@@ -36,7 +36,7 @@ addhl -group /sh/code regex [\[\]\(\)&|]{1,2} 0:operator
 addhl -group /sh/code regex (\w+)= 1:identifier
 addhl -group /sh/code regex ^\h*(\w+)\h*\(\) 1:identifier
 
-addhl -group /sh/code regex \$(\w+|\{.+?\}|#|@|\?|\$|!|-|\*) 0:identifier
+addhl -group /sh/code regex \$(\w+|\{.+?\}|#|@|\?|\$|!|-|\*) 0:value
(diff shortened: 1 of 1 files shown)


rc/core/sh.kak
```diff
@@ -33,10 +33,10 @@ addhl -group /sh/comment fill comment
 }
 
 addhl -group /sh/code regex [\[\]\(\)&|]{1,2} 0:operator
-addhl -group /sh/code regex (\w+)= 1:identifier
+addhl -group /sh/code regex (\w+)= 1:attribute
```
mawww: This is an identifier, why change its highlighting to something else ?

rc/core/sh.kak
```diff
@@ -33,10 +33,10 @@ addhl -group /sh/comment fill comment
...
 addhl -group /sh/code regex ^\h*(\w+)\h*\(\) 1:identifier
 
-addhl -group /sh/code regex \$(\w+|\{.+?\}|#|@|\?|\$|!|-|\*) 0:identifier
+addhl -group /sh/code regex \$(\w+|\{.+?\}|#|@|\?|\$|!|-|\*) 0:value
```
mawww: idem

lenormf: You could make the argument that expansions are values, especially subshells. I want shell scripts to be highlighted with more than two colors (string/identifier), I think this is a valid change.

rc/core/sh.kak
lenormf: Maybe I got carried away changing this, but it makes highlighting so much nicer:


Maybe this shouldn't changed, in hindsight.

PR #916, opened 2016-11-07
Update go keywords

casimir:
Add missing builtins and remove `interface` from the keywords since it is in the types already.

cf https://golang.org/pkg/builtin/

### DIFF
--- a/rc/base/go.kak
+++ b/rc/base/go.kak
@@ -29,7 +29,8 @@ addhl -group /go/code regex %{-?([0-9]*\.(?!0[xX]))?\b([0-9]+|0[xX][0-9a-fA-F]+)
 %sh{
     # Grammar
     keywords="break|default|defer|else|fallthrough|for|func|go|goto|if|import"
-    keywords="${keywords}|interface|make|new|package|range|return|select|case|switch|type|continue"
+    keywords="${keywords}|make|new|package|range|return|select|case|switch|type|continue"
+    keywords="${keywords}|append|cap|close|complex|copy|delete|imag|len|panic|print|println|real|recover"
     attributes="const|var"
     types="bool|byte|chan|complex128|complex64|float32|float64|int|int16|int32"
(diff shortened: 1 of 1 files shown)


lenormf:
I had intentionally left those out of the list of keywords, as they are not keywords, and they might also make the buffers unreadable if too many builtin functions are used (many things would be highlighted as keywords).

This isn't the first time we want to highlight something but are lacking a category, maybe a new builtin face should be added to the themes (for instance to highlight builtin functions)?

casimir:
Builtin are parts of the language specification and have special meaning. Syntax highlight is the only visual difference with function call. Note that I don't want to highlight dynamic things like function definition/call and other things like that in this PR. Plus why take only `new` and `make` and leave the others builtins?

I know a lot of modern vim themes provides different categories for keyword and builtin. Although a builtin type would be nice, I don't mind having the same time, I see builtins as functional keywords.

To be clear I don't like working in a rainbow buffer but visual hints on builtins are really important to read the flow of the code where in the loop do I `append`, where do I `panic`, etc.

From my personal view, 2 examples (only the `len` builtin in this code but you see the idea):

*Not good, rainbow clusterfuck - atom, base16 tomorrow night*
screen shot 2016-11-07 at 14 51 33

*Good (don't get why `func` has its own color though) - nvim/vim-go, papercolor*
screen shot 2016-11-07 at 14 52 42

lenormf:
I only added `new` and `make` because they looked more like keywords than operations like `append`, `copy` etc. The exact same dilemn happened with the python support script, with functions like `len`.

@mawww let's add a default face to all existing colorschemes to highlight builtin functions instead of either highlighting them as keyword or not highlighting them at all.

casimir:
Like keywords, builtins are present in almost every languages. Maybe this is better if we just add them to highlighters and don't make assumption about their importance or usage.

Adding a default face is a good idea, if someone is willing to do it. A lot of scripts will have to be updated, maybe this PR could be merged so that `go.kak` could be part of this refactoring.

mawww:
I'll merge that for now, and an eventual project wide introduction of builtin functions could tweak it afterwards.

PR #924, opened 2016-11-09
Add column information

kurkale6ka:


lenormf:
Cheers.

rc/base/lint.kak
```diff
@@ -41,7 +41,7 @@ def lint -docstring 'Parse the current buffer with a linter' %{
                 }
             }
             /:[0-9]+:[0-9]+:/ {
-                errors = errors $2 "," $3 "," substr($4,2) ":"
+                errors = errors $2 "," $3 ",c" $3 " " substr($4,2) ":"
```
mawww: So, we just prepend 'c<column> ' in front of the error description ? that `c` looks strange, but I dont have any totally satisfying alternative, maybe `:<column> ` would be closer to the traditional error format.

alexanderjeurissen:
Could you elaborate what this PR does ? the motivation for creating it etc. ?

lenormf:
The issue is that when a linter reports several issues on the same line (or even one on a large line), it can be hard to identify what part triggers the error, so the column number fixes that.

PR #926, opened 2016-11-11
Introduce the `builtin` face, modify scripts accordingly

lenormf:
Hi,

The `function` face was defined as just bold text, it should work fine with most themes. I used it to highlight keywords in several languages (C, C++, D, Go, Python) that have default symbols that are not quite keywords, but still need to be highlighted as they're predefined by the standard.

The C/C++ case is special, I only highlighted macros defined by the preprocessor or the compilers.

HTH.

### DIFF
--- a/rc/base/d.kak
+++ b/rc/base/d.kak
@@ -35,26 +35,36 @@ addhl -group /d/code regex "\b(this)\b\s*[^(]" 1:value
 
 %sh{
     # Grammar
-    keywords="alias|asm|assert|body|cast|class|delegate|delete|enum|function"
-    keywords="${keywords}|import|in|interface|invariant|is|lazy|mixin|module"
-    keywords="${keywords}|new|out|pragma|struct|super|typeid|typeof|union"
-    keywords="${keywords}|unittest|__parameters|__traits|__vector|break|case"
-    keywords="${keywords}|catch|continue|default|do|else|finally|for|foreach"
-    keywords="${keywords}|foreach_reverse|goto|if|return|switch|throw|try|with|while"
+
+    keywords="abstract|alias|align|asm|assert|auto|body|break|case|cast"
+    keywords="${keywords}|catch|cent|class|const|continue|debug"
+    keywords="${keywords}|default|delegate|delete|deprecated|do|else|enum|export|extern"
+    keywords="${keywords}|final|finally|for|foreach|foreach_reverse|function|goto"
+    keywords="${keywords}|if|immutable|import|in|inout|interface|invariant"
+    keywords="${keywords}|is|lazy|macro|mixin|module|new|nothrow|out|override"
+    keywords="${keywords}|package|pragma|private|protected|public|pure|ref|return|scope"
+    keywords="${keywords}|shared|static|struct|super|switch|synchronized|template"
+    keywords="${keywords}|throw|try|typedef|typeid|typeof|union"
+    keywords="${keywords}|unittest|version|volatile|while|with"
     attributes="abstract|align|auto|const|debug|deprecated|export|extern|final"
     attributes="${attributes}|immutable|inout|nothrow|package|private|protected"
-    attributes="${attributes}|public|pure|ref|override|scope|shared|static|synchronized|version|__gshared"
-    types="bool|byte|cdouble|cfloat|char|creal|dchar|double|dstring|float"
+    attributes="${attributes}|public|pure|ref|override|scope|shared|static|synchronized|version"
+    attributes="${attributes}|__gshared|__traits|__vector|__parameters"
+    types="bool|byte|cdouble|cent|cfloat|char|creal|dchar|double|dstring|float"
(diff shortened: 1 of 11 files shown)


casimir:
While I like the introduction of this face I find the name misleading. Given the name I'd guess this face apply to every function call, depending on the mood even function declaration, which is not the case. Does it cover more than builtins (aka function keywords)?

lenormf:
I too dislike the name, but I don't have any alternative to propose. Since it's not visible to users, I went with it to have a PoC ready, leaving us the option to change it at a later time if we ever find a better name.

rc/core/c-family.kak
```diff
@@ -196,6 +196,21 @@ addhl -group /cpp/code regex %{\b-?(0x[0-9a-fA-F]+|\d+)[fdiu]?|'((\\.)?|[^'\\])'
     "
 }
 
+# c and c++ compiler macros
```
mawww: Almost all of these are non standard, I dont think we should support them.

rc/core/c-family.kak
```diff
@@ -145,7 +145,7 @@ def -hidden _c-family-insert-on-newline %[ eval -draft %[
 }
 
 # c specific
-addhl -group /c/code regex %{\bNULL\b|\b-?(0x[0-9a-fA-F]+|\d+)[fdiu]?|'((\\.)?|[^'\\])'} 0:value
+addhl -group /c/code regex %{\b-?(0x[0-9a-fA-F]+|\d+)[fdiu]?|'((\\.)?|[^'\\])'} 0:value
```
mawww: Why remove NULL here ? The fact that its a macro is an implementation detail, NULL means the null pointer value in C.

lenormf: Forgot to add highlighting for values in the shell scope.

rc/core/c-family.kak
lenormf: There are no standard pre-processing macro anyway, I thought I would highlight those provided by `gcc` (most of those are supported by `clang` as well), since they are fairly common.

I could trim this list down to the bare minimum (`__FILE__`, `__LINE__` and the likes) if needed, I didn't think it wuold be an issue since I'm using the `function` face and not -say- the `keywords` one.

lenormf: Fixed in f0524bd.

lenormf:
About the name of the face, do we have something more appropriate?

casimir:
@lenormf the more I think about it the more I come back to `builtin`. What were the arguments against it again?

lenormf:
`function` was the named suggested by @mawww when I mentioned the feature, so I went with it. Maybe builtin is a bit broad, and I originally presented the face as being used to highlight default functions in python and go.

casimir:
What about more specific name like `builtinFunction`?

mawww:
I dont mind much between `function` and `builtin`, I think we should define what is it this is supposed to highlight. Taking a look at python, seems the doc references 'built-in functions' and 'built-in constants', the latter would be highlighted as 'values', so if we want just to support built-in functions, `builtin` seems ok.

I think `builtinFunction` does not fit well with all other defined faces, I'd prefer a single word face name.

Taking a look at this commit, it seems `function` is rarely used for actual functions (builtin or not). So, as I think we should not go above 10 standard faces for code (we have 9 as of now), I'd favor `builtin` as its both vague (so could be apply for different things depending on the language), and different from existing faces (`function` and `identifier` are overlapping for example).

(Yeah, this comment is a bit "train of thought").

rc/base/d.kak
```diff
@@ -35,21 +35,31 @@ addhl -group /d/code regex "\b(this)\b\s*[^(]" 1:value
...
     types="${types}|idouble|ifloat|int|ireal|long|ptrdiff_t|real|size_t|short"
-    types="${types}|string|ubyte|uint|ulong|ushort|void|wchar|wstring"
-    values="true|false|null|__FILE__|__MODULE__|__LINE__|__FUNCTION__"
-    values="${values}|__PRETTY_FUNCTION__|__DATE__|__EOF__|__TIME__"
-    values="${values}|__TIMESTAMP__|__VENDOR__|__VERSION__"
+    types="${types}|string|ubyte|ucent|uint|ulong|ushort|void|wchar|wstring"
+    values="true|false|null"
+    tokens="__FILE__|__MODULE__|__LINE__|__FUNCTION__"
+    tokens="${tokens}|__PRETTY_FUNCTION__|__DATE__|__EOF__|__TIME__"
+    tokens="${tokens}|__TIMESTAMP__|__VENDOR__|__VERSION__|#line"
```
lenormf: Those are called tokens, and are the equivalent of preprocessing macros except they are defined in the language. `#line` is a [special token sequence](https://dlang.org/spec/lex.html#special-token-sequence), which I consider also as a preprocessing macro.

lenormf:
Renamed `function` to `builtin`, rebased on `master`. It's ready for merging,

rc/core/c-family.kak
mawww: gcc_macros is a wrong name for that.

PR #932, opened 2016-11-19
Enhance latex.kak

wheatdog:
Enhance the syntax highlight of the latex filetype.

1. Comments that don't start at the beginning.
1. Escape symbols like \\%, \\_, \\$.


mawww:
Thanks !

PR #941, opened 2016-11-28
go: add `error` to types

casimir:
For whatever reason I though this was included in #926. Looks like it wasn't...


lenormf:
`error` is a type, not a keyword, it was lost in the process when I modified the lists.

casimir:
What a dumbass. I'll make another PR to fix it...

PR #947, opened 2016-12-05
Indent go buffers with tabs

doppioandante:


lenormf:
Why? There's already a `gofmt` tool that will normalize the contents of the buffer, no need to enforce any coding style in the support script.

doppioandante:
Why not? This gives correct indentation right away with hooks that can't use the previous line indentation, for example when you write a new function, it feels nicer.

casimir:
I'm for this change, it's really more comfortable with it. One downside is that the filetype hook is not reversible anymore.

lenormf:
Users might want to use spaces for indentation while they code, and use `gofmt` right before it's shared. Language support scripts have no business enforcing one choice over others, particularly when you can pick an intentation style in your own `kakrc` file.

casimir:
On second though I'm against this change. Don't get me wrong it's really annoying to get the wrong indent when adding lines but this config should stay in `kakrc`.
All official `<lang>.kak` provide syntactic coloration and indent hooks and a way to remove them, but that's all (for now at least). This config can not be revert with a `remove-hooks` and thus can be counter intuitive for some users.

Also, personal preference but [I set this option](https://github.com/casimir/kakoune-extra/blob/master/lang/go.kak) on the `window` scope instead of the `buffer` scope. I find it more logical to set visual option to the visual scope.

doppioandante:
I find it strange that a Go programmer would not want tabs in their code, but I agree that this would not follow our current conventions.

PR #948, opened 2016-12-06
 Allow setting custom make error pattern

pauldub:
The current pattern used by the commands `make-next` and `make-prev` 
to find error messages are not suitable for all usages:

For example the go compiler will not suffix error lines with `error: ` and is
not usable with these functions.

This change allows the user to define a custom error pattern, instead of
having to work around the error (for example using sed to insert the 
`error: ` suffix).

What do you think of this? I have not followed the current convention
of having options without separators (like `makecmd`) because it has 
a verbose name. 

Also this does not feel to be the right solution because the pattern has 
to be set at global level.

### DIFF
--- a/rc/core/make.kak
+++ b/rc/core/make.kak
@@ -1,5 +1,5 @@
 decl str makecmd make
-decl str make_error_pattern "[0-9]+: (?:fatal )?error:"
+decl str make_error_pattern " (?:fatal )?error:"
 
 decl str toolsclient
 decl -hidden int _make_current_error_line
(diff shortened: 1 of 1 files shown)


mawww:
Hello,

I wonder about this change, the option actually controls which lines are considered for finding next/previous error, but we still keep the general format `<filename>:<line>:(column:)?.*`. if we decide to only support that format (which is very prevalent in the unix word), then the option should not contain the `[0-9]+:' part.

Seems that would be pretty nice, as it would allow to choose if make-next takes warnings, notes, etc... into account. And I dont think we need a generalized error format handling system (which would be pretty tricky to put in place).

I'd suggest to adapt that change so that we search for next/previous '(?:\w:)?[^:]+:\d+:(?:\d+:)?%opt{make_error_pattern}'.

Regarding the name, we still need to do a general unification on the option naming conventions, current trend is to go towards more explicit names, less abbreviations (as we did for commands), so `make_error_pattern` seems ok.

Thanks for that PR,

Maxime.

rc/core/make.kak
```diff
@@ -1,4 +1,6 @@
 decl str makecmd make
+decl str make_error_pattern " (?:fatal )?error:"
```
pauldub: Should this be the default value or should it be left empty?

mawww: I'd let that as a default for now, as it was the previous behaviour, we can discuss later on if the default should be changed.

casimir:
@pauldub how do you set the value? The only way I found is `set global make_error_pattern ''`. But I would rather not do that.

pauldub:
@casimir at the moment I set it in a `WinCreate` hook.

PR #949, opened 2016-12-06
Transform mimetype to a fallback option

casimir:
The option is now used as a fallback when detection by extension fails. Some scripts like `base/mail.kak` and `base/html.kak` still rely heavily on it.

This PR collides with #878.

### DIFF
--- a/rc/base/file.kak
+++ b/rc/base/file.kak
@@ -1,0 +1,13 @@
+decl str mimetype
+
+hook global BufOpen .* %{ %sh{
+    if [ -z "${kak_opt_filetype}" ]; then
+        mime=$(file -b --mime-type "${kak_buffile}")
+        printf %s\\n "${mime}" | grep -q '^text/x-'
+        if [ $? -eq 0 ]; then
+            printf "set buffer filetype '%s'\n" "${mime:7}"
+        else
+            printf "set buffer mimetype '%s'\n" "${mime}"
+        fi
+    fi
+} }
--- a/rc/base/file.kak
+++ b/rc/base/file.kak
@@ -3,9 +3,12 @@ decl str mimetype
 hook global BufOpen .* %{ %sh{
     if [ -z "${kak_opt_filetype}" ]; then
         mime=$(file -b --mime-type "${kak_buffile}")
-        printf %s\\n "${mime}" | grep -q '^text/x-'
-        if [ $? -eq 0 ]; then
-            printf "set buffer filetype '%s'\n" "${mime:7}"
+        case "${mime}" in
+            text/x-*) filetype="${mime#text/x-}" ;;
+            text/*)   filetype="${mime#text/}" ;;
+        esac
+        if [ -n "${filetype}" ]; then
+            printf "set buffer filetype '%s'\n" "${filetype}"
         else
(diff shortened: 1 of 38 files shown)


rc/base/file.kak
```diff
@@ -0,0 +1,13 @@
+decl str mimetype
```
mawww: We dont need this option anymore, do we ?

rc/base/file.kak
```diff
@@ -0,0 +1,13 @@
...
+
+hook global BufOpen .* %{ %sh{
+    if [ -z "${kak_opt_filetype}" ]; then
+        mime=$(file -b --mime-type "${kak_buffile}")
+        printf %s\\n "${mime}" | grep -q '^text/x-'
+        if [ $? -eq 0 ]; then
+            printf "set buffer filetype '%s'\n" "${mime:7}"
+        else
+            printf "set buffer mimetype '%s'\n" "${mime}"
```
mawww: this is not used for anything, remove

rc/base/file.kak
```diff
@@ -0,0 +1,13 @@
+decl str mimetype
+
+hook global BufOpen .* %{ %sh{
+    if [ -z "${kak_opt_filetype}" ]; then
+        mime=$(file -b --mime-type "${kak_buffile}")
+        printf %s\\n "${mime}" | grep -q '^text/x-'
```
mawww: Looking at the other changes that were needed, I suggest we use a mapping here in the end, something like:
```
case $mime in
  text/x-c++) filetype=cpp ;;
  text/x-tex) filetype=latex ;;
  text/x-shellscript) filetype=sh ;;
  text/x-*) filetype=${mime%text/x-} ;;
  text/plain) ;;
  text/*) filetype=${mime%text/} ;;
esac
[ -n "$filetype" ] && printf "set buffer filetype '%s'\n" ${filetype}
```

rc/base/file.kak
lenormf: User scripts might want to know what the current mimetype is, or even for debugging purpose (if `file` gives the current buffer an eroneous mimetype, I want a way to know what it is).

rc/base/file.kak
```diff
@@ -0,0 +1,13 @@
...
+        if [ $? -eq 0 ]; then
+            printf "set buffer filetype '%s'\n" "${mime:7}"
```
lenormf: The variable expansion with and index isn't POSIX.

rc/base/file.kak
lenormf: Not all mimetypes are easily linkable to their respective filetype, e.g. `message/rfc822` for mails. This list is going to be big if it's implemented, and not having it would allow weird mimetypes to be contained in their respective language scripts.

mawww: So, part of my reasoning for moving that here is that mime types are not standardised for most of these (any text/x- is explicitely not standard), so whatever is used is mostly specific to the `file` command. For the few that are standardised (html, xml and mail basically), it does not seem worth it to continue to keep that mimetype option around.

In other words, as `mimetype` values are not standard, I'd rather have the `file.kak` know about most kakoune filetypes than filetype scripts know about whatever `file` generates as mimetype for their filetype. Keeping the coupling in a single optional place.

PR #950, opened 2016-12-08
Fix python comment chars

casimir:
Add line chars and fix selection chars (`"""` is for docstrings).


rc/core/commenting.kak
```diff
@@ -78,7 +78,8 @@ hook global BufSetOption filetype=(pug|rust) %{
 }
 
 hook global BufSetOption filetype=python %{
-    set buffer comment_selection_chars '""":"""'
+    set buffer comment_line_chars '#'
```
lenormf: This is the default already, no need to re-initialize the variable :)

PR #953, opened 2016-12-12
editorconfig, fix indent options and add whitespaces ones

casimir:
I was just going to fix a syntax error but got carried away and refactored the whole script plus implemented some missing features for editorconfig.

Options reference: http://editorconfig.org/#file-format-details

- **indent_style**, **indent_size**, **tab_width**: follow the specification or default to 4 spaces when needed
- **end_of_line**: set `eolformat` to `lf` or `crlf`
- **charset**: set `BOM` to `utf8` if the value is `utf-8-bom`
- **trim_trailing_whitespace**: remove trailing whitespaces on save when set to `true`
- **insert_final_newline**: remove trailing empty lines when the value is `false`, leave only one empty line at the end of the file when the value is `true`, does nothing when the option is not present

### DIFF
--- a/rc/extra/editorconfig.kak
+++ b/rc/extra/editorconfig.kak
@@ -18,10 +18,11 @@ def -hidden _editorconfig-bufwritepre-hook %{
     } }
 }
 
-def editorconfig-load -docstring "Set indentation options and whitespace behavior according to editorconfig" %{
+def editorconfig-load -params ..1 -docstring "editorconfig-load [file]: set formatting behavior according to editorconfig" %{
+    remove-hooks buffer editorconfig-hooks
     %sh{
-        command -v editorconfig >/dev/null 2>&1 || { echo 'echo -color Error The editorconfig tool could not be found'; exit 1; }
-        editorconfig "${kak_buffile}" | awk -F= -- '
+        command -v editorconfig >/dev/null 2>&1 || { echo 'echo -color Error editorconfig could not be found'; exit 1; }
+        editorconfig "${1:-$kak_buffile}" | awk -F= -- '
             /indent_style=/            { indent_style = $2 }
             /indent_size=/             { indent_size = $2 == "tab" ? 4 : $2  }
             /tab_width=/               { tab_width = $2 }
--- a/rc/extra/editorconfig.kak
+++ b/rc/extra/editorconfig.kak
@@ -2,21 +2,6 @@
 # ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
 
 decl -hidden bool editorconfig_trim_trailing_whitespace false
-decl -hidden int  editorconfig_insert_final_newline     0
-
-def -hidden _editorconfig-bufwritepre-hook %{
-    hook buffer BufWritePre %val{buffile} -group editorconfig-hooks %{ %sh{
-        if [ ${kak_opt_editorconfig_trim_trailing_whitespace} = "true" ]; then
-            printf %s\\n "try %{ exec -draft %{ %s\h+$<ret>d } }"
-        fi
-        if [ ${kak_opt_editorconfig_insert_final_newline} -gt 0 ]; then
-            printf %s\\n "try %{ exec -draft %{ %s\v+\Z<ret>d } }"
-        fi
-        if [ ${kak_opt_editorconfig_insert_final_newline} -gt 1 ]; then
-            printf %s\\n "try %{ exec -draft %{ geo<esc> } }"
-        fi
-    } }
-}
 
 def editorconfig-load -params ..1 -docstring "editorconfig-load [file]: set formatting behavior according to editorconfig" %{
     remove-hooks buffer editorconfig-hooks
@@ -29,7 +14,6 @@ def editorconfig-load -params ..1 -docstring "editorconfig-load [file]: set form
             /end_of_line=/             { end_of_line = $2 }
             /charset=/                 { charset = $2 }
             /trim_trailing_whitespace/ { trim_trailing_whitespace = $2 }
-            /insert_final_newline/     { insert_final_newline = $2 }
 
             END {
                 if (indent_style == "tab") {
@@ -52,11 +36,12 @@ def editorconfig-load -params ..1 -docstring "editorconfig-load [file]: set form
                 if (trim_trailing_whitespace == "true") {
                     print "set buffer editorconfig_trim_trailing_whitespace true"
                 }
-                if (insert_final_newline) {
-                    print "set buffer editorconfig_insert_final_newline " (insert_final_newline == "true" ? 2 : 1)
-                }
             }
         '
     }
-    _editorconfig-bufwritepre-hook
+    hook buffer BufWritePre %val{buffile} -group editorconfig-hooks %{ %sh{
+        if [ ${kak_opt_editorconfig_trim_trailing_whitespace} = "true" ]; then
+            printf %s\\n "try %{ exec -draft %{ %s\h+$<ret>d } }"
+        fi
+    } }
 }
--- a/rc/extra/editorconfig.kak
+++ b/rc/extra/editorconfig.kak
@@ -18,7 +18,7 @@ def -hidden _editorconfig-bufwritepre-hook %{
     } }
 }
 
-def editorconfig-load -docstring "Set indentation optins and whitespace behavior according to editorconfig file" %{
+def editorconfig-load -docstring "Set indentation options and whitespace behavior according to editorconfig" %{
     %sh{
         command -v editorconfig >/dev/null 2>&1 || { echo 'echo -color Error The editorconfig tool could not be found'; exit 1; }
         editorconfig "${kak_buffile}" | awk -F= -- '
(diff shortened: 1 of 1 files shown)


rc/extra/editorconfig.kak
```diff
@@ -1,35 +1,61 @@
...
+        if [ ${kak_opt_editorconfig_insert_final_newline} -gt 0 ]; then
+            printf %s\\n "try %{ exec -draft %{ %s\v+\Z<ret>d } }"
+        fi
+        if [ ${kak_opt_editorconfig_insert_final_newline} -gt 1 ]; then
+            printf %s\\n "try %{ exec -draft %{ geo<esc> } }"
+        fi
+    } }
+}
+
+def editorconfig-load -docstring "Set indentation optins and whitespace behavior according to editorconfig file" %{
```
doppioandante: Typo

casimir:
@mawww if you have some time soon this begin to look like a annoyance for a few people.

rc/extra/editorconfig.kak
```diff
@@ -1,35 +1,61 @@
...
                 }
-                if (end_of_line) {
-                    if (end_of_line == "lf" || end_of_line == "crlf")
-                        print "set buffer eolformat " end_of_line
-                    else
-                        print "error"
+                if (end_of_line == "lf" || end_of_line == "crlf") {
+                    print "set buffer eolformat " end_of_line
+                }
+                if (charset == "utf-8-bom") {
```
tmiller: `utf-8-bom` is not a valid property of charset from EditorConfig, It should be changed to `utf-8`

rc/extra/editorconfig.kak
```diff
@@ -1,35 +1,61 @@
...
+                    print "set buffer BOM utf-8"
```
tmiller: Setting the buffer BOM seems to have changed at some point too. The README specifies it as `utf8` not `utf-8`.

casimir: Nice catch.

rc/extra/editorconfig.kak
casimir: Kakoune only handles UTF-8 therefore the encoding is ignored, the purpose is to activate BOM handling. I had hesitations to include this one.

rc/extra/editorconfig.kak
```diff
@@ -1,35 +1,61 @@
...
+def editorconfig-load -docstring "Set indentation options and whitespace behavior according to editorconfig" %{
     %sh{
         command -v editorconfig >/dev/null 2>&1 || { echo 'echo -color Error The editorconfig tool could not be found'; exit 1; }
-        editorconfig $kak_buffile | awk -F= -- '
-            /indent_style=/ { indent_style = $2 }
-            /indent_size=/  { indent_size = $2 == "tab" ? 4 : $2  }
-            /tab_width=/    { tab_width = $2 }
-            /end_of_line=/  { end_of_line = $2 }
-            /charset=/      { charset = $2 }
+        editorconfig "${kak_buffile}" | awk -F= -- '
```
kballard: If `kak_buffile` is `*scratch*` (or, more generally, if it doesn't start with `/`), `editorconfig` is going to complain "Input file must be a full path name.". Much more useful behavior would be to detect this and use `$HOME/.` instead (the `/.` is because with just `$HOME`, editorconfig will look at `$HOME/../.editorconfig` instead of `$HOME/.editorconfig`). This way users can load editorconfig in `BufNew` and it will work for the scratch buffer using their defaults.

casimir: I don't think I get this use case. This is an opt-in feature so you can choose when to use it, I use this line in my config `hook global BufCreate /.+ %{ editorconfig-load }`.

casimir:
@kballard with the last commit you should be able to do whatever you want with scratch buffers. Something like:
```
hook global BufCreate \*.+\* { editorconfig-load /some/file/path }
```

rc/extra/editorconfig.kak
kballard: What's unclear about this? If I have a file `$HOME/.editorconfig`, it's reasonable to assume that this is the default configuration I want for my editor in the absence of other configuration, right? Given that, scratch buffers should use this default configuration. If you don't want scratch buffers to use this default configuration then you can use your hook with `/.+` as you show, but if you do want scratch buffers to use it then you can use the hook with `.*`. As it stands, if you use `.*` right now, you'll just cause `editorconfig-load` to spew an error when presented with `*scratch*`.

kballard:
@casimir I shouldn't have to write a separate hook for scratch buffers. I should be able to write a hook for `.*` that works for all buffers. Doing anything else is just begging for problems.

casimir:
You should, kakoune encourages users to define hook themselves, you can find a lot of discussion about that in this bug tracker actually ([an example](https://github.com/mawww/kakoune/pull/851#discussion_r86684090)). We try to leave a maximum of behaviour decisions to the user, this reduce complexity in scripts and promote discoverability of configuration patterns for newcomers while staying simple enough.

rc/extra/editorconfig.kak
casimir: This is personal preference, I don't get the point of a global `.editorconfig` this is the same as a partial config in your `kakrc`, plus I consider EditorConfig designed for a (sub)project scope. Also I use scratch buffers as, well, scratch buffers: typeless and volatiles.
My way is not better than yours and not better than another ones. This script implements the same behaviour than the official binary with a unique parameter, it won't do more.

kballard: The reason for a global `.editorconfig` instead of putting it in `kakrc` is because `.editorconfig` will work for a variety of editors, not just Kakoune, and so any editor I use will pick up the same configuration.

When you say this script implements the same behavior as the official binary, that doesn't actually make any sense. We're talking here about scratch buffers, which do not have a corresponding file on disk. The official binary is only capable of operating given a file path on disk. Therefore, the official binary has no idea about the concept of scratch buffers, and therefore does not provide any opinion either way as to how scratch buffers should behave.

casimir:
This is getting nowhere, sorry we fail to comprehend ourselves on usage. To
be clear, this script provides a function to use EditorConfig for a buffer,
that's all. Hooks are out of scope here but you are free to create yours
script to handle that.

kballard:
Yes, I know what this script does. What I don't understand is your continued insistence on having this script *fail to work* for scratch buffers. It's really pretty easy to make it fall back to loading `$HOME/.editorconfig`, and that seems like reasonable behavior for a scratch buffer. The user should not be forced to understand that it will simply fail for scratch buffers and that they have to write a second hook specifically for them. If they want to customize the behavior for scratch buffers then they can do that with a separate hook, but if they simply call `editorconfig-load` from a hook on a scratch buffer, then it should do something reasonable.

Incidentally, it occurs to me that the correct behavior here isn't actually to load `$HOME/.editorconfig`, but instead to load a file relative to Kakoune's cwd. Or in other words, if the `$kak_buffile` doesn't start with `/` then it should simply prepend it with `$PWD`.

rc/extra/editorconfig.kak
tmiller: Ah, I see what you mean, I believe I agree with adding it.

rc/extra/editorconfig.kak
```diff
@@ -1,35 +1,62 @@
...
+
+decl -hidden bool editorconfig_trim_trailing_whitespace false
+decl -hidden int  editorconfig_insert_final_newline     0
+
+def -hidden _editorconfig-bufwritepre-hook %{
+    hook buffer BufWritePre %val{buffile} -group editorconfig-hooks %{ %sh{
+        if [ ${kak_opt_editorconfig_trim_trailing_whitespace} = "true" ]; then
+            printf %s\\n "try %{ exec -draft %{ %s\h+$<ret>d } }"
+        fi
+        if [ ${kak_opt_editorconfig_insert_final_newline} -gt 0 ]; then
```
mawww: Kakoune always guarantees the file ends with a newline, because Kakounes buffers are arrays of lines and lines always ends with an end of line char. This seems to be inserting additional newlines, which is not how I interpret the editorconfig spec, I believe that option is there to guarantee the file ends with `\n`, which is what Kakoune will do anyway, and which seems to be the correct thing with text files in a posix environment.

casimir: Good to know, let's drop this part then.

mawww:
@kballard Regarding the support for `*scratch*` buffers, I see `editorconfig-load` support passing the filename as argument, so the user hook that will do `editorconfig-load` can probably take advantage of it.

kballard:
@mawww We already had a long discussion about this. The problem is that you've now got a command that *does the wrong thing* when used in a straightforward fashion. Yes, users can work around this issue with a correctly-written hook, but users shouldn't *have to* understand the implementation details of the command in order to use it. It's just bad API design. Users should be able to write a hook on `.*` that calls `editorconfig-load` and have it do something reasonable for all buffers, and the ability to override the filename using an argument should only be used if the user wishes to further customize the behavior away from the default.

mawww:
@kballard its pretty tricky to decide whats a reasonable behaviour for buffers that dont have any associated paths, using CWD ? using $HOME ? not doing anything ?. Which filename to pass to the editorconfig binary ? just the $CWD or $CWD/$kak_bufname ?

We can detect if a buffer has a file or not by looking at $kak_buffile (if starts with a `/` or not), so the logic should be easy to implement.

I dont really mind if we add a default here, I merely say that I dont see which default makes the most sense. PR are welcome.

kballard:
As [previously commented](https://github.com/mawww/kakoune/pull/953#issuecomment-269562925), I believe CWD makes the most sense (or rather, `$CWD/$kak_bufname`, if `$kak_buffile` does not start with `/`, otherwise editorconfig won't look at `$CWD/.editorconfig`).

PR #962, opened 2016-12-13
doc.kak: add aliases for :h and :help

Phlogistique:
This would help VIM users find the documentation


lenormf:
We've kind of steered away from single letter/non-descriptive aliases because of the completion engine, I think users that really want a `h` alias can declare it in their `kakrc`.

hxegon:
was just about to open an issue when I saw this, thanks @Phlogistique!

PR #965, opened 2016-12-13
Updated readme to include sudo make install

cvoege:
Currently you need to run `sudo make install` to get the install the kak command to your path and for syntax highlighting to work. I'd say 99% of people using this will want to do that immediately, and otherwise they will be quickly confused. This was not included in the README tldr, so I added it.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -18,6 +18,7 @@ TL;DR
 git clone http://github.com/mawww/kakoune.git
 cd kakoune/src
 make
+sudo make install
 ./kak
 ---------------------------------------------
 


lenormf:
Could you please rebase the branch, two commits to add a single line looks weird.

Also I'm not convinced this change is necessary, as users who don't understand they need to have elevated privileges to install files in the default `PREFIX` will more likely use a `kakoune` package provided by their distribution rather than compile the project by hand.

mawww:
The quick start guide does not require Kakoune to be installed, it should run fine directly from the src/ directory.

Issue #963, opened 2016-12-13
Homebrew formula needs to set XML_CATALOG_FILES


kballard:
When installing on macOS using the Homebrew instructions, the install fails. The formula depends on docbook and asciidoc, but it doesn't set the `XML_CONFIG_FILES` environment variable, so it fails to work properly. Looking at other formulae that use docbook and asciidoc, they either pass something like `--with-xml-catalog=#{etc}/xml/catalog` to their `configure` script, or they set `ENV["XML_CATALOG_FILES"] = etc/"xml/catalog"`.

doppioandante:
Yes, dup of #954.
If you want feel free to make a PR for this.

mawww:
Should be fixed now.

Issue #954, opened 2016-12-13
xsltproc errors when trying to install


dlee:
Here's my `xsltproc --version`:
```
❯ xsltproc --version
Using libxml 20902, libxslt 10128 and libexslt 817
xsltproc was compiled against libxml 20902, libxslt 10128 and libexslt 817
libxslt 10128 was compiled against libxml 20902
libexslt 817 was compiled against libxml 20902
```

Here's the log from ~/Library/Logs/Homebrew/kakoune/01.make:
```
2016-12-13 06:39:03 -0800

make
install
PREFIX=/usr/local/Cellar/kakoune/HEAD-e441295

clang++ -DKAK_DEBUG -I/usr/local/opt/ncurses/include -pedantic -std=gnu++11 -g -Wall -Wno-reorder -Wno-sign-compare -Wno-address -MD -MP -MF .alias_registry.debug.d -c -o .alias_registry.debug.o alias_registry.cc
clang++ -DKAK_DEBUG -I/usr/local/opt/ncurses/include -pedantic -std=gnu++11 -g -Wall -Wno-reorder -Wno-sign-compare -Wno-address -MD -MP -MF .assert.debug.d -c -o .assert.debug.o assert.cc
clang++ -DKAK_DEBUG -I/usr/local/opt/ncurses/include -pedantic -std=gnu++11 -g -Wall -Wno-reorder -Wno-sign-compare -Wno-address -MD -MP -MF .backtrace.debug.d -c -o .backtrace.debug.o backtrace.cc
clang++ -DKAK_DEBUG -I/usr/local/opt/ncurses/include -pedantic -std=gnu++11 -g -Wall -Wno-reorder -Wno-sign-compare -Wno-address -MD -MP -MF .buffer.debug.d -c -o .buffer.debug.o buffer.cc
clang++ -DKAK_DEBUG -I/usr/local/opt/ncurses/include -pedantic -std=gnu++11 -g -Wall -Wno-reorder -Wno-sign-compare -Wno-address -MD -MP -MF .buffer_manager.debug.d -c -o .buffer_manager.debug.o buffer_manager.cc
clang++ -DKAK_DEBUG -I/usr/local/opt/ncurses/include -pedantic -std=gnu++11 -g -Wall -Wno-reorder -Wno-sign-compare -Wno-address -MD -MP -MF .buffer_utils.debug.d -c -o .buffer_utils.debug.o buffer_utils.cc
[...]

PR #966, opened 2016-12-13
Fix Homebrew formula

lehmacdj:
I fixed the homebrew formula as per #954 and #963. I tested installing using this formula and it worked (at least for me).


SikhNerd:
Was just about to submit the same, and this identical change to what I was testing locally works for me as well fwiw

lehmacdj:
Is it okay that my copyright waver is after my first commit? I can rebase so that the order is different if I need too.

mawww:
Thanks a lot.

PR #968, opened 2016-12-13
"Fewer"

joshthewebgeek:
https://www.youtube.com/watch?v=UJPeg1b-jVo

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -11,7 +11,7 @@ TL;DR
 
 {website}
 
-*Vim inspired* -- *Faster as in less keystrokes* --
+*Vim inspired* -- *Faster as in fewer keystrokes* --
 *Multiple selections* -- *Orthogonal design*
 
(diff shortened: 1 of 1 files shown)


lenormf:
Could you make the same change on the `gh-pages` branch, which hosts http://kakoune.org/ ?

joshthewebgeek:
Good idea. [Done.](https://github.com/mawww/kakoune/pull/981)

lenormf:
Thanks!

mawww:
According to https://en.wikipedia.org/wiki/Fewer_vs._less it does not look like fewer is clearly more correct, and that less has been used with countable objects for a very long time in the English language.

That said, I dont mind much, if people really feel fewer is better, I can merge that, I expect everybody to understand both words.

mawww:
No activity on that, I'll take it as no strong support for Fewer rather that Less.

PR #989, opened 2016-12-14
Document alt period

### LINKED ISSUE #986: Anything equivalent to the vim ; command?
I couldn't find any "repeat last f/t" equivalent in the docs. Is this implemented somewhere?

hxegon:
closes #986


mawww:
Thanks for that, I had already a non pushed commit doing the same.

lenormf:
Just for information, the reason why the documentation wasn't mentioning `<a-.>` is that this binding was introduced recently as an experiment, and we (users) were waiting to see if it was really that useful before making it official.

PR #1011, opened 2016-12-16
Update documentation about builtin faces

iology:


mawww:
Nice to get that corrected, thank you.

PR #1028, opened 2016-12-18
Fix editorconfig specification

tmiller:
I believe that kakoune was looking for the incorrect values for "indent_style" and "charset". For "indent_style" it appears to be looking for "spaces" instead of "space" and for "charset" it appears to be looking for "utf-8-bom" instead of "utf-8". I've linked the relevant sections of the spec for others to confirm my suspicions.

https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#indent_style
https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#charset

### DIFF
--- a/rc/extra/editorconfig.kak
+++ b/rc/extra/editorconfig.kak
@@ -14,7 +14,7 @@ def editorconfig-load -docstring "Set indentation options according to editorcon
                     print "set buffer indentwidth 0"
                     print "set buffer aligntab true"
                 }
-                if (indent_style == "spaces") {
+                if (indent_style == "space") {
                     print "set buffer indentwidth " (indent_size == "tab" ? 4 : indent_size)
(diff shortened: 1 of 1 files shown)


leemachin:
Thanks :) There is a similar PR here by @casimir (which is a bigger refactoring): https://github.com/mawww/kakoune/pull/953

casimir:
Can you confirm #953 fixes it for you?

tmiller:
@casimir, Sorry I searched the issues but I forgot to search the pull requests. I've looked at your code and it should fix the issue I was seeing with spaces. It appears to still be searching for utf-8-bom instead of utf-8. But I'm not using that, I just noticed it was wrong as well.

tmiller:
I'm closing this pr, since it is a duplicate.

PR #1032, opened 2016-12-19
Shouldn't use hardcoded '/tmp', tmpdir might not be '/tmp' on some platform

leira:
Used hardcoded '/tmp' as tmpdir in close_session(), session socket file wasn't
removed when session exit on macOS, caused conflict when start or connect to the
same session name

### DIFF
--- a/src/remote.cc
+++ b/src/remote.cc
@@ -489,15 +489,20 @@ static StringView tmpdir()
     return "/tmp";
 }
 
+static StringView session_file(StringView session)
+{
+    if (find(session, '/')!= session.end())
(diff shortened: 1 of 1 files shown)


leira:
@doppioandante u mean zstr() on StringView? Let me change it.

mawww:
Hi, thanks for catching that, went with another solution because yours was returning a stringview to dead data, and I'd like to avoid allocating a string when we are writting to a sockaddr_un anyway.

leira:
No problem~

I can't express how much I appreciate your work, haven't seen a software I love so much for a long time!

PR #1035, opened 2016-12-19
Add the -version option to print the version

Phlogistique:
The version is obtained from the command "git describe --always --dirty".

A bit of Makefile magic is used to generate a "version.cc" that is only
recompiled when the output of "git describe" changes.

In the event of a release outside of Git, the code could be modified to use git
when available and to fall back on a hard-coded version number. The current
fall back is the empty string.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -66,6 +66,25 @@ endif
 
 CXXFLAGS += -std=gnu++11 -g -Wall -Wno-reorder -Wno-sign-compare -Wno-address
 
+
+# Version number handling. This save the current version number to a file and
+# rebuilds version.cc iff the version number has actually changed.
+version := $(shell git describe --always --dirty 2>/dev/null)
+version_old = $(shell cat .version 2>/dev/null)
+
+ifneq ($(version_old),$(version))
+# Mark version.cc for rebuild. Do not use .PHONY because this needs to be
+# persistent if make is interrupted. Do not use the $(file) function as it
+# segfaults in GNU Make 4.0
+$(shell touch version.cc)
+$(file > .version,$(version))
+endif
+
+# Special treatment for version.cc: define KAK_VERSION
+# This needs to be an "=" rather than ":=" to be expanded in rule context
+version_cppflag = $(if $(filter version.cc,$<),-DKAK_VERSION=\"$(version)\")
+
+
 all : kak
 kak : $(objects)
 	$(CXX) $(LDFLAGS) $(CXXFLAGS) $(objects) $(LIBS) -o $@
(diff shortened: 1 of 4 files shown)


lenormf:
An easier way of implementing this would have been to generate a `-DVERSION=` pre-processing flag, which would have subsequently been used in the code, instead of having a global variable. Also generating source files is ugly.

That said, even though I'd like to see versioning in the project, having a git digest as a version number isn't very useful, and will even allow packager to use this as reference and create confusing packages where there's no way to know what release is considered major, the changelogs will almost constantly be out of date etc.

So in the end, I think it's too early for that, there's no need to force versioning using a value that's not very relevant (except for debugging purposes, for users to identify if they're out of date, but even so that's not much of a big help anyway) until @mawww has decided it's time to have proper major/minors.

Phlogistique:
> An easier way of implementing this would have been to generate a -DVERSION= pre-processing flag, which would have subsequently been used in the code, instead of having a global variable

The point of the global variable is to prevent the need to recompile the code when the version string changes.

> Also generating source files is ugly.

OK, I uploaded a new version which does not use code generation.

Here is an extract of the original diff using code generation, for future reference:
```
diff --git a/src/Makefile b/src/Makefile
index 966582d..33c3633 100644
--- a/src/Makefile
+++ b/src/Makefile
@@ -19,6 +19,12 @@ ifeq ($(pedantic),yes)
 endif
 
 sources := $(wildcard *.cc)
+
+# version.cc is generated, include it explicitly:
+ifeq ($(filter version.cc,$(sources)),)
+sources += version.cc
+endif
+
 objects := $(addprefix ., $(sources:.cc=$(suffix).o))
 deps := $(addprefix ., $(sources:.cc=$(suffix).d))
 docs := $(wildcard ../doc/manpages/*.asciidoc)
@@ -138,3 +146,19 @@ uninstall:
 
 .PHONY: check TAGS clean distclean installdirs install install-strip uninstall
 .PHONY: tags test man doc
+
+# Version code
+version = $(shell git describe --always --dirty 2>/dev/null)
+version_cc_old = $(shell cat version.cc 2>/dev/null)
+define version_cc_new
+#include "version.hh"
+
+const char* const Kakoune::version = "$(version)";
+endef
+
+ifneq ($(version_cc_old),$(version_cc_new))
+.PHONY: version.cc
+endif
+
+version.cc:
+       $(file > $@,$(version_cc_new))
```

> having a git digest as a version number isn't very useful

It's still better than having no way whatsoever to know which source your binary was compiled from!

> will even allow packager to use this as reference and create confusing packages

I don't think having a "-version" switch changes anything to what a packager will do or not do.

> except for debugging purposes, for users to identify if they're out of date

Well, this does sound pretty useful to me.

leemachin:
I feel like I might be imagining this but I feel really sure that it was said somewhere that kakoune has no version for a reason. I'm happy to be wrong because I have no idea where I saw this comment (or read something else and misunderstood it :D)

In any case, semver _has_ been discussed.  More than once. A version number constructed from a git hash doesn't reveal any useful information (downloading `v.cccccc` doesn't mean it is a later version than `v.aaaaaa` unless you have the repo history at hand to query it), and if versioning becomes a thing I don't see what is so bad about manually updating a version file along with a changelog.

Phlogistique:
@leemachin Your comment is about the versionning policy of Kakoune.

The point of this pull request is to provide a mechanism for the user to be able to figure out which version of Kakoune they are using, regardless of whatever versionning policy the Kakoune project might or might not implement.

lenormf:
Could this PR be rebased on `master`, to make use of the `Kakoune::version` variable that was implemented?

PR #1055, opened 2016-12-21
Add handlebars syntax highlighters

haarts:
Please review this with some care as I'm very new to Kakoune.


lenormf:
Could you squash all the commits please?

haarts:
Funny you would ask that. I just spent some time making pretty commit out of the one large one I had. :)

lenormf:
It's a good practice generally, but since it's only a single file with not many changes we don't need to add much more than one commit upstream.

mawww:
Hello,

Can you add a copyright waiver commit as well ? See the CONTRIBUTING file.

Cheers.

haarts:
@mawww Done!

PR #1064, opened 2016-12-22
Strip \n in Atoms

### LINKED ISSUE #775: How does bold work?
I'm experimenting with a QML UI for kakoune. The following JSON is what you get from the draw call when you do the following:
* create a dir with a file containing 'test1' and 'test2' on two different lines
* run :grep test in kakoune
* grep-jump on the first line in the group output

What I get is a sequence of whitespaces in bold just before the second entry, which for some reason the ncurses UI doesn't display. Is this expected?
```
  {
      "face": {
        "fg": "default",
        "bg": "default",
        "attributes": [
          "bold"
        ]
      },
      "contents": "                                                                                                                                                                                       "
    }
```

Full JSON:
```javascript
[
  [
    {
      "face": {
        "fg": "cyan",
        "bg": "default",
        "attributes": [
          "bold"
        ]
      },
      "contents": "a.txt"
    },
    {
      "face": {
        "fg": "default",
        "bg": "default",
        "attributes": [
          "bold"
        ]
      },
      "contents": ":"
    },
    {
      "face": {
        "fg": "green",
        "bg": "default",
        "attributes": [
          "bold"
        ]
      },
      "contents": "1"
    },
    {
      "face": {
        "fg": "default",
        "bg": "default",
        "attributes": [
          "bold"
        ]
      },
      "contents": ":"
    },
    {
      "face": {
        "fg": "black",
        "bg": "white",
        "attributes": [
          "bold"
        ]
      },
      "contents": "1"
    },
    {
      "face": {
        "fg": "default",
        "bg": "default",
        "attributes": [
          "bold"
        ]
      },
      "contents": ":test1\n"
    },
    {
      "face": {
        "fg": "default",
        "bg": "default",
        "attributes": [
          "bold"
        ]
      },
      "contents": "                                                                                                                                                                                       "
    }
  ],
  [
    {
      "face": {
        "fg": "cyan",
        "bg": "default",
        "attributes": []
      },
      "contents": "a.txt"
    },
    {
      "face": {
        "fg": "default",
        "bg": "default",
        "attributes": []
      },
      "contents": ":"
    },
    {
      "face": {
        "fg": "green",
        "bg": "default",
        "attributes": []
      },
      "contents": "2"
    },
    {
      "face": {
        "fg": "default",
        "bg": "default",
        "attributes": []
      },
      "contents": ":"
    },
    {
      "face": {
        "fg": "green",
        "bg": "default",
        "attributes": []
      },
      "contents": "1"
    },
    {
      "face": {
        "fg": "default",
        "bg": "default",
        "attributes": []
      },
      "contents": ":test2\n"
    }
  ],
  [
    {
      "face": {
        "fg": "default",
        "bg": "default",
        "attributes": []
      },
      "contents": "\n"
    }
  ]
]
```

doppioandante:
Fixes #775

### DIFF
--- a/doc/json_ui.asciidoc
+++ b/doc/json_ui.asciidoc
@@ -20,6 +20,9 @@ Here are the data structures used:
 * Line : Array of Atom
 * Coord { int line; int column }
 
+The newline is implicit after every Line and a Line containing no atoms
+should be displayed as an emtpy line.
+
 Here are the requests that can be written by the json ui on stdout:
 
 * draw(Array<Line> lines, Face default_face, Face padding_face)
--- a/src/json_ui.cc
+++ b/src/json_ui.cc
@@ -49,11 +49,14 @@ String to_json(StringView str)
         if (next == end)
             break;
 
-        char buf[7] = {'\\', *next, 0};
-        if (*next >= 0 and *next <= 0x1F)
-            sprintf(buf, "\\u%04x", *next);
+        if (*next != '\n')
+        {
+            char buf[7] = {'\\', *next, 0};
+            if (*next >= 0 and *next <= 0x1F)
+                sprintf(buf, "\\u%04x", *next);
 
-        res += buf;
+            res += buf;
+        }
         it = next+1;
     }
     res += '"';


lenormf:
Seems that this changeset still allows the json-rpc backend to emit atoms with no contents. Also according to the bug tracker there was nothing to be fixed there (apparently), it was merely closed without further comments, is removing all newlines from atoms the way to go?

doppioandante:
> Seems that this changeset still allows the json-rpc backend to emit atoms with no contents.

This is true, but that's not the problem.
What was happening there is that the line is modified afterwards by the highlighter, which adds atoms after the last one. The only problem is that the "\n" is sticked to the last actual atom, so you get a "\n" in the middle of the line which doesn't make sense.

The alternative is to leave this as it is, but document that clients should actual get rid of all the "\n" that it finds (which should be what the terminal UI does right now).

lenormf:
So it could be an issue caused by highlighters? I'm afraid this patch fixes the symptoms rather than the issue, because the original bug report was closed without much info about why.

doppioandante:
Yeah I still have to think about the solution a bit more.

I don't feel like the line highlighter is broken: [this](https://github.com/mawww/kakoune/blob/master/src/highlighters.cc#L528) is the line where the empty space is added.
It is a push back on the line element, and it is conceptually ok because the newline should be implicit in the representation. Changing it so that the "\n" is taken away from the previous atom and put at the end is ugly.

The other way is to get rid of the "\n" altogether, and this is what I did, but only at the JSON level. It would probably make sense to do this when atoms are constructed.

The third option is leaving everything as it is and leaving the correction to all the clients, and I don't like it.

doppioandante:
I have been thinking about this and realized that this is not the correct fix: the cursor can go upon newlines, so it shouldn't be stripped away from the atom.
As matter of fact we want it to be treated as a space, so it could make sense to replace it with a space, or left to be intended as a space from the UI.

mawww:
Yeah, the fact that lines last atom ends with a `\n` is expected, we could get rid of it in the core and replace it with a space, or let the UI handle `\n` themselves, but we need to have something there.

Keeping `\n` is more efficient with local UIs, as we directly access buffer data with no string copy, so we cannot replace that `\n` with something else without first duplicating the atom data. It seems reasonable to continue doing that.

In any case, we cant just strip them out.

PR #1065, opened 2016-12-22
Basic implementation of a buffer list

lenormf:
Hi,

Here's what it looks like (my theme+default):

a

b

c

d

As you can see, the separator is customizable, as well as the faces used to render the bar. A unicode ellipsis is inserted at the beginning/end of the list when there wasn't enough space to print an entire buffer name.

Read the first commit message for all the details.


lenormf:
Following the advice of @doppioandante, I've decided not to implement any kind of smart behaviour in the buffer list, the buffers are now displayed from left to right, with no attempts at showing buffers surrounding the active one.

lisael:
Isn't there a naming clash with `%val{buflist}` ?

lenormf:
I don't think so, what makes you say that? The commits merely add an ncurses UI widget, it doesn't interfere with options.

lisael:
Not technically speaking, but for the user, it may be confusing reading the same name for two different things. (The fact is I had to make an effort to forget about %opt{buflist} while reading the PR. Maybe I used this a lot, recently :) )

lenormf:
Branch rebased against 3a0d9481ac0d037deb43fe93f35318acae763d6b, and the second commit has been squashed into the other for easiness of maintainance purposes.

Delapouite:
Thanks for this PR.

How do you use this buffer list in practice? By using `buffer-next` and `buffer-previous` commands?
I tried to analyze my workflow with a similar buffer bar offered by vim. The main difference is that each buffer name in this list is prepended by a number, like so:

1 buffer-foo |  2 buffer-bar | 3 buffer-qux …

This way I've bound key mapping to quickly jump to a specific buffer. Like `<a-1>`.

Is this something convenient that's worth adding?

lenormf:
I consider this more of a widget and a visual representation of kakoune's state than an actual tool. Consequently I don't think I would like to add numbering to the bar, however you can add that feature very easily.

Also you could probably implement the bindings you mention with a shell scope (assuming the buffers in `%val{buflist}` are in a deterministic order -which I don't know if they are), it should get you pretty close to what you had with `vim` if combined with numbering.

Delapouite:
If someone is interested after reading this thread, here's an alternative suggestion by danr, using `info`

```
# Buffers (by danr 2017, public domain)
map global normal <a-,> :bp<ret>:bufinfo<ret>
map global normal <a-.> :bn<ret>:bufinfo<ret>
map global normal <a--> ga:bufinfo<ret>
map global normal <a-d> :db<ret>:bufinfo<ret>
map global normal <a-q> :db!<ret>:bufinfo<ret>

def bufinfo %{
    %sh{
        echo -n info -- %^
        (while read -d : buf; do
            if [[ "$buf" == "$kak_bufname" ]]; then
                echo "> $buf <"
            else
                echo "  $buf  "
            fi
        done) <<< "$kak_buflist"
        echo ^
    }
}
```
https://gist.github.com/danr/e69a55129a3fd27f1f098d4e167afccd

alexherbo2:
I would like the feedback on buffer change being built-in.

It’s something missing when doing `ga` or `:buffer-next` commands.

Delapouite:
> I would like the feedback on buffer change being built-in.

You mean something like a `BufChange` hook?

occivink:
What you can do since recently is add
```
hook global WinDisplay .* bufinfo
```
in addition to the bufinfo implementation so that you get the popup without having to remap every commands that changes the current buffer.

alexherbo2:
@Delapouite No, I mean the feedback being built-in.

lenormf:
I documented the options/faces added, and rebased on 55d1d1020d215730737554045537cf3a8877c46a.

lenormf:
Fixed some issues that glitched the rendering.

Delapouite:
Here's an alternative solution using external bars like lemonbar : https://github.com/mawww/kakoune/wiki/Bar

mawww:
In my view, #1942 superseeds this PR, and I have no intention to merge this PR, as it is pretty intrusive, closing.

PR #1067, opened 2016-12-25
Add 'extend' to ruby grammar (meta)

knubie:
Technically `include` and `extend` are both just methods on `Module` and `Object`, respectively, but I think we should have both.


leemachin:
I think it would be worth adding `prepend`, `refine` and `using`, since they're also missing.

PR #1069, opened 2016-12-26
prevent `->` from appearing as an arrow on github

vitaly:
`->` was interpreted as a 'smart arrow' and replaced with `→`

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -409,7 +409,7 @@ Changes
  * `alt-m`: merge contiguous selections together (works across lines as well)
 
  * `>`: indent selected lines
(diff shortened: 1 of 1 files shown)


kballard:
Why? I thought the smart arrow was intentional. It's a single keypress, having one glyph is fine, and `→` is commonly used to mean the right arrow key.

kballard:
Huh, I actually thought it was literally the right arrow key. I wrote a comment to that effect, and then realized that, hey, it's probably alt plus `>`. So yeah, :+1:

mawww:
This is not relevant anymore with the recent changes on the README unifying the key syntax.

PR #1080, opened 2016-12-27
Consistent notation for key chords in Readme.

akkartik:
Is notation like `<a-<>` and `<a->>` correct? Clear?


mawww:
`<a-lt>` and `<a-gt>` would be the correct ones, we use `<lt>` to avoid `<` being interpreted as a key start.

akkartik:
Fixed. Any other acronyms like these?

PR #1088, opened 2016-12-28
Add nested ruby support to html. (e.g. hl in erb templates)

alexanderjeurissen:
This diff makes it so that the HTML filetype is also applied for filetypes that have the extension `.html.erb` and adds ruby highlighting to the `<%= %>` and `<% %>` tags in html files.

**before:**
screen shot 2016-12-28 at 15 34 11

**after:**
screen shot 2016-12-28 at 15 32 58

### DIFF
--- a/rc/base/html.kak
+++ b/rc/base/html.kak
@@ -8,6 +8,10 @@ hook global BufCreate .*\.html %{
     set buffer filetype html
 }
 
+hook global BufCreate .*\.html\..* %{
+    set buffer filetype html
+}
+
 hook global BufCreate .*\.xml %{
     set buffer filetype xml
 }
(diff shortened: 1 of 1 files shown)


alexherbo2:
I would prefer having an `erb.kak` rather than merging in base `html.kak`.

alexanderjeurissen:
The reason I added this to `html.kak` is because I was afraid that having all html highlighting work in `erb.kak` would lead to a lot of duplication of highlighting rules already present in `html.kak`

alexherbo2:
You can `ref html` as well.

casimir:
I second that. This is weird to include a template language in the HTML highlighter. This would also be a good example to integrate more template languages.

alexanderjeurissen:
Thanks for the advice, I'll create a separate ERB highlighter then.

mawww:
Closing that, erb highlighter is welcome :).

PR #1089, opened 2016-12-28
Add nested html support to javascript files. (For React components)

alexanderjeurissen:
This diff adds nested HTML highlighting to javascript files. This is especially necessary for the increasing number of people that work with view frameworks such as React.

**before**:
screen shot 2016-12-28 at 15 41 19

**after**:
screen shot 2016-12-28 at 15 39 40

### DIFF
--- a/rc/base/javascript.kak
+++ b/rc/base/javascript.kak
@@ -12,7 +12,8 @@ addhl -group / regions -default code javascript \
     double_string '"'  (?<!\\)(\\\\)*"        '' \
     single_string "'"  (?<!\\)(\\\\)*'        '' \
     comment       //   '$'                    '' \
-    comment       /\*  \*/                    ''
+    comment       /\*  \*/                    '' \
+    tag          <       >                    ''
 
 # Regular expression flags are: g → global match, i → ignore case, m → multi-lines, y → sticky
 # https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp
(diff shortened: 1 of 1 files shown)


alexherbo2:
If you could do same as #1088 if that’s possible.

alexanderjeurissen:
@alexherbo2 I think in this case it's a bit more tricky unless multiple highlighters can be active at the same time, reason being that react files often end with `.js` as file extension. So distinguishing javascript files from files that contain jsx can be tricky.

alexherbo2:
That is what I was afraid of, but that’s fine if there is no other way.

casimir:
Don't take that as a request but this is actually possible. I would see a `extra/react.kak` defining a highlighter but also a filetype setter. If you can detect react specific syntax in the file you can set the filetype of the buffer yourself. For some inspiration there is [file.kak](https://github.com/mawww/kakoune/blob/master/rc/base/file.kak) that uses the `file` to deduce the filetype.

mawww:
This wont distinguish between `<` and `>` used as operators, and tags.

Additionally, I agree this should be in a separate file, that would reuse javascript highlighting, but define regions differently. Note that `ref` highlighters are not limited to root highlighters, you can do `add-highlighter ref javascript/code`, so its pretty easy to create a new regions highlighter for jsx, and reference the normal javascript highlighters for the matching regions, while referencing the root html highlighter for other regions.

mawww:
Waiting for the jsx.kak or react.kak :), closing this one.

Issue #974, opened 2016-12-14
building on macOS


nhooyr:
When trying to build with clang++, I get the following error:

```
/p/t/k/src
❯ make CXX=clang++                                                                                                                                                                                         i
clang++ -DKAK_DEBUG -I/usr/local/opt/ncurses/include -pedantic -stdlib=libc++ -g -Wall -Wno-reorder -Wno-sign-compare -Wno-address -MD -MP -MF .color.debug.d -c -o .color.debug.o color.cc
In file included from color.cc:1:
In file included from ./color.hh:4:
./hash.hh:4:10: fatal error: 'type_traits' file not found
#include <type_traits>
         ^
1 error generated.
make: *** [.color.debug.o] Error 1
```

my clang++ version is
```
clang version 3.7.1 (tags/RELEASE_371/final)
Target: x86_64-apple-darwin16.1.0
Thread model: posix
```

Any ideas what is wrong?

nhooyr:
If I use the clang++ that is included in macOS (I was using clang++ from my package manager nix above), I get

```
/p/t/k/src
❯ make CXX=clang++                                                                                  i
clang++ -DKAK_DEBUG -I/usr/local/opt/ncurses/include -pedantic -stdlib=libc++ -g -Wall -Wno-reorder -Wno-sign-compare -Wno-address -MD -MP -MF .color.debug.d -c -o .color.debug.o color.cc
In file included from color.cc:1:
In file included from ./color.hh:4:
./hash.hh:14:18: warning: variadic templates are a C++11 extension [-Wc++11-extensions]
template<typename... Type>
                 ^
./hash.hh:29:24: warning: rvalue references are a C++11 extension [-Wc++11-extensions]
size_t hash_values(Type&& t)
                       ^
./hash.hh:39:33: warning: variadic templates are a C++11 extension [-Wc++11-extensions]
template<typename Type, typename... RemainingTypes>
                                ^
./hash.hh:40:24: warning: rvalue references are a C++11 extension [-Wc++11-extensions]
size_t hash_values(Type&& t, RemainingTypes&&... rt)
                       ^
./hash.hh:40:44: warning: rvalue references are a C++11 extension [-Wc++11-extensions]
size_t hash_values(Type&& t, RemainingTypes&&... rt)
                                           ^
In file included from color.cc:1:
./color.hh:14:10: error: ISO C++ forbids forward references to 'enum' types
    enum NamedColor : char
         ^
[...]

PR #1096, opened 2016-12-30
Fix boost linking on OSX for manual make

casimir:
Adresses #974^W manual build using boost from homebrew. This PR adds explicit flags for homebrew's boost. My guess is that `brew` has some implicit `{LD,CPP}FLAGS`, so it shouldn't conflict. 

cc @nhooyr

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -38,8 +38,8 @@ os := $(shell uname)
 
 ifeq ($(os),Darwin)
     LIBS += -lncurses -lboost_regex-mt
-    CPPFLAGS += -I/usr/local/opt/ncurses/include
-    LDFLAGS += -L/usr/local/opt/ncurses/lib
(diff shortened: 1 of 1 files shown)


nhooyr:
Is there something missing from this PR? It only modifies `rc/extra/editorconfig.kak`.

casimir:
Fixed @nhooyr. Sorry too much branches on my repo 😨.

nhooyr:
Either way, I can't test this, I don't use homebrew, I use nix.

casimir:
I had the exact same error so I assume you used homebrew too. Anyway the fix for nix should be an equivalent, either add include and lib path in flags or a `PKG_PATH` when running make.

nhooyr:
Somehow, it is building fine though now, with the system clang++. See my latest comment on #974.

PR #1104, opened 2017-01-01
Fix for -l and -clear options on OSX

### LINKED ISSUE #1097: kak -l shows no sessions on OSX
I have several instances running, some just started with "kak somefile", but also kak -d -s somename, none of which show up with "kak -l" on OSX. It is however possible to connect to the named session with "kak -c somename".
In contrast to README.md (2.3 Running), the session fifos don't seem to show up under what is "/tmp" on my mac but seem to be stored someplace else.

casimir:
Fixes #1097. 

@mawww this is a reimplementation of the logic in [remote.cc](https://github.com/mawww/kakoune/blob/master/src/remote.cc#L483) but I'm not sure how to organise this, I'll let you factorise if needed.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -761,8 +761,13 @@ int main(int argc, char* argv[])
         const bool clear_sessions = (bool)parser.get_switch("clear");
         if (list_sessions or clear_sessions)
         {
+            StringView tmpdir = getenv("TMPDIR");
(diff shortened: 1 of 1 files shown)


mawww:
Already fixed by e42881fa380ee2f299dc00bfbc38356f7919b0d2, thanks for that PR.

PR #1108, opened 2017-01-04
update command names

alexherbo2:


casimir:
I would advise against changing all files at once, this will create conflicts against all currently opened PRs. It would be a better approach to do it step by step, encouraging people to do it in their respective PR.

alexherbo2:
I was thinking it would be the right time to change all that in once, as there is no merge conflict against the other pull requests (at the time being).

casimir:
I didn't check. I agree on the timing.

README.asciidoc
```diff
@@ -1185,24 +1185,24 @@ Manipulation of the displayed text is done through highlighters, which can be ad
 or removed with the command
 
 -----------------------------------------------------
```
mawww: Please fix these lines as well so that they match the length of the content.

alexherbo2: Done for all docs, but I did not touch tabs that could mess with the alignment.

PR #1109, opened 2017-01-04
Allow non-standard homebrew location for boost and ncurses linking

hoop33:
Enhances #1096 to allow non-standard homebrew installation location.


mawww:
Hello, doesn't that break non brew installation ?

I suspect `$(brewdir)` is passed as `PREFIX` to `make`, so maybe we could use that value to control the include directories ?

mawww:
Thank you for that.

Issue #1115, opened 2017-01-09
Binding to copy selections vertically to equal substrings


rouanth:
For example, regarding this snippet:
```rust
/// Brief description
///
/// Long
/// really long
/// ...
/// description
struct MyStruct {
    field: i8
}
```

I wish to select the whole documentation comment.
`a-ai` selects the whole text.
`a-ap` selects the comment along with the struct definition.
I could to to the start of the paragraph, extend to `struct` and go back a line, but it is quite hacky.

Another example:

```latex
\begin{tabular}{|c|c|c|}
\hline
n & Time (s) & k     \\ \hline
1 & 232.4    & 123.3 \\ \hline
2 & 344.8    & 552.9 \\ \hline
3 & 45       & 45    \\ \hline
4 & 44       & 44    \\ \hline
\end{tabular}

Some long, really long line.
```

The cursor is at the beginning. I need to delete the column `Time (s)`:
```latex
\begin{tabular}{|c|c|c|}
\hline
n & k     \\ \hline
1 & 123.3 \\ \hline
2 & 552.9 \\ \hline
3 & 45    \\ \hline
4 & 44    \\ \hline
\end{tabular}

Some long, really long line.
```

The following does the trick: `/T<ret>b;4Ct&d`. But I have multiple tables that are alike to this one, and they have a differing number of rows. So `4C` isn't really a solution. There is a workaround, however: indent a table multiple times, run `C` many times, then deindent after I'm done: `/T<ret>?end<ret>K>>><a-;>b;CCCCCCCCCCCt&d<<<`. I can't run `10C` because then it would affect the lines below. I hope we all agree that while this solution allows me to patch my tables it isn't the best one.

There are probably better ways to do this, but I think that the most lucid one would be possible if there was a way to select all `&` in the current continuous column if the cursor is on one of them. The same goes for selection of comments, selection of brackets in the example "Easily swap parameters with the 'rotate selection contents' command" on http://kakoune.org and in many other places.

danr:
Here are some ideas that do not use `C`:

For the comment+struct, you can select the paragraph, split it at lines `<A-s>`, keep the ones that contain a comment `<A-k>//<ret>`. If you want, you can now merge consecutive selections with `<A-m>`.

For the latex tables, select the paragraph with the tables, split it at lines, `<A-s>`, keep the ones that contain for instance `&` or `\\`. Now you have a selection at each line, and you can navigate around so you only select the second column and remove it.

[...]

PR #1116, opened 2017-01-09
Keybinding for copying selections on matching substrings vertically

rouanth:
A new keybinding, `^`, which allows for selecting substrings that are directly above or below the currently selected one if their contents match that of the original selection.

For example:
```rust
/// Long
/// comment
/// ...
fn main() { }
```
If the cursor is located at the beginning of the snippet, pressing `^` causes the `/` at the beginning of each of the first three lines to be selected.


danr:
This is an interesting keybinding (and I now understand what you meant with it)!
It does not need to be implemented in the C++ kakoune source though, it should be possible to get it going with some scripting.

lenormf:
As said previously, I don't think adding this to the primitives makes much sense, it sounds too specific, better keep it as a mapping in your user configuration.

mawww:
This is an interesting binding, but I am not convinced it has wide enough applications to warrant being a built-in, I think it should be doable through a mapping.

Do you have examples of editing problems that are cumbersome to do without that binding ? It seems in many cases `C` and `<a-k>` can be used.

rouanth:
I've provided some use cases in the issue which this request closes: https://github.com/mawww/kakoune/issues/1115. In general, I often encounter text which is distinguished by a vertical structure but isn't on a separate paragraph and doesn't have a fixed number of lines. For these cases, both `C` and `<a-k>` can be used perfectly well when editing manually; however, when doing automation with macros, `C` needs to know the number of lines, and the regular expressions required to do it with `<a-k>` in general case and without relying on its interactivity are often unnecessarily complex.

I will try to write a script that implements the keybinding, but I must warn that it can take very long for me to get to it. I guess the pull request can be rejected, but I'll leave the issue open if it's fine with you.

Thank you all for your analysis!

alexherbo2:
https://github.com/occivink/kakoune-vertical-selection

PR #1128, opened 2017-01-11
Fix fix of autoinsertion of comment starts

danr:
I didn't touch c-family.kak because I don't really dare touching it.

I added some tests for haskell and markdown.

It's a shame that there is so much copy-paste between the rc files.


mawww:
I applied a slightly different, fix, and cherry picked the test cases from this. Thanks for that PR.

PR #1133, opened 2017-01-13
Implement a test engine in python, deprecate the `run` shell script

lenormf:
Hi,

I implemented this a while ago, and I think it's a solid replacement for the current shell implementation. It's faster, has a more compact output than the current test suite script, and has builtin options.

```
usage: run [-h] [-d] [-v] [-f] [-t] [-p PATH_KAKOUNE]
           [directories [directories ...]]

Kakoune test engine

positional arguments:
  directories           Directories that contain the test suites

optional arguments:
  -h, --help            show this help message and exit
  -d, --debug           Display debug information
  -v, --verbose         Display more information
  -f, --fails-only      Only display tests that have failed
  -t, --no-timers       Do not time all the tests and give a final summary
                        about the time spent executing each test
  -p PATH_KAKOUNE, --path-kakoune PATH_KAKOUNE
                        Path to the Kakoune executable to test
```


lenormf:
The failing test looks a bit suspicious, could it have been written to make it work with a shell testing environment (the extra escapes are particularly weird) ?

lenormf:
For some reason, changing `echo "${kak_selections}"` to its `printf` equivalent fixed it, which is weird because the shell wasn't supposed to interpret the escape sequences in the variables at all (running the tests on my own system didn't not return _any_ failed test)… I'm getting a shellshock déjà-vu there.

lenormf:
Rebased on `706202218700c989b760942170515c7f3757290b`, migrated the code to python3.

PR #1134, opened 2017-01-13
Fixes to Perl syntax highliting

tadzik:
Whether variable declarations should be coloured as `identifier` or `value` is a topic for a discussion, but the other change should be pretty straightforward (unless there's a better way to express it in kakscript).


rc/base/perl.kak
```diff
@@ -8,6 +8,10 @@ hook global BufCreate .*\.pl %{
     set buffer filetype perl
 }
 
+hook global BufCreate .*\.pm %{
```
casimir: You don't need to create a new hook, just change the regex of the existing one.

tadzik: Good, point :) Fixed, thanks!

PR #1143, opened 2017-01-19
Tiny tweaks

ekie:


README.asciidoc
```diff
@@ -99,8 +100,8 @@ Building
 
 Kakoune dependencies are:
 
- * A C\+\+11 compliant compiler (GCC >= 5 or clang >= 3.4) along with its
-   associated C\+\+ standard library (libstdc\+\+ or libc\+\+)
+ * A {cpp} compliant compiler (GCC >= 5 or clang >= 3.4) along with its
```
Delapouite: Why removing the explicit version number in your adjustment?

README.asciidoc
```diff
@@ -99,8 +100,8 @@ Building
 
 Kakoune dependencies are:
 
- * A C\+\+11 compliant compiler (GCC >= 5 or clang >= 3.4) along with its
-   associated C\+\+ standard library (libstdc\+\+ or libc\+\+)
+ * A {cpp}11 compliant compiler (GCC >= 5 or clang >= 3.4) along with its
```
mawww: The readme is expected to be readable as text as well, so I'd rather keep `C++` written as-is in the source.

ekie: I understand but i think the way it gets rendered on github is weird:

`A C\\+11 compliant compiler (GCC >= 5 or clang >= 3.4) along with its associated C\\+ standard library (libstdc\\ or libc\\)`

mawww: OK yeah, its pretty tricky to fix due to how asciidoc works, fine...

Issue #973, opened 2016-12-14
Website only has WebM videos


kballard:
All of the videos on the website are WebM, which isn't supported by all browsers (notably, Safari doesn't support it). This means the website has a bunch of broken videos scattered around.

pho:
[asciinema](https://asciinema.org/) would be a nice alternative for this use case

lenormf:
I guess we could re-encode the videos with `x264` and have an imgur-type format selection, but at the time we decided that enough browsers could read webm so we didn't bother.

The alternative is also asciinema, but I don't like centralizing documentation resources (can we export asciinema sequences to a video, to back them up?).

pho:
Asciinema has a standalone player which could be embedded statically on the web, if thats what you were worried about: https://github.com/asciinema/asciinema-player

Although after a second thought, the work required to redo every example in asciinema is quite a lot, compared to reencoding and setting multiple sources on the <video>...

kballard:
What do you mean by "imgur-type format selection"? All you need is 2 `<source>` tags inside your `<video>` tag and the browser will pick whichever one it supports.

lenormf:
Imgur as a `gifv` filetype that does exactly what you described.

mawww:
Thats now fixed.

kballard:
The videos are still broken. I see you have .mp4 files now, but neither Safari nor QuickTime Player can play them (though Chrome can). Unfortunately, I'm not sure why it can't, but the problem is likely due to the mp4 profile or level.

lenormf:
Indeed, the videos were encoded with the `High` profile instead of the `Baseline` one. I'll fix it if someone opens an issue.

kballard:
Filed as #1178.

lenormf:
Thanks!

PR #1159, opened 2017-01-26
Update the showcase videos

lenormf:
The new videos are wider, and show exactly what keys were hit during the demo.

You can see the changes in action [on my repository](http://lenormf.github.io/kakoune/).


mawww:
The last video is missing.

lenormf:
Actually I decided no to redo it, as it's just showing the `:new` and `:make` commands, spelled out completely on the prompt. Should I redo it ?

alexherbo2:
I think because they are not the same size, just disable your tool to display keystrokes here.

danr:
Now also using paragraph objects 👍 
I think the fonts and colours will have a broader appeal, too.

lenormf:
>I think because they are not the same size, just disable your tool to display keystrokes here.

The reason is re-recorded the videos is to have the keys displayed as they're typed.

mawww:
I'd prefer for all videos to be same size, and use the same coloscheme.

occivink:
@lenormf 
>Also is it me or the video look a bit blurry?

vp8 is kind of a bad codec, you might get better results with vp9 which should have the same browser support.

lenormf:
I think most browsers only support vp8 for now (if they support it at all), that's why I went for it.

occivink:
I don't think that's correct, vp9 has been supported for a while by most browsers (with the notable exception of safari, which doesn't support vp8 anyway).

lenormf:
According to wikipedia:

>Internet Explorer and Apple Safari are missing VP9 support completely. In March 2016 an estimated 65 to 75% of webbrowsers in use on desktop and notebook systems were able to play VP9 videos in HTML5 webpages, based on data from StatCounter.

I guess that's good enough for us? @mawww are those numbers good enough? I don't know what browser users who end up on kakoune.org use mostly, but chances are they're on firefox/chrome.

occivink:
It's probably safe to ignore IE users, but there is a sizable osx userbase (#973) so ideally we should also have h264 versions of these videos. But my point is that vp8 is not going to have any more support than vp9.

lenormf:
I already have the x264 encoded videos (I just transcoded them into vp8 after recording them), I think I'll put them as alternative for the vp9 encoded clips. I might also add a couple other examples of basic things to do with the editor, so I promoted this PR to WIP, in the meantime,

lenormf:
I decided not to re-encode in vp9, as the only difference I saw on the videos was the size (⅓ bigger vp9 clips). The blurry effect I'm seeing comes from my browser, which is resizing the videos for no reason. Internet Explorer and Safari users will now be able to watch the clips, I've added the original mp4 recordings to the page.

I also added a new demo showing off how to wrap arbitrary lines in a document.

mawww:
Any chance to get the multi client video redone as well ? I'd really prefer if all videos were the same size, using the same colorschemes.

lenormf:
It will be done, yes.

lenormf:
I updated the last video on the list, I won't modify this branch anymore unless somebody has a suggestion.

PR #1160, opened 2017-01-27
Support the vim behaviour for +line syntax

ekie:
* Ommitting the linenumber will move to buffer end

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -819,6 +819,11 @@ int main(int argc, char* argv[])
         {
             if (not name.empty() and name[0_byte] == '+')
             {
+                if (name == "+" or name  == "+:")
+                {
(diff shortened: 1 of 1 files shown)


src/main.cc
```diff
@@ -819,6 +819,11 @@ int main(int argc, char* argv[])
         {
             if (not name.empty() and name[0_byte] == '+')
             {
+                if (name.begin()+1 == name.end() or *(name.begin()+1)  == ':')
```
lenormf: This check looks weird, are you sure you can't just check for the length of the `name` string and be done with it? Also apparently the `+:…` argument format is supported too?

mawww: Yeah, or just compare the whole string: `if (name == "+" or name == "+:")` will be more explicit, and this is not performance critical code.

PR #1176, opened 2017-02-01
Fix gc when buffer is smaller than window

### LINKED ISSUE #1170: gc should center with small buffers
When a buffer does not fill the window (eg. tildes appears), `gc` still center on window.
So if the buffer is less than half the screen, `gc` is behaving like `gj` and `gb`.

I think it should center of buffer. A 10 lines buffer on a 50 lines window, `gc` should go on line 5.

4d47:
An attempt to fix #1170


4d47:
I played a bit more with it and found the result of the division is better when ceil. Eg. gc on 3 lines should go to ln 2 no 1. So it's getting more ugly, maybe not worth it.

```c++
auto line = window.position().line - 1 + (std::min(buffer.line_count(), window.dimensions().line) + 2 - 1) / 2;
```

danr:
I don't think this keybinding should be changed in the C++ code. That said, here's an attempt to script the behaviour you want:

    def gc-4d47 %{
        eval -save-regs ce %{
            eval -draft %{
                exec gg
                exec gggb
                reg c %val{cursor_line}
                exec ge
                reg e %val{cursor_line}
            }
            %sh{
                echo echo $kak_reg_e $kak_reg_c
                if [[ "$kak_reg_e" -eq "$kak_reg_c" ]]; then
                    echo exec $(( "$kak_reg_c" / 2 ))g
                fi
            }
        }
    }
    map global goto gc-4d47

4d47:
Thanks for sharing this snippet @danr, I'll try this approch

4d47:
Just for reference, ended with this version.

```sh
def goto-middle -docstring "goto first non blank of middle displayed line" %{
    eval -save-regs tb %{
        eval -draft %{
            exec gt
            reg t %val{cursor_line}
            exec gb
            reg b %val{cursor_line}
        }
        %sh{
            echo exec gt$(( ($kak_reg_b - $kak_reg_t - 1 + 2 - 1) / 2 ))jgi
            # Ceil the result of division;
            # Using `j` instead of `g` so it does not change text position on screen;
            # Added `gi` at the ended for extra Vim like behavior;
        }
    }
}
```

mawww:
I think the current behaviour is correct, we are trying to jump to the middle of the window, and we end up clamped at the end of the buffer.

I think we could discuss which behaviour is the best one, but that in the end this corner case is not particularly important, and keeping the simpler code is better. Besides, I am happy to see that it can be implemented for those who really want the alternate behaviour.

PR #1181, opened 2017-02-02
Adds tomorrow-night theme and include and included faces.

nicholastmosher:
I made a colorscheme modeled after my favorite vim theme, tomorrow-night. As a side effect of the new theme, I created two new faces for use with the colorschemes, `include` and `included`. The intent is to be able to uniquely highlight a language's "include" statement (or whichever variation of an import keyword) and the header or package or whatever's being "included".

capture

### DIFF
--- a/colors/base16.kak
+++ b/colors/base16.kak
@@ -27,6 +27,8 @@
         face attribute ${orange_dark}
         face comment ${grey_dark}
         face meta ${orange_light}
+        face include default
+        face included default
         face builtin default+b
     "
 
--- a/colors/default.kak
+++ b/colors/default.kak
@@ -10,6 +10,8 @@ face operator yellow
 face attribute green
 face comment cyan
 face meta magenta
+face include default
+face included default
 face builtin default+b
 
 # For markup
--- a/colors/github.kak
+++ b/colors/github.kak
@@ -13,6 +13,8 @@ face operator yellow
 face attribute rgb:A71D5D
 face comment rgb:AAAAAA
 face meta rgb:183691
+face include default
+face included default
 face builtin default+b
 
 ## markup
--- a/colors/lucius.kak
+++ b/colors/lucius.kak
@@ -33,6 +33,8 @@
         face attribute ${lucius_light_blue}
         face comment ${lucius_grey}
         face meta ${lucius_purple}
+        face include default
+        face included default
         face builtin default+b
 
         # and markup
--- a/colors/reeder.kak
+++ b/colors/reeder.kak
@@ -31,7 +31,9 @@
         face attribute  ${green}
         face comment    ${brown_light}
         face meta       ${brown_dark}
-        face builtin   default+b
+        face include    default
+        face included   default
+        face builtin    default+b
 
         # and markup
         face title      ${orange}+b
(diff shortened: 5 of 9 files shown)


lenormf:
This reminds of something I've asked myself for a long time: should we allow language-specific faces into the colorscheme files? In this particular example, the `meta` face is ignored and overridden with a custom `include` one, and on top of that a subset of a `meta` line is also highlighted differently. with `included` (which I would have called `includeparam` I think, so that we could have a specialization for macros and macro names/definitions).

On one hand there is a precedent with the markup faces, but on the other hand every language has its own unique feature that could be highlighted with a custom face. Should we follow that logic and allow language scripts to modify default themes, or send the faces back to the language scripts themselves (which, thinking about it, might not be doable)?

mawww:
include/included seems too specific to me, I think we can already get good result by just adding an highlighter that applies the 'identifier' (or another one) to the filename part of `#include`.

I am thinking for example that in a language where we define and import modules, we probably want the module names to be highlighted the same way regardless of if its getting defined or imported, so included seems a wrong name.

A possible direction would be to rename `identifier` to `variable`, and add a `module` face (and eventually a `function` face as well).

nicholastmosher:
> should we allow language-specific faces into the colorscheme?

My thought when creating the new faces was that although not every language supports some sort of import statement, quite a large set of them do. In particular, when I'm reading c or cpp (usually the header files), it's easy for "#include" statements to get lost under the other # macros, so I wanted a way for them to be distinguishable, even though they are technically macros.

> include/included seems too specific to me

Do you mean the naming convention, or the specialization of the faces themselves?

> A possible direction would be to rename `identifier` to `variable`, and add a `module` face (and eventually a `function` face as well).

Would the idea behind renaming `identifier` be to use it for another purpose, or just to clarify the name? And about the module face, correct me if I'm wrong but that would be used to identify the name of the module (as in, replacing what I have currently as `included`) right?.

If you're also thinking about having a function face, should we add that now (populating it with `default` in the existing colorschemes) and make an issue for adding the rules for modules and functions to each language?

mawww:
> Would the idea behind renaming identifier be to use it for another purpose, or just to clarify the name? And about the module face, correct me if I'm wrong but that would be used to identify the name of the module (as in, replacing what I have currently as included) right?.

Yeah, the idea is that `identifier` is pretty wide as a face, so if we need to distinguish between variable names, module names, type names... We could replace it with a more detailed set.

That said, in my view just displaying the filename in `#include <filename>` as an identifier would be good enough to visualy distinguish it (keeping `#include` highlighted as a macro).

Maybe we could split that PR in two, I'll be happy to merge the colorscheme itself, and then we can see how expand highlighting support for `#includes`.

nicholastmosher:
Closed by #1185. The temporary solution for anyone who wants unique highlighting for "#include" statements is to place the following into `~/.config/kak/kakrc`.

```
%sh{
    includes="\h*(#include)\h+(\"[^\"]+?\"|<[^>]+?>)"
    printf %s "
        face include rgb:87afaf
        face included rgb:b5bd68
        add-highlighter -group /c/macro regex ${includes} 1:include 2:included
        add-highlighter -group /cpp/macro regex ${includes} 1:include 2:included
    "
}
```

lenormf:
I think the idea was to have another PR done in which there's a `variable` face replacing the `identifier` one, and a new `module` face that could be used to highlight the name of imported files/modules, directly from the language support scripts.

PR #1189, opened 2017-02-04
Tweak mapping docs

ekie:


doc/manpages/keys.asciidoc
```diff
@@ -629,4 +629,14 @@ map <scope> <mode> <key> <keys>
 *keys*::
 	string that lists the keys that will be executed when *key* is hit
 
-Keys are always executed within the *normal* mode.
+All mappings, except those bound to the `insert` mode, will be executed in
```
lenormf: Actually this bit (as well as what I wrote about the `map` command a while ago) is redundant with what's on the `commands` page. I've been thinking for a while now that mapping/unmapping probably deserve their own documentation page, considering the sustancial amount of things that could be said about the subject, what do you think?

README.asciidoc
```diff
@@ -1455,14 +1455,16 @@ You can redefine a key's meaning using the map command
 :map <scope> <mode> <key> <keys>
 --------------------------------
 
-with `scope` being one of `global`, `buffer` or `window` (or any prefix),
-mode being `insert`, `normal`, `prompt`, `menu` or `user` (or any prefix), `key` being
-a single key name and `keys` a list of keys.
+with `scope` being one of `global`, `buffer` or `window`, mode being `insert`,
+`normal`, `prompt`, `menu`, `goto`, `view`, `object`, `user` or a one character
+long arbitrary prefix, `key` being a single key name and `keys` a list of keys.
```
mawww: `one char long arbitrary prefix` seems strange, we allow any prefix.

doc/manpages/keys.asciidoc
mawww: Actually its only user and normal mode mapping that are executed in normal mode.

mawww:
I think this is superseeded by #1263, so closing.

Issue #166, opened 2014-06-16
Add show_whitespaces_in_selection highlighter


alexherbo2:

nicholastmosher:
Along the same lines as this issue, is there currently a way to customize how whitespaces are displayed? The only mention I can see in the repository of `show_whitespaces` is in a c++ file, which makes it seem inflexible to me.

Does there exist a way to create a highlighter (either with a regex or otherwise) that could achieve something like the following?

![capture](https://cloud.githubusercontent.com/assets/4210949/22610904/d0aa2ab6-ea35-11e6-9c3a-5badcc5a6979.png)

That's a capture of my vim setup that I've been trying to replicate with kakoune. @mawww, do you have any suggestions of how to do that with existing features, or would that require new features?

lenormf:
The characters used by the `show_whitepaces` highlighter are hardcoded, create a new issue if you want modularity on that front.

PR #1190, opened 2017-02-04
Allow modifying the characters used when highlighting whitespace

lenormf:
This commit adds the following flags to the `show_whitespaces`
highlighter, with a one character long parameter:

 * `-lf`: character replacing line feeds
 * `-spc`: character replacing spaces
 * `-nbsp`: character replacing non breakable spaces
 * `-tab`: character replacing a tabulation
 * `-tabpad`: character used as padding after a tabulation to satisfy
              the `tabstop` option

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -1227,9 +1227,8 @@ general highlighters are:
        <option_name>.
  * `show_matching`: highlight matching char of the character under the selections
        cursor using `MatchingChar` face.
- * `show_whitespaces`: display symbols on top of whitespaces to make them more explicit
-       using the Whitespace face.
- * `number_lines \<-relative> \<-hlcursor> \<-separator <separator text> >`: show line numbers.
+ * `show_whitespaces \<-tab <separator> \<-tabpad <separator> \<-lf <separator> \<-spc <separator> \<-nbsp <separator>`: display symbols on top of whitespaces to make them more explicit using the Whitespace face.
+ * `number_lines \<-relative> \<-hlcursor> \<-separator <separator text>`: show line numbers.
        The -relative switch will show line numbers relative to the main cursor line, the
        -hlcursor switch will highlight the cursor line with a separate face. With the
        -separator switch one can specify a string to separate the line numbers column with
--- a/doc/manpages/highlighters.asciidoc
+++ b/doc/manpages/highlighters.asciidoc
@@ -51,9 +51,25 @@ General highlighters
 	highlight matching char of the character under the selections cursor
 	using MatchingChar face
 
-*show_whitespaces*::
+*show_whitespaces* [options]::
 	display symbols on top of whitespaces to make them more explicit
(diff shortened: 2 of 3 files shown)


lenormf:
@nicholastmosher I've tried this with the settings you suggested in #166, here is what it looks like:



I couldn't figure out how to use a minus sign as an argument, so I used `~`.

Delapouite:
Great addition. Thanks for this lenormf!

Nothing dramatic: what about `-space` instead of `-spc`, in order to align with other kakoune long names everywhere? (Bonus, it aligns well with vim `listchars` options)

alexherbo2:
@Delapouite I was thinking the same for the other flags as well.

- `-line-feed`
- `-space`
- `-non-breaking-space`
- `-tab(ulation)?`
- `tab(ulation)?-padding`

lenormf:
I thought that using something like `-non-breaking-space` would be too annoying, especially when we when to use all five flags at once, so I abreviated everything. I'll change them to whatever is best, just let me know!

doppioandante:
Tab and nbsp are common abbreviations but where is spc used instead of space?

lenormf:
Because I didn't spend more than .1µs thinking about the names of the flags, I just wanted to get something to work first.

nicholastmosher:
I just merged this into my local install of kakoune, one thing I'm trying to figure out is how to set the face of the whitespace characters. @lenormf do you have a custom face specifically for the wihtespaces in your screen capture or is that just how it turned out? And if there isn't a way to specify that, is that something you could add?

ekie:
:face Whitespace facespec

lenormf:
Set the `Whitespace` face to whatever you want, you can override at runtime from the prompt.

nicholastmosher:
I found something that's not necessarily a bug; the behavior is exactly as prescribed, but something that might be worth looking for a solution. I'm using the following for whitespace options: `addhl show_whitespaces -tab > -tabpad - -lf $`. What I've found is that when aligning text using `&`, the tabs stay fixed-width, so that even though the representation of tabs is correct, they're displayed incorrectly. Below is a snapshot of a selection after running `s=<ret>&`:

capture

Opening this in another editor shows that the `=`s are properly aligned, but the explicit whitespace is causing an incorrect display. Below is how I'd expect it to look (how it's shown in my vim setup):

capture

lenormf:
You split using `=`, which kept the spaces around the equal signs, so the output shown in the screenshot seems normal to me. To align properly, you would have to trim the whitespace around the equal sign beforehand, which can be annoying that's why I came up with an `alignr` script that does that for you (it would be used as `:alignr wS=`).

In other editors, sometimes tabs are not rendered entirely to favor visual comfort, we can see that on your second screenshot: some tabs have less than 3 minus signs, and kakoune doesn't try to do that as far as I know which might be why you're confused.

PR #1193, opened 2017-02-05
Allow flags to have a minus sign as argument

lenormf:
Hi,

While implementing #1190 I noticed that we can't use `-` as argument for a flag, so I fixed this. Note that this only applies to the argument of flags that require one, the `--` separator still has to be used for positional parameters.

HTH.


Delapouite:
Thanks for fixing this issue! This one is also somehow related : https://github.com/mawww/kakoune/issues/1220

Issue #1186, opened 2017-02-03
Make clarifications and specializations in colorscheme faces.


nicholastmosher:
A continuation of some points brought up in #1181.

The primary ideas are:
* Disambiguate the `identifier` face by renaming it to `variable`.
* Add new faces, `module` and `function`:
  * `module` to highlight the names of imported resources (e.g. `<stdio.h>` or `java.lang.String`)
  * `function` to highlight functions defined previously in the file, working directory, or included modules.
* Discuss the merit of adding some sort of `import` face (may need better name suggestions), which would highlight the keyword(s) of a language that performs the module importing (e.g. `#include` or `require( )`).

mawww:
This has been fixed.

PR #1205, opened 2017-02-10
Adds faces module and function. Renames identifier face to variable.

nicholastmosher:
Addresses some ideas from #1186.

The renamed face `variable` keeps the color it originally held when named `identifier` in order to keep the consistent feel of each theme.

In most cases, the `module` face adopts the same color as the colorscheme's `string` face, except in some cases when a colorscheme has the same color for `meta` and `string`, in which case `module` adopts another reasonable color from the scheme. Currently the only highlighter that uses the `module` face is in `c-family.kak`, which highlights terms following a `#include`.

The `function` face adopts colors from each colorscheme so as to be consistent with the theme. In each case, the adopted color was chosen in an attempt not to over saturate any given color in a theme. Since no highlighters currently use the `function` face, it's hard to tell whether I was successful in that sense.

I'm eager for suggestions of improvements, and for followup contributions to implement the `function` and `module` faces (which should probably reference #1186 to make a nice record of changes directly caused by that issue).


lenormf:
As much as having a `function` face would be nice, it would be very hard to parse function names in C/C++ like languages with a simple regex, so I'm not sure this new face is useful yet (unless there are cases in which it can be used reliably, in some language).

nicholastmosher:
Yeah, I was thinking the same thing. I'm going to be brainstorming over the next few days to see if I can come up with a fair solution to that. One thing I've thought of so far is that in C/C++, functions are usually immediately followed by parentheses, so that could be a naive way to do it. This wouldn't work in the case of function references, like `void (*my_func_ptr)(void) = my_func;`. I'm still fairly new to the kak syntax, but I'm thinking there's probably a way to create a hidden variable that can be populated with function names (provided we can reliably identify those), and add those to a dynamic regex highlighter. I believe @mawww did something similar with his `curword` highlighter in his [kakrc](https://github.com/mawww/config/blob/master/kakrc#L31).

lenormf:
Unfortunately, it's impossible to get function names with a regex, we would need to actually parse the buffer to get reliable results, and using a naive regex that isn't too big would cause false positives. We don't need a dynamic highlighter either, and storing function names in a variable in an option wouldn't be very clean, and also cause false positives.

I don't think it's doable with a reasonable solution for C/C++, but maybe other languages could benefit from it? Python, Go, Lua could use that, I reckon.

mawww:
I dont think highlighting functions in c like languages is very important, I'd think lisp-like languages are the ones where highlighting the first element of a s-expression differently would be nice.

We dont have to use the faces everywhere, so not highlighting C like function names with the function face seems ok, we could even decide to use the `function` face for `function-like` builtins (sizeof, offsetof, .*_cast...).

PR #1211, opened 2017-02-13
Add elm language support

tmiller:
Adding language support for [elm](http://elm-lang.org). Using the Haskell language file for a starting point since the syntax is very similar.


lenormf:
Could you please squash all those commits that modify the file into a single one?

casimir:
Also I would have put this one in `rc/extra/` instead of `rc/base/`, just like `coffee.kak`.

tmiller:
Thanks, I've squashed the commits and moved the file to `rc/extra`.

PR #1213, opened 2017-02-13
Tighten the build options in travis

lenormf:
Hi,

This commit adds the `-Werror` flag to the compilation flags on travis, in order to treat warnings as errors. I realize it might not sound so great because there are warning that we know won't do any harm, but down the line I think it's worth it, as some errors might actually cause issues and would have otherwise been ignored (e.g. the 'no return in a function returning non-void' warning that currently stops the compilation).

I also wanted to compile using clang3.4 on travis, since that's the minimum version that is documented in the build requirements, but setting it up proved to be a hassle so I let 3.5 live for now.

HTH.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ install:
     elif [ $TRAVIS_OS_NAME = osx ]; then
        brew outdated boost || brew upgrade boost;
     fi;
+    export CXXFLAGS="${CXXFLAGS} -Werror -Wno-unknown-attributes";
 env:
   global:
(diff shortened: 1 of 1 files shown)


lenormf:
Relies on #1241 to pass the travis build.

mawww:
I dont think we should use different options between travis and the regular build. We could enable -Werror, but we still have some irrelevant warnings in release build (about unused variables in unit test code), and I dont feel like introducing a macro to silence them, and the maybe_unused attribute is only C++17.

In any case, I wont enable -Werror only during CI, but I agree we should try to have it on all the time.

PR #1226, opened 2017-02-19
Properly wrap `kak_assert` into a do-while scope

lenormf:
Expanding the `kak_assert` macro to either an `if` statement or nothing
leads to issues when the macro is used in a conditional statement that
doesn't use braces.

Example: ncurses_ui.cc:476, in non debug mode, the macro will expand to
an empty line, resulting in the `ungetch` call not being executed if the
`ioctl` call succeeds (line 448).

### DIFF
--- a/src/assert.hh
+++ b/src/assert.hh
@@ -17,12 +17,13 @@ void on_assert_failed(const char* message);
 #define TOSTRING(X) STRINGIFY(X)
 
 #ifdef KAK_DEBUG
-    #define kak_assert(...) \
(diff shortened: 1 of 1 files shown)


mawww:
I dont think the ncurses_ui.cc:476 case is problematic, as we still have the `;` after the macro.

However the current implementation will do the wrong thing with the following construct:
```
if (condition)
   kak_assert(something);
else
  somthing_else();
```
As the else will bind to the if contained inside kak_assert.

Issue #1216, opened 2017-02-14
Pipes: | appends a newline to the selection, if necessary, while <a-|> does not.


bobi32:
Yesterday I encountered some strange behavior while piping a selection into my clipboard with both these commands (long story short: `<a-|>` systematically lost the last character of the selection while `|` did not).
Then I led further tests and it turns out that `|` does append a newline to the selection, while `<a-|>` does not.

After a quick search on the repository, it seems that the issue comes from **src/normal.cc**. The git diff below implements a quick working patch:
```
diff --git a/src/normal.cc b/src/normal.cc
index 6a34f6a..db346cb 100644
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -454,8 +454,11 @@ void pipe(Context& context, NormalParams)
                 for (int i = 0; i < selections.size(); ++i)
                 {
                     selections.set_main_index(i);
+                    String in = content(buffer, selections.main());
+                    if (in.back()!='\n')
+                        in += '\n';
                     ShellManager::instance().eval(real_cmd, context,
-                                                  content(buffer, selections.main()),
+                                                  in,
                                                   ShellManager::Flags::None);
                 }
             }
```

mawww:
Hello,

I am not sure about this, in `|`, we assume the command is a filter, which usually wants full lines in the unix world. In the case we do add a trailing `\n`, we then remove it from the filter output.

For `<a-|>`, we dont know what the command is doing, and adding an eol is often a bad choice. The common use case for `<a-|>` is to pipe to `xsel` or `xclip` to copy the selection to X11 clipboard. Adding an eol there is clearly not the desired behaviour.

So, I dont think we can improve consistency between `|` and `<a-|>` by adding an eol in `<a-|>`, because `|` compensate that eol afterwards, `<a-|>` has no way to do that, and the eol is going to break the most common use case for that feature.

Out of curiosity, what was the use case that lead you to want this change ?

bobi32:
Well, I started using `|` and `<a-|>` with X clipboard, but rather than using xsel or xclip, I used my clipboard manager (fyi, GPaste CLI tool `gpaste-client`: https://github.com/Keruspe/GPaste).

It is when I started to cut (with `|`, taking advantage that `gpaste-client`, with no further options, outputs nothing when reading from input pipe — just like `xsel -i` and `xclip -i` afaik) and copy (with `<a-|>`) selections that I noticed the different behaviors aforementioned.

[...]

PR #1232, opened 2017-02-21
Consistent | and <a-|>

bobi32:
Fixing the different behaviors between | and <a-|> commands (now both append a newline to each selection if necessary)


lenormf:
Assuming this change gets merged in, could you please amend your first commit and have the last line say "Fixes #1216" ?

bobi32:
This is done!

PR #1234, opened 2017-02-21
Improved haskell comment regex

omar-polo:
As title says, I tried to improve a bit the way kakoune highlights haskell comments, even if I'm not an haskell developer.

the following screen shows the difference

09-30-2017-02-21_1366x768


mawww:
Thank you.

PR #1238, opened 2017-02-23
[WIP] contrib: Add a Dockerfile for CI builds

lenormf:
Hi,

Travis doesn't offer much margin when it comes to software versions, so I came up with a `docker` image that allows to compile the project using different compilers.

Even though it takes more space because of base files duplication, I went with `multistrap` so that the build environments would be segregated, and also to let the door open to cross-compiling if we ever need it (an ARM build could be nice).

Not sure if this really belongs to `contrib`, but here it is. Read the commit message for more info.

### DIFF
--- a/contrib/docker-build/Dockerfile
+++ b/contrib/docker-build/Dockerfile
@@ -1,0 +1,134 @@
+##
+## Dockerfile by lenormf
+##
+## Usage: docker build -t $USER/kakoune-build .
+## Once the image has been created: docker run --privileged $USER/kakoune-build [compiler]...
+## Privileges are needed in order to forward the host's /proc to the underlying chroot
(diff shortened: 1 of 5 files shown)


lenormf:
There's still room for improvement, I think being able to select a branch/fork and specifying arbitrary `CXXFLAGS` would be really useful, so I'm putting this in WIP.

lenormf:
Other needed feature: different toolsets per system, i.e. `bsd`, `gnu`, `busybox`.

mawww:
Is that still WIP, or do we want to merge it now ?

lenormf:
I might have to change the way the Dockerfile behaves and I'm not sure when I'm going to do that, so no merge for now. I can close the PR and open it when that's done, if needed.

PR #1243, opened 2017-02-26
add rc/ocaml.kak

alexherbo2:


rc/base/ocaml.kak
```diff
@@ -0,0 +1,50 @@
+# http://ocaml.org
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+# require ocp-indent
+
+decl str-list ocaml_keywords and:as:asr:assert:begin:class:constraint:do:done:downto:else:end:exception:external:false:for:fun:function:functor:if:in:include:inherit:initializer:land:lazy:let:lor:lsl:lsr:lxor:match:method:mod:module:mutable:new:nonrec:object:of:open:or:private:rec:sig:struct:then:to:true:try:type:val:virtual:when:while:with
```
mawww: I'd rather not have this as an option, but in a temporary shell variable like in other support files. We dont need to store this value through the whole process.

alexherbo2: Done

PR #1249, opened 2017-03-01
Add racer completion for rust

casimir:
Finally found some time to rewrote this. All you need is racer installed and in your `$PATH`.

### DIFF
--- a/rc/extra/racer.kak
+++ b/rc/extra/racer.kak
@@ -1,0 +1,62 @@
+decl -hidden str racer_tmp_dir
+decl -hidden completions racer_completions
+
+def racer-complete -docstring "Complete the current selection with racer" %{
+    %sh{
+        dir=$(mktemp -d -t kak-racer.XXXXXXXX)
+        printf %s\\n "set buffer racer_tmp_dir ${dir}"
+        printf %s\\n "eval -no-hooks %{ write ${dir}/buf }"
+    }
+    %sh{
+        dir=${kak_opt_racer_tmp_dir}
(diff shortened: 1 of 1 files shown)


dpc:
OK, so after some debugging I've figured out that it somewhat works, but not display a lot of stuff. What I'm doing is just completing stuff like `std::sync::` and other prefixes, and a lot of things are missing.

dpc:
I added some dumps here and there and it looks that `racer_completion` is being set to stuff that `kak` doesn't like. Eg. when match has an example.

dpc:
My current workaround. Change the `candidate = ...` line to:

```
candidate = word "|" word "|" word 
```

casimir:
It looks like the description need more escaping. I'll look into it. If that's the case you should have some output in *debug*.

dpc:
I'm a kak noob, and while I did play with `*debug*` when investigating my own scripting mitsakes, I couldn't get it to help me with this one. I actually resorted to adding bunch of `echo "$somethig" > /tmp/dump_something` in `racer.kak`.

casimir:
Ok, I could look into it. It was simpler than I expected and is fixed by #1258. 

@dpc if you want some more details: completion list syntax in Kakoune use `:` and `|` as separators, I escaped the `:`s but totally forgot about rust's syntax for closures and thought `|`s could not be found in the doc. Escaping them too fixes the behavior. Note that Kakoune can't show an info window when it's too big so you won't have the documentation for most packages anyway.

PR #1258, opened 2017-03-06
racer: escape pipe from closure in the description

casimir:
The rust documentation can be pretty extensive for stdlib containing for example... examples. This patch fixes the completion when there is a closure in this documentation.

### DIFF
--- a/rc/extra/racer.kak
+++ b/rc/extra/racer.kak
@@ -31,6 +31,11 @@ def racer-complete -docstring "Complete the current selection with racer" %{
                         sub(word, "{default+e}" word "{default+d}", menu)
                         menu = "{default+d}" menu
                         word = word "("
+                    } else if (type == "Enum") {
+                        menu = substr(menu, 0, length(menu) - 2)
+                        sub(word, "{default+e}" word "{default+d}", menu)
+                        menu = "{default+d}" menu
+                        word = word "::"
                     } else {
                         menu = "{default+e}" word "{default+d} " menu
                     }
--- a/rc/extra/racer.kak
+++ b/rc/extra/racer.kak
@@ -22,6 +22,7 @@ def racer-complete -docstring "Complete the current selection with racer" %{
                     word = $2
                     type = $7
                     desc = substr($9, 2, length($9) - 2)
+                    gsub(/\|/, "\\|", desc)
                     gsub(/\\n/, "\n", desc)
                     menu = $8
                     sub(/^pub /, "", menu)


dpc:
It definitely fixes most of the problems, but I wonder if there's more into it.

Eg. `std::` does give me a long list of completions, bug `std::sync::` does not show anything until I type another letter.

lenormf:
@dpc Actually I've had this problem too using a custom completion function in one of my scripts, and I couldn't figure out what's happening but the symptoms are exactly what you describe.

@mawww when the last letter of a word that should be completed upon is typed, the completion menu doesn't show up even though the `completion` option associated to the completer is indeed set to the proper list of candidates with a timestamp/coordinates. The timestamp is the right one, the candidates are all simple (e.g. `a|a|a`), but for some reason the menu tends to show up on the previous-before-last letter typed (triggering the `InsertIdle` hook that sets the `completion` variable), and not on the last letter.

Code:

```
decl int dict_max_suggestions 3
decl int dict_min_chars 3
decl str dict_lang "en"

decl -hidden completions dict_completions

def -hidden -params 1 dict-complete %{
    set buffer dict_completions %sh{
        export kak_coord_word_x=$((kak_cursor_column - ${#1}))

        printf %s\\n "${1}" | aspell -a -l "${kak_opt_dict_lang:-en}" | awk '
            /^&[^:]+: (.+)/ {
                FS=", "
                sub(/^&[^:]+: /, "", $0)
                prefix = ENVIRON["kak_cursor_line"] "." ENVIRON["kak_coord_word_x"] "@" ENVIRON["kak_timestamp"]
                max_suggestions = ENVIRON["kak_opt_dict_max_suggestions"]

                if (NF < 1) exit
                printf "%s", prefix
                for (i = 1; i <= NF; i++) {
                    if (i > max_suggestions) break
                    printf ":%s|%s|%s", $i, $i, $i
                }
            }
        '
    }
}

hook global WinSetOption filetype=(plain|asciidoc|markdown|git-commit) %{
    set window completers "option=dict_completions:%opt{completers}"
    hook -group dictcomplete buffer InsertIdle .* %{
        try %{
            eval -save-regs m %{
                exec -draft h<a-i>W "<a-k>[^\n]{%opt{dict_min_chars},}<ret>" \"my
                dict-complete %reg{m}
            }
        } catch %{
            set buffer dict_completions ''
        }
    }
}
```

dpc:
@casimir Slightly OT, but do you happen to have any other Rust enhancements? `cargo` as a make command, https://github.com/dan-t/rusty-tags?

casimir:
Got it, that was an unbalanced `{` when completing enum (e.g. `std::sync::TryLock`).

@dpc I don't use a lot more than the builtin kak scripts, I forced myself to share my file so here you go: https://github.com/casimir/kakoune-extra/blob/master/lang/rust.kak. Although `cargo` has make command work perfectly I tend to open a window with a `watchexec cargo test` to get the color. Cargo output can be hard to read sometimes. 
I don't know what rusty-tags does but universal-ctags support rust perfectly, and `ctags.kak` also. I just execute `ctags --languages=rust -R src` because I also want to implement source jump with racer.

dpc:
@casimir Thank you very much. I'll try it out when I have a chance.

PR #1263, opened 2017-03-07
doc: Write a dedicated "mapping" page

lenormf:


doc/manpages/mapping.asciidoc
```diff
@@ -0,0 +1,45 @@
...
+		normal, commands mode
+	*prompt*::
+		prompts, inputs
+	*menu*::
+		multiple selection menus
+	*user*::
+		user mode entered when the prefix (default: *,*) is hit
+
+Note that despite the *mode* in which the bindings are initialized, the
+keys are always executed in a 'normal' execution context.
```
danr: Not so; they are executed in the same scope as they are bound to. The modes `view` and `goto` are missing.

lenormf: It's weird, that line came from `keys.asciidoc` and even then it wasn't the case IIRC.

danr: There was a discussion about this in #1189. I now realized `object` is also missing.

lenormf:
Moves the changes requested over from #1189.

doc/manpages/mapping.asciidoc
mawww: It was the case at some point, but we realised it prevented some legitimate use cases and went back to running them directly in the current mode.

doc/manpages/mapping.asciidoc
```diff
@@ -0,0 +1,52 @@
+KAKOUNE(1)
+==========
+
+NAME
+----
+mapping - a
+
+Description
+-----------
+Creating and removing shortcuts boils down to the following functions,
```
mawww: commands, not functions

doc/manpages/mapping.asciidoc
```diff
@@ -0,0 +1,52 @@
...
+map [flags] <scope> <mode> <key> <keys>
+unmap <scope> <mode> <key> [<expected>]
+---------------------------------------
+
+The *map* command makes *key* behave as if the *keys* sequence was typed.
+
+*mode* dictates in what context the mapping will be available:
+
+	*insert*::
+		insert mode (default: 'i')
```
mawww: not sure what that `default: 'i'` means, I'd think insert mode is explicit enough

doc/manpages/mapping.asciidoc
```diff
@@ -0,0 +1,52 @@
...
+	*normal*::
+		normal, commands mode
```
mawww: `commands` might just add confusion, normal mode ought to be explicit enough.

doc/manpages/mapping.asciidoc
```diff
@@ -0,0 +1,52 @@
...
+	*prompt*::
+		prompts, inputs (e.g. '/')
```
mawww: `prompts, such as when entering a command through *:*, or a regex through */*`

doc/manpages/mapping.asciidoc
```diff
@@ -0,0 +1,52 @@
...
+	*menu*::
+		multiple items dropdown menus (e.g. ':')
```
mawww: Hum, not really, the menu mode is a specific one, which is actually pretty rare, AFAIR only ctags.kak and spell.kak use it. In menu mode we dont have anything else we can do than select a menu item.

That makes me think it might be time to rework that menu mode, but its another story.

lenormf: I really had no idea what the difference was with `prompt`, so I used my best judgement. Also I put default keys on the other mode explanations because I put one for this mode, resulting in all the above remarks.

lenormf:
Still need to come up with a good description for the `menu` mode, unless this is going to be changed in a near future in which case we can probably afford to be vague about it.

danr:
The menu mode is entered when a `:menu` command is run.

lenormf:
I added a description for the `menu` mode, should be good to go now.

PR #1264, opened 2017-03-07
src: Support the `-help` flag

lenormf:
Output:

```
> ./kak -help
Usage: ./kak [options] [file]... [+<line>[:<col>]|+:]

Options:
 -c <arg>: connect to given session
 -e <arg>: execute argument on initialisation
 -n : do not source kakrc files on startup
 -s <arg>: set session name
 -d : run as a headless session (requires -s)
 -p <arg>: just send stdin as commands to the given session
 -f <arg>: act as a filter, executing given keys on given files
 -q : in filter mode, be quiet about errors applying keys
 -ui <arg>: set the type of user interface to use (ncurses, dummy, or json)
 -l : list existing sessions
 -clear : clear dead sessions
 -ro : readonly mode
 -help : display a help message

Prefixing a positional argument with a plus (`+`) sign will place the
cursor at a given set of coordinates, or the end of the buffer if the plus
sign is followed only by a colon (`:`)
```

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -801,11 +801,32 @@ int main(int argc, char* argv[])
                    { "ui", { true, "set the type of user interface to use (ncurses, dummy, or json)" } },
                    { "l", { false, "list existing sessions" } },
                    { "clear", { false, "clear dead sessions" } },
-                   { "ro", { false, "readonly mode" } } }
+                   { "ro", { false, "readonly mode" } },
+                   { "help", { false, "display a help message and quit" } } }
     };
+
     try
     {
+        auto show_usage = [&]()
+        {
+            write_stdout(format("Usage: {} [options] [file]... [+<line>[:<col>]|+:]\n\n"
+                    "Options:\n"
+                    "{}\n"
+                    "Prefixing a positional argument with a plus (`+`) sign will place the\n"
+                    "cursor at a given set of coordinates, or the end of the buffer if the plus\n"
+                    "sign is followed only by a colon (`:`)\n",
+                    argv[0], generate_switches_doc(param_desc.switches)));
+            return 0;
+        };
+
+        if (contains(ConstArrayView<char*>{argv+1, (size_t)argc-1}, StringView{"--help"}))
+            return show_usage();
+
         ParametersParser parser(params, param_desc);
+
+        const bool show_help_message = (bool)parser.get_switch("help");
+        if (show_help_message)
+            return show_usage();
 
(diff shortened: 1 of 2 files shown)


src/main.cc
```diff
@@ -798,6 +798,20 @@ int main(int argc, char* argv[])
                    { "clear", { false, "clear dead sessions" } },
                    { "ro", { false, "readonly mode" } } }
     };
+
+    if (argc == 2 and String(argv[1]) == "--help")
```
mawww: Use the parameter parser here, and `-help` instead of `--help`, we dont use the long option syntax anywhere.

lenormf: Actually I didn't only do this PR to have a nice summary of the CLI parameters available, but to also follow the long help option convention (most if not all utilities implement this, regardless of the amount of minus signs they use for other options).

lenormf:
I think these last changes work better, c.f. commit message.

src/main.cc
```diff
@@ -810,6 +812,19 @@ int main(int argc, char* argv[])
 
         ParametersParser parser(params, param_desc);
 
+        const bool show_help_message = (bool)parser.get_switch("help");
+        if (show_help_message or (argc == 2 and String(argv[1]) == "--help"))
```
mawww: how about `contains(ArrayView{argv+1, argv+argc},  StringView{"--help"})`

lenormf:
I rebased the branch against `master`, fixed an issue where the `or` part of my `if` statement was never parsed (because the `ParameterParser` was throwing an exception before we reached that statement), and used the following more idiomatic way to check for a value in an array:

```c++
        if (contains(ConstArrayView<char*>{argv+1, argc-1}, StringView{"--help"}))
```

src/main.cc
```diff
@@ -801,12 +801,38 @@ int main(int argc, char* argv[])
...
                    { "l", { false, "list existing sessions" } },
                    { "clear", { false, "clear dead sessions" } },
-                   { "ro", { false, "readonly mode" } } }
+                   { "ro", { false, "readonly mode" } },
+                   { "help", { false, "display a help message and quit" } } }
     };
+
     try
     {
+        auto show_usage = [&](const ParameterDesc &param_desc)
```
mawww: Its always going to be the local param_desc, so no need to pass it as a a parameter.

lenormf:
Fixed the `show_usage` lambda's prototype, and made it return zero to save a few lines.

PR #1277, opened 2017-03-14
src: Update the macro register after each keypress

lenormf:


mawww:
Whats the goal there ? If its for displaying the keys being recorded, I think a hard coded behaviour in `generate_context_info` showing the keys in `recording (<reg>=<keys>...)` would be good enough.

lenormf:
I don't care much for the issue, having the register up-to-date at any point in time makes more sense if it doesn't introduce new issues.

mawww:
We are recording, it seems to make as much sense to consider the register should be updated once we finish recording than update it on the go. This is why I wonder if we have a use case for that feature, or if some behaviour gets better once we do that.

lenormf:
One possible case would be the issue linked to this PR, I don't have any personal motivation to have that merged otherwise.

mawww:
Reopen if we have a proper reason to do that, for now lets keep things simple.

PR #1278, opened 2017-03-14
src: Align the assistant in the middle of the popup

lenormf:
It gets pretty ugly when there are many rows in a popup, and the assitant is top-aligned (e.g. bindings available on the user prefix).

HTH.

### DIFF
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -859,11 +859,17 @@ String make_info_box(StringView title, StringView message, ColumnCount max_width
 
     auto line_count = max(assistant_size.line-1,
                           LineCount{(int)lines.size()} + 2);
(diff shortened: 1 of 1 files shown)


mawww:
Not sure about this one, anybody got an opinion on assistant alignement (and no, I wont add support for an option to choose that) ?

lenormf:
I didn't realize it would be that big of a deal! No need for an option, I think it makes sense to center the assistant in a window with many rows, instead of keeping it at the top (dead space isn't really sexy).

ekie:
i dont have the assistant enabled, but i would even say that the alignment of the assistant as it is now is distracting with many rows and disturbs me parsing the info box. seems like a sane addition.

PR #1279, opened 2017-03-15
src: Make the cursor character an opening delimiter

lenormf:
This commit allows the editor to consider the character under the cursor
as an opening delimiter when using an object selector, instead of
ignoring it and looking for one before the cursor.


casimir:
I second that.

PR #1291, opened 2017-03-22
src: Introduce a `-i` suffix flag for filter backups

### LINKED ISSUE #1288: .kak-bak files from running in filter mode ( kak -f )
Running kakoune in filter mode leaves a backup file behind. Would it be OK with you to allow **not** creating that file?

The way `sed` works is `-i[SUFFIX]` meaning if you don't provide any suffix no backup is created, otherwise `[FILE][SUFFIX]` is created as a backup file.

So the options I see are:
- changing how `-f` works by allowing to (or not to) specify a custom suffix, dropping the `.kak-bak` default
- adding another flag that disables that `.kak-bak` file generation

Thoughts?

lenormf:
This commit allows the user to chose to backup the files on which a
filter has been run, by specifying a suffix for the backup file. The
former implementation always backed up the files with a hardcoded
".kak-bak" suffix.

When no suffix is specified on the command line, the files are not
saved.

Fixes #1288

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -642,7 +642,7 @@ int run_server(StringView session,
     return 0;
 }
 
-int run_filter(StringView keystr, StringView commands, ConstArrayView<StringView> files, bool quiet)
+int run_filter(StringView keystr, StringView commands, ConstArrayView<StringView> files, bool quiet, StringView suffix_backup)
 {
     StringRegistry  string_registry;
     GlobalScope     global_scope;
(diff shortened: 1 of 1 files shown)


src/main.cc
```diff
@@ -689,7 +689,8 @@ int run_filter(StringView keystr, StringView commands, ConstArrayView<StringView
         for (auto& file : files)
         {
             Buffer* buffer = open_file_buffer(file);
-            write_buffer_to_file(*buffer, file + ".kak-bak");
+            if (suffix_backup.length())
```
mawww: `not suffix_backup.empty()` is more explicit.

lenormf:
Now using `not suffix_backup.empty()` instead of `suffix_backup.length()`.

PR #1296, opened 2017-03-24
Added gruvbox colorscheme

laelath:
Adapted my favorite vim colorscheme to kakoune, tried to be as faithful as possible to the original vim colors.


lenormf:
Nice work, I had migrated that theme over a while back myself, but found that the colors do not contrast enough with each other, and ended up modifying it to end up with a customized gruvbox.

Left is yours, right is my modified version:

PR #1302, opened 2017-03-29
Make ctags.kak work without coreutils

casimir:
This is not the prettiest fix but with this you don't need GNU readlink to use ctags. I've used it for some time on macOS and Archlinux.

### DIFF
--- a/rc/base/ctags.kak
+++ b/rc/base/ctags.kak
@@ -7,9 +7,12 @@ decl str-list ctagsfiles 'tags'
 
 def -params ..1 \
     -shell-candidates '
+        realpath() {
+            echo $(cd -P -- "$(dirname -- "$1")" && echo "$(pwd -P)/$(basename -- "$1")")
+        }
         printf %s\\n "$kak_opt_ctagsfiles" | tr \':\' \'\n\' |
         while read -r candidate; do
-            [ -f "$candidate" ] && readlink -f "$candidate"
+            [ -f "$candidate" ] && realpath "$candidate"
         done | awk \'!x[$0]++\' | # remove duplicates
         while read -r tags; do
             namecache="${tags%/*}/.kak.${tags##*/}.namecache"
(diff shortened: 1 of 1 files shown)


rc/base/ctags.kak
```diff
@@ -7,9 +7,12 @@ decl str-list ctagsfiles 'tags'
 
 def -params ..1 \
     -shell-candidates '
+        realpath() {
+            echo $(cd -P -- "$(dirname -- "$1")" && echo "$(pwd -P)/$(basename -- "$1")")
```
lenormf: You could save a subshell by using `${1##*/}` instead of `basename`.

casimir:
I'm not familiar with variable expansion. Is it POSIX?

ekie:
yes it is.

lenormf:
[Yes](http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_02).

casimir:
I'll gladly change it then.

casimir:
And it's done.

casimir:
I guess `dirname` could also be replace by `${1##/*}`.

lenormf:
You probably meant `${1%/*}`, and `dirname` handles cases such as `/root` which would turn into an empty string with the shell expansion.

I couldn't get this function to do what it's supposed to on my system, I'll look into it (also in the script we don't need `readlink -f`, we could use `readlink -e` directly since we check that the candidate exists beforehand, maybe there's a shortcut to be taken there).

lenormf:
A possible alternative using less subshells:

```
expr "$(ls -l /path/to/link)" : '.*[[:space:]]->[[:space:]]\([^>][^>]*\)$'
```

It's ugly, but I don't think there's any elegant solution to this unfortunately.

casimir:
@lenormf I'm not sure what it's supposed to do but on my machine it only print a `\n`.

lenormf:
The idea:

```
> ls -l /tmp/testfile
lrwxrwxrwx 1 fle users 11 Mar 30 14:04 /tmp/testfile -> /etc/passwd
> expr "$(ls -l testfile)" : '.*[[:space:]]->[[:space:]]*\(.*\)'
/etc/passwd
```

Note that I dropped the final `$` sign in the second argument.

casimir:
Hum, it doesn't work for sub levels.

```
$ ls -l ~/.config/kak
lrwxr-xr-x 1 casimir staff 23 Jul 22  2016 /Users/casimir/.config/kak -> /Users/casimir/dots/kak
$ realpath ~/.config/../.config/kak/autoload/rc/base/autowrap.kak
/Users/casimir/dev/src/github.com/casimir/kakoune/rc/base/autowrap.kak
$ ls -l ~/.config/../.config/kak/autoload/rc/base/autowrap.kak
-rw-r--r-- 1 casimir staff 1972 Jan 10 15:07 /Users/casimir/.config/../.config/kak/autoload/rc/base/autowrap.kak
```

rc/base/ctags.kak
```diff
@@ -7,9 +7,12 @@ decl str-list ctagsfiles 'tags'
 
 def -params ..1 \
     -shell-candidates '
+        realpath() {
+            printf %s\\n "$(cd -P -- "$(dirname -- "$1")" && printf %s\\n "$PWD/${1##*/}")"
```
mawww: why the two printfs ? isnt that an if/then/else disguised ?

casimir: One to print the expanded `$PWD` in the command expansion scope and the other to print the result of `realpath`.
If you have a better way to do it I'll take it, especially the shell expansion in the shell expansion.

casimir:
I took another stab at it but this is still the better way I found. Any more concerns or this can be merge ?

lenormf:
If the last component of the path passed to the function is a symbolic link, it's not resolved. But do we really need to resolve any link at all in this file anyway?

casimir:
Adressed by 510be03910081f320fae600629f23591f4fc207c.

PR #1306, opened 2017-03-31
src: Try to detect possible overflows properly

lenormf:
The previous check tried to catch the overflow after the fact, which
relied on the undefined behaviour of the operation.


lenormf:
Anything blocking here ?

PR #1320, opened 2017-04-09
rc: Export $TMPDIR to new `tmux` processes

lenormf:
Hi,

Read the first commit message for the description of the changeset. Still unsure whether `iterm2.kak` would need similar changes, my guess it probably does.

HTH.


leira:
This change broke on fish shell. On fish shell, the command is supposed to be:

`"env TMPDIR='${TMPDIR}' kak -c ${kak_session} ${kakoune_params}"`

But I'm not sure how to make it work in both bash and fish.

lenormf:
The change you suggest is the right one, I should have thought about it. It's POSIX, so it will work in most shells.

PR #1325, opened 2017-04-16
[WIP] src: Implement an highlighter that hints at long lines

lenormf:
This commit introduces the `show_long_line_hints` highlighter which will
display a one-column wide hint character on the last column of the current
window, on lines that are longer than the window's column count.

The `-hint` parameter controls what character will be displayed (defaults
to a dollar sign), and the `LongLineHint` face dictates how the hint should
be displayed.


lenormf:
Would this be considered for merging if I adapted it to the new highlighter API?

lenormf:
I guess that means no.

PR #1328, opened 2017-04-19
src: Implement a `write!` command

lenormf:
This commit allows "forced" writes to a write-protected file, by
attempting to temporarily grant the current user write permissions on
it. After the buffer has been written, the previous permissions are
restored if the file existed, or set to 0644 otherwise.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -676,8 +676,10 @@ command `q!` has to be used).
  * `e[dit][!] <filename> [<line> [<column>]]`: open buffer on file, go to given
      line and column. If file is already opened, just switch to this file.
      use edit! to force reloading.
- * `w[rite] [<filename>]`: write buffer to <filename> or use it's name if
-     filename is not given.
+ * `w[rite][!] [<filename>]`: write buffer to <filename> or use it's name if
+     filename is not given. If the file is write-protected, its
+     permissions are temporarily changed to allow saving the buffer and
+     restored afterwards when the write! command is used.
  * `w[rite]a[ll]`: write all buffers that are associated to a file.
  * `q[uit][!]`: exit Kakoune, use quit! to force quitting even if there is some
      unsaved buffers remaining.
--- a/doc/manpages/commands.asciidoc
+++ b/doc/manpages/commands.asciidoc
@@ -24,8 +24,11 @@ command *q!* has to be used).
 	open buffer on file, go to given line and column. If file is already
 	opened, just switch to this file. Use edit! to force reloading
 
-*w[rite]* [<filename>]::
-	write buffer to <filename> or use it's name if filename is not given
+*w[rite][!]* [<filename>]::
+	write buffer to <filename> or use it's name if filename is not
+	given. If the file is write-protected, its permissions are temporarily
+	changed to allow saving the buffer and restored afterwards when
+	the write! command is used.
 
 *w[rite]a[ll]*::
 	write all buffers that are associated to a file
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -309,6 +309,7 @@ const CommandDesc force_edit_cmd = {
     edit<true>
 };
 
+template<bool force = false>
 void write_buffer(const ParametersParser& parser, Context& context, const ShellContext&)
 {
     Buffer& buffer = context.buffer();
@@ -326,7 +327,7 @@ void write_buffer(const ParametersParser& parser, Context& context, const ShellC
                     buffer.name() : parse_filename(parser[0]);
 
     context.hooks().run_hook("BufWritePre", filename, context);
-    write_buffer_to_file(buffer, filename);
+    write_buffer_to_file(buffer, filename, force);
     context.hooks().run_hook("BufWritePost", filename, context);
 }
 
@@ -340,6 +341,18 @@ const CommandDesc write_cmd = {
     CommandHelper{},
     filename_completer,
     write_buffer,
+};
+
+const CommandDesc force_write_cmd = {
+    "write!",
+    "w!",
+    "write [filename]: write the current buffer to its file "
+    "or to [filename] if specified, even when the file is write protected",
+    single_optional_param,
+    CommandFlags::None,
+    CommandHelper{},
+    filename_completer,
+    write_buffer<true>,
 };
 
 void write_all_buffers(Context& context)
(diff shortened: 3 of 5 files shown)


src/commands.cc
```diff
@@ -342,6 +342,52 @@ const CommandDesc write_cmd = {
     write_buffer,
 };
 
+void force_write_buffer(const ParametersParser& parser, Context& context, const ShellContext&)
```
mawww: make `write_buffer` a template function with a `<bool force>` parameter, and pass that force flag to the write_buffer_to_file function, add logic in that function to do the mode change.

src/commands.cc
```diff
@@ -342,6 +342,52 @@ const CommandDesc write_cmd = {
...
+            throw runtime_error("couldn't change file permissions");
+    }
+    else
+        st.st_mode = S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH;
+
+    context.hooks().run_hook("BufWritePre", filename, context);
+    write_buffer_to_file(buffer, filename);
+    context.hooks().run_hook("BufWritePost", filename, context);
+
+    if (::chmod(filename.c_str(), st.st_mode) < 0)
```
mawww: This should be done in an `on_scope_end`, as we want to restore the permissions even if something else fails.

lenormf:
The requested changes have been implemented.

About the permissions change itself, I think it's fine as long as we restore the original permissions after writing the file. We also don't try very hard at writing the file (we could try a `chown` beforehand, or try to modify group permissions with `chmod`), this is the least invasive change we can implement that allows lazy users such as myself to not have to use a `sudo` write function, or manually change the permissions of the file in a shell.

README.asciidoc
```diff
@@ -676,8 +676,10 @@ command `q!` has to be used).
  * `e[dit][!] <filename> [<line> [<column>]]`: open buffer on file, go to given
      line and column. If file is already opened, just switch to this file.
      use edit! to force reloading.
- * `w[rite] [<filename>]`: write buffer to <filename> or use it's name if
-     filename is not given.
+ * `w[rite][!] [<filename>]`: write buffer to <filename> or use it's name if
```
occivink: its name

doc/manpages/commands.asciidoc
```diff
@@ -24,8 +24,11 @@ command *q!* has to be used).
 	open buffer on file, go to given line and column. If file is already
 	opened, just switch to this file. Use edit! to force reloading
 
-*w[rite]* [<filename>]::
-	write buffer to <filename> or use it's name if filename is not given
+*w[rite][!]* [<filename>]::
+	write buffer to <filename> or use it's name if filename is not
```
occivink: same

src/commands.cc
```diff
@@ -325,6 +326,23 @@ void write_buffer(const ParametersParser& parser, Context& context, const ShellC
     auto filename = parser.positional_count() == 0 ?
                         buffer.name() : parse_filename(parser[0]);
 
+    struct stat st;
+    auto restore_mode = on_scope_end([&]{
```
mawww: This should not be done before the call to `::stat`, imagine if `::stat` fails, then we will be doing a `::chmod` using an uninitialized `st`.

src/commands.cc
```diff
@@ -325,6 +326,23 @@ void write_buffer(const ParametersParser& parser, Context& context, const ShellC
...
+        if (force and ::chmod(filename.c_str(), st.st_mode) < 0)
+            throw runtime_error("couldn't restore file permissions");
+    });
+    if (force)
```
mawww: Move all that `::chmod` code into `write_buffer_to_file`, which would take an additional `force` argument. I'd like to limit the amount of posix filesystem code that ends up out of `file.cc`

lenormf:
Changes requested implemented, the branch has also been rebased on `master`.

As for the linguistic mistakes, I'll leave them out of this PR to make the commits about the feature itself only, despite the fact that my documentation extends sentences that we would like to fix. I'll most likely proof read the document entirely and make a commit to fix all small mistakes such as the ones mentioned here.

src/file.cc
```diff
@@ -278,9 +278,28 @@ void write_buffer_to_fd(Buffer& buffer, int fd)
     }
 }
 
-void write_buffer_to_file(Buffer& buffer, StringView filename)
+void write_buffer_to_file(Buffer& buffer, StringView filename, bool force)
 {
-    int fd = open(filename.zstr(), O_CREAT | O_WRONLY | O_TRUNC, 0644);
+    int fd;
```
mawww: Why is fd declared here, instead of at point of initialization ?

src/file.cc
```diff
@@ -278,9 +278,28 @@ void write_buffer_to_fd(Buffer& buffer, int fd)
...
+    struct stat st;
+
+    if (force)
+    {
+        if (::stat(filename.zstr(), &st) == 0)
+        {
+            const mode_t mode = st.st_mode | S_IWUSR;
```
mawww: Do we need a variable here ? just pass `st.st_mode | S_IWUSR` to `::chmod`

src/file.cc
```diff
@@ -278,9 +278,28 @@ void write_buffer_to_fd(Buffer& buffer, int fd)
...
+            if (::chmod(filename.zstr(), mode) < 0)
+                throw runtime_error("couldn't change file permissions");
+        }
+        else
+            st.st_mode = S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH;
```
mawww: or maybe just set `force` to `false`, something like:

```
struct stat st;
if (force and ::stat(filename.zstr(), &st) == 0)
{
      if (::chmod(filename.zstr(), mode) < 0)
          throw runtime_error("couldn't change file permissions");
}
else
    force = false;
```

0644 is what open is going to set as mode to by default, so if the file is not existing, `restore_mode` does not need to run.

src/file.cc
```diff
@@ -278,9 +278,28 @@ void write_buffer_to_fd(Buffer& buffer, int fd)
...
+    }
+    auto restore_mode = on_scope_end([&]{
+        if (force and ::chmod(filename.zstr(), st.st_mode) < 0)
+            throw runtime_error("couldn't restore file permissions");
+    });
+
+    fd = open(filename.zstr(), O_CREAT | O_WRONLY | O_TRUNC, 0644);
```
mawww: Maybe call zstr only once, as it might allocate a string if needed (`auto zfilename = filename.zstr();`)

src/file.cc
lenormf: I thought it would look cleaner to move this variable at the top, since it's used several times. Up to you.

src/file.cc
lenormf: It looked cleaner at the time, it's up to you.

lenormf:
Changes requested implemented.

Issue #880, opened 2016-10-18
Question: possibility of implementing GDB plugin.


aytim:
Sorry, didn't find another place to ask question. Want to try this editor for daily work. 
Ideally, I want something like SublimeGDB or vim-lldb.
Do kakoune's design or philosophy have something against plugins like those mentioned?

lenormf:
What features would you expect from such a plugin? This topic was already discussed a bit in the past, and we came to the conclusion that it would take non negligible effort to implement such a compatibility script, and it wouldn't really make things easier to debug an application compared to having the editor set to Kakoune.

aytim:
Thanks for reply! I didn't find any discussions about this topic (I tried). Still interested to read it.

Answering your question about features. Currently I debugging remotely by gdb via ssh terminal window. My main problem (as many others, I suppose) is lack of code navigation functionality in gdb.
Basically, I want some convenience :). For example:
- navigate to function, I want set breakpoint with few keystrokes. Not remembering line and filename and writing to gdb `b foo.cpp:123`
- same as before, when i see variable, for example, `veryLongName`, I don't wan't to remember name and write to gdb `p veryLongName`.
- when inspecting stack, I want to see caller function in editor, so I can use 'Go to definition' and such convenient tools.

At least implementing these cases would make me happier.

I don't mind to implement plugin by myself, for fun. But my main concern is that kakoune isn't designed to work that way.

So the main question is: is kakoune extension capabilities allow implementing debugger plugin?
Or maybe there exists other way to integrate debugging?

mawww:
Hello,

I have been wanting this for some time, and I think it can be achieved. Here is a potential design:

With gdb 7.12, that was just released, there is the `new-ui` command that allows to make gdb add a new ui (of the wanted type, gdb/mi would be what we want here). We could have a program running, connected to that gdb/mi ui, transforming gdb events into kakoune commands (when the gdb frame changes for example), and listiening for commands comming from kakoune (add breakpoint, display value of that variable...).

That program can just be a shell script implemented in an eventual `gdb.kak`, it does not have to be very complicated: parsing gdb/mi outputs and generating kakoune commands, and parsing commands comming from kakoune (on a fifo for example) and generating the gdb/mi command associated, and potentially the kakoune command to run afterwards (getting a variable value for example would run a gdb/mi command, then parse the result and run a kakoune command to display that result in say an info box).

[...]

PR #1333, opened 2017-04-24
[wip] gdb integration script

occivink:
The script works similarly to how mawww described it in [this comment](https://github.com/mawww/kakoune/issues/880#issuecomment-254656973)

This [asciinema](https://asciinema.org/a/116839) gives a quick overview of what the script can do currently. In the second run the execution is controlled in gdb directly.

Known issues:

- the `tail | socat | awk` connection thing is kind of weird, and killing it sometimes also kill the gdb session. I did it that way because that's what I could get to work. The gdb `new-ui` command expects communication to be done through a pty (?) which I don't really understand to be honest.
- ~~pre-existing breakpoints are not known to the script, and the `info breakpoints` command lists them in a different format which is a pain~~
- ~~multi-threading? I haven't looked at all into how that works with the `new-ui` communication, especially with regards to the execution pointer~~ (seems to be fine)

Regarding the first two points I'll probably end up asking for help on the gdb side of things

Some nice additions would be to start the gdb process directly from kakoune into a new terminal, and some way to view/manipulate the callstack

### DIFF
--- a/rc/extra/gdb.kak
+++ b/rc/extra/gdb.kak
@@ -242,6 +242,8 @@ define-command gdb-disable-autojump %{
     set global gdb_jump_client ""
 }
 
+declare-option int backtrace_current_line
+
 define-command gdb-backtrace %{
     try %{
         %sh{
@@ -252,12 +254,55 @@ define-command gdb-backtrace %{
                 echo raise
             fi
         }
-        edit! -fifo "%opt{gdb_dir}/backtrace" *gdb-backtrace*
-        hook -group fifo buffer BufCloseFifo .* %{
-            nop %sh{ rm -f "${kak_opt_gdb_dir}/backtrace" }
-            remove-hooks buffer fifo
+        eval -try-client %opt{toolsclient} %{
+            edit! -fifo "%opt{gdb_dir}/backtrace" *gdb-backtrace*
+            set buffer filetype backtrace
+            set buffer backtrace_current_line 0
+            hook -group fifo buffer BufCloseFifo .* %{
+                nop %sh{ rm -f "${kak_opt_gdb_dir}/backtrace" }
+                remove-hooks buffer fifo
+            }
         }
     }
+}
+
+hook -group backtrace-highlight global WinSetOption filetype=backtrace %{
+    add-highlighter group backtrace
+    add-highlighter -group backtrace regex "^([^\n]*?):(\d+)" 1:cyan 2:green
+    add-highlighter -group backtrace line '%opt{backtrace_current_line}' default+b
+}
+
+hook global WinSetOption filetype=backtrace %{
+    hook buffer -group backtrace-hooks NormalKey <ret> gdb-backtrace-jump
+}
+
+def -hidden gdb-backtrace-jump %{
+    eval -collapse-jumps %{
+        try %{
+            exec -save-regs '' 'xs^([^\n]*?):(\d+)<ret>'
+            set buffer backtrace_current_line %val{cursor_line}
+            eval -try-client %opt{jumpclient} "edit -existing %reg{1} %reg{2}"
+            try %{ focus %opt{jumpclient} }
+        }
+    }
+}
+
+def gdb-backtrace-next %{
+    eval -collapse-jumps -try-client %opt{jumpclient} %{
+        buffer *backtrace*
+        exec "%opt{find_current_line}ggl/^[^:]+:\d+:<ret>"
+        gdb-backtrace-jump
+    }
+    try %{ eval -client %opt{toolsclient} %{ exec %opt{find_current_line}g } }
+}
+
+def gdb-backtrace-prev %{
+    eval -collapse-jumps -try-client %opt{jumpclient} %{
+        buffer *backtrace*
+        exec "%opt{find_current_line}g<a-/>^[^:]+:\d+:<ret>"
+        gdb-backtrace-jump
+    }
+    try %{ eval -client %opt{toolsclient} %{ exec %opt{find_current_line}g } }
 }
 
 # implementation details
--- a/rc/extra/gdb.kak
+++ b/rc/extra/gdb.kak
@@ -80,7 +80,10 @@ define-command -hidden gdb-session-connect-internal %{
                 line = get(frame, "line=\"", "[0-9]+", "\"")
                 return line " \"" file "\""
             }
-            BEGIN { connected = 0 }
+            BEGIN {
+                connected = 0
+                printing = 0
+            }
             // {
                 if (!connected) {
                     connected = 1
@@ -131,7 +134,7 @@ define-command -hidden gdb-session-connect-internal %{
             /\^done/ {
                 if (printing) {
                     # trim trailing \n
-                    print_value = substr(print_value, 0, length(print_value) - 3)
+                    print_value = substr(print_value, 0, match(print_value, "(\n|\\\\n)*$") - 1)
                     # QUOTE => \\QUOTE
                     gsub("'\''", "\\\\'\''", print_value)
                     # eval -client $client QUOTE info  -- \QUOTE $string \QUOTE QUOTE
--- a/rc/extra/gdb.kak
+++ b/rc/extra/gdb.kak
@@ -1,0 +1,320 @@
+# script summary:
+# a long running shell process connects to an existing gdb session and handles input/output
+# input (common gdb commands) to the gdb session is done by writing to a fifo
+# output is converted into corresponding kak commands to update the presented information
+
+declare-option -hidden str gdb_dir
+
+# contains all known breakpoints in this format:
+# id|enabled|line|file:id|enabled|line|file|:...
+declare-option -hidden str-list gdb_breakpoints_info
+# if execution is currently stopped, contains the location in this format:
+# line|file
+declare-option -hidden str-list gdb_location_info
+# note that these variables may reference locations that are not in currently opened buffers
+
+# corresponding flags generated from the previous variables
+# these are only set on buffer scope
+declare-option -hidden line-flags gdb_breakpoints_flags
+declare-option -hidden line-flags gdb_location_flag
+
+set-face GdbBreakpoint red,default
+set-face GdbLocation blue,default
+
+add-highlighter -group / group -passes move gdb
+add-highlighter -group /gdb flag_lines GdbLocation gdb_location_flag
+add-highlighter -group /gdb flag_lines GdbBreakpoint gdb_breakpoints_flags
+
+hook global WinCreate .* %{
+    add-highlighter ref -passes move gdb
+}
+
+define-command -params .. -file-completion gdb-session-new %{
+    gdb-session-connect
+    %sh{
+        # can't connect until socat has created the pty thing
+        while [ ! -e "${kak_opt_gdb_dir}/pty" ]; do
+            sleep 1
+        done
+        {
+            setsid -w $kak_opt_termcmd " \
+              gdb $@ --init-eval-command=\"new-ui mi3 ${kak_opt_gdb_dir}/pty\""
+            printf gdb-session-stop\\n | kak -p "$kak_session"
+        } 2>/dev/null >/dev/null &
+    }
+}
+
+define-command gdb-session-connect %{
+    gdb-session-stop
+    %sh{
+        tmpdir=$(mktemp --tmpdir -d gdb_kak_XXX)
+        mkfifo "${tmpdir}/pipe"
+        {
+            tail -n +1 -f "${tmpdir}/pipe" | socat "pty,link=${tmpdir}/pty" STDIO,nonblock=1 | awk '
+            function send(what) {
+                cmd = "kak -p '"$kak_session"'"
+                print(what) | cmd
+                close(cmd)
+            }
+            function get(input, prefix, pattern, suffix) {
+                s = match(input, prefix pattern suffix)
+                return substr(input, s + length(prefix), RLENGTH - length(prefix) - length(suffix))
+            }
+            function breakpoint_info(breakpoint) {
+                id = get(breakpoint, "number=\"", "[0-9]+", "\"")
+                enabled = get(breakpoint, "enabled=\"", "[yn]", "\"")
+                line = get(breakpoint, "line=\"", "[0-9]+", "\"")
+                file = get(breakpoint, "fullname=\"", "[^\"]*", "\"")
+                return id " " enabled " " line " \"" file "\""
+            }
+            function frame_info(frame) {
+                file = get(frame, "fullname=\"", "[^\"]*", "\"")
+                line = get(frame, "line=\"", "[0-9]+", "\"")
+                return line " \"" file "\""
+            }
+            BEGIN { connected = 0 }
+            // {
+                if (!connected) {
+                    connected = 1
+                    print("-break-list") >> "'"$tmpdir/pipe"'"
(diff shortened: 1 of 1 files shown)


rc/extra/gdb.kak
```diff
@@ -0,0 +1,282 @@
...
+define-command gdb-set-breakpoint   %{ gdb-cmd "break %val{buffile}:%val{cursor_line}" }
+define-command gdb-clear-breakpoint %{ gdb-cmd "clear %val{buffile}:%val{cursor_line}" }
+define-command gdb-toggle-breakpoint %{
+    try %{
+        %sh{
+            printf %s\\n "$kak_opt_gdb_breakpoints_info" | tr ':' '\n' |
+            while read -r current; do
+                buffer=$(printf %s "$current" | cut -d \| -f 4-)
+                line=$(printf %s "$current" | cut -d \| -f 3)
+                if [ "$buffer" = "$kak_buffile" -a "$line" = "$kak_cursor_line" ]; then
```
lenormf: It's better to use `if [ … ] && [ … ]; then` rather than the `-a` flag, [according to shellcheck](https://github.com/koalaman/shellcheck/wiki/SC2166).

rc/extra/gdb.kak
```diff
@@ -0,0 +1,282 @@
...
+                    printf %s\\n raise
```
lenormf: I think doing `echo gdb-clear-breakpoint` and getting rid of the following `catch` is cleaner.

rc/extra/gdb.kak
```diff
@@ -0,0 +1,282 @@
...
+    set-option global gdb_location_info "%arg{1}|%arg{2}"
+    gdb-refresh-location-flag
+}
+
+define-command -hidden gdb-refresh-location-flag %{
+    try %{
+        %sh{
+            if [ -n "$kak_opt_gdb_location_info" ]; then
+                line=$(printf %s "$kak_opt_gdb_location_info" | cut -d \| -f 1)
+                buffer=$(printf %s "$kak_opt_gdb_location_info" | cut -d \| -f 2-)
```
lenormf: You could save a subprocess here and use shell expansions:

```
line="${kak_opt_gdb_location_info%%|*}"
buffer="${kak_opt_gdb_location_info#*|}"
```

rc/extra/gdb.kak
```diff
@@ -0,0 +1,282 @@
...
+                printf "set-option -add \"buffer=%s\" gdb_location_flag \"%s|->\"\n" "$buffer" "$line"
+            fi
+        }
+    }
+}
+
+define-command -hidden gdb-clear-location %{
+    %sh{
+        if [ -n "$kak_opt_gdb_location_info" ]; then
+            buffer=$(printf %s "$kak_opt_gdb_location_info" | cut -d \| -f 2-)
```
lenormf: Shell expansion here as well.

rc/extra/gdb.kak
```diff
@@ -0,0 +1,282 @@
...
+define-command -hidden -params 4 gdb-handle-breakpoint-created %{
+    set-option -add global gdb_breakpoints_info "%arg{1}|%arg{2}|%arg{3}|%arg{4}"
+    gdb-refresh-breakpoints-flags %arg{4}
+}
+
+define-command -hidden -params 1 gdb-handle-breakpoint-deleted %{
+    %sh{
+        printf "set-option global gdb_breakpoints_info \"\"\n"
+        while read -r current; do
+            id=$(printf %s "$current" | cut -d \| -f 1)
```
lenormf: Shell expansion here as well.

rc/extra/gdb.kak
```diff
@@ -0,0 +1,282 @@
...
+            if [ "$id" != "$1" ]; then
+                printf "set-option -add global gdb_breakpoints_info \"%s\"\n" "$current"
+            else
+                buffer=$(printf %s "$current" | cut -d \| -f 4-)
```
lenormf: Shell expansion here as well.

rc/extra/gdb.kak
```diff
@@ -0,0 +1,282 @@
...
+            fi
+        done <<-EOF
+			$(printf %s "$kak_opt_gdb_breakpoints_info" | tr ':' '\n')
+		EOF
```
lenormf: Is there something amiss here? There's unexpected indentation and an here-document delimiter.

rc/extra/gdb.kak
```diff
@@ -0,0 +1,282 @@
...
+        printf "gdb-refresh-breakpoints-flags \"%s\"\n" "$buffer"
+    }
+}
+
+define-command -hidden -params 4 gdb-handle-breakpoint-modified %{
+    %sh{
+        printf "set-option global gdb_breakpoints_info \"\"\n"
+        printf %s\\n "$kak_opt_gdb_breakpoints_info" | tr ':' '\n' |
+        while read -r current; do
+            id=$(printf %s "$current" | cut -d \| -f 1)
```
lenormf: Shell expansion here as well.

rc/extra/gdb.kak
```diff
@@ -0,0 +1,282 @@
...
+            if [ "$id" != "$1" ]; then
+                printf "set-option -add global gdb_breakpoints_info \"%s\"\n" "$current"
+            else
+                printf "set-option -add global gdb_breakpoints_info \"%s|%s|%s|%s\"\n" "$1" "$2" "$3" "$4"
+            fi
+        done
```
lenormf: Is another process in charge of writing to this loop's `stdin`? Or is something missing here?

rc/extra/gdb.kak
occivink: Since the while loop is in a subshell I wouldn't be able to `exit` without also calling `set-breakpoint`. I don't think a heredoc solution like I did in `gdb-handle-breakpoint-deleted` would be cleaner, so unless you have a simpler solution to this I'll leave it as is.

rc/extra/gdb.kak
```diff
@@ -0,0 +1,283 @@
...
+    hook -group gdb global KakEnd .* %{
+        gdb-stop
+    }
+}
+
+define-command gdb-stop %{
+    try %{
+        %sh{
+            if [ -n "$kak_opt_gdb_dir" ]; then
+                pkill -P $(cat "${kak_opt_gdb_dir}/pid")
```
lenormf: This command isn't provided on some systems, is there anyway to emulate its effects using a plain `kill`?

mawww:
After the `wrap` highlighter change, you need to tweak the highlighter group you create with `-passes move` as you put line_flags highlighters in it, which are group highlighters.

Also, I think using Unicode characters, like `●` for breakpoints, and say `➡` for current position would be nicer looking.

Finally, maybe having a single line_flags option and highlighter would be simpler than two options and two highlighters (although I dont mind much if it seems better to you to have both).

occivink:
Ok for the nice glyphs and the adpation to the add-highlighter changes. I somehow messed up the commits so I ended up squashing the commits. 
I'll try to think about using a single highlighter and see if it's better in terms of readability/maintainability.

I've reported the bug I mentioned on gdb's side: https://sourceware.org/bugzilla/show_bug.cgi?id=21477

Remaining todo that I see:
- [ ] Tweak the printing code, I noticed some wrongly handled data types
- [ ] tmux support for new gdb session
- [ ] Ideally call `gdb-stop` when the gdb connection stops for whatever reason. Unfortunately there is no MI notification for gdb exiting, surprisingly.
- [ ] Document options and variables

lenormf:
Shouldn't there be a way for users to pick the symbols used by the scripts? I'm guessing not everyone uses a terminal that can render fancy glyphs, so maybe they should be stored in modifiable options, to allow changing them to ascii characters (those options could be hidden, in order not to spoil the namespace)?

mawww:
I dont think thats important, Kakoune embraces unicode, modern terminal should be able to display unicode symbols (this is not emojies, just basic symbols). The linting support already uses a block symbol to tag the lines that have an error or warning for example.

occivink:
I've added an "autojump" functionality to go directly where the execution stopped, but it doesn't fully work: the buffer changes appropriately, but it's not jumping to the line like calling `gdb-jump-to-location` directly does. 
Maybe I should use this script to see why...

rc/extra/gdb.kak
```diff
@@ -228,12 +238,29 @@ decl -hidden str gdb_jump_client
...
+            if [ -n "$kak_opt_gdb_dir" ]; then
+                mkfifo "$kak_opt_gdb_dir"/backtrace
+                echo "-stack-list-frames" > "$kak_opt_gdb_dir"/pipe
+            else
+                raise
+            fi
+        }
+        edit! -fifo "%opt{gdb_dir}/backtrace" *gdb-backtrace*
+        hook -group fifo buffer BufCloseFifo .* %{
+            %sh{ rm "${kak_opt_gdb_dir}/backtrace" }
```
lenormf: May I suggest a more idiotmatic `nop %sh{ rm -f "${kak_opt_gdb_dir}/backtrace" }` here? It's not a big deal if the file isn't removed, and we don't want to take any eventual (although unlikely) output from the `rm` command.

occivink: Good point

rc/extra/gdb.kak
```diff
@@ -174,7 +174,7 @@ define-command gdb-session-stop %{
                 rm "${kak_opt_gdb_dir}/pid" "${kak_opt_gdb_dir}/pipe"
                 rmdir "$kak_opt_gdb_dir"
             else
-                printf %s\\n raise
+                echo raise
```
lenormf: I think that since the code after this scope is fully static, you could just print it with an `echo` in the other conditional branch and not rely on this trick (maybe one day there will be a valid `raise` keyword`, and this code will be broken).

```
if [ -n "$kak_opt_gdb_dir" ]; then
  kill $(ps -o pid= --ppid $(cat "${kak_opt_gdb_dir}/pid"))
  rm "${kak_opt_gdb_dir}/pid" "${kak_opt_gdb_dir}/pipe"
  rmdir "$kak_opt_gdb_dir"

  echo '
         set-option global gdb_dir ""

         gdb-clear-location
         gdb-clear-breakpoints

         remove-hooks global gdb
  '
fi
```

Unless you rely on an exception being raised here, in the calling scope of the command?

rc/extra/gdb.kak
```diff
@@ -0,0 +1,413 @@
...
+define-command gdb-clear-breakpoint %{ gdb-cmd "clear %val{buffile}:%val{cursor_line}" }
+define-command gdb-toggle-breakpoint %{
+    try %{
+        %sh{
+            printf %s\\n "$kak_opt_gdb_breakpoints_info" | tr ':' '\n' |
+            while read -r current; do
+                buffer="${current#*|*|*|}"
+                line=$(printf %s "$current" | cut -d \| -f 3)
+                if [ "$buffer" = "$kak_buffile" ] && [ "$line" = "$kak_cursor_line" ]; then
+                    printf %s\\n raise
```
lenormf: Similarly to the previous command, you could potentially simplify the code here by clearing the breakpoints immediately on error (`echo gdb-clear-breakpoint`), and setting one after the `while` loop (`echo gdb-set-breakpoint`). The entire command would be a shell scope, and it would get rid of the `try`.

rc/extra/gdb.kak
occivink: Okay, I did it this way because in general I prefer writing kakoune commands not in a shell scope if I can avoid it, but I don't mind changing it.

rc/extra/gdb.kak
occivink: However I'll leave this one as is because I think the logic would more complex if I were to implement it entirely in the shell scope (need to keep track in a variable if there is a breakpoint here, get rid of the subshell).

lenormf: Wouldn't the following work?

```
define-command gdb-session-stop %{ %sh{
    if [ -n "$kak_opt_gdb_dir" ]; then
        kill $(ps -o pid= --ppid $(cat "${kak_opt_gdb_dir}/pid"))
        rm "${kak_opt_gdb_dir}/pid" "${kak_opt_gdb_dir}/pipe"
        rmdir "$kak_opt_gdb_dir"

        echo '
            set-option global gdb_dir ""

            gdb-clear-location
            gdb-clear-breakpoints

            remove-hooks global gdb
        '
    fi
} }
```

I think it looks clearer than the actual version, also it doesn't try to control the execution flow artificially with code that is intended to fail.

lenormf: Wouldn't the following work?

```
define-command gdb-toggle-breakpoint %{ %sh{
    printf %s\\n "$kak_opt_gdb_breakpoints_info" | tr ':' '\n' |
    while read -r current; do
        buffer="${current#*|*|*|}"
        line=$(printf %s "$current" | cut -d \| -f 3)
        if [ "$buffer" = "$kak_buffile" ] && [ "$line" = "$kak_cursor_line" ]; then
            echo gdb-clear-breakpoint
            exit
        fi
    done
    gdb-set-breakpoint
} }
```

I think it looks clearer than the actual version, also it doesn't try to control the execution flow artificially with code that is intended to fail.

occivink: Ah yeah I guess you're right

occivink: Ah no it doesn't work actually, because `exit` only applies to the subshell, so we end up setting the breakpoint again. Guess I'll go with a heredoc solution

lenormf: The `echo gdb-set-breakpoint` is not reached when `exit` is called, I think my code works as intended?

mawww:
Where are we at with this ?

I reallly like this script, however I wonder if it should be provided with Kakoune upstream, or in a separate repository.

On one hand, I'd like to have it available easily, and having it upstream means we can fix it directly when doing breaking changes. On the other hand, its a complex piece of script that could have a life on his own, and its not really requiered out of the box for a nice experience (unlike highlighting for various, including obscure, languages, as that really improves the experience not to have to get a script whenever you open a file from a new language).

What do you think about that ?

alexherbo2:
I have the same opinion.

How about Git sub-modules?

occivink:
Right now I'm pretty happy with the script in terms of features. I've used it with success on C and C++ projects , although not with other languages supported by gdb. 
The biggest issue that I see right now is the communication with gdb. When I've talked to gdb people it seemed like they would be open for `new-ui` to accept a socket in addition to a pty, which would solve a couple of issues. I had a quick look into how that could be implemented, but not being very proficient in C I didn't go very far.
Aside from features, there's probably things to fix in terms of user interaction but I don't have the "recul" (?) needed. If somebody feels like testing script end-to-end, I'd appreciate.

Since, as you mention it, it's a complex script, the documentation concerns start to rear their head. At least with a separate repo we would be able to propose a dedicated README. 

I don't mind keeping a separate repository for the script like @Delapouite is doing. Submodule would also be acceptable if we want to separate the script from the core, but still provide it out of the box.

Delapouite:
As a side note, I contacted github to know if the [kakoune account](https://github.com/kakoune) was available to turn it into an organisation. Here is the answer I received on the 1st of June:

```
Hi Bruno,

I took a look at the account in question, and it doesn't appear that it is dormant (not all activity on GitHub is public). Sorry to be the bearer of bad news with this one.

Thanks,
Jacqui
```

doppioandante:
I think that this is important enough that we should merge it upstream, although I'm still fine with an external repo.

occivink:
I'm closing this PR in favor of a dedicated repository
https://github.com/occivink/kakoune-gdb

PR #1342, opened 2017-04-30
rc: Replace `setsid` calls with POSIX shell scripts

lenormf:
Better rely on the shell rather than on an external binary which might not be installed.

### DIFF
--- a/rc/base/x11.kak
+++ b/rc/base/x11.kak
@@ -28,7 +28,9 @@ The optional arguments will be passed as arguments to the new client} \
            exit
         fi
         if [ $# -ne 0 ]; then kakoune_params="-e '$@'"; fi
-        setsid ${kak_opt_termcmd} "kak -c ${kak_session} ${kakoune_params}" < /dev/null > /dev/null 2>&1 &
+        {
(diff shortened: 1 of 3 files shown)


mawww:
Are you sure this pattern will create the processes in a new process group ? the point of that `setsid` is to make sure the terminal emulator processes are not in the same process group as the Kakoune server, so that when hitting `<c-c>` in Kakoune, they dont get killed.

lenormf:
New terminals are spawned in those files, which means their parent process will be the initd (`pid=1`). I verified that using `ps axo ppid,pid,cmd`, I think we should be fine?

mawww:
Nope, that is not sufficient, here, with rxvt-unicode, if I apply your changes then run kak, `:new` then `<c-c>` the second client dies. That was the reason for the setsid call, we need to detach those terminal processes from the kakoune server process group, and putting them in a new posix session does just that.

lenormf:
`<c-c>` does nothing here, I must not have any job control in my newly created terminal?

mawww:
Maybe the terminal emulator you are using does some session management itself. Whats your terminal emulator ?

Issue #538, opened 2015-12-22
Describe briefly the sessions listed with the -l flag


lenormf:
Hi,

When listing sessions using `kak -l`, all we get back from the server is numbers, it would be nice to add some kind of info that would serve as a hint-name of the current buffer ? something else ?

Thanks.

mawww:
We could have that as a script using kak -l, something like

```
for session in $(kak -l); do
   echo  'nop %sh{ echo $kak_session : $kak_buflist >> outfile }' |  kak -p $session
done
cat outfile
```

having that built into kak -l would not be nice though, as it means kak -l now needs to connect to the sessions (instead of just listing the sockets)

lenormf:
It'd be best if there was a builtin way to have more info about sessions, without having to drag a script around everywhere `kak` has to be deployed.

Another option flag, maybe (`-ll`) ?

tony:
> having that built into kak -l would not be nice though, as it means kak -l now needs to connect to the sessions (instead of just listing the sockets)

i view tmux as a pretty conservative approach to handling these things, tmux connects to a socket via a client just to run simple commands to list session/window/pane info, heh. it prints the info, then closes

it uses something sort of similar to the kakoune variables in that script, called [formats](http://man7.org/linux/man-pages/man1/tmux.1.html#FORMATS).

lenormf:
I implemented a script in #2580 that describes sessions, but as you can see, it's not trivial to get right. I still think a `-ll` flag would make sense, or we could revive my PR that allows the server to write back on `stdout`.

PR #1344, opened 2017-04-30
src: Add a `server_pid` default environment variable

lenormf:
Hi,

Since I don't use process IDs as session names, I'd like to have a simple way of gettting the process ID of the server.

Also this feature allows making use of a feature of LSP, namely allowing the language server to have the same lifespan as the process that spawns it[1].

[1] https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#initialize-request

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -157,6 +157,10 @@ void register_env_vars()
             "window_height", false,
             [](StringView name, const Context& context) -> String
             { return to_string(context.window().dimensions().line); }
+        }, {
+            "server_pid", false,
+            [](StringView name, const Context& context) -> String
+            { return to_string(Server::instance().pid()); }
         }
     };
 
@@ -544,7 +548,8 @@ int run_server(StringView session,
 
     GlobalScope::instance().options().get_local_option("readonly").set<bool>(flags & ServerFlags::ReadOnly);
 
-    Server server{session.empty() ? to_string(getpid()) : session.str()};
+    const pid_t pid_server = getpid();
+    Server server{pid_server, session.empty() ? to_string(pid_server) : session.str()};
 
     bool startup_error = false;
     if (not (flags & ServerFlags::IgnoreKakrc)) try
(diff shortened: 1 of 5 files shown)


src/main.cc
```diff
@@ -157,6 +157,10 @@ void register_env_vars()
             "window_height", false,
             [](StringView name, const Context& context) -> String
             { return to_string(context.window().dimensions().line); }
+        }, {
+            "server_pid", false,
+            [](StringView name, const Context& context) -> String
+            { return to_string(Server::instance().pid()); }
```
mawww: Or we could just call `getpid` here and not bother storing that in the `Server`

lenormf: It would issue a syscall everytime the variable is expanded, since it's not going to change I think storing it is cleaner.

mawww:
This is going to be `$PPID` in any `%sh{...}` block, so we dont need to expose it.

lenormf:
You still need to spawn a shell, does it not make sense to have a dedicated option accessible with a simple expand?

mawww:
I find it highly unlikely that you need the server pid, but are not forking a shell to do something with it at some point.

lenormf:
Killing the server from outside a renamed session.

mawww:
Not sure i get you, what command would you use ?

lenormf:
I wanted to be able to run `kak -d -s my-session`, and list information about sessions (in a script, looping over `kak -l`), such as server PID, number of open buffers etc from the CLI. To do that, I wanted to introduce an `-F` flag that would act like `tmux` formats (mentioned [in this issue](https://github.com/mawww/kakoune/issues/538#issuecomment-223156458)), and make the server expand a string passed to it (`echo '%val{server_pid}' | kak -s my-session -F`). That way I could easily kill a session from the command line, and/or make a script that lists information as requested in the issue linked previously.

Arguably this `server_pid` variable isn't really useful in itself, but I thought that if I was to implement an `-F` flag, I wouldn't want the server to have to spawn a shell or issue a syscall to get such basic info.

lenormf:
Actually my implementation would use `-P`, which would be like `-p` but would write the expanded string passed on `stdin` on `stdout`.

mawww:
I am still unconvinced, if the point of having the server_pid is to facilitate killing a session, might I suggest just using `echo 'kill!' | kak -p <session>` ? the pid of the server is really just an implementation detail, I am not sure how relevant it is to make it available.

PR #1349, opened 2017-05-05
doc: Document guidelines about writing kak scripts

lenormf:


Delapouite:
Thanks for this @lenormf . Your docs are always great and detailed.

lenormf:
Removed the paragraph that mentions underscore prefixes of hidden options, and make options documented with a docstring as well as commands.

PR #1350, opened 2017-05-05
Elixir highlighter

zambal:


rc/extra/elixir.kak
```diff
@@ -0,0 +1,98 @@
...
+hook -group elixir-highlight global WinSetOption filetype=elixir %{ add-highlighter ref elixir }
+
+hook global WinSetOption filetype=elixir %{
+    hook window InsertEnd  .* -group elixir-hooks  elixir-filter-around-selections
+    hook window InsertChar \n -group elixir-indent elixir-indent-on-new-line
+}
+
+hook global WinSetOption filetype=elixir %{
+    set buffer tabstop 2
+    set buffer indentwidth 2
```
lenormf: Are these indentation settings mandatory in elixir code? If not it's better to let users set those as they see fit.

zambal: You are right, although I think most people work with these settings, there are no standard rules as for example in go and it's definitely not mandatory.

I'll add a commit without these indentation settings.

zambal: Commit is done and I also cleaned up the file a bit and tweaked a few colors. If you see no other issues, I think this is good enough for basic Elixir support.

lenormf:
Should be fine, could you squash your intermediary commits into a single one please? So that there's basically one commit with a waiver and one with the Elixir support script?

zambal:
Sure, I just squashed the commits.

rc/extra/elixir.kak
```diff
@@ -0,0 +1,85 @@
...
+# Faces
+# -----
+
+face ElixirAtom yellow
+face ElixirUnderscore white+d
+face ElixirAttribute green+d
+face ElixirSpecialKeyword blue+b
+face ElixirEnv red+b
+face ElixirOperator rgb:a8a8a8
+face ElixirInterpolation rgb:00d7ff
```
mawww: Using custom faces here means colorschemes wont be redefining these. I think it will be better to use the generic ones (see for example `colors/default.kak`)

If you think we really need to introduce new generic ones, we can discuss that either here or on IRC, but I dont think adding filetype specific faces is the way to go.

zambal: It felt a bit dirty to add them indeed, but I wanted to have a bit more colors to select from, but the fact different colorschemes are now effectively not supported for the Elixir highlighter is indeed not good.

I'll commit a version without the extra faces.

zambal: Would you like me to squash this commit into the previous commit too?

PR #1355, opened 2017-05-09
Fix #1353

### LINKED ISSUE #1353: `autoload` in kakrc fails on ksh-like `/bin/sh` providers
When the system's `/bin/sh` provider is a ksh-like (zsh, mksh, so on), `autoload` is a shell builtin. Therefore, setting it in the distributed kakrc (predictably) doesn't work.

To reproduce:
- Install mksh (faster/simpler example).
- Back up your /bin/sh and symlink mksh into its place (`cp /bin/sh /bin/oldsh && ln -s /bin/mksh /bin/sh`)
- Ensure that kakrc.kak is in your autoload directory.
- `kak system_kakrc`

Expected: syntax highlighting is enabled
Actual: syntax highlighting is not enabled

You can then run something along the lines of...
```kak
source autoload/kakrc.kak
set buffer filetype kak
```
to have it reappear temporarily.

If you rename `autoload()` to something else (e.g `kak_autoload()` and replace the calls with that name) everything functions as expected as well - something that can be used as the solution.

I'll make a PR (and ref. this) when I get home.

5paceToast:
See bug #1353 and commits for more details.
tl;dr - fix instances where autoloading would fail due to a shell having `autoload` as a reserved keyword.

### DIFF
--- a/share/kak/kakrc
+++ b/share/kak/kakrc
@@ -23,7 +23,7 @@ def -params 1 -docstring "colorscheme <name>: enable named colorscheme" \
 }}
 
 %sh{
-    autoload() {
+    autoload_directory() {
         dir=$1
         for rcfile in ${dir}/*.kak; do
             if [ -f "$rcfile" ]; then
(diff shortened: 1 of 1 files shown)


5paceToast:
Commit amended - `kak_autoload` -> `autoload_directory`. (history directly modified since this repo appears to prefer single commits as in #1350)

mawww:
Thanks for that fix :)

PR #1365, opened 2017-05-15
kakrc: Simplify and optimize the autoload function

lenormf:
If we don't care about the order of loading of the scripts in the autoload directory, this should save us some time (spawning process is expensive).


lenormf:
I pushed this commit in #1379.

PR #1366, opened 2017-05-15
Add some missing languages to comment.kak

danr:


rc/core/comment.kak
```diff
@@ -50,6 +51,10 @@ hook global BufSetOption filetype=html %{
     set buffer comment_block '<!--:-->'
 }
 
+hook global BufSetOption filetype=kak %{
+    set buffer comment_line '#'
```
lenormf: This is the default value already, the hunk is not needed.

rc/core/comment.kak
```diff
@@ -78,7 +83,8 @@ hook global BufSetOption filetype=(pug|rust) %{
 }
 
 hook global BufSetOption filetype=python %{
-    set buffer comment_block '\'\'\':\'\'\''
+    set buffer comment_line '#'
+    set buffer comment_block "''':'''"
```
lenormf: The `comment_line` is already set to `#` by default, so we don't need to specify it here once again.

danr:
Thank you for the comments!

PR #1368, opened 2017-05-15
Improve kakrc highlighter with more static words

danr:


rc/core/kakrc.kak
```diff
@@ -15,25 +15,38 @@ add-highlighter -group / regions -default code kakrc \
     comment (^|\h)\K\# $ '' \
     double_string %{(^|\h)"} %{(?<!\\)(\\\\)*"} '' \
     single_string %{(^|\h)'} %{(?<!\\)(\\\\)*'} '' \
-    shell '%sh\{' '\}' '\{'
```
lenormf: Considering how many combinations are possible, maybe it would be worth trying to generate those lines in a shell scope? Or do we consider that those are the most commonly used delimiters and we don't want to be exhaustive anyway?

danr: Are you thinking of `%sh'`, `%sh"` etc?

lenormf: Any punctuation sign can be used for scopes, including `!`, `~`, `#` etc. I've seen those being used in some scripts already, so do we aim at providing exhaustive support for all types of scopes here? If yes (I don't know if it's worth it or not), I think these lines could be generated for a shell `for` loop.

PR #1369, opened 2017-05-15
Remove ad-hoc rules in kakrc highlighter

danr:
These are less useful with more static words, and they are
woefully incomplete: no support -docstring for map, set
uses the variable face, but there is no corresponding
highlighter for decl or %opt{..}.

### DIFF
--- a/rc/core/kakrc.kak
+++ b/rc/core/kakrc.kak
@@ -38,9 +38,6 @@ add-highlighter -group / regions -default code kakrc \
 }
 
 add-highlighter -group /kakrc/code regex \brgb:[0-9a-fA-F]{6}\b 0:value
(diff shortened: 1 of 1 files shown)


danr:
The static words PR is #1368

lenormf:
Those are supposed to provide with a small argument checker that highlights unexpected keywords with the `Error` face, why not fix it instead of removing it?

danr:
I know this, but supporting docstrings for map would be a big and presumably messy undertaking.

I understand it's a controversial PR as it removes features, so just close it if you disapprove of it :)

lenormf:
I guess the proper way would be to implement a linter for `kak` source and let the `:lint` command take care of all this. I'm not particularly against the removal, but maybe it could be improved to handled simple cases such as calls to `set`.

PR #1371, opened 2017-05-15
Don't execute hooks when indenting Makefiles with tab

danr:


mawww:
Hum, I am wondering about this one, I suspect this is to avoid the tab to space hook, but it should arguably not be added on Makefiles. My problem here is that there is no reason to disable hooks here and not in other places where we do inserts. Maybe the solution would be to rather use `<a-gt>` instead of `i<tab>` to insert an indent tab.

mawww:
ping

PR #1374, opened 2017-05-16
rc: Document non-hidden options with `-docstring`

lenormf:
Hi,

This is a preliminary PR that adds a `-docstring` flag to all non-hidden options declared in the `rc` directory. 

There's no way that I know of to interactively read those documentation strings as of today, but I've been working on another PR to include documentation generated directly from the scripts, and it needs as many documented completable options/commands as possible, so there you go.

### DIFF
--- a/rc/core/c-family.kak
+++ b/rc/core/c-family.kak
@@ -299,7 +299,7 @@ def -hidden c-family-insert-include-guards %{
 
 hook global BufNewFile .*\.(h|hh|hpp|hxx|H) c-family-insert-include-guards
 
-decl -docstring "colon separated list of path in which header files will be looked for" \
+decl -docstring "semi-colon separated list of path in which header files will be looked for" \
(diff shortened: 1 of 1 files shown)


lenormf:
Rebased on 55d1d1020d215730737554045537cf3a8877c46a.

mawww:
I think I merged a previous version of that, could you rebase it so that the new changes can be merged ?

lenormf:
It looks like this PR was already merged (b9cdccd53aa9c66c619efb0c38b2c8064f70f79a), but after rebasing it turned out that one change wasn't taken into account.

The branch was rebased on 6906e6924bf97b5974a14db3b49cfa21dbc64c1a, and now only changes a single word.

lenormf:
Dang this change is wrong, string lists *are* separated by a colon, and not a semi-colon as was the case in this variable before :/

My bad, I'll make another PR to revert this change.

PR #1380, opened 2017-05-17
Add modal info option.

### LINKED ISSUE #1375: Expose modal placement for info boxes
Feature request for `info -placement modal` (just like the popup for reloading file on external file change.)

Thanks!

ekie:
Closes #1375


src/client.cc
```diff
@@ -382,11 +382,8 @@ void Client::info_show(String title, String content, BufferCoord anchor, InfoSty
     m_ui_pending &= ~InfoHide;
 }
 
-void Client::info_hide(bool even_modal)
```
ekie: is the even_modal logic needed ?

lenormf:
There are cases where we don't want a modal to be cleared because of another info menu, that's why the changes that your commit undoes were introduced in b3674a2f.

ekie:
i could not trigger the info_hide unintentionally when the buffer reload popup was visible and i thought this code was obsolete

mawww:
I think `-placement center` is wrong here, as `-placement` is intended to be relative to the anchor of the info box. Whats the use case for exposing the modal info mode ? Maybe we need proper modal support here (with a way to explicitely hide the modal info).

ekie:
i agree that placement is not the right option for this. maybe @danr can jump in here for his usecase. i extended this implementation to accept a face in order to stand out visually, introducing a warning face for more contrast that would be useful for me.

danr:
I'm using the info in the lower-right corner to show my open buffers with the current buffer highlighted. Since I want my attention to be on this popup when changing buffer I thought it would be nice if it could be in the center rather than in the corner.

PR #1389, opened 2017-05-20
Changed wording of completers description

ManDay:


occivink:
You should also add a copyright waiver commit as described in [CONTRIBUTING](https://github.com/mawww/kakoune/blob/master/CONTRIBUTING).

PR #1391, opened 2017-05-21
Solarized toggle background

stevenyvr987:
The solarized color theme can be easily changed between a light and dark background.

- Therefore, I've added a function in file colors/solarized.kak to toggle between light and dark.
- I've also added an option within the theme to select the background to be initially light.
- Also, I've modified the indentation of some lines to align rgb values.

### DIFF
--- a/colors/solarized.kak
+++ b/colors/solarized.kak
@@ -1,24 +1,70 @@
 # solarized theme
 
+# Solarized colorscheme will start in light or dark background,
+# depending on the following option.
+# 
+decl -hidden str solarized_initially light
+
+# Map the following toggle function to an alias or a key sequence
+# for easy toggling between light and dark, eg,
+# map global user t :solarized-toggle-background<ret>
+# 
+def -docstring 'toggle solarized background between light and dark' -allow-override \
+solarized-toggle-background %{
+    %sh{
+        set_bg() printf %s\\n "set global solarized_background $1"
+        case "$kak_opt_solarized_background" in
+        ( dark) set_bg 'light';;
+        (light) set_bg  'dark';;
+        esac
+    }
+    colorscheme solarized
+}
+
 %sh{
     # Base color definitions
-    base03="rgb:002b36"
-    base02="rgb:073642"
-    base01="rgb:586e75"
-    base00="rgb:657b83"
-    base0="rgb:839496"
-    base1="rgb:93a1a1"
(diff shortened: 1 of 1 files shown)


lenormf:
The idiomatic way is rather to separate the themes into files, e.g. `solarized.kak` and `solarized_light.kak`, otherwise it mixes logic with data.

colors/solarized.kak
```diff
@@ -1,24 +1,70 @@
...
+decl -hidden str solarized_initially light
+
+# Map the following toggle function to an alias or a key sequence
+# for easy toggling between light and dark, eg,
+# map global user t :solarized-toggle-background<ret>
+# 
+def -docstring 'toggle solarized background between light and dark' -allow-override \
+solarized-toggle-background %{
+    %sh{
+        set_bg() printf %s\\n "set global solarized_background $1"
```
lenormf: Does that work in your shell? There are no braces for the implementation of the function.

colors/solarized.kak
```diff
@@ -1,24 +1,70 @@
...
+        case "$kak_opt_solarized_background" in
+        ( dark) set_bg 'light';;
```
lenormf: Why are the values matched against surrounded with parens? The usual syntax is `case $foo in …);; esac`.

colors/solarized.kak
stevenyvr987: When defining one-liner functions, I usually leave out the braces; I put them for multi-line functions. I'm using dash.

colors/solarized.kak
stevenyvr987: The left bracket is optional and most people leave it out, but I find it visually unsettling not to have the bracket closed; also, having it missing may mess up any bracket matching logic in an editor.

stevenyvr987:
I considered writing a solarized_light.kak, but it would mostly repeat solarized.kak, since there is a simple transform between light and dark for solarized. Also, I didn't know where to define the toggle function and I didn't want to add a new file for such a slight feature.

doppioandante:
I understand your reasoning but I would also be for a separate colorscheme script, the cost of duplication is not that bad in this case and changing color scheme through the :colorscheme command is more predictable than an ad hoc command.

stevenyvr987:
In light of the above conversation, I'm closing this PR and updating it with PR #1394.

PR #1393, opened 2017-05-23
[WIP] Add ncurses_cursor ui_options.

ekie:
Let the ncurses_ui interface handle the cursor style.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -304,6 +304,8 @@ void register_options()
                        "The ncurses ui supports the following options:\n"
                        "<key>:                        <value>:\n"
                        "    ncurses_assistant             clippy|cat|dilbert|none|off\n"
+                       "    ncurses_cursor                block|ibeam|underline\n"
+                       "    ncurses_cursor_blink          bool\n"
                        "    ncurses_status_on_top         bool\n"
                        "    ncurses_set_title             bool\n"
                        "    ncurses_enable_mouse          bool\n"
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -1039,7 +1039,7 @@ void NCursesUI::enable_mouse(bool enabled)
     fflush(stdout);
 }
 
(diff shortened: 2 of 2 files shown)


lenormf:
This doesn't really make sense, the "cursor" is always a one character wide selection, and whatever the terminal decides to render the hardware cursor as shouldn't be taken into account by the editor. We should let the terminal handle this.

ekie:
well the terminal will still handle the hardware cursor this just offers a way to set the visual representation, moreover this assures consistent user experience with a default setting. there was some confusion when 91ed57c broke the colorschemes and failed to highlight the cursor.

blastrock:
+1 on this

Even if there is always a one-char wide selection, when we press 'a', we clearly see the selection before the cursor and a cursor which is one-char wide after that selection. The cursor in this case is not part of the selection.

I think it makes a lot of sense to change the cursor shape to reflect that you are in insert mode. It is far easier to see than the "insert" that appears in the status bar. Also, as a user coming from neovim with NVIM_TUI_ENABLE_CURSOR_SHAPE, I find that it is a bit of a regression.

blastrock:
I actually tried to make this pull request work. The problem is that the cursor is rendered by the Window class, it just highlights the cursor block with a different color. The NCursesUI class disables the terminal cursor rendering in its constructor.

Cursors must be rendered by the Window class because when there are multiple ones, the terminal will not be able to handle them. So I made a compromise and made the Window class render all cursors except the main one and let it be rendered by NCursesUI.

I am not particularly happy with the code, so I am leaving what I have done here in case anyone has a better idea: https://github.com/blastrock/kakoune/commit/ae1b92469148c37c57846c307dd4dd7a9bb47052

mawww:
I dont plan to merge this, we have experimented with visible ncurses cursor some time ago, and it broke many terminals. Kakoune will keep full control of cursor display for the time being.

PR #1394, opened 2017-05-23
Solarized light

stevenyvr987:
Here is a simple PR to add a solarized light color theme to the existing dark theme. This updates my previous PR #1391, which I'm closing.

### DIFF
--- a/colors/solarized_light.kak
+++ b/colors/solarized_light.kak
@@ -1,15 +1,15 @@
-# solarized dark theme
+# solarized light theme
 
 %sh{
     # Base color definitions
-    base03="rgb:002b36"
-    base02="rgb:073642"
-    base01="rgb:586e75"
-    base00="rgb:657b83"
-     base0="rgb:839496"
-     base1="rgb:93a1a1"
-     base2="rgb:eee8d5"
-     base3="rgb:fdf6e3"
+     base3="rgb:002b36"
+     base2="rgb:073642"
+     base1="rgb:586e75"
+     base0="rgb:657b83"
+    base00="rgb:839496"
+    base01="rgb:93a1a1"
+    base02="rgb:eee8d5"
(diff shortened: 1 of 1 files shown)


lenormf:
There are weird hunks in the original solarized file, which modify some whitespace, and those have apparently been mirrored onto the new file which results in wonky alignment.

lenormf:
Use github to open the files in this PR in "raw" mode, you'll see that things are mis-aligned. You might have to do some `%@` in the buffers.

stevenyvr987:
The files are as intended; there are no mixed uses of tabs and whitespaces. The irregular leading spaces in some of the lines are a side effect of my intention to vertcally align the rgb values so that they can be easily checked by eye against each other and against the corresponding values listed in the solarized website.

lenormf:
Ok, I wasn't looking at things that way.

Delapouite:
Thanks for your contribution @stevenyvr987 

I tried to use your solarized_light colorscheme. and I had some trouble. My terminal is in dracula (https://draculatheme.com/), and you did not redefine all the colors. So in the end I have a broken mix of dracula and solarized color.

For example:
```
face PrimarySelection   white,blue
```
should be:
```
face PrimarySelection   ${base03},${blue}
```
and so on…

stevenyvr987:
@Delapouite sorry for the delayed response, I saw your message just now.

I think you've identified a shortcoming with the original solarized.kak file: some of the face def'ns have hard-coded the colors black and white, but like you pointed out they should use base03 and maybe base0, respectively. Also, I think that the faces using default+b should maybe use a base color combination instead.

Let me wrap my head around this a bit longer and later submit a corrective commit.

Delapouite:
yes, some work may need to be done to the original solarized (dark)

bhajneet:
I made these fixes in #1627

PR #1396, opened 2017-05-24
doc: Turn general documentation into actual man pages

lenormf:
Hi,

Getting `man` to display compressed pages stored in an arbitrary directory has proven difficult, since not two implementations follow the same conventions (environment variables, flags etc). After receiving complaints that the documentation pages would fail to be displayed, I've decided to make this PR to turn all our "static" documentation into actual man pages, which should fix all our issues.

The pages are prefixed with `kak-` and installed into section `7`, as is I believe the case for this kind of on board documentation. The only issue is that we wouldn't be able to show those manpages from a locally compiled `kak` binary (as is the case as of today), but I think that's for the best.

Also this is only a proof concepts to gather comments, so the branch will be rebased in the future.

HTH.


lenormf:
Another issue that we have with the current implementation is shown in the following bug report on the Gentoo tracker:
https://bugs.gentoo.org/show_bug.cgi?id=619562

Users who compile the project with a different `docdir` variable end up unable to read the documentation onboard because scripts and compilation options are separated.

Right now the completion candidates for the `:doc` command are a static list, not sure how to handle that dynamically.

lenormf:
I allows the `:doc` command to take (as it was before) a keyword, as last optional parameter, which contains a string to select in the resulting buffer once the man page has been opened. This resulted into breaking changes:

* the `:man` command know takes an optional second `keyword` parameter, which is forwarded from the `:doc` command when specified
* the current selection isn't read anymore when no page name is specified in the command to allow this `keyword` parameter to be optional
* the changes made to the `:doc` command previously have been applied to the `:man-impl` command as well, i.e. using `sed` instead of `col` to allow simplifying the code and reduce the amount of processes spawned by the shell script
* the buffer is now properly rewritten when the window is resized

lenormf:
Fixed a bug with weird section names (`ssl(7ssl)`?), which actually reduced the amount of processes spawned when populating the completion candidates.

I think this is ready to be merged now.

lenormf:
Rebased on f9c48237a7d3a8c2aae44c532a398f62ea0100f1.

mawww:
I have two problems with that change:

1. The need to maintain a hardcoded list of doc pages for the doc command completion.
2. The fact that `:doc` does not work anymore if Kakoune is not installed.

That second point is my biggest gripe, up to now Kakoune has been working the same way installed or running from src/.

I wonder if we can go a different way, maybe using something else than man as a backend (or maybe using the roff toolchain manually).

lenormf:
On the other hand right now some users don't have access to the documentation at all because their implementation of `man` doesn't allow reading local files, which is my biggest concern.

1. I couldn't find any other way than hardcoding a list (assuming we're not pre-processing the files to insert the list of all documentation files detected at build-time).

2. I'd rather have everybody be guaranteed to be able to read the documentation even if that means installing the man pages to the system, rather than allow those who use the editor from the build directory to read the documentation at the depends of those who don't use a compatible `man` implementation (those won't be able to open the documentation in any case).

One of the other advantages of this new way to handle the documentation is that we don't have to maintain two shell scripts anymore, everything is handled by the `man` command. I'm afraid that trying to use `nroff` is going to induce the same kind of issues we've been struggling with with the `man` implementations, meaning that some users will have `nroff` on their system, others `groff`, etc.

Issue #1130, opened 2017-01-12
expose $kak_command_history and $kak_search_history


Delapouite:
Hi

Use cases for these variables in rc scripts :
- save a history to a file
- pass a history to an external interactive filter like fzf to browse through the list of previous commands or searches

lenormf:
Passing an entire history of commands through a single environment variable might yield undefined behaviour due to its size, depending on the shell that's run. Also it's going to take a massive hit on performance, as everything would have to be escaped to allow parsing using a delimiter.

The idea is good, but variables probably shouldn't be how it should be implemented.

Delapouite:
You're right.

My issue could be more oriented about what I want to achieve than how this can be implemented.
Maybe these lists can be retrieved by asking the kak executable itself with a flag? Something like that:

```
kak -list-history ${kak_session_id}
```

mawww:
I am not sure there is better than env vars for that:
* Not convinced we will hit any size limitations, history is not that big
* Escaping is going to be a problem no matter what way we send history, as long as it ultimately goes through text, we will have to separate entries somehow.

`kak -list-history` means we introduce "answers" in the remote protocol, which I still want to avoid.

So, in my opinion, the real problem here is to choose a nice delimiter. `\n` is not an option because we support multiline commands (write a multiline command in a buffer, select it and do `:<c-r>.` to see). `\0` Wont work if we use env vars (but thats not a strong enough argument to use something else).

Or we could use `\n` and escape the ones that are not history delimiters.

lenormf:
>Not convinced we will hit any size limitations, history is not that big

I have a kak client running on a remote server I've been using for a couple months, if I had a way of counting I many commands I've issued in it I'm sure I would hit a 3 digits number. Now imagine concatenating all that together, adding delimiters, escaping, and you get a very big variable.

There's no POSIX limit to an environment variable's length that I know of, but I know there an implementation-defined limit to the length of the values that will be handled by shells. Even if there was an environment variable that held all of the commands history, chances are using it would be unpredictable because it depends on `/bin/sh` (which can be anything as we know).

>Escaping is going to be a problem no matter what way we send history, as long as it ultimately goes through text, we will have to separate entries somehow.

Taking into account the fact that the length of the commands history is not predictable (and can be fairly big as is my case), I'd rather not have my shell scopes randomly spawning with a delay caused by the editor escaping everything. This also becomes a concern if we assume that at some point we'll have a mechanism that restores command history across restarts.

[...]

PR #1397, opened 2017-05-25
Expose last entered command in register :

danr:
I've often written some command in the command line only to later wish I had written it in a buffer. This patch puts the last entered command in register `:`, similar to the `|` and `/` registers.

(Related feature request: #1130 would make this information available as shell expansion)

### DIFF
--- a/src/input_handler.cc
+++ b/src/input_handler.cc
@@ -149,7 +149,8 @@ constexpr StringView register_doc =
     "    * @: default macro register\n"
     "    * /: default search register\n"
     "    * ^: default mark register\n"
-    "    * |: default shell command register\n";
+    "    * |: default shell command register\n"
+    "    * :: last entered command\n";
 
 class Normal : public InputMode
 {
--- a/src/main.cc
+++ b/src/main.cc
@@ -171,7 +171,7 @@ void register_registers()
 {
     RegisterManager& register_manager = RegisterManager::instance();
 
-    for (auto c : "abcdefghijklmnopqrstuvwxyz/\"|^@")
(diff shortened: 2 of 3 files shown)


lenormf:
Are there any other uses of this feature, other than the occasional access to the last validated command?

danr:
None comes to mind. 

It seems as useful as the `|` register. What are the uses of the `|` register?

lenormf:
It holds the command that will be automatically used if hitting the return key on an empty prompt, after using e.g. the `|` key.

I don't think a similar behavior would be really useful on the command line, which makes this new functionality quite limited in usefulness.

mawww:
This gives me an idea, we could maybe store actual histories in the registers.

Registers are list of strings, in the case of `/|:` we could store the history directly in them, which would provide a way to get the history (through normal register access, `"/<a-p>` would paste the whole search history), and a way to restore it (`"/y`)).

I'd need to think a bit more about that to see if its really a good idea, but it could solve a few issues in one go.

Any opinions on that ?

lenormf:
Using registers as a long-term storage data structure doesn't sound that hot.

Screwtapello:
> we could maybe store actual histories in the registers

Sometimes I `"/y` to avoid the magic behaviour of `*`, and it would be a shame to lose my entire search history for such a thing. On the other hand, prompt registers are already somewhat special in that they only care about the last item, so maybe we could make them more special by saying that yanking something to that register (or using the `:set-register` command) only appends new items (or whatever Kakoune's interactive behaviour is), rather than replacing. Then it would be really easy to write a plugin to persist them.

PR #1405, opened 2017-05-30
fix: remove duplicate include to containers.hh in face_registry.cc

Delapouite:


lenormf:
Does this fix an error in particular? Did you use a tool to find that out?

Delapouite:
No tools used. I just noticed it by skimming through the code.

PR #1422, opened 2017-06-06
Update command names

alexherbo2:


mawww:
Thats, I think, a desirable change, but it'll break user configuration, so we need to discuss it a bit on IRC to see if its worth it.

lenormf:
I would actually go further and have even more explicit command names:

* `lint-previous-line`
* `grep-previous-match`
* `make-previous-error`

The corresponding `-next` commands would consequently have to be renamed as well, potentially.

PR #1423, opened 2017-06-07
Clarify rotation direction between <'> and <a-'> in info help

Delapouite:


occivink:
When do these messages appear?

Delapouite:
When you press on either `'` or `<a-'>` and you have the `autoinfo` option enabled

PR #1429, opened 2017-06-09
rc: Fix calls to `mktemp`

lenormf:
Allow `mktemp` to make use of the `TMPDIR` environment variable when
calling it with a template.

Don't use the deprecated `-t` flag.

### DIFF
--- a/rc/base/lint.kak
+++ b/rc/base/lint.kak
@@ -8,7 +8,7 @@ decl -hidden range-specs lint_errors
 
 def lint -docstring 'Parse the current buffer with a linter' %{
     %sh{
-        dir=$(mktemp -d -t kak-lint.XXXXXXXX)
+        dir=$(mktemp --tmpdir -d kak-lint.XXXXXXXX)
         mkfifo "$dir"/fifo
         printf '%s\n' "eval -no-hooks write $dir/buf"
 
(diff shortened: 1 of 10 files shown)


casimir:
This PR break all the modified scripts on macOS.

Example for lint.kak:
```
shell stderr: <<<
mktemp: illegal option -- -
usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix 
mkfifo: /fifo: Permission denied
>>>
Error: 1:1: 'lint' 1:1: 'eval' 1:1: 'write' /buf: Permission denied
```

lenormf:
So the mac implementation excepts a `-t` flag to know what format to use, but the GNU implementation considers this flag deprecated (and will potentially remove it in newer releases) and doesn't seem to take `$TMPDIR` into account all the time unless the mac-unsupported `-p` flag is passed…

mawww:
Can we just use `mktemp -d ${TMPDIR:-/tmp}/blah-XXXX` in that case ? or will that break on OSX ?

lenormf:
I think this will work, I'll adapt my changes.

casimir:
It's working with just `-d` on OSX.

lenormf:
It doesn't on all versions, i.e. on OS X 10.10.5.

mawww:
Whats the problem there ?

lenormf:
It simply shows a usage message [apparently](https://unix.stackexchange.com/questions/30091/fix-or-alternative-for-mktemp-in-os-x#comment522946_84980), the template was probably made optional in recent versions.

lenormf:
Follow up in #1449.

PR #1430, opened 2017-06-09
[WIP] share: Introduce the `sh` directory, `mktemp` implementation

lenormf:
This commit installs an `sh` directory in the share directory, in which
implementations of popular tools that might not be available on a user's
system sit.

In order for a shell scope to use those implementations if the regular
binaries are unavailable, the `PATH` variable has to be modified as follows,
at the beginning of the scope:

```
%sh{
    export PATH="${PATH}:${kak_runtime}/sh"

    <code>
}
```

### DIFF
--- a/rc/base/lint.kak
+++ b/rc/base/lint.kak
@@ -8,6 +8,9 @@ decl -hidden range-specs lint_errors
 
 def lint -docstring 'Parse the current buffer with a linter' %{
     %sh{
+        # fallback implementation: mktemp
+        export PATH="${PATH}:${kak_runtime}/sh"
+
         dir=$(mktemp -d "${TMPDIR:-/tmp}"/kak-lint.XXXXXXXX)
         mkfifo "$dir"/fifo
         printf '%s\n' "eval -no-hooks write $dir/buf"
--- a/rc/base/spell.kak
+++ b/rc/base/spell.kak
@@ -11,6 +11,9 @@ Formats of language supported:
     spell %{
     try %{ add-highlighter ranges 'spell_regions' }
     %sh{
+        # fallback implementation: mktemp
+        export PATH="${PATH}:${kak_runtime}/sh"
+
         file=$(mktemp -d "${TMPDIR:-/tmp}"/kak-spell.XXXXXXXX)/buffer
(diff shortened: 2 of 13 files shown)


rc/base/lint.kak
```diff
@@ -8,6 +8,8 @@ decl -hidden range-specs lint_errors
 
 def lint -docstring 'Parse the current buffer with a linter' %{
     %sh{
+        export PATH="${PATH}:${kak_runtime}/sh"
```
mawww: I think a '# mktemp fallback' comment should follow, to make it explicit which fallback we try to import

lenormf:
I added a comment hinting that we modify the path to have access to the `mktemp` custom implementation.

lenormf:
As it turned out after merging #1429, even such a simple utility uses different flags depending on the environment (GNU or mac basically). I think we should not extend the `PATH` and use the implementation as a fallback in those cases (i.e. `mktemp`), and have a wrapper/reimplementation named differently instead (`kak_mktemp`).

mawww:
Yeah, I am still on the fence about this change (I know I suggested it 😛), I tend to think we should not start adding a set of additional commands bundled with Kakoune, but we are hitting these portability problems from time to time.

Kakoune should not try to fix posix problems itself, we try to stay as posix as possible to maximize portability, but if we dont have an acceptable tool provided by posix, I am not sure we should be rolling our own instead of just depending on the lowest common denominator that is usually available.

This needs more discussions and thoughts...

lenormf:
Note that the original idea was to provide a minimal fallback implementation in case the tools were not available on the system. I'm OK with not trying to abstract already existing APIs (e.g. GNU vs BSD `mktemp`).

I'll rework the `mktemp` implementation to make it only accept a `-d` flag, and require a template as argument (it seems to be what's working on both linux and mac).

lenormf:
Rebased on f9c4823, simplify the `mktemp` alternative implementation to be seamlessly usable with the current code.

mawww:
I am wondering on which platform dont we have mktemp available ?

lenormf:
Good point, `mktemp` is also provided by busybox, which I use as a reference for availability of a command.

So really the only script we could use now is `readlink`, right ?

PR #1434, opened 2017-06-12
Allow reverse ranges in highlighter

occivink:
Rationale: I'm using a "marks" highlighter and marks are defined from cursor to anchor, which may sometimes be "reversed".


occivink:
Thanks!

PR #1438, opened 2017-06-14
rc ctags: Jump into the `jumpclient` client

lenormf:


mawww:
Not sure this is desirable, jumping to a tag is a manual process, its hard to know if we want to jump in the current client or in another one (seems its clearer from make/grep than when you type `:tag`). I'd rather let the user choose how they want to jump by wrapping the command themselves instead of forcing one policy on them.

lenormf:
I thought the `jump` client existed to open all the buffers that were jumped to, whether it was user induced or not (that includes `lint` too by the way).

Arguably whether or not the target buffer is opened in a separate client is equally as annoying depending on the behavior that you want, because there's no simple way to make the `tag` command jump/not jump in its current state that I know of.

What changes could be done to the `ctags.kak` module so that both workflows are available?

lenormf:
I ended up writing a wrapper, which is probably best for now:

```
def ctags-jump-client %{
    eval -try-client %opt{jumpclient} "
        edit %val{buffile}
        select %val{selections_desc}
        exec %{<a-i>w: tag<ret>}
        focus
    "
}
map -docstring 'lookup the work under the cursor in the tags file' global user '}' %{:ctags-jump-client<ret>}
```

PR #1439, opened 2017-06-14
src: Strip commands' documentation string

lenormf:
Arguably this is a cosmetics commit, but the more heavily "documented" commands might benefit from a shell here-document type behaviour, splitting clearly the documentation from the flags and the implementation.

More information in the commit message.

HTH.


lenormf:
Maybe this change would also be welcome on the options' docstring, they've gotten fluffier lately.

PR #1446, opened 2017-06-15
rc ctags: Make commands follow the naming convention

lenormf:


rc/base/ctags.kak
```diff
@@ -19,9 +19,9 @@ def -params ..1 \
             fi
             cat "$namecache"
         done' \
-    -docstring %{tag [<symbol>]: jump to a symbol's definition
+    -docstring %{ctags-search [<symbol>]: jump to a symbol's definition
 If no symbol is passed then the current selection is used as symbol name} \
-    tag \
+    ctags-search \
```
casimir: What about `ctags-jump` instead?

lenormf: @mawww If I rename the command to `ctags-jump`, will it be enough proof that the command needs to jump into the `jump` client :p ?

mawww: I am not sure thats what we want, if I type `:ctags-jump tag`, would I really want to have the focus change to another client and the file be opened there ? it seems wrong to me, as thats pretty different from hitting `<ret>` in a `*make*` or `*grep*` buffer.

Switching client is easy (at least your WM/terminal multiplexer should make it easy), so it seems better to let the user chose in which client they want to jump by just typing the `:ctags-jump` in that client.

pickfire:
Can we do `gd` like vim `ctrl-]` to jump between tags?

PR #1448, opened 2017-06-15
Add 'Source' hook

Delapouite:
There's currently no easy way to know which `kak` files are sourced.
This hook provides info about this, it's triggered every-time the `source` command is used and give the path.

This way we can ease coordination about conditional or/and dependent sourcing of kak scripts.
Examples in @danr 's [kakrc](https://github.com/danr/dotfiles/blob/4f0b117d7d15ea6f164610298c1d02c1786d5383/config/kak/kakrc) import / plug logic

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -1467,6 +1467,8 @@ Existing hooks are:
        hidden.
  * `RawKey`: Triggered whenever a key is pressed by the user, the key is
        used for filtering.
+ * `Source`: Triggered when a file is sourced. The filtering text is
+       the file path
 
 When not specified, the filtering text is an empty string.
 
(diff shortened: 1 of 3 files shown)


mawww:
I dont get the use case for that, what kind of hook on `Source` would help with sourcing dependent scripts ? Would that be used to delay execution of commands from one script until another script is eventually loaded ?

I'd like to get a good understanding of how that might be useful before merging.

lenormf:
If the end goal is to allow deterministic loading order of scripts, I think working on that first would be better than merging a feature that eventually could be used to the same ends. The use case presented for this feature seems very specific, is it the simplest/only way to synchronize scripts loading?

Delapouite:
This PR has merge conflicts. It didn't reach consensus, so closing.

PR #1450, opened 2017-06-16
Remove rc/ranger

alexherbo2:
Rationale is `rc/ranger` is superseded by [Opening files in specific client][Open files in specific client] approach.

- https://github.com/mawww/kakoune/wiki/Open-files-in-specific-client
- https://youtu.be/v_Ffno9wiJ4

[Open files in specific client]: https://github.com/mawww/kakoune/wiki/Open-files-in-specific-client


Delapouite:
I think the `terminal.kak` and `shell.kak` (more generic) must be included in `rc/extra` before deleting this script.

alexherbo2:
I think same, but I don’t get the occasion to discuss that with @mawww.

`:shell` and `:terminal` commands are not that useful without providing the `kak` which makes use of `$KAK_{SESSION,CLIENT}` to open files.

Delapouite:
I found that I like ranger to open with the current file selected by default

`
ranger --selectfile=$KAK_BUFFILE
`

To do so, I had to modify `shell.kak` to add : 
```
 # Ensure Kakoune sets these variables by making them appear in the source:
  # – kak_session
  # – kak_client
  # – kak_buffile
```

alexherbo2:
I added to the wiki and my dotfiles.

PR #1455, opened 2017-06-17
Use user-supplied autoshowcompl option's value.

nuao:
I couldn't switch autoshowcomplete off with 

`:set buffer autoshowcompl no`

The replacement of `m_autocompl(true)` here seems to fix this.


nuao:
I closed this PR and created another for the same issue because I forgot to add the contributor's empty commit message. This duplicate PR should be completely deleted, but I'm not aware of a way to do that.

PR #1456, opened 2017-06-17
Use user-supplied autoshowcompl option's value.

nuao:
I couldn't switch autoshowcomplete off with

`:set buffer autoshowcompl no`

The replacement of `m_autocompl(true)` here seems to fix this.


mawww:
`autoshowcompl` was intended to control the prompt completions, but I think it makes sense to use it for insert completion as well. If we end up with users wanting distinct options for the two behaviours we can always revisit that later.

Thanks for that PR.

PR #1458, opened 2017-06-19
Rust and Haskell highlighters: don't treat character literal '"' as the start of a string

dianne:
Previously, the character literal `'"'` would cause the rest of the file to be highlighted as a string. The C-family highlighter used a regex for string openings that prevents this, so I've added it to the Rust and Haskell highlighters as well.


dianne:
I've updated the copyright waiver to be in my name. Sorry if there was a better way than a force push; I'm new to github.

PR #1463, opened 2017-06-23
rc formatter: Make sure the formatter returned successfully

### LINKED ISSUE #1357: Format wipes buffer on error with formatcmd
As per @casimir 's request, I'm logging this issue here and closing one in @lenormf 's `kakoune-extras` repo.

The `:format` command will wipe the buffer if the command specified by the `formatcmd` setting returns an error instead of a copy of the code. I'm happy to take a crack at a PR against `rc/core/formatter.kak` if that's the best place to deal with this.

lenormf:
The previous implementation used to replace the contents of the buffer with
whatever the `formatcmd` was returning, regardless of the exit code of the
command, which led to the buffer being wiped out on error.

This commit does the formatting in a temporary file, and only replaces the
current buffer with the contents of the -formatted- temporary file if the
`formatcmd` returned successfully.

Fixes #1357

### DIFF
--- a/rc/core/formatter.kak
+++ b/rc/core/formatter.kak
@@ -1,17 +1,35 @@
 decl -docstring "shell command to which the contents of the current buffer is piped" \
     str formatcmd
 
-def format -docstring "Format the contents of the current buffer" %{
+def format -docstring "Format the contents of the current buffer" %{ eval -draft %{
     %sh{
-        if [ ! -z "${kak_opt_formatcmd}" ]; then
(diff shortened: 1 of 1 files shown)


Delapouite:
That's a welcomed behavior for this tool!
https://github.com/prettier/prettier was busting my buffers from time to time on silly syntax errors.
Thanks lenormf

lenormf:
You're welcome.

I've tested the code with `clang-format` and several combinations of arguments, but I wouldn't mind some extra testing to make this dodgy script works properly.

PR #1473, opened 2017-06-29
Docs: update completions_extra_word_chars → extra_word_chars

Delapouite:
To reflect the changes in https://github.com/mawww/kakoune/commit/f41d78083aa8d3883bcd4ed819e6784e09e69dba

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -958,7 +958,7 @@ Some options are built in Kakoune, and can be used to control its behaviour:
    - `option=<opt-name>` where <opt-name> is a _completions_ option.
  * `static_words` _str-list_: list of words that are always added to completion
      candidates when completing words in insert mode.
- * `completions_extra_word_chars` _str_: a string containing all additional character
+ * `extra_word_chars` _codepoint-list_: list of all additional codepoints
      that should be considered as word character for the purpose of insert mode
      completion.
  * `autoreload` _enum(yes|no|ask)_: auto reload the buffers when an external
--- a/doc/manpages/options.asciidoc
+++ b/doc/manpages/options.asciidoc
@@ -18,7 +18,7 @@ Types
 *regex*::
 	as a string but the set commands will complain if the entered text
 	is not a valid regex
-*int-list*, *str-list*::
+*int-list*, *str-list*, *codepoint-list*::
 	a list, elements are separated by a colon (:) if an element needs
 	to contain a colon, it can be escaped with a backslash
 *range-specs*::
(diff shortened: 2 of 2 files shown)


README.asciidoc
```diff
@@ -958,7 +958,7 @@ Some options are built in Kakoune, and can be used to control its behaviour:
    - `option=<opt-name>` where <opt-name> is a _completions_ option.
  * `static_words` _str-list_: list of words that are always added to completion
      candidates when completing words in insert mode.
- * `completions_extra_word_chars` _str_: a string containing all additional character
```
mawww: Actually, this is not a `str` anymore, its a `codepoint-list` and is `:` separated as other lists.

lenormf: Do you mean `str-list`? This type doesn't exist on `master`.

mawww: Juste because its not available to `declare-option` does not mean it does not exists :)

lenormf: I `grep`'d the type in the source and couldn't find it, to check if it was simply just not completable upon. The search didn't return any match for `codepoint-list`.

Delapouite: Should I update this PR replacing `str` with `codepoint-list` then? Would that be enough?

mawww: Well, it would be nice to describe the fact that its a `:` separated list of codepoints.

mawww: @lenormf you wont find `codepoint-list` directly, its a `Vector<Codepoint>` option, and we get the option type name automatically (Vector<T> options are automatically named `option_type<T>() + "-list"`)

lenormf: >you wont find codepoint-list directly, its a Vector<Codepoint> option

That's my point, if it's not a declarable/complete-able type, then to the users it doesn't exist, so it will be confusing to see its name appear in the documentation.

Delapouite:
I've completed this PR to add doc, autoinfo and autocompletion support for the `codepoint-list` option type.

mawww:
Hum, I am afraid I was misunderstood, there is no reason to make `codepoint-list` available as a custom option type, we have some option types that are specific to one option (`completers` for example), but make not much sense for users to declare themselves. I dont think there is a relevant use case for exposing `codepoint-list` yet. That should not prevent the documentation to say that this option is a `:` separated list of codepoint though (and even use the name `codepoint-list`).

mawww:
I'd say, remove the changes in kakrc highlighting and commands.cc, the rest is fine.

Delapouite:
Thanks. done.

PR #1482, opened 2017-07-05
rc: Split `c-family.kak` into C/C++/ObjC scripts

### LINKED ISSUE #1476: Split `c-family.kak` into language dedicated scripts
Hi,

From IRC:

```
ironzorg | I think the `c-family.kak` script should be split into proper `c.kak`, `cpp.kak` and `objectivec.kak` scripts
ironzorg | it's always been our line of thought that duplicating code across scripts isn't a problem if it simplifies scripts and makes them readable, and the `c-family.kak` script doesn't follow this
ironzorg | also I wouldn't be against potentially saving on the few shells spawned to generate code dynamically
```

HTH.

lenormf:
Closes #1476

### DIFF
--- a/rc/core/c.kak
+++ b/rc/core/c.kak
@@ -1,0 +1,232 @@
+hook global BufCreate .*\.c$ %{
+    set buffer filetype c
+}
+
+hook global BufCreate .*\.(h|inl)$ %{
+    try %{
+        exec -draft %{%s\b::\b|\btemplate\h*<lt>|\bclass\h+\w+|\b(typename|namespace)\b|\b(public|private|protected)\h*:<ret>}
+    } catch %{
+        set buffer filetype c
+    }
+}
+
+def -hidden c-trim-autoindent %[ eval -draft -itersel %[
+    # remove the line if it's empty when leaving the insert mode
+    try %[ exec <a-x> 1s^(\h+)$<ret> d ]
+] ]
+
+def -hidden c-indent-on-newline %< eval -draft -itersel %<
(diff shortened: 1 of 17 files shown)


lenormf:
This PR is ready now.

mawww:
I do like the effort, but it seems it gives us lots of duplicated code between c.kak and cpp.kak.

I agree c-family.kak is uglier than necessary, but I am not convinced we get in a much better state by duplicating all that code around.

lenormf:
As mentioned in the linked issue report, `duplicating code across scripts isn't a problem if it simplifies scripts and makes them readable`.

Also users who don't want to enable support for e.g. objC by not linking `c-family.kak` in their `autoload` directory have to link this language factory that spawns shells every which way and slows down -ever so slightly- startup time.

Finally it makes customization of the highlighters of C/C++/objC more complicated to new users (or just those that can't read shell) because the file compiles a template with different values.

andreyorst:
> I do like the effort, but it seems it gives us lots of duplicated code between c.kak and cpp.kak.

Can't `c.kak` be sourced from within `cpp.kak` and therefore `cpp,kak` will contain only cpp specific stuff, like highlighting for templates, classes, etc? I'm looking forward to extend c-family.kak, to have more highlighters for example for method like syntax, and user types, such as `mytype_t`, as well as fixing define overlapping, since it colors inline comments as defines.

mawww:
I think this is sufficiently out-of-date to be closed now.

PR #1483, opened 2017-07-05
[WIP] rc alternative: Implement a generic command to open the header/implementation buffer

lenormf:
Hi,

This commit implements a generic command to open the counter part of the current file in a header/implementation architecture. It should be slightly faster than the current implementation, handle filenames containing spaces and be less cumbersome to add new header/implementation pairs.

Mostly WIP because I haven't really surveyed what supported languages should be added to the script.

HTH.


lenormf:
The issue with this current implementation is that the `open_alternative` function will loop on every single extension, even when of them matched the file we're looking for. This is due to the nested `while` loops, which we can't break -cleanly- out of from the line that prints the `edit` command.

lenormf:
The issue with `while` loops is that their behavior isn't consistent from one shell to the other when it comes to leaving them.

Example:

```bash
$ bash -c "echo | while read; do exit 1; done; echo end"
end
$ zsh -c "echo | while read; do exit 1; done; echo end"
```

We have to use `exit` here to push a return code up to the parent process, as we can't modify variables from within a `while` scope either (no `found=1` possible).

The following does work on all shells, but forces the creation of a subshell:

```bash
$ bash -c "echo | ( while read; do exit 1; done; ); echo end"
end
$ zsh -c "echo | ( while read; do exit 1; done; ); echo end"
end
```

So in the end I had a choice, either let the loops look for alternative files even though one had already been found (we don't even try to break out of the loops), or additional subshells are spawn to run each loop to allow stopping the iterations as soon as a candidate is found.

I went for the latter, I'd rather create too many shells than issue a syscall for all the possible files that could be considered but that we still know won't be opened.

Rebased on c702c915c1e3ad2df354f38132645a72c5d577f2.

mawww:
I am unconvinced that we have a need to abstract that into a general feature. It would make sense if we had multiple implementations of alternative-file for various file types and they were very close in implementation, but here the generic one only supports c-family, and even uses C specific lingo (header), so its basically extracting some code from c-family.kak into its own file.

lenormf:
The current implementation of the command is not specific to the C-family, it simply parses extensions, so it makes sense to move it to its own file and make it generic.

Another issue is that the extensions handled by it are hardcoded, it sets a limit on the re-usability of the function if users want to use it with files with unsupported extensions. Users also have to override the entire language file if they want to modify the `alt` command to support a custom extension.

This implementation is cross-compatible with the current one as far as I know.

Delapouite:
lua, moonscript and ruby also register an `alt` command

mawww:
Yep, what I want to see is lua, moonscript and ruby ported to that new system, naming thats not referencing a concept absent from all these languages (header), especially as the alternate file is actually a unit test file in some of those, and end up with code thats smaller/clearer than the sum of code it replaces.

What I dont like about the current PR is that it just moves the c-family functionality out of c-family.kak, and gives us another instance of the "where to set the option" dilemma (in the filetype.kak or in the feature.kak).

lenormf:
I didn't see that other languages had their own custom implementation, although the "spec" pattern in lua and moon doesn't seem to be standard.

In that case, in order to fix the issues I was mentioning, how about we keep in `c-family.kak` a somewhat generic implementation that only opens a header/source file, which will then be re-used wherever needed (through an alias)? As for the other two languages, we could have one generic implementation that looks for a file with a given prefix/suffix? Both would need to be somewhat generic though, to allow tweaking the extensions that would be detected.

PR #1487, opened 2017-07-06
Docs: add incipit for options manpage presenting common commands

Delapouite:
Hi

Here's a small but comprehensive intro about options, following the existing ones in [hooks](https://github.com/mawww/kakoune/blob/master/doc/manpages/hooks.asciidoc) and [mapping](https://github.com/mawww/kakoune/blob/master/doc/manpages/mapping.asciidoc) man pages.

### DIFF
--- a/doc/manpages/options.asciidoc
+++ b/doc/manpages/options.asciidoc
@@ -5,6 +5,37 @@ NAME
 ----
 options - a
 
+Description
+-----------
+
+Options are used for user configuration.
+You can set the builtin ones or declare your owns.
+
+The following commands are available to interact with them:
+
+----------------------------------------------
+declare-option [-hidden] <type> <name> [value]
+set-option [-add] <scope> <name> <value>
+update-option <scope> <name>
+unset-option <scope> <name>
+----------------------------------------------
+
+Refer to the 'scopes' documentation page for more information about
+scopes.
+
+To get the current value of an option:
+
+--------------------------------
+echo %opt{name}
+echo %sh{ echo "$kak_opt_name" }
+--------------------------------
+
+To display the current values of all options in the debug buffer:
+
+-------------
+debug options
+-------------
+
 Types
 -----
 *int*::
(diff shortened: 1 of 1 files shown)


doc/manpages/options.asciidoc
```diff
@@ -46,9 +77,6 @@ Types
 	a set of flags, taking a combination of the given values joined by a
 	'|' character
 
-Refer to the 'scopes' documentation page for more information about
```
lenormf: Why was this removed?

Delapouite: This line was moved a bit further up.

doc/manpages/options.asciidoc
```diff
@@ -5,6 +5,37 @@ NAME
...
+declare-option [-hidden] <type> <name> [value]
+set-option [-add] <scope> <name> <value>
+update-option <scope> <name>
+unset-option <scope> <name>
+----------------------------------------------
+
+Refer to the 'scopes' documentation page for more information about
+scopes.
+
+To get the current value of an option:
```
lenormf: This bit duplicates the contents of the `expansions` page, which increases the probability of having discrepancies in the documentation when either page changes.

#1470 did the same thing as well, if pages' topics intersect then we should redirect one to the other with a "for more information about XXX, refer to the YYY page" type message, instead of potentially creating more opportunities for desynchronization.

Delapouite: Yes, I see what you mean.
It puts more burden on our shoulders as doc maintainers with the benefits for kakoune newcomers to have the info right where they need it.

What mechanism could be used to let the user easily jump from one topic to another like hyperlink? vim's `:help` tag system is very nice in this regard.

lenormf: You mean having e.g. `|expansions|` in the document, and have the page be opened automatically when hitting `<ret>` (or clicking?) on it? It would be nice, also we could use a character other than `|`, if those sequences aren't interpreted by `a2x`.

Delapouite:
Similar incipit has been added.
https://github.com/mawww/kakoune/blob/master/doc/pages/options.asciidoc
Closing

Issue #1479, opened 2017-07-03
Preserve declaration order in user mapping infobox


occivink:
Currently the order of the user mappings in the infobox is seemingly arbitrary . It would be nice if it respected the order of declaration so that we could group similar mappings together.

Delapouite:
> is seemingly arbitrary

Isnt it alphabetical ?

occivink:
Oh you're right, the `<a-*>` mappings threw me off.

lenormf:
>>is seemingly arbitrary

>Isnt it alphabetical ?

The entries are sorted by ASCII value (which means "alphabetically" in programming parlance) in `keymap_manager.cc:49`. Since this order doesn't really make sense to the average user (who will wonder why capitalized keys are placed after punctuation signs, apart from lower case keys), I suppose we could fix it in `normal.cc:134`, where this list is generated.

mawww:
It would be possible to follow the order of definition, although it would not be perfect as we'll still have mappings grouped by scope (global mappings first, then buffer mappings, and finally window mapping).

If thats preferred, it should not be too hard to switch to that.

occivink:
I don't use any scope-specific mappings so I can't say much, but even then the grouping would be more "logical" I think.

PR #1494, opened 2017-07-11
Preserve declaration order for mappings.

ekie:
#1479


mawww:
Sorry, I missed that and did the change myself. Although this was not correct, std::unique only works for sorted lists.

ekie:
thank you for the feedback.

PR #1499, opened 2017-07-11
rc formatter: Restore selections after formatting

### LINKED ISSUE #1492: formatter.kak does not restore cursor position
Hello

I noticed a broken behavior since the last update on `formatter.kak` https://github.com/mawww/kakoune/commit/6376aa6206aefb408fef75ecfb21648418656301
The enclosing `eval` is done in a draft context, therefore I think we do not have access to the correct location for `$x` and `$y` anymore.
The visual effect is that the cursor always ends up at the bottom of file after each formatting.

lenormf:
The previous implementation was using a draft context, which prevented the
selection to be properly restored. This commit makes use of the updating
mechanism provided through the `update-option` command, in order to try to
restore as much of the original selection as possible onto the formatted
buffer.

Multiple selections are also now handled (only the main selection was
handled before).

Fixes #1492.

### DIFF
--- a/rc/core/formatter.kak
+++ b/rc/core/formatter.kak
@@ -1,13 +1,19 @@
 decl -docstring "shell command to which the contents of the current buffer is piped" \
     str formatcmd
 
-def format -docstring "Format the contents of the current buffer" %{ eval -draft %{
-    %sh{
-        if [ -n "${kak_opt_formatcmd}" ]; then
-            ## Save the current position of the cursor
-            readonly x=$((kak_cursor_column - 1))
-            readonly y="${kak_cursor_line}"
+# This option holds the selection as it was at the time of calling the `format` command
+# After the buffer has been formatted, the selection will be restored according to
+# the value held in this option
+decl -hidden range-specs formatter_prevsel
 
+def format -docstring "Format the contents of the current buffer" %{
+    set buffer formatter_prevsel "%val{timestamp}"
+    eval -itersel %{
+        set -add buffer formatter_prevsel "%val{selection_desc}|Default"
+    }
+
+    eval -draft %{ %sh{
+        if [ -n "${kak_opt_formatcmd}" ]; then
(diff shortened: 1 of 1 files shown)


rc/core/formatter.kak
```diff
@@ -1,13 +1,19 @@
...
-def format -docstring "Format the contents of the current buffer" %{ eval -draft %{
-    %sh{
-        if [ -n "${kak_opt_formatcmd}" ]; then
-            ## Save the current position of the cursor
-            readonly x=$((kak_cursor_column - 1))
-            readonly y="${kak_cursor_line}"
+# This option holds the selection as it was at the time of calling the `format` command
+# After the buffer has been formatted, the selection will be restored according to
+# the value held in this option
+decl -hidden range-specs formatter_prevsel
```
mawww: I dont think that will work until piping does diff the input and the output and uses that as the actual modifications. Currently piping just replaces the input with the output, which means update-option will not give use anything interesting.

lenormf: I assumed the command would work on any modification made to the buffer, and it seems that this implementation leaves the main selection (and occasionally the secondary selections as well) in the proper location… Should the call to the option be removed, or should the primitives that modify the buffer do a diff?

mawww: if you do `%|cat<ret>`, the buffer contents is unchanged, but for Kakoune, we did remove the whole content, then insert it again, so any selections will get pushed back to the last char of the buffer (remove whole content makes all selection select the first char, then inserting the buffer in front of that moves all these selections to the end).

The pipe implementation needs to diff the input and output, and then to apply those edits to the buffer. We have the building blocks to generate the diff, what we need is the code that can apply that diff to a buffer.

lenormf:
I created issue #1501, which once resolved should make the changes this PR introduces relevant.

lenormf:
#1501 has been resolved, and the PR apparently works as expected without needing any change.

PR #1510, opened 2017-07-17
rc man: Avoid undefined behavior on `expr`

lenormf:
Looking up the man page for `index` was failing on systems using
GNU/coreutils. The `:man` command matched whatever page it was given with
the `expr` utility. This tool behaves as expected when it follows strictly
the POSIX standard but the GNU implementation introduces additional commands
(including `index`), about which the standard states:

```
The use of string arguments length, substr, index, or match produces unspecified results.
```

As a result, parsing the man page number is now implemented with pure
shell expansions, to avoid triggering an undefined behavior when the topic
searched is one of the keywords above.

### DIFF
--- a/rc/core/man.kak
+++ b/rc/core/man.kak
@@ -67,10 +67,13 @@ The page can be a word, or a word directly followed by a section number between 
     subject=${@-$kak_selection}
 
     ## The completion suggestions display the page number, strip them if present
-    pagenum=$(expr "$subject" : '.*(\([1-8].*\))')
-    if [ -n "$pagenum" ]; then
-        subject=${subject%%\(*}
-    fi
+    case "${subject}" in
+        *\([1-8]*\))
+            pagenum="${subject##*(}"
+            pagenum="${pagenum%)}"
+            subject="${subject%%(*}"
+            ;;
+    esac
 
     printf %s\\n "eval -collapse-jumps -try-client %opt{docsclient} man-impl $pagenum $subject"
 } }


doppioandante:
I don't know whether it makee sense in this case but have you tried setting POSIXLY_CORRECT before calling man?

lenormf:
The GNU implementation of `expr` doesn't even try to check for that environment variable, unfortunately (note that some other tools such as `echo` do try to though). This variable also seems to have been introduced by GNU, so using it would favor hinting at GNU tools whereas several implementations are available for POSIX tools (e.g. busybox).

It's probably less error-prone to use tools in a POSIX way straight from the beginning, their behavior will be deterministic that way.

rc/core/man.kak
```diff
@@ -67,10 +67,14 @@ The page can be a word, or a word directly followed by a section number between
...
 
     ## The completion suggestions display the page number, strip them if present
-    pagenum=$(expr "$subject" : '.*(\([1-8].*\))')
-    if [ -n "$pagenum" ]; then
-        subject=${subject%%\(*}
-    fi
+    case "${subject}" in
+        *\([1-8]*\))
+            pagenum="${subject##*(}"
+            pagenum="${subject%%*)}"
```
mawww: Suspicious line is suspcious

lenormf: Failed c/p, it's fixed now.

lenormf:
Fixed page number parsing, removed the default `*` case in the switch to save up one line.

PR #1519, opened 2017-07-27
Add more details in wrong_argument_count error message

Delapouite:
Hi

I often trigger by mistake `wrong_argument_count` errors, especially when typing custom commands where I don't remember (or know) the number of params and the docstring is not clear (or absent).

With this PR, the user can more distinctly see what's going on, the new message provides additional data:

`"wrong argument count - given: {}, expected: min {} max {}`


Delapouite:
I'll do another more general PR to homogenize error messages: some start with a capital letter, other don't …

Issue #745, opened 2016-07-26
Have an interface for the history tree, document it


lenormf:
Hi,

I edited the title of this issue to reflect the fact that it's not so much about documenting keys as it is about deciding what the interface should be and documenting how the whole undo-tree thing works.

lenormf:
I had a look at what other editors have, and I think vim's [gundo](https://github.com/sjl/gundo.vim) is pretty nice:

* interactive tree visualizer: unlike emacs' implementation, the changes are a stack, not a top-to-bottom list; there's always a left-aligned straight path (representing the current branch) from which the changes branch off as opposed to a centered tree
* diff printing: fancy syntax highlighted diffing to know what the changes do before they're applied

All this combined with tmux panes would be p gr8.

Delapouite:
I have started a draft here : https://github.com/mawww/kakoune/wiki/Undo-redo----History-tree-of-changes to better illustrate the difference between `u` and `<a-u>`.

This document deserves a bit of editing and when it will be mature enough, it will be turned into a proper `doc/manpages/undo-history.asciidoc`.

PR #1520, opened 2017-07-27
Add history-move command

Delapouite:
Hi

In this PR, I'd like to start thinking about the problematic exposed by @lenormf in this issue https://github.com/mawww/kakoune/issues/745 about having an interface for the history tree.

I think to build a tool such as gundo we must have way to:

1.  move through the history tree
2. query/dump info about the history tree

For point 1) we have almost all the primitives:

`u` and `U` let us move through the tree in child -> parent and parent -> child manner
`<a-u>` and `<a-U>`  let us move the through the tree by decrementing / incrementing the unique `change id`

But these commands only work in relative fashion (even with a `count`) and linearly.

The `history-move` command proposed in this PR let us move to an arbitrary `change id` in an absolute fashion.
I don't think it really deserves a dedicated key, so a command is fine in this case.
Currently, it does not follow kakoune classic naming scheme `verb-noun` because I didn't find a good name (don't hesitate to offer you suggestions) and it's a way to group all the future `history-*` commands under a namespace like `history-dump`, `history-list` etc… for point 2) in future PRs.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -705,6 +705,7 @@ command `q!` has to be used).
      `:%sh{ echo echo tchou }` will echo tchou in Kakoune, whereas
      `:nop %sh{ echo echo tchou }` will not, but both will execute the
      shell command.
+ * `history-move <change_id>`: move to change <change_id> in undo history
 
 Multiple commands
 ~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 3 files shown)


doppioandante:
One thing that puzzles me about this is that if we ever expose a GUI for the undo tree it would probably bypass these commands and it would exchange messages directly with the kakoune server, and we'd end up with some duplication.

Delapouite:
kakoune now provides a `kak_history_id`:
https://github.com/mawww/kakoune/commit/d4fa94c3564738f642027e352bdabb67f0bc2471

Therefore, the `history-move` command can be implemented through scripting. I'll try to link to an implementation soon.

PR #1523, opened 2017-08-01
Markdown indent fixes

Screwtapello:
When I write Markdown documents, I tend to use "Setext" header syntax, like this:

    Some Header
    -----------

When I put the cursor on the line of `-` and press `o` to start adding new text underneath the header, the current version of Kakoune assumes that all those `-` are list bullets, and so copies them to the next line:

    Some Header
    -----------
    -----------

...which isn't actually what I wanted.

This PR includes:

  - a commit that adds some more test-cases for adding newlines to Markdown documents (which I wrote to make sure I didn't regress anything)
  - a commit that adds test-cases like the example above, plus a change to `markdown.kak` that makes the tests pass
  - the required copyright waiver


mawww:
Thanks very much for this thorough pull request.

PR #1534, opened 2017-08-18
Improve Haskell highlighter

danr:
Import keywords are put in keyword face instead of meta face.
This leaves room for pragmas and macros to be in the meta face.

Operator keywords are put in keyword face too.

Finally, expression keywords are put in face attribute.

NB: Now with comments at start of line

### DIFF
--- a/rc/base/haskell.kak
+++ b/rc/base/haskell.kak
@@ -12,19 +12,61 @@ hook global BufCreate .*[.](hs) %{
 # ‾‾‾‾‾‾‾‾‾‾‾‾
 
 add-highlighter -group / regions -default code haskell \
-    string   %{(?<!')"} (?<!\\)(\\\\)*"      ''  \
-    comment  (--[^>])   $                    ''  \
-    comment \{-         -\}                  \{- \
-    macro   ^\h*?\K#    (?<!\\)\n            ''
+    string   '(?<!\'\\)(?<!\')"'         (?<!\\)(\\\\)*" ''   \
+    macro   ^\h*?\K#                     (?<!\\)\n       ''   \
+    pragma  \{-#                         \#-\}           \{-  \
+    comment \{-                            -\}           \{-  \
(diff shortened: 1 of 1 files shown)


rc/base/haskell.kak
```diff
@@ -12,19 +12,53 @@ hook global BufCreate .*[.](hs) %{
...
 add-highlighter -group /haskell/string  fill string
 add-highlighter -group /haskell/comment fill comment
+add-highlighter -group /haskell/pragma  fill meta
 add-highlighter -group /haskell/macro   fill meta
 
-add-highlighter -group /haskell/code regex \b(import)\b 0:meta
-add-highlighter -group /haskell/code regex \b(True|False)\b 0:value
-add-highlighter -group /haskell/code regex \b(as|case|class|data|default|deriving|do|else|hiding|if|in|infix|infixl|infixr|instance|let|module|newtype|of|qualified|then|type|where)\b 0:keyword
-add-highlighter -group /haskell/code regex \b(Int|Integer|Char|Bool|Float|Double|IO|Void|Addr|Array|String)\b 0:type
+add-highlighter -group /haskell/code regex (?<=\W)0x+[A-Fa-f0-9]+ 0:value
```
lenormf: Wouldn't `\b` be enough, instead of the look-behind?

danr: Yeah, it should be the same thing, right? I don't remember why I wrote it like this, do you want me to change it to `\b`?

lenormf: The simpler the better, unless you have a good reason for using the current expression. Also if we ever switch over to `std::regex`, we'll have to get rid of look-behinds since they're not implemented by it.

danr:
I realized due to your comment that `\b` cannot be used since `'` is valid in identifiers. Fixed this and some other tidbits.

lenormf:
Couldn't this be handled with `\b'?` then?

danr:
No that wouldn't work since I want it to _not_ match.

PR #1535, opened 2017-08-18
Show error when using go-jump and jq is not installed

blastrock:


lenormf:
The command will probably be executed regardless of whether `jq` is installed, since the hunk only prints a message and doesn't leave the execution flow.

Checking for the existence of a command in `PATH` should be done in a POSIX way, using `command -v … >/dev/null`.

This check will be done every time the `jump` command is executed, which is sub-optimal, if you need to check for dependencies I think it makes more sense to do it in a shell scope in the root namespace once.

Finally, the scripts generally don't check for dependencies, they are usually documented at the head of the file (or not ;)) and we let the commands fail at runtime if the user hasn't installed the dependencies required to run the script (errors are logged in the debug buffer, and the messages are usually enough to figure out what's wrong quickly). I'm not sure if the current lazy behavior should be kept, or if all scripts should check for non POSIX dependencies, as it would introduce a lot of superficial code in otherwise already complicated scripts.

blastrock:
> The command will probably be executed regardless of whether jq is installed, since the hunk only prints a message and doesn't leave the execution flow.

Right, I forgot a return there ><

> Checking for the existence of a command in PATH should be done in a POSIX way, using command -v … >/dev/null.

I can do it that way if you prefer. You talk about optimality though, such a test would fork a process while "which" is a shell built-in and (I hope) it takes no time to execute.

> This check will be done every time the jump command is executed, which is sub-optimal, if you need to check for dependencies I think it makes more sense to do it in a shell scope in the root namespace once.

I can put the check there, but isn't the file always sourced even when the user isn't editing go files? It feels weird to show an error when the user didn't ask for anything and doesn't know what is a go file.

> Finally, the scripts generally don't check for dependencies, they are usually documented at the head of the file (or not ;)) 

They are in this case, but I didn't read them :P 

> errors are logged in the debug buffer, and the messages are usually enough to figure out what's wrong quickly

This script fails silently though, it took me some time to understand how to make it work, hence this patch.

I think the silent fail is due to the fact that commands are run like "printf blabla | kak -p $kak_session", and the whole %sh{} block is redirected to /dev/null at the end. I don't know why it is done like this. Can't the script just print the commands on stdout for kakoune to run them?

lenormf:
>I can do it that way if you prefer. You talk about optimality though, such a test would fork a process while "which" is a shell built-in and (I hope) it takes no time to execute.

`which` isn't defined by the standard, which means that some shells might not define it, so we've been using `command` in all scripts (which is [defined by POSIX](http://pubs.opengroup.org/onlinepubs/009696899/utilities/command.html)).

>I can put the check there, but isn't the file always sourced even when the user isn't editing go files? It feels weird to show an error when the user didn't ask for anything and doesn't know what is a go file.

There are two paradigms possible, either the scripts are all loaded unconditionally off of the usual `/usr/share/kak` directory (might vary depending on your distribution), or the current user has a `${XDG_CONFIG_HOME}/autoload` directory which contains the scripts (or symbolic links to the scripts) they have chosen to load at startup. I was mostly thinking of the second case when I made my comment, but I think you can compromise the two cases by simply printing an error on the debug buffer (`echo -debug Dependency unmet: jq` or something along those lines). That way the message remains hidden to users who don't care about installing `jq`, but available to Go users.

>I think the silent fail is due to the fact that commands are run like "printf blabla | kak -p $kak_session", and the whole %sh{} block is redirected to /dev/null at the end. I don't know why it is done like this. Can't the script just print the commands on stdout for kakoune to run them?

It's a [shell pattern](https://github.com/mawww/kakoune/blob/master/doc/writing_scripts.asciidoc#running-a-process-in-the-background) we use to run potentially time consuming commands in the background, we have to send commands (and eventually error messages) to `kak -p`, because the output of this parallel execution flow is redirected to `/dev/null`.

blastrock:
Fair enough, I'll use command -v and put the lines at init time.

Thanks for all the explanation and links, that was useful :)

blastrock:
I'm still a bit sad that go-jump will continue to fail silently, but this is better than nothing.

BTW, if I may ask another question, about the ```(...) > /dev/null &``` pattern, wouldn't it be better to do something like ```(...) | kak -p $kak_session &```? So that if a line fails in the script, the output will not go unnoticed.

lenormf:
The pattern used allows the command to be sent to the background and more importantly to not block the termination of the parent shell. With a regular `&` pipeline, the shell scope will hang until the command has ended, which creates a significant lag.

PR #1537, opened 2017-08-21
doc faq: Document the expansion of shell scopes

lenormf:


doc/manpages/faq.asciidoc
```diff
@@ -133,6 +133,16 @@ The shell scope allows users to spawn any interpreter they want, for a minimal
 cost in terms of performance, it is therefore the reason why it's the only
 one available by default.
 
+What shell is used to expand `%sh{}` scopes ?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The server expands shell scopes using `/bin/sh` (which is, in most
```
mawww: Thats not strictly true, we go through the dirs specified by the `_CS_PATH` confstr (see `man confstr`) and find the first one that contains `sh`. This is *guaranteed* by posix to be a posix shell, its probably going to be /bin/sh on almost all distribution, but it seems that posix does not let us assume that `/bin/sh` is going to be a posix compliant shell.

lenormf: I just tried to show that `/bin/sh` isn't necessarily a shell by itself (as is the case in FreeBSD and OpenBSD), but can be a link to another shell (as is the case in Archlinux, Debian, Gentoo).

I could have sworn there was a hardcoded `/bin/sh` in the shell manager at some point, which mislead me, so I need to rewrite this entire answer.

mawww: It used to be that, then we discovered that `/bin/sh` was not stricly guaranteed to be posix, that the only guaranteed way to get a posix tool on a posix system was through that `confstr` call. I think that changed around a year ago.

lenormf:
I rewrote the answer, and didn't mention any implementation detail to keep it generic.

PR #1539, opened 2017-08-23
travis: Make sure that all autoload scripts load correctly

doppioandante:
I think this could be useful in case something gets changed in kak script and breaks existing stuff.
I've tested this locally by taunting a .kak file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,3 +51,9 @@ addons:
     branch_pattern: coverity-scan
 
 script: cd src && make && make test
+script:
+    - cd src
+    - echo $(pwd)
+    - ./kak -n -e "buffer *debug*; write clean_debug_buffer; quit"
+    - ./kak -e "buffer *debug*; write debug_buffer; quit"
+    - cmp debug_buffer clean_debug_buffer


.travis.yml
```diff
@@ -51,3 +51,4 @@ addons:
     branch_pattern: coverity-scan
 
 script: cd src && make && make test
+script: ./kak -e "buffer *debug*; write debug_buffer\n; quit" && echo "*** This is the debug buffer, where debug info will be written ***\n" | cmp debug_buffer -
```
mawww: I think this should use printf if you want to interpret `\n`

lenormf: This line is also depending on some string in the source code that might change or be removed in the future, I don't think this is the best way of testing whether all scripts have been loaded.

Also when the test suite is run, the contents of the debug buffer will be displayed, which will show that some scripts haven't been loaded, so we might not need to run this pipeline at all.

doppioandante: I could change it so that it compares it with the content of the debug buffer of kakoune run with -n. But it's not a big deal to change the message string here.
AFAIR, the test suite runs without loading external script (I tried to fix this in the past but it didn't land), I will check.

.travis.yml
```diff
@@ -51,3 +51,9 @@ addons:
     branch_pattern: coverity-scan
 
 script: cd src && make && make test
+script:
+    - cd src
+    - echo $(pwd)
```
lenormf: `pwd`.

doppioandante:
If somebody more familiar than me with travis could chime in on why the executable is not found I would be very happy. I hope the result of the previous make is not thrown away.
Otherwise I will look at this again in a few days.

PR #1542, opened 2017-08-27
Docs: Fix wording in goto commands (h,l,i)

nedp:
Hi,

Currently the docs said `gh`, `gl`, and `gi` "select to" their target; as far as I can tell they in fact "go to" their target just like the other goto commands.

This PR updates the wording for those commands to make them consistent.


mawww:
Thanks for that, they changed behaviour long time ago, but seems we never updated the doc.

PR #1547, opened 2017-08-28
Adding Ubuntu TIP to the README

ZehCnaS34:


mawww:
Hello, could you rebase that branch to get rid of that merge commit ?

lenormf:
As far as I know, `libboost-dev` isn't needed, and you can get away with only installing `asciidoc-base` to generate the documentation. Also that final command that installs the editor to `~/.local` seems very specific to you need, maybe this should be kept generic with a simple `sudo make intall`.

Screwtapello:
I haven't tried to build kakoune on Ubuntu, but I do use it on Debian.

For me, `libboost-regex-dev` depends on `libboost-regex1.62-dev` which depends on `libboost1.62-dev` so yeah, the extra `libboost-dev` dependency is unnecessary.

Installing `asciidoc-base` instead of `asciidoc` seems to work, but it does also require `xsltproc`. Easier to install `xsltproc` than the hundreds of megabytes of TeX that asciidoc normally depends on, though.

When you `make install`, you probably should add `debug=no` so that you get an optimized build instead of a slow debug build.

Regarding installing to `~/.local`, that's the standard place to install things you don't want to install system wide, but given that all the other install-instruction tips install Kakoune system-wide and require root access, maybe `sudo make install` isn't such a terrible thing.

lenormf:
>Installing asciidoc-base instead of asciidoc seems to work, but it does also require xsltproc. Easier to install xsltproc than the hundreds of megabytes of TeX that asciidoc normally depends on, though.

I think `a2x` is provided by the `asciidoc-base` package, which is why we need it.

>When you make install, you probably should add debug=no so that you get an optimized build instead of a slow debug build.

We want users who compile from source to use a debug build by default, so that they can include a readable stacktrace in their issue if needed.

PR #1548, opened 2017-08-29
Check for gocode, goimports and gogetdoc on go-tools loading

blastrock:
These tools also need to be installed and don't come with go's default installation.

### DIFF
--- a/rc/extra/go-tools.kak
+++ b/rc/extra/go-tools.kak
@@ -6,9 +6,11 @@
 # - jq for json deserializaton
 
 %sh{
-    if ! jq --version > /dev/null 2>&1; then
-        echo 'echo -debug %{Dependency unmet: jq, please install it to use go-tools}'
-    fi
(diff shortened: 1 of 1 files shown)


lenormf:
Could you factorize all that into a `for` loop? Also the error code check seems weird, I think it's easier to check for a command's presence in the `PATH` using `command -v … >/dev/null`.

blastrock:
Actually they don't have a -v flag or a -help flag or anything. Giving -v triggers the "unknown argument" error which gives error code 2. I can't find a way to make these command not trigger an error without using them on real code.

Ok for the for loop, I'll do it later.

lenormf:
`command` is an actual command defined by POSIX that allows you to test whether a given program is available in the `PATH`, e.g. `command -v jq >/dev/null && echo OK || echo KO`.

blastrock:
Oh! I didn't understand that, my bad! Fixed.

PR #1553, opened 2017-09-02
Add current history id to env variables

Franciman:
The curr_history_id env variable could help implementing LSP's textDocument/didChange more efficiently
because we could track all the changes since last sync point and only send those changes instead of sending the whole text document every time.

Maybe also #745 could benefit from this?
Also related to #1520

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -792,7 +792,7 @@ Some of Kakoune state is available through environment variables:
  * `kak_buf_line_count`: the current buffer line count
  * `kak_timestamp`: timestamp of the current buffer, the timestamp is an
        integer value which is incremented each time the buffer is modified.
- * `kak_curr_history_id`: history id of the current buffer, the history id is an integer value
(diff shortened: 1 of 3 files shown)


Delapouite:
I like this PR.

You're right, introducing this var would provide a less intrusive alternative to my proposal about a new `history-move` command in https://github.com/mawww/kakoune/pull/1520.
This command could be implemented in kak scripts with `<a-u>` and `<a-U>` combined with `curr_history_id` with simple arithmetic and a `count`.

In the doc we could put more emphasize about the difference between `kak_curr_history_id` and `kak_timestamp` which *may* confuse some users.

mawww:
I like this change, however we are moving away from abbrevations in general in Kakoune naming, could you rename the variable to `kak_current_history_id`, or more simply `kak_history_id` as its implicit in most variable names that they refer to the current state.

Franciman:
Ok, I fixed the naming.
@Delapouite I agree the current description is confusing. Hope we'll find a clearer way to explain it.

PR #1560, opened 2017-09-08
rc: Create a face for every hardcoded color

lenormf:
The `make.kak` script still has hardcoded colors in it because I have no idea what data they highlight.

I removed the highlighter for some keywords that are not language-specific in comments in `swift.kak`.

HTH.


lenormf:
I think some implementations of `make` point at the exact location in a line that wasn't parsed properly, as compilers do, which was what that regex was highlighting.

Rebased on c702c915c1e3ad2df354f38132645a72c5d577f2.

mawww:
I do not like the explosion of very specific faces this leads to.

I think various hard coded colors should use generic, already existing faces (for the language support files), and we should try to merge together some others (we do not need separate grep and make faces for filenames, line count, column count, they should be the same face).

Could you try to make those changes, avoiding introducing new faces as much as possible ?

lenormf:
Why is having many faces a problem, if they are all contained within the file they are used in? They all have a prefix as well, so they are kind of sorted by scope.

It's definitely simpler to use already existing faces, because it's just simpler first and second because every buffer is highlighted with colors within the same palette; but it takes away ways of customization that could be achieved for a minimal cost (c.f. above).

Another potential issue that could arise is for data that's supposed to be highlighted with a specific color that doesn't have any face that's similar to it already declared in the theme.

mawww:
My reasoning is that the main benefit of naming the face is that colorschemes can define it. As a colorscheme writer, I'd much prefer to have a small set of reused faces (say "ColumnNumber", "Filename", "LineNumber") to define rather than needing to duplicate those for every script that define its own (say "ScriptColumnNumber", "ScriptFilename", "ScriptLineNumber"...).

I see introducing a new face as a message to theme writer: please add this color to your themes. Hence I think it should be done only when there is a very clear need for it, and we should try to provide generic ones as much as possible.

For example, admitting we want to have colorscheme control over the color of added line/removed line in diffs, we should probably have a reusable face, not linked to diff, that can be used for added/removed files in git status files. That would make colorscheme much more usefull, as I think it makes much more sense for a scheme to contain a general face (Lets call it "DiffAdded", although I am not entirely satisfied with that name) that will get used by both git, diff, and other eventual scripts that want to colorize text that represent an "addition" concept.

lenormf:
Rebase on 9b216e0e79612267b4a3065615e6fb71f8fec7f6, the new faces have been removed.

mawww:
I think we misunderstood each other, I actually liked the introduction of some faces, namely for filenames, line number and column number, diff add/remove. What I wanted was for those faces to be shared between scripts rather than having one specific face for filenames in grep results, another one for filenames in make results...

For the other ones, I think using existing faces as you did is the right call.

lenormf:
About the `diff` example specifically, buffers that display patches (e.g. `git-commit` in verbose mode) would typically add the `diff` highlighter to the window, and never need to access the faces directly (they would be encapsulated), so I didn't add any face.

As for the other ones, wouldn't that create a hard dependency between scripts, only to get access to a given face? e.g. `grep.kak` would require `make.kak` to be loaded, in order to access the `filename` face declared in the latter. Unless you want to those to be promoted as defaults?

mawww:
The same faces faces can be defined in various scripts, as the command `face` both declare and sets the value of a face. So we can just have the different scripts declare the face they will use, this will not introduce conflicts or dependencies between scripts.

lenormf:
We would have to remember to change all the declarations of the same face in separate files if the default color pair was to change, it sounds inconvenient.

mawww:
Its not perfect, I agree, but if we want colorscheme to support those faces, we need to limit their number, it still looks better to me to have a single shared face for  this shared concept, potentially blessing it into default.kak once it gets popular/used enough, than the two other alternatives I see (different name in each script, or hard coded color)

lenormf:
I re-worked the PR to split it into more commits and rebase on `master`.

andreyorst:
I've also noticed that `diff show-diff` command adds a flag column that has background color that doesn't match theme background.

PR #1564, opened 2017-09-09
docs: add command aliases

Delapouite:
Hi

This PR also fixes things like `write-all-quit` not having a `!` version.

Also I don't know why the beginning of the document has bold and not the end in the Github interface

### DIFF
--- a/doc/manpages/commands.asciidoc
+++ b/doc/manpages/commands.asciidoc
@@ -10,51 +10,62 @@ Primitives
 
 Some commands take an exclamation mark (*!*), which can be used to force
 the execution of the command (i.e. to quit a modified buffer, the
-command *q!* has to be used).
+command *q!* has to be used). Aliases are mentionned below each commands.
 
-*cd* [<directory>]::
+*change-directory* [<directory>]::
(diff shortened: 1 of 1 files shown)


mawww:
I'd prefer them to be written in another way, here it makes the rendered document look funny with two lines for each section titles. Maybe just put the aliases in parens next to the full command name.

lenormf:
I faced this issue as well when auto-generating documentation from the built-in scripts, and I fixed it by adding a small subsection to all commands containing aliases (and default value for options):

```
*command* [param]::
    *aliases* cmd
    This command does things.
```

Adding the alias on the same line clutters a lot, writing it in the description's paragraph takes it away slightly from the command (because of the indentation), and writing it right below the command looks weird - I don't know how it could be done nicely with the tools we use.

Delapouite:
Yes. You are right, the previous version was not satisfying.

In this new attempt I followed lenormf layout, but asciidoc requires an extra `+` at the end of the line to avoid weird indentation: http://asciidoctor.org/docs/asciidoc-writers-guide/#list-continuation

PR #1567, opened 2017-09-11
doc faq: Document how to fix the "insert mode escape lag"

lenormf:


lenormf:
Actually I think that setting this option for the entire `tmux` session is not good practice, using a hook that sets this option in the current `tmux` window seems more appropriate.

Screwtapello:
Why is it not good practice to set the option globally? The "correct" value largely depends on the network latency and packet-loss between the user and the tmux client, and that's going to be the same for all windows in the tmux session.

You may also wish to look at [the vis FAQ for the same question](https://github.com/martanne/vis/wiki/FAQ#m-key-is-inserted-whenever-i-press-escape-and-key-in-short-succession) and see if there's any details you want to copy across to the Kakoune FAQ.

lenormf:
Users who rely on this behavior to have time to do complicated meta sequences will have the carpet pulled under them if they blindly set the `escape-time` option for the whole session. I couldn't get anything to work in a more elegant way anyway, so I guess this will have to do.

PR #1569, opened 2017-09-12
Add Context::clear_status()

Delapouite:
Hello

Through the `input_handler` source, two slightly different ways to clear the status are used:
`context().print_status({});`
`context().print_status(DisplayLine{});`

In order to better convey the meaning, what about homogenizing both forms in a dedicated `clear_status()` function?

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -75,6 +75,11 @@ void Context::print_status(DisplayLine status, bool immediate) const
         client().print_status(std::move(status), immediate);
 }
 
+void Context::clear_status() const
+{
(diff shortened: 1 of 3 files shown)


mawww:
`context().print_status({})` and `context().print_status(DisplayLine{})` are the same, in one we just let C++ deduction to know we are constructing a DisplayLine, in the other we specify that explicitely, I favor the first style, but might have used the other in moments of inattention, I dont think we need an explicit `clear_status` method here, but I will welcome the removal of redundant `DisplayLine` specifiers.

Issue #675, opened 2016-05-14
Expose information about the actually usable buffer space in scripts


lenormf:
Hi,

An issue I noticed while using the `:man` command is that the man pages are always too large for the kakoune buffer to fit in when the `number_lines` highlighter is active (there's always a couple characters out of the screen that I need to `<a-l>` to). That's because the `man` executable is given the total width of the window (`window_width`), instead of the width of the area that will hold the data: the width passed to `man` has to be `window_width` minus the width that the `number_lines` highlighter takes up on the screen (including the seperator).

As such this rendering issue can be solved (at least) two ways: either highlighters are exposed along with the options that they were passed, or another default variable is created that holds the coordinate of the top-left corner of the "rendering area" (for lack of a better term), as well as its width and height. Either way, the `:man` command would be able to pass the `man` binary a more accurate width, and avoid off-screen litter.

Thanks.

Delapouite:
Hi.

I tried to expose the data needed in this PR: https://github.com/mawww/kakoune/pull/1570
The new `window_range_column` *seems* to do the job.

PR #1570, opened 2017-09-12
Add window_{top,left,range_column,range_line} vars

Delapouite:
Hello

This PR is an attempt at solving https://github.com/mawww/kakoune/issues/675

The naming of `window_range_*` is not that great, but it follows the internals.

Also while displaying the value of `window_left` in my status line, I found this value to be a bit *flacky* moving from `0` to `1` randomly with no apparent reason.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -810,6 +810,10 @@ Some of Kakoune state is available through environment variables:
  * `kak_cursor_byte_offset`: offset of the main selection from the beginning of the buffer (in byte).
  * `kak_window_width`: width of the current kakoune window
  * `kak_window_height`: height of the current kakoune window
+ * `kak_window_top`: top offset of the current kakoune window
+ * `kak_window_left`: left offset of the current kakoune window
+ * `kak_window_range_column`: width (without line numbers) of the current kakoune window
+ * `kak_window_range_line`: height (without line numbers) of the current kakoune window
  * `kak_hook_param`: filtering text passed to the currently executing hook
  * `kak_hook_param_capture_N`: text captured by the hook filter regex capture N
  * `kak_client_env_<name>`: value of the <name> variable in the client environment.
--- a/doc/manpages/expansions.asciidoc
+++ b/doc/manpages/expansions.asciidoc
@@ -110,6 +110,14 @@ informations about Kakoune's state:
 	width of the current kakoune window
 *kak_window_height*::
 	height of the current kakoune window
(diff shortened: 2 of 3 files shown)


danr:
You can calculate `window_top` this way: 

```
decl int window_top
def update_window_top %{
    eval -save-regs ct %{
        eval -draft -no-hooks %{
            reg c %val{cursor_line}
            exec gt
            reg t %val{cursor_line}
        }
        set window window_top %sh{ echo $(($kak_reg_c - $kak_reg_t)) }
    }
}
```

By pressing `gb` instead you could get the bottom line of the window.

Note that we have the `window_height` value availabe too.

I don't know of any way of getting the left and right columns since there are no corresponding goto commands.

Delapouite:
Thanks! Yes you're right about `window_top`.
The initial demand was more about exposing the `window_range_*` ones, but after looking at the code, it felt no harm to also exposed the other two so it can avoid this `eval -draft exec` dance.

lenormf:
It's simpler to expose variables manage by the editor directly, to handle updates due to e.g. terminal resizing.

Note that the above snippet won't work properly as `gt` will take `scrolloff` into account, making the window offset incorrect if context lines are displayed above the selection upon hitting `gt`.

My original issue mentioned the offset of the code-rendering area and its length to be able to know where on the terminal it is located, which boils down to the current part of the buffer that is show minus the statusbar (and other widgets displayed), as well as any highlighter added to the current window.

danr:
The snippet works with `scrolloff`, it simply looks at the cursor line.

lenormf:
If `window_top` is supposed to be how far away the top-most line currently displayed is from the first line, then `cursor_line` will be off by as many lines are set in `scrolloff`.

If it's supposed to be the offset on the terminal of the buffer rendering area compared to the top-most border of the terminal, then it's incorrect.

mawww:
I think it makes sense to expose that data, but I am not really keen on adding 4 new env vars just for that. I am wondering if it would be better to expose the window range as a single var, something close to the selection syntax (say `<pos line>.<pos column>,<height>.<width>`). Not sure if thats really better as it makes accessing the data a bit more painful (as you now need to go through `cut` or similar to get the data).

Yet another good reason to find a way to expose structured data more nicely to the shell...

Delapouite:
Closing because `window.range()` was removed in https://github.com/mawww/kakoune/commit/0bdde

PR #1579, opened 2017-09-19
Add max_history_id in status printed with <a-u> and <a-U>

Delapouite:
Hi

Follow up to the work done on https://github.com/mawww/kakoune/pull/1553

This PR adds valuable info about how *far* we are from the original / final version of a buffer while using successive `<a-u>` and `<a-U>`, which is handy to better combine with a `count`.

The format `#id (total)` follows the one recommended in this PR about selection index: https://github.com/mawww/kakoune/pull/1500

### DIFF
--- a/src/buffer.hh
+++ b/src/buffer.hh
@@ -142,6 +142,7 @@ public:
     bool           redo(size_t count = 1) noexcept;
     bool           move_to(size_t history_id) noexcept;
     size_t         current_history_id() const noexcept;
+    size_t         next_history_id() const noexcept { return m_next_history_id; }
 
     String         string(BufferCoord begin, BufferCoord end) const;
     StringView     substr(BufferCoord begin, BufferCoord end) const;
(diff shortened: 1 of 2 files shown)


src/normal.cc
```diff
@@ -1770,18 +1770,21 @@ void move_in_history(Context& context, NormalParams params)
     const int count = std::max(1, params.count);
     const int history_id = (int)buffer.current_history_id() +
         (direction == Direction::Forward ? count : -count);
+    const int total_history_id = (int)buffer.next_history_id() - 1;
```
mawww: I think `max_history_id` would be a better name here

src/buffer.hh
```diff
@@ -142,6 +142,7 @@ public:
     bool           redo(size_t count = 1) noexcept;
     bool           move_to(size_t history_id) noexcept;
     size_t         current_history_id() const noexcept;
+    size_t         next_history_id() const { return m_next_history_id; }
```
mawww: `noexcept` as well please

Delapouite:
Done. Thanks again for your help and advice.

PR #1583, opened 2017-09-21
Add count support for indent / deindent

Delapouite:


src/normal.cc
```diff
@@ -1054,12 +1055,14 @@ void indent(Context& context, NormalParams)
...
 template<bool deindent_incomplete = true>
-void deindent(Context& context, NormalParams)
+void deindent(Context& context, NormalParams params)
 {
+    ColumnCount count = params.count ? params.count : 1;
     ColumnCount tabstop = context.options()["tabstop"].get<int>();
     ColumnCount indent_width = context.options()["indentwidth"].get<int>();
     if (indent_width == 0)
         indent_width = tabstop;
+    indent_width = indent_width * count;
```
lenormf: You can use the `*=` operator to simplify the line, also this has a big potential for integer overflow.

Issue #1584, opened 2017-09-22
lua-insert-on-new-line puts comment delimiters within text


exitium:
If you type a Lua comment like so:
```lua
-- First comment. Second comment.
```
And then split it by inserting a new line, the comment delimiter is put within the text:
```lua
Sec-- ond comment.
```
Instead of the expected:
```lua
-- First comment.
-- Second comment.
```

This problem occurs with both single line and multi-line comments.

PR #1585, opened 2017-09-22
Fix Lua comment delimiter insertion #1584

exitium:


exitium:
I replaced 'John Doe' with my username. If that is unacceptable then I will close my pull request and allow someone else to make the fix.

mawww:
No problem with using your nickname. Thanks for that change.

PR #1593, opened 2017-09-26
Add selections_desc format to select autoinfo

Delapouite:
Hi

For no apparent reasons, I struggle every time I have to use the `select` command.
My mind never remembers which version is correct `x.y,x2.y2` or `x,y.x2,y2`

The correct format is available in `:doc commands` but I think it's useful to also have it in the autoinfo, which is displayed right when we need it.
The same strategy is already used for the `set-face` command where the format is also mentioned in the autoinfo.


lenormf:
I don't think data implementation formats should be documented in more than one place (the documentation pages), if they come to change as has happened a few times in the past, the places that reference the old format will be forgotten and mislead users.

PR #1599, opened 2017-09-28
Do not restore backup files if they are older than actual file.

nochiel:
Old backup files might be in a directory for myriad reasons.  Restoring them will overwrite a more recently edited file.  Don't restore them but report that they were found.

### DIFF
--- a/rc/extra/autorestore.kak
+++ b/rc/extra/autorestore.kak
@@ -7,20 +7,29 @@ def autorestore-restore-buffer -docstring "Restore the backup for the current fi
         buffer_basename="${kak_buffile##*/}"
         buffer_dirname=$(dirname "${kak_buffile}")
 
-        ## Find the name of the latest backup created for the buffer that was open
-        ## NB. kak doesn't automatically cd to current buffer_dirname.
-        buffer_backup_glob="${buffer_dirname}/.${buffer_basename}.kak.*"
-        backup_path=$(ls -1t ${buffer_backup_glob} 2>/dev/null | head -n 1)
+        if [ -f "${kak_buffile}" ]; then
+            newer=$(find "${buffer_dirname}"/".${buffer_basename}.kak."* -newer "${kak_buffile}" -exec ls -1t {} + 2>/dev/null | head -n 1)
 
-        if [ -z "${backup_path}" ]; then
-            exit
+            older=$(find "${buffer_dirname}"/".${buffer_basename}.kak."* \! -newer "${kak_buffile}" -exec ls -1t {} + 2>/dev/null | head -n 1)
+        else
+            # New buffers that were never written to disk.
+            newer=$(ls -1t "${buffer_dirname}"/".${buffer_basename}.kak."* 2>/dev/null | head -n 1)
+        fi
+
+        if [ -z "${newer}" ]; then
+            if [ -n "${older}" ]; then
+                printf %s\\n "
+                    echo -debug Old backup file(s) found: will not restore ${older} .
+                "
+            fi
+            exit   
         fi
 
         printf %s\\n "
             ## Replace the content of the buffer with the content of the backup file
-            echo -debug \"Restoring file: ${backup_path}\" 
+            echo -debug Restoring file: ${newer}
 
-            exec -draft %{ %d!cat<space>${backup_path}<ret>d }
+            exec -draft %{ %d!cat<space>\"${newer}\"<ret>d }
 
             ## If the backup file has to be removed, issue the command once
             ## the current buffer has been saved
@@ -28,10 +37,9 @@ def autorestore-restore-buffer -docstring "Restore the backup for the current fi
             ## buffer was restored, do not remove the backup
             hook -group autorestore buffer BufWritePost '${kak_buffile}' %{
                 nop %sh{
-                    if [ \"\${kak_opt_autorestore_purge_restored}\" = true ]; then
-                        ls -1 '${buffer_dirname}'/.'${buffer_basename}'.kak.* 2>/dev/null | while read -r f; do
-                            rm -f \"\${f}\"
-                        done
+                    if [ \"\${kak_opt_autorestore_purge_restored}\" = true ];
+                    then
+                        rm -f \"${buffer_dirname}/.${buffer_basename}.kak.\"* 
                     fi
                 }
             }
@@ -41,19 +49,18 @@ def autorestore-restore-buffer -docstring "Restore the backup for the current fi
 
 ## Remove all the backups that have been created for the current buffer
 def autorestore-purge-backups -docstring "Remove all the backups of the current buffer" %{
-    nop %sh{
-        if [ ! -f "${kak_buffile}" ]; then
-            exit
-        fi
+    %sh{
+        [ ! -f "${kak_buffile}" ] && exit
 
         buffer_basename="${kak_bufname##*/}"
         buffer_dirname=$(dirname "${kak_bufname}")
 
-        ls -1 "${buffer_dirname}"/."${buffer_basename}".kak.* 2>/dev/null | while read -r f; do
-            rm -f "${f}"
-        done
+        rm -f "${buffer_dirname}/.${buffer_basename}.kak."*
+
+        printf %s\\n "
+            echo -markup {Information}Backup files removed.
+            "
     }
-    echo -markup '{Information}Backup files removed'
 }
 
 ## If for some reason, backup files need to be ignored


rc/extra/autorestore.kak
```diff
@@ -14,6 +14,9 @@ def autorestore-restore-buffer -docstring "Restore the backup for the current fi
 
         if [ -z "${backup_path}" ]; then
             exit
+        elif [ ${kak_buffile} -nt ${backup_path} ]; then
```
mawww: Unfortunately, `-nt` is not posix... IIRC you need to use `find ... -newer $kak_buffile` to do the same.

This also reminds me why we were using `backup_path=$(ls -1t ."${kak_bufname}".kak.* 2>/dev/null | head -n 1)` to get the candidate backup buffer. So we'd need to fix that as well (get the most recent backup file).

rc/extra/autorestore.kak
```diff
@@ -9,32 +9,36 @@ def autorestore-restore-buffer -docstring "Restore the backup for the current fi
 
         ## Find the name of the latest backup created for the buffer that was open
         ## NB. kak doesn't automatically cd to current buffer_dirname.
-        buffer_backup_glob="${buffer_dirname}/.${buffer_basename}.kak.*"
-        backup_path=$(ls -1t ${buffer_backup_glob} 2>/dev/null | head -n 1)
+        buffer_backup_glob=".${buffer_basename}.kak.*"
+#        backup_path=$(ls -1t ${buffer_backup_glob} 2>/dev/null | head -n 1)
```
lenormf: The comment should go, the line is already in versioning.

rc/extra/autorestore.kak
```diff
@@ -9,32 +9,36 @@ def autorestore-restore-buffer -docstring "Restore the backup for the current fi
...
+        newer=$(find "${buffer_dirname}" -maxdepth 1 -name "${buffer_backup_glob}" -newer "${kak_buffile}" 2>/dev/null | head -n1)
+        older=$(find "${buffer_dirname}" -maxdepth 1 -name "${buffer_backup_glob}" \! -newer "${kak_buffile}" 2>/dev/null | head -n1)
```
lenormf: Unfortunately, `-maxdepth` isn't [POSIX](http://pubs.opengroup.org/onlinepubs/009695399/utilities/find.html) which is why we usually use `ls -1`, in that case.

rc/extra/autorestore.kak
```diff
@@ -9,32 +9,36 @@ def autorestore-restore-buffer -docstring "Restore the backup for the current fi
...
-        if [ -z "${backup_path}" ]; then
-            exit
-        elif [ ${kak_buffile} -nt ${backup_path} ]; then
-          printf "%s\\n" "echo -debug \"Old backup file found: ${backup_path}; will not restore.\""
-          exit
+
+        if [ -z "${newer}" ];
+        then
+          if [ -n "${older}" ];
+          then
```
lenormf: You could save a few lines here by writing `if [ -z "${newer}" ] && [ -z "${older}" ]; then`

rc/extra/autorestore.kak
```diff
@@ -9,32 +9,36 @@ def autorestore-restore-buffer -docstring "Restore the backup for the current fi
...
             ## buffer was restored, do not remove the backup
             hook -group autorestore buffer BufWritePost '${kak_buffile}' %{
                 nop %sh{
-                    if [ \"\${kak_opt_autorestore_purge_restored}\" = true ]; then
-                        ls -1 '${buffer_dirname}'/.'${buffer_basename}'.kak.* 2>/dev/null | while read -r f; do
-                            rm -f \"\${f}\"
-                        done
+                    if [ \"\${kak_opt_autorestore_purge_restored}\" = true ];
+                    then
+                      rm \"${buffer_dirname}/.${buffer_basename}.kak.\"* 2>/dev/null
```
lenormf: Instead of redirecting `stderr` to `/dev/null`, use the `-f` flag that won't print any error, it's more elegant.

rc/extra/autorestore.kak
```diff
@@ -44,19 +48,20 @@ def autorestore-restore-buffer -docstring "Restore the backup for the current fi
...
 def autorestore-purge-backups -docstring "Remove all the backups of the current buffer" %{
-    nop %sh{
+    %sh{
         if [ ! -f "${kak_buffile}" ]; then
             exit
         fi
 
         buffer_basename="${kak_bufname##*/}"
         buffer_dirname=$(dirname "${kak_bufname}")
+        log=$(rm -v "${buffer_dirname}/.${buffer_basename}.kak."* 2>&1)
```
lenormf: `rm -v` isn't [POSIX](http://pubs.opengroup.org/onlinepubs/009695399/utilities/rm.html), and you can use `-f` to avoid error messages.

rc/extra/autorestore.kak
```diff
@@ -44,19 +48,20 @@ def autorestore-restore-buffer -docstring "Restore the backup for the current fi
...
 
-        ls -1 "${buffer_dirname}"/."${buffer_basename}".kak.* 2>/dev/null | while read -r f; do
-            rm -f "${f}"
-        done
+        printf %s\\n "
+          echo -markup \"{Information}Backup files removed\"
+          echo -debug \"Removed backup files: ${log}\"
```
lenormf: Protip: the `echo` command will join all of its arguments with a single space, so you don't need to escape the message, e.g. `echo -debug Removed backup files: ${log}` (no quotes needed).

mawww:
Could you as well rebase the PR to get rid of that merge commit ?

mawww:
And I dont think we have a copyright waiver commit from you, could you add one (see the CONTRIBUTING file)

rc/extra/autorestore.kak
nochiel: Sorry about that. Old habits. I'm still trying to get used to posix.

rc/extra/autorestore.kak
nochiel: Cases:
1. no backup file.
2. no newer backup file; no older backup file.
3. no newer backup file; 1 or more older backup files.

rc/extra/autorestore.kak
nochiel: Nice. But can I write, e.g., `echo -debug foo \n bar` ?

mawww:
Could you squash all the non-copyright-waiver commits together in a single one ?

rc/extra/autorestore.kak
```diff
@@ -7,31 +7,42 @@ def autorestore-restore-buffer -docstring "Restore the backup for the current fi
...
 
-        ## Find the name of the latest backup created for the buffer that was open
-        ## NB. kak doesn't automatically cd to current buffer_dirname.
-        buffer_backup_glob="${buffer_dirname}/.${buffer_basename}.kak.*"
-        backup_path=$(ls -1t ${buffer_backup_glob} 2>/dev/null | head -n 1)
+        buffer_backup_glob=".${buffer_basename}.kak.*"
 
-        if [ -z "${backup_path}" ]; then
-            exit
+       	if [ -f "${kak_buffile}" ]; then
```
mawww: is that a tab used for indentation ?

We use 4 spaces for Kakoune.

rc/extra/autorestore.kak
```diff
@@ -7,31 +7,42 @@ def autorestore-restore-buffer -docstring "Restore the backup for the current fi
...
+       		newer=$(find "${buffer_dirname}" \! -path "${buffer_dirname}" -prune -name "${buffer_backup_glob}" -newer "${kak_buffile}" 2>/dev/null | head -n1)
```
mawww: I think instead of `\! -path "${buffer_dirname}"` it would be clearer to use `-type f`. Also, I dont think find guarantees the first path printed is going to be the newest. Unfortunately I cant find a way to tell posix find to do that, so I think you'd still need to go through `ls -t -1`.

I'd guess something like `find ... | xargs ls -t -1 | head -n1` will do.

rc/extra/autorestore.kak
```diff
@@ -41,19 +52,19 @@ def autorestore-restore-buffer -docstring "Restore the backup for the current fi
...
         buffer_dirname=$(dirname "${kak_bufname}")
 
-        ls -1 "${buffer_dirname}"/."${buffer_basename}".kak.* 2>/dev/null | while read -r f; do
-            rm -f "${f}"
-        done
+        rm -f "${buffer_dirname}/.${buffer_basename}.kak."*
+
+        printf %s\\n "
+          echo -markup {Information}Backup files removed
+          echo -debug Backup files removed with exit status: $?
```
mawww: I think this is a remaining debug line that should not be in the commited version.

rc/extra/autorestore.kak
nochiel: `\! -path .. -prune` is to avoid finding the wrong backup file in a subdirectory when we have:
```
./readme.txt
./.readme.txt.kak.z48y
./doc/readme.txt
./doc/.readme.txt.kak.z48y
```

rc/extra/autorestore.kak
```diff
@@ -7,31 +7,39 @@ def autorestore-restore-buffer -docstring "Restore the backup for the current fi
         buffer_basename="${kak_buffile##*/}"
```
nochiel: @mawww How come `basename` wasn't used here?

lenormf: No need to spawn an entire process for something a POSIX expansion can handle!

PR #1601, opened 2017-09-28
Docs: add missing -command-completion and -shell-candidates switches

Delapouite:


danr:
Note: these have `$token_to_complete` in the shell context, and also `-shell-completion` has `$pos_in_token`, as per:
https://github.com/mawww/kakoune/blob/76f072a786c84195bd04e65afd76ccc01da8b69b/src/commands.cc#L946

Issue #1465, opened 2017-06-24
`.` doesn't remember counts


lenormf:
Hi,

Something weird is happening with `.`, when used in the following scenarios:

* create 3 new lines: `3o<esc>`
* repeat the last insert: `.`
* a single line has been created above every previously created selection, whereas I expected *3* new lines to have been created *below* each selection

The opposite is also true when using `O`, single lines are created below each selection upon hitting the `.` key, whereas you would expect 3 lines to be created above.

HTH.

PR #1602, opened 2017-09-28
Add count support for repeat_last_insert '.' and repeat_last_select `<a-.>`

Delapouite:


lenormf:
The repetition primitives already remember the count used during the last operation, allowing a count on those would not serve much purpose as they would only multiply the cached count.

Delapouite:
You're right. The `.` already remembers the count used in the command it repeats as you asked in https://github.com/mawww/kakoune/issues/1465

But it's not the case of `<a-.>`. 
Currently: `]p` and `3]p` acts the same. With this PR you can do `]p`, and then `3<a-.>`.

Also, back to the `.`, it offers interesting perspective.
For example making markdown headers. I'm a the beginning on the line I want to turn in a h1 : `I#<esc>`. I jump to the line I want to turn in a h2: `2.`. Here comes a h4: `4.`…
So it's useful when you made a command with *1 count* and want to repeat a various count times depending of the context.

Finally, multiplying the command count with the repeating_last_insert count could lead to powerful edit operations in its own way.

lenormf:
OK.

>Currently: ]p and 3]p acts the same. With this PR you can do ]p, and then 3<a-.>.

As much as I understand the example, this would only be fixing a symptom, and not the root problem (which is that object selection doesn't support counts). Could be interesting in the mean time though.

>Finally, multiplying the command count with the repeating_last_insert count could lead to powerful edit operations in its own way.

Such as? The current idiomatic way of repeating actions a fixed amount of times is through macros, which support counts already, do you have an example in mind?

JJK96:
I second this, It happens quite often that I do an insert and want to repeat it a couple of times, so a count would be nice.

PR #1619, opened 2017-10-05
Add client_list var

Delapouite:
Hello

This list can be valuable to get more details about the current running kakoune sessions as in https://github.com/mawww/kakoune/issues/538
Also, it helps external scripts to deal with `jumpclient`, `toolsclient`… coordinations, positioning by the window manager etc…


danr:
Does this give a list of client pids or client  names?

lenormf:
From looking at the code, I would say -unescaped- client names.

PR #1627, opened 2017-10-10
colors/*.kak fixes

bhajneet:
Removed terminal colors from Solarized and added a light variant. Default theme was fixed/tested in tty for readability (lots of white+blue were changed to black+blue). This theme works well in both tty and in modern terminals where blue and white are often too similar in brightness.


lenormf:
I don't know what terminals you tested the default theme on, but on `termite` black/blue looks much less legible than white/blue. I think the current colors shouldn't be modified for this theme, unless the vast majority of the popular terminals used in the community is proven to mis-render the particular color pair (screenshots?).

No two terminals render colors the same way, so there's no way to make everybody happy, in any case this theme can still be overloaded from the user configuration.

bhajneet:
I tested on gnome-terminal and the default tty1 session. You're right that black/blue is less legible than white/blue in tty (and I suppose termite). However, for most modern browsers it's almost impossible to read blue on white versus the alternative.

Also, I didn't swap black for white (with the blue), I actually mis-typed that. What I meant to say was white/blue was changed to white/black for the menu (blue was swapped out).

So this:
screenshot from 2017-10-11 07-49-08

To this:
screenshot from 2017-10-11 07-48-46

Now we could fix the black+cyan for foreground menu, but this combo worked just fine in tty1 and I'm sure it does in termite as well. These same colors are used for auto-completion, and I think it works well there too.

The other major change I did to the default theme was the primary + secondary cursors/selections. A picture is most helpful again, so here you go:

Before:
screenshot from 2017-10-11 07-53-30

After:
screenshot from 2017-10-11 07-53-53

In the old way, you can't really tell which is primary / secondary for two lines (easier when you have 3 like in this example). Also white/blue is hard to read in modern color schemes (I tried a few like solarized, tango, base16), but you're right that black/blue is hard to read in tty1 / xterm / rxvt schemes. So perhaps, we could do the same thing we do for the menu and use cyan for the selection color:
screenshot from 2017-10-11 08-03-49

This works well with the modern terminals (matching "highlight" color of menus/autocomplete) and in older tty1 or other xterm-colored terminals.

The other minor changes were:

* Separate current line number with other line numbers by making it reversed and bold
* Hide wrapped line numbers (they can always be added by the user, but this default makes it obvious where lines are wrapped without getting in your way, and the default kakrc doesn't use wrapped lines anyways)
* Menu info (extra text) is changed to just match the auto-completion text. Cyan on white is difficult to read in either scenario and the whitespace buffer that's always been there is pretty good for distinction.
* Matching character doesn't work in old terminals well enough, so I made it reverse of the default color (if the text was red, it will be black on red). This is only a problem now when the text is white itself (reverse becomes black on white, similar to the off-black on white). So I also added an underline for the matching. I am on the fence for this one, but I still think in most cases it will be more readable to use the color of the text as the background.

Overall, I agree there is no way to make everyone happy, but there is at least a readable standard we can achieve for many more people.

mawww:
I think the change to the default theme should be discussed separately, could you remove the changes to `default.kak` from this PR ?

PR #1630, opened 2017-10-12
Add support for Uppercase registers with extending behavior

Delapouite:
Hello

## Context

Yesterday during a tupperVim meeting, a fellow vim user showed to the audience a few ways about how he was using registers in his workflow. One nice bit was using the uppercase registers to append new data to a register instead of just replacing it.

## How it fits in kakoune

I immediately thought "Well in kakoune we have more powerful primitives, like `<a-z>` and `<a-Z>` to do crazy stuffs like appending, union, intersection with selections etc…".
But then I realized that it does not really offer the same mechanisms. It's mainly focused on working with the `^` register (at least by default) which store full selection descriptions: coordinates, filename etc…

But as it's in the case in vim, sometimes you only want to work with raw contents in order to process it with another task or command.

Currently,  kakoune does not make any distinction between lowercase and uppercase alpha-registers.
Which means that `"ay` and `"Ay` is treated the same (despite the info in the status bar, but it's another matter).
I think that adding support for uppercase registers as a way to "extend" their content instead of replacing it, is a great parallel about how uppercase moves are the "extend" counter parts of the lower case ones:

`w`: select to next word start
`W`: extend to next word start
`"ay`: store in 'a' register
`"Ay`: extend 'a' register

For now I've limited the support to `y`, `c` and `d`, but I think it could probably be done with macro recording (in a future PR). It supports multiselection, which mimics the way the `<a-p>` command works. I've also added a small example about how it works in the register manpage (I don't know if it deserves to bloat the README even more).


nochiel:
- Is there anything you need from users of this branch? 
- Is it necessary to vote so that this feature gets merged upstream?

---

Thanks again. I've *really* needed this.  
<kbd>Z</kbd>move about, select something <kbd><a-Z>a</kbd>move about<kbd>zy</kbd> is useful and powerful but primitive accumulate/append-to-register is very useful when I've already yanked something into a register and need to quickly append instead of using a different register then having to paste multiple times from each register.

alexherbo2:
@Delapouite Could you feed us some GIFs? :smile_cat:

Delapouite:
I rewrote the looping logic of the PR taking into account the potential discrepancy between the number of current selections and the number of strings in the target register.

Items from the largest collection are now discarded:

- 1-1: sels [a] + reg [d] = reg[a, d]
- 2-2: sels [a, b] + reg [d, e] = reg[ad, be]
- 3-2: sels [a, b, c] + reg [d, e] = reg [ad, be]
- 2-3: sels [a, b] + reg [d, e, f] = reg[ad, be]

This time around I've added test cases for each scenarios.

mawww:
> 1-1: sels [a] + reg [d] = reg[a, d]

Why isn't that reg[ad] ?

Delapouite:
oops, this is a "typo" in my previous comment (I fixed it to avoid confusion for future readers).
You're absolutely right, the result is `reg[ad]`. This is what I intended to write in the comment and that's what is implemented codewise.

Abraxas-Knister:
Looks like reg[a,b] + sel[c] = reg[a,bc] might be an option in its own right, as maybe is reg[a,b,c]. Probably best to let the user decide like in the case of `<a-z>`?  

Another option might also be to use uppercase registers as another, unrelated set of registers, but that would be a different feature request. My idea for that would be to make the lowercase set window local registers and the uppercase set global ones. I'd like the register yank extend behavior with a new command `Y` in this case (if I see that correctly `Y` isn't doing anything for now?) Then it's clear that this only applies to yanking primitives and not to `*` and `Z`

PR #1631, opened 2017-10-13
Regex: Fix a few mistakes in the documentation

lenormf:
>For performance reasons lookaround contents cannot be an arbitrary regular expression, it must be sequence of literals, character classes or any-character (`.`), and the use of quantifiers are not supported.

Is the limitation of the engine that the lookaround must be of constant length, or that it cannot be a regular expression (with the exception of the `.` character)? Would `\d{4}` be acceptable?

Depending on which one it is, the quoted bit should be changed.

### DIFF
--- a/doc/manpages/regex.asciidoc
+++ b/doc/manpages/regex.asciidoc
@@ -11,7 +11,7 @@ Regex Syntax
 Kakoune regex syntax is based on the ECMAScript syntax, as defined by the
 ECMA-262 standard.
 
-Kakoune's regex always run on unicode codepoint sequences, not on bytes.
+Kakoune's regex always run on Unicode codepoint sequences, not on bytes.
 
 Literals
 --------
@@ -26,7 +26,7 @@ Some additional literals are available as escape sequences:
 * `\n` matches the line feed character.
 * `\r` matches the carriage return character.
 * `\t` matches the tabulation character.
-* `\v` matches the the vertical tabulation character.
+* `\v` matches the vertical tabulation character.
 
 Character classes
 -----------------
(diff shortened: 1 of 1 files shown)


doc/manpages/regex.asciidoc
```diff
@@ -161,11 +161,11 @@ preceded by `bar` and where `foo` matches from the current position
 Modifiers
 ---------
 
-Some modifiers can control the matching behaviour of the atoms following
+Some modifiers can control the matching behavior of the atoms following
```
mawww: What ! American spelling ? This goes too far mister.

doc/manpages/regex.asciidoc
```diff
@@ -116,7 +116,7 @@ by a quantifier, which specifies the number of times they can match.
 
 By default, quantifiers are *greedy*, which means they will prefer to
 match more characters if possible. Suffixing a quantifier with `?` will
-make it non-greedy, meaning it will prefer to match less characters.
+make it non-greedy, meaning it will prefer to match fewer characters.
```
mawww: alright for this time, lets let the prescriptivists have their fun.

mawww:
Regarding lookarounds, the limitation is that we only accept literals, character classes and `.`, and no quantifiers can be used.

I know `aaaa` and `a{4}` are the same, but seems simpler both in implementation and in description to refuse every quantifier instead of making an exception for this form. (And I suspect its not very useful for lookaround use cases anyway)

doc/manpages/regex.asciidoc
lenormf: The rest of the documentation is in american english as well, so I stayed within the same lexicon, even though I personally prefer the queen's english.

lenormf:
The comment on that particular paragraph wasn't about the technical aspect of the restriction, but rather how it was explained.

For instance, `(?=....)` is a regular expression, but the documentation says that they're not allowed, then exempts `.` from this rule. I would have stated things straighter, e.g.

>The contents of a lookaround must be sequence of literals, character classes or any-character (`.`); quantifiers are not supported.

mawww:
I am not sure I get you, the paragraph says:

> For performance reasons lookaround contents cannot be an arbitrary
> regular expression, it must be sequence of literals, character classes
> or any-character (`.`), and the use of quantifiers are not supported.

Are you bothered by the first part of the sentence "For performance reasons lookaround contents cannot be an arbitrary regular expression" ? The second part is basically the same as what you are proposing, right ?

lenormf:
"Cannot be a regular expression" even though any amount of dots is a regular expression, e.g. `....`. Instead of stating a general rule with exceptions, we make the text clearer by listing the few requirements directly, it's less confusing I think (and shorter).

mawww:
Understood, I changed that text in d776e06aca5dfa52971ec82ce926b92045a81831.

PR #1632, opened 2017-10-13
iterm: fix setting TMPDIR with fish shell, close pane on exit

vito:
This PR changes the iterm window handling to run the following:

```sh
export TMPDIR='${TMPDIR}'
exec kak ...
```

This does two things:

1. Fixes support for `fish`; `FOO=1 bar` doesn't work, but it does include bash-compatibility for `export`.
1. Makes it so that when you close the editor, the pane closes, consistent with the `tmux` rc.

Tested this manually by overriding the `def`s in my own [`kakrc`](https://github.com/vito/dotfiles/blob/master/kakoune/.config/kak/kakrc#L65-L120).

### DIFF
--- a/rc/extra/iterm.kak
+++ b/rc/extra/iterm.kak
@@ -25,7 +25,8 @@ def -hidden -params 1.. iterm-new-split-impl %{
         -e "    tell current session of current window"               \
         -e "        tell (split ${direction} with same profile)"      \
         -e "            select"                                       \
-        -e "            write text \"TMPDIR='${TMPDIR}' ${sh_cmd}\"" \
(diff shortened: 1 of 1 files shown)


vito:
I've actually got a follow-up commit for this ready, which does the following instead:

```
export TMPDIR='${TMPDIR}'
exec kak ...
```

This should work with both `bash` and `fish` (as `fish` provides a [compatibility function for `export`](https://github.com/fish-shell/fish-shell/blob/master/share/functions/export.fish)), and makes the behavior of quitting the editor consistent with the `tmux` integration (close the pane).

mawww:
Thanks !

casimir:
That bothered me too. Thanks!

PR #1636, opened 2017-10-16
Refactor comment.kak

occivink:
The `echo throw` pattern might be controversial, if you don't like it I can just as well change it to use an `-impl` command.

### DIFF
--- a/rc/core/comment.kak
+++ b/rc/core/comment.kak
@@ -59,7 +59,7 @@ hook global BufSetOption filetype=haml %{
 
 hook global BufSetOption filetype=html %{
     set-option buffer comment_line ''
-    set-option buffer comment_block_begin '<lt>!--'
+    set-option buffer comment_block_begin '<!--'
     set-option buffer comment_block_end '-->'
 }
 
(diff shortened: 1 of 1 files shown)


mawww:
I just introduced a built-in `fail` command that takes text as argument and fails for this use case. So I'd recommend to use `echo fail description`

rc/core/comment.kak
```diff
@@ -3,41 +3,50 @@ decl -docstring "characters inserted at the beginning of a commented line" \
...
+decl -docstring "characters inserted before a commented block" \
+    str comment_block_begin
+decl -docstring "characters inserted after a commented block" \
+    str comment_block_end
 
 ## Default comments for all languages
 hook global BufSetOption filetype=asciidoc %{
-    set buffer comment_block '///:///'
+    set buffer comment_block_begin '///'
+    set buffer comment_block_end '///'
```
lenormf: I originally stored the separators in a single command to keep things simple, why do we need two now? Parsing it is trivial with shell expansions, as long as no language needs a colon `:` as a commenting character.

rc/core/comment.kak
```diff
@@ -50,7 +59,8 @@ hook global BufSetOption filetype=haml %{
 
 hook global BufSetOption filetype=html %{
     set buffer comment_line ''
-    set buffer comment_block '<!--:-->'
+    set buffer comment_block_begin '<lt>!--'
```
lenormf: I also originally avoided doing this to make things simpler even if that meant spawning a process to escape the separators later on.

occivink:
I don't think having a third option is much of a negative, it doesn't cost anything. The beginning and end characters are distinct things so I think it makes sense to use separate options.
The `:` consideration is also a hard limitation, we wouldn't be able to handle such a language. Granted, very few languages use colons in their comments but it's still unnecessary. On the same note, just as few languages use `<` as part of their comments so it seems overkill to have generic code to handle them instead of just hardcoding it (I can add a comment about `<` to `<lt>` so that it doesn't trip up newcomers).
The combination of the two also allows us to remove the `sed` processing. It's not that it was too complex, just that having less logic is in my view a plus.
Finally, I think that doing the same text processing over and over on constant input is a bit of a code smell.

occivink:
Even better, we can put the values directly into the `"` and `/` registers so that we won't need to do any `<` to `<lt>` at all

occivink:
Any interest in this? Are there further changes you want to see?

lenormf:
My comments still stand.

occivink:
I don't have anything to add to my answer

mawww:
Hi hesitate on that,

I do not have a strong opinion one the two options/one option for block comments debate, I think the pro and cons of both do balance each other.

However we still require a shell block to validate the options, which means we can extend that shell block to do more necessary logic, such as transforming `<` into `<gt>`. I wonder if there are some cases where we do not want to have the `<gt>` version. Say to select a comment block, we would use custom text objects, and that would probably be triggered through a mapping, or direct exec, needing the `<gt>` form. If we were trying to put the comment text into a register and pasting it however we would like not to have the `<` form.

So, I do not mind much between the two versions, they do seem to be close to equivalent. Which makes it hard to decide whether to merge that PR or not.

occivink:
Ok, I didn't think about the possibility that someone might want to use the options outside of the script. I've now taken a different approach that doesn't need any `<` to `<lt>` conversion, but I had to modify the logic slightly

PR #1642, opened 2017-10-19
ini.kak: add ';' as a valid comment character

helvetie:
This is a very tiny fix.
Most implementations of the INI syntax I've come across understand `#` and `;` as comments, and so does Vim, which uses `;` as its default comment character for this file type.

Given that comment.kak uses `;` but kakoune only highlights `#`, this simple edit should fix it.

Thank you!


lenormf:
The semi-colon is a part of the specification for comments in INI files, so it should have been there in the first place regardless.

Thanks for fixing that!

PR #1649, opened 2017-10-22
Support bright named colors

### LINKED ISSUE #988: Support for bright terminal colors
I prefer to have my editor match my terminal color scheme. In Kakoune, you can define faces by name (e.g. "black", "red", "green"), but there are no valid strings for the corresponding "bright" colors. It would make sense to allow color assignment with integer values 0-255 like Vim does.

net:
Without this change it is impossible to accurately render color schemes requiring 16 precise colors, (e.g. Solarized, Atom One, Base 16, and many many more).

Closes #988.


lenormf:
Can you simply add another flag (e.g. `+l`) and document it as behaving as the bold flag on terminals that don't support bright colors?

Screwtapello:
"bold" and "bright colours" are different things. Bold makes the text wider, while "bright" draws the text at the same width, but with using a colour from a different palette entry.

Back in the days when the IBM PC supported 16 colour text but only one font weight, while most serial terminals supported 8 colours (or less) but did support bold, terminal emulators for the PC often used the PC's extra colours to represent the bold attribute, but that was an implementation choice due to the limited hardware at the time. Likewise, the PC could support bright background colours *or* blinking text but not both... but that doesn't make them the same thing, it just means the PC hardware was limited.

For the usual legacy cruft reasons, many modern terminals do support interpreting "bold" as "add 8 to the colour index if it's between 0 and 7 inclusive", but you can usually turn that off, and you'd probably want to in order to take proper advantage of 16-colour palettes like Solarized (as the OP suggests).

For example, here's `xterm -pc`:
image

The command being executed there is `tput setaf 1; printf "hel"; tput bold; printf "lo"; tput sgr0; tput setaf 9; printf 'wor'; tput bold; printf 'ld\n'` which prints "hel" in dark red, "lo" in bold dark red, "wor" in bright red and "ld" in bold bright red. You can see that the "o" and "w" are clearly different colours, despite one being bold dark red and the other being bright red.

net:
@lenormf as I understand them attributes apply to color pairs, not colors.

doppioandante:
How should this be interpreted in a GUI?

net:
@doppioandante the same as any other named color would be.

src/color.cc
```diff
@@ -19,6 +19,14 @@ static constexpr const char* color_names[] = {
     "magenta",
     "cyan",
     "white",
+    "brblack",
```
mawww: we have been moving away from abbreviations in Kakoune source code and in scripts, I suggest to use `bright-something`, fuzzy completion should make it quick to write those (br will match bright-red pretty strongly as it matches 2 word start).

src/color.hh
```diff
@@ -22,6 +22,14 @@ struct Color
         Magenta,
         Cyan,
         White,
+        BrBlack,
```
mawww: Same thing, name that `BrightSomething`

src/ncurses_ui.cc
```diff
@@ -238,6 +238,14 @@ default_colors = {
     { Color::Magenta, COLOR_MAGENTA },
     { Color::Cyan,    COLOR_CYAN },
     { Color::White,   COLOR_WHITE },
+    { Color::BrBlack,    8 },
```
mawww: Lets either use number for all colors, or (what I think would be my preference), use `COLOR_BLAH + 8` for bright colors.

src/color.cc
net: `brblack` has widespread use, but I understand moving away from abbreviations. Would `brightblack` (like tmux) suffice?

mawww: I'd rather have it in two words to make insert completion work well with it.

net:
@mawww changes made.

PR #1655, opened 2017-10-26
Add support for line completion in all buffers

Delapouite:
- via completers option with line=all vs line=buffer
- via <c-x> L mapping

### DIFF
--- a/src/insert_completer.cc
+++ b/src/insert_completer.cc
@@ -33,7 +33,7 @@ String option_to_string(const InsertCompleterDesc& opt)
         case InsertCompleterDesc::Option:
             return "option=" + (opt.param ? *opt.param : "");
         case InsertCompleterDesc::Line:
-            return "line";
+            return "line=" + (opt.param ? *opt.param : "");
     }
     kak_assert(false);
     return "";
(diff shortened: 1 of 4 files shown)


src/insert_completer.cc
```diff
@@ -335,18 +340,34 @@ InsertCompletion complete_line(const SelectionList& sels, const OptionManager& o
...
     InsertCompletion::CandidateList candidates;
-    for (LineCount l = 0_line; l < buffer.line_count(); ++l)
+
+    auto add_candidates = [&](const Buffer& buf) {
+        for (LineCount l = 0_line; l < buf.line_count(); ++l)
+        {
+            if (buf.name() == buffer.name() && l == cursor_pos.line)
+                continue;
+            ByteCount len = buf[l].length();
+            if (len > cursor_pos.column and std::equal(prefix.begin(), prefix.end(), buf[l].begin()))
```
mawww: We are going to need to be smarter than that I am afraid, we cannot be going around and storing every single line from every buffer in the candidate list and then try to sort it. It will get pretty slow, especially if it triggers automatically.

(If we have an empty line, we are collecting every single line from every single buffer)

I am not sure what the best way would be here, but I am afraid this will get pretty slow and unusable fast. Maybe limit our completion to cases where we have at least N chars on the completed line, or only get up to say 100 candidates and stop there (its unlikely the user intends to search for his line among >10 candidates anyway).

Delapouite:
I followed your perf recommandation and added this check at the beginning of the gathering loop:
```
if (candidates.size() == 100)
    break;
```

PR #1664, opened 2017-11-03
Improve scripts' readability

lenormf:


src/commands.cc
```diff
@@ -1694,9 +1694,9 @@ const CommandDesc exec_string_cmd = {
 };
 
 const CommandDesc eval_string_cmd = {
+    "evaluate-commands",
     "eval",
-    nullptr,
-    "eval <switches> <commands>...: execute commands as if entered by user",
+    "evaluate-commands <switches> <commands>...: execute commands as if entered by user",
```
danr: > _evaluate_ commands as if entered by the user

danr:
Readability is subjective so I will chime in with my opinion: I think this makes readability worse.

lenormf:
Since the aliases are not highlighted in the scripts anymore, I always have to stop and think whether the first non-highlighted keyword of a line has a typo. Now the code is more verbose, but I know that the code is lexicaly correct because it's highlighted as expected.

Also aliases allegedly are mostly for use interactively, that's why they aren't being highlighted anymore.

PR #1684, opened 2017-11-09
Fix use of default face attributes in ncurses set_face

Delapouite:
Hi

Currently it is not possible to give face attributes like `bold` or `italic` to gui parts like the `Menu` or the `StatusLine`:

`face MenuForeground +b` does not work

Since the atoms composing menus don't have their own face attributes, we rely on the ones provided by the default face.


Delapouite:
Sorry my example was a bit misleading. This one does not work either:

```
face MenuForeground red,green+b
```

The selected menu is red on green, but not bold.

PR #1695, opened 2017-11-12
c-familiy.kak: improve highlighting of C++ literals

fsub:
Support almost all types of literal values as given at [1,2,3].

[1] http://en.cppreference.com/w/cpp/language/character_literal
[2] http://en.cppreference.com/w/cpp/language/integer_literal
[3] http://en.cppreference.com/w/cpp/language/floating_literal

### DIFF
--- a/rc/core/c-family.kak
+++ b/rc/core/c-family.kak
@@ -178,7 +178,23 @@ add-highlighter shared/c/code regex %{\b-?(0x[0-9a-fA-F]+|\d+)[fdiu]?|'((\\.)?|[
 }
 
 # c++ specific
-add-highlighter shared/cpp/code regex %{\b-?(0x[0-9a-fA-F]+|\d+)[fdiu]?|'((\\.)?|[^'\\])'} 0:value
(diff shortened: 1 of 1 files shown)


lenormf:
It'd be interesting to know if these have any impact on performance.

fsub:
Me too. ;-) Do you have an idea how to get a meaningful result?
So far, I don't have any real world experience with those expressions. I just manually tested all cases that I could derive from the grammar.
Certainly, the expressions can be optimized for speed without changing the accepted language. That, on the other hand, may impact readability.
I'm perfectly okay if this won't get merged. I secretly hope that someone will propose a better solution. ;-)

mawww:
It will be a bit slower, but it should not be noticeable, Kakoune's highlighting *must* stay performant, and if adding those regex makes it too slow, I'd consider that a bug.

If that happens, one possible direction is extending the regex engine to support matching multiple regexes at the same time (doing a single iteration on the text while running multiple regex matching).

PR #1696, opened 2017-11-13
Add count support to jumps (<c-o> and <c-i>). Add jumps tests

Delapouite:


lenormf:
Out of curiosity, do you use that often?

PR #1698, opened 2017-11-13
Integration with GNU screen.

nochiel:
- [x] **FIXME: BUG:** If screen session is open in >1 terminals, `screen-new` commands will create a region in the most recently connected terminal.  Expected: region should be created in the terminal that has focus. wtf!
  - @mawww, @lenormf, @Delapouite: Might this be because [screen checks `ttyname(3)`](http://git.savannah.gnu.org/cgit/screen.git/tree/src/screen.c#n868)?
    - `eval %{%sh{ echo echo -debug $(tty) }}` -> "not a tty" 
    - Both emacs and vim return a valid tty and screen behaves as expected.

### DIFF
--- a/rc/base/screen.kak
+++ b/rc/base/screen.kak
@@ -1,0 +1,50 @@
+# http://gnu.org/software/screen/
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+ 
+hook global KakBegin .* %{
+    %sh{
+        [ -z "kak_client_env_$STY" ] && exit
+        echo "
+            alias global focus screen-focus
+            alias global new screen-new-vertical
+        "
+    }
+}
+
+
+def screen-new-vertical -params .. -command-completion -docstring "Create a new vertical region" %{
+     %sh{
+
+        tty="$(ps -o tty ${kak_client_pid} | tail -n 1)"
+        screen -X eval 'split -h' 'focus down' "screen kak -c \"${kak_session}\" -e \"$*\"" < "/dev/$tty"
+    }
+}
+
+def screen-new-horizontal -params .. -command-completion -docstring "Create a new horizontal region" %{
+     %sh{
+        tty="$(ps -o tty ${kak_client_pid} | tail -n 1)"
+        screen -X eval 'split -v' 'focus right' "screen kak -c \"${kak_session}\" -e \"$*\"" < "/dev/$tty"
+    }
+}
+
+def screen-new-window -params .. -command-completion -docstring "Create a new window" %{
+    %sh{
+        tty="$(ps -o tty ${kak_client_pid} | tail -n 1)"
+        screen -X screen kak -c "${kak_session}" -e "$*" < "/dev/$tty"
+    }
+}
+
+def -docstring %{screen-focus [<client>]: focus the given client
+If no client is passed then the current one is used} \
+    -params ..1 -client-completion \
+    screen-focus %{ %sh{
+        if [ $# -eq 1 ]; then
+            printf %s\\n "
+                evaluate-commands -client '$1' %{ %sh{
+                    screen -X focus
+            }}"
+        elif [ -n "kak_client_env_$STY" ]; then
+            tty="$(ps -o tty ${kak_client_pid} | tail -n 1)"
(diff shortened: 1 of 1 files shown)


mawww:
I suspect you can fix the bug by using the `kak_client_env_STY` env var instead of `STY`. `STY` is going to be the value of that variable for the *server* process, but you most likely want to use the one of the current *client* process.

rc/base/screen.kak
```diff
@@ -0,0 +1,48 @@
+# http://gnu.org/software/screen/
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+ 
+hook global KakBegin .* %{
+    %sh{
+        [ -z "$STY"] && exit
```
lenormf: There needs to be a space before the closing bracket.

rc/base/screen.kak
```diff
@@ -0,0 +1,48 @@
...
+            alias global focus screen-focus
+            alias global new screen-new-vertical
+        "
+    }
+}
+
+
+def screen-new-vertical -params .. -command-completion -docstring "Create a new vertical region" %{
+     %sh{
+       cmd="screen -X eval 'split -h' 'focus down' 'screen kak -c ${kak_session} -e \"$@\"'"
```
lenormf: * Why not use `eval` directly there?
* `$@` doesn't make sense here, you need to use `$*`
* `${kak_session}` needs to be enclosed in quotes

rc/base/screen.kak
```diff
@@ -0,0 +1,48 @@
...
+       eval ${cmd} 2>&1 >/dev/null
```
lenormf: If `cmd` has to stay, then this expansion needs to be enclosed in quotes. Also not hiding errors may be better, to track issues.

nochiel:
> @mawwwr: Or maybe you need to pass to screen another value, hopefully accessible from client env, that identifies the window of the current client.

If I run: `eval %{ %sh{ echo echo -debug $(stat /proc/self/fd/0) } }`, I get `File: /proc/self/fd/0 -> /dev/null`.  Is this supposed to be the case? In my admittedly limited experience, I haven't seen any programs that do that.

lenormf:
This is the result of caed8a55c7a5ac2754641769e1f603482886d5f1.

nochiel:
> @lenormf: This is the result of caed8a5.

Well, that commit is tricky because programs in the subshell might call `ttyname(0)` to find the current pts (I think it's what posix calls the "controlling terminal").  

I've added a workaround to screen.kak.

rc/base/screen.kak
```diff
@@ -0,0 +1,50 @@
...
+    -params ..1 -client-completion \
+    screen-focus %{ %sh{
+        if [ $# -eq 1 ]; then
+            printf %s\\n "
+                evaluate-commands -client '$1' %{ %sh{
+                    screen -X focus
+            }}"
+        elif [ -n "kak_client_env_$STY" ]; then
+            tty="$(ps -o tty ${kak_client_pid} | tail -n 1)"
+            screen -X select "$kak_client_env_WINDOW" < "$/dev/$tty"
```
lenormf: I think there's one `$` too many here.

rc/base/screen.kak
```diff
@@ -0,0 +1,50 @@
...
+            alias global focus screen-focus
+            alias global new screen-new-vertical
+        "
+    }
+}
+
+
+def screen-new-vertical -params .. -command-completion -docstring "Create a new vertical region" %{
+     %sh{
+
```
lenormf: Empty line.

mawww:
One thing to keep in mind is that child processes are forked from the *server* process of Kakoune, so trying to access the terminal there is usually wrong.

Kakoune uses `fork` and `execve` because that was the simplest way to achieve our goal. Currently forked processes should not assume they have a terminal attached to them. Not only dont we guarantee the server to be running inside a terminal, but we do not even do that for clients, as ncurses clients is the common case, but not the only one, we can have other clients (such as kakoune-qml) that do not rely on a terminal.

I am not sure I understand what `forkpty` would bring here.

nochiel:
> @mawww I am not sure I understand what forkpty would bring here.

I hear you.  I'll try to implement a poc of shellmanager that uses it and see what the results are.

PR #1702, opened 2017-11-18
Makefile: use PKG_CONFIG, not pkg-config

somasis:
As to follow one of the nice autotools standards.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -17,6 +17,8 @@ sources := $(sort $(wildcard *.cc))
 objects := $(addprefix ., $(sources:.cc=$(suffix).o))
 deps := $(addprefix ., $(sources:.cc=$(suffix).d))
 
+PKG_CONFIG ?= pkg-config
+
 PREFIX ?= /usr/local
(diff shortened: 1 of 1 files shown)


lenormf:
Since the project doesn't use the `autotools` at all, why follow this particular convention? As far as I know, this variable won't be used on any of the systems that have `pkg-config`, so it doesn't help much.

somasis:
Because it very well could be used. PKG_CONFIG is something that often 
gets set when cross-compiling.

lenormf:
That's very weird, we don't even support GNU targets in the `Makefile`, but we now have one variable used by a build system that's not being used to build the project?

mawww:
That fixes a problem I experienced while updating my build script for kakoune on exherbo, using plain pkg-config fails because its usually wrong when cross compiling, respecting an eventual PKG_CONFIG env var fixes it.

mawww:
We do respect `CXX` to set the C++ compiler, so I think we have `prior art` already, `PKG_CONFIG` is part of the build system tools we now rely on.

PR #1703, opened 2017-11-18
Makefile: Add ability to disable compressing manpage

somasis:
Some distributions don't compress them, so having an option to just disable it is nice.


somasis:
@lenormf Certainly, but having a single option to do it without any post-install work would be nice. I see your point, though.

src/Makefile
```diff
@@ -1,5 +1,12 @@
 debug ?= yes
 static ?= no
+GZIP_MAN ?= yes
```
mawww: Is that a conventional name ? if not, can it use the same naming conventions as the rest, namely lowercase ?

somasis: It's not a conventional name, I'll lowercase it.

PR #1704, opened 2017-11-18
Fix: GNU Screen integration

nochiel:
Ref. #1698
- [x] bug: `screen-focus` doesn't work as expected.  My initial testing was stupid and did not account for the fact that `focus` command's default is to focus regions in a cycle.
  - `screen` is very annoying.  There doesn't seem to be a way to focus an arbitrary region except by navigating in a sequence.  What is the correct solution for scripting `focus`?
  - [x] test: `screen -X select` does not actually switch focus to the region in which it is run so we can't use this.
  - possible fixes:
    - remove `screen-focus` and avoid using the `focus` command.
    - send a patch upstream: `focus [n|title]` 
    - [x] use the current region to switch to the client window.
- [x] `ranger.kak`: integrate with screen.
  - [x] fix: passing directory parameters to `ranger` command in screen doesn't work.
  - [x] fix: If directory name has quotes e.g. `"foo bar"`, then screen will be unable to pass it to ranger and the `ranger` command will fail.
  - [x] fix: If a directory name has a backslash in it e.g. `\"foo bar\"`, then it won't be passed to ranger correctly. The backslashes will be stripped.
  - [x] fix: If `$kak_{session,client}` have spaces in them, then the `ranger` command will fail.
  - [x] fix: passing command-line parameters to ranger fails.

### DIFF
--- a/rc/extra/ranger.kak
+++ b/rc/extra/ranger.kak
@@ -23,6 +23,29 @@ All the optional arguments are forwarded to the ranger utility} \
             kak -p $kak_session; \
             tmux select-pane -t $kak_client_env_TMUX_PANE'.format(file=fm.thisfile.path)) \
           if fm.thisfile.is_file else fm.execute_console('move right=1')"
+
+  elif [ -n "${STY}" ]; then
+
+    script="/tmp/kak-ranger-${kak_client}-${kak_session}.sh"
+    selections="/tmp/kak-ranger-${kak_client}-${kak_session}.txt"
+    cat > "$script" << EOF
+#! /usr/bin/env sh
+cd "$PWD"
+ranger --choosefiles="$selections" $@
+while read -r f; do
+    printf %s  "evaluate-commands -client '${kak_client}' edit '\"\$f\"'" | kak -p '${kak_session}'
+done < "$selections"
+screen -X remove
+rm -f "$selections" "$script"
+EOF
+
+        tty="$(ps -o tty ${kak_client_pid} | tail -n 1)"
+        screen -X eval \
+            'split -h' \
+            'focus down' \
+            "screen sh '$script'" \
+        < "/dev/$tty"
+
   elif [ -n "$WINDOWID" ]; then
     setsid $kak_opt_termcmd " \
       ranger $@ --cmd "'"'" \
--- a/rc/extra/ranger.kak
+++ b/rc/extra/ranger.kak
@@ -4,31 +4,31 @@
 define-command ranger-open-on-edit-directory \
     -docstring 'Start the ranger file system explorer when trying to edit a directory' %{
         hook global RuntimeError "\d+:\d+: '\w+' (.*): is a directory" %{ %sh{
-          directory=$kak_hook_param_capture_1
-          echo ranger $directory
+            directory=$kak_hook_param_capture_1
+            echo ranger $directory
     }}
 }
 
 define-command \
-  -params .. -file-completion \
-  -docstring %{ranger [<arguments>]: open the file system explorer to select buffers to open
-All the optional arguments are forwarded to the ranger utility} \
-  ranger %{ %sh{
-  if [ -n "$TMUX" ]; then
-    tmux split-window -h \
-      ranger $@ --cmd " \
-        map <return> eval \
-          fm.execute_console('shell \
-            echo evaluate-commands -client $kak_client edit {file} | \
-            kak -p $kak_session; \
-            tmux select-pane -t $kak_client_env_TMUX_PANE'.format(file=fm.thisfile.path)) \
-          if fm.thisfile.is_file else fm.execute_console('move right=1')"
+    -params .. -file-completion \
+    -docstring %{ranger [<arguments>]: open the file system explorer to select buffers to open
+    All the optional arguments are forwarded to the ranger utility} \
+    ranger %{ %sh{
+    if [ -n "${TMUX}" ]; then
+        tmux split-window -h \
+            ranger $@ --cmd " \
+                map <return> eval \
+                    fm.execute_console('shell \
+                        echo evaluate-commands -client ' + ranger.ext.shell_escape.shell_escape('$kak_client') + ' edit {file} | \
+                        kak -p '.format(file=fm.thisfile.path) + ranger.ext.shell_escape.shell_escape('$kak_session') + '; \
+                        tmux select-pane -t $kak_client_env_TMUX_PANE') \
+                    if fm.thisfile.is_file else fm.execute_console('move right=1')"
 
-  elif [ -n "${STY}" ]; then
+    elif [ -n "${STY}" ]; then
 
-    script="/tmp/kak-ranger-${kak_client}-${kak_session}.sh"
-    selections="/tmp/kak-ranger-${kak_client}-${kak_session}.txt"
-    cat > "$script" << EOF
+        script="/tmp/kak-ranger-${kak_client}-${kak_session}.sh"
+        selections="/tmp/kak-ranger-${kak_client}-${kak_session}.txt"
+        cat > "$script" << EOF
 #! /usr/bin/env sh
 cd "$PWD"
 ranger --choosefiles="$selections" $@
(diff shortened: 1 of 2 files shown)


rc/extra/ranger.kak
```diff
@@ -23,6 +23,18 @@ All the optional arguments are forwarded to the ranger utility} \
...
+  elif [ -n "$kak_client_env_STY" ]; then
+    screen -X eval \
+        'split -h' \
+        'focus down' \
+        'screen ranger $@ --cmd " \
+            map <return> eval \
+                fm.execute_console('shell \
+                    echo evaluate-commands -client $kak_client edit {file} | \
+                        kak -p $kak_session; \
+                    kak -p $kak_session <<< \"evaluate-commands -client $kak_client %{ %sh{
```
lenormf: Here strings `<<<` are not POSIX, unfortunately.

rc/extra/ranger.kak
```diff
@@ -23,6 +23,18 @@ All the optional arguments are forwarded to the ranger utility} \
...
             tmux select-pane -t $kak_client_env_TMUX_PANE'.format(file=fm.thisfile.path)) \
           if fm.thisfile.is_file else fm.execute_console('move right=1')"
+  elif [ -n "$kak_client_env_STY" ]; then
+    screen -X eval \
+        'split -h' \
+        'focus down' \
+        'screen ranger $@ --cmd " \
+            map <return> eval \
+                fm.execute_console('shell \
+                    echo evaluate-commands -client $kak_client edit {file} | \
```
lenormf: Prefer using `printf %s\\n` instead of `echo`, to avoid portability issues.

Wrap `$kak_client` with double quotes as well.

nochiel: @lenormf Shell quoting is driving me insane. Unfortunately I don't know how to do this correctly so that if ${kak_client} or ${kak_session} have spaces in them, they are handled.  Can you offer a fix?

lenormf: Maybe this would work:

```
echo \\\"evaluate-commands -client \\\\\"${kak_client}\\\\\" edit {file}\\\" | kak -p \\\"${kak_session}\\\"
```

nochiel: > @lenormf: Maybe this would work [...]

Does that work for you?  Unfortunately, I can't get it to work.

nochiel: @lenormf Would it make sense to write `ranger.kak` in python given that running ranger implies that there's a python interpreter available?  That would at least help avoid all the shell-quoting woes, I think.

mawww:
@nochiel your link leads me to a 404 error, I am not sure what indentwidth/tabstop vars your are refering to.

lenormf:
In most case we never want to modify a language support script to tweak how code is displayed on the client, unless it's a requirement by the language (e.g. Makefile targets have to start with a tab).

On a related note I found this yesterday, and I don't think it's right:

https://github.com/mawww/kakoune/blob/master/rc/extra/pony.kak#L94,L95

nochiel:
> @lenormf: n most case we never want to modify a language support script to tweak how code is displayed on the client,

Understood. Would it be better instead to include this modeline in .kak files: 
```
# kak:indentwidth=4:
```
and in .cc or .hh files
```
// kak:indentwidth=4:
```

So that those of us who use different indentation don't accidentally change what's used for kakoune source code?

mawww:
Kakoune modeline support is currently in `rc/extra/` which means its considered pretty secondary. If we start to use modelines in the core code, it should move to `rc/core` (`rc/core` contains whatever is needed to edit kakoune itself).

However I am not really fond of modelines in general, as I do not like the idea of putting tool specific metadata inside the file itself, so I'd rather not add modelines to Kakoune's sources files for now (until I change my mind).

mawww:
Hi,

If you are satisified with the PR, could you squash those commits into two, one that fixes screen.kak, and another one for ranger.kak changes ?

nochiel:
> @mawww : If you are satisified with the PR [...] 

I am.

lenormf:
>- Bug: Doesn't work if ${kak_client} or ${kak_session} have spaces. To fix this, both these vars should be quoted when passed to kak but deeply nested shell-quoting has proved difficult to use reliably.

This bit is still a problem though, and within reach.

nochiel:
> @lenormf: This bit is still a problem though, and within reach

Indeed.  I would appreciate some help.  I tried the previous solution you suggested but it also didn't quite work.  The multiple levels of quoting needed for the shell and embedded python code have me thoroughly flummoxed.  I don't know what else to try.

lenormf:
I tried using your version of `screen.kak`, and it doesn't work. I start `kak` within a `screen` session, and `:new` creates an X11 window. Apparently the check for `${kak_client_env_STY}` fails, but checking for `${STY}` directly works as expected.

nochiel:
> @lenormf: [...]apparently the check for ${kak_client_env_STY} fails, but checking for ${STY} directly works as expected.

In the previous PR @mawww [noted that](https://github.com/mawww/kakoune/pull/1698#issuecomment-344142359):
> [...] STY is going to be the value of that variable for the server process, but you most likely want to use the one of the current client process.

Problem: 
- If the kak server is started in screen but a kak client is in some other terminal, not screen.
or
- If the kak server is started not in screen, but a kak client is started in screen.

Desired: We want to alias `new` and `focus` for the client only. If there were a KakClientBegin hook, that's the one I would suggest using.

Bug (?): `$kak_client_env_STY` ≠ `$STY` if we start kak normally e.g. `kak foo.txt` without opening a client? I'm not sure I understand why this is the case. (I was planning on reporting it as [a separate issue](https://github.com/mawww/kakoune/issues/1716#issuecomment-346769395) once this PR was done unless there is a different solution already possible.)

lenormf:
The other multiplexer support scripts assume that the server runs within the multiplexer's session, it might make the code a bit ugly otherwise.

rc/extra/ranger.kak
```diff
@@ -14,30 +14,30 @@ define-command \
...
           if fm.thisfile.is_file else fm.execute_console('move right=1')"
-  elif [ -n "$kak_client_env_STY" ]; then
+  elif [ -n "${STY}" ]; then
     tty="$(ps -o tty ${kak_client_pid} | tail -n 1)"
     args="${@:-$PWD}"
     screen -X eval \
         'split -h' \
         'focus down' \
-        "screen ranger $args
+        "screen ranger \"$args\"
```
lenormf: Actually maybe this shouldn't be escaped, as the variable will contain command line parameters?

rc/extra/ranger.kak
```diff
@@ -14,30 +14,30 @@ define-command \
...
         --cmd '
 map <return> eval \
 fm.execute_console(\"shell \
-echo evaluate-commands -client ${kak_client} edit {file} | \
-kak -p ${kak_session};\
+printf %s evaluate-commands -client \" + shell_escape(\"${kak_client}\") + \" edit {file} | \
```
lenormf: Not sure how this works, as in theory everything after `%s` is supposed to be a single argument.

rc/extra/ranger.kak
nochiel: I was quoting this in order to ensure that the current directory (e.g. with spaces in them) was passed correctly.  I've since changed this to use the screen 'chdir` command.

As you've noted, quoting causes command line arguments to fail.  Indeed, I hadn't been testing for this correctly.

However, even unquoted, the problem remains that passing command line arguments to ranger doesn't work.  Consider, in kak: `ranger --cmd 'cd /etc'`
Using the unquoted forumula, screen seems to separate the command its evaluating at `--cmd` and tries to evaluate `cd` as a screen command.

I don't yet have a solution.

nochiel: @lenormf: Does passing command line arguments to ranger work in tmux or x11? I can't get it to work with the `ranger.kak` in master. My recent commit uses  `--choosefiles` and a heredoc shell script instead of `map`.  It's the only way I could get everything to work with screen.  If passing ranger command line args doesn't work in tmux or x11 perhaps we should switch them to `--choosefiles` and a heredoc shell script as well?

rc/extra/ranger.kak
nochiel: > @lenormf: in theory everything after %s is supposed to be a single argument.

Bizarrely, infuriatingly, this stopped working consistently "on my machine" and now sometimes only produces a single concatenated argument.

`printf "%s "` or `printf %s$' ' [...]` or `printf %s "[...]"` produces the expected result consistently.

rc/extra/ranger.kak
lenormf: You could wrap every argument in quotes, and pass this escaped string to `screen`, something like:

```bash
args=""
for i in "$@"; do
     args="${args} \"${i}\""
done

screen -X eval … "screen ranger ${args}…
```

nochiel:
@lenormf, @mawww: While testing, I found that x11 and tmux integration don't work for me. I'm not certain how to fix them.  My PR was intended to deal only with ranger.kak integration with screen which now works properly afaict.

Perhaps a different PR can be opened to fix ranger.kak x11 and tmux integration if it's broken?

nochiel:
This PR is now complete (with the exception of the bug (opening filenames with spaces) that I found in the tmux and x11 ranger.kak integration, which I think are out of scope for this PR).

rc/extra/ranger.kak
nochiel: I've reverted to using echo in the tmux and x11 sections so as not to break what was working in master.

mawww:
Hey, were are we at with this ? anything still blocking ? Note that I dont think the client/session names containing spaces use case is too problematic, I think we could actually reject those in general (Kakoune could refuse to use that, as they are just going to cause trouble for various scripts).

nochiel:
@mawww Unless @lenormf has any further comments, I believe it's ready.

Issue #1707, opened 2017-11-19
[TEST] Templated issue


lenormf:
> If you're not reporting a bug (e.g. requesting a feature or asking a general question), feel free to remove this template.
> Otherwise, edit the following sections with the relevant information.

### Steps to reproduce

List of minimal amounts of actions needed to trigger the bug, e.g.

* fire up `kak`
* `…`

### Outcome

What is the outcome of the above instructions?

If the result is a program crash, please include a *unmangled* stacktrace (c.f. `c++filt`) if possible, e.g.

```
assert failed "m_valid" at optional.hh:72
[Debug Infos]
pid: 1234
callstack:
kak(Kakoune::Backtrace::Backtrace() 0x2b) [0x845111]
[…]
kak(_start 0x29) [0x83e599]
```

### Expected behavior

How should the editor have behaved instead?

### Configuration

**Commit hash**: 079b006

**Operating System**: KakouneOS

**Terminal**: KakouneTerm

**Variable parameters**:

* [ ] release mode: was `kak` compiled with `debug=no` as argument to `make` (default: no)?
* [ ] no autoload: is the bug reproducible when the `-n` flag is passed to the `kak` program?

Issue #1606, opened 2017-09-30
Create a Kakoune organization on Github


lenormf:
Hi,

I'd like to have a "Github Organization" for the project, as it would have a couple advantages:

* centralize all the repositories related to the same project in one group to allow better discovery: the `kakoune`, `kakoune-golf` and `kakoune-website` (which should have its own repositories to let it have its own issue tracker) could be listed on a single dedicated page, instead of being mixed with the other repositories of the owner

* allow regular contributors to use management tools such as https://waffle.io/ to allow better discovery of what issues they could fix or what general improvement they could make in the project, without navigating the issue tracker by themselves; the owner of the project would also have better visibility over pressing issues and their current implementation state

* have its members display a fancy Kakoune logo on their profile, which is one more way to help potentially new users discover the project through developers they follow, and is also something recruiters like to see when they screen candidates

Note that being part of the organization in no way equals having push access to the repository, members all have a specific set of access rights in it which are assigned to by the owner: read only, read write, full member, outside collaborator etc.

Thanks.

robertmeta:
Another important benefit of this is if (at some far future date!) our benevolent dictator for life decides he doesn't want to maintain kakoune anymore, he can hand off the organization to someone else without everyone having to switch to a new fork.

robertmeta:
One additional point (and a bit of a bump to this) is that it stops anyone from taking the organization name Kakoune which could end up very confusing.

mawww:
I am on the fence about this one, on one hand I agree moving Kakoune to an organization would be nice, it would make Kakoune feel a bit more "official". But on the other hand, I am not sure I want my baby to fly to far away from its nest yet :).

lenormf:
There's already a release out, so the left the nest, the baby has. Also you wouldn't lose any control over it, it would just give you more granularity.

As for the organization name, I have created `kakoune-editor`, which I think is better than just `kakoune`, as it's more descriptive. Kind of like what the Dolphin emulator did.

robertmeta:
As @lenormf said -- the baby has a release, and this is just a new nest, still under your control.

maximbaz:
@mawww on a related note, how do you feel about giving collaborator access to someone you trust to help with the load? There are currently 50 open PRs, a big part of them are code-complete and just waiting for a code review or to be merged straight away. Also, there are over 200 issues, some of which are not actionable and should just be closed.

In my mind, an obvious candidate for this role would be @lenormf.

robertmeta:
[...]

Issue #1472, opened 2017-06-27
Use a wiki engine to host the pages from the Github wiki


lenormf:
Hi,

I think the Github wiki for the project has reached critical mass, which makes it hard to find things in it. The structure is also completely flat (it's based on pages), and could use some categories/sections. How about a more advanced wiki engine (self-hosted or not) for all the data that's regularly added to the Github wiki?

PR #1708, opened 2017-11-19
github: Add a template for bug reports

lenormf:
Demo issue at #1707.

I constantly follow the same format when I report an issue, making use of github's templates could save us some time as the same questions often come back when investigating a bug.

HTH.

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,10 @@
+<!--
+If you're not reporting a bug (e.g. requesting a feature or asking a general question), feel free to remove this template.
+Otherwise, please edit the following sections with the relevant information ().
+-->
+
+### Steps <!-- how to reproduce the issue? -->
+
+### Outcome <!-- what is the result of the above steps? -->
+
+### Expected <!-- what should have been the result of the above steps? -->


lenormf:
Any thoughts? Does the template make issues look too scary?

Delapouite:
I feel this PR belongs to the same bucket that:

better wiki engine: https://github.com/mawww/kakoune/issues/1472
kakoune github org: https://github.com/mawww/kakoune/issues/1606

The kakoune community is *microscopic* (even counting its 3000 stars). And there's no real need for big guns to handle the small stream of questions and contributions.

Now we could talk about the project's vision for the future, but it may deserve a dedicated conversation.

lenormf:
It doesn't matter how many users report bugs, the problem is that new users don't know what information to include in their issues.

mawww:
I am not sure we need such a template, I am afraid it will draw away some reports due to the apparent difficulty to get all the informations requested (like, whats my kakoune version, I installed it through the package manager of my distribution).

I think a simpler template, close to what @nochiel usually uses, like https://github.com/mawww/kakoune/issues/1718 would be better.

lenormf:
The only difference is the last section, which contains a checklist and info that is almost always needed. I could remove it to make the issue look more like the one you like, but then it would be pointless to have a template if all there is in it is sections that could be merged into a single sentence.

I could turn the template into a generic quote, giving a list of pieces of data that could potentially be of interest, the layout of the issue would be free.

lenormf:
I edited the template, it should now only show three sections, but has comments to guide users.

PR #1726, opened 2017-11-23
rc: Show a debug warning for unmet command dependencies

lenormf:


mawww:
I dont know about this, do we really want to clutter the debug buffer with all those messages by default ?

I think I'd prefer to get an error message once I try to use a command that does not have its dependencies satisfied, rather than always get them.

I think the underlying principle is that to me, having screen.kak loaded even though screen is not installed should not be viewed as an error, only trying to use screen.kak commands with no screen installed (or more simply without running inside a screen session). I dont think we should have lots of startup errors in the debug buffer for scripts loaded by default that the user is probably not interested in.

lenormf:
I only modified script that provided support for a specific tool/set of tools, so the amount of warnings is rather low (~3) on a typical desktop setup.

I thought it would help people who fine tune their custom `autoload` directory, if they only load scripts that they expect to be using, they'll know right away what they need to install by checking the debug buffer instead of having to read the script to see what commands will be needed.

Also it's a one-time silent warning sort-of hidden away in the debug buffer. The first thing regular users are going to see when they check it because a command they're trying to use doesn't do anything is the name of the script that implements that command, and the name of the expected dependencies.

mawww:
I think we should get an error message when we try to *use* the script, rather than just on load. Its usually the case, as Kakoune will automatically write whatever arrived on a `%sh{...}` stderr to the debug buffer.

I prefer that behaviour because I think displaying those warnings on load clutters the debug buffer, and gives us the habit to ignore that "header" that always appears in it, making us miss some real errors when they do not look different enough from what usually ends up there.

I agree that it can simplify a bit things, for example in the only script that does check on load at the moment `go-tools.kak`, moving the check to commands (needed as some of them run in the background, so detach themselves from stderr) will result in more complex code. So I understand the value of just checking on load, but I still think checking in commands is a better trade-off.

lenormf:
What about commands that have side effects before a non-installed shell command is being executed (e.g. creating temporary files)? Once the error breaks the execution flow, we do have an error, but any side effect that has been triggered to setup the actually execution of the shell command doesn't have a chance of being canceled, so users have to litter their filesystem to have a chance to see what they need to run a command.

I think checking in shell scopes that the program exists before it's being executed is kind of redundant by the way, since -as you said- a "no such file" error is going to be printed in the debug buffer anyway.

If this PR has to be denied, can I make another-one that removes the check in `go-tools.kak` to keep things consistent? After all that shell scope was the only thing that prompted this PR.

TeddyDD:
I'v got an idea:
1. Factor out command check to new core script `requires.kak` (or other suitable name). It would provide`requires <exec name>` command that would check if given executable is in PATH.
All script commands could use it to declare dependencies:
```
define-command foo %{
    require "git hg jq"
    ...
}
```

2. Add option in `reqires.kak` controlling `requires` behaviour when dependencies are not satisfied.
(fail/log to debug/ignore).

mawww:
@TeddyDD that could be a nice direction, but we need to solve another problem before: dependencies between scripts.

Right now we do not handle them, a script has no way to say "I require functionality provided by that other script". This is linked to the current design, where Kakoune knows almost nothing about script files:  we just open the script, execute all commands in it, and we are done. At no point do we know what comes from which script.

The most basic functionality we need is to allow dependencies to be loaded before processing a dependent script. which could be provided by a `source-if-not-yet-sourced <script>` command, requiring minimum infrastructure (we only need to maintain a list of sourced scripts). I fear this would, however, open a big can of worms, as we complexify loading of scripts to be more "programming language like".

lenormf:
I fail to see how a dedicated command to check for a program in the `PATH` is better than a simple scope: 

* the command would basically be a single call to `command`
* factorizing a one-liner into a dedicated script is overkill
* we still can't see what binary dependencies a script has without reading its code, the most complex scripts have one visible command (which would make use of the theoretical `requires` command) buried below more private commands - this is too much to ask from a non-savvy user
* allowing users to control the behavior of the command when a dependency is missing would be a major double standard considering features implementable through options were rejected until now

The PR is going nowhere, so I'm closing it, and users will have to hope for the best when they try commands, for now.

PR #1754, opened 2017-12-08
Added support for fairy assistant

aThorp96:
Added Navi as an assistant


lenormf:
Putting this here for future reference:

>mawww | I'd say from now on you have to be a long time contributor for your assistant pull request to be considered :D

mawww:
Thank you very much for this PR, however, as @lenormf posted, I dont think we should continue merging alternative assistant. I'd rather go (as previously stated) with a more general system allowing to specify custom assistant rather than add yet another one if the ncurses UI source code.

PR #1763, opened 2017-12-10
Fix highlight references in Markdown fenced code blocks

### LINKED ISSUE #1735: Go syntax highlighting in Markdown code block
**Type**: Bug

**Severity**: Low

**Version**: Not sure how to get the version via CLI, but whichever is in the AUR https://aur.archlinux.org/packages/kakoune-git. I just updated, but the issue was there prior to the update as well (and the last time I updated kakoune was at least a month ago).

**Description**

When writing a `go` code block in a `markdown` file, e.g.

    ```go
    func f() { /* ... */ }
    ```

The entire code block, including the leading \`\`\`go and trailing \`\`\`, are treated as one big string literal (which start and end with backticks) which means everything is just a single color. My guess is that the issue is with the [`markdown.kak`](https://github.com/mawww/kakoune/blob/master/rc/base/markdown.kak) file rather than [`go.kak`](https://github.com/mawww/kakoune/blob/master/rc/base/go.kak), since the highlighting inside a code block shouldn't be affected by the characters defining the block itself.

**Edit** Also note that the following modification, putting a \` after `go` fixes the highlighting inside the block (which isn't a solution, of course, but helps identify the issue):

    ```go`
    func f() { /* ... */ }
    ```

**Edit 2** Lot of backticks here, editing the formatting for clarity. Also, I suppose it likely isn't treating it as 'one big string literal', rather the first two backticks make an empty string literal, then the next backtick starts another, which is then closed by the first backtick in the closing ```, and finally the last two backticks in the closing trio make another empty string literal.

danr:
Fixes #1735

We need \K to not interfer with languages own interpretation of ` like multiline strings in javascript

We need \b in e.g. java\b otherwise it blocks javascript

I couldn't get the bare ``` to not block the other highlighters when introducing \K any other way than negative lookahead of all possible highlighers, this is quite icky

### DIFF
--- a/rc/base/markdown.kak
+++ b/rc/base/markdown.kak
@@ -12,51 +12,52 @@ hook global BufCreate .*[.](markdown|md|mkd) %{
 # ‾‾‾‾‾‾‾‾‾‾‾‾
 
 add-highlighter shared/ regions -default content markdown \
-    c          ```\h*c          ```          '' \
-    cabal      ```\h*cabal      ```          '' \
-    clojure    ```\h*clojure    ```          '' \
-    coffee     ```\h*coffee     ```          '' \
-    cpp        ```\h*cpp        ```          '' \
-    css        ```\h*css        ```          '' \
-    cucumber   ```\h*cucumber   ```          '' \
-    d          ```\h*d          ```          '' \
-    diff       ```\h*diff       ```          '' \
-    dockerfile ```\h*dockerfile ```          '' \
-    fish       ```\h*fish       ```          '' \
-    gas        ```\h*gas        ```          '' \
-    go         ```\h*go         ```          '' \
-    haml       ```\h*haml       ```          '' \
-    haskell    ```\h*haskell    ```          '' \
-    html       ```\h*html       ```          '' \
-    ini        ```\h*ini        ```          '' \
-    java       ```\h*java       ```          '' \
-    javascript ```\h*javascript ```          '' \
-    json       ```\h*json       ```          '' \
-    julia      ```\h*julia      ```          '' \
-    kak        ```\h*kak        ```          '' \
-    kickstart  ```\h*kickstart  ```          '' \
-    latex      ```\h*latex      ```          '' \
-    lisp       ```\h*lisp       ```          '' \
-    lua        ```\h*lua        ```          '' \
-    makefile   ```\h*makefile   ```          '' \
-    moon       ```\h*moon       ```          '' \
-    objc       ```\h*objc       ```          '' \
-    perl       ```\h*perl       ```          '' \
-    pug        ```\h*pug        ```          '' \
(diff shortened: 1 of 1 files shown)


rc/base/markdown.kak
```diff
@@ -12,51 +12,52 @@ hook global BufCreate .*[.](markdown|md|mkd) %{
...
+    rust       ^```\h*rust\b\K       ```          '' \
+    sass       ^```\h*sass\b\K       ```          '' \
+    scala      ^```\h*scala\b\K      ```          '' \
+    scss       ^```\h*scss\b\K       ```          '' \
+    sh         ^```\h*sh\b\K         ```          '' \
+    swift      ^```\h*swift\b\K      ```          '' \
+    tupfile    ^```\h*tupfile\b\K    ```          '' \
+    typescript ^```\h*typescript\b\K ```          '' \
+    yaml       ^```\h*yaml\b\K       ```          '' \
+    code       ^```((!?=(c|cabal|clojure|coffee|cpp|css|cucumber|diff|dockerfile|fish|gas|go|haml|haskell|html|ini|java|javascript|json|julia|kakrc|kickstart|latex|lisp|lua|makefile|moon|objc|perl|pug|python|ragel|ruby|rust|sass|scala|scss|sh|swift|tupfile|typescript|yaml)[^\n])*)$          ```          '' \
```
danr: hum this is fishy, `!?=` is not negative lookahead...

|            | lookahead | lookbehind
------------ | --------- | ----------
**positive** | `?=`      | `?<=`
**negative** | `?!`      | `?<!`

dgrisham: Good point, maybe that regex is just failing and thus doesn't override the previous blocks. I tried to modify the `code` regex, but I get the error:

```
markdown.kak:14:1: 'add-highlighter' regex parse error: Lookaround can only contain literals, any chars or character classes at '^```(?!(c|cabal|clojure|coffee|cpp|css|cucumber|diff|dockerfile|fish|gas|go|haml|haskell|html|ini|java|javascript|json|julia|kakrc|kickstart|latex|lisp|lua|makefile|moon|objc|perl|pug|python|ragel|ruby|rust|sass|scala|scss|sh|swift|tupfile|typescript|yaml))<<<HERE>>>[^\n]*$'
```

So maybe you can't use `|` in the lookahead.

dgrisham: This seems to fix it:

```
code       ^```(c|cabal|clojure|coffee|cpp|css|cucumber|diff|dockerfile|fish|gas|go|haml|haskell|html|ini|java|javascript|json|julia|kakrc|kickstart|latex|lisp|lua|makefile|moon|objc|perl|pug|python|ragel|ruby|rust|sass|scala|scss|sh|swift|tupfile|typescript|yaml)\K          ```          '' \
```

I'll make a PR with that.

PR #1765, opened 2017-12-12
Generate the markdown highlighter code

danr:
An extra round of regions does the trick to only highlight the inner parts of the block without the backticks.

### DIFF
--- a/rc/base/markdown.kak
+++ b/rc/base/markdown.kak
@@ -11,99 +11,31 @@ hook global BufCreate .*[.](markdown|md|mkd) %{
 # Highlighters
 # ‾‾‾‾‾‾‾‾‾‾‾‾
 
-add-highlighter shared/ regions -default content markdown \
-    c          ^```\h*c\b\K          ```          '' \
-    cabal      ^```\h*cabal\b\K      ```          '' \
-    clojure    ^```\h*clojure\b\K    ```          '' \
-    coffee     ^```\h*coffee\b\K     ```          '' \
-    cpp        ^```\h*cpp\b\K        ```          '' \
-    css        ^```\h*css\b\K        ```          '' \
-    cucumber   ^```\h*cucumber\b\K   ```          '' \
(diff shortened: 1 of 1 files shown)


alexherbo2:
Can you add Markdown it-self please?

alexherbo2:
And [Elvish][].

[Elvish]: https://elvish.io

danr:
Added `markdown`, but I cannot add elvish because it would refer to a non-existing highlighter

rc/base/markdown.kak
```diff
@@ -11,100 +11,32 @@ hook global BufCreate .*[.](markdown|md|mkd) %{
...
+%sh{
+  languages="
+    c cabal clojure coffee cpp css cucumber d diff dockerfile fish gas go
+    haml haskell html ini java javascript json julia kak kickstart latex
+    lisp lua makefile markdown moon objc perl pug python ragel ruby rust
+    sass scala scss sh swift tupfile typescript yaml
+  "
+  echo 'add-highlighter shared/ regions -default content markdown \'
+  for lang in ${languages}; do
+    printf '%s    ^```\h*%s\\b   ```   "" ' "${lang}" "${lang}"
```
dgrisham: Should the `\\b` here be `\K\b` ?

danr: No it is actually not needed anymore, the regions introduced for each language make sure that the triple-quotes are not to be highlighted by the language's highlighter.

dgrisham:
Copied this down to test it, it seems that codeblocks without a language aren't highlighted.

danr:
Can you paste the file you're testing here? My tests have been working.

Maybe I should test this using the json-ui. It's quite heavy, but it's what we have for testing highlighters afaik.

dgrisham:
Yeah, the file I'm testing is just:

    hello
    =====
    
    test
    
    ```
    int main();
    ```

I don't see a color change if I toggle between `` and ``` at the start of the code block.

PR #1766, opened 2017-12-12
Bug fix: Markdown fence regex

dgrisham:
Adjusting the fix in https://github.com/mawww/kakoune/pull/1763, since the regex engine doesn't support `|` in lookarounds here. As a result, the previous regex was failing and there was no color change on fenced codeblocks unless they were for a specific language -- this should fix that.


dgrisham:
@danr

danr:
Hey, thanks, I made another fix here instead which I think is more general because it avoids the copy-paste code throughout the file: https://github.com/mawww/kakoune/pull/1765

dgrisham:
@danr Ah, that looks so much better, nice! Was hoping for that kind of approach to this file anyway.

PR #1767, opened 2017-12-12
kakrc: Load files sorted by name

fsub:
In certain cases configuration files must be sourced in a particular order.
File systems usually do not enable users to reorder entries. However,
users usually have control over the naming.

### DIFF
--- a/share/kak/kakrc
+++ b/share/kak/kakrc
@@ -24,8 +24,8 @@ def -params 1 -docstring "colorscheme <name>: enable named colorscheme" \
 
 %sh{
     autoload_directory() {
-        find -L "$1" -type f -name '*\.kak' \
-            -exec printf 'try %%{ source "%s" } catch %%{ echo -debug Autoload: could not load "%s" }\n' '{}' '{}' \;
+        find -L "$1" -type f -name '*\.kak' | sort | xargs -I {} \
+            printf 'try %%{ source "%s" } catch %%{ echo -debug Autoload: could not load "%s" }\n' '{}' '{}'
     }
 
     localconfdir=${XDG_CONFIG_HOME:-${HOME}/.config}/kak


ekie:
i still use the old autoload implementation to enforce a particular order. https://github.com/mawww/kakoune/commit/d5726c0cfd3f4e872c6f85c6596ecb78614589f2#diff-7de0d503ebc5a6c79735c5965a426404
i find the directory hierachy to determine a order more intuitive.

mawww:
That could be a stop gap solution, but unfortunately it does not solve the general problem we want to solve, which is handling of dependencies between scripts.

I did not find a satisfying solution for that yet, basically what we want is:

* Make sure one script is loaded before another one
* Make sure a script is only loaded once

One solution which comes to mind right now would be to:

* Store the list of already sourced scripts in core kakoune
* Make `source` accept switches to specify
    * To allow reloading of an already loaded script (say `-reload`)
    * Not to fail if the script was already loaded (but not load it) (say `-if-unsourced`)
* Use `source -if-unsourced` at the begining of scripts to ensure their dependencies are loaded

Now this is not fully satisfying because
* It makes scripts a more complicated concept than just a list of commands executed as if they were written by the user
* Specifying a script we depend on is not easy (what to use to specify the script in a robust fashion ? `core/comment.kak` ? but then what happens if the use has its own autoload folder and has a `comment.kak` link there ? how to handle dependencies between non-bundled script if we use paths relative to `$kak_runtime`...)

I think once we solve that in a elegant and simple enough way, we wont need half solutions such as predictible sourcing order, which is very simple, but full of problems as well (ugly 01-naming required, makes it the user's job to ensure dependencies are satisfied...)

lenormf:
@ekie the output of the `find` command isn't deterministic, so you're relying either on a non-POSIX behavior, or on an implementation that might not be installed on all systems.

danr:
What about using standard unix tools for this? We could use `tsort` for ordering dependencies correctly. Here's how:

We can use `find` to get the kak files under `rc`:

```sh
$ find -L rc -type f -name '*\.kak'
rc/extra/latex.kak
rc/extra/jedi.kak
rc/extra/cabal.kak
rc/extra/haml.kak
rc/extra/coffee.kak
rc/extra/editorconfig.kak
...
```

Now let's allow a file to say what files it want to have been sourced before it.
Let's make it a bit simpler and say `rc/d.kak` needs to be after on `rc/a.kak` and `rc/c.kak`
and that `rc/b.kak` needs to be after `rc/a.kak`.
We modify the files like this:

* `rc/d.kak`
  ```
  # requires rc/a.kak
  # requires rc/c.kak
  ... rest of the file ...
  ```

* `rc/b.kak`
  ```
  # requires rc/a.kak
  ... rest of the file ...
  ```

Now let's generate a textual representation of the dependency graph.
We add self-edges to conform to the input format of `tsort`:

```
$ for file in $(find -L rc -type f -name '*\.kak'); do   \
    echo $file $file                                     \
    grep requires $file | sed 's,#\s*requires,'"$file,"  \
  done
rc/c.kak rc/c.kak
rc/b.kak rc/b.kak
rc/b.kak rc/a.kak
rc/d.kak rc/d.kak
rc/d.kak rc/a.kak
rc/d.kak rc/c.kak
rc/a.kak rc/a.kak
```

Sort the files topologically using `tsort`:

```
$ for file in $(find -L rc -type f -name '*\.kak'); do \
    echo $file $(grep requires $file | sed 's/#\s*requires//') \
  done | tsort | tac
rc/a.kak
rc/c.kak
rc/d.kak
rc/b.kak
```

The files can safely be sourced in this order.

Here's a little script that sets this up for you to experiment with:

```sh
mkdir -p depgraph/rc
cd depgraph/
touch rc/{a..d}.kak
echo '# requires rc/a.kak' >> rc/b.kak
echo '# requires rc/a.kak' >> rc/d.kak
echo '# requires rc/c.kak' >> rc/d.kak
for file in $(find -L rc -type f -name '*\.kak'); do echo $file $file; grep requires $file | sed 's,#\s*requires,'"$file,"; done | tsort | tac
```

mawww:
@danr, I am afraid your solution still the "how to identify a script" problem.

Relative to my suggested direction, it is more complex, but could be faster. Its hard to tell for performances as it requires grepping every sourced files and running tsort, compared with running each sourced files potentially multiple times (although only one run will fully source, all others would block at the `module` command, which is expected to be first).

danr:
Ok!

I understand your suggestion with a command with functionality 
* assignes the module an identifier,
* works as `#pragma once` which makes sure the file is only loaded once, 
* can mark modules depend on other modules which means that evaluation of it will be suspended until all its dependents are loaded

In my example we only need to give names to a and c. So we make these changes:

```
a.kak:
module a
```

```
b.kak
module -require a
```

```
c.kak:
module c
```

```
d.kak
module -require a -require c
```

The syntax and naming needs some work but the idea is that `module` takes 0-1 arguments of self-identifications and any number of `-require` arguments.

Is this the essence of your suggestion?

danr:
Would something like this below be OK? It is a little bit more flexible because you are not limited to modules and can specify any piece of code to have dependencies that must be loaded before it. This is the example I outlined above:

```
require a %{
    require c %{
        provide d
        echo -debug d loaded
    }
}

require a %{
    provide c
    echo -debug c loaded
}

provide a
provide b
```

This could be provided by the c++ code but it's also doable using what's provided by kakoune now:

```
decl int unique 0
decl str-list when ''
decl str-list provided ''

def require -params 2 %{
    %sh{
        name=suspended_$kak_opt_unique
        echo -n "
            set global unique $(($kak_opt_unique + 1))
            set -add global when '$1=$name'
            def -allow-override $name "
    } %arg{2}
    thaw %arg{1}
}

def thaw -params 1 %{
    %sh{
        echo "$kak_opt_provided" | tr : '\n' | grep "^$1\$" >/dev/null
        if [ $? = 0 ]; then
            for func in $(echo "$kak_opt_when" | tr : '\n' | grep "^$1=" | sed 's/^.*=//'); do
                echo "$func"
                echo def -allow-override "$func" nop
            done
        fi
    }
}

def provide -params 1 %{
    set -add global provided %arg{1}
    thaw %arg{1}
}
```

fsub:
@mawww The change proposed was not intended to solve the general dependency problem (although it does in a certain sense).
My concrete use case is [kakoune-mark](https://github.com/fsub/kakoune-mark). I need to ensure that mark.kak is sourced after all other files that are fiddling with highlighting. This requirement seems to be slightly different from what is mainly discussed here.
I'm not sure if declaring dependencies intrusively is superior to directly defining a list of files to be loaded. (The list could be given in a text file. Using file naming and the directory structure for defining the list has advantages and drawbacks.) A list is a simple concept that defines a unique ordering. Avoiding duplicate entries does not seem to be difficult.
In contrast, a declarative approach seems to be quite complex. The user needs to ensure that his declarations actually define a partial order. Probably, we need to detect cycles. If dependencies are specified intrusively, adding custom files seems to be messy. Depending on which kinds of rules are supported, introducing one file might involve editing multiple files (potentially some of which belong to kakoune).

mawww:
@danr I think you misunderstood my suggestion, its actually much simpler than that:

```
a.kak:
module a-identifier

commands...
```

```
b.kak:
source %val{runtime}/extra/a.kak

commands...
```

`module a-identifier` works like
```
#ifndef a-identfier
#define a-identifier

...
#endif
```

so `a.kak` will be loaded only on the first time we see `source a.kak`. The identifier is independant from the filename of the script, which makes it possible to "override" a script by reusing the same identifier (it remains an unsolved problem how to ensure that an user overriden implementation of  `a.kak` gets sourced first and prevents the `%val{runtime}/extra/a.kak` file from being sourced, maybe we still want to have sorting by filename in the autoload folder.

There is no need to defer loading of a script until another one is sourced, we just ensure its sourced at that script begining with the `source` command, knowing that we will not source the script multiple times thanks the the `module` command.

@fsub Why does kakoune mark needs to be sourced *after* other scripts ? I am afraid this approach is doomed because highlighting is often touched in hooks. So the scripts load but they do not touch the highlighting yet, its only when a buffer is opened that this happens, in a WinSetOption hook for example.

My main problem with the list approach is that we expose dependencies to the user, which must set up the list of files to load according to dependencies between them, and this is not very explicit (sometimes there is a `require` comment inside the script, but I dont expect each user to go read each script to know what needs to be sourced before what).

danr:
I see. We could use one kak function for that:

```
decl str-list sourced ''
def once -params 2 %{
    %sh{
        echo "set -add global sourced $1"
        echo "$kak_opt_sourced" | tr : '\n' | grep "^$1\$" >/dev/null
        if [ $? = 0 ]; then                
            echo -n "nop "
        else
            echo -n "eval "
        fi
    } %arg{2}
}
``` 

Usage: 

```
a.kak:
once a-identifier %{
   commands
   ...
}
```

mawww:
@danr yep, that would do the trick, but I think I'd rather have the command built-in for a few reasons:
* Avoid adding one layer of quoting to the whole script (as everything ends up inside the `once` parameter, needing to be quoted one way or another)
* Avoid the need for the script defining `once` to be loaded first (making it possible to source a file from `kak -n` instead of requiring it to first source whatever provides `once`)

fsub:
@mawww If I mark a word I want to override all existing highlighting. mark.kak must be sourced last so that it can install a hook that is executed last:
```
hook -group mark global WinSetOption filetype=.* %{ try %{
   # hack to override syntax highlighting
   remove-highlighter window/mark
   add-highlighter window ref mark
}}
```
I am curious to learn if there is a better approach.

If we restrict ourselves to 'require' dependencies it should be possible to avoid bothering users. Otherwise, if we want to support 'after' dependencies (like some OS init systems), I think that we cannot avoid exposing dependency information to users (who want a custom configuration) because we do not know (and cannot know) all nodes in their graph. We control only what is distributed along with kakoune. Only the user can decide where to put additional nodes.

Using the list approach we would deliver a file that contains everything that is distributed along with kakoune:
```
a.kak
b.kak   # requires a.kak  
c.kak   # requires a.kak
d.kak
e.kak   # after c.kak, requires d.kak
```
Comments can be used to clarify interdependencies. It should also be possible to skip loading a file by commenting the corresponding line.

If the user does not want a.kak he can follow a simple set of rules to get:
```
#a.kak
#b.kak   # requires a.kak  
#c.kak   # requires a.kak
d.kak
e.kak   # after c.kak, requires d.kak
```

If the user wants to add f.kak only he can know its dependencies. If f.kak must not be loaded before d.kak and requires b.kak he could arrive at:
```
a.kak
b.kak   # requires a.kak  
#c.kak   # requires a.kak
d.kak
f.kak   # after d.kak, requires b.kak
e.kak   # after c.kak, requires d.kak
```

Instead of using comments, the dependency specification could be formalized. It could then be used to check consistency or automatically insert a new line at a valid spot provided a user defined specification.

The alternatives seem to be:
- declaring a dependency graph intrusively, distributed along a set of relevant files;
- describing dependency information as a list stored at a central location;

mawww:
@fsub I think the best way for what you try to achieve is to use the `KakBegin` hook, which runs *after* sourcing the kak scripts.

Regarding the list approach, it still makes it the user job to know which script requires which, and will give non-obvious errors when the users gets it wrong. I think the intrusive way is better, because it gives that responsibility to the script writer, which is what makes most sense, and failure should be more explicit (`source blah...` is going to fail if blah... is nowhere to be found). Moreover, it will source a dependent script even if its not explicitely supposed to be autoloaded (a.kak depends on b.kak, but only a.kak is linked in the user autoload folder, sourcing a.kak will still trigger sourcing of b.kak) which I think the the correct behaviour (I want the feature in a.kak, and it so happen that b.kak must be sourced to get them), rather than failing because b.kak was not in autoload.

fsub:
@mawww Thanks for the feedback. I'll try using the `KakBegin` hook.

The problem I see is with scripts developed independently without knowledge of each other. Let `C` be some component and let `D` denote the set of all other components directly depending on `C`. Some developer does not like `C` and invents a replacement `R`. He also takes care to adapt some of the components from `D`. Some user learns about `R` and wants to use it. Alas, he uses components from `D` that have not been adapted. Now both `C` and `R` are loaded. Unfortunately, they are incompatible.

What I'm saying is that I don't see a simple approach that can solve this mess automatically.

The mess can be avoided if there is some authority that maintains a component repository and takes care of dependencies and compatibility. The fundamental property here is knowlege of the complete graph.

fsub:
@mawww Using the `KakBegin` hook solved the issue with kakoune-mark. Thanks!

PR #1780, opened 2017-12-23
src: Handle files bigger than 2G

lenormf:
I managed to open a 4G file with this patch, and scroll through it.

However adding a line at the end of file took 100% of the CPU and was very slow. Not sure if it's my modifications causing this, or if it hints at a loop somewhere that could be optimized.

The process also took 10G of RAM.


mawww:
I would like to get that fixed at some point, but it will not be with this approach. I think limiting lines to 2G bytes is reasonable, so we do not need to change the coord type. What would need to change is StringView.

PR #1785, opened 2017-12-25
src: Skip the current buffer in completion candidates

### LINKED ISSUE #1782: Buffer name completion
Completion for `:buffer` command shouldn't include the current buffer name.

lenormf:
Fixes #1782

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -127,6 +127,10 @@ static Completions complete_buffer_name(const Context& context, CompletionFlags 
     for (const auto& buffer : BufferManager::instance())
     {
         StringView bufname = buffer->display_name();
+
+        if (bufname == context.buffer().display_name())
+            continue;
+
         if (buffer->flags() & Buffer::Flags::File)
         {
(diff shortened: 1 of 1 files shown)


mawww:
I dont think this is a good idea, I agree we could get rid of the current buffer for the `:buffer` command, but this will also remove it from the `:delete-buffer` command or any custom command using `-buffer-completion` which is not always what you want.

Also, on an implementation standpoint, dont compare the buffer name to check if this is the current one, compare the object addresses.

lenormf:
>dont compare the buffer name to check if this is the current one, compare the object addresses

I thought doing the opposite of that would be better, comparing addresses is often a slippery slope.

mawww:
@lenormf at least in the way I code, either a type is a value-type and then you'll have an existing `operator==` (`BufferCoord` for example is a value-type, `Face` as well), or it is an object thats expected to be long lived and accessed indirectly (through a pointer or a reference, such as `Buffer`, `Window`...) and this object identity is usually its address. This is C++, objects do not move around in memory without explicit code.

PR #1787, opened 2017-12-27
doc: Don't present Kakoune as a Vim alternative

### LINKED ISSUE #1536: Remove references to `vim` on the site and in the documentation
Hi,

Many new comers seem to be mislead by the multiple references to `vim` on the website or in the documentation, thinking that Kakoune is another similar implementation aimed at competing with it. We regularly get questions about plugins cross-compatibility, window splits etc, and the answer to those questions usually push new users away because they consider any difference with `vim` to be a missing feature, whereas some are really a design choice complimentary to the ones made by `vim`.

This shows that the message doesn't get across well after a superficial look at the project's landing page/`README`: Kakoune is an implementation of a "selection-first" paradigm, wrapped in modal edition primitives and with a strong UNIX flavor. It has some similarities with `vim` (and other modal editors), and is often compared to it in terms of efficiency (golf) because it's a pillar of the world of code edition, but I feel like the way Kakoune is presented at first glance makes it look like a replacement, rather than simply a different tool (in the UNIX way) with different strengths.

As such, I'd like some mentions to `vim` to be removed from the site/documentation when they are not strictly necessary (e.g. the "Vim inspired" item in the project's motto), and the others kept to show how well/badly the editor performs compared to its "cousin".

Maybe Kakoune could be compared to other modal editors as well, to avoid polarizing the presentation of the project?

lenormf:
Fixes #1536

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -15,7 +15,7 @@ TL;DR
 
 {website}
 
-*Vim inspired* -- *Faster as in fewer keystrokes* --
+*Modal editor* -- *Faster as in fewer keystrokes* --
 *Multiple selections* -- *Orthogonal design*
 
 ---------------------------------------------
@@ -26,7 +26,7 @@ make
 ---------------------------------------------
 
 See http://github.com/mawww/golf for kakoune solutions to vimgolf challenges,
-regularly beating the best vim solution.
+regularly beating the best Vim solution.
 
 See the link:doc/design.asciidoc[design document] for more information on
 Kakoune philosophy and design.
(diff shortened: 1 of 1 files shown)


README.asciidoc
```diff
@@ -15,7 +15,7 @@ TL;DR
 
 {website}
 
-*Vim inspired* -- *Faster as in fewer keystrokes* --
+*Modal edition* -- *Faster as in fewer keystrokes* --
```
Screwtapello: While "edition" is an English word, it's usually in the context of a published book. Perhaps "editing" or "editor" would be more appropriate here?

lenormf: I wrote "modal editor" on the site too, fixed.

PR #1796, opened 2018-01-07
Highlight JSX

### LINKED ISSUE #1700: Jsx Support
I would like to create a jsx highlighter, are there any tips I should know before I start? I've looked at that way markdown handles nested highlighters, but I think jsx would be a bit more complex since it can handle recursive highlighters e.g. js>html>js>html. Any help would be greatly appreciated.

danr:
Additionally merges the javascript and typescript highlighters.

Fixes #1700


danr:
I haven't battle-tested this yet, I'll probably find some quirks to iron out in the upcoming days

mawww:
I am not very familiar with all that Javascript ecosystem, but does it makes much sense to have those supported directly in the javascript.kak file ?

danr:
It might not make sense, true. There is some collision as the JSX highlighter will be activated on the comparison that looks like a start tag in `a<b` (but not with a space in between: `< b`). JSX is only meant to be enabled on file endings `jsx` (`tsx` for TypeScript).  

Introducing more filetypes is the only way I see to selectively enable JSX. What do you think about filetypes named for example`javascript-jsx` and `typescript-jsx`? Only those filetypes would then have this additional highlighting.

gerardabello:
@danr @mawww I think it makes a lot of sense to do this for `.js` files. It is true that technically JSX code should use the `.jsx` extension, but the reality is that nowadays a lot of projects just use `.js`.

mawww:
Ok, I guess I'll merge that. @danr can you fix the conflicts first ?

danr:
Done!

Issue #1685, opened 2017-11-09
Support for creating more KeymapModes


Delapouite:
Hello

Kakoune offers these builtins KeymapModes : 

```
enum class KeymapMode : char
{
    None,
    Normal,
    Insert,
    Prompt,
    Menu,
    Goto,
    View,
    User,
    Object,
};
```

My `user` mode is being pretty crowded, with many unrelated mappings fighting for the same letter.
I would love to be able to **declare new modes at runtime**.

For examples one to keep all the `buffer navigation` stuffs together, another to keep my `JavaScript Flow` mappings under the same prefix, one for `git`… Well this list can be pretty big.

The new commands could look something like this (I'm opened to suggestions) :
- `declare-mode <mode_name>`
    ->  :declare-mode git,
    -> :declare-mode flow
    …
- `enter-mode <mode_name>` 
    -> :enter-mode grep
    -> :map global normal , ':enter-mode user<ret>' (today's equivalent of `,` key)

This way you can use `map` and `unmap` as usual:
```
declare-mode grep
map global grep n ':grep-next-match<ret>' -docstring 'next match'
map global grep p ':grep-previous-match<ret>' -docstring 'previous match'
map global normal <a-g> ':enter-mode grep<ret>' -docstring 'enter grep mode'
```

This is an ambitious architectural change codewise but I feel it would be a great enhancement in flexibility. Other editors like *Spacemacs* hugely benefits from this.

I know that today, you can emulate this use case as described in the [wiki]( https://github.com/mawww/kakoune/wiki/Implementing-user-mode) with a mix of `on-key` and `info` commands. I already did it multiple times in some of my plugins, like in [kakoune-buffers](https://github.com/Delapouite/kakoune-buffers/blob/master/buffers.kak#L125-L157) for instance.
But it's cumbersome, you have to rely on manual shell scripting to switch through the keys and implement the info with hardcoded values. With the proposed solution above, you let other users of your plugin extend/change the mapping the way they already know, with `map` and the infobox is autoupdated for free.

This proposal is still a bit rough, don't hesitate to share your ideas.

Delapouite:
alex shared an intesting thought on IRC.
For now I've only described "one shot" mode, like the current 'user' mode. But the implementation should offer a way to enter "lock" modes, like `V`. In this case `<esc>` can't be remapped since it's the way to go back to the previous mode.

mawww:
[...]

PR #1803, opened 2018-01-18
Add declare-user-mode / enter-user-mode commands

Delapouite:
Hi.

This is a proof a concept for this issue: https://github.com/mawww/kakoune/issues/1685

It *works* and does not introduce drastic changes and millions of new lines to the code base, or paradigm shift. But keep in mind that I'm not well versed in C++ and memory safety so they may be obvious flaw in the design.

## Usage example

Lets say I want to keep all the commands dedicated to navigation between buffers, clients… in the same keymap called `nav`:

`:declare-user-mode global nav`

Now I can start populating it, as usual:

`:map global nav n '<esc>:buffer-next<ret>' -docstring 'next buffer'`
`:map global nav p '<esc>:buffer-previous<ret>' -docstring 'previous buffer'`
`:map global nav o '<esc>:buffer-only<ret>' -docstring 'keep only current buffer'`

Finally I assign a normal key to enter the `nav` mode:

`:map global normal <a-,> ':enter-user-mode global nav<ret>' -docstring 'enter nav mode'`

That's it.

Now when I press `<a-,>` an autoinfo pops, showing the docstring I registered for the `nav` mode.
If I press `n`, I will go to next buffer and be back to normal mode. So like `goto` or `view` modes, it's a one shot action.

## About the code

It introduces a *global* vector `keymap_modes` (I thought building a whole KeymapModeRegistry class was a bit overkill).

`declare-user-mode` push the new mode inside.
`enter-user-mode` is mostly a copy paste of the `,` command.


src/keymap_manager.cc
```diff
@@ -8,6 +8,10 @@
 namespace Kakoune
 {
 
+// the define-user-mode command will push into this vector
+Vector<String> keymap_modes = { "normal", "insert", "prompt", "menu",
```
mawww: I really dont like this global, it needs to stay in sync with the 'KeymapMode` enum.

src/commands.cc
```diff
@@ -1098,16 +1099,12 @@ const CommandDesc echo_cmd = {
 
 KeymapMode parse_keymap_mode(const String& str)
 {
-    if (prefix_match("normal", str)) return KeymapMode::Normal;
```
mawww: We are loosing that prefix behaviour here, but I think its actually better, we have completion for that.

Delapouite:
Thanks for your input!

I've fixed the easy parts:
- the line wrongly removed in the docs
- I've added a check for only alphanumeric characters when declaring a new user-mode

To be honest, I was not really proud of the global vector either. As I said it was just a proof of concept using my humble skillset in the language to try to make something work with the minimal amount of changes. To get the "feeling" of this new feature.

I'll follow your advice and try to come up with a more robust solution where additional modes are stored on the KeymapManagers themselves.

src/commands.cc
```diff
@@ -2101,6 +2095,63 @@ const CommandDesc fail_cmd = {
...
+    single_param,
+    CommandFlags::None,
+    CommandHelper{},
+    CommandCompleter{},
+    [](const ParametersParser& parser, Context& context, const ShellContext&)
+    {
+        if (contains(keymap_modes, parser[0]))
+            throw runtime_error(format("mode '{}' already defined", parser[0]));
+
+        if (std::any_of(parser[0].begin(), parser[0].end(),
```
mawww: `contains_that(parser[0], <<lambda here>>)` to the rescue

Delapouite:
Hi mawww. Here's a refactor of the PR trying to follow your guidelines.

The `user_modes` vector is not a global anymore but is a member of `KeymapManager`. Which means that user-modes are now scoped. So this change offers even more flexibility than my previous attempt.

A basic test of the feature has also been added in `/tests`

You're right, after a phase of deprecation, the current `user` mode could be removed in favor of letting people crafting their own specialized user modes. (side effect: iit also frees the `,` key for other purposes).
A new `user-modes` doc page could potentially be added in another PR if needed.

src/commands.cc
```diff
@@ -1107,7 +1108,12 @@ KeymapMode parse_keymap_mode(const String& str)
     if (prefix_match("user", str))   return KeymapMode::User;
     if (prefix_match("object", str)) return KeymapMode::Object;
 
-    throw runtime_error(format("unknown keymap mode '{}'", str));
+    auto it = std::find(user_modes.begin(), user_modes.end(), str);
```
mawww: use the range based find wrapper provided by ranges.hh: `find(user_modes, str)` (it will actually use std::find, but does the iterator boilerplate for you).

src/commands.cc
```diff
@@ -1415,11 +1421,13 @@ auto map_key_completer =
...
                      complete(params[0], pos_in_token, scopes) };
+
         if (token_to_complete == 1)
         {
-            constexpr const char* modes[] = { "normal", "insert", "menu", "prompt", "goto", "view", "user", "object" };
+            auto modes = { "normal", "insert", "menu", "prompt", "goto", "view", "user", "object" };
+            auto user_modes = get_scope(params[0], context).keymaps().user_modes();
             return { 0_byte, params[1].length(),
-                     complete(params[1], pos_in_token, modes) };
+                     complete(params[1], pos_in_token, concatenated(modes, user_modes) | gather<Vector<String>>()) };
```
mawww: `complete` takes an arbitrary container as last argument, so no need to gather the concatenated result into a vector, we can avoid copying all that data.

src/keymap_manager.hh
```diff
@@ -43,6 +43,10 @@ public:
     };
     const KeymapInfo& get_mapping(Key key, KeymapMode mode) const;
 
+    using UserModeList = Vector<String>;
+    const UserModeList user_modes() const { return m_user_modes; }
```
mawww: return a const reference here, not a copy.

src/commands.cc
```diff
@@ -1107,7 +1108,12 @@ KeymapMode parse_keymap_mode(const String& str)
     if (prefix_match("user", str))   return KeymapMode::User;
     if (prefix_match("object", str)) return KeymapMode::Object;
 
-    throw runtime_error(format("unknown keymap mode '{}'", str));
+    auto it = std::find(user_modes.begin(), user_modes.end(), str);
+    if (it == user_modes.end())
+        throw runtime_error(format("unknown keymap mode '{}'", str));
+
+    char offset = static_cast<char>(KeymapMode::Object) + 1;
```
mawww: I think we should introduce a `KeymapMode::FirstUserMode` as the last KeymapMode, and use that rather than rely on the fact that `KeymapMode::Object` is the last one. With this code we'll get a bug as soon as a new non-user mode gets added at the end of the KeymapMode enum.

src/keymap_manager.cc
```diff
@@ -49,4 +51,15 @@ KeymapManager::KeyList KeymapManager::get_mapped_keys(KeymapMode mode) const
     return res;
 }
 
+void KeymapManager::add_user_mode(const String& user_mode_name)
+{
+    if (contains(m_user_modes, user_mode_name))
```
mawww: We might want to validate against built-in mode names as well.

src/keymap_manager.hh
```diff
@@ -43,6 +43,10 @@ public:
...
+    void add_user_mode(const String& user_mode_name);
```
mawww: We do not pass const ref to strings in Kakoune code anymore (except very special cases), we either use a StringView (passed by value) or, if we are going to copy the string, a String by value that gets std::move'd. In the current case, both seems acceptable to me.

src/commands.cc
Delapouite: I had to use `gather` because without it the code triggers this static_assert error about long live string:

```
completion.hh: In instantiation of ‘Kakoune::CandidateList Kakoune::complete(Kakoune::StringView, Kakoune::ByteCount, const Container&) [with Container = Kakoune::ConcatView<std::initializer_list<const char*>&, std::vector<Kakoune::String, Kakoune::Allocator<Kakoune::String, (Kakoune::MemoryDomain)0> >&>; Kakoune::CandidateList = std::vector<Kakoune::String, Kakoune::Allocator<Kakoune::String, (Kakoune::MemoryDomain)22> >]’:

commands.cc:1430:87:   required from here

completion.hh:67:5: error: static assertion failed: complete require long lived strings, not temporaries
     static_assert(not std::is_same<decltype(*begin(container)), String>::value,
```

What would be the alternative?

mawww: Ahh, right, ok I guess you can gather there for now.

src/commands.cc
```diff
@@ -1415,11 +1421,13 @@ auto map_key_completer =
         if (token_to_complete == 0)
             return { 0_byte, params[0].length(),
                      complete(params[0], pos_in_token, scopes) };
+
         if (token_to_complete == 1)
         {
-            constexpr const char* modes[] = { "normal", "insert", "menu", "prompt", "goto", "view", "user", "object" };
+            auto modes = { "normal", "insert", "menu", "prompt", "goto", "view", "user", "object" };
```
mawww: keep that constexpr

src/commands.cc
```diff
@@ -1415,11 +1421,13 @@ auto map_key_completer =
...
+            auto user_modes = get_scope(params[0], context).keymaps().user_modes();
```
mawww: `auto&` to avoid an uneeded copy

Delapouite:
Thanks! I've added back `constexpr` and used `auto&`

Delapouite:
Unfortunately Travis CI is green for my combo (gcc + linux) but failed for the 3 others:

```
commands.cc:1427:28: error: constexpr variable 'modes' must be initialized by a constant expression
            constexpr auto modes = { "normal", "insert", "menu", "prompt", "goto", "view", "user", "object" };
```

What is the correct way to express this?

mawww:
Try to specify the type, `constexpr StringView modes[] =`

Delapouite:
I followed similar statements in `commands.cc` and added `static` in front of the `constexpr`. CI is finally green on all platforms.

PR #1806, opened 2018-01-21
doc: Document the `config` option

lenormf:


danr:
The config is a value, `%val`, not an option: https://github.com/mawww/kakoune/blob/299e22ca7c3a976c7f2edf99f4ff248f6e4c9f85/src/main.cc#L124

mawww:
And it is already documented as of 07dfcd336d31987ea30c6345a0674cd5c16ed21a.

lenormf:
I was sort of thinking that all variables were already in the `options` documentation page, so I blindly put `config` in there, but it turns out the default "values" are in `expansions`… Maybe the two lists could be shown on the same page?

PR #1810, opened 2018-01-24
Add heredoc support for ruby.kak

jjthrash:


rc/base/ruby.kak
```diff
@@ -22,6 +22,7 @@ add-highlighter shared/ regions -default code ruby       \
     literal       '%[iqrswxIQRSWX]\{' \}     \{ \
     literal       '%[iqrswxIQRSWX]\[' \]     \[ \
     literal       '%[iqrswxIQRSWX]<'   >      < \
+    heredoc '<<-?(\w+)'         '^\h*(\w+)$' '' \
```
mawww: I suspect you want the words to match here, so you'd need to add `-match-capture` to the switches of the regions highlighter.

jjthrash: Thanks. I updated via force-push.. hopefully that won't confuse things.

PR #1811, opened 2018-01-25
Add preserve-previous-line-indent

danr:
This is initial work on trying to reduce the code copypaste in the rc-files by introducing commands to do common indentation heuristics.

If this approach is approved by other contributors and merged we can introduce additional commands, such as increasing the indent after a set of keywords passed as a regexp argument to the command.

This was made using the new find and replace in https://github.com/occivink/kakoune-find :star2:

### DIFF
--- a/rc/base/css.kak
+++ b/rc/base/css.kak
@@ -45,8 +45,7 @@ define-command -hidden css-filter-around-selections %{
 
 define-command -hidden css-indent-on-new-line %[
     evaluate-commands -draft -itersel %[
-        # preserve previous line indent
-        try %[ execute-keys -draft \; K <a-&> ]
+        preserve-previous-line-indent
         # filter previous line
         try %[ execute-keys -draft k : css-filter-around-selections <ret> ]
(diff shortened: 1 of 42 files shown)


mawww:
Hello,

Unfortunately, I will consider introducing such helpers only once we have a solution to handle dependency between scripts. I described an initial approach that I am rather happy with in #1767, it still lacks a nice way to source a script without knowing its exact path.

Once we have that, we can start looking at providing a nice set of indentation helpers that can be used in all the support scripts.

danr:
This does not require any particular loading order: only that `indent.kak` will be loaded before an `InsertChar` is triggered for a particular filetype.

mawww:
This does however require that the `indent.kak` script gets loaded at some point, which is already a dependency problem (if you just add symlinks to some language support file, auto indent will break with no clear reason, so the user now needs to be aware of this dependency to manage his config).

danr:
Ok! Closing this until we have another solution then.

PR #1812, opened 2018-01-25
src: Allow lone \E atoms in regular expressions

lenormf:


lenormf:
The last two commits allow `*` to be used on selections containing the `\E` atom (bug reported by @occivink). There's probably a better way to implement that, but the idea is there.

lenormf:
Marking as WIP, as there's still cases not handled by the current implementation.

lenormf:
I removed the two trailing commits, the feature they implemented is not-trivial and needs more thinking.

mawww:
I dont see a good reason to merge that, closing.

lenormf:
Back when <kbd>*</kbd> still used `\Q` and `\E`, it was sometimes tedious to reuse the (modified) pattern because I had to remove both sequences.

With that commit, I only had to remove `\Q`, and it sort of made sense as a shortcut to allow `\E` to be by itself. Now I don't really need this anymore.

PR #1815, opened 2018-01-26
Add [scratch] indicator in context info

Delapouite:
Hi
This indication may seems redundant at first because the *default* scratch buffer is unambiguously called `*scratch*`.
But it is useful for subsequent buffers with arbitrary names created by the `:e -scratch` command.


src/client.cc
```diff
@@ -130,10 +130,12 @@ String generate_context_info(const Context& context)
         s += "[+]";
     if (context.client().input_handler().is_recording())
         s += format("[recording ({})]", context.client().input_handler().recording_reg());
-    if (context.buffer().flags() & Buffer::Flags::New)
-        s += "[new file]";
     if (context.hooks_disabled())
         s += "[no-hooks]";
+    if (context.buffer().flags() == Buffer::Flags::None)
```
lenormf: I think you should rather check that there's no `File` flag set here, rather than expect no flags at all to be set.

Delapouite: I took inspiration from here https://github.com/mawww/kakoune/blob/8e3e5b10c1e05cc174a8c1ddf44dc7bc3fd095e4/src/buffer_manager.cc#L87

lenormf: To me that line creates a buffer that has no flag, it doesn't mean that all scratch buffers don't have any flags (in theory).

mawww: Yep, I think its the `File` flag that should be checked, as we could have various combinations (`NoUndo`, `ReadOnly`, `Debug`)... Should a `debug` buffer be considered a `scratch` buffer as well ?

lenormf: I think a debug buffer is a particular type of scratch buffer, it's probably better to let that be.

mawww: I tend to see them as that as well, but do we want to have both `[debug]` and `[scratch]` displayed for them ? seems a bit redundant.

Delapouite:
Thanks for your advice. `[scratch]` indicator is now based on the `File` flag with the exception of `Debug` to avoid `[scratch][debug]`.

PR #1819, opened 2018-01-28
Various improvements to Homebrew formula

alyssais:
* Add missing Homebrew dependencies on Linux
* `require "formula"` is unnecessary
* `brew audit --strict` warns on single quotes
* Remove unnecessary option

### DIFF
--- a/contrib/kakoune.rb
+++ b/contrib/kakoune.rb
@@ -1,17 +1,20 @@
-require "formula"
-
 class Kakoune < Formula
   homepage "https://github.com/mawww/kakoune"
   head "https://github.com/mawww/kakoune.git"
 
-  depends_on 'docbook-xsl' => :build
-  depends_on 'ncurses' => [:build, :recommended]
-  depends_on 'asciidoc' => [:build, 'with-docbook-xsl']
+  depends_on "docbook-xsl" => :build
+  depends_on "ncurses" => [:build, :recommended]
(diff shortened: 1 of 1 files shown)


mawww:
Hello,

Thanks for those improvements, I do not think I have a copyright waiver from you yet, could you add one (see the CONTRIBUTING file).

Cheers,

Maxime.

contrib/kakoune.rb
```diff
@@ -1,17 +1,20 @@
-require "formula"
-
 class Kakoune < Formula
   homepage "https://github.com/mawww/kakoune"
   head "https://github.com/mawww/kakoune.git"
 
-  depends_on 'docbook-xsl' => :build
-  depends_on 'ncurses' => [:build, :recommended]
-  depends_on 'asciidoc' => [:build, 'with-docbook-xsl']
```
lenormf: Why was this option removed?

lenormf:
I think all three commits could be squashes as well?

contrib/kakoune.rb
alyssais: Because it [no longer](https://github.com/Homebrew/homebrew-core/commit/90aae3e04c9af9d834c16c6820aff930fed5c3f4) exists.

alyssais:
Added the copyright waiver and squashed

PR #1824, opened 2018-01-31
Add justfile support

TeddyDD:
[Just](https://github.com/casey/just) is cool little make alternative (more like command runner than build tool)

### DIFF
--- a/rc/extra/just.kak
+++ b/rc/extra/just.kak
@@ -1,0 +1,55 @@
+# Detection
+# ‾‾‾‾‾‾‾‾‾
+
+hook global BufCreate .*/?[jJ]ustfile %{
+    set-option buffer filetype justfile
+}
+
+# Indentation
+# ‾‾‾‾‾‾‾‾‾‾‾
+
+define-command -hidden just-indent-on-new-line %{
+     evaluate-commands -draft -itersel %{
+        # preserve previous line indent
+        try %{ execute-keys -draft \;K<a-&> }
+        # cleanup trailing white spaces on previous line
+        try %{ execute-keys -draft k<a-x> s \h+$ <ret>"_d }
+        # copy '#' comment prefix and following white spaces
+        try %{ execute-keys -draft k <a-x> s ^\h*//\h* <ret> y jgh P }
+    }
+}
+
+# Highlighters
+# ‾‾‾‾‾‾‾‾‾‾‾‾
+
+add-highlighter shared/ regions -default content justfile \
+    comment  '#' '$'  ''                 \
+    double_string '"' (?<!\\)(\\\\)*" '' \
+    single_string "'" (?<!\\)(\\\\)*' '' \
+    inline   '`' '`' ''                  \
+    shell    '^\h+' '^[^\h]' ''
+
+add-highlighter shared/justfile/comment fill comment
+add-highlighter shared/justfile/single_string fill string
+add-highlighter shared/justfile/double_string fill string
+add-highlighter shared/justfile/shell ref sh
+add-highlighter shared/justfile/inline ref sh
+
+add-highlighter shared/justfile/content regex '^(@)?([\w-]+)(?:\s(.+))?\s?(:)(.+)?$' 1:operator 2:function 3:value 4:operator 5:type
+add-highlighter shared/justfile/content regex '([=+])' 1:operator
+add-highlighter shared/justfile/content regex '^([\w-]+)\s=' 1:value
+
+hook global WinSetOption filetype=justfile %{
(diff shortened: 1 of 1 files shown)


rc/extra/just.kak
```diff
@@ -0,0 +1,61 @@
...
+        # copy '#' comment prefix and following white spaces
+        try %{ execute-keys -draft k <a-x> s ^\h*//\h* <ret> y jgh P }
+    }
+}
+
+# Highlighters
+# ‾‾‾‾‾‾‾‾‾‾‾‾
+
+add-highlighter shared/ regions -default content justfile \
+	comment  '#' '$'  ''                 \
```
lenormf: Can you indent the lines with 4 spaces please?

rc/extra/just.kak
```diff
@@ -0,0 +1,61 @@
...
+    double_string '"' (?<!\\)(\\\\)*" '' \
+    single_string "'" (?<!\\)(\\\\)*' '' \
```
lenormf: Small indenting glitch here.

rc/extra/just.kak
```diff
@@ -0,0 +1,61 @@
...
+add-highlighter shared/justfile/inline ref sh
+
+#                                                                                    @          recipe     params  :          depends
+add-highlighter shared/justfile/content regex '^(@)?([\w-]+)(?:\s(.+))?\s?(:)(.+)?$' 1:operator 2:function 3:value 4:operator 5:type
+add-highlighter shared/justfile/content regex '([=+])' 1:operator
+add-highlighter shared/justfile/content regex '^([\w-]+)\s=' 1:value
+
+#hook -group justfile-highlight global WinSetOption filetype=justfile %{
+    #add-highlighter window ref justfile
+#}
```
lenormf: If this hook isn't needed, remove it from the file.

rc/extra/just.kak
```diff
@@ -0,0 +1,61 @@
...
+	inline   '`' '`' ''                  \
+	shell    '^\h+' '^[^\h]' ''
+
+add-highlighter shared/justfile/comment fill comment
+add-highlighter shared/justfile/single_string fill string
+add-highlighter shared/justfile/double_string fill string
+add-highlighter shared/justfile/shell ref sh
+add-highlighter shared/justfile/inline ref sh
+
+#                                                                                    @          recipe     params  :          depends
```
lenormf: If the current regex engine allows, you could name the capture groups in the regular expression directly, to avoid annotating the faces. Not sure it's supported though, if not maybe we could ask for it!

rc/extra/just.kak
```diff
@@ -0,0 +1,61 @@
...
+    #add-highlighter window ref justfile
+#}
+
+hook global WinSetOption filetype=justfile %{
+    hook window InsertChar \n -group justfile-indent just-indent-on-new-line
+    add-highlighter window ref justfile
+
+    # hack to add global variable substititution highlighting
+    add-highlighter buffer group justfile
+    add-highlighter buffer/justfile regex '(\{{2})([\w-]+)(\}{2})' 1:operator 2:variable 3:operator
```
lenormf: I think all highlighters should be set on the window, rather than the buffer.

TeddyDD: This hack allows to highlight `{{VALUE}}` variable substitutions in sh regions. I don't know how to do this better, any ideas?

rc/extra/just.kak
TeddyDD: I tried this `regex '(?<operator>[=+])'`, it returns `'add-highlighter' wrong parameter count`.

rc/extra/just.kak
TeddyDD: whoops 😊

rc/extra/just.kak
lenormf: I think `add-highlighter shared/justfile/shell regex […]` should be enough, without need for declaring a hook in the `buffer` scope.

TeddyDD: Thanks, works like a charm 👍

TeddyDD:
BTW: Let me know if you prefer squashed commits

lenormf:
Yes, since only a single file was modified, it's better to squash the commits.

TeddyDD:
I squashed commits. Is there anything else I should change?

mawww:
Did we loose the copyright waiver commit in the squashing ? Can you add that commit back please ?

TeddyDD:
:man_facepalming: looks like chanses I do rebasing correctly in the first try are close to zero.
Sorry, added copyright waiver back :)

PR #1826, opened 2018-02-01
Add check for tmux compiled from master.

nochiel:
The expr incorrectly sets VERSION_TMUX to the empty string if tmux -V == master causing the shell script to fail to parse.  
tmux -V == master if compiled directly from repository.

### DIFF
--- a/rc/extra/tmux-repl.kak
+++ b/rc/extra/tmux-repl.kak
@@ -5,9 +5,10 @@ hook global KakBegin .* %{
     %sh{
         if [ -n "$TMUX" ]; then
             VERSION_TMUX=$(tmux -V)
-            VERSION_TMUX=$(expr "${VERSION_TMUX}" : 'tmux \([0-9]*\).*')
+            VERSION_TMUX=$(expr "${VERSION_TMUX}" : 'tmux \([0-9]*\|master\)')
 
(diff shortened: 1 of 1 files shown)


lenormf:
I think that's a bug from the `tmux` package of your distribution, you should contact its maintainer instead.

nochiel:
> @lenormf: I think that's a bug from the tmux package of your distribution, you should contact its maintainer instead.

I meant, if you install, as per the tmux [readme](https://github.com/tmux/tmux/blob/master/README):
```sh
# To get and build the latest from version control:

git clone https://github.com/tmux/tmux.git
cd tmux
sh autogen.sh
./configure && make
```
And tmux master [configure.ac](https://github.com/tmux/tmux/blob/master/configure.ac#L3) has:
```ac
AC_INIT(tmux, master)
```

nochiel:
> @lenormf: Note: your commit breaks the functionality, the parens in the regular expression are mandatory to extract the data they capture.

Drat. The code was a bit confusing. The `if [ "${VERSION_TMUX}" -gt 1 ]; then` didn't seem to correspond to the error message: `expected >= 2.x` As a result I thought that the intent was to use expr to return number of characters matched.  In which case `[0-9]` should have been `[2-9]`, I think.

PR #1830, opened 2018-02-04
Implement a plugins pages

lenormf:
This commit adds a page to the navigation bar that will list
repositories on Github that have the "kakoune" topic set. The order
of the repositories is decided by Github (stars), and an attacker
could have their rogue plugin listed on the page, therefore users
are warned to use the plugins at their own risks.

The page works on mobile as well.

Preview:


TeddyDD:
I think you might want to exclude repo mawww/kakoune from the result, besides this I love the idea ❤️

lenormf:
There are a few false positives because the only criteria is `topic=kakoune`. I'd rather we decide what topics a plugin should have, rather than filter out specific projects from the list fetched from github.

TeddyDD:
It's not big deal, just kinda funny since it's first on the list :)

PR #1844, opened 2018-02-12
lua: highlight nested strings/comments

jacobdufault:


rc/base/lua.kak
```diff
@@ -11,11 +11,12 @@ hook global BufCreate .*[.](lua) %{
...
-add-highlighter shared/ regions -default code lua \
-    string  '"'      (?<!\\)(\\\\)*"  '' \
-    string  "'"      (?<!\\)(\\\\)*'  '' \
-    comment '--\[\[' '\]\]'           '' \
-    comment '--'     '$'              '' \
+add-highlighter shared/ regions -default code -match-capture lua \
+    string  '"'           (?<!\\)(\\\\)*" '' \
+    string  "'"          (?<!\\)(\\\\)*'  '' \
+    string  '\[(=*)\['   '\](=*)\]'       '' \
+    comment '--\[(=*)\[' '\](=*)\]'       '' \
```
MilanVasko: Won't this highlight weird cases like this incorrectly?

``` lua
--[=[
	Hello, this is a comment!
]]
]=]--

function test()
	print("Hello")
end
```
I think the number of `=` somehow needs to be guarded to be the same on the opening and on the closing side, but I'm not sure how it would be implemented.

jacobdufault: image

This is handled by the `-match-capture`. The `(=*)` is capturing the number of `=` entries.

MilanVasko: Oh, right, I've missed that part. That's pretty neat then :)

Issue #1657, opened 2017-10-27
Select "content on line" surrounding object


nuao:
I'd like to suggest an addition to the menu "select inner/outer surrounding object"

This is to select the current line then trim whitespace at either end. Or viewed another way, go to first non-whitespace character on line then extend to final non-whitespace character.

The effect can currently be achieved with `giGl`, but the object menu would be more convenient.

More generally, given the "custom object desc" already exists, it would therefore also be nice to have the ability to customize the list of selectable objects with user-chosen keybindings. This way users could make this type of change independently and have their own list.

Thanks for your consideration.

occivink:
You can actually customize the text objects. For example, you can do 

map -docstring line global object l '<esc>giGl'

The `<esc>` is here to exit the selection menu.

nuao:
Ah, thanks very much. I never knew.
I submitted a possible PR so that the main Readme matches the information given in the map command's info box.

PR #1846, opened 2018-02-13
Add trim_selections primitive

Delapouite:
Hello!

## TL;DR

This PR introduces a new `trim selections` primitive bound to `_` key.
Ex: `%(   foo  )` -> `%(foo)`

## Rational

Kakoune already provides quite a few useful primitives to deal with blank chars:

`&`: align selection, align the cursor of selections by inserting spaces before the first character of the selection
`<a-&>`: copy indent, copy the indentation of the main selection (or the count one if a count is given) to all other ones
`@`: convert tabs to spaces in current selections, uses the buffer tabstop option or the count parameter for tabstop
`<a-@>`: convert spaces to tabs in current selections, uses the buffer tabstop option or the count parameter for tabstop
`<a-j>`: join selected lines
`<a-J>`: join selected lines and select spaces inserted in place of line breaks
…

Surprisingly there are no easy way (if I'm not mistaken) to omit white-spaces on both sides of selections.
I think this is a common operation that deserves to be built-in.

### Markdown / ASCII tables editing

```
+|       |       |        |                |
+|:------|------:|:------:|----------------|
+|%(foo )|%( bar)|%( qux )|%( pop )%( hip )|
```

When you need to only grab the content of cells. `<a-i>|` and then trim.

### Copy/paste

Often I need to copy an indented line, and paste in in the middle of another one. Since `x` select the starting spaces and the final `\n`, it's cumbersome. Now you can do `x_`, which again only select the *real meat* of the line to be pasted elsewhere.

### Content rotation of multiple lines

Select a paragraph with lines indented at different level:

```
line 1
  line 2
    line 3
```
to rotate the lines and respect the original indentation: `%<a-s>_<a-">`. Split by lines, trim and rotate.

```
line 3
  line 1
    line 2
```

I'm sure we'll find many more interesting use-cases along the way. It may even offers interesting stuff to simplify existing scripts.

## Relation to text objects

I've always been surprised that there's no `line` text-objects.  Some kind of `w` vs `<a-i>w` symmetry.
One of the reason is that the `x` primitive is a first class citizen and almost equivalent to `<a-a>x` (if we decide that here the `x` means `line`). But what about `<a-i>x`? "select inside line"? In fact this issue was raised not so long ago: https://github.com/mawww/kakoune/issues/1657

With the `trim` primitive you can emulate this behavior with a `x_`. The added benefits is that we reinforce the orthogonality of kakoune design.

## Precedents

Almost every programming languages has some kind of `trim` primitive in their standard library: 
JavaScript `String.prototype.trim`, Python and Ruby `strip()` etc…

The `vis` editor (kakoune's cousin) already provides this feature: https://github.com/martanne/vis/blob/5963289d0257c7bd422c3cf54866a846168f5c7a/config.def.h#L154
It's bound to the `underscore` key which is free in kakoune and is therefore a perfect candidate.

### User land

Can this be done in a `kak` script? Well kind of. Here's one solution I came up but which is not very satisfying:
```
# map global normal _ '1s\A\s*(.*?)\s*\z<ret>'  -docstring 'trim'
```
As is, it pollutes the search register and display annoying messages in the status bar. (Perfs may be lame as well in comparaison of a native implementation). There may be better solutions of course, but
again the trim primitive is so basic it should be available out of the box for everyone without this hacky regex magic. And you get the auto-info box for free ;)

## Implementation

In other PRs I've already proven that I'm far from being proficient in C++.
So I hope the code is almost good this time. It's small so it does not bloat the editor. As usual mawww, don't hesitate to fix any obvious mistake.

Thanks for reading dear kakouners and I hope you like this proposal.


aver-d:
Thanks for the write-up.
Another nice use for this would be splitting comma-separated values (lists, arrays, arguments…) into multiple selections:

    <a-i>bS,_

Would be enough to have three selections `a b c` in both these inconsistently-spaced cases below:

    ( a,  b,c)

    (
      a, b  ,
     c
    )

aver-d:
Actually… just playing around here I see the inner object argument selection `<a-i>u` would already handle those examples I gave.
But I still like the trim idea and think @Delapouite makes a good case.

occivink:
Regarding the mapping, you can make it not pollute the search register by doing prepending `_` to `1s`. But overall I agree that this is a useful primitive.

Delapouite:
Thanks for giving this proposal a chance mawww.
I think I share your concern about feature creep in kakoune.

For example that's why the two major PRs I proposed recently, the one about `ModeChange` hook and `declare-user-mode`, share the same vision of trying to find common denominators on which other stuff can be expressed.
In the case of `ModeChange` it was a common ground to encompass `*Begin/*End` hooks and for `declare-user-mode` we could potentially remove the builtin `user` mode and give power to the user to declare many custom ones.

So I believe that trimming selections belong to the same end of the feature spectrum: a fundamental brick that will benefit many user land commands. It's very generic and by opposition to features too specific that I would consider feature creep like a command to "turn text to PascalCase" or a command to "toggle emojis from their unicode representation"…

lenormf:
Is that used so often that it warrants to get its own -scarce&precious- mapping? Is it used in the builtin scripts?

Delapouite:
This rhetoric about "-scarce&precious-" is getting old.
Why keeping free keyboard space if we never use it in the end?
Do you have a hidden killer feature you're expecting to unleash in 2050?
Sorry if I sound rude or pedantic, it's not my intention. I'm genuinely interested about your aim.

lenormf:
If a trivial feature is implement-able using a mapping in the user's configuration file, then it should probably be. If we ever want to use the `_` key for something not-trivial to implement in the future but it has already been bound, then it will be tough on users who got used to the first side effect. Even more so if more mappings are based on it (e.g. alt-$key, alt-shift-$key), that will also need to be (re)moved because the original primitive they're based on is gone and their key mapping doesn't make sense anymore.

Until now, trivial mappings that made it into the builtin set of primitives are those that were so frequently used that it was worth it to make them builtin (e.g. `<a-s>`), so I'm asking if this particular feature is used so often that it deserves that same treatment.

mawww:
@lenormf I am not sure we need to make sure a feature is commonly requested or used before adding it as a built-in, I think that if we have a strong hunch that it would become a commonly used feature once available then it would be good enough. Its not exactly the same thing, although it is close. For example, I believe people trim whitespaces out of their selections pretty often without really considering that they are doing that (I often use `e` instead of `w` because I know I do not want trailing whitespaces, not to say `_` would be superior in that case, but that made me realise I already often care about surrounding spaces and try to avoid them).

Another use case is trimming whitespaces before using `*`... The automatic behaviour was pretty well received, and kept for a long time until I reverted it recently.

I am still a bit hesitant, but it seems like a reasonable bult-in, and the implementation is very low complexity.

PR #1866, opened 2018-02-22
kakrc: Allow `colorscheme` to search directories recursively

lenormf:


share/kak/kakrc
```diff
@@ -9,14 +9,23 @@ def -params 1 -docstring "colorscheme <name>: enable named colorscheme" \
       done
   } \
   colorscheme %{ %sh{
-    if [ $# -eq 1 ]; then
-        if [ -f "${kak_config}/colors/$1.kak" ];then
-            echo "source '${kak_config}/colors/$1.kak'"
+    find_colorscheme() {
```
mawww: How about using a single find command with multiple paths (according to the posix manpage, find can take multiple path to search in, so we dont need to invoke find multiple times, and hence dont need that function definition).

lenormf: I think the order is important here, we don't want a colorscheme stored in `runtime` to be picked over the same file stored in `config` (to allow users to override a builtin theme). I don't see anything related to output order in the [POSIX man page](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/find.html), so I don't think we can expect files found in the first directory to be returned first.

mawww: fair enough, then another complaint, completion needs to be changed as well :)

PR #1871, opened 2018-02-23
faq: Document how to render colors properly in `tmux`

lenormf:
Hi,

I was on the fence between writing a wiki page or a FAQ entry and/or add the terminfo file to `contrib`… Let me know if this is too specific to be merged in the project directly.

HTH.

### DIFF
--- a/doc/pages/faq.asciidoc
+++ b/doc/pages/faq.asciidoc
@@ -45,9 +45,30 @@ This variable should be set by the terminal emulator, and not overridden with
 an arbitrary value, otherwise it might interfere with general UI rendering on
 the terminal's window.
 
-Note that if you're using `tmux`, the proper -official- way to get Kakoune to
-work is to add the following line to your `tmux` configuration file:
-`set -sg default-terminal screen-256color`
+== I'm using `tmux` and colors look weird
+
+If you're using a tool that doesn't support "palette switching", colors will
+still be slightly off: they are being rounded down to values supported by the
+regular color palette by Kakoune. Fortunately, on recent versions of `tmux`
+(>=2.6), you can work around that by using the following configuration
+settings:
+
+~/.tmux.conf
+```
+set -g default-terminal "tmux-256color"
+set -ga terminal-overrides ",*col*:Tc"
+```
+
+Note: `\*col*` is a regular expression that matches your terminal's `$TERM`
+value, modify it if necessary, e.g. `xterm-termite`.
+
+Next, run the following command to create a local `terminfo` override:
+
+```
+$ tic /path/to/kakoune/contrib/tmux-256color.terminfo
+```
+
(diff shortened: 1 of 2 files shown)


doc/pages/faq.asciidoc
```diff
@@ -45,9 +45,30 @@ This variable should be set by the terminal emulator, and not overridden with
 an arbitrary value, otherwise it might interfere with general UI rendering on
 the terminal's window.
 
-Note that if you're using `tmux`, the proper -official- way to get Kakoune to
-work is to add the following line to your `tmux` configuration file:
-`set -sg default-terminal screen-256color`
+== I'm using `tmux` and colors look weird
+
+If you're using a tool that doesn't support "palette switching", colors will
```
Screwtapello: By "tool", do you mean "terminal emulator"?

Screwtapello:
I've [poked](https://lists.gnu.org/archive/html/bug-ncurses/2018-02/msg00029.html) the ncurses maintainer to make this change upstream, so Kakoune won't need to ship a custom terminfo blob forever.

doc/pages/faq.asciidoc
lenormf: Well, `tmux` isn't a terminal emulator and yet it has its own terminfo file and has to implement OSC4 itself, so I /think/ I stand correct?

lenormf:
I thought about reporting this to `ncurses` directly as well, but then thought that it would take forever to have that merged and propagated upstream, so I'd rather have that explanation merged now (and remove it later, when the time comes).

doc/pages/faq.asciidoc
Screwtapello: To my mind, anything that interprets terminal escape sequences like OSC4 is a terminal emulator; `xterm` is a teminal emulator that draws its output with X11, PuTTY is a terminal emulator that draws its output with Win32, and `tmux` is a terminal emulator that draws its output with ncurses.

Saying "tool" here makes me think of other tools that might or might not support palette switching; Kakoune is a tool that does (in that it will try to set palette entries if it can), Vim (as far as I know) is a tool that does not, but those tools aren't relevant to this section of the docs. Ruling them out with a phrase like "terminal emulator" (or something similar) would make it clearer to me.

Screwtapello:
The tmux maintainer [replied](https://lists.gnu.org/archive/html/bug-ncurses/2018-02/msg00030.html) to my email saying "pls no update"; apparently the Real Solution is going to involve waiting for ncurses 6.1 with official 24-bit colour support to be widely available.

So yes, I agree this PR represents the current best practice for getting Kakoune colours to work properly inside tmux.

lenormf:
Just force-pushed to escape `\*col*`, which was rendered as `*col*`.

lenormf:
Anything blocking here? I can move all this to the wiki if it won't be merged in the FAQ directly.

PR #1875, opened 2018-02-23
Display number of combined/saved/restored selections in status

Delapouite:
Hi

This PR provides more feedback when dealing with selection lists.
It follows convention found in previous status messages like:

https://github.com/mawww/kakoune/blob/master/src/normal.cc#L604
```
context.print_status({ format("yanked {} selections to register {}",
                                  context.selections().size(), reg),
                           get_face("Information") });
```

### DIFF
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -1740,7 +1740,8 @@ void combine_selections(Context& context, SelectionList list, Func func)
                                  else
                                  {
                                      if (list.size() != sels.size())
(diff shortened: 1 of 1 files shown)


lenormf:
That's something I've been wondering about: should all primitives that have a side effect show a message? E.g. <kdb>y</kbd> shows a message, but <kdb>p</kbd> or <kbd>r</kbd> based primitives don't; <kbd>$</kbd> doesn't show any message either.

PR #1876, opened 2018-02-24
[POC] Implement a `-F` flag

lenormf:
Hi,

There was some interest in this feature, so here is the PoC I came up with. An alternative was also to simply allow commands executed with `-p` to return their output to the client, but I don't think that's something we'd want.

### DIFF
--- a/doc/kak.1.asciidoc
+++ b/doc/kak.1.asciidoc
@@ -7,7 +7,7 @@ kak - a vim inspired, selection oriented code editor
 
 SYNOPSIS
 --------
-*kak* [-help] [-q] [-n] [-l] [-ro] [-clear] [-ui ui_type] [-e command] [-E command] [-f keys] [-p session_id] [-c session_id|[[-d] -s session_id] [+line[:column]|+:] file ...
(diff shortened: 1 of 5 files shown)


mawww:
Changing `fd_readable` not to use a zero timeout is going to break Kakoune, we need fd_readable to be non blocking if we want Kakoune to stay responsive to multiple clients.

I am not sure what the use case for that `-F` flag is, I assume it is a way to get some data out  of Kakoune from scripts, but thats already provided by env vars, and assuming there is a legitimate use cases where env vars are not good enough, `-F` is not going to be of much help as it runs in a very limited context (no client, no window, no buffer...). I think if we would need something similar to `eval -client <blah>` for this to make sense.

lenormf:
A UI could extract information from the underlying server, to e.g. reconstruct a modeline without relying on the already expanded string sent by the server.

mawww:
As is, this is of very limited usefulness, I agree the potential for such a functionality would be quite nice, and its worth exploring what can be done, but I'll close that for now until we come back with a stronger design.

PR #1882, opened 2018-02-25
Ease reading of command-line options

aver-d:
This change displays command-line options in grid format. Each parameter is indented with two spaces and then padded to maintain vertical alignment of each description.

I think the visual spacing makes the options much easier to read and gives a friendlier appearance. This is particularly important for people new to Kakoune who use `-help` as a way to become familiar with the program.

Existing:

    Options:
    -c <arg>: connect to given session
    -e <arg>: execute argument on client initialisation
    -E <arg>: execute argument on server initialisation
    -n : do not source kakrc files on startup
    -s <arg>: set session name
    -d : run as a headless session (requires -s)
    -p <arg>: just send stdin as commands to the given session
    -f <arg>: act as a filter, executing given keys on given files
    -i <arg>: backup the files on which a filter is applied using the given suffix
    -q : in filter mode, be quiet about errors applying keys
    -ui <arg>: set the type of user interface to use (ncurses, dummy, or json)
    -l : list existing sessions
    -clear : clear dead sessions
    -ro : readonly mode
    -help : display a help message and quit

Proposed with this change:

    Options:
      -c <arg>   connect to given session
      -e <arg>   execute argument on client initialisation
      -E <arg>   execute argument on server initialisation
      -n         do not source kakrc files on startup
      -s <arg>   set session name
      -d         run as a headless session (requires -s)
      -p <arg>   just send stdin as commands to the given session
      -f <arg>   act as a filter, executing given keys on given files
      -i <arg>   backup the files on which a filter is applied using the given suffix
      -q         in filter mode, be quiet about errors applying keys
      -ui <arg>  set the type of user interface to use (ncurses, dummy, or json)
      -l         list existing sessions
      -clear     clear dead sessions
      -ro        readonly mode
      -help      display a help message and quit

### DIFF
--- a/src/parameters_parser.cc
+++ b/src/parameters_parser.cc
@@ -7,11 +7,26 @@ namespace Kakoune
 
 String generate_switches_doc(const SwitchMap& switches)
 {
+    Vector<int> lengths(switches.size());
+    int i = 0;
+    for (auto& sw : switches) {
+        lengths[i++] = (int)sw.key.length() + (sw.value.takes_arg ? 5: 0);
+    }
+    int maxlen = *std::max_element(lengths.begin(), lengths.end());
+
     String res;
-    for (auto& sw : switches)
-        res += format("-{} {}: {}\n", sw.key,
+    i = 0;
+    for (auto& sw : switches) {
+        int len = lengths[i++];
+        String pad = "  ";
+        while (len++ < maxlen)
+            pad += ' ';
+        res += format("  -{} {}{}{}\n",
+                      sw.key,
                       sw.value.takes_arg ? "<arg>" : "",
+                      pad,
                       sw.value.description);
+    }
     return res;
 }
 


src/parameters_parser.cc
```diff
@@ -7,11 +7,26 @@ namespace Kakoune
...
+
     String res;
-    for (auto& sw : switches)
-        res += format("-{} {}: {}\n", sw.key,
+    i = 0;
+    for (auto& sw : switches) {
+        int len = lengths[i++];
+        String pad = "  ";
+        while (len++ < maxlen)
+            pad += ' ';
```
mawww: Use `String(Codepoint, CharCount)` constructor here, `const String pad{' ', maxlen - len + 1}`

src/parameters_parser.cc
```diff
@@ -7,11 +7,26 @@ namespace Kakoune
 
 String generate_switches_doc(const SwitchMap& switches)
 {
+    Vector<int> lengths(switches.size());
```
mawww: I would avoid storing lengths, they are simple to compute, I suggest to write a small named lambda and use ranges:
```
auto switch_len = [](auto& sw) { return sw.key.length() + (sw.value.take_arg ? 5 : 0); };
auto switches_len = switches | transform(switch_len);
const CharCount maxlen = *std::max_element(switches_len.begin(), switches_len.end());
```

Reuse that switch_len lambda in the second loop to get the len. I believe the extra computations will be largely cheaper than allocating a vector.

src/parameters_parser.cc
```diff
@@ -7,11 +7,26 @@ namespace Kakoune
...
+    int i = 0;
+    for (auto& sw : switches) {
+        lengths[i++] = (int)sw.key.length() + (sw.value.takes_arg ? 5: 0);
+    }
+    int maxlen = *std::max_element(lengths.begin(), lengths.end());
```
mawww: This will lead to a memory error if switches are empty.

aver-d:
Hi and thanks for considering this.

Sorry, the reason for not doing the check on switches length is because I'd thought this function was only for command-line parameters (which are known in advance). I hadn't noticed that this same function is also called within `command_manager.cc`. Sorry about that, and, of course, you're right that this should be fixed.

Anyway, you'll see that C++ isn't one of my languages, explaining my "old-fashioned" implementation with minimal use of language features.

In the same spirit, I'd normally just add this at the top:

    if (switches.empty())
        return "";

But I'm of course happy to change the implementation to whatever you think best. I wanted to use your code exactly as you gave (only replacing a `take_arg`, `takes_arg` typo)

    auto switch_len = [](auto& sw) { return sw.key.length() + (sw.value.takes_arg ? 5 : 0); };
    auto switches_len = switches | transform(switch_len);
    const CharCount maxlen = *std::max_element(switches_len.begin(), switches_len.end());

but the compiler does not seem to like it:

    /usr/include/c++/7.3.0/bits/stl_algo.h:5654:13: error: use of deleted function ‘Kakoune::TransformView<const Kakoune::HashMap<Kakoune::String, Kakoune::SwitchDesc, (Kakoune::MemoryDomain)9>&, Kakoune::generate_switches_doc(const SwitchMap&)::<lambda(auto:8&)> >::Iterator& Kakoune::TransformView<const Kakoune::HashMap<Kakoune::String, Kakoune::SwitchDesc, (Kakoune::MemoryDomain)9>&, Kakoune::generate_switches_doc(const SwitchMap&)::<lambda(auto:8&)> >::Iterator::operator=(const Kakoune::TransformView<const Kakoune::HashMap<Kakoune::String, Kakoune::SwitchDesc, (Kakoune::MemoryDomain)9>&, Kakoune::generate_switches_doc(const SwitchMap&)::<lambda(auto:8&)> >::Iterator&)’
and

    ranges.hh:122:12: error: non-static reference member ‘const Kakoune::TransformView<const Kakoune::HashMap<Kakoune::String, Kakoune::SwitchDesc, (Kakoune::MemoryDomain)9>&, Kakoune::generate_switches_doc(const SwitchMap&)::<lambda(auto:8&)> >& Kakoune::TransformView<const Kakoune::HashMap<Kakoune::String, Kakoune::SwitchDesc, (Kakoune::MemoryDomain)9>&, Kakoune::generate_switches_doc(const SwitchMap&)::<lambda(auto:8&)> >::Iterator::m_view’, can’t use default assignment operator

amidst other interesting reading.

However… I'm not looking to take up your time with this. I won't be offended if you either close this PR and add the code yourself, or, alternatively, since it's only a short function, you could just copy and paste back here the complete implementation you're happy with and I'll add that directly.

PR #1885, opened 2018-02-26
Fix broken rust racer support (plus tweaks)

marcesquerra:
I wanted to give rust editing with kakoune a try and I found out that racer integration was broken. With a bit of experimentation I found out what was broken: awk is adding an extra ':' at the end of the list of completions.

And working on it I couldn't resist the temptation to try and see if I could improve a bit how the racer suggestions where presented :)

By the way, since this is my first PR, let me congratulate you for what you have done with kakoune, it's amazing!

### DIFF
--- a/rc/extra/racer.kak
+++ b/rc/extra/racer.kak
@@ -45,7 +45,7 @@ define-command racer-complete -docstring "Complete the current selection with ra
                 }'
             )
             printf %s\\n "evaluate-commands -client '${kak_client}' %{
-                set-option buffer=${kak_bufname} racer_completions %@${compl}@
+                set-option buffer=${kak_bufname} racer_completions %@${compl: : -1}@
             }" | kak -p ${kak_session}
             rm -r ${dir}
         ) > /dev/null 2>&1 < /dev/null &
--- a/rc/extra/racer.kak
+++ b/rc/extra/racer.kak
@@ -29,13 +29,40 @@ define-command racer-complete -docstring "Complete the current selection with ra
                     gsub(/\|/, "\\|", menu)
                     if (type == "Function") {
                         sub(word, "{default+e}" word "{default+d}", menu)
+                        gsub(/^fn /, "    fn ", menu)                            # The extra spaces are there to vertically align
+                                                                                 # the type of element on the menu to make it easy
+                                                                                 # to read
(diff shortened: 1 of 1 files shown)


rc/extra/racer.kak
```diff
@@ -45,7 +72,7 @@ define-command racer-complete -docstring "Complete the current selection with ra
                 }'
             )
             printf %s\\n "evaluate-commands -client '${kak_client}' %{
-                set-option buffer=${kak_bufname} racer_completions %@${compl}@
+                set-option buffer=${kak_bufname} racer_completions %@${compl: : -1}@
```
ul: This doesn't work on MacOS. Could we do POSIX style, like `${compl%?}` ?

mawww: Yeah, sorry I did not notice that during review, `${blah:<pos>:<len>}` is a bash extension, seems here you want to remove the last character, so `${compl%?}` should do the same portably. I dont have racer setup here, so I cannot test. Could you validate that this works ?

lenormf: If the last character is unknown, you could still use the following:

```
expr "${compl}" : '\(..*\).$'
```

marcesquerra: I don't have much bash experience (I had to google for this solution), so...

@mawww I've tried the `${compl%?}` approach and seems to work on linux (I haven't been able to test on this on a mac).

Question for you all: the problem is that awk adds an extra `:` (the row separator) at the end, do you think something more specific that ensures that what's removed is actually a `:` is needed?

marcesquerra:
I've created another PR to try and fix the problem on Mac: #1903 

The conversation can follow there!

rc/extra/racer.kak
ul: I tested `${compl%?}` in on Mac, it does work there.

ul: I don't speak awk, but I guess this line `gsub(/:/, "\\:", candidate)` (https://github.com/marcesquerra/kakoune/blob/f6a3fd2e4ed8857ffb1900fbb17370eaab277500/rc/extra/racer.kak#L70) adds `:` after every candidate, therefore we could be pretty sure that `:` is always the last character, except of empty candidates list case.

Issue #1521, opened 2017-07-28
Question: modify buffer's content on create and write


atomrc:
I'd like to code two simple hooks to handle gpg files. I'd like to decrypt them on `BufCreate` and encrypt on `BufWritePre`, but I am not sure what I do is the right way to do this

Here is the gist

```
hook global BufCreate .*[.](gpg|pgp|asc) %{
    # decrypt here
    decl -hidden str command "\%di%sh{ gpg -d $kak_buffile }<esc>"
    exec %opt{command}
}

hook global BufWritePre .*[.](gpg|pgp|asc) %{
    # encrypt here
}
```

The exec really feels hacky (and set the buffer state to `modified`) and I don't know how to do the encryption part (do I have to write the encrypted content back to the buffer, write the buffer myself and then reset the content of the buffer to the decrypted content?)

Any pointer on what variables or hooks I could use to achieve that would be appreciated

Thanks guys

mawww:
Not sure about defining that `command` option, note that Kakoune commands are not a script language at all, so decl is not declaring anything like a local variable, just an option.

The best bet is probably to use `|` to pipe the encypted/decrypted buffer content through gpg.

It will still have some rough edges (we'll get the encrypted content in the undo), an alternative for encrypting is to write the decrypted content, then encrypt it in place with gpg in a `BufWritePost`.

We might need a proper hook for modifying the buffer content before we start recording undo, some kind of `BufReadPost` to try to clean that up, but it should be functional without.

nochiel:
@atomrc Would you like to help test PR #1888 ?

tmccombs:
Proposal: how about adding two options which contain shell strings for performing reads and writes.

A `read_action` command which should write the contents of the file to stdout, and a `write_action`, that reads the buffer content and does something with it (like encrypt and store). 

This seems like it would be straightforward to implement, and makes it easy to make plugins for encryption, compression, and even things like editing remote files.

tmccombs:
Oh, and that would avoid having to put the plaintext on the filesystem, as the referenced PR has to.

caksoylar:
[...]

PR #1888, opened 2018-02-26
Enable kak to edit gpg encrypted files.

nochiel:
- [x] Help wanted: testing/code-review.
~- [ ] The script caters to my typical workflow but is it the common case?~
~- [ ] Bug #1889.~
- [x] Enable the options. (currently they don't do anything)
- [x] Make it POSIX.

### DIFF
--- a/rc/extra/gpg.kak
+++ b/rc/extra/gpg.kak
@@ -1,0 +1,65 @@
+# Enable kakoune to open, edit, write files encrypted
+# using [GNU privacy guard](https://www.gnupg.org/)
+
+# options {
+declare-option -docstring 'Set this to true to add a signature when saving the file.' bool gpg_sign false
+declare-option -docstring 'Encrypt for this user id. If not set, it is a assumed that a default was configured in ~/.gnupg/gpg.conf.' str gpg_recipient 
+declare-option -docstring 'Set this to a secure deletion command e.g. shred(1).' str gpg_rmcmd 'rm -f'
+# }
+
+hook -group gpg global BufOpenFile .*\.gpg %{
+    set-option buffer autoreload no
+    evaluate-commands %sh{
+        f="${kak_buffile%.gpg}"
+        umask 066
+        gpg --output "$f" --yes --decrypt "${kak_buffile}"
+
+        if [ 0 -eq $? ]; then
+            printf %s\\n "
+                evaluate-commands %{
+                    edit! '$f'
+                    set-option buffer modelinefmt '[GPG] ${kak_opt_modelinefmt}'
+                    hook -group gpg buffer BufWritePost %val{buffile} gpg-encrypt
+                    hook -group gpg buffer BufClose %val{buffile} gpg-cleanup
+                    echo -markup {Information} Decrypted: '${kak_buffile}'
+                }
+            "
+        else 
+            printf %s\\n "
+                echo -markup {Error} Failed to decrypt ${kak_buffile}
+            "    
+        fi
+    }
+}
+
+define-command -hidden gpg-encrypt %{ evaluate-commands %sh{
+        options=""
+        [ -n "${kak_opt_gpg_recipient}" ] && options="-r '${kak_opt_gpg_recipient}'"
+        [ "${kak_opt_gpg_sign}" = "true" ] && options="--sign ${options}"
+
+        cp -f "${kak_buffile}.gpg" "${kak_buffile}.gpg~" 
+        eval gpg " ${options}" --yes --encrypt "${kak_buffile}"
+
+        if [ 0 -eq $? ]; then
+            printf %s\\n "
+                echo -markup {Information} Encrypted: ${kak_buffile}.gpg
+            "
+        else
+            printf %s\\n "
+                echo -markup {Error} Failed to encrypt: ${kak_buffile}.gpg
+            "
+        fi
+    }
+
+}
+
+define-command -hidden gpg-cleanup  %{ evaluate-commands %sh{
+        if [ -f "${kak_buffile}" ] && [ -n "${kak_opt_gpg_rmcmd}" ]; then
+            eval " ${kak_opt_gpg_rmcmd} '${kak_buffile}'" -- </dev/null >/dev/null 2>&1 &
+        fi
+
+        printf %s\\n "
+            try 'evaluate-commands delete-buffer ${kak_buffile}.gpg'
+        "
+    }
+}


Delapouite:
gpg related: https://github.com/mawww/kakoune/issues/1521

rc/extra/gpg.kak
```diff
@@ -0,0 +1,52 @@
...
+# options {
+declare-option str gpg_recipient 
+declare-option bool gpg_sign false
+# }
+
+hook -group gpg global BufOpenFile .*\.gpg %{
+    set buffer autoreload no
+    %sh{
+        d="$(dirname ${kak_buffile})"
+        f="$(basename -s .gpg ${kak_buffile})"
```
lenormf: Unfortunately, `-s` isn't defined by [POSIX](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/basename.html).

rc/extra/gpg.kak
```diff
@@ -0,0 +1,52 @@
...
+        f="${d}/${f}"
+
+        umask 066
+        gpg --output "$f" --yes --decrypt ${kak_buffile}  2>/dev/null
+        printf %s\\n "
+            # Show in the modeline that this is a gpg file
+            evaluate-commands %{
+                e "$f"
```
lenormf: `edit!` is probably better here.

rc/extra/gpg.kak
```diff
@@ -0,0 +1,52 @@
...
+                echo  -markup {Information} Decrypted: "${kak_buffile}"
+            }
+        "
+    }
+}
+
+define-command -hidden gpg-encrypt %{
+    %sh{
+        options=""
+        [ -n "${gpg_recipient}" ] && options="-r ${gpg_recipient}"
```
lenormf: I think you'll need to quote the recipient, or otherwise to check that it contains valid characters.

rc/extra/gpg.kak
```diff
@@ -0,0 +1,52 @@
...
+        [ "${gpg_sign}" = "true" ] && options="--sign ${options}"
+
+        cp -f "${kak_buffile}.gpg"{,~}
```
lenormf: I'm not sure whether this shortcut is POSIX, is it? If not, prefer the long form.

rc/extra/gpg.kak
```diff
@@ -0,0 +1,52 @@
...
+}
+
+define-command -hidden gpg-encrypt %{
+    %sh{
+        options=""
+        [ -n "${gpg_recipient}" ] && options="-r ${gpg_recipient}"
+        [ "${gpg_sign}" = "true" ] && options="--sign ${options}"
+
+        cp -f "${kak_buffile}.gpg"{,~} 
+        gpg ${options} --yes --encrypt "${kak_buffile}"
```
lenormf: Prefer using `eval` when injecting options into a command.

rc/extra/gpg.kak
```diff
@@ -0,0 +1,52 @@
...
+        printf %s\\n "
+            echo  -markup {Information} Encrypted: ${kak_buffile}.gpg
+        "
+    }
+
+}
+
+define-command -hidden gpg-cleanup  %{
+    %sh{
+        [ -f "${kak_buffile}" ] && shred -u "${kak_buffile}"
```
lenormf: `shred` isn't defined by POSIX, so unfortunately we can't reasonably assume it's going to be installed. You could use `rm`, or maybe define a `gpgrmcmd` option that would hold the command to use to remove a flag.

rc/extra/gpg.kak
```diff
@@ -0,0 +1,52 @@
...
+        printf %s\\n "
+            try 'evaluate-commands delete-buffer ${kak_buffile}.gpg'
```
lenormf: You'll need quoting here.

mawww:
Where are we at with this one ? Its a script I'll be happy to merge (or, should it be provided separately ?).

rc/extra/gpg.kak
```diff
@@ -0,0 +1,57 @@
...
+declare-option -docstring 'Encrypt for this user id. If not set, it is a assumed that a default was configured in ~/.gnupg/gpg.conf.' str gpg_recipient 
+declare-option -docstring 'Set this to a secure deletion command e.g. shred(1).' str gpg_rmcmd 'rm -f'
+# }
+
+hook -group gpg global BufOpenFile .*\.gpg %{
+    set-option buffer autoreload no
+    %sh{
+        f="${kak_buffile%.gpg}"
+        umask 066
+        gpg --output "$f" --yes --decrypt "${kak_buffile}" 2>/dev/null
```
lenormf: The return code should be checked here, and an error printed if the decryption couldn't happen. Also let `gpg` print errors to the debug buffer.

rc/extra/gpg.kak
```diff
@@ -0,0 +1,57 @@
...
+}
+
+define-command -hidden gpg-encrypt %{
+    %sh{
+        options=""
+        [ -n "${kak_opt_gpg_recipient}" ] && options="-r '${kak_opt_gpg_recipient}'"
+        [ "${kak_opt_gpg_sign}" = "true" ] && options="--sign ${options}"
+
+        cp -f "${kak_buffile}.gpg" "${kak_buffile}.gpg~" 
+        eval gpg "${options}" --yes --encrypt "${kak_buffile}"
```
lenormf: Same as above, regarding error checking.

rc/extra/gpg.kak
```diff
@@ -0,0 +1,57 @@
...
+}
+
+define-command -hidden gpg-cleanup  %{
+    %sh{
+        if [ -f "${kak_buffile}" ] && [ -n "${kak_opt_gpg_rmcmd}" ]; then
+            # Using shred(1) is naturally very slow for large files because it has
+            # to overwrite the whole temp file with zeros. Improve kakoune's
+            # responsiveness upon closing large files by forking off the
+            # deletion process.
+            (eval "${kak_opt_gpg_rmcmd} ${kak_buffile}") </dev/null >/dev/null 2>&1 &
```
lenormf: Braces `{`/`}` could be used here instead of parenthesis to prevent a subprocess to be spawn.

rc/extra/gpg.kak
nochiel: However, I can't seem to get messages to display in the mode line. :disappointed:

rc/extra/gpg.kak
nochiel: When I use braces it doesn't parse in Kakoune. :disappointed:

rc/extra/gpg.kak
```diff
@@ -0,0 +1,65 @@
...
+                echo -markup {Error} Failed to encrypt: ${kak_buffile}.gpg
+            "
+        fi
+    }
+
+}
+
+define-command -hidden gpg-cleanup  %{ evaluate-commands %sh{
+        if [ -f "${kak_buffile}" ] && [ -n "${kak_opt_gpg_rmcmd}" ]; then
+            eval "${kak_opt_gpg_rmcmd} ${kak_buffile}" </dev/null >/dev/null 2>&1 &
```
lenormf: You'll need quotes around the `kak_buffile` expansion, and also `--` after the expansion for the rm command to avoid parameter injection.

bitonic:
@nochiel @mawww is this good to merge? it'd be very useful.

nochiel:
I'm not currently able to maintain this PR. Closing for now.

PR #1890, opened 2018-02-27
Add -lock switch to enter-user-mode command

Delapouite:
Hello

This PR complements the previous one on user-modes.

It adds a handy `-lock` switch to the `enter-user-mode` command.

The behavior is the same than `v` vs `V` for the builtin *view* commands. It's useful for plugins that need to repeat the same command several times quickly in a row.
Example in *kakoune-gdb*: https://github.com/occivink/kakoune-gdb/blob/master/gdb-helper.kak
where you can do several `sssss` for multiple *step in*.

The code is almost a copy paste of the previous one with the additions

at the beginning:
```
 if (key == Key::Escape)
    return;
```
and at the end:

```
if (lock)
    enter_user_mode(context, std::move(mode_name), mode, true);
```


Delapouite:
The falling travis build is unrelated to this PR. It's because of : https://github.com/mawww/kakoune/commit/bebc81ebe1972eb33b89917bd520af2055d1fb6d (which has been fixed since)

PR #1898, opened 2018-03-03
Support extended function keys F13-F63

fsub:
Use case: mapping Shift-F11 (alias F21 on rxvt) to gdb-finish. Shift-F11 is the standard short cut for 'step out' on Visual Studio and Qt Creator.


Screwtapello:
Although ncurses traditionally gives you `f1` through `f63` keys, most keyboards have 12 function keys, and [ncurses convention](http://invisible-island.net/ncurses/ncurses.faq.html#modified_keys) is to map the rest to modified versions of the standard 12:

> F13-F24 are shifted F1-F12
> F25-F36 are control F1-F12 
> F37-F48 are shift+control F1-F12

Comparing `infocmp xterm-256color` with xterm's [ctlseqs.txt](http://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h2-PC-Style-Function-Keys), I deduce:

> F49-F60 are alt F1-F12
> F61-F63 are shift+alt F1-F3

So rather than mapping ncurses function keys directly to `<f1>` through `<f63>` symbols, perhaps we could follow the pattern of the rest of Kakoune's key-names and use `<f1>` through `<f12>`, `<F1>` through `<F12>`, `<c-f1>` through `<c-f12>`, `<c-F1>` through `<c-F12>`, `<a-f1>` through `<a-f12>`, etc?

fsub:
@Screwtapello I don't think that these relations are universally correct. Your first source clearly points out that Shift-F11 corresponds to F23 for `scoansi` but F21 for `rxvt`. AFAICS, all we have is [terminfo](http://pubs.opengroup.org/onlinepubs/7908799/xcurses/terminfo.html). That gives us `kf0`, ..., `kf63`.

Screwtapello:
The convention I described is, I think, pretty common across terminal emulators designed for keyboards with 12 function keys (since Thomas Dickey maintains the entire terminfo database and tries to keep such things consistent). As you can see from the table on the page I linked, rxvt is designed for keyboards with 10 or 20 function keys, and so it's never going to fit neatly into any pattern. In fact, the Debian's `rxvt-unicode-256color`  terminfo record only includes `kf1` through `kf20`, so no shift-F11 mapping will ever work there, regardless of what key-naming Kakoune uses.

I think it will be a usability issue for Kakoune if mapping Ctrl-C is `<c-c>` but mapping Ctrl-F3 is "run `cat` in a terminal and press the key you want to bind, then run `infocmp -1 $TERM` and grep for the characters your bound key produces, except you have to change `^[` to `\\E`, and then look for the `kfXX` code that maps to those characters, then take the `k` off the beginning and use `<fXX>` as the key name".

Also, @mawww tries to avoid tying Kakoune's language and configuration to closely to the details of ncurses and terminals; for example, Kakoune tries to use #rrggbb colours everywhere even though ncurses and terminals don't properly support them yet. For pretty much any other UI, supporting modified function keys as `<c-f3>` is going to be easier than terminfo's 64-function-key scheme.

fsub:
@Screwtapello I agree that it would be nice to support modified function keys like `<c-f3>`. I fail to see how this interferes with supporting `<f21>`. On platforms that do not support these extended function keys they, obviously, do not work.
Tying Shift-F11 to `<F23>`, on the other hand, is not correct for all platforms. I don't think that kakoune should make such assumptions. ([libtermkey](http://www.leonerd.org.uk/code/libtermkey/) seems to solve these problems but it certainly is not as ubiquitous as ncurses.)

That said, I made the change since I need support for Shift-F11. Once it's available, I'll happily switch to a better solution.

mawww:
Yeah, I am on the fence about this one, more than 12 function keys seems like legacy support to me, which I'd like to avoid, but at the same time it does solve a problem that we have no other solution for

fsub:
@mawww I just checked the situation in neovim: On my machine, mapping `<S-F11>` does not work. Provided that correct terminfo is used, mapping `<F21>` works like Shift-F11.
It seems as if accepting fancy keys is much easier when all keys are viewed as logical rather than physical entities. Does it really matter that a typical keyboard now has 12 function keys? But I'm not advertising. I agree with @Screwtapello's usability concerns. Fortunately, the decision is yours.

Screwtapello:
@fsub What terminal are you using? In gnome-terminal and (regular) Vim, mapping `<S-F11>` works and `<F21>` does not; nor does using the custom termcap code `<t_FB>`.

Actually, that's a thought: Kakoune could follow Vim's lead and have both "portable" key names (used directly on other UIs and via the semi-conventional mapping on ncurses UI), and ncurses-specific names (in some kind of namespace like `nc_`.

fsub:
Should this be closed or kept open?

mawww:
This can be closed, I want to pursue an alternate way to fix that:

* Merge the shift modifier support
* Add a ncurses specific option (in ui_option) to set the first shifted F key. (default either to kf13 of kf21), ncurses ui will the be responsible to rewrite f13 into shift-f1.

So, we keep our support for F1-F12, and add shifted support though a ncurses specific hack.

PR #1903, opened 2018-03-04
[WIP] Fix racer completion on Mac

marcesquerra:
This is a continuation to #1885 

The solution I implemented there doesn't work on Mac, and this intends to fix it.


mawww:
Thanks

PR #1911, opened 2018-03-07
src: Allow the `buffer` command to jump to line/column

lenormf:
Hi,

When I have several buffers opened in different directories and want to jump to a known location in one of them, using `edit` is too tedious if the buffer is far away from the CWD.

HTH.


danr:
This can also be achieved by using `select` after `edit`.

lenormf:
I can also use `<n>g` but that's still one more command I don't want to use if possible.

Delapouite:
On a side note, there's no easy way to goto a specific column.

src/commands.cc
```diff
@@ -514,11 +514,14 @@ const CommandDesc write_all_quit_cmd = {
     }
 };
 
+ParameterDesc buffer_params{
+    {}, ParameterDesc::Flags::None, 1, 3
```
mawww: This is not reused, it can just be written in line in the CommandDesc

src/commands.cc
```diff
@@ -529,6 +532,18 @@ const CommandDesc buffer_cmd = {
         {
             context.push_jump();
             context.change_buffer(buffer);
+
+            const size_t param_count = parser.positional_count();
+            if (param_count > 1 and not parser[1].empty())
+            {
+                int line = std::max(0, str_to_int(parser[1]) - 1);
+                int column = param_count > 2 and not parser[2].empty() ?
```
mawww: const LineCount/ByteCount instead of int would be nice here

src/commands.cc
```diff
@@ -529,6 +532,18 @@ const CommandDesc buffer_cmd = {
         {
             context.push_jump();
             context.change_buffer(buffer);
+
+            const size_t param_count = parser.positional_count();
+            if (param_count > 1 and not parser[1].empty())
```
mawww: I am not sure we need to test for empty, str_to_int will return 0 in those cases. (Maybe it should fail with "not a number", in any case we dont gain much by testing here).

lenormf:
I think all the comments have been addressed.

lenormf:
Bump!

Delapouite:
Should we also document these new optional params in https://github.com/mawww/kakoune/blob/master/doc/pages/commands.asciidoc ?

eraserhd:
I'm not clear on what the use case is.  I assume you want to use the `:buffer` command to auto-complete the buffer name at the prompt (which is why this helps when it is far away from the CWD), but then you want to type in the line and column manually?

mawww:
Yeah, it is still unclear to me what the use case for this is.

lenormf:
When I `grep` for something in another shell and get a line number, using `:buffer` and its fuzzy matching completion engine is more convenient than using `:edit` (which returns all files on disk as candidates).

eraserhd:
I think you are saying `:buffer`'s autocomplete is better.  Have you looked at `:find` in Maxime's kakrc?

mawww:
Thinking of it, for consistency with edit it could make sense to support this, but at the same time it means we lose the selections that were in whatever window we end-up recycling.

Issue #909, opened 2016-11-03
[Questions] Terminal and bindings limitations


occivink:
While trying to remap some keys I experienced some limitations as to what can and cannot be bound.
I don't know much about the inner workings of terminal emulation, but some cursory experimentation led me to the following:
* `alt` modifier bindings are pretty much okay with any key
* `shift` doesn't really exist as a modifier only the alternate key representation is sent, so something like `<s-space>` or `<s-ret>` is not possible
* `ctrl` modifier bindings are only okay with `a-z`

So now the questions:
Do these limitations come from the terminal emulation and cannot be overcome?
If so, do they extend to non-terminal UIs?

mawww:
Hello,

We dont currently treat `shift` as a modifier, because doing that means we would have to handle equivalence of `A` and `<s-a>` or `<` and `<s-,>` in Kakoune code, meaning we would either have to assume a qwerty keyboard, or have a way to know the actual layout of the user keyboard to map correctly `<s-whatever>`.

Regarding `ctrl`, we are hitting limitations of the terminal, Kakoune itself can support any keys with the `ctrl` modifier, but terminals do not, we receive the exact same thing for `<c-m>` or `<ret>`, regardless of the actual key that was pressed, same thing for `<c-i>` and `<tab>`. For that, non terminal UIs will have no problem to send these keystrokes to Kakoune, and Kakoune should handle them just fine.

Cheers,

Maxime.

Issue #147, opened 2014-05-23
Add shift-arrow commands to extend the selections


alexherbo2:
[ncurses.cc](https://github.com/mawww/kakoune/blob/master/src/ncurses.cc)

``` cpp
else switch (c)
{
    case KEY_SLEFT:  return Key::ShiftLeft;
    case KEY_SRIGHT: return Key::ShiftRight;
...
```

[keys.cc](https://github.com/mawww/kakoune/blob/master/src/keys.cc)

``` cpp
static const KeyAndName keynamemap[] = {
    { "s-up",    Key::ShiftUp    },
    { "s-down",  Key::ShiftDown  },
    { "s-left",  Key::ShiftLeft  },
    { "s-right", Key::ShiftRight },
...
```

[normal.cc](https://github.com/mawww/kakoune/blob/master/src/normal.cc)

``` cpp
KeyMap keymap =
{
    { Key::ShiftLeft, move<CharCount, Backward, SelectMode::Extend> },
    { Key::ShiftDown, move<LineCount, Forward,  SelectMode::Extend> },
    { Key::ShiftUp,   move<LineCount, Backward, SelectMode::Extend> },
    { Key::ShiftDown, move<CharCount, Forward,  SelectMode::Extend> },
...
```

lenormf:
Modifiers on arrow keys are not supported by `ncurses`, we can close this issue for the time being.

PR #1912, opened 2018-03-07
Support more keys with shift modifiers.

Screwtapello:
This includes all the special shifted keys I could generate with `gnome-terminal` and a standard PC keyboard; ncurses supports other keys like "shift-cancel" and "shift-message" that I have no idea how to trigger, so I didn't bother supporting them.

Since the only shift-modifiers previously supported by Kakoune were single-letter or `<tab>` → `<backtab>`, I wasn't sure how to name shifted keys. The difference between `<right>` and `<Right>` looked too subtle, so I went with `<RIGHT>` instead.

I also took the liberty of mapping the shifted arrows keys to match shifted HJKL, and shifted Home and End to match `<a-H>` and `<a-L>` since the unshifted ones matched `<a-h>` and `<a-l>`. I didn't map `<DEL>` to anything in normal mode because `<del>` wasn't mapped, and `<pageup>` and `<pagedown>` weren't mapped to anything that looked like it could use a shift modifier, so I left them alone.


Delapouite:
related: https://github.com/mawww/kakoune/issues/147

mawww:
Hello,

I believe we would need a more general support for shifted keys, `<s-key>` seems like the best bet. That means we also need to implement canonicalization of keys (`<s-a>` and `<A>` are the same, we could either have `Key` able to compare equal for the two representation, or ensure keys are canonicalized under one representation thats used internally).

If you want to give that a go, you are welcome to, or we can just open an issue for that and I'll try to look at it soon.

Screwtapello:
@occivink on IRC pointed me to #909, which brought up concerns with the shift modifier and keyboard layouts: Should maps for `<s-2>` activate when `@` is received? Not on Japanese keyboards! Should maps for `<s-i>` activate when `I` is received? Not in Turkey! It's not generally possible to turn a stream of Unicode codepoints back into the raw keypresses that produced them.

However, I also agree that (for non-Unicode-codepoint keys) having different syntax for the shift modifier versus all the other modifiers is a bit weird. I can think of a couple of alternatives:

- Status quo: No shift modifiers for anything.
    - Least features
    - Least effort
    - Definitely consistent with existing practice
    - Works the same way for everybody
    - Hypothetical GUI-based clients that can map arbitrary shifted keys are unduly limited
- Standardize on "shift makes uppercase", so shifted `a` is `A`, shifted `up` is `UP`, maybe one day shifted `f1` is `F1` (but, see #1898).
    - More features than the status quo
    - Basically implemented in this PR
    - (somewhat) consistent with existing practice, but new users are likely to expect `<s-…>` syntax to match `<c-…>` and `<a-…>` (and Vim uses similar key-modifier syntax too)
    - Works the same way for everybody
    - Awkward for GUI clients to support, especially for characters that do not have upper and lower case variants
- Standardize on the `<s-…>` syntax for the shift modifier, but deliberately exclude Unicode codepoint keys. Trying to map something like `<s-2>` should provide a helpful error message along the lines of "Cannot use the shift modfier with printable keys, just bind the shifted version directly".
    - Same functionality as the previous option
    - More difficult than the first two options, but existing modifier code should give me something to copy
    - Mostly consistent with existing system, but at least the surprising corner-case gives a helpful error
    - Works the same way for everybody
    - GUI clients can easily extend this by just removing the error message
- Standardize on the `<s-…>` syntax, and always normalize according to the rules of US-ASCII and the standard US keyboard layout.
    - More featureful than the above options
    - More difficult than the above options
    - Consistent with the existing system, but some users would likely complain when mappings for `<s-ä>` are not triggered by `Ä`
    - Works the same way for everybody
    - GUI clients would probably find this awkward to implement
- Standardize on the `<s-…>` syntax, and normalize shift mappings using locale-sensitive functions like `toupper()`.
    - Most featureful option
    - Most difficult option
    - Consistent with the existing system, but still not perfect: for example, the upper-case version of the German character ß is either SS or ẞ, but `<s-ß>` on [a German keyboard](https://en.wikipedia.org/wiki/German_keyboard_layout#General_information) actually produces `?`
    - Kakoune scripts could work differently in different environments, making it difficult for people to share plugins
    -  GUI clients would likely find this awkward to implement directly, but it's probably close enough to the "right" behaviour that they'd just ignore it and do the right thing, especially since behaviour already differs from one Kakoune instance to the next

I'd recommend the third option ("reject shift-modifier on Unicode codepoint mappings"), myself. What do other people think?

jussiarpalahti:
I have recently started trying out Kakoune, so I don't have that much experience. However, as a fresh user I can vouch for caution in relying on modifiers. I use finnish/swedish keyboard layout and have sticky keys on by default on MacOS and iOS devices (on latter I use Kak on Linux through SSH). This has meant that all <alt-something> keys have to be mapped specifically to what Kak expects them to be instead of Apple's alternative chars. Meaning []{}, but also <a-h> and <a-l> and so on. Also, many keys such as / and ; require shift already.

So, I guess I would favor allowing shift as modifier since it might actually make it easier for me to get a working command palette in Kak. As things stand I already have to adapt my terminal for Kak.

Screwtapello:
@jussiarpalahti When you say "all keys have to be mapped specifically to what Kak expects them to be instead of Apple's alternative chars", do you mean things like [Option+J typing ∆](https://fsymbols.com/keyboard/mac/) instead of sending the regular `<a-j>` command? There should be a "Use option as meta" setting [somewhere in the Terminal preferences](http://osxdaily.com/2013/02/01/use-option-as-meta-key-in-mac-os-x-terminal/) which should fix all of those at once, instead of you having to configure them individually.

jussiarpalahti:
@Screwtapello Yes, I do have option key work as meta key by default. However, this means that whilst Kakoune keys now work with alt modifier, I lose access to alternative chars (such as [ ] and others that Github Markdown apparently mangled in my previous message...).

I now have a separate terminal profile in iTerm where I've made individual mappings for necessary option + key combinations and everything else works as meta + key. Kak wiki suggests setting left or right option key to meta, but this does not work with sticky keys accessibility setting.

Of course this situation is not specific to Kakoune. It just so happens that all other terminal commands and apps I use work with option key by default.

mawww:
Hello,

I think we could standardize on `<s-key>` and consider that for base ascii, `<s-x>` is `X`. For other characters, I am not familiar enough with all the various existing keyboards, but based on the french keyboard, accentuated characters tend not to have a capital form on the keyboard anyway.

Once we have the shift modifier, we can express various special keys, and let the users who prefer `<s-a>` to `A` use that if they want (I think internally using `A` might be the simplest, but I believe Kakoune should be tolerant for base ascii and accept the shifted representation as well).

Screwtapello:
Closing this PR to create a new one that follows @mawww's suggested approach.

Issue #1611, opened 2017-10-03
add a ctagscmd option


lisael:
Hi,

The ctags command is hard-coded in `ctags.kak`. I'd like to be able to exclude build directories from ctags-generate.

We should use the same patern as `lintcmd`, `grepcmd`,  `makecmd`...

It seems pretty easy, I can make the PR if this design is accepted.

mawww:
Yep, thats fine by me, I'll merge such a PR.

PR #1915, opened 2018-03-07
Add ctagscmd option

Delapouite:
Hi. To close: https://github.com/mawww/kakoune/issues/1611


mawww:
I think we should do the same as with grep/make: remove the ctagsopt option and let that be part of the ctagscmd option, the two are a bit redundant.

PR #1924, opened 2018-03-12
Add jumpToDefinition and showDoc support for rust (with racer)

marcesquerra:
A couple of notes:
- I've only tested this with linux
- There is a bit where the code has to replace the literal '\n' with a new line. And the only way I managed to make it work is by writing an actual new line in the bash code, which is quite ugly. Any suggestion on how to make it nicer would be really welcome. (Writing '\n' as the string to put as the replacement replaces all the '\n' with 'n', and Writing '\\n' as the string to put as the replacement replaces all the '\n' with '\n')

### DIFF
--- a/rc/extra/racer.kak
+++ b/rc/extra/racer.kak
@@ -93,3 +93,53 @@ define-command racer-disable-autocomplete -docstring "Disable racer completion" 
     remove-hooks window racer-autocomplete
     unalias window complete racer-complete
 }
+
+define-command racer-go-definition -docstring "Jump to where the rust identifier below the cursor is defined" %{
+    %sh{
+        dir=$(mktemp -d "${TMPDIR:-/tmp}"/kak-racer.XXXXXXXX)
+        printf %s\\n "set-option buffer racer_tmp_dir ${dir}"
+        printf %s\\n "evaluate-commands -no-hooks %{ write ${dir}/buf }"
+    }
+    %sh{
+        dir=${kak_opt_racer_tmp_dir}
+        (
+            cursor="${kak_cursor_line} $((${kak_cursor_column} - 1))"
+            racer_data=$(racer --interface tab-text  find-definition ${cursor} ${kak_buffile} ${dir}/buf | head -n 1)
+            racer_match=$(echo "$racer_data" | cut -f1 )
+            if [[ "$racer_match" == "MATCH" ]]; then
+              racer_line=$(echo "$racer_data" | cut -f3 )
+              racer_column=$(echo "$racer_data" | cut -f4 )
+              racer_file=$(echo "$racer_data" | cut -f5 )
+              printf %s\\n "edit -existing '$racer_file' $racer_line $racer_column"
+              printf %s\\n "set-option buffer readonly true"
+            else
+              printf %s\\n "echo -debug 'racer could not find a definition'"
+            fi
+        )
+    }
+}
+
+define-command racer-show-doc -docstring "Show the documentation about the rust identifier below the cursor" %{
+    %sh{
+        dir=$(mktemp -d "${TMPDIR:-/tmp}"/kak-racer.XXXXXXXX)
+        printf %s\\n "set-option buffer racer_tmp_dir ${dir}"
+        printf %s\\n "evaluate-commands -no-hooks %{ write ${dir}/buf }"
+    }
+    %sh{
+        dir=${kak_opt_racer_tmp_dir}
+        (
+            cursor="${kak_cursor_line} $((${kak_cursor_column} - 1))"
+            racer_data=$(racer --interface tab-text  complete-with-snippet  ${cursor} ${kak_buffile} ${dir}/buf | head -n 2 | tail -n 1)
+            racer_match=$(echo "$racer_data" | cut -f1 )
+            printf %s\\n "info '$racer_data'"
+            if [[ "$racer_match" == "MATCH" ]]; then
+              racer_doc=$(echo "$racer_data" | cut -f9 | xargs )
+              racer_doc=$(echo "${racer_doc//\\n/
+}") # This is a bit ugly and probably should be replaced with something better than a new line in the code itself, but I couldn't work how
+              printf %s\\n "info '$racer_doc'"
+            else
+              printf %s\\n "echo -debug 'racer could not find a definition'"
+            fi
+        )
+    }
+}
--- a/rc/extra/racer.kak
+++ b/rc/extra/racer.kak
@@ -105,11 +105,11 @@ define-command racer-go-definition -docstring "Jump to where the rust identifier
         (
             cursor="${kak_cursor_line} $((${kak_cursor_column} - 1))"
             racer_data=$(racer --interface tab-text  find-definition ${cursor} ${kak_buffile} ${dir}/buf | head -n 1)
-            racer_match=$(echo "$racer_data" | cut -f1 )
+            racer_match=$(printf %s\\n "$racer_data" | cut -f1 )
             if [[ "$racer_match" == "MATCH" ]]; then
-              racer_line=$(echo "$racer_data" | cut -f3 )
-              racer_column=$(echo "$racer_data" | cut -f4 )
-              racer_file=$(echo "$racer_data" | cut -f5 )
+              racer_line=$(printf %s\\n "$racer_data" | cut -f3 )
+              racer_column=$(printf %s\\n "$racer_data" | cut -f4 )
+              racer_file=$(printf %s\\n "$racer_data" | cut -f5 )
               printf %s\\n "edit -existing '$racer_file' $racer_line $racer_column"
               printf %s\\n "set-option buffer readonly true"
             else
@@ -128,15 +128,16 @@ define-command racer-show-doc -docstring "Show the documentation about the rust 
     %sh{
         dir=${kak_opt_racer_tmp_dir}
         (
-            cursor="${kak_cursor_line} $((${kak_cursor_column} - 1))"
+            printf -v NEWLINE '\n'
+            cursor="${kak_cursor_line} ${kak_cursor_column}"
             racer_data=$(racer --interface tab-text  complete-with-snippet  ${cursor} ${kak_buffile} ${dir}/buf | head -n 2 | tail -n 1)
-            racer_match=$(echo "$racer_data" | cut -f1 )
-            printf %s\\n "info '$racer_data'"
+            racer_match=$(printf %s\\n "$racer_data" | cut -f1)
+            printf %s\\n "info %@.$racer_match.@"
             if [[ "$racer_match" == "MATCH" ]]; then
-              racer_doc=$(echo "$racer_data" | cut -f9 | xargs )
-              racer_doc=$(echo "${racer_doc//\\n/
-}") # This is a bit ugly and probably should be replaced with something better than a new line in the code itself, but I couldn't work how
-              printf %s\\n "info '$racer_doc'"
+              racer_doc=$(printf %s\\n "$racer_data" | cut -f9 | xargs )
+              racer_doc=$(printf %s\\n "$racer_doc" | sed -e "s/\\\n/\\$NEWLINE/g")
+              racer_doc=$(printf %s\\n "$racer_doc" | sed -e "s/@/\\@/g")
+              printf %s\\n "info %@$racer_doc@"
             else
               printf %s\\n "echo -debug 'racer could not find a definition'"
             fi
--- a/rc/extra/racer.kak
+++ b/rc/extra/racer.kak
@@ -102,20 +102,18 @@ define-command racer-go-definition -docstring "Jump to where the rust identifier
     }
     %sh{
         dir=${kak_opt_racer_tmp_dir}
-        (
-            cursor="${kak_cursor_line} $((${kak_cursor_column} - 1))"
-            racer_data=$(racer --interface tab-text  find-definition ${cursor} ${kak_buffile} ${dir}/buf | head -n 1)
-            racer_match=$(printf %s\\n "$racer_data" | cut -f1 )
-            if [[ "$racer_match" == "MATCH" ]]; then
-              racer_line=$(printf %s\\n "$racer_data" | cut -f3 )
-              racer_column=$(printf %s\\n "$racer_data" | cut -f4 )
-              racer_file=$(printf %s\\n "$racer_data" | cut -f5 )
-              printf %s\\n "edit -existing '$racer_file' $racer_line $racer_column"
-              printf %s\\n "set-option buffer readonly true"
-            else
-              printf %s\\n "echo -debug 'racer could not find a definition'"
-            fi
-        )
+        cursor="${kak_cursor_line} $((${kak_cursor_column} - 1))"
+        racer_data=$(racer --interface tab-text  find-definition ${cursor} "${kak_buffile}" "${dir}/buf" | head -n 1)
+        racer_match=$(printf %s\\n "$racer_data" | cut -f1 )
+        if [ "$racer_match" = "MATCH" ]; then
+          racer_line=$(printf %s\\n "$racer_data" | cut -f3 )
+          racer_column=$(printf %s\\n "$racer_data" | cut -f4 )
+          racer_file=$(printf %s\\n "$racer_data" | cut -f5 )
+          printf %s\\n "edit -existing '$racer_file' $racer_line $racer_column"
+          #printf %s\\n "set-option buffer readonly true"
+        else
+          printf %s\\n "echo -debug 'racer could not find a definition'"
+        fi
     }
 }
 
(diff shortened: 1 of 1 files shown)


rc/extra/racer.kak
```diff
@@ -93,3 +93,53 @@ define-command racer-disable-autocomplete -docstring "Disable racer completion"
...
+
+define-command racer-go-definition -docstring "Jump to where the rust identifier below the cursor is defined" %{
+    %sh{
+        dir=$(mktemp -d "${TMPDIR:-/tmp}"/kak-racer.XXXXXXXX)
+        printf %s\\n "set-option buffer racer_tmp_dir ${dir}"
+        printf %s\\n "evaluate-commands -no-hooks %{ write ${dir}/buf }"
+    }
+    %sh{
+        dir=${kak_opt_racer_tmp_dir}
+        (
```
lenormf: Why create a subshell? If you really need it, consider using `{`/`}` in order not to spawn a new process.

rc/extra/racer.kak
```diff
@@ -93,3 +93,53 @@ define-command racer-disable-autocomplete -docstring "Disable racer completion"
...
+            cursor="${kak_cursor_line} $((${kak_cursor_column} - 1))"
+            racer_data=$(racer --interface tab-text  find-definition ${cursor} ${kak_buffile} ${dir}/buf | head -n 1)
```
lenormf: Consider using `eval` when injecting parameters into a command, to avoid whitespace issues.

rc/extra/racer.kak
```diff
@@ -93,3 +93,53 @@ define-command racer-disable-autocomplete -docstring "Disable racer completion"
...
+            racer_match=$(echo "$racer_data" | cut -f1 )
```
lenormf: Don't use `echo` to display the contents of a variable directly, it's not portable. Use `printf %s\\n` instead.

rc/extra/racer.kak
```diff
@@ -93,3 +93,53 @@ define-command racer-disable-autocomplete -docstring "Disable racer completion"
...
+            if [[ "$racer_match" == "MATCH" ]]; then
```
lenormf: Double brackets and the `==` operator are not portable, use `[`/`]` and `=` instead.

rc/extra/racer.kak
```diff
@@ -93,3 +93,53 @@ define-command racer-disable-autocomplete -docstring "Disable racer completion"
...
+              racer_line=$(echo "$racer_data" | cut -f3 )
```
lenormf: `printf` here.

rc/extra/racer.kak
```diff
@@ -93,3 +93,53 @@ define-command racer-disable-autocomplete -docstring "Disable racer completion"
...
+              racer_column=$(echo "$racer_data" | cut -f4 )
```
lenormf: `printf` here.

rc/extra/racer.kak
```diff
@@ -93,3 +93,53 @@ define-command racer-disable-autocomplete -docstring "Disable racer completion"
...
+              racer_file=$(echo "$racer_data" | cut -f5 )
```
lenormf: `printf` here.

rc/extra/racer.kak
```diff
@@ -93,3 +93,53 @@ define-command racer-disable-autocomplete -docstring "Disable racer completion"
...
+    }
+}
+
+define-command racer-show-doc -docstring "Show the documentation about the rust identifier below the cursor" %{
+    %sh{
+        dir=$(mktemp -d "${TMPDIR:-/tmp}"/kak-racer.XXXXXXXX)
+        printf %s\\n "set-option buffer racer_tmp_dir ${dir}"
+        printf %s\\n "evaluate-commands -no-hooks %{ write ${dir}/buf }"
+    }
+    %sh{
```
lenormf: The same remarks above apply in this scope as well.

rc/extra/racer.kak
```diff
@@ -93,3 +93,53 @@ define-command racer-disable-autocomplete -docstring "Disable racer completion"
...
+        dir=${kak_opt_racer_tmp_dir}
+        (
+            cursor="${kak_cursor_line} $((${kak_cursor_column} - 1))"
+            racer_data=$(racer --interface tab-text  complete-with-snippet  ${cursor} ${kak_buffile} ${dir}/buf | head -n 2 | tail -n 1)
+            racer_match=$(echo "$racer_data" | cut -f1 )
+            printf %s\\n "info '$racer_data'"
+            if [[ "$racer_match" == "MATCH" ]]; then
+              racer_doc=$(echo "$racer_data" | cut -f9 | xargs )
+              racer_doc=$(echo "${racer_doc//\\n/
```
lenormf: This is not portable, use `tr`.

Screwtapello: I don't think `tr` can replace strings, only individual characters, so this would need to be `sed` instead.

That also solves the other problem, since `sed` *does* interpret backslash escapes.

danr: There was a discussion on IRC some weeks ago that  `sed` only replaces newlines if you give it a literal newline character (on macintosh). Otoh, `tr` can delete newlines, there are recent commits changing sed to tr because of this.

lenormf: I had originally overlooked that the expansion replaces a literal `\n`, not the newline character, so `sed` needs to be used here.

rc/extra/racer.kak
marcesquerra: Can you expand on this? I'm relatively new to kakoune and I'm not entirely sure I follow your comment.

If you are talking about the two "%sh{" blocks, when I try to put the bit that creates the temporal file in the same block that runs racer, for some reason racer complains the file doesn't exist.

marcesquerra:
I've tweaked most (I think) of the concerns raised, if you want to look again.

rc/extra/racer.kak
lenormf: The parenthesis line 105 creates a subshell, it's a feature of the shell scripting language. If you don't need the code enclosed by these parenthesis (the closing one is a few lines below) to be run in a separate shell, remove them. Otherwise, consider using curly braces `{`/`}` if possible, because they don't spawn a new process.

rc/extra/racer.kak
```diff
@@ -93,3 +93,54 @@ define-command racer-disable-autocomplete -docstring "Disable racer completion"
...
+        printf %s\\n "set-option buffer racer_tmp_dir ${dir}"
+        printf %s\\n "evaluate-commands -no-hooks %{ write ${dir}/buf }"
+    }
+    %sh{
+        dir=${kak_opt_racer_tmp_dir}
+        (
+            cursor="${kak_cursor_line} $((${kak_cursor_column} - 1))"
+            racer_data=$(racer --interface tab-text  find-definition ${cursor} ${kak_buffile} ${dir}/buf | head -n 1)
+            racer_match=$(printf %s\\n "$racer_data" | cut -f1 )
+            if [[ "$racer_match" == "MATCH" ]]; then
```
lenormf: `if [ "$racer_match" = "MATCH" ]; then`

rc/extra/racer.kak
```diff
@@ -93,3 +93,54 @@ define-command racer-disable-autocomplete -docstring "Disable racer completion"
...
+    }
+    %sh{
+        dir=${kak_opt_racer_tmp_dir}
+        (
+            printf -v NEWLINE '\n'
+            cursor="${kak_cursor_line} ${kak_cursor_column}"
+            racer_data=$(racer --interface tab-text  complete-with-snippet  ${cursor} ${kak_buffile} ${dir}/buf | head -n 2 | tail -n 1)
+            racer_match=$(printf %s\\n "$racer_data" | cut -f1)
+            printf %s\\n "info %@.$racer_match.@"
+            if [[ "$racer_match" == "MATCH" ]]; then
```
lenormf: `if [ "$racer_match" = "MATCH" ]; then`

rc/extra/racer.kak
```diff
@@ -93,3 +93,54 @@ define-command racer-disable-autocomplete -docstring "Disable racer completion"
...
+define-command racer-show-doc -docstring "Show the documentation about the rust identifier below the cursor" %{
+    %sh{
+        dir=$(mktemp -d "${TMPDIR:-/tmp}"/kak-racer.XXXXXXXX)
+        printf %s\\n "set-option buffer racer_tmp_dir ${dir}"
+        printf %s\\n "evaluate-commands -no-hooks %{ write ${dir}/buf }"
+    }
+    %sh{
+        dir=${kak_opt_racer_tmp_dir}
+        (
+            printf -v NEWLINE '\n'
```
lenormf: The `-v` flag is not [POSIX](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/printf.html), you will have to remove it.

rc/extra/racer.kak
```diff
@@ -93,3 +93,54 @@ define-command racer-disable-autocomplete -docstring "Disable racer completion"
...
+            cursor="${kak_cursor_line} ${kak_cursor_column}"
+            racer_data=$(racer --interface tab-text  complete-with-snippet  ${cursor} ${kak_buffile} ${dir}/buf | head -n 2 | tail -n 1)
+            racer_match=$(printf %s\\n "$racer_data" | cut -f1)
+            printf %s\\n "info %@.$racer_match.@"
+            if [[ "$racer_match" == "MATCH" ]]; then
+              racer_doc=$(printf %s\\n "$racer_data" | cut -f9 | xargs )
```
lenormf: `xargs` isn't guaranteed to be installed on all systems, you'll have to remove it. I think this subshell can be replaced with an `awk` one-liner, but I'm not sure the behavior of `xargs` without any command is the same everywhere, so I don't know what the expected result is here.

rc/extra/racer.kak
lenormf: Or maybe simply escaping the arguments will be enough, if you don't inject flags.

rc/extra/racer.kak
Screwtapello: I'm guessing this whole `racer_doc` manipulation is supposed to do something like "remove all real newlines, unescape escaped newlines", yes? Then something like this should work:

    racer_doc=$(printf '%s\n' "$racer_data" | sed -E -e '
        # Grab the 9th tab-delimited field of every line
        s/([^\t]*\t){8}([^\t]*)/\2/
        # Copy the first line to the hold buffer
        1 h
        # Append the second and subsequent lines to the hold buffer
        2,$ H
        # When we get to the last line...
        $ {
            # Move the whole blob back to the pattern buffer
            x
            # Delete all the real newlines in the pattern buffer
            s/\n/ /g
            # Unescape all the escaped newlines
            s/\\n/\n/g
            # Escape any @ signs so we can use @ as a delimiter for Kakoune
            s/@/\\@/g
        }
    ')

lenormf: Unfortunately, `-E` isn't defined by POSIX.

rc/extra/racer.kak
marcesquerra: Thanks, I got it now. Fixed (it works), going to be in on the next commit

marcesquerra:
I think that now looks a lot better.

I've found that I was swimming against the current, trying to remove all the escapes that racer was putting in place in the returned data, when this escapes, in truth, were making my life easier allowing me to add the value directly where it needed to be used.

rc/extra/racer.kak
Screwtapello: Dammit! I checked `sed(1)`, which says "the -E option has been supported for years by GNU sed, and is now included in POSIX.", but I failed to check `sed(1posix)` or [the online version](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/sed.html).

StackOverflow [points](https://unix.stackexchange.com/questions/310446/are-there-standard-versions-of-sed-on-which-e-is-unsupported) to [an issue in the POSIX issue-tracker](http://austingroupbugs.net/view.php?id=528) that adds `-E`, but although the issue was resolved at the end of 2012, apparently the change didn't make it into the 2016 revision.

rc/extra/racer.kak
```diff
@@ -93,3 +93,48 @@ define-command racer-disable-autocomplete -docstring "Disable racer completion"
...
+    }
+    %sh{
+        dir=${kak_opt_racer_tmp_dir}
+        cursor="${kak_cursor_line} ${kak_cursor_column}"
+        racer_data=$(racer --interface tab-text  complete-with-snippet  ${cursor} "${kak_buffile}" "${dir}/buf" | head -n 2 | tail -n 1)
+        racer_match=$(printf %s\\n "$racer_data" | cut -f1)
+        if [ "$racer_match" = "MATCH" ]; then
+          racer_doc=$(printf %s\\n "$racer_data" | cut -f9 )
+          racer_doc=$(printf %s\\n "$racer_doc" | sed -e 's/^"\(.*\)"$/\1/g')
+          racer_doc=$(printf %s\\n "$racer_doc" | sed -e "s/@/\\\\@/g")
```
Screwtapello: You might as well put both sed replacement commands in the one invocation.

rc/extra/racer.kak
```diff
@@ -127,20 +125,16 @@ define-command racer-show-doc -docstring "Show the documentation about the rust
...
-              racer_doc=$(printf %s\\n "$racer_data" | cut -f9 | xargs )
-              racer_doc=$(printf %s\\n "$racer_doc" | sed -e "s/\\\n/\\$NEWLINE/g")
-              racer_doc=$(printf %s\\n "$racer_doc" | sed -e "s/@/\\@/g")
-              printf %s\\n "info %@$racer_doc@"
-            else
-              printf %s\\n "echo -debug 'racer could not find a definition'"
-            fi
-        )
+        cursor="${kak_cursor_line} ${kak_cursor_column}"
+        racer_data=$(racer --interface tab-text  complete-with-snippet  ${cursor} "${kak_buffile}" "${dir}/buf" | head -n 2 | tail -n 1)
```
lenormf: It's not critical, but the `head -n 2 | tail -n 1` can be replaced with `sed -n 2p`, assuming the goal is to print the second line.

marcesquerra: Changed. (Yes, only the second line is need; the first line has headers and the rest are not the proper matches)

rc/extra/racer.kak
marcesquerra: Fixed, thanks. It's cleaner now.

marcesquerra:
I've put back the bit that make readonly the buffers opened by "go to definition", but this time only for files from the external crates or the standard library (this way avoiding setting your own files as readonly when jumping around)

rc/extra/racer.kak
```diff
@@ -104,13 +104,20 @@ define-command racer-go-definition -docstring "Jump to where the rust identifier
...
         racer_data=$(racer --interface tab-text  find-definition ${cursor} "${kak_buffile}" "${dir}/buf" | head -n 1)
+
         racer_match=$(printf %s\\n "$racer_data" | cut -f1 )
         if [ "$racer_match" = "MATCH" ]; then
           racer_line=$(printf %s\\n "$racer_data" | cut -f3 )
           racer_column=$(printf %s\\n "$racer_data" | cut -f4 )
           racer_file=$(printf %s\\n "$racer_data" | cut -f5 )
           printf %s\\n "edit -existing '$racer_file' $racer_line $racer_column"
-          #printf %s\\n "set-option buffer readonly true"
+          if [[ "$racer_file" == ${RUST_SRC_PATH}* ]]; then
```
lenormf: `[[` and `==` are bashisms, they're not portable that will work on arbitrary POSIX-compliant shells, you need to use:

`if [ "$foo" = "abc" ]; then`

rc/extra/racer.kak
```diff
@@ -104,13 +104,20 @@ define-command racer-go-definition -docstring "Jump to where the rust identifier
...
+            printf %s\\n "set-option buffer readonly true" # The definition resides on the standard library, and the new buffer should be readonly
+          else
+            if [[ "$racer_file" == ${HOME}/.cargo/registry/src* ]]; then
```
lenormf: Same remark.

rc/extra/racer.kak
```diff
@@ -126,12 +133,13 @@ define-command racer-show-doc -docstring "Show the documentation about the rust
...
-        racer_data=$(racer --interface tab-text  complete-with-snippet  ${cursor} "${kak_buffile}" "${dir}/buf" | head -n 2 | tail -n 1)
+        racer_data=$(racer --interface tab-text  complete-with-snippet  ${cursor} "${kak_buffile}" "${dir}/buf" | sed -n 2p )
         racer_match=$(printf %s\\n "$racer_data" | cut -f1)
         if [ "$racer_match" = "MATCH" ]; then
           racer_doc=$(printf %s\\n "$racer_data" | cut -f9 )
-          racer_doc=$(printf %s\\n "$racer_doc" | sed -e 's/^"\(.*\)"$/\1/g')
-          racer_doc=$(printf %s\\n "$racer_doc" | sed -e "s/@/\\\\@/g")
+          racer_doc=$(printf %s\\n "$racer_doc" |
+            sed -e 's/^"\(.*\)"$/\1/g' | # Remove leading and trailing quotes
+            sed -e "s/@/\\\\@/g")        # Escape all @ so that it can be properly used in the string expansion
```
lenormf: I think the idea was to have something like the following, to avoid spawning an extra process:

```
racer_doc=$(printf %s\\n "$racer_doc" |
       sed -e 's/^"\(.*\)"$/\1/g' \
           -e "s/@/\\\\@/g")
```

marcesquerra:
Since this PR has already been merged, I'm trying to address the last two comments from @lenormf at #1933

lenormf:
Thanks!

PR #1931, opened 2018-03-15
WIP: Add support for the shift modifier.

Screwtapello:
Currently, special keys just use the shift modifier directly, uppercase-ASCII chars are normalised to use the shift modifier and lowercase chars (for forwards compatibility with front-ends that can actually read raw keyboard events), and other chars are not allowed a shift modifier.

I haven't updated the ncurses UI to construct the new `Key` structs, nor updated `normal.cc` to expect them, so Kakoune should work the same as usual, with the exception of custom mappings for uppercase characters.

Unfortunately, a huge number of integration tests fail, even ones that don't use any uppercase keys, let alone custom mappings. For example, `compose/backward-search`.

### DIFF
--- a/src/keys.cc
+++ b/src/keys.cc
@@ -67,7 +67,16 @@ KeyList parse_keys(StringView str)
     {
         if (*it != '<')
         {
-            result.emplace_back(Key::Modifiers::None, *it);
+            Key::Modifiers modifier = Key::Modifiers::None;
+            Codepoint key = *it;
+
+            if (is_upper(key))
+            {
+                modifier |= Key::Modifiers::Shift;
+                key = to_lower(key);
+            }
+
+            result.emplace_back(modifier, key);
             continue;
         }
 
@@ -92,6 +101,7 @@ KeyList parse_keys(StringView str)
             {
                 case 'c': modifier |= Key::Modifiers::Control; break;
                 case 'a': modifier |= Key::Modifiers::Alt; break;
+                case 's': modifier |= Key::Modifiers::Shift; break;
                 default:
                     throw runtime_error(format("unable to parse modifier in '{}'",
                                                full_desc));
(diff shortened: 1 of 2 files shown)


Delapouite:
I haven't verified yet, but the issue with failing tests coud be related to the init phase:

```
exec -save-regs / %{%s%\(\K[^)]+\)<ret>a<backspace><esc>i<backspace><backspace><c-u><esc><a-;>}
```

https://github.com/mawww/kakoune/blob/master/test/run#L14

Delapouite:
The tests are indeed failing because of the init command.

To see what's going, on I've added this the `write_to_debug_buffer` line in the for loop of the `exec` command in `commands.cc`:

```
const CommandDesc exec_string_cmd = {
    "execute-keys",
    "exec",
    "execute-keys <switches> <keys>: execute given keys as if entered by user",
    context_wrap_params,
    CommandFlags::None,
    CommandHelper{},
    CommandCompleter{},
    [](const ParametersParser& parser, Context& context, const ShellContext&)
    {
        context_wrap(parser, context, [](const ParametersParser& parser, Context& context) {
            KeyList keys;
            for (auto& param : parser)
            {
                KeyList param_keys = parse_keys(param);
                keys.insert(keys.end(), param_keys.begin(), param_keys.end());
            }

            ScopedEdition edition(context);
            for (auto& key : keys) {
                write_to_debug_buffer(key_to_str(key));
                context.input_handler().handle_key(key);
            }
        });
    }
};
```

Then, when `exec -save-regs / %{%s%\(\K[^)]+\)<ret>a<backspace><esc>i<backspace><backspace><c-u><esc><a-;>}` is executed:

- **regular kakoune**
```
%
s
%
\
(
\
K
[
^
)
]
<plus>
\
)
<ret>
a
<backspace>
<esc>
i
<backspace>
<backspace>
<c-u>
<esc>
<a-;>
```

- **this PR**
```
%
s
%
\
(
\
<s-k>
[
^
)
]
<plus>
\
)
<ret>
Error: 1:1: 'exec' nothing selected

```

Screwtapello:
The thing that puzzles me about that explanation is that the `exec` is executed for every test, but only some of them fail.

Perhaps the tests that pass are ones that don't have/care about an initial selection?

Screwtapello:
However, I notice there's a call to `handle_key()` in that code, a function I wasn't familiar with. Turns out, it's built to ignore modifiers it doesn't recognise.

So I guess it boils down to this:

 1. normalize `<s-x>` to `X` with an en_US-and-ASCII-specific hack in `keys.cc`
 2. normalize `X` to `<s-x>` with en_US-and-ASCII-specific hacks in `input_handler.cc`
 3. don't support the shift-modifier or code-points at all

I'm leaning more heavily toward that third option, but I guess I'll have another bash at the first one.

Issue #1930, opened 2018-03-15
[jump list] save-current-selection doesn't restore jumps correctly when text is inserted after selection.


nochiel:
### Steps 
- `:execute-keys %{i1<a-;><c-s><space>2<a-;><c-s><space>3<esc><c-o><c-o>}`
### Outcome 
- cursor stays in place.

### Expected 
- cursor should jump to point after 2 then jump to point after 1.

lenormf:
Does the same error occur without using `<a-;>` ?

nochiel:
Yes.

nochiel:
I should add that the following also doesn't work:
`execute-keys %{i1<esc>Zi<space>2<esc>z}`

The cursor remains in place even though kak reports that selections have been restored.

The selection is being saved but it is not being restored.

mawww:
I believe Kakoune's behaviour is correct, you are saving a cursor which is *after* 1, then you insert *before* that cursor ` 2`, save it again (it is still on eol), and insert ` 3` *before* it.

The logical position you saved has always been the end of line (this is nothing eol specific, if you would be on any character it would be the same), and you have always been inserting *before* it, so I would not expect <c-o> to jump elsewhere.

mawww:
To be more precise, if you'd save the position with cursor onto `1`, `2` or `3` (so before the point of next insertion), things would work as you expect.

To understand better what is going on, just open another client and put his cursor onto the eol before executing the keys, you'll see that this other client selections are correctly updated and stay on eol.

PR #1932, opened 2018-03-16
Add tests for saving selections while inserting.

nochiel:
Failing tests for #1930.


mawww:
As said in #1930, I believe current behaviour is correct.

PR #1938, opened 2018-03-17
Docs: add missing [<switches>] in commands signatures

Delapouite:


Screwtapello:
What's the difference between `[<flags>]` and `[<switches>]`?

We might as well make the documentation match the terminology used in the code.

mawww:
Yeah, we should choose one and use only that. Also, various commands actually support flags/switches in many positions, not only after the command name itself. Thats generally the rule, and only certain commands will only allow switches at start.

lenormf:
I always thought flags were switches that didn't take any argument.

Delapouite:
All `<flags>` have now been renamed to `<switches>` since it was both used in infoboxes and in the code itself.

PR #1942, opened 2018-03-19
Add ncurses_layout option to customize the UI (buffer-list, grep summary…)

Delapouite:
Hi

## TL;DR

This PR removes `ncurses_status_on_top` option in favor of a more powerful and flexible `ncurses_layout`.

`ncurses_layout` is `main,status` by default. If you want the status on top behavior: `status,main`.

But it offers more potential. You can add extra `zones` (a better name can be discussed).
Let say you want to add a _buffer list_ on top of the screen.

1 - declare a regular string option, and name it like you desire, like `buffer_list`.
2 - add it to the `ncurses_layout` list. Example `buffer_list,main,status` to display the content of the `buffer_list` option in a new line on top of the screen.
3 - now it's up to a kak script to compute and update `buffer_list` when needed. It's value will be updated on screen. You can use faces.

That's it. No new commands involved, we just reuse and extend a feature pioneered by the `modelinefmt` option.

## UI paradigm

Kakoune already offers two ways to display extra bits of data on screen:
- `info`: only one at a time, no styles, get in the way if too big
- `menu`: only one at a time, interactive features

So having a system to just display `exta lines` of data like described in this PR provides an alternative persistent approach which is non intrusive and useful.

## Prior attempts

- https://github.com/mawww/kakoune/pull/1065:  this PR only focus on a buffer list. With `ncurses_layout` you can add as many _zones_ you need with any content you want.
- https://github.com/mawww/kakoune/wiki/Bar: this solution uses an external `bar` program to display things. It's hard to keep it displayed in the correct workspace, in the right position, always tied to the correct terminal window…
- [kakoune-electron](https://github.com/delapouite/kakoune-electron): using a complete foreign GUI system may have other advantages but it's a lot of work and rewriting for such a nice feature that can be implemented right in the current ncurses_ui and benefit every kakoune users out of the box.

## About the code

It's about 100ish new lines of code (including doc), with mainly plumbing stuffs for the different UIs (Remote, JSON…). Of course there's room to improve / fix things, it's mainly a proof of concept.

## Screenshot and example

image

```
declare-option str zone_buflist

define-command -hidden zone-buflist %{ %sh{
  list=''
  while read buf; do
    index=$(($index + 1))
    if [ "$buf" = "$kak_bufname" ]; then
      # markup specific to lemonbar
      list="$list{+r} $index $buf "
    else
      list="$list{Default} $index $buf "
    fi
  done <<< $(printf '%s\n' "$kak_buflist" | tr ':' '\n')
  echo "set-option global zone_buflist '$list'"
} }

hook global WinDisplay .* zone-buflist

set global ui_options ncurses_assistant=none:ncurses_layout=zone_buflist,main,status
```

### DIFF
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -1,7 +1,9 @@
 #include "ncurses_ui.hh"
 
+#include "buffer_utils.hh"
 #include "display_buffer.hh"
 #include "event_manager.hh"
+#include "face_registry.hh"
 #include "keys.hh"
 #include "ranges.hh"
 #include "string_utils.hh"
@@ -256,7 +258,8 @@ NCursesUI::NCursesUI()
       }},
       m_assistant(assistant_clippy),
       m_colors{default_colors},
-      m_cursor{CursorMode::Buffer, {}}
+      m_cursor{CursorMode::Buffer, {}},
+      m_layout{"main", "status"}
 {
     initscr();
     raw();
@@ -317,16 +320,17 @@ void NCursesUI::Window::refresh()
 void NCursesUI::redraw()
 {
     pnoutrefresh(m_window, 0, 0, 0, 0,
-                 (int)m_dimensions.line + 1, (int)m_dimensions.column);
+                 (int)m_dimensions.line + (int)m_layout.size() - 1, (int)m_dimensions.column);
 
     m_menu.refresh();
     m_info.refresh();
+    for (auto zone : m_zones)
+        draw_zone(zone.key, zone.value, get_face("StatusLine"));
 
     if (m_cursor.mode == CursorMode::Prompt)
-        wmove(newscr, m_status_on_top ? 0 : (int)m_dimensions.line,
-              (int)m_cursor.coord.column);
+        wmove(newscr, get_zone_line("status"), (int)m_cursor.coord.column);
     else
-        wmove(newscr, (int)m_cursor.coord.line + (m_status_on_top ? 1 : 0),
+        wmove(newscr, (int)m_cursor.coord.line + get_zone_line("main"),
               (int)m_cursor.coord.column);
 
     doupdate();
@@ -390,7 +394,7 @@ void NCursesUI::draw(const DisplayBuffer& display_buffer,
 
     check_resize();
 
-    LineCount line_index = m_status_on_top ? 1 : 0;
+    LineCount line_index = get_zone_line("main");
     for (const DisplayLine& line : display_buffer.lines())
     {
         wmove(m_window, (int)line_index, 0);
@@ -402,7 +406,7 @@ void NCursesUI::draw(const DisplayBuffer& display_buffer,
     wbkgdset(m_window, COLOR_PAIR(get_color_pair(padding_face)));
     set_face(m_window, padding_face, default_face);
 
-    while (line_index < m_dimensions.line + (m_status_on_top ? 1 : 0))
+    while (line_index < m_dimensions.line + get_zone_line("main"))
     {
         wmove(m_window, (int)line_index++, 0);
         wclrtoeol(m_window);
@@ -416,7 +420,7 @@ void NCursesUI::draw_status(const DisplayLine& status_line,
                             const DisplayLine& mode_line,
                             const Face& default_face)
 {
-    const int status_line_pos = m_status_on_top ? 0 : (int)m_dimensions.line;
+    const int status_line_pos = get_zone_line("status");
     wmove(m_window, status_line_pos, 0);
 
     wbkgdset(m_window, COLOR_PAIR(get_color_pair(default_face)));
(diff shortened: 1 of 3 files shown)


danr:
I like this, especially so since it generalises already existing functionality.

Delapouite:
mawww's feedback on IRC:

- it would need to be a more general thing (having Client interpret ui_options seems wrong, having it read a ncurses_* value even more)
- we could isolate that in ncurses ui by using 2 options: the zones one, and have zones reference another entry in the ui_options map. `ui_options 'ncurses_zones=main,status,blah:blah=the text to have in blah zone'`. Then using hooks to update blah when buffer are opened/closed (if blah was used to list buffers). I believe using set -add on a  hash map option, such as ui_option should permit to just change the value of one entry 'set -add ui_options blah=...' will set/reset blah entry to whatever is given

Delapouite:
The changes are now only limited to `ncurses_ui` and do not affect the client and other UIs anymore.

src/ncurses_ui.cc
```diff
@@ -1061,9 +1082,31 @@ void NCursesUI::set_ui_options(const Options& options)
...
+        if (it != options.end())
+        {
+            auto layout = it->value | split<String>(',') | gather<Vector<String>>();
+            auto main = find(layout, "main");
+            auto status = find(layout, "status");
+            if (main == m_layout.end() or status == m_layout.end())
+                write_to_debug_buffer("main and status zones are mandatory in ncurses_layout");
+
+            m_layout = layout;
+            m_status_on_top = status < main;
```
mawww: Do we still need m_status_on_top ? It looks like we should be able to get rid of it.

src/ncurses_ui.cc
```diff
@@ -1061,9 +1082,31 @@ void NCursesUI::set_ui_options(const Options& options)
...
+        auto it = options.find("ncurses_layout"_sv);
+        if (it != options.end())
+        {
+            auto layout = it->value | split<String>(',') | gather<Vector<String>>();
+            auto main = find(layout, "main");
+            auto status = find(layout, "status");
+            if (main == m_layout.end() or status == m_layout.end())
+                write_to_debug_buffer("main and status zones are mandatory in ncurses_layout");
+
+            m_layout = layout;
```
mawww: No need to copy the layout, use `m_layout = std::move(layout)` (and iterate on m_layout or move that line after the iteration).

src/ncurses_ui.cc
```diff
@@ -1061,9 +1082,31 @@ void NCursesUI::set_ui_options(const Options& options)
...
-        auto it = options.find("ncurses_status_on_top"_sv);
-        m_status_on_top = it != options.end() and
-            (it->value == "yes" or it->value == "true");
+        auto it = options.find("ncurses_layout"_sv);
+        if (it != options.end())
+        {
+            auto layout = it->value | split<String>(',') | gather<Vector<String>>();
+            auto main = find(layout, "main");
+            auto status = find(layout, "status");
+            if (main == m_layout.end() or status == m_layout.end())
```
mawww: if we get rid of the m_status_on_top assignment below, this can become `not contains(layout, "main") or not contains(layout, "status")`. Additionally, you are comparing to **m_**layout here, so that condition is going to be always false.

src/ncurses_ui.cc
```diff
@@ -1106,4 +1149,15 @@ void NCursesUI::set_ui_options(const Options& options)
     }
 }
 
+int NCursesUI::get_zone_line (StringView zone_name)
```
mawww: I dont like the use of `get_` here, should be `compute_` or just `zone_line`. Also, remove the space before `(`.

src/ncurses_ui.cc
```diff
@@ -933,12 +954,13 @@ void NCursesUI::info_show(StringView title, StringView content,
...
 
-    const Rect rect = {m_status_on_top ? 1_line : 0_line, m_dimensions};
+    auto main_line_pos = get_zone_line("main");
+    const Rect rect = {{main_line_pos}, m_dimensions};
     InfoBox info_box;
     if (style == InfoStyle::Prompt)
     {
         info_box = make_info_box(m_info.title, m_info.content, m_dimensions.column, m_assistant);
-        anchor = DisplayCoord{m_status_on_top ? 0 : m_dimensions.line, m_dimensions.column-1};
+        anchor = DisplayCoord{m_status_on_top ? main_line_pos-1 : main_line_pos + m_dimensions.line, m_dimensions.column-1};
```
mawww: I dont get why we use `m_status_on_top` here, the `get_zone_line` should give use the correct line offset.

Clyybber:
Why was this closed?

Delapouite:
Since faces were scoped, there is no easy way to rebase this PR.
Further efforts will continue on this branch which I intend to update soon: https://github.com/Delapouite/hakoune/tree/layout

Jackojc:
Any update on this?

sgraf812:
Well, I guess the answer is https://github.com/mawww/kakoune/wiki/Bar. But it would be neat to have all this logic wrapped in a plugin rather than having to copy and paste from a wiki page.

Jackojc:
Wish there was a way to do it without external support of X or Wayland.

PR #1947, opened 2018-03-20
Shift modifiers v3

Screwtapello:
This branch teaches Kakoune that Shift is a possible modifier, including the `s-` prefix in key names. `<s-x>` in normalized to `<X>` when `x` is a lowercase ASCII character, but other printable characters (including aliases for printable characters like `lt`, `gt`, `plus` and `minus`) are disallowed because the behaviour of the shift modifier depends on the user's keyboard layout, which Kakoune cannot determine from inside a terminal.

Shifted keys that ncurses recognises are mapped to properly-modified keys so they can be mapped by users, and added to the default normal mappings where appropriate.

This still has some potential problems, but I don't think addressing them is practical:

  - It's possible to create a mapping for a key like `<c-s-a>`, but it cannot be triggered because terminals ignore the shift modifier for control keys
  - You can create a mapping for `<s-f11>`, but it cannot be triggered because there's no universal standard for what key-code ncurses will use for modified function keys
  - It is not universally true that applying the shift modifier to a lower-case ASCII letter will produce the corresponding uppercase ASCII letter (see [Dotted and dotless I](https://en.wikipedia.org/wiki/Dotted_and_dotless_I)), but Kakoune's default keymap already assumes this is the case for mnemonic reasons (`i` is insert, `I` is insert at the beginning of the line) so that ship has already sailed

This branch also includes a commit that teaches Kakoune to normalise `<s-tab>` to `<backtab>` in the same way that it normalises `<s-x>` to `<X>`. Backtab is [theoretically distinct](https://deskthority.net/wiki/Tab_key) from the tab key, but I bet it will be `<s-tab>` on every computer Kakoune is ever run on, and somebody wanting to bind it will probably guess `<s-tab>` before they guess `<backtab>`. However, I can imagine reasons not to do this, or not to do it this way, so I left it as a separate commit that can be left out if desired.

### DIFF
--- a/doc/pages/mapping.asciidoc
+++ b/doc/pages/mapping.asciidoc
@@ -65,14 +65,19 @@ be used:
     Keys can also be wrapped in angle-brackets for consistency
     with the non-alphabetic keys below.
 
-*X*, *<X>*::
-    Holding down Shift while pressing the *x* key.
-
 *<c-x>*::
     Holding down Control while pressing the *x* key.
 
 *<a-x>*::
     Holding down Alt while pressing the *x* key.
+
+*<s-x>*, *X*, *<X>*, *<s-X>*::
+    Holding down Shift while pressing the *x* key.
+    *<s-x>*, *<s-X>* and *<X>* are treated as the same key. The *s-* modifier
+    only works with ASCII letters and cannot be used with other printable keys
+    (non-ASCII letters, digits, punctuation) because their shift behaviour
+    depends on your keyboard layout. The *s-* modifier _can_ be used with
+    special keys like *<up>* and *<tab>*.
 
 *<c-a-x>*::
     Holding down Control and Alt while pressing the *x* key.
@@ -92,9 +97,6 @@ be used:
 *<tab>*::
     The Tab key.
 
-*<backtab>*::
-    The reverse-tab key. This is Shift-Tab on most keyboards.
-
 *<backspace>*::
     The Backspace (delete to the left) key.
 
@@ -110,3 +112,8 @@ be used:
 
 *<f1>*, *<f2>*, ...*<f12>*::
     Function keys.
+
+NOTE: Although Kakoune allows many key combinations to be mapped, not every
+possible combination can be triggered. For example, due to limitations in
+the way terminals handle control characters, mappings like *<c-s-a>* are
+unlikely to work in Kakoune's terminal UI.
--- a/src/input_handler.cc
+++ b/src/input_handler.cc
@@ -461,15 +461,15 @@ public:
         }
         else if (key == ctrl('w'))
             to_next_word_begin<Word>(m_cursor_pos, m_line);
-        else if (key == ctrlalt('w'))
+        else if (key == ctrl(alt('w')))
             to_next_word_begin<WORD>(m_cursor_pos, m_line);
         else if (key == ctrl('b'))
             to_prev_word_begin<Word>(m_cursor_pos, m_line);
-        else if (key == ctrlalt('b'))
+        else if (key == ctrl(alt('b')))
             to_prev_word_begin<WORD>(m_cursor_pos, m_line);
         else if (key == ctrl('e'))
             to_next_word_end<Word>(m_cursor_pos, m_line);
-        else if (key == ctrlalt('e'))
+        else if (key == ctrl(alt('e')))
             to_next_word_end<WORD>(m_cursor_pos, m_line);
         else if (key == ctrl('k'))
             m_line = m_line.substr(0_char, m_cursor_pos).str();
@@ -623,7 +623,7 @@ public:
                 it = std::find_if(m_choices.begin(), m_selected, match_filter);
             select(it);
         }
-        else if (key == Key::Up or key == Key::BackTab or
+        else if (key == Key::Up or key == shift(Key::Tab) or
                  key == ctrl('p') or (not m_edit_filter and key == 'k'))
         {
             ChoiceList::const_reverse_iterator selected(m_selected+1);
@@ -837,9 +837,9 @@ public:
                 m_refresh_completion_pending = true;
             }
         }
-        else if (key == Key::Tab or key == Key::BackTab) // tab completion
+        else if (key == Key::Tab or key == shift(Key::Tab)) // tab completion
         {
-            const bool reverse = (key == Key::BackTab);
+            const bool reverse = (key == shift(Key::Tab));
             CandidateList& candidates = m_completions.candidates;
             // first try, we need to ask our completer for completions
             if (candidates.empty())
@@ -1568,8 +1568,10 @@ InputHandler::ScopedForceNormal::~ScopedForceNormal()
 
 static bool is_valid(Key key)
 {
+    constexpr Key::Modifiers valid_mods = (Key::Modifiers::Control | Key::Modifiers::Alt | Key::Modifiers::Shift);
+
     return key != Key::Invalid and
-        ((key.modifiers & ~Key::Modifiers::ControlAlt) or key.key <= 0x10FFFF);
+        ((key.modifiers & ~valid_mods) or key.key <= 0x10FFFF);
 }
 
 void InputHandler::handle_key(Key key)
--- a/src/keys.cc
+++ b/src/keys.cc
@@ -11,6 +11,11 @@
 namespace Kakoune
 {
 
+struct key_parse_error : runtime_error
+{
+    using runtime_error::runtime_error;
+};
+
 static Key canonicalize_ifn(Key key)
 {
     if (key.key > 0 and key.key < 27)
@@ -19,6 +24,22 @@ static Key canonicalize_ifn(Key key)
         key.modifiers = Key::Modifiers::Control;
         key.key = key.key - 1 + 'a';
     }
+
+    if (key.modifiers & Key::Modifiers::Shift)
+    {
+        if (is_basic_alpha(key.key))
+        {
+            // Shift + ASCII letters is just the uppercase letter.
+            key.modifiers &= ~Key::Modifiers::Shift;
+            key.key = to_upper(key.key);
+        }
+        else if (key.key < 0xD800 || key.key > 0xDFFF)
+        {
+            // Shift + any other printable character is not allowed.
+            throw key_parse_error(format("Shift modifier only works on special keys and lowercase ASCII, not '{}'", key.key));
+        }
+    }
+
     return key;
 }
 
@@ -53,7 +74,6 @@ static constexpr KeyAndName keynamemap[] = {
     { "pagedown", Key::PageDown },
     { "home", Key::Home },
     { "end", Key::End },
-    { "backtab", Key::BackTab },
     { "del", Key::Delete },
     { "plus", '+' },
     { "minus", '-' },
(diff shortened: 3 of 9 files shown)


src/keys.cc
```diff
@@ -192,6 +231,35 @@ UnitTest test_keys{[]()
...
+    kak_assert(parse_keys("<s-x>") == KeyList{ {'X'} });
+    kak_assert(parse_keys("<X>") == KeyList{ {'X'} });
+    kak_assert(parse_keys("X") == KeyList{ {'X'} });
+    kak_assert(parse_keys("<s-up>") == KeyList{ {Key::Modifiers::Shift, Key::Up} });
+    kak_assert(parse_keys("<s-tab>") == KeyList{ {Key::BackTab} });
+
+    try {
+        parse_keys("<s-X>");
+    } catch (runtime_error& err) {
+	kak_assert(err.what() == "Shift modifier only works on special keys and lowercase ASCII");
```
lenormf: Relying on error messages is error prone, wouldn't it be enough to check that an exception is thrown?

lenormf: I think it would make sense to have a custom exception for parsing errors, @mawww ?

src/keys.cc
```diff
@@ -102,9 +103,41 @@ KeyList parse_keys(StringView str)
         auto name_it = find_if(keynamemap, [&desc](const KeyAndName& item)
                                            { return item.name == desc; });
         if (name_it != std::end(keynamemap))
-            result.push_back(canonicalize_ifn({ modifier, name_it->key }));
+        {
+            Codepoint key = name_it->key;
+
+            if (modifier & Key::Modifiers::Shift)
```
mawww: move that logic in canonicalize_ifn

src/keys.cc
```diff
@@ -102,9 +103,41 @@ KeyList parse_keys(StringView str)
...
+            {
+                if (key < 0xD800 || key > 0xDFFF)
+                {
+                    if (!is_lower(name_it->key))
```
mawww: we can probably tolerate upper case as well `<s-X> == <s-x> == <X>`, use `is_basic_alpha` to check that.

src/keys.cc
```diff
@@ -167,9 +200,13 @@ String key_to_str(Key key)
 
     switch (key.modifiers)
     {
-    case Key::Modifiers::Control:    res = "c-" + res; named = true; break;
-    case Key::Modifiers::Alt:        res = "a-" + res; named = true; break;
-    case Key::Modifiers::ControlAlt: res = "c-a-" + res; named = true; break;
+    case Key::Modifiers::Control:         res = "c-" + res; named = true; break;
```
mawww: now that we have all those combination, a loop on Control, Alt, and Shift that prepend respectively "c-", "a-", and "s-" would be cleaner.

src/keys.cc
mawww: yeah, introducing a specific exception type, inheriting from runtime_error for that is fine. But frankly, parse_keys has not much other ways to fail than failing to parse a key... Also, introduce a macro for that in unit_test.hh `kak_expect_throw(exception_type, expression)` to keep the unit test easy to follow.

src/keys.hh
```diff
@@ -19,17 +19,21 @@ struct Key
...
         Alt     = 1 << 1,
+        Shift   = 1 << 2,
         ControlAlt = Control | Alt,
-
-        MousePress   = 1 << 2,
-        MouseRelease = 1 << 3,
-        MousePos     = 1 << 4,
-        MouseWheelDown = 1 << 5,
-        MouseWheelUp = 1 << 6,
+        ControlShift = Control | Shift,
```
mawww: No need to add all those combined alternatives, I'd rather remove ControlAlt.

src/keys.hh
```diff
@@ -109,6 +117,18 @@ constexpr Key ctrlalt(Key key)
 {
     return { key.modifiers | Key::Modifiers::ControlAlt, key.key };
 }
+constexpr Key ctrlshift(Key key)
```
mawww: same thing as for the modifiers, just let the user write `ctrl(alt(key))`, and remove `ctrlalt`

src/ncurses_ui.cc
```diff
@@ -569,14 +569,23 @@ Optional<Key> NCursesUI::get_next_key()
         {
         case KEY_BACKSPACE: case 127: return {Key::Backspace};
         case KEY_DC: return {Key::Delete};
+        case KEY_SDC: return {{Key::Modifiers::Shift, Key::Delete}};
```
mawww: use `shift(Key::Delete)`

src/normal.cc
```diff
@@ -2055,6 +2055,11 @@ static const HashMap<Key, NormalCmd, MemoryDomain::Undefined, KeymapBackend> key
     { {'K'}, {"extend up", move<LineCount, Backward, SelectMode::Extend>} },
     { {'L'}, {"extend right", move<CharCount, Forward, SelectMode::Extend>} },
 
+    { {Key::Modifiers::Shift, Key::Left}, {"extend left", move<CharCount, Backward, SelectMode::Extend>} },
```
mawww: `shift(Key::Left)`

Screwtapello:
I've pushed a new version of my changes, rebased on top of current master and squashed to a single commit as requested on IRC. I believe this addresses all the code-review comments so far, although I may have gone a little overboard on key-parsing exceptions: of the seven exception-types now in `exception.hh`, over half are specific ways in which key-parsing can fail. At least it makes the tests reliable!

src/exception.hh
```diff
@@ -23,6 +23,30 @@ private:
     String m_what;
 };
 
+struct key_parse_error : runtime_error
```
mawww: This does not belong to exception.hh, which does not contain any specific option type. it should be directly in keys.cc as nothing else is going to know about it except the unit test (and I am still on the fence that we really need to introduce an exception for that, so please keep it to a single one)

src/exception.hh
```diff
@@ -23,6 +23,30 @@ private:
...
+{
+    key_parse_error(String what)
```
mawww: `using runtime_error::runtime_error;` to inherit that constructor.

src/input_handler.cc
```diff
@@ -1568,8 +1568,10 @@ InputHandler::ScopedForceNormal::~ScopedForceNormal()
 
 static bool is_valid(Key key)
 {
+    Key::Modifiers valid_mods = (Key::Modifiers::Control | Key::Modifiers::Alt | Key::Modifiers::Shift);
```
mawww: constexpr

src/keys.cc
```diff
@@ -19,6 +19,28 @@ static Key canonicalize_ifn(Key key)
...
+
+    if (key.modifiers & Key::Modifiers::Shift)
+    {
+        if (is_basic_alpha(key.key))
+        {
+            // Shift + ASCII letters is just the uppercase letter.
+            key.modifiers &= ~Key::Modifiers::Shift;
+            key.key = to_upper(key.key);
+        }
+        else if (key.key == Key::Tab)
```
mawww: You know what, this is a good occasion to get rid of backtab, and move the backtab == shift-tab hack into ncurses.

src/keys.cc
```diff
@@ -165,13 +188,10 @@ String key_to_str(Key key)
...
         res = String{key.key};
 
-    switch (key.modifiers)
-    {
-    case Key::Modifiers::Control:    res = "c-" + res; named = true; break;
-    case Key::Modifiers::Alt:        res = "a-" + res; named = true; break;
-    case Key::Modifiers::ControlAlt: res = "c-a-" + res; named = true; break;
-    default: break;
-    }
+    if (key.modifiers & Key::Modifiers::Shift)   { res = "s-" + res; named = true; }
```
mawww: might be nitpicking, but I'd put them in the control, alt, shift order, as I tend to say "ctrl alt blah", "ctrl shift blah" "alt shift blah" rather than "alt ctrl blah" "shift ctrl blah" or "shift alt blah".

src/ncurses_ui.cc
```diff
@@ -570,14 +570,23 @@ Optional<Key> NCursesUI::get_next_key()
         {
         case KEY_BACKSPACE: case 127: return {Key::Backspace};
         case KEY_DC: return {Key::Delete};
+        case KEY_SDC: return {shift({Key::Delete})};
```
mawww: no need to braces here, `shift` already returns a Key.

src/normal.cc
```diff
@@ -2062,6 +2062,11 @@ static const HashMap<Key, NormalCmd, MemoryDomain::Undefined, KeymapBackend> key
     { {'K'}, {"extend up", move<LineCount, Backward, SelectMode::Extend>} },
     { {'L'}, {"extend right", move<CharCount, Forward, SelectMode::Extend>} },
 
+    { {shift(Key::Left)}, {"extend left", move<CharCount, Backward, SelectMode::Extend>} },
```
mawww: same, no need for braces here

Screwtapello:
Once again, thanks for being patient with my C++ inexperience.

I've pushed an updated branch that addresses comments from the latest review, rebased and squashed again.

As suggested, I've removed all support for `<backtab>` and left `<s-tab>` as the One True Syntax for that particular key combination. For the sake of users who might have `<backtab>` mappings configured, I added a section to `mappings.asciidoc` to help people migrate. I started by normalizing `<backtab>` to `<s-tab>` while keeping support for the old name, but it started getting complicated so I just went with removal instead.

doc/pages/mapping.asciidoc
```diff
@@ -110,3 +112,18 @@ be used:
 
 *<f1>*, *<f2>*, ...*<f12>*::
     Function keys.
+
+NOTE: Although Kakoune allows many key combinations to be mapped, not every
+possible combination can be triggered. For example, due to limitations in
+the way terminals handle control characters, mappings like *<c-s-a>* are
+unlikely to work in Kakoune's terminal UI.
+
+== Unsupported keys
```
mawww: The place to document breakage is the startup message in main. I dont want to litter the documentation with references to how things used to be. (Just a single sentence, "<backtab> is gone, use <s-tab>").

src/keys.cc
mawww: You are right, I was confused.

src/keys.cc
```diff
@@ -182,16 +200,39 @@ UnitTest test_keys{[]()
     KeyList keys{
          { ' ' },
          { 'c' },
+         { Key::Up },
          { Key::Modifiers::Alt, 'j' },
-         { Key::Modifiers::Control, 'r' }
+         { Key::Modifiers::Control, 'r' },
```
mawww: can you convert those to `alt('j'), ctrl('r'), and `shift(Key::Up)` while your are at it ?

src/ncurses_ui.cc
```diff
@@ -570,15 +570,24 @@ Optional<Key> NCursesUI::get_next_key()
         {
         case KEY_BACKSPACE: case 127: return {Key::Backspace};
         case KEY_DC: return {Key::Delete};
+        case KEY_SDC: return shift({Key::Delete});
```
mawww: I dont think you need the braces inside the shift call.

PR #1948, opened 2018-03-20
Support JSX fragments

danr:
These are tags with no name: <>...</>


danr:
Found a bug, closing this for now

PR #1950, opened 2018-03-21
Add extra_word_chars '-' to css.kak

Delapouite:
Hi

Very handy for virtually every CSS property names: `background-color`, `font-width`, `box-shadow`…


lenormf:
Using `-add` is probably better here.

mawww:
I think setting the value is fine, we dont want to add anything here, thats the word definition for css.

PR #1962, opened 2018-03-25
doc json_ui: Document the `refresh` request

lenormf:


Screwtapello:
OK, but what is the client supposed to *do* when it receives a refresh request?

lenormf:
Same as what they do when they get a `menu_hide` or `info_hide` request, it's in the name!

Screwtapello:
In my experience, "refresh" usually means "ncurses or your terminal might be confused, please redraw the screen from scratch", but in this context I'd expect that to either be an existing `draw()` request (if Kakoune is responsible for maintaining the for-lack-of-a-better-word framebuffer), or for that not to be part of the protocol at all (if the client is responsible for maintaining the framebuffer).

Looking at how `refresh()` is called in the existing Kakoune codebase, this is how I think it works:

- All the messages that sound like they draw something (`draw()`, `draw_status()`, `menu_show()`, etc.) should *not* be drawn to the screen immediately. Instead, the most recent value associated with each kind of message should be cached on the client, and a client-internal "dirty" flag set.
- When `refresh(false)` is received and the "dirty" flag is set, then draw the cached data to the screen.
- When `refresh(false)` is received and the "dirty" flag is clear, do nothing.
- When `refresh(true)` is received, ignore the "dirty" flag and draw the cached data to the screen.

lenormf:
The way this request is handled is implementation-specific, it's sent when the state of the UI has changed.

mawww:
I guess it would not hurt to have a description, but getting the name in the doc is already a step forward.

doppioandante:
When I was hacking together the qml client I never had the need to honor the refresh request, I always assumed it was needed specifically for ncurses
https://github.com/doppioandante/kakoune-qml/blob/f690e369a6426a957cf6c6ed67585d7affccdaca/src/KakounePane.qml#L102

mawww:
@doppioandante I believe the refresh request came into existence afterwards, however everything would work fine without it, its mostly an optimization to avoid useless redraws.

PR #1963, opened 2018-03-25
[RFC] src: Compare buffer contents before prompting for a reload

lenormf:
Hi,

It's sometimes annoying that the editor prompts for a buffer reload when the underlying file has been modified on the file system (e.g. saved or `touch`ed), but its contents haven't changed.

The CRC32 algorithm should be fast enough for hash computations, and collisions hopefully rare with text files.

HTH.

### DIFF
--- a/src/buffer.cc
+++ b/src/buffer.cc
@@ -25,6 +25,7 @@ struct ParsedLines
     BufferLines lines;
     ByteOrderMark bom = ByteOrderMark::None;
     EolFormat eolformat = EolFormat::Lf;
+    size_t hash = 0;
 };
 
 static ParsedLines parse_lines(StringView data)
@@ -49,9 +50,15 @@ static ParsedLines parse_lines(StringView data)
     while (pos < data.end())
     {
         const char* eol = std::find(pos, data.end(), '\n');
-        res.lines.emplace_back(StringData::create({{pos, eol - (crlf and eol != data.end() ? 1 : 0)}, "\n"}));
+        StringView line = {pos, eol - (crlf and eol != data.end() ? 1 : 0)};
+
+        res.lines.emplace_back(StringData::create({line, "\n"}));
+        res.hash = combine_hash(res.hash, hash_value(line));
         pos = eol + 1;
     }
+
+    res.hash = combine_hash(res.hash, hash_value(res.bom));
+    res.hash = combine_hash(res.hash, hash_value(res.eolformat));
 
     return res;
 }
@@ -74,9 +81,11 @@ Buffer::Buffer(String name, Flags flags, StringView data,
       m_flags{flags | Flags::NoUndo},
       m_history{m_next_history_id++, nullptr}, m_history_cursor{&m_history},
       m_last_save_history_cursor{&m_history},
-      m_fs_timestamp{fs_timestamp.tv_sec, fs_timestamp.tv_nsec}
+      m_fs_status{{fs_timestamp.tv_sec, fs_timestamp.tv_nsec}, data.length(), 0}
 {
     ParsedLines parsed_lines = parse_lines(data);
+
+    m_fs_status.file_hash = parsed_lines.hash;
 
     if (parsed_lines.lines.empty())
         parsed_lines.lines.emplace_back(StringData::create({"\n"}));
@@ -121,7 +130,7 @@ void Buffer::on_registered()
             run_hook_in_own_context("BufNewFile", m_name);
         else
         {
-            kak_assert(m_fs_timestamp != InvalidTime);
+            kak_assert(m_fs_status.timestamp != InvalidTime);
             run_hook_in_own_context("BufOpenFile", m_name);
         }
     }
@@ -305,7 +314,9 @@ void Buffer::reload(StringView data, timespec fs_timestamp)
     apply_options(options(), parsed_lines);
 
     m_last_save_history_cursor = m_history_cursor;
-    m_fs_timestamp = fs_timestamp;
+    m_fs_status.timestamp = fs_timestamp;
+    m_fs_status.file_size = data.length();
+    m_fs_status.file_hash = parsed_lines.hash;
 }
 
 void Buffer::commit_undo_group()
@@ -634,7 +645,7 @@ void Buffer::notify_saved()
 
     m_flags &= ~Flags::New;
     m_last_save_history_cursor = m_history_cursor;
-    m_fs_timestamp = get_fs_timestamp(m_name);
+    m_fs_status.timestamp = get_fs_timestamp(m_name);
 }
 
 BufferCoord Buffer::advance(BufferCoord coord, ByteCount count) const
@@ -693,13 +704,25 @@ BufferCoord Buffer::char_prev(BufferCoord coord) const
 timespec Buffer::fs_timestamp() const
 {
     kak_assert(m_flags & Flags::File);
-    return m_fs_timestamp;
+    return m_fs_status.timestamp;
 }
 
 void Buffer::set_fs_timestamp(timespec ts)
 {
     kak_assert(m_flags & Flags::File);
-    m_fs_timestamp = ts;
+    m_fs_status.timestamp = ts;
+}
+
+size_t Buffer::get_file_hash() const
+{
+    kak_assert(m_flags & Flags::File);
+    return m_fs_status.file_hash;
+}
+
+ByteCount Buffer::get_file_size() const
+{
+    kak_assert(m_flags & Flags::File);
+    return m_fs_status.file_size;
 }
 
 void Buffer::on_option_changed(const Option& option)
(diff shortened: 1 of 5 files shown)


src/crc32.cc
```diff
@@ -0,0 +1,84 @@
+#include "crc32.hh"
```
mawww: we already have hash_data which uses murmur3 and should be fast enough. I dont think we really need to add a crc32 implementation.

src/crc32.cc
```diff
@@ -0,0 +1,84 @@
...
+        update_hash = crc32_constants[(update_hash ^ *data++) & 0xFF] ^ (update_hash >> 8);
+
+    return ~update_hash;
+}
+
+crc32_t crc32_hash_file(StringView path)
+{
+    MappedFile fd{path};
+
+    return crc32_hash((uint8_t*)fd.data, fd.st.st_size);
```
mawww: This is not in sync with buffer content, it does not take into account end of line normalization.

src/crc32.cc
```diff
@@ -0,0 +1,84 @@
...
+{
+    update_hash = ~update_hash;
+
+    while (length--)
+        update_hash = crc32_constants[(update_hash ^ *data++) & 0xFF] ^ (update_hash >> 8);
+
+    return ~update_hash;
+}
+
+crc32_t crc32_hash_file(StringView path)
```
mawww: If we want a file hashing function, it probably should be in file.cc.

src/crc32.cc
lenormf: Arguably it would be simpler/faster to instantiate a `Buffer` on the same file, and compare two objects directly without hashing.

mawww: I would not think that, creating a buffer means we need to copy the whole data from the file buffer, whereas we can hash it in place, we just need to ignore the `\r`.

lenormf: Do lines in `m_lines` contain the full `LF`/`CRLF` line ending? I'd rather filter those out in the `Buffer` function and only have to parse the whole original mmap'ed file.

src/crc32.cc
lenormf: Side question, is `murmur3` that much better than a plain old CRC32? It seems to be using substantially more cycles, maybe be it's not the best algorithm to hash entire files of arbitrary length? Using it for key hashes is sound enough (because of collisions), I wonder about files (whose contents are generally plain text).

src/crc32.cc
lenormf: Another point is that incremental hashing using Murmur3 is not implemented, so I would need to do it to be able to hash lines from `m_lines` and compare that to the file's total hash.

https://github.com/aappleby/smhasher/blob/master/src/PMurHash.c

danr:
Why do you even hash? Why not just compare the buffer contents with the file contents? Same complexity and no false negatives.

lenormf:
I went the complicated way and implemented per-character checks, it seems to be working.

lenormf:
Doesn't work when the file is originally empty! I'll look into it.

src/buffer.cc
```diff
@@ -15,25 +15,19 @@
...
 #include "window.hh"
 
+#include <cstring>
 #include <algorithm>
 
 namespace Kakoune
 {
 
-struct ParsedLines
+static BufferLines parse_lines(StringView data, ByteOrderMark* bom, EolFormat* eolformat)
```
mawww: why change that signature ? That now requires to create those objects before.

src/buffer.hh
```diff
@@ -249,14 +252,17 @@ private:
         StringView back() const { return BufferLines::back()->strview(); }
     };
     LineList m_lines;
+    ByteOrderMark m_byte_order_mark = ByteOrderMark::None;
+    EolFormat m_eol_format = EolFormat::Lf;
```
mawww: Those are out of sync with the options, and only represent the state of the buffer at last reload, I am not keen on storing that.

src/buffer.hh
```diff
@@ -191,6 +191,9 @@ public:
     // returns true if the buffer is in a different state than
     // the last time it was saved
     bool is_modified() const;
+    // returns true if the buffer is in a different state than
+    // the underlying file
+    bool is_file_modified() const;
```
mawww: I dont like having that in the Buffer class, we have everything we need to implement that externally, `buffer_utils.cc` would be a good place to have this.

src/buffer.hh
```diff
@@ -249,14 +252,17 @@ private:
         StringView back() const { return BufferLines::back()->strview(); }
     };
     LineList m_lines;
+    ByteOrderMark m_byte_order_mark = ByteOrderMark::None;
+    EolFormat m_eol_format = EolFormat::Lf;
 
     String m_name;
     String m_display_name;
     Flags  m_flags;
+    ByteCount m_contents_size;
```
mawww: Same comment, I am not keen on storing that as well, why would we want to keep around the buffer size at last reload ? We can compute it if needed.

src/buffer.cc
```diff
@@ -627,6 +625,37 @@ bool Buffer::is_modified() const
...
+        if (m_byte_order_mark == ByteOrderMark::Utf8)
+            fd.data += 3;
+
+        for (size_t i = 0; i < line_count(); i++)
+        {
+            const StringDataPtr& bufdata_str = line_storage(i);
+            const size_t bufdata_len = bufdata_str->length - 1;
+
+            if ((!bufdata_len and fd.data[eol_offset - 1] != '\n')
+                or (bufdata_len and std::strncmp(fd.data, bufdata_str->data(), bufdata_len)))
```
mawww: use `std::equal`

src/buffer.cc
```diff
@@ -627,6 +625,37 @@ bool Buffer::is_modified() const
...
+
+        if (m_byte_order_mark == ByteOrderMark::Utf8)
+            fd.data += 3;
+
+        for (size_t i = 0; i < line_count(); i++)
+        {
+            const StringDataPtr& bufdata_str = line_storage(i);
+            const size_t bufdata_len = bufdata_str->length - 1;
+
+            if ((!bufdata_len and fd.data[eol_offset - 1] != '\n')
```
mawww: That looks strange, you are looking at the first or second character of `fd.data`, and checking if its `\n`, I think thats not what you meant.

lenormf: I don't know if the behavior of `strncmp()` is well defined when its size argument is 0, so I'm verifying that the corresponding line in the file is empty.

src/buffer.hh
lenormf: Getting the file size requires issuing an expensive syscall, we can simply store the integer since it's computed right before every time we need it.

src/buffer.hh
lenormf: I would need getters to get the BOM and EOL format.

src/buffer.cc
lenormf: I need to know the BOM and EOL format every a reload is checked for, the only thing that's temporarily needed after this function returns is the `BufferLines` object.

src/buffer.hh
mawww: Those are in the options already

src/buffer.hh
mawww: Which syscall ? my understanding is that this is the size of the file at last load. Is that correct to compare to size at last load instead of current size ? In that case why are we comparing the current content of the buffer instead of the last loaded content ?

src/buffer.cc
mawww: `std::equal` will make all that cleaner.

src/buffer.cc
mawww: those are enums, they are extremely cheap to copy around, there is no need to pass them by pointer to the function, we can just copy the ones from ParsedLines. Moreover, they are stored in options, and I am not convinced they should be duplicated in the buffer.

src/buffer.hh
lenormf: Getting a file's size is generally done using `stat()`.

Saving the file size only on reload is probably not a good idea, as it will be different from the size of the buffer on a simple `:write`.

lenormf:
That was what my first implementation was doing, why revert back?

mawww:
My concerns on the first implementation were mostly about the crc32 hashing and the general code organization. One thing that changed is that I was initially thinking we should compare current buffer state to new filesystem state, but I now think that we should compare last reload buffer state to new filesystem state.

Besides, I think my suggestion is still pretty different from what you had, I'd like to extend m_fs_timestamp to be a more general m_fs_status struct that records additional data regarding the file state on the filesystem. I would prefer the filesystem hashing to be done with no memory allocations, but I think as a start, to avoid complex hashing implementation, we could just use the existing parse_lines function and generate the hash from that (iterate on lines and hash combine them, hash combine eol and bom state with that).

lenormf:
I haven't had much time lately, so I pushed what I did to take the requested changes into account. It's not tested at all, and I still don't like that watered-down implementation of `buffer_needs_reload()`, but I want to know if this is the right direction.

lenormf:
Re-worked in #3225.

PR #1965, opened 2018-03-27
Docs: add missing `*` for -buffer switch of exec/eval

Delapouite:


lenormf:
I think that this was originally not intended to be documented, being kind of a hack. I suppose we're not getting rid of it now, so it's probably better to promote this to an actual feature.

PR #1982, opened 2018-04-01
src: Brand plain text files

lenormf:
The editor is slowly gaining traction, and I think it's time we hold
the project to the same standards as other professional ones. The
community will never grow, and the editor will never become popular
if attention isn't brought to it, so I've taken the matter into my
own hands.

A short branding message is appended to the end of text buffers, to
show everybody who reads them what editor they were written with. But
don't worry: the message will not appear on the user interface as
you're editing files and saving them, everything happens in the
background.

Sent with my iPhone.


SolitudeSF:
this, but unironically

PR #1988, opened 2018-04-03
Docs: fix asciidoc list formatting issue

Delapouite:
Hi

image

Similar problem than in https://github.com/mawww/kakoune/pull/1923
A new line is needed between each item in a list.


lenormf:
A more robust way to handle that would be to parse those asciidoc line endings (`:`), and insert as many newlines as there are extra colons, dynamically. It might not be worth it to get closer to an asciidoc parser though.

PR #1998, opened 2018-04-13
Update README.asciidoc to mention FreeBSD

t6:
I've just committed a FreeBSD port of kakoune to the ports tree [1]. A binary package should also be available in the next couple of days.
It would be nice if it could be mentioned in the README.

Thank you :)

[1] https://www.freshports.org/editors/kakoune


mawww:
Thanks !

Issue #1995, opened 2018-04-11
Mapping `<f11>` has no effect when `ncurses_shift_function_key` is set to `10`


fsub:
### Steps 
```
set-option global ui_options ncurses_shift_function_key=10   # rxvt
map global normal <f11> :gdb-step<ret>
```

### Outcome 
Pressing F11 has no effect.

### Expected 
Pressing F11 executes `gdb-step`.

mawww:
I am not sure that this is a buggy behaviour, we basically tell NCursesUI that <s-F1> is at `F11`, so I'd expect any key that generates F11 to be interpreted as <s-F1>. Frankly its a strange choice rxvt made to generates F11 for <s-F1>, but I am not convinced we should change the logic to accommodate the fact that rxvt makes F11 and <s-F1> ambiguous, I think if you say the shift_function_keys start after F10, then F11 should be treated as <s-F1>.

Hopefully there is a way to tell rxvt to change that and use xterm compatible mappings (F13 = <s-F1>).

fsub:
@mawww It seems arbitrary whether <s-f1> or <f11> prevails. @Screwtapello suggested to support multi-mappings. In my opinion that's the best option since it reflects the situation that, in rxvt, Shift-F1 and F11 are the same.

I don't think that rxvt allows changing function key logic.

fsub:
@mawww Contrary to what I thought, rxvt allows you to redefine keys. Not sure whether that's what I want, though.

mawww:
As I said on #2000, I dont think we should compensate a design flaw in rxvt by complexifying Kakoune. Sending two keystrokes seems wrong as the user only hit one. That means we need to disambiguate.

I think disambiguating using <s-F1> when we get <F11> is what makes the most sense because thats the simplest rule regarding `ncurses_shift_function_keys`. There is no strong reason to prefer one over the other, so I'd pick the simplest to implement.

What that means is that you'd need either to tell rxvt to send less ambiguous keys for shifted function keys, or tweak your kakoune configuration to map <s-F1> instead of F11. Note that you should be able to have terminal dependent configuration by switching on the kak_client_env_TERM env var.

fsub:
@mawww Thank you for explaining your point of view. I have no problems working around the issue.

Putting all the blame to rxvt does not seem to be fair. `kf11` and `kf21` are completely unambiguous.

PR #2000, opened 2018-04-14
Multi mapped function keys

fsub:
Proof of concept.
See #1995 and #1996.

### DIFF
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -253,8 +253,9 @@ NCursesUI::NCursesUI()
         if (not m_on_key)
             return;
 
-        while (auto key = get_next_key())
-            m_on_key(*key);
+        for (auto keys = get_next_keys(); not keys.empty(); keys = get_next_keys())
+            for (const auto& key : keys)
+                m_on_key(key);
       }},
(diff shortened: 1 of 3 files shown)


src/ncurses_ui.cc
```diff
@@ -612,15 +613,31 @@ Optional<Key> NCursesUI::get_next_key()
...
+            {
+                if (c == KEY_F(i + 1))
+                {
+                    keys.emplace_back(Key::F1 + i);
+                    break;
+                }
+            }
+            for (int i = 0; i < 12; ++i)
+            {
+                if (c == KEY_F(m_shift_function_key + i + 1))
```
Screwtapello: Since this function returns a list now, I think you can do this in a single loop like before:

    for (int i = 0; i < 12; ++i)
    {
        if (c == KEY_F(i + 1)) keys.emplace_back(Key::F1 + i);
        if (c == KEY_F(m_shift_function_key + i + 1)) keys.push_back(shift(Key::F1 + i));
    }

fsub: Yes, I wasn't sure about this. Let us first see if @mawww agrees with the general approach.

PR #2005, opened 2018-04-17
Adds openbsd support

### LINKED ISSUE #2004: OpenBSD support
Currently it's not compiling on openbsd 6.3-current.

I believe the two main culprits are the Makefile and file.cc

I was able to get the Makefile working by copy pasting the entry for dragonfly and changing it to openbsd, but that doesnt work for file.cc (it will compile, but it will abort) 

Obviously this is due to openbsd working differently than dragonfly, I'll investigate further when I have time, but I'm opening an issue here to track my progress and / or get the ball rolling.

For the record, the makefile error was that it could not find ncursesw, but that error disappeared after copy pasting the dragonfly / freebsd entry. Not sure if it should be adapted to openbsd yet.

The file.cc error is `finding executable path is not supported on this platform" in this case, copying the freebsd / dragonfly entry makes it compile, but abort at runtime. I'm fairly sure the abort is due to file.cc mod and not the makefile mod.

Superpat:
Seems to work on openbsd 6.3-current but needs more testing. Had to
hardcode the binary path as openbsd considers getting the executable
path at runtime a security flaw. 

Closes #2004

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -57,6 +57,10 @@ else ifeq ($(os),DragonFly)
     LIBS += -lncursesw
     CPPFLAGS += -I/usr/local/include
     LDFLAGS += -L/usr/local/lib
+else ifeq ($(os),OpenBSD)
+    LIBS += -lncursesw
+    CPPFLAGS += -I/usr/local/include
+    LDFLAGS += -L/usr/local/lib
 else ifneq (,$(findstring CYGWIN,$(os)))
     CPPFLAGS += -D_XOPEN_SOURCE=700
     LIBS += -lncursesw -ldbghelp
--- a/src/file.cc
+++ b/src/file.cc
@@ -22,9 +22,6 @@
 #if defined(__FreeBSD__)
 #include <sys/sysctl.h>
 #endif
-#if defined(__OpenBSD__)
-#include <sys/sysctl.h>
-#endif
 
 #if defined(__APPLE__)
 #include <mach-o/dyld.h>
--- a/src/file.cc
+++ b/src/file.cc
@@ -22,6 +22,9 @@
 #if defined(__FreeBSD__)
 #include <sys/sysctl.h>
 #endif
+#if defined(__OpenBSD__)
+#include <sys/sysctl.h>
+#endif
 
 #if defined(__APPLE__)
 #include <mach-o/dyld.h>
@@ -571,6 +574,8 @@ String get_kak_binary_path()
     kak_assert(res != -1);
     buffer[res] = '\0';
     return buffer;
+#elif defined(__OpenBSD__)
+    return "/usr/local/bin/";
 #else
 # error "finding executable path is not implemented on this platform"
 #endif


Superpat:
I also ran the tests on my personal machine and I noticed that only one test fails.

chiron$ gmake test
/test && ./run
1J
ackward-search
ulti-delete
undo-multi-replace-at-end
-scroll-onto-tab

nt
uct
     indent-if-brace
kell
tag
de-list-item
    open-after-setext-heading-1
shouldnt-duplicate
    normal
      align
      align-tab
      all
      append
      append-at-eol
      change
      clear-selections
      codepoint-width
        tab-width
        vertical-movement
      concat-paste
      concat-paste-line
      convert-spaces-to-tabs
      convert-tabs-to-spaces
      copy-sel-above
      copy-sel-below
      copy-sel-below-multi-line
      deindent
      delete
      extend-lines
      extra-word-chars
      facedesc-invalid-syntax
        attributes
        bg-color
      find-char-backward
      find-char-backward-extending
      find-char-forward
      find-char-forward-extending
      goto
      goto-extending
        buffer-bottom
        buffer-end
        buffer-top
        last-change
        line-begin
        line-end
        buffer-bottom
        buffer-end
        buffer-top
        file
        last-buffer
        last-change
        line-begin
        line-end
      indent
      insert
      insert-at-line-start
      insert-replace
      jump
        backward
        backward-count
        backward-invalid-count
        forward
        forward-count
        forward-invalid-count
      keep-cmd
--- /home/patrick/Repositories/kakoune/test/./normal/keep-cmd/selections Fri Feb 16 08:15:29 2018
+++ selections   Mon Apr 16 21:01:59 2018
@@ -1 +1 @@
-foo:bar
+foo:rha:bar

debug buffer:

      lower-case
      match-char
      move
        down
        down-extending
        left
        left-extending
        right
        right-extending
        up
        up-extending
      next-big-word
      next-big-word-extending
      next-match
      next-match-appending
      next-word
      next-word-extending
      object
        around
        around-parent
          angle
          braces
          brackets
          parenthesis
          angle
          argument
            multi-level
            single-level
          big-word
          braces
          brackets
          double_quote
          grave_quote
          indent
          paragraph
          parenthesis
          sentence
          single_quote
          slash
          word
        drop-non-whitespace
        end
        end-extending
          angle
          argument
          big-word
          braces
          brackets
          double_quote
          grave_quote
          indent
          paragraph
          parenthesis
          sentence
          single_quote
          word
          angle
          argument
          big-word
          braces
          brackets
          double_quote
          grave_quote
          indent
          paragraph
          parenthesis
          sentence
          single_quote
          word
        inner
          angle
          argument
            multi-level
            single-level
          big-word
          braces
          brackets
          double_quote
          grave_quote
          indent
          paragraph
          parenthesis
          sentence
          single_quote
          slash
          word
        on-end
          around
            angle
            braces
            brackets
            parenthesis
          inner
            angle
            braces
            brackets
            parenthesis
          to-end
            angle
            braces
            brackets
            parenthesis
        on-start
          around
            angle
            braces
            brackets
            double_quote
            grave_quote
            parenthesis
            single_quote
            slash
          inner
            angle
            braces
            brackets
            double_quote
            grave_quote
            parenthesis
            single_quote
          to-start
            angle
            braces
            brackets
            parenthesis
        start
        start-extending
          angle
          argument
          big-word
          braces
          brackets
          double_quote
          grave_quote
          indent
          paragraph
          parenthesis
          sentence
          single_quote
          word
          angle
          argument
          big-word
          braces
          brackets
          double_quote
          grave_quote
          indent
          paragraph
          parenthesis
          sentence
          single_quote
          word
      open-above
      open-below
      open-multiple-above
      open-multiple-below
      paste-after
      paste-before
      pipe
      pipe-to
      previous-big-word
      previous-big-word-extending
      previous-word
      previous-word-extending
      record-macro
      redo
      reload
      repeat-insert
      repeat-select
        repeat-end-paragraph
        repeat-find-char
      replace
      replace-lines
      replay-complex-insert
      replay-macro
      rotate
      rotate-content
      save-selections
      search
      search-extending
      search-reverse
      search-reverse-extending
      search-reverse-rightmost
        missed-match
        overlap
      select
      select-line
      select-line-extending
      split
      split-at-begin
      split-multiple-lines
      switch-case
      to-char-backward
      to-char-backward-extending
      to-char-forward
      to-char-forward-extending
      trim
      trim-lines
        table
      undo
      undo-after-replace-lines
      upper-case
      user-modes
        lock
        once
      yank
    regression
      0-comment-after-command
      0-compute-modified-range-crash
      0-crash-on-backspace-at-buffer-begin
      0-crash-on-pipe-with-selection-access
      0-crash-on-specialy-crafted-modeline
      0-open-below-should-not-move-cursor-on-eol
      0-replace-last-eol-with-eol
      0-undo-change-at-eof
      1014-ambiguous-kak_selections
      1051-crash-on-empty-param
      1053-crash-on-deletion-and-paste
      1074-comment-leader-autoinsert-error
      1105-object-selection-behaviour
      1111-unexpected-behaviour-on-new-line-in-comment
      1118-misaligned-comment-start
      1129-capture-groups-are-broken
      1195-infinite-loop-in-regex-matching
      1227-segfault-on-option-access
      1233-corrent-cursor-handling-for-indent-objects
      1275-replaced-range-split
      1308-wrong-behaviour-of-kak-indent-on-newline
      1382-column-highlighter-broken-on-horizontal-scroll
      1433-scrolloff-broken-with-soft-wrap
      1435-misplaced-cursor-with-show_matching-hl
      1453-show_whitespaces-highlighter-breaks-tab-alignment
      1459-assertion-on-wrapped-long-line
      1469-assert-on-repeat-insert
      1504-assertion-on-incorrect-pipe-use
      1518-wrong-undo-handling-with-fifo-buffers
      1525-lua-indent-error
      1580-A-not-moving-to-eol
      1680-crash-with-dot-and-alt-semicolon
      1731-wrap-hidden-buffer
      1741-scrolloff-hides-end-of-lines
   cd ..29-unicode-as-  kak-tests.DTmeRR      compose
      beh      del-empty-line
      eof-m        inline-sort
      pipe-at-eof
      -b      unicode
    display
      horizontal-scroll
      horizontalis      horizontal-scroll-with-tab
      vertical-scroll
    highlight
      column          multi-columns
      regions
      wrap
        avoid-odd-places
    inderi      c-family
        align-while-expr
        close-namespace
        close-strac        construct-align
        indent-after-parenthesis
        indent-if-body
   r-        move-brace
        multiline-nested-align
        nested-align
        open-function
      has33        before-comment
        indented-comment
        inside-comment
      html
        indent-closing-o
      lua
        insert-end-only-when-needed
      markdown
        inside-complex-blockquote
        insird        inside-nested-list-item
        inside-simple-blockquote
        inside-strong-emphasis
        list-item
    -i        open-after-setext-heading-2
      ruby
        following-blocks-should-not-prevent-end
        indent-after-class
        method-definition-1

Superpat:
Technically on openbsd you would use ${PREFIX} to define the path for the files, but that only works in the context of a port in the ports tree. So I'm wondering if I should switch it here which would require openbsd users building from the git to modify the source but make the port simpler or keep it hardcoded here and apply a patch in the port.

mawww:
It would be nice to find whats happening for that test case.

I wonder about hard coding the path, I think passing prefix as a `-DKAK_PREFIX=...` and using that macro in the openbsd code would be cleaner.

dahlbaek:
Using the last OpenBSD snapshot, I had to use `CXX=clang++ gmake` to compile the openbsd branch. Unfortunately, I am not sufficiently familiar with OpenBSD or kakoune to know if this has to do with weird configuration on my end, or if it is just the standard procedure on OpenBSD.

mawww:
Superseeded by #2131

PR #2006, opened 2018-04-17
tmux version check fails on FreeBSD

t6:
tmux-repl fails on FreeBSD and claims that tmux' version is too old
even though tmux 2.7 is installed.

The problem is that
```
expr "$(tmux -V)" : 'tmux \([0-9]*\|master\)'
```
fails on FreeBSD.  It works fine with GNU expr.

Attempt to fix it by expanding the expression a bit.

### DIFF
--- a/rc/extra/tmux-repl.kak
+++ b/rc/extra/tmux-repl.kak
@@ -4,8 +4,8 @@
 hook global KakBegin .* %{
     %sh{
         if [ -n "$TMUX" ]; then
-            VERSION_TMUX=$(tmux -V)
-            VERSION_TMUX=$(expr "${VERSION_TMUX}" : 'tmux \([0-9]*\|master\)')
+            VERSION_TMUX=$(tmux -V | cut -d' ' -f2)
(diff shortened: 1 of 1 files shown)


rc/extra/tmux-repl.kak
```diff
@@ -5,7 +5,7 @@ hook global KakBegin .* %{
     %sh{
         if [ -n "$TMUX" ]; then
             VERSION_TMUX=$(tmux -V)
-            VERSION_TMUX=$(expr "${VERSION_TMUX}" : 'tmux \([0-9]*\|master\)')
+            VERSION_TMUX=$(expr \( "${VERSION_TMUX}" : 'tmux \([0-9]*\)' \) \| \( "${VERSION_TMUX}" : 'tmux \(master\)' \))
```
mawww: that starts to be pretty complex, what we want to get is whatever is before the first dot of the second word. Can you try to just use something like:
```
VERSION_TMUX=$(tmux -V | cut -d' ' -f2)
VERSION_TMUX=${VERSION_TMUX%%.*}
```

PR #2010, opened 2018-04-22
Fix building with ncurses on DragonFly 5.2

shassard:
Kakoune fails to build on my DragonFlyBSD 5.2 box due to pathing issues related to ncurses. This patch fixes the build.


lenormf:
Doesn't `pkg-config` work? It'd be more efficient than hardcoding paths.

shassard:
@lenormf the pkg-config default behavior seems to work fine on dragonfly. I'll send an update removing this platform specific logic that isn't needed anymore.

PR #2016, opened 2018-04-28
Small improvements of TRAMPOLINE

1nquisitive:
Did some small improvements of the Trampoline.

* Changed the typo "edition" to the correct word "editing".
* Clarified for beginners which mode you start in and in which mode you input the commands/primitives.
* Removed the <, V, ^, and > keys from the paragraph about movements because they're confusing.

### DIFF
--- a/contrib/TRAMPOLINE
+++ b/contrib/TRAMPOLINE
@@ -9,11 +9,11 @@
                   Mawww's experiment for a better code editor
 +=----------------------------------------------------------------------------=+
 
-This walk-through is an introduction to Kakoune's basic edition capabilities
+This walk-through is an introduction to Kakoune's basic editing capabilities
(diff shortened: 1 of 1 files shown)


lenormf:
The arrow keys are still heavily used as far as I know, so removing them is not a good idea. Replace them with the unicode symbols (→, ← etc) ?

1nquisitive:
That works for me. My main issue was that the keys was incorrect. I'll upload a new commit restoring the keys but change them to the unicode characters. It doesn't seems that it will create problems with Asciidoc.

@lenormf How did you do the initial formatting of the document? Did you have a tools for creating the keys "icons" or did you do all the formatting manually?

lenormf:
I used Kakoune to write the document.

PR #2028, opened 2018-05-05
Improve SPEC file, use release tag instead of commit hash

scadu:
Since we already have a tagged release I think it would be nice to update the existing SPEC file. Built and tested on Fedora 28 x86_64, everything seems to be just fine.

@jkonecny12 would it be possible to use the release instead of playing with commit hashes?

### DIFF
--- a/contrib/kakoune.spec
+++ b/contrib/kakoune.spec
@@ -1,11 +1,11 @@
 Name:           kakoune
-Version:        0
-Release:        208.20160511git84f62e6f%{?dist}
+Version:        2018.04.13
+Release:        1%{?dist}
 Summary:        Vim inspired editor
 
 License:        Unlicense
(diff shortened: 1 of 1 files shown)


scadu:
For some reason the build fails on copr (cannot find src directory), while it works perfectly locally. When providing copr with locally prepared SRPM it's also fine. 
SRPM on copr is build with rpkg, locally with `rpmbuild -bs $specfile`.

Will check rpkg and update the PR appropriately.

scadu:
For some reason the build fails on copr (cannot find src directory), while it works perfectly locally. When providing copr with locally prepared SRPM it's also fine.
SRPM on copr is build with rpkg, locally with rpmbuild -bs $specfile.

Will check rpkg and update the PR appropriately if required.

PR #2047, opened 2018-05-15
Add <c-f> in prompt mode to insert default value

alyssais:
I'd like to bind the right arrow to this if the prompt is empty (as would ironzorg). I think that could probably be implemented in an rc file, except that currently there's no way to access the value of the default value at all. This PR adds <c-f> to insert the default value in a prompt.

### DIFF
--- a/doc/pages/keys.asciidoc
+++ b/doc/pages/keys.asciidoc
@@ -703,6 +703,9 @@ The following keys are recognized by this mode to help edition.
 *<c-v>*::
     insert next keystroke without interpreting it
 
+*<c-f>*::
+    insert the default value for this prompt
+
 *<c-o>*::
     disable auto completion for this prompt
 
--- a/src/input_handler.cc
+++ b/src/input_handler.cc
@@ -896,6 +896,13 @@ public:
                 return;
             }
         }
+        else if (key == ctrl('f'))
+        {
+            m_line_editor.insert(m_empty_text);
+            display();
+            m_line_changed = true;
+            m_refresh_completion_pending = true;
+        }
         else
         {
             m_line_editor.handle_key(key);


alyssais:
```
09:18 <qyliss^work> In search/select mode, is there a way to make the default value (the value of /) editable in the search field?
09:18 <qyliss^work> Instinctively I'd hit right and expect to find my cursor at the end of that value, but it doesn't do anything.
09:18 <ironzorg> yea I was wondering the same
09:25 <ironzorg> are you creating the issue qyliss^work or am I doing it?
09:26 <qyliss^work> I'm wondering if I can just bind a key
09:28 <ironzorg> `/<ret>/<up>` ?
09:28 <ironzorg> it works, but really that's a workaround for a sensible feature request
09:30 <qyliss^work> Yeah I guess
```

lenormf:
PING

lenormf:
PING

lenormf:
```diff
commit 3f3257315f6feb2e7b1062d397e8db04cba73fa6
Author: Alyssa Ross <hi@alyssa.is>
Date:   Tue May 15 09:45:12 2018 +0100

    Add <c-f> in prompt mode to insert default value

diff --git a/doc/pages/keys.asciidoc b/doc/pages/keys.asciidoc
index 674541a5..6f9c8526 100644
--- a/doc/pages/keys.asciidoc
+++ b/doc/pages/keys.asciidoc
@@ -810,6 +810,8 @@ The following keys are recognized by this mode to help with editing
 
     *w*:::
         explicit word completion (from current buffer)
+*<c-f>*::
+    insert the default value for this prompt
 
 *<c-o>*::
     toggle automatic completion
diff --git a/src/input_handler.cc b/src/input_handler.cc
index 85904e7f..14bb1e99 100644
--- a/src/input_handler.cc
+++ b/src/input_handler.cc
@@ -1001,6 +1001,13 @@ public:
             push_mode(new Normal(context().input_handler(), true));
             return;
         }
+        else if (key == ctrl('f'))
+        {
+            m_line_editor.insert(m_empty_text);
+            display();
+            m_line_changed = true;
+            m_refresh_completion_pending = true;
+        }
         else
         {
             if (key == ' ' and
```

krobelus:
I had thought that there was a shortcoming in `<c-p>` and `<c-n>` which would give a good motivation for this.

However, I'm struggling to find such a scenario, maybe something has changed..

I have found what looks like a related bug in `<c-n>`: if I run

- `:set-register / first second<ret>`
- `/`

And then type the four keys below, the prompt value updates as follows

```
<c-p> # second
<c-p> # first
<c-n> # first <--- I expect "second" here, is this a bug?
<c-n> # back to the empty prompt (looks correct)
```

krobelus:
meh, this patch breaks the default readline binding of `<c-f>` (move one character to the right) :(

lenormf:
What mapping would you suggest instead?

krobelus:
`<c-p>` is already strictly better so I don't see the motivation

PR #2052, opened 2018-05-16
Allow mouse-wheel scrolling while dragging a selection.

### LINKED ISSUE #2051: Cannot scroll with the mouse-wheel while selecting text
### Steps 

1. Open a document long enough that it doesn't all fit on screen
2. Hold down the mouse button on some text to begin selecting it
3. While holding down the mouse button with one finger, roll the scroll-wheel to bring more of the document into view

### Outcome 

* Document scrolls
* Anchor and cursor are moved to where the mouse-cursor was when the scroll-wheel was touched
* Selection has ended, so moving the mouse cursor no longer moves the selection cursor.

### Expected 

* Document scrolls
* Anchor stays where it was when selection began
* Cursor still follows the mouse around, so more than a single screen of text can be selected with the mouse.

**Note:** Keys like `<c-f>` and `<c-b>` work while selecting with the mouse, it's just scroll-wheel-scrolling that's broken.

Screwtapello:
Fixes #2051.

### DIFF
--- a/src/input_handler.cc
+++ b/src/input_handler.cc
@@ -125,13 +125,21 @@ struct MouseHandler
             return true;
 
         case Key::Modifiers::MouseWheelDown:
-            m_dragging = false;
             scroll_window(context, 3);
+            if (m_dragging) {
+                cursor = context.window().buffer_coord(key.coord());
+                selections.main() = {buffer.clamp(m_anchor), cursor};
+                selections.sort_and_merge_overlapping();
+            }
             return true;
 
         case Key::Modifiers::MouseWheelUp:
-            m_dragging = false;
             scroll_window(context, -3);
+            if (m_dragging) {
+                cursor = context.window().buffer_coord(key.coord());
+                selections.main() = {buffer.clamp(m_anchor), cursor};
+                selections.sort_and_merge_overlapping();
+            }
             return true;
 
         default: return false;


src/input_handler.cc
```diff
@@ -125,13 +125,21 @@ struct MouseHandler
             return true;
 
         case Key::Modifiers::MouseWheelDown:
-            m_dragging = false;
             scroll_window(context, 3);
+            if (m_dragging) {
```
mawww: I am not sure how that would give correct results as  the `buffer_coord` call is using an out of date, pre-scrolling display buffer. Are you confident you get the correct cursor position on scroll ?

Screwtapello: If I hold down the button 1 and scroll by one "notch", carefully keeping my pointing device steady, then no: the cursor stays where it was before the scroll occurred, it's not updated to the post-scroll position.

However, after scrolling by a second "notch", the cursor moves to where it was after the first scroll (before the second scroll), so the lag between scrolling and (text) cursor movement is limited. Additionally, in practice after scrolling I move the mouse to the place I where want to end the selection, which updates the cursor position correctly.

If there's a more correct way to scroll the window, though, I'm happy to make changes.

mawww: It would be a bigger change to get that to work correctly,, however I suspect we can get the exact same behaviour by just removing the `m_dragging = true;` lines. Could you check that ? If thats the case, such a change will be good enough for now.

Screwtapello: Hmm... do you mean removing all the `m_dragging` handling in `MouseHandler`, including setting `m_dragging` on mouse-press, clearing it on mouse-release and testing it in the mouse-wheel handlers?

If I do that, drag-to-select works just as it did before (since ncurses doesn't send mouse-move events unless a button is held anyway), but now the mouse-wheel always acts like a click as well as a scroll, disrupting the current selection.

Screwtapello: As discussed on IRC, I tried going back to the original code and just removing the `m_dragging = false;` lines from the mouse-wheel handling.

After that change, scrolling while dragging moves the anchor to the cursor position before the scroll occurs, but it returns to its original position on mouse-move or mouse-release. Scrolling while *not* dragging moves the anchor to the cursor position and it never goes back.

PR #2074, opened 2018-05-24
Add CommandCompleter for reg command

Delapouite:
Hello

This PR follows the existing autocompletion mechanism like with `set-option`. While typing : 

```:set-option global aligntab ``` the autocompletion menu shows `false` which is the current value of `aligntab` option.

The behavior is now the same with the `set-register` command. While typing:
```:set-register g ``` if the content of the register `g` is `foo`, `foo` is displayed in the autocomplete menu.

This has the nice bonus effect of letting the user inspect the content of a given register.


Delapouite:
I improved this PR to not be limited to the first token anymore.

lenormf:
PING

lenormf:
PING

PR #2075, opened 2018-05-24
docs: Add more info about which keys use a default register

Delapouite:
Hello

This PR attemps to clarifies which keys use registers, the current page is not exhaustive (only mentionning `@` and `^` registers)

- there was a small `Searching` section far down the page only listing `*` and `<a-*>`. So I regrouped all search commands together in this section and move it in the upper part of the page. This way the user can see that they all share the `/` register in common.
- add mention about yank/paste using the `"` register
- add mention about pipe/insert/append using the `|` register
- add mention about the command prompt using the `:` register


doc/pages/keys.asciidoc
```diff
@@ -367,6 +327,61 @@ the Shift modifier and moving will extend each selection instead.
 *<a-(>*::
     rotate selections content backward
 
+== Changes through external programs
+
+Shell expansions are available, (See <<expansions#shell-expansions,`:doc expansions shell-expansions`>>)
+Piping uses the *|* register by default (See <<registers#,`:doc registers`>>)
```
Screwtapello: It wasn't immediately obvious to me what piping would use the `|` register *for*; my first guess was that the pipe output would be stored in the register for later pasting.

Perhaps "The default command comes from the `|` register by default..."

doc/pages/keys.asciidoc
```diff
@@ -640,8 +645,9 @@ the wanted object:
 == Prompt commands
 
 When pressing `:` in normal mode, Kakoune will open a prompt to enter a command.
-The following keys are recognized by this mode to help edition.
-(See <<commands#,`:doc commands`>>)
+On validation the command line is stored in the *:* register (See <<registers#,`:doc registers`>>).
```
Screwtapello: I'm not sure what "validation" means here. Is it something to do with the way the `:` turns red if your command-line has a syntax error?

If it's just "when you press `<ret>`", then we can just say that.

doc/pages/keys.asciidoc
```diff
@@ -640,8 +645,9 @@ the wanted object:
...
+
+The following keys are recognized by this mode to help edition (See <<commands#,`:doc commands`>>).
```
Screwtapello: "..to help with editing" feels more natural to me here.

Delapouite:
Thanks a lot for the proofreading @Screwtapello, I've followed your suggestions.

doc/pages/keys.asciidoc
```diff
@@ -367,6 +327,61 @@ the Shift modifier and moving will extend each selection instead.
...
+    search (select previous match)
+
+*?*::
+    search (extend to next match)
+
+*<a-?>*::
+    search (extend to previous match)
+
+*n*::
+    select next match
```
mawww: While you are at it, make it clear that `n` only select the next match from the main selection, while `/` select the next match from each selections.

doc/pages/keys.asciidoc
```diff
@@ -367,6 +327,61 @@ the Shift modifier and moving will extend each selection instead.
...
+*N*::
+    add a new selection with next match
+
+*<a-n>*::
+    select previous match
+
+*<a-N>*::
+    add a new selection with previous match
+
+***::
```
mawww: Does that display correctly as a bold `*` ?

Delapouite:
Thanks. I've added the precisions about `/` vs `n` behavior in regards of `each` vs `main` selection.

The `***` seems to be correctly rendered as a `<strong>*</strong>` on GitHub.

PR #2076, opened 2018-05-24
Fix truncated message in lint.kak

Delapouite:
Hi

The truncated messages have been bugging me for quite a long time and also been noticed by a few people on IRC.
I put the `:` in leading positions to avoid the extraneous `substr(f, 1, length(f)-1)` in the original awk script.

Here's a before / after.

The before output is missing the 2 closing parenthesis on each line `)`

lint kak


Delapouite:
Yes it should. On the screeshot, in the *before* version, the Error (line 1) is missing one paren after `(no-unused-vars` which is part of the Eslint message and `(col 7` which is added by awk.

PR #2077, opened 2018-05-24
Syntax highlighter for i3/sway

JJK96:


rc/extra/i3.kak
```diff
@@ -0,0 +1,41 @@
...
+add-highlighter shared/i3/code regex "--(release|border|whole-window|toggle|no-startup-id)" 0:attribute
+
+# color
+add-highlighter shared/i3/string regex "#[0-9a-fA-F]{6}" 0:value
+
+# attributes
+add-highlighter shared/i3/code regex "client\.(background|statusline|background|separator|statusline)" 1:attribute
+add-highlighter shared/i3/code regex "client\.(focused_inactive|focused|unfocused|urgent|inactive_workspace|urgent_workspace|focused_workspace|active_workspace|placeholder)" 1:attribute
+
+hook -group i3-highlight global WinSetOption filetype=i3 %{ add-highlighter window ref i3 }
```
mawww: Shouldn't we have a hook to detect i3 config files (.config/i3/config) and set the i3 filetype on them ?

JJK96: You are right, I forgot to move those lines from my kakrc file. 
I added them to the commits

PR #2082, opened 2018-05-25
Remove make_current_error_line in clang.kak and lint.kak

Delapouite:
Hi

This option was probably introduced by a wrong copy-paste from `make.kak`.

### DIFF
--- a/rc/base/lint.kak
+++ b/rc/base/lint.kak
@@ -15,7 +15,6 @@ define-command lint -docstring 'Parse the current buffer with a linter' %{
         printf '%s\n' "evaluate-commands -draft %{
                   edit! -fifo $dir/fifo -debug *lint-output*
                   set-option buffer filetype make
-                  set-option buffer make_current_error_line 0
                   hook -always -group fifo buffer BufCloseFifo .* %{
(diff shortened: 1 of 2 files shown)


mawww:
I dont think those were introduced by error, we set those buffer filetype to `make`, because we want to have jumping support provided by `make.kak`. At first I thought we could leave them out because the value in `global` scope would be 0, so for a new buffer we would setting those to 0 would be redundant. However for fifo buffers we do not actually recreate them between runs.

If you do `:make` twice in a row, we are going to reuse the `*make*` buffer because deleting it and recreating it would mean clients showing the `*make*` buffer would switch to another one on deletion, and would hence not show `*make*` anymore when the new `*make*` buffer would be created.

This behaviour was introduced by 4ef5c807247ec28612f5d1330b23ee56679c5c6b (2014, thats pretty entrenched), and I believe it to be correct, that consequence is that options are preserved on those buffers, so if you'd do `:make`, jump to some error (updating the `make_current_error_line` in the process), then run `:make` again, you'd want that option to switch back to 0.

I have been using `*make*` as an example, but any fifo buffer of `make` filetype needs to follow that behaviour, which is why we need those lines.

PR #2084, opened 2018-05-26
plus sign missed

co-dh:


doc/pages/highlighters.asciidoc
```diff
@@ -102,7 +102,7 @@ add-highlighter window regex //\h*(TODO:)[^\n]* 0:cyan 1:yellow,red
     This highlights all the current search matches in italic:
 
 ---------------------------------------------
```
mawww: can you fix this line length as well (add a `-` here and on the closing line line).

PR #2085, opened 2018-05-26
Display a more comprehensive status message for lint command

Delapouite:
Hi

Currently, after a linting operation a bland `linting done` message pops in the status bar.

With this PR, the message gives more useful indications: `linting results: 2 error(s) 11 warning(s)` (see bottom of screenshot below).

## Other benefits giving more control to users

- 2 options are available `lint_error_count` and `lint_warning_count`, so users can use them elsewhere. For example to permanently display their values in `modelinefmt` or do other calculations.
- if users don't remember how many errors/warnings were displayed in the status, they can run the `lint-show-counters` again.
- if users don't want to be annoyed with the status message, they can alias `lint-show-counters` as `nop`.
- users can `-override` `lint-show-counters` for their needs (if they don't like the default colors or the formatting).

lint-status

### DIFF
--- a/rc/base/lint.kak
+++ b/rc/base/lint.kak
@@ -5,6 +5,8 @@ The output returned by this command is expected to comply with the following for
 
 declare-option -hidden line-specs lint_flags
 declare-option -hidden range-specs lint_errors
+declare-option -hidden int lint_error_count
+declare-option -hidden int lint_warning_count
 
 define-command lint -docstring 'Parse the current buffer with a linter' %{
     %sh{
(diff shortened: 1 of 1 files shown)


rc/base/lint.kak
```diff
@@ -70,6 +80,10 @@ define-command -hidden lint-show %{
         fi
     } }
 
+define-command -hidden lint-show-counters %{
+    echo -markup linting results:{red} %opt{lint_error_count} error(s){yellow} %opt{lint_warning_count} warning(s)
```
lenormf: I think the `Error` and `Prompt faces should be used here, instead of hard-coded colors. The result should be same , with the default theme.

mawww: Yeah, I think that would make sense.

Delapouite: I used explicit `{red}` and `{yellow}` colors to mirror the same hardcoded value of the `█` blocks in the gutter:
```
flags = flags ":" $2 "|{red}█"
flags = flags ":" $2 "|{yellow}█"
```

In default.kak:
```
face global Information black,yellow
face global Error black,red
```
these two faces use yellow and red as background, but `█` needs it as foreground.

mawww: Right, yeah, thats reasonable.

PR #2094, opened 2018-05-31
tmux-send-text can take 1 parameter now.

co-dh:
This is useful when send any text that is not even exists in current buffer. 
For example send string to reload current script: \l<space><c-r>%<ret>,
which can be then used to bind with convenient hotkey.

### DIFF
--- a/rc/extra/tmux-repl.kak
+++ b/rc/extra/tmux-repl.kak
@@ -50,9 +50,15 @@ define-command tmux-repl-window -params 0..1 -command-completion -docstring "Cre
     tmux-repl-impl 'new-window' %arg{@}
 }
 
-define-command -hidden tmux-send-text -docstring "Send the selected text to the repl pane" %{
+define-command -hidden tmux-send-text -params 0..1 -docstring "Send the selected text to the repl pane if no parameter.
+  Otherwise send parameter with new line" %{
     nop %sh{
-        tmux set-buffer -b kak_selection "${kak_selection}"
+        if [ $# -eq 0 ]; then 
+            tmux set-buffer -b kak_selection "${kak_selection}"
+        else
+            tmux set-buffer -b kak_selection "$1
+"
+        fi
         kak_orig_window=$(tmux display-message -p '#I')
         kak_orig_pane=$(tmux display-message -p '#P')
         tmux select-window -t:$(tmux show-buffer -b kak_repl_window)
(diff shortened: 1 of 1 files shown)


rc/extra/tmux-repl.kak
```diff
@@ -50,9 +50,15 @@ define-command tmux-repl-window -params 0..1 -command-completion -docstring "Cre
     tmux-repl-impl 'new-window' %arg{@}
 }
 
-define-command -hidden tmux-send-text -docstring "Send the selected text to the repl pane" %{
+define-command -hidden tmux-send-text -params 0..1 -docstring "Send the selected text to the repl pane if no parameter.
```
lenormf: Here is a more idiomatic help message:

```
tmux-send-text [text]: Send text to the REPL pane
If no text is passed, then the selection is used
```

rc/extra/tmux-repl.kak
```diff
@@ -50,9 +50,15 @@ define-command tmux-repl-window -params 0..1 -command-completion -docstring "Cre
...
+  Otherwise send parameter with new line" %{
     nop %sh{
-        tmux set-buffer -b kak_selection "${kak_selection}"
+        if [ $# -eq 0 ]; then 
+            tmux set-buffer -b kak_selection "${kak_selection}"
+        else
+            tmux set-buffer -b kak_selection "$1
```
lenormf: There's a trailing newline before the ending quote, is that on purpose?

co-dh: Yes. For example, you can select a word and send with new line, which will print its value by most interpreters.
Do you know a better way to add new line?

rc/extra/tmux-repl.kak
co-dh: Thanks. Will modify later.

rc/extra/tmux-repl.kak
lenormf: We should not assume that users only want to send whole commands to the REPL prompt, so I think the newline character should be removed. Users could declare a command that sends text followed by a newline, if they want that behavior.

Also, `$'\n'`.

co-dh: Agree. Consider it done.

mawww: `$'\n'` is not posix :(

lenormf: In that case:

```bash
readonly NEWLINE=$(echo)
```

Issue #1273, opened 2017-03-09
Show keys in the status line while recording macros.


Delapouite:
Hi

While crafting macros you often have to cross your fingers, hoping everything will work as expected after typing a long sequence of keys.
I think the main issue of this syndrome is that during the recording, we have no trace of what has been done so far.

Kakoune sets live-feedback as one of its core value, but cruelly left us partially in the dark for macros.
Currently, after pressing `Q`, a mention *[ recording (<reg>) ]* appears in the status line.

A nice improvement would be to also display (on the right?) the keys sequence of the macros while it's being recorded.
This way you can pause your flow of thoughts midway during macro recording, and visually checked that the keys you recorded so far are good.

I guess this may be implemented right now by showing the content of the chosen register in a custom `modelinefmt` https://github.com/mawww/kakoune/pull/1267
But it would be great to have this improved interactive behavior by default.

danr:
I tried to adding a hook like so:

    hook -group test buffer NormalIdle .* %{echo %reg{@}}

but alas, the `@` register is not set until the macro is finished! So there might be an easy fix here if the register is updated after each key-press.

lenormf:
#1277 might provide a beginning of solution, if you make it so a menu containing the contents of the current register pops up in macro recording mode. Hardly doable as of today though.

PR #2104, opened 2018-06-03
Display recorded keys during macro recording

Delapouite:
Hi

To close https://github.com/mawww/kakoune/issues/1273


src/input_handler.hh
```diff
@@ -86,6 +86,7 @@ public:
     bool is_recording() const;
     void stop_recording();
     char recording_reg() const { return m_recording_reg; }
+    StringView recorded_keys() const { return m_recorded_keys; }
```
lenormf: I think the implementation of this function is missing from the commit.

PR #2106, opened 2018-06-04
Improve unmap command key-completer

Delapouite:
Hi

Here's a small enhancement on the `unmap` command completer.
For example typing in the prompt `unmap global user ` will suggest all the keys registered on the `user` mode.

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -1131,6 +1131,8 @@ KeymapMode parse_keymap_mode(StringView str, const KeymapManager::UserModeList& 
     return (KeymapMode)(std::distance(user_modes.begin(), it) + offset);
 }
 
+static constexpr auto modes = { "normal", "insert", "menu", "prompt", "goto", "view", "user", "object" };
+
 const CommandDesc debug_cmd = {
     "debug",
     nullptr,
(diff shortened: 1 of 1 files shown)


src/commands.cc
```diff
@@ -1448,22 +1448,32 @@ const CommandDesc declare_option_cmd = {
...
                      complete(params[0], pos_in_token, scopes) };
-
         if (token_to_complete == 1)
         {
-            static constexpr auto modes = { "normal", "insert", "menu", "prompt", "goto", "view", "user", "object" };
             auto& user_modes = get_scope(params[0], context).keymaps().user_modes();
             return { 0_byte, params[1].length(),
                      complete(params[1], pos_in_token, concatenated(modes, user_modes) | gather<Vector<String>>()) };
         }
+        if (unmap && token_to_complete == 2)
```
lenormf: `else if` here?

mawww: I dont mind much about the `else if`, as the previous if returns from the function, however Kakoune code uses the alternate boolean operator syntax (and, or, not) rather than the "classic" one (&&, ||, !), please fix that.

Delapouite:
Only 1 of the 4 registered compilers on Travis is failing

PR #2107, opened 2018-06-04
Add support for regex flag to toggle dot-matches-newline

occivink:
This used to be supported by boost regex with the `s` modifier as well. 

Using a character class for this seems like it could be expensive, maybe it's worth introducing an `Op` for this, or simply reusing the same `CharacterClass`?

### DIFF
--- a/src/regex_impl.cc
+++ b/src/regex_impl.cc
@@ -201,12 +201,21 @@ private:
         auto it = m_pos.base();
         if (m_regex.end() - it >= 4 and *it++ == '(' and *it++ == '?')
         {
-            auto m = *it++;
-            if ((m != 'i' and m != 'I') or *it++ != ')')
-                return false;
-            m_ignore_case = (m == 'i');
-            m_pos = Iterator{it, m_regex};
-            return true;
+            while (true)
+            {
+                auto m = *it++;
+                if (m == 'i' or m == 'I')
+                    m_ignore_case = (m == 'i');
+                else if (m == 's' or m == 'S')
+                    m_dot_maches_newline = (m == 's');
+                else if (m == ')')
+                {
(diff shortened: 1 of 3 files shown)


Delapouite:
The timing is perfect since this `s` flag is also part of ECMAScript 2018 which is due this month: https://github.com/tc39/proposal-regexp-dotall-flag

src/regex_impl.cc
```diff
@@ -201,12 +201,21 @@ struct RegexParser
...
-            if ((m != 'i' and m != 'I') or *it++ != ')')
-                return false;
-            m_ignore_case = (m == 'i');
-            m_pos = Iterator{it, m_regex};
-            return true;
+            while (true)
+            {
+                auto m = *it++;
+                if (m == 'i' or m == 'I')
+                    m_ignore_case = (m == 'i');
```
mawww: Lets move those to a single m_flags (using with_bit_ops) to store all the enabled flags, instead of a boolean for each.

src/regex_impl.cc
```diff
@@ -274,7 +283,19 @@ struct RegexParser
...
         switch (const Codepoint cp = *m_pos)
         {
-            case '.': ++m_pos; return new_node(ParsedRegex::AnyChar);
+            case '.':
+                ++m_pos;
+                if (m_dot_maches_newline)
+                    return new_node(ParsedRegex::AnyChar);
+                else
+                {
+                    CharacterClass c;
```
mawww: I think this could warrant its own vm op, or at the very least, the character class should be shared, here we'll get a character class for each '.' appearing in the regex.

occivink:
I'll add tests at a later point and squash everything if/when you're happy with the changes.

PR #2113, opened 2018-06-08
[WIP] ghcid.kak: Simple ghcid integration

joamaki:
Add a simple integration for ghcid that streams ghcid errors
to a buffer and allows jumping between errors.

### DIFF
--- a/rc/extra/ghcid.kak
+++ b/rc/extra/ghcid.kak
@@ -24,8 +24,11 @@ declare-option -docstring "name of the client in which to jump" \
 declare-option -docstring "name of the client in which to process output" \
   str ghcidtoolsclient
 
+declare-option -docstring "shell command to use to execute ghcid" \
+  str ghcidcmd ghcid
+
 declare-option -docstring "ghcid error regexp" \
-  str ghcid_error_pattern "[^:\n]+:\d+:\d+: error:$"
+  str ghcid_error_pattern "[^:\n]+:\d+:\d+(-\d+)?: error:$"
 
 declare-option -hidden int ghcid_current_error_line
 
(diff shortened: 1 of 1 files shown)


joamaki:
Don't merge. This still has issues with GHC 8.4.x and I'd like to add some range highlighters.

rc/extra/ghcid.kak
```diff
@@ -0,0 +1,129 @@
...
+  -docstring %{ghcid-start [<arguments>]: ghcid utility wrapper
+All optional arguments forwarded to ghcid} \
+  ghcid-start %{ %sh{
+    workdir=$(mktemp -d "${TMPDIR:-/tmp}"/kak-ghcid.XXXXXXXXX)
+    output=${workdir}/output
+    errors=${workdir}/errors
+    fifo=${workdir}/fifo
+    exec 2> ${errors}
+    touch ${output}
+    mkfifo ${fifo}
```
lenormf: Ideally you would put double quote around these three expansions, as the directory in `$TMPDIR` could contain spaces.

rc/extra/ghcid.kak
```diff
@@ -0,0 +1,129 @@
...
+    cmd="ghcid -S -n -o ${output} $@"
```
lenormf: Quotes around the expansion.

rc/extra/ghcid.kak
```diff
@@ -0,0 +1,129 @@
...
+    echo "starting: ${cmd}" > ${output}
```
lenormf: Using `echo` with a string that has arbitrary expansions is not portable, prefer using `printf`.

rc/extra/ghcid.kak
```diff
@@ -0,0 +1,129 @@
...
+    $cmd > /dev/null < /dev/null &
```
lenormf: Prefer using `eval` rather than expanding the command inline.

rc/extra/ghcid.kak
```diff
@@ -0,0 +1,129 @@
...
+    pid=$!
+    (tail -qf ${errors} ${output} > ${fifo} 2>&1) > /dev/null 2>&1 < /dev/null &
```
lenormf: The `-q` flag is not POSIX, expansions should be enclosed in double quotes, and you could use braces `{`/`}` instead of parenthesis `(`/`)` to avoid spawning an entire process.

rc/extra/ghcid.kak
```diff
@@ -0,0 +1,129 @@
...
+    pid2=$!
+
+    printf %s\\n "evaluate-commands -try-client '$kak_opt_ghcidtoolsclient' %{
+      edit! -fifo ${fifo} -scroll *ghcid*
```
lenormf: Quotes around the expansion.

rc/extra/ghcid.kak
```diff
@@ -0,0 +1,129 @@
...
+      set-option buffer filetype ghcid
+      set-option buffer ghcid_current_error_line 0
+      hook -group ghcid buffer BufCloseFifo .* %{
+        nop %sh{ kill ${pid} ${pid2}; rm -r ${output} }
```
lenormf: Quotes around the expansion (the `${output}` one, PIDs are fine).

rc/extra/ghcid.kak
```diff
@@ -0,0 +1,129 @@
...
+        remove-hooks buffer ghcid
+        remove-hooks global ghcid
+      }
+      # hook to writes to .(l)hs files and clear the buffer
+      hook -group ghcid global BufWritePost .*\.l?hs %{
+        evaluate-commands -try-client %opt{ghcidtoolsclient} %{
+          buffer '*ghcid*'
```
lenormf: You should probable use the `edit` command here, in case that buffer doesn't exist when the hook is triggered.

rc/extra/ghcid.kak
```diff
@@ -0,0 +1,129 @@
...
+          execute-keys '%d'
```
lenormf: A `try`/`catch` would be good here, in case the buffer exists but is empty.

rc/extra/ghcid.kak
```diff
@@ -0,0 +1,129 @@
...
+            # switch to the file and show the error
+            evaluate-commands -try-client %opt{ghcidclient} "edit -existing %reg{1} %reg{2} %reg{3}; echo -markup {Information}%reg{4}; try %{ focus }"
+        }
+    }
+}
+
+define-command ghcid-next-error -docstring 'Jump to next ghcid error' %{
+    evaluate-commands -try-client %opt{ghcidclient} %{
+        try %{
+          buffer '*ghcid*'
```
lenormf: `edit` instead of `buffer`.

rc/extra/ghcid.kak
```diff
@@ -0,0 +1,129 @@
...
+          execute-keys "ga"
+          echo -markup "{Information}No more errors"
+        }
+    }
+    try %{ evaluate-commands -client %opt{ghcidtoolsclient} %{ execute-keys %opt{ghcid_current_error_line}g vc } }
+}
+
+define-command ghcid-previous-error -docstring 'Jump to previous ghcid error' %{
+    evaluate-commands -try-client %opt{ghcidclient} %{
+        buffer '*ghcid*'
```
lenormf: `edit` instead of `buffer`.

lenormf:
If the PR is not ready for merging, you could add a `[WIP]` tag to the title.

rc/extra/ghcid.kak
joamaki: this was almost verbatim from 'make.kak'. should that be changed as well?

rc/extra/ghcid.kak
joamaki: Ah that's unfortunate as without "-q" the output is quite noisy. That option is supported by both GNU and BSD variants so IMO it would be fine here.

rc/extra/ghcid.kak
joamaki: if the buffer doesn't exist we shouldn't do anything. This should get unhooked when the buffer is destroyed, but I'll wrap this in a try anyway.

rc/extra/ghcid.kak
joamaki: how would I do edit here when the filename is only resolved within the ghcid function?

rc/extra/ghcid.kak
hurryabit: Also, `-n` expects an integer argument in ghcid 0.7.

rc/extra/ghcid.kak
lenormf: I see that `-q` is fairly largely supported across the implementations I've checked, despite not being POSIX compliant.

If there's no other way to implement that bit, I suppose it's fine to leave the flag until somebody complains about it?

rc/extra/ghcid.kak
lenormf: Yes, use a `try`/`catch` then.

rc/extra/ghcid.kak
joamaki: I think here a try + buffer makes more sense. I don't think it should do anything if the buffer doesn't exist (as there wouldn't be a next error anyway...).

rc/extra/ghcid.kak
joamaki: I'm having some issues with having tail being executed from the main shell instead of subshell. Kakoune seems to get stuck no matter how stdin/stdout/stderr is being redirected. Tried also disowning the tail process. Suggestions welcome on what might be going wrong.

danr:
Relevant: https://github.com/danr/kakoune-ghci-bridge

joamaki:
@danr ah, you beat me to it. was about to try doing something similar. I'm now a bit unsure on how much value there is with the simpler ghcid wrapper, or whether it would make more sense to polish `kakoune-ghci-bridge`...

PR #2131, opened 2018-06-19
Rebase/openbsd

dahlbaek:
Rebase of the pull request by Superpat #2005 with an alternative to hard coding the binary path. Instead, the path is passed in as a macro, by using the preprocessor option `-DKAK_BIN_PATH=$(bindir)/kak`.

### DIFF
--- a/src/file.cc
+++ b/src/file.cc
@@ -573,7 +573,7 @@ String get_kak_binary_path()
     buffer[res] = '\0';
     return buffer;
 #elif defined(__OpenBSD__)
-    return "/usr/local/bin/";
+    return KAK_BIN_PATH;
 #else
 # error "finding executable path is not implemented on this platform"
 #endif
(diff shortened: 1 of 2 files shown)


src/file.cc
```diff
@@ -572,6 +572,10 @@ String get_kak_binary_path()
     kak_assert(res != -1);
     buffer[res] = '\0';
     return buffer;
+#elif defined(__OpenBSD__)
+    size_t res = strlcpy(buffer, KAK_BIN_PATH, 2048);
```
mawww: return KAK_BIN_PATH should be enough, no need to copy it into the buffer.

dahlbaek:
Squashed the commits and changed to simply returning `KAK_BIN_PATH`. As far as I can tell, the failed test is not related to anything in the pull request...

Superpat:
In my limited use, I've noticed problems with multiple selections, havent really had the time to dive into the problem though.

PR #2132, opened 2018-06-19
 Fix remaining static_words and spell.kak for the new list syntax

shachaf:


shachaf:
Note that jedi.kak is still broken. I'm not quite sure how it was supposed to work, so someone who uses it should probably fix it.

mawww:
cherry-picked

PR #2135, opened 2018-06-20
Fix clang.kak and part of lint.kak

shachaf:
* Make clang.kak compatible with POSIX `[`.
* Make lint.kak dump range/line-specs correctly. It still doesn't read them
  correctly -- that'll be easier after the upcoming $kak_ changes for lists.


shachaf:
While testing this I noticed that lint.kak has a race condition -- sometimes the linter runs before kak has finished writing the buffer -- which should probably be addressed elsewhere.

mawww:
cherry-picked

PR #2139, opened 2018-06-21
Explicitly discard unused variables in selectors.cc

dahlbaek:
Using `clang++` on OpenBSD, the compiler emits warnings related to `selectors.cc` (see [selectors.log](https://github.com/mawww/kakoune/files/2122684/selectors.log)). These warnings disappear if the unused variables are explicitly recast to void.

Casting to void is suggested in [https://stackoverflow.com/a/34289000](https://stackoverflow.com/a/34289000) and [https://stackoverflow.com/a/17711748](https://stackoverflow.com/a/17711748)

### DIFF
--- a/src/selectors.cc
+++ b/src/selectors.cc
@@ -289,7 +289,7 @@ find_opening(Iterator pos, const Container& container,
     {
         if (nestable)
         {
-            for (auto m : RegexIt{match[0].second, pos, container.begin(), container.end(), closing})
+            for (auto m [[gnu::unused]] : RegexIt{match[0].second, pos, container.begin(), container.end(), closing})
                 ++level;
         }
 
@@ -317,7 +317,7 @@ find_closing(Iterator pos, const Container& container,
     {
         if (nestable)
         {
-            for (auto m : RegexIt{pos, match[0].first, container.begin(), container.end(), opening})
+            for (auto m [[gnu::unused]] : RegexIt{pos, match[0].first, container.begin(), container.end(), opening})
                 ++level;
         }
 


src/selectors.cc
```diff
@@ -290,7 +290,10 @@ find_opening(Iterator pos, const Container& container,
         if (nestable)
         {
             for (auto m : RegexIt{match[0].second, pos, container.begin(), container.end(), closing})
```
mawww: I'd rather disable them with an attribute, can you use `for (auto m [[gnu::unused]] : ...`

dahlbaek:
@mawww I was not aware that force pushing would also make the review unavailable(?). I made the change you requested. Perhaps the attribute [[maybe_unused]] could be of interest, but seeing as that is from the C++17 standard, you may not want to use it.

mawww:
We will migrate to [[maybe_unused]] as soon as we can move to C++17, the current guideline has been "migrate to a C++ standard once Debian Stable has a compiler for it", but I am pretty eager to use it, as its fun to use new C++ features (and it can help cleanup part of the code).

lenormf:
As this is a non [standard attribute](https://en.cppreference.com/w/cpp/language/attributes), it would be better to refrain on sprinkling compiler-specific instructions and use the portable `(void)m;`. It's ugly, but I think it doesn't make the code more complicated to read as this would rest on its own line.

dahlbaek:
I can change it back to `(void)m;` if you like. Perhaps it is easier to go for `[[gnu::unused]]` until you can switch to C++17, since a simple `grep -r '\[\[gnu::' *` will expose any use of such compiler-specific attributes. Running it just now revealed to me that a few other compiler specific attributes are also used in the source code.

mawww:
The C++ standard committee clarified that unknown attributes should be ignored, gcc and clang both support `[[gnu::unused]]` and any other conforming compiler will ignore it. Moreover, `(void)m;` is not a standard mandated way to disable warnings, its just a hack some compiler support.

lenormf:
The point is specifically to prevent warnings during compilation, which the void cast won't emit any with `clang` and `gcc`.

dahlbaek:
I guess there are two options.

1) `(void)m;`

   Pros:
   - Works on `gcc` and `clang`

   Cons:
   - Not standard mandated.
   - Undocumented in terms of silencing warnings (?).
   - If this is used, it will probably be forgotten and never replaced.

2) `[[gnu::unused]]`

   Pros:
   - Works on `gcc` and `clang`
   - Well documented in terms of silencing warnings.
   - If this is used, it is easy to find and replace later on (`grep -r '\[\[gnu::' *`).

   Cons:
   - Not standard mandated.

mawww:
@lenormf I dont really get your point, `[[gnu::unused]]` disables those warnings, and is much more explicit (IMHO) than casting to void. Neither are guaranteed by the standard to disable unused warnings (a compiler might recognize that casting to void is a no-op and still emit the warning), but the attribute matches whats standard mandated from C++17 (we'll need to update it to `[[maybe_unused]]`).

PR #2140, opened 2018-06-21
 Run git in the directory of the buffer being edited

shachaf:


mawww:
I believe this should be based of master.

I am not sure we want this behaviour, it means you cannot use "<c-r>%" in git command line anymore. This should probably not be a git specific change, if you generally want commands to run from the directory of the current file, an auto cd hook might make more sense. Although it comes with its own set of problems.

I am still having trouble understanding why running a Kakoune session from a project root (and having one session per project) does not work well enough.

shachaf:
I didn't think this would be controversial -- maybe this should move to an issue.

If I have a Kakoune session running from the repository root then of course this isn't a problem. But sometimes I edit a file from my home directory to look at it, and then I can't git show-diff and git-log-lines and so on. Getting information about the current file is my primary use case -- if I want to commit or anything like that I do it from the command line.

maximbaz:
What I would prefer is kak to auto-jump in the project root of the current buffer, I have this configuration in nvim for a long time and I like it a lot. That way, I e.g. can run `nvim ~/.config/kak/kakrc` from my home directory, and nvim detects that `~/.config/kak/kakrc` is in fact a symlink to `~/.dotfiles/.config/kak/kakrc`, and in turn `~/.dotfiles` is a git repo - so nvim sets `cwd` to `~/.dotfiles`. Very convenient for git operations, `:e`, `:grep` and many other cases.

alexherbo2:
I prefer Git not to change to the root of the current file but the given files (which defaults to the current file), so that `<c-r>%` and `git add <c-r>.` from Git status buffer do not break.

PR #2149, opened 2018-06-23
rc: Highlight embedded documentation

### LINKED ISSUE #1944: Implement a builtin `documentation` face
Hi,

Some frameworks and languages allow writing embedded documentation, generally using comment blocks e.g. doxygen strings.

It'd be nice to highlight those differently than comments to improve readability, using a builtin `documentation` face that the builtin colorschemes would initialize.

Thanks.

lenormf:
This commit adds a `documentation` face to the builtin themes, used
to highlight common documentation syntaxes:

	/**
	 * JavaDoc
	 */

	/*!
	 * QtDoc
	 */

	/// Inline documentation

	## Inline documentation

The face is only an alias to the `comment` one for now.

Closes #1944

### DIFF
--- a/colors/base16.kak
+++ b/colors/base16.kak
@@ -28,6 +28,7 @@ evaluate-commands %sh{
         face global operator ${cyan_light}
         face global attribute ${orange_dark}
         face global comment ${grey_dark}
+        face global documentation comment
         face global meta ${orange_light}
         face global builtin default+b
     "
--- a/colors/default.kak
+++ b/colors/default.kak
@@ -11,6 +11,7 @@ face global keyword blue
 face global operator yellow
 face global attribute green
 face global comment cyan
+face global documentation comment
 face global meta magenta
 face global builtin default+b
 
--- a/colors/desertex.kak
+++ b/colors/desertex.kak
@@ -1,15 +1,16 @@
 # desertex theme
 
 # Code
-face global value      rgb:fa8072
-face global type       rgb:dfdfbf
-face global identifier rgb:87ceeb
-face global string     rgb:fa8072
-face global error      rgb:c3bf9f+b
-face global keyword    rgb:eedc82
-face global operator   rgb:87ceeb
-face global attribute  rgb:eedc82
-face global comment    rgb:7ccd7c+i
+face global value         rgb:fa8072
+face global type          rgb:dfdfbf
+face global identifier    rgb:87ceeb
+face global string        rgb:fa8072
+face global error         rgb:c3bf9f+b
+face global keyword       rgb:eedc82
+face global operator      rgb:87ceeb
+face global attribute     rgb:eedc82
+face global comment       rgb:7ccd7c+i
+face global documentation comment
 
 # #include <...>
 face global meta rgb:ee799f
--- a/colors/github.kak
+++ b/colors/github.kak
@@ -14,6 +14,7 @@ face global keyword rgb:A71D5D+b
 face global operator yellow
 face global attribute rgb:A71D5D
 face global comment rgb:AAAAAA
+face global documentation comment
 face global meta rgb:183691
 face global builtin default+b
 
--- a/colors/greyscale.kak
+++ b/colors/greyscale.kak
@@ -24,6 +24,7 @@ evaluate-commands %sh{
     set-face global builtin ${grey}+b
     set-face global module ${grey_dark_1}
     set-face global comment ${grey}+i
+    set-face global documentation comment
     set-face global function Default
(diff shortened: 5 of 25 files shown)


Screwtapello:
`///` and `//!` are also docstring prefixes for Rust; could you add them to `base/rust.kak` too?

SolitudeSF:
`nim` supports `##` and `##[` `]##` pair for multiline doc comments.

lenormf:
Quick note: I thought of implementing this primarily because of C/C++, which use comments for embedded documentation. However, `python` has docstrings, which are highlighted as strings currently, but would be highlighted as comments with this PR - the `documentation` face is an alias to the `comment` face.

Screwtapello:
Python is a weird one; `python.kak` refers to both `"""triple-double-quoted strings"""` and `'''triple-single-quoted strings'''` as "docstrings", but that's not strictly true: they're both just multi-line strings. A Python docstring is any string literal (single-, double-, raw-, triple-quoted or any combination) that appears immediately inside a module, class or function.

I don't think Kakoune's syntax-highlighting is smart enough to distinguish Python docstrings from other string literals, so for the purposes of this PR it's probably best to rename that syntax region from "docstring" to "multiline" or something, and leave Python without any syntax feature mapped to the "documentation" face at all (well, except for `##` comments, I guess).

lenormf:
The official use for docstrings is to document parts of the code, and they are not a way to store multi-line text, even though that's how they're being used generally.

So I don't think they should have been highlighted as strings in the first place, although I understand why that happened (misnomer), and not highlighting them as documentation when there's a face for that purpose would feel like neglecting the standard.

Also users who actually use docstrings only for documentation purposes might be confused by the fact the editor supports highlighting documentation, but still colors them as strings.

Screwtapello:
According to [the Python language reference](https://docs.python.org/3.6/reference/lexical_analysis.html#string-and-bytes-literals), the official term for triple-quoted strings is "longstring"; "docstring" doesn't appear anywhere in the "String and Byte Literals" section. Triple-quoted strings are definitely used for many things besides documentation, like queries:

```python
cursor.execute("""
    insert into users (
        name,
        pwhash
    ) values (?, ?)
    """,
    name,
    pwhash,
)
```

Meanwhile, PEP257 "Docstring Conventions" [says one-line docstrings should still use a triple-quoted string](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings) — if docstrings could *only* be triple-quoted strings, they wouldn't need to say anything.

lenormf:
They are used as regular multi-line strings, but they are supposed to be containing documentation for the code beneath, so highlighting them as strings is incorrect.

Screwtapello:
I confess, I haven't encountered the rule "triple-quoted strings may only be used for docstrings" before. Can you point me to some documentation or a style-guide or something that spells that out? I've had a look around myself (see the Python Language Reference and PEP257 linked previously and also [Google's Python style guide](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#310-strings), and haven't found a source.

dahlbaek:
@lenormf :
> Docstrings can only be declared with triple quotes, and editors don't know the difference between a docstring and a multi-line string, which leads to code being highlighted wrongly. Therefore Python developers conventionally don't use triple quoted strings unless really necessary, even though you can technically use triple-quoted string everywhere.


I do not think there is anything in the [python string documentation](https://docs.python.org/3/library/stdtypes.html#text-sequence-type-str) which suggests that triple double quoted strings should be used only for docstrings. According to [PEP 257](https://www.python.org/dev/peps/pep-0257/):

> A docstring is a string literal that occurs as the first statement in a module, function, class, or method definition. Such a docstring becomes the __doc__ special attribute of that object.

We can easily verify that docstrings do not need to be declared using triple quotes:

```python
def f():
    'This does nothing.'
    pass

print(f.__doc__)
```

outputs

```
This does nothing.
```

PEP 257 does go on to say

> For consistency, always use """triple double quotes""" around docstrings.

i.e. docstrings should be enclosed in triple double quotes. It does not say that anything enclosed in triple double quotes is — or should be — a docstring.

Further comments:

- According to an accepted (and highly upvoted) answer on stackoverflow [triple quotes are the pythonic way to create a long multi-line string](https://stackoverflow.com/a/10660443/7599943).

- For a typical use of multiline strings outside of docstrings see the [psycopg2 documentation](http://initd.org/psycopg/docs/usage.html).

lenormf:
Rebased on `master`. Highlighting of all docstrings as documentation was removed to avoid false positives.

PR #2153, opened 2018-06-23
Improve coloring of wrapped line numbers in gruvbox

maximbaz:
The default coloring of wrapped line numbers in gruvbox is simply too bright:

image

I used the same pattern as in solarized-dark and made these numbers blend in the background:

image

In my mind, this is a much better default value to have.

### DIFF
--- a/colors/gruvbox.kak
+++ b/colors/gruvbox.kak
@@ -57,6 +57,7 @@
         face global SecondaryCursorEol ${bg},${fg4}
         face global LineNumbers        ${bg4}
         face global LineNumberCursor   ${yellow},${bg1}
+        face global LineNumbersWrapped ${bg}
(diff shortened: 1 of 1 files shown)


maximbaz:
I disagree that it "breaks functionality" and I know I can change things locally, but I argue that this is a much better default to have. Not only the current default is way too bright, but hiding the numbers is how `solarized-dark` and `solarized-light` themes behave (the most recently updated themes in this repo), and this is how the `vim-gruvbox` theme behaves (the grandfather of this theme in this repo, according to git history).

maximbaz:
This is not worth arguing about, I'm happy as long as we get rid of bright white color and these numbers don't stand out. I set it to `${bg1}`, no objections about getting this merged?

image

mawww:
Good compromise, thanks a lot.

PR #2155, opened 2018-06-23
Protect against buffer overflows and unchecked string truncations

dahlbaek:
I went through calls to sprintf and readlink and did my best to rewrite them in ways protected against buffer overflows and unchecked string truncations.

- I replaced calls to sprintf by calls to snprintf and added a check to ensure that the operation succeeded and did not truncate the string. I followed the [secure idiom](https://www.freebsd.org/cgi/man.cgi?snprintf(3)) found on the FreeBSD man pages:

      snprintf(buffer, sizeof(buffer), "%s", string);

- I made sure to avoid that the terminating null character in calls to readlink would overflow the buffer, following [this page](https://wiki.sei.cmu.edu/confluence/display/c/POS30-C.+Use+the+readlink%28%29+function+properly). See also [https://stackoverflow.com/a/5525712/7599943](https://stackoverflow.com/a/5525712/7599943).

### DIFF
--- a/src/color.cc
+++ b/src/color.cc
@@ -65,7 +65,8 @@ String to_string(Color color)
     if (color.color == Color::RGB)
     {
         char buffer[11];
-        sprintf(buffer, "rgb:%02x%02x%02x", color.r, color.g, color.b);
+        int res = snprintf(buffer, sizeof(buffer), "rgb:%02x%02x%02x", color.r, color.g, color.b);
+        kak_assert(0 <= res && res < sizeof(buffer));
         return buffer;
     }
     else
--- a/src/file.cc
+++ b/src/file.cc
@@ -545,19 +545,21 @@ timespec get_fs_timestamp(StringView filename)
 
 String get_kak_binary_path()
 {
-    char buffer[2048];
 #if defined(__linux__) or defined(__CYGWIN__)
-    ssize_t res = readlink("/proc/self/exe", buffer, 2048);
+    char buffer[2048];
+    ssize_t res = readlink("/proc/self/exe", buffer, sizeof(buffer) - 1);
     kak_assert(res != -1);
     buffer[res] = '\0';
     return buffer;
 #elif defined(__FreeBSD__)
+    char buffer[2048];
     int mib[] = {CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME, -1};
     size_t res = sizeof(buffer);
     sysctl(mib, 4, buffer, &res, NULL, 0);
     return buffer;
 #elif defined(__APPLE__)
-    uint32_t bufsize = 2048;
+    char buffer[2048];
+    uint32_t bufsize = sizeof(buffer);
     _NSGetExecutablePath(buffer, &bufsize);
     char* canonical_path = realpath(buffer, nullptr);
     String path = canonical_path;
@@ -571,7 +573,8 @@ String get_kak_binary_path()
     BPath path(&info.ref);
     return path.Path();
 #elif defined(__DragonFly__)
-    ssize_t res = readlink("/proc/curproc/file", buffer, 2048);
+    char buffer[2048];
+    ssize_t res = readlink("/proc/curproc/file", buffer, sizeof(buffer) - 1);
     kak_assert(res != -1);
     buffer[res] = '\0';
     return buffer;
--- a/src/json_ui.cc
+++ b/src/json_ui.cc
@@ -52,7 +52,10 @@ String to_json(StringView str)
 
         char buf[7] = {'\\', *next, 0};
         if (*next >= 0 and *next <= 0x1F)
-            sprintf(buf, "\\u%04x", *next);
+        {
+            int result = snprintf(buf, sizeof(buf), "\\u%04x", *next);
+            kak_assert(0 <= result && result < sizeof(buf));
+        }
 
         res += buf;
         it = next+1;
@@ -66,7 +69,8 @@ String to_json(Color color)
     if (color.color == Kakoune::Color::RGB)
     {
         char buffer[10];
-        sprintf(buffer, R"("#%02x%02x%02x")", color.r, color.g, color.b);
+        int res = snprintf(buffer, sizeof(buffer), R"("#%02x%02x%02x")", color.r, color.g, color.b);
+        kak_assert(0 <= res && res < sizeof(buffer));
         return buffer;
     }
     return to_json(to_string(color));
--- a/src/string_utils.cc
+++ b/src/string_utils.cc
@@ -121,49 +121,63 @@ int str_to_int(StringView str)
 InplaceString<15> to_string(int val)
 {
     InplaceString<15> res;
-    res.m_length = sprintf(res.m_data, "%i", val);
+    int result = snprintf(res.m_data, res.m_length + 1, "%i", val);
+    kak_assert(0 <= result && result <= res.m_length);
+    res.m_length = result;
     return res;
 }
 
 InplaceString<15> to_string(unsigned val)
 {
     InplaceString<15> res;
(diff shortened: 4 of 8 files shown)


src/color.cc
```diff
@@ -65,7 +65,8 @@ String to_string(Color color)
     if (color.color == Color::RGB)
     {
         char buffer[11];
-        sprintf(buffer, "rgb:%02x%02x%02x", color.r, color.g, color.b);
```
lenormf: This call cannot overflow the buffer?

dahlbaek: You are probably right. But the call
```
snprintf(buffer, sizeof(buffer), "rgb:%02x%02x%02x", color.r, color.g, color.b);
```
is guaranteed to not overflow the buffer, regardless of code changes anywhere else. Isn't that prefereable?

src/file.cc
```diff
@@ -542,19 +542,21 @@ timespec get_fs_timestamp(StringView filename)
 
 String get_kak_binary_path()
 {
-    char buffer[2048];
```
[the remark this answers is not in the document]
dahlbaek: The buffer is not re-declared in the HAIKU or OpenBSD scope (where it is not used).

src/json_ui.cc
```diff
@@ -52,7 +52,10 @@ String to_json(StringView str)
 
         char buf[7] = {'\\', *next, 0};
         if (*next >= 0 and *next <= 0x1F)
-            sprintf(buf, "\\u%04x", *next);
```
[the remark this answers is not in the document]
dahlbaek: My understanding is that the c/c++ standards do not offer any guarantees regarding maximum bit size of the integer types (only relative size between the different types, and guarantees regarding minimal size). Likewise, my understanding is that the length specifiers do not truncate, they only pad strings in order to ensure that strings end up being at least of a given size. It follows that relying on ints being smaller than the number `2^31 - 1 = 2147483647` or using length specifiers can not be relied upon in order to defend against buffer overflows.

Example: Suppose integers and long long integers on my system are both 64 bit. Then the program

```c
#include <stdio.h>

int main ()
{
    char buffer [15];
    long long int integer = 123456789012345;

    size_t result = snprintf (buffer, sizeof(buffer), "%014lli", integer);
    bool was_truncated = result >= sizeof(buffer);
    printf("Integer value: %lli\n", integer);
    printf("Buffer contains: %s\n", buffer);
    printf("String was truncated: %s\n", was_truncated ? "true" : "false");

    return 0;
}
```

produces the output

```
Integer value: 123456789012345
Buffer contains: 12345678901234
String was truncated: true
```

If `long long int` is replaced by `int`, the output would be the same.

dahlbaek:
@lenormf: I did my best to argue in favor of making the changes. Briefly, I tried to suggest code which will (to the best of my knowledge) not produce overflows or unchecked truncations regardless of input, and such that verification of correct behaviour should be as straightforward as possible.

src/json_ui.cc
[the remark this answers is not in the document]
dahlbaek: Of course, the format string should have been `"%#012llx"` in order to also make room for the preceding `0x`! But the outcome is the same: The length specifier does not protect against a buffer overflow, snprintf protects against a buffer overflow and the check catches the string truncation. For completeness, if you change the above program into
```c
#include <stdio.h>

int main ()
{
    char buffer [15];
    long long int integer = 0x1000000000000000;

    size_t result = sprintf (buffer, "%#012llx", integer);
    bool overflowed = result >= sizeof(buffer);
    printf("Integer value: %#llx\n", integer);
    printf("Buffer contains: %s\n", buffer);
    printf("Buffer overflow:: %s\n", overflowed ? "true" : "false");

    return 0;
}
```
the outcome becomes
```
Integer value: 0x1000000000000000
Buffer contains: 0x1000000000000000
Buffer overflow:: true
```

dahlbaek:
I found some errors in my proposed changes, where the return value of `snprintf`was being recast to an unsigned integer type, clearly not what I had intended. The two commits "Do not recast snprintf return [...]" fix these errors. If you decide to merge these commits, I would suggest that I first squash those two commits; I left them around only for transparency.

src/color.cc
dahlbaek: Just a follow-up remark here: Whether or not this call can overflow the buffer is an implementation detail, depending on the bit size of the integer type `unsigned char`. The `unsigned char` type is capable of containing values *at least* up to 0xff, but there is no standard mandated guarantee that it could not contain higher numbers, such as 0x100, in which case the call could cause a buffer overflow.

src/file.cc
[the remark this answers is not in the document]
dahlbaek: Perhaps you would prefer an application of the `[[gnu::unused]]` attribute?

src/string_utils.cc
```diff
@@ -121,49 +121,63 @@ int str_to_int(StringView str)
 InplaceString<15> to_string(int val)
 {
     InplaceString<15> res;
-    res.m_length = sprintf(res.m_data, "%i", val);
```
dahlbaek: @lenormf wrote:
> This call cannot overflow the buffer?
> 
> Same remark for all the following hunks,[...]

- The buffer may overflow of the bit size of the `integer` type (implementation dependent) is sufficiently large.
- If the `sprintf` call fails, its negative return value will be recast to the same type as `res.m_length`, which is `unsigned char`. Thus, the failed call will go unchecked, and may cause difficult-to-debug problems further down line (I suppose).

My proposed change takes both of these possibilities into account.

Similar remarks apply to the other hunks.

src/json_ui.cc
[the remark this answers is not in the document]
dahlbaek: @lenormf wrote:
> This call cannot overflow the buffer?

Follow-up remark: For this (and only this) call, I think you are right. The string being read into buf is guaranteed to be of size precisely 7 characters (including terminating null character). Still, I think a check for failure of the `sprintf` call is missing. Is there really any compelling reason to go for

```c
sprintf(buf, "\\u%04x", *next)
```

instead of the bullet-proof

```c
snprintf(buf, sizeof(buf), "\\u%04x", *next)
```

?

src/file.cc
```diff
@@ -545,19 +545,21 @@ timespec get_fs_timestamp(StringView filename)
...
 #if defined(__linux__) or defined(__CYGWIN__)
-    ssize_t res = readlink("/proc/self/exe", buffer, 2048);
```
dahlbaek: If nothing else, I think you should consider changing `2048` to `2047`. Otherwise, this is a buffer overflow if the content of `"/proc/self/exe"` is precisely 2048 characters long.

src/file.cc
```diff
@@ -571,7 +573,8 @@ String get_kak_binary_path()
     BPath path(&info.ref);
     return path.Path();
 #elif defined(__DragonFly__)
-    ssize_t res = readlink("/proc/curproc/file", buffer, 2048);
```
dahlbaek: Same as above, I think you should consider changing 2048 to 2047.

src/file.cc
lenormf: `readlink()` does not append a null byte, so this won't overflow. The problem is how the buffer is handled assuming that it will contain at most 2047 bytes.

dahlbaek: I'm not sure I understand your comment. Do you agree that, in the highlighted line of code, 2048 should be changed to 2047?

dahlbaek: I'm not sure i understand your point. Are you pointing out that "this is a buffer overflow" should have been "this causes a buffer overflow two lines below", or are you disagreeing that the solution is to change 2048 to 2047 in the highlighted line of code?

dahlbaek:
abandoned

Issue #2156, opened 2018-06-24
multi-line selection should be commented on the same level with `comment-line`


maximbaz:
Let's use this code snippet:

```python
1: def main():
2:     for i in range(5):
3:         print(i)
4:     print('done')
```

--------------

<details>
<summary>Case 1: Select everything with <code>%</code>, run <code>comment-line</code>.</summary>

Expected:

```python
1: # def main():
2: #     for i in range(5):
3: #         print(i)
4: #     print('done')
```

Actual:

```python
1: # def main():
2:     # for i in range(5):
3:         # print(i)
4:     # print('done')
```
</details>

-------------

<details>
<summary>Case 2: Select lines 2-4, run <code>comment-line</code>.</summary>

Expected: 

```python
1: def main():
2:     # for i in range(5):
3:     #     print(i)
4:     # print('done')
```

Actual:

```python
1: def main():
2:     # for i in range(5):
3:         # print(i)
4:     # print('done')
```
</details>

-----------------

I was hoping to make a PR instead of an issue, but I got stuck at coming up with a smart way to make this selection.

Currently the selection is being trimmed with `<a-s>gi<a-l>`, but it leads to the "Actual" behavior that I want to avoid. So far I was thinking to also trim the selection with `<a-s>gi`, but then to align cursors with the most _left_ one - kind of like the opposite of `&` operator. But there is no such operator, as far as I can see... Maybe there is another way to achieve what I want?

shachaf:
I had a similar issue and wrote my own comment-line -- this the current state of it, with the description of the behavior I want in the comment. Does this sound like the behavior you want as well?

```
# To comment/uncomment a selection containing multiple lines:
# Drop blank lines, unless any line contains unindented text.
# Find the leftmost non-whitespace column among all lines.
# If every line has a comment string starting at this column, delete them.
#   (Optional: Among the lines not ending at the comment string,
#    find the smallest number of spaces shared by all of them
#    immediately after the comment string, and delete that as well.)
# Otherwise insert a comment string before this column.
def comment-line-better %{
  eval %sh{[ -z "$kak_opt_comment_line" ] && echo "fail '%opt{comment_line} not set!"}
  eval -draft -itersel -save-regs '/"a' %{
    exec '<a-x><a-s>gi"aZ'
    try %{
      exec -itersel 'gh<a-K>\S<ret>' # Fail if any line is unindented.
      exec '"az<a-K>^$<ret>"aZ' # Drop blank lines.
[...]

PR #2163, opened 2018-06-27
Uncomment the selected lines only if all of them were commented

### LINKED ISSUE #2161: The selected lines should be uncommented only if all of them were commented
Suppose you have the following snippet:

```yaml
abc:
    # def: ha
    eg: bla
```

You want to comment it all, so you select everything with `%` and execute `comment-line` command.

Expected:

```yaml
# abc:
#     # def: ha
#     eg: bla
```

Actual:

```yaml
abc:
    def: ha
    eg: bla
```

I tried to come up with a solution in order to make PR, but stumbled upon a couple of roadblocks. Below is the snippet I could get to myself, and it has a couple of issues:

1. `Z` and `z` don't seem to have any effect within the command, so I cannot test whether all lines are commented with `<a-K>` and then safely restore the selection.
2. The `execute-keys <a-K>\A\Q%opt{comment_line}\E\h?<ret>` doesn't seem to raise an error when there are no lines left, so the code doesn't go into `catch` block, even though when testing manually I do see a red error message in the bottom saying that "no selections remaining".

Please point me in the right direction, or make your own PR, I don't mind as long as I can get the functionality in `master` :)

```kak
define-command comment-line -docstring '(un)comment selected lines using line comments' %{
    %sh{
        if [ -z "${kak_opt_comment_line}" ]; then
            echo "fail \"The 'comment_line' option is empty, could not comment the line\""
        fi
    }
    evaluate-commands -draft %{
        # Select the content of the lines, without indentation
        execute-keys <a-s>gi<a-l>

        try %{
            # Keep non-empty lines and save the selection
            execute-keys <a-K>\A\s*\z<ret>Z

            try %{
                # See if there are any uncommented lines in the selection
                execute-keys <a-K>\A\Q%opt{comment_line}\E\h?<ret>

                # There are uncommented lines, so restore the selection and comment everything
                set-register '"' "%opt{comment_line} "
                execute-keys zP
            } catch %{
                # All lines were commented, so restore the selection and uncomment them
                set-register / "\A\Q%opt{comment_line}\E\h?"
                execute-keys zs<ret>d
            }
        }
    }
}
```

maximbaz:
Fixes #2161 

-----------

Suppose you have the following snippet:

```yaml
abc:
    # def: ha
    eg: bla
```

You want to comment it all, so you select everything with `%` and execute `comment-line` command.

Expected:

```yaml
# abc:
#     # def: ha
#     eg: bla
```

Actual:

```yaml
abc:
    def: ha
    eg: bla
```

**NOTE:** Because of #2156, the result of this PR doesn't exactly look as "Expected", but it is still better than what we have in `master`:

```yaml
# abc:
  # # def: ha
  # eg: bla
```

### DIFF
--- a/rc/core/comment.kak
+++ b/rc/core/comment.kak
@@ -150,15 +150,18 @@ define-command comment-line -docstring '(un)comment selected lines using line co
         try %{
             # Keep non-empty lines
             execute-keys <a-K>\A\s*\z<ret>
+            set-register / "\A\Q%opt{comment_line}\E\h?"
 
             try %{
-                # Select the comment characters and remove them
-                set-register / "\A\Q%opt{comment_line}\E\h*"
-                execute-keys s<ret>d
-            } catch %{
-                # Comment the line
+                # See if there are any uncommented lines in the selection
+                execute-keys -draft <a-K><ret>
+
+                # There are uncommented lines, so comment everything
                 set-register '"' "%opt{comment_line} "
                 execute-keys P
+            } catch %{
+                # All lines were commented, so uncomment everything
+                execute-keys s<ret>d
             }
         }
     }


mawww:
Would it be possible to add a test case for that as well (take a look at `test/indent/lua/insert-end-only-when-needed/` for an example of how tests can run with bundled scripts), somewhere like `test/indent/comment/comment-all-on-partially-commented` ?

maximbaz:
Sure, I'll have a look

maximbaz:
Added a better implementation with tests in #2167, let me know what you think about that approach.

PR #2167, opened 2018-06-28
Improve comment-line command

### LINKED ISSUE #2156: multi-line selection should be commented on the same level with `comment-line`

### LINKED ISSUE #2161: The selected lines should be uncommented only if all of them were commented

maximbaz:
This basically combines all the improvements I tried to make to the `comment-line` in one change.

---------

Fixes #2156
Fixes #2161
Closes #2160
Closes #2163


maximbaz:
Oh no, it seems the `align-cursors-left` doesn't work on macOS. Does anyone have a Mac to help figure out the command that can work on both OS? Just run these two commands and see if they produce the same output, and if not, tell me which part of the pipeline fails and why:

```
❯ echo '2.5,2.9:1.3,1.11'  | tr ':' '\n' | cut -d',' -f1 | cut -d'.' -f2 | sort -n | head -n1
3

❯ echo '2.5,2.9:1.3,1.11' | sed 's/\.[0-9]\+,/.xxx,/g'                 
2.xxx,2.9:1.xxx,1.11
```

maximbaz:
macOS issues seem to be solved, this is ready for your feedback 😉

net:
I would rather blank lines be commented. If you don't want to comment blank lines, you can use `<a-s><a-k><ret>`; the reverse is more difficult. @shachaf

maximbaz:
I don't mind either way, can you suggest a simple way to avoid adding trailing whitespaces? Currently the comment char is `"# "`, so by adding this to a blank like we will end up with trailing spaces.

P.S. [Should we make more things configurable](https://github.com/mawww/kakoune/pull/2184#issuecomment-402171170) instead of trying to find a solution that everyone will like?

shachaf:
If you use `<a-s>` then the comment character will go on each line's column separately which makes the command pointless. If you want blank lines to be commented, can you describe the behavior you want? Do you want spaces inserted before the comment character, or what?

I'm fine with making things like this configurable if people can't agree on them, but Kakoune's scripting language really doesn't lend itself to it.

shachaf:
Instead of putting `"%opt{comment_line} "` in the register, you can insert `%opt{comment_line}`, drop lines where the cursor is on \n, and then insert a space.

maximbaz:
Just tried your suggestion, I can't get it to work on the following example:

```
1: a:
2:     b: 3
3:         c: 4
4:
5:     e: 5
```

Select lines 2-5 and execute `:comment-line`, I want to achieve the following result:

```
1: a:
2:     # b: 3
3:     #     c: 4
4:     #
5:     # e: 5
```

Without skipping empty lines we will get this result:

```
1: a:
2: #     b: 3
3: #         c: 4
4: #
5: #     e: 5
```

(This example exists also as a unit test `comment-multiple-lines-indented` by the way)

The magic is happening in `align-selections-left` command, so far it was easy - we align the selections to the leftmost cursor of the non-empty selection. But with including empty lines, the leftmost column becomes column 1. So now we need to pad the empty line with a correct amount of spaces somehow 🤔

Feels complicated, I'm happy to hear suggestions, but if you are puzzled too, I tend to keep the current logic and exclude empty lines.

PR #2174, opened 2018-06-30
src: Select the data inserted by `!` and <a-!>

### LINKED ISSUE #1468: Select the data inserted by the `!` and `<a-!>` primitives
Hi,

It would be convenient to have the text inserted by the `!` and `<a-!>` primitives automatically selected after they've been added to the buffer.

Thanks.

lenormf:
Closes #1468

### DIFF
--- a/doc/pages/keys.asciidoc
+++ b/doc/pages/keys.asciidoc
@@ -374,10 +374,10 @@ The default command comes from the *|* register by default (See <<registers#,`:d
     ignore its output.
 
 *!*::
-    insert command output before each selection.
+    insert and select command output before each selection.
 
 *<a-!>*::
-    append command output after each selection.
+    append and select command output after each selection.
 
 == Searching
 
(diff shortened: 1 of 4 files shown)


aver-d:
I've tested this patch for a few months. It works nicely, thanks.

The main advantage is being able to immediately press `s` to select inside the newly inserted text, and it's also useful for moving the entire text for alignment or indentation.

aver-d:
I forgot to mention that I've noticed that when the command returns no output or has an error (eg. `! printf <ret>` or `!unknown-cmd <ret>` ), then the selection extends upwards towards the first character in the buffer.

I typically use `!` programmatically inside scripts, so this has not come up much for me, but I suppose in the case of an error I'd expect no change in the current selection. Maybe an error message would be useful too.

lenormf:
I think I see where the problem is, I'll look into it.

lenormf:
Updated the documentation.

lenormf:
Rebased on master.

PR #2176, opened 2018-07-01
Improve git gutter

### LINKED ISSUE #129: git show-diff flags modified lines as deleted lines

maximbaz:
The current git gutter is quite primitive and doesn't account for many cases, it produces weird result with relatively complicated diffs, tries to mark the same line with different icons and doesn't account for modified lines (#129).

This code is heavily inspired by https://github.com/airblade/vim-gitgutter, but rewritten from scratch in a way that shouldn't introduce any licensing issues for you.

Besides showing modified lines, it also uses `_` and `‾` instead of `-` to better display the position of deleted lines, and it can show the lines with deletions and modifications in the same place (using `~_`).

image

Before:

image

`update-diff` generates:

```
set-option buffer git_diff_flags  :0|{red}-:0|{green}+:0|{red}-:2|{red}-:2|{green}+:4|{red}-:4|{red}-:4|{green}+:7|{green}+:8|{red}-
```

After:

image

`update-diff` generates:

```
set-option buffer git_diff_flags :1|{red}‾:2|{blue}~:4|{blue}~_:7|{green}+:8|{red}-
```

Fixes #129

### DIFF
--- a/rc/extra/git-tools.kak
+++ b/rc/extra/git-tools.kak
@@ -97,25 +97,55 @@ Available commands:\n  add\n  rm\n  blame\n  commit\n  checkout\n  diff\n  hide-
     }
 
     update_diff() {
-        git diff -U0 $kak_buffile | awk '
-            BEGIN {
-                line=0
-                flags=ENVIRON["kak_timestamp"]
+        git --no-pager diff -U0 $kak_buffile | perl -e '
+            $flags = $ENV{"kak_timestamp"};
+            foreach $line (<STDIN>) {
+                if ($line =~ /@@ -(\d+)(?:,(\d+))? \+(\d+)(?:,(\d+))?/) {
+                    $from_line = $1;
+                    $from_count = ($2 eq "" ? 1 : $2);
+                    $to_line = $3;
+                    $to_count = ($4 eq "" ? 1 : $4);
+
+                    if ($from_count == 0 and $to_count > 0) {
+                        for $i (0..$to_count - 1) {
+                            $line = $to_line + $i;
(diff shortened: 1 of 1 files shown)


lenormf:
Did you use `perl` on purpose, knowing that `git` depends on it anyway :p?

maximbaz:
Funnily enough this is actually the first time I write `perl` code in my life 🙂 

I started by looking how to adapt the `awk` code, found out that it is impossible to reference matching regex groups, and looked for alternatives. Perl seemed like a good choice, it is widely available on different distributions and has good regex support... and now I know `git` depends on it too 😉

mawww:
Hi, could you add a copyright waiver commit (see the CONTRIBUTING file), ideally in a new PR as you have a lot of PRs opened and I'd like to merge it first.

maximbaz:
Done: #2182 

Looking forward to see some PRs merged now 😛

mawww:
Thanks

net:
The way I handle modifications next to deletions is by underlining (+u) the modification character. This has the effect of putting the deletion and the modification indicator in the same column. I prefer it to two separate columns. (I also use "▹" instead of "~", and purple instead of blue.)

screen shot 2018-07-02 at 10 38 47 am

maximbaz:
Good idea, thanks guys, I'll make a follow-up PR to use underlined `~` (I also prefer this char), and I got some extra feedback in https://github.com/mawww/kakoune/commit/379e7c2d2d0968ad654e8e60b7164e175d66624c#r29565351 which I will address at the same time.

PR #2181, opened 2018-07-02
Add Red Phoenix colorscheme

boj:
Dark terminal colorscheme based on dkeg's work located on https://terminal.sexy

2018-07-02-140149_1345x1079_scrot

### DIFF
--- a/colors/red-phoenix.kak
+++ b/colors/red-phoenix.kak
@@ -1,0 +1,88 @@
+##
+## Red Phoenix (dkeg) - adapted by boj
+##
+
+evaluate-commands %sh{
+    black="rgb:000000"
+    blue="rgb:81a2be"
+
+    orange1="rgb:F2361E"
+    orange2="rgb:ED4B19"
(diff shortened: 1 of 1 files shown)


net:
That looks great. The highlight contrast seems poor though.

mawww:
I agree that the selection with red text seems a bit hard to read, I dont mind that much but if you wish to tweak it please do.

boj:
No problem, I'll experiment a little further and push a change with better contrast.

boj:
2018-07-04-131433_923x449_scrot

That seems to be a tad better.

lenormf:
Also split those selections, so we can see how secondary selections are highlighted!

mawww:
Seems the `face` commands in this colorscheme are using the old, unscoped syntax, you are probably using a pretty old version of Kakoune, faces can now be set to different values in different scopes (global/buffer/window). the `face` command now takes a scope arguments as first argument.

boj:
I've updated this to use the modern syntax.

PR #2184, opened 2018-07-02
Further git-gutter improvements

maximbaz:
Follow-ups from https://github.com/mawww/kakoune/commit/379e7c2d2d0968ad654e8e60b7164e175d66624c#r29565351 and #2176

image

### DIFF
--- a/rc/extra/git-tools.kak
+++ b/rc/extra/git-tools.kak
@@ -97,7 +97,6 @@ Available commands:\n  add\n  rm\n  blame\n  commit\n  checkout\n  diff\n  hide-
     }
 
     update_diff() {
-        git rev-parse --show-toplevel >/dev/null 2>&1 || return
         git --no-pager diff -U0 "$kak_buffile" | perl -e '
             $flags = $ENV{"kak_timestamp"};
             foreach $line (<STDIN>) {
--- a/rc/extra/git-tools.kak
+++ b/rc/extra/git-tools.kak
@@ -97,7 +97,8 @@ Available commands:\n  add\n  rm\n  blame\n  commit\n  checkout\n  diff\n  hide-
     }
 
     update_diff() {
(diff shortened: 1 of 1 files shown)


rc/extra/git-tools.kak
```diff
@@ -97,7 +97,8 @@ Available commands:\n  add\n  rm\n  blame\n  commit\n  checkout\n  diff\n  hide-
     }
 
     update_diff() {
-        git --no-pager diff -U0 "$kak_buffile" 2>/dev/null | perl -e '
+        git rev-parse --show-toplevel >/dev/null 2>&1 || return
+        git --no-pager diff -U0 "$kak_buffile" | perl -e '
```
maximbaz: @lenormf what are your thoughts on this approach? I can put the check in my hook, but maybe having it here is better to benefit everyone. Basically the update_diff command will now do nothing if we are not in git repo.

maximbaz: Actually, I realized that it defeats the purpose, your point was that people need to see the error, and in this case people will still see nothing in `*debug*`. I'll update.

lenormf:
Looks good.

net:
I personally find the tilde to be a pretty ugly character to display in the gutter. Perhaps we could consider other ASCII characters to represent modification. I do realize I may be alone in this.

maximbaz:
Can you suggest some other characters that you would prefer over tilde? 

---------------

To be honest, my current impression with kakoune is that too few things are configurable. It's not okay that you are forced to copy-paste the entire thing if you want to use a pencil icon instead of tilde, just like it's not okay to ask @shachaf to copy-paste the entire `comment.kak` if they don't want to have a space after a comment character (ref: #2156). I wonder, has this topic been brought up before?

------------

😛 

image

PR #2185, opened 2018-07-02
Add missing python3 keyword: nonlocal

maximbaz:


lenormf:
Ah, it was introduced in 3.7! Maybe we should come up with a one-liner that prints out all the language's keywords, and put that in a comment in the language file?

Screwtapello:
I was wondering how one could possibly write a Python script to determine the list of Python keywords, and then I discovered:

    $ python3 -c 'import keyword; print("\n".join(keyword.kwlist))'

maximbaz:
That's what I used as well, there is a hint just two lines above in `python.kak` 🙂

```
# Keyword list is collected using `keyword.kwlist` from `keyword`
```

PR #2186, opened 2018-07-03
src: Propagate to clients the `kill` exit code

### LINKED ISSUE #2177: Killing a server with connected clients results in a non-zero exit code
### Steps 

* start a server: `kak -n -s test -d`
* connect a client to it: `kak -c test`
* from another shell, kill the server: `echo kill | kak -p test`

### Outcome 

The client outputs:

```
socket read failed: Interrupted system call
disconnecting                              
```

The exit code is set to 255.

### Expected 

I think the `kill` command should cleanly disconnect all clients, who should subsequently return 0.

Additionally, I think the `kill` command should take an optional exit code, which would then be sent to the clients so that they can cleanly quit with that same exit code.

HTH.

lenormf:
Fixes #2177

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -434,30 +434,44 @@ static void ensure_all_buffers_are_saved()
     throw runtime_error(message);
 }
 
+template<bool force>
+void kill(const ParametersParser& parser, Context& context, const ShellContext&)
+{
+    auto& client_manager = ClientManager::instance();
+
+    if (not force and client_manager.count() == 1)
+        ensure_all_buffers_are_saved();
+
+    const int status = parser.positional_count() > 0 ? str_to_int(parser[0]) : 0;
+    while (not client_manager.empty())
+        client_manager.remove_client(**client_manager.begin(), true, status);
+
+    throw kill_session{};
+}
+
 const CommandDesc kill_cmd = {
     "kill",
     nullptr,
-    "kill current session, quit all clients and server",
-    no_params,
+    "kill [exit_code]: terminate the current session, the server and all clients connected"
+    "; the given integer will be used as exit code by the server and client processes",
+    { {}, ParameterDesc::Flags::SwitchesAsPositional, 0, 1 },
     CommandFlags::None,
     CommandHelper{},
     CommandCompleter{},
-    [](const ParametersParser&, Context&, const ShellContext&){
-        ensure_all_buffers_are_saved();
-        throw kill_session{};
-    }
+    kill<false>
 };
 
 
 const CommandDesc force_kill_cmd = {
     "kill!",
     nullptr,
-    "kill current session, quit all clients and server, do not check for unsaved buffers",
-    no_params,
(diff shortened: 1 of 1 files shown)


src/commands.cc
```diff
@@ -412,30 +412,42 @@ static void ensure_all_buffers_are_saved()
...
 }
 
+template<bool force>
+void kill(const ParametersParser& parser, Context& context, const ShellContext&)
+{
+    if (not force and ClientManager::instance().count() == 1)
+        ensure_all_buffers_are_saved();
+
+    const int status = parser.positional_count() > 0 ? str_to_int(parser[0]) : 0;
+    for (auto& client : ClientManager::instance())
```
mawww: This will be buggy, as once we start removing clients, the iterators are going to be invalidated. This should be something like `while (not ClientManager::instance().empty()) { ... }`

src/commands.cc
```diff
@@ -412,30 +412,47 @@ static void ensure_all_buffers_are_saved()
...
+    auto& client_manager = ClientManager::instance();
+
+    if (not force and client_manager.count() == 1)
+        ensure_all_buffers_are_saved();
+
+    const int status = parser.positional_count() > 0 ? str_to_int(parser[0]) : 0;
+    while (not client_manager.empty())
+    {
+        auto client = client_manager.begin()->get();
+        client_manager.remove_client(*client, true, status);
```
mawww: `**client_manager.begin()` would do, although I am tempted to add a method to ClientManager to clear all clients.

src/main.cc
```diff
@@ -725,7 +725,10 @@ int run_server(StringView session, StringView server_init,
             }
         }
     }
-    catch (const kill_session&) {}
+    catch (const kill_session& e)
+    {
+        local_client_exit = e.status;
```
mawww: There should be no need to do that as the local client will get removed as every other clients and will set that value already.

src/commands.cc
lenormf: I thought that using so many levels of indirection at once was a bit ugly and made the code less easy to understand, do you really want to get rid of that temporary variable?

mawww: Yeah, I'd rather have the double indirection.

lenormf:
Rebased the code on `master` and updated it to comply with the requested changes.

src/commands.hh
```diff
@@ -6,7 +6,10 @@ namespace Kakoune
 
 void register_commands();
 
-struct kill_session {};
+struct kill_session
+{
```
mawww: We dont need this status field anymore, do we ?

lenormf: Actually, don't we need to set `local_client_exit` so that the server itself can exit with the status code stored in the exception?

mawww: local_client_exit is set by the client m_on_exit callback (see line 700).

lenormf: Fixed in the latest push.

Issue #2191, opened 2018-07-05
Things that are broken in master after big-breaking-tidy-up


maximbaz:
Making a list as I wouldn't be able to look into these right now, maybe someone has a chance to have a look:

- [x] `git show-diff` now silently fails. The issue is probably here:

https://github.com/mawww/kakoune/blob/b06ad0650747f79034e64dce60832ff824b3cb84/rc/extra/git-tools.kak#L188

- [x] `'lint-enable' 2:6: 'add-highlighter' no parent in path`

The above is fixed with #2192 

----------

More issues:

- [x] `'go-disable-autocomplete' 2:6: 'set-option' invalid completer description: '''filename' 'word=all'''`
- [ ]   `lint`:
    ```
    error running command 'set-option "buffer=/path/to/code.go" lint_flags 1
    set-option "buffer=/path/to/code.go" lint_errors 1
    set-option "buffer=/path/to/code.goo" lint_error_count 0
    set-option "buffer=/path/to/code.go" lint_warning_count 0
    evaluate-commands -client  lint-show-counters
    ': 5:484: 'evaluate-commands' wrong argument count
    ```
- [x] `lint`: even if I "fix" the above by replacing `${kak_client}` with `${kak_client:-default0}`, I see the lint highlighters in the gutter, and I see all the issues in `*lint-output*` buffer, but navigation with `:lint-next-error` doesn't work, it prints an error `no lint diagnostics`.

maximbaz:
I looked into `lint` issue, the problem is with command:
```
evaluate-commands -client  lint-show-counters
```

In the code it looks as `-client $kak_client`, and as you can see the parameter `$kak_client` got expanded to an empty string.

I think it's some kind of race condition, I have `lint` call running in `hook global WinSetOption filetype=XX` because I want to immediately lint a buffer that I've just opened (or if I change its filetype). It used to work before the tidy-up, but now the kak_client is not set during WinSetOption hook.

@mawww does it ring any bell, why $kak_client is not set during WinSetOption hook? Was it an accidental or intended change?

maximbaz:
Regarding $kak_client I've just stumbled upon #1872, except I think my case is different. First of all, it definitely worked before the big-breaking-tidy-up. But also I only have one window, and I have issues with missing $kak_client even if I open kak first, and then open the file with `:e file.go` (so the window definitely exists at that time).

mawww:
[...]

PR #2192, opened 2018-07-06
Fix highlighters after big-breaking-tidy-up

maximbaz:
- [x] Fix git-tools highlighter
- [x] Fix lint highlighter

See #2191

### DIFF
--- a/rc/base/lint.kak
+++ b/rc/base/lint.kak
@@ -87,13 +87,13 @@ define-command -hidden lint-show-counters %{
 }
 
 define-command lint-enable -docstring "Activate automatic diagnostics of the code" %{
-    add-highlighter window flag_lines default lint_flags
+    add-highlighter window/lint-highlighter flag_lines default lint_flags
     hook window -group lint-diagnostics NormalIdle .* %{ lint-show }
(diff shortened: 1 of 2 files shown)


rc/base/lint.kak
```diff
@@ -87,13 +87,13 @@ define-command -hidden lint-show-counters %{
 }
 
 define-command lint-enable -docstring "Activate automatic diagnostics of the code" %{
-    add-highlighter window flag_lines default lint_flags
+    add-highlighter window/lint-highlighter flag_lines default lint_flags
```
mawww: why all those `-highilghter` in highlighter names ? from the context we already know those are highlighters, so it seems unnecessary to add that in the name as well. I'd prefer just `lint` here, `git-blame` there...

maximbaz: I saw in `git-tools` the existing highlighters have this prefix, I thought it's a pattern that you introduced and I wanted to keep the consistency 🙂

add-highlighter window/git-log-highlight
add-highlighter window/git-status-highlight

Do you want me to drop the `-highlight` suffix from everywhere?

maximbaz: There's a commit now that removes the suffix in the files I touched

Issue #2152, opened 2018-06-23
kak starts noticeably slower than neovim


maximbaz:
Install `kakoune` on Arch Linux from the official repo.

Run `kak -n` — kakoune starts instantly (perfect).
Run `kak` — kakoune starts noticeably slowly, even though I didn't even create `~/.config/kak/` folder yet.
I use `neovim` with lots of plugins and huge config file, and it starts faster than `kak`.

Given that I open and close editor many times during the day, it is essential to have the editor start as fast as possible.

I'm not sure how I can provide more information to narrow down the cause of slowness. What does `kak` do, that `kak -n` doesn't?

TeddyDD:
You could try to measuer startup time like this:
```
time -p kak -e 'quit'
real 0.48
user 0.21
sys 0.31
```
It takes about half a second to lanuch in my case, and I have shitload of plugins. I wonder how long it takes in your case.

SolitudeSF:
manually sourcing each file from kakrc is faster than load same files from autoload.
autoloading all default `kak` files takes `~115ms` while when putting them in autoload takes `~170ms`.
why is that?

maximbaz:
I did some additional benchmarking:

---------------

nvim with no config: `25ms`

```
❯ time (repeat 100 nvim -u NONE -c 'quit')
1.33s user 0.87s system 88% cpu 2.490 total
```

nvim with huge config and lots of plugins: `114ms`

```
❯ time (repeat 100 nvim -c 'quit')
9.20s user 1.61s system 94% cpu 11.390 total
```

`kak -n`: `7ms` (awesome result!)

```
❯ time (repeat 100 kak -n -e 'quit')
0.35s user 0.35s system 99% cpu 0.703 total
```

`kak` with no custom config (except whatever ArchLinux package ships): `400ms`

```
❯ time (repeat 100 kak -e 'quit')
31.48s user 12.14s system 108% cpu 40.057 total
```

-------------

So before I even start configuring `kak`, it already takes almost 4 times longer to load than my fully customized `nvim` — let's try to get this number down 😉 

`:echo %val{runtime}/autoload<ret>` outputs: `/usr/share/kak /autoload`

I don't have the folder `/autoload`, but I do have the folder `/usr/share/kak` present.

Creating a folder `~/.config/kak/autoload` does make `kak` load much faster, as if I provided the `-n` argument. However, I do want to have all those files loaded 🙂 

------------------

[...]

PR #2196, opened 2018-07-08
Improve kakoune startup time by 33%

maximbaz:
Before:

```
❯ time (repeat 300 kak -e 'quit')
115.89s user 48.61s system 108% cpu **2:31.93 total**
```

After:

```
❯ time (repeat 300 kak -e 'quit')
82.50s user 31.23s system 111% cpu **1:42.31 total**
```

Related to #2152, but not good enough to close it just yet.

**Can anyone test this on Mac?**

### DIFF
--- a/share/kak/kakrc
+++ b/share/kak/kakrc
@@ -28,7 +28,7 @@ def -params 1 -docstring "colorscheme <name>: enable named colorscheme" \
 evaluate-commands %sh{
     autoload_directory() {
         find -L "$1" -type f -name '*\.kak' \
-            -exec printf 'try %%{ source "%s" } catch %%{ echo -debug Autoload: could not load "%s" }\n' '{}' '{}' \;
+            | sed 's/.*/try %{ source "&" } catch %{ echo -debug Autoload: could not load "&" }/'
     }
 
     echo "colorscheme default"


maximbaz:
@SolitudeSF https://github.com/mawww/kakoune/issues/2152#issuecomment-399688866:

> manually sourcing each file from kakrc is faster than loading same files from autoload.
manually loading all default kak files takes ~115ms while when putting them in autoload takes ~170ms.
> why is that?

Now you have the answer, because spawning `printf` with `-exec` for every file is expensive it seems.

maximbaz:
For historical purposes, these are the approaches I've tested:

```sh
# Status quo, slow
find -L "$1" -type f -name '*\.kak' \
    -exec printf 'try %%{ source "%s" } catch %%{ echo -debug Autoload: could not load "%s" }\n' '{}' '{}' \;

# 33% improvement, -printf exists only on Linux
find -L "$1" -type f -name '*\.kak' \
    -printf 'try %%{ source "%p" } catch %%{ echo -debug Autoload: could not load "%p" }\n'

# 33% improvement
find -L "$1" -type f -name '*\.kak' \
    | sed 's/.*/try %{ source "&" } catch %{ echo -debug Autoload: could not load "&" }/'

# 27% improvement
find -L "$1" -type f -name '*\.kak' | while read script; do
    echo "try %{ source '$script' } catch %{ echo -debug Autoload: could not load '$script' }"
done
```

casimir:
Applying the patch locally on macOS I can confirm it's starting between 35 and 40% on my laptop.

maximbaz:
Awesome, thanks for testing!

dahlbaek:
Also verifying on a slow laptop running OpenBSD:

Before patch:

```
localhost% time (for i in {1..100}; do kak -e "quit"; done;)
( for i in {1..100}; do; kak -e "quit"; done; )  82.01s user 219.24s system 88% cpu 5:41.68 total
```

After patch:
```
localhost% time (for i in {1..100}; do kak -e "quit"; done;)
( for i in {1..100}; do; kak -e "quit"; done; )  75.93s user 177.15s system 87% cpu 4:48.80 total
```

So, approximately 15% reduction in startup time for me.

Issue #2199, opened 2018-07-08
Navigate completions list by columns, not by rows


maximbaz:
I cannot get used to navigating kakoune completions horizontally:

![kak-completions-navigation](https://user-images.githubusercontent.com/1177900/42423425-402fa9d4-82fa-11e8-92a3-751336c8a24b.gif)

Everywhere else I know a column navigation is used instead, e.g. here's `zsh`:

![vertical-list-navigation](https://user-images.githubusercontent.com/1177900/42423340-0d3f224e-82f9-11e8-97fd-2686d3b0d3cd.gif)

Bash also draws completions vertically:

![image](https://user-images.githubusercontent.com/1177900/42423548-6330f10c-82fc-11e8-87dc-fa49613c70db.png)

`ls` and `exa` also order files in columns:

![image](https://user-images.githubusercontent.com/1177900/42423561-849f9d02-82fc-11e8-9c7c-772d6e435585.png)

Can we please make the completions list in kakoune also conform to this pattern? At least optionally.

It's much easier to compare alphabetically-sorted files when they are drawn in columns, if I'm looking for the file "client.cc", it is easy to see at a glance that files whose names start with a letter "c" are in the second column.

Related: #1508

lenormf:
We read text from left to right, and then go to the next line downwards, which makes the current implementation _theoretically_ more intuitive.

This issue has probably more to do with habits than an actual problem we can solve using logic (neither layout seem fundamentally flawed, they're just a design choice).

maximbaz:
I'll be happy as long as any way to choose the vertical layout, so an option as proposed in #2212 sounds like a perfect solution 😉

maximbaz:
Fixed with #2286

PR #2212, opened 2018-07-13
Added an option for vertical menu layout

laelath:
Added the ability to have menus in the ncurses ui laid out alphabetically in vertical columns. Also added an option to manage the layout with `ncurses_menu_layout` in `ui_options`.
Based on the request by #2199

### DIFF
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -741,16 +741,31 @@ void NCursesUI::draw_menu()
     const LineCount mark_height = min(div_round_up(sq(win_height), menu_lines),
                                       win_height);
     const LineCount top_line = m_menu.first_item / m_menu.columns;
-    const LineCount mark_line = (win_height - mark_height) * top_line / max(1_line, menu_lines - win_height);
+
+    const int menu_cols = div_round_up(item_count, (int)m_menu.size.line);
+    const int first_col = m_menu.first_item / (int)m_menu.size.line;
+
+    const LineCount mark_line = m_menu_vertical ? (win_height - mark_height) * first_col / max(1, menu_cols - m_menu.columns) :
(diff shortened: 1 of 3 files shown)


maximbaz:
Sooo cool, thank you for making this happen!

src/ncurses_ui.cc
```diff
@@ -1228,6 +1256,14 @@ void NCursesUI::set_ui_options(const Options& options)
         m_wheel_down_button = wheel_down_it != options.end() ?
             str_to_int_ifp(wheel_down_it->value).value_or(5) : 5;
     }
+
+    {
+        auto it = options.find("ncurses_menu_layout"_sv);
+        if (it == options.end() or it->value == "horizontal")
+            m_menu_vertical = false;
+        else if (it->value == "vertical")
+            m_menu_vertical = true;
```
lenormf: Those lines can be simplified into

```cpp
m_menu_vertical = it != options.end() and it->value == "vertical";
```

laelath: There's a slight difference, if it was set to something other than "horizontal" or "vertical" the value isn't changed with the current code.

lenormf: I understand, but since no errors are shown anyway, it's simpler to treat any value that doesn't set the boolean to `true` as `false`. That's how the other options are handled as well.

laelath: That makes sense, I've implemented the change.

Screwtapello:
I don't want to derail this, but have you considered removing the option and just making the layout always vertical? I mean, phonebooks don't have an option to toggle them back to horizontal-layout, and @mawww generally seems to be OK with making incompatible changes without an option, if the result is a strict improvement.

laelath:
I created an option for it because I figured it was easier to change from that to just vertical columns if @mawww requested it.

PR #2218, opened 2018-07-17
Add a new hook InsertCharPre

yshui:
This hook is the same as InsertChar, except it's triggered
right before the character is inserted.

Rational: plugins might want the cursor position right before the character is inserted (In my case, I want to save the cursor position when the buffer is unmodified, so I can restore it after closing kakoune). This is non-trivial to work out in the InsertChar hook.

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -803,7 +803,7 @@ const CommandDesc remove_highlighter_cmd = {
 static constexpr auto hooks = {
     "BufCreate", "BufNewFile", "BufOpenFile", "BufClose", "BufWritePost",
     "BufWritePre", "BufOpenFifo", "BufCloseFifo", "BufReadFifo", "BufSetOption",
-    "InsertBegin", "InsertChar", "InsertDelete", "InsertEnd", "InsertIdle", "InsertKey",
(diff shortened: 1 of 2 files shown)


yshui:
Hmm, I don't think this is good enough for my goal. What I really want is to hook the change of "modified" state.

This could still be useful for other things

maximbaz:
Shouldn't `InsertChar` be renamed to `InsertCharPost` in such case, to stay consistent with other hooks?

yshui:
@maximbaz Yes, but that will be a breaking change.

maximbaz:
That's not an issue, we've just had a few breaking changes merged to master a couple of weeks ago - see the [changelog for development version](https://github.com/mawww/kakoune/blob/master/doc/pages/changelog.asciidoc#development-version).

laelath:
Couldn't you save the position on the last `NormalIdle` or `InsertIdle`?

yshui:
@laelath I guess I could. I didn't notice the default idle timeout is just 50ms, meaning I need to type 20 characters per second to keep idle hooks from being triggered. Isn't hooking into that basically the same as hooking RawKey?

maximbaz:
So it sounds @yshui you won't be using this hook and there aren't any other examples where it could be useful, shall we close this PR until we see a real need?

yshui:
@maximbaz Agreed.

PR #2221, opened 2018-07-18
rc lint: Fix the previous/next navigation functions, fix the diagnosis anchors, add an error message

lenormf:


maximbaz:
I'm curious, do you actually see the lint messages now? Since big-breaking-tidy-up I see the red icon in gutter, but when my cursor is on the line, I don't see actual lint message anymore. There's nothing interesting in `*debug*` buffer, and in fact I do have the `*lint-output*` with lint messages, but they don't show up when cursor is on the line with the error.

lenormf:
Good point, I actually don't see the message anymore either!

I'll update this PR when I've fixed it.

lenormf:
Should be working now.

maximbaz:
Yeah, I confirm, great job! This fixes point 5 in #2191.

PR #2224, opened 2018-07-19
src: Rename `auto_complete` to `autocomplete`

lenormf:
Removing the underscore seems to make the option name more
in line with the others (even though some do use a separator,
e.g. `disabled_hooks`).


Delapouite:
Yes we have to choose one side. All of them with an underscore or none of them.

lenormf:
I think the separator is justified when an option name is clearly made of two words, but in that case it's a bit confusing (`autoinfo` doesn't use an underscore for example).

lenormf:
Bump!

PR #2226, opened 2018-07-19
Add g++ as new build dependency

jkonecny12:
From Fedora 29 the gcc package is not part of the buildroot environment.


jkonecny12:
The builds are working correctly now. Thanks @mawww :+1:

PR #2232, opened 2018-07-20
src: Don't quote integer lists expanded in a shell

### LINKED ISSUE #2223: int-list: first item is printed with quotes
### Steps 
```
decl int-list blub
set global blub 1 2 3
eval %sh{ printf "echo %s\\n" "$kak_opt_blub" }
eval %sh{ printf "echo (%s)\\n" "$kak_opt_blub" }
eval %sh{ printf "echo ( %s )\\n" "$kak_opt_blub" }
```
### Outcome 
```
1 2 3
('1' 2 3 )
( 1 2 3 )
```

### Expected 
```
1 2 3
(1 2 3)
( 1 2 3 )
```

Not sure if my expectation is reasonable.

lenormf:
Fixes #2223

### DIFF
--- a/src/option_types.hh
+++ b/src/option_types.hh
@@ -79,6 +79,12 @@ String option_to_string(const Vector<T, domain>& opt)
     return join(opt | transform([](const T& t) { return quote(option_to_string(t)); }), ' ', false);
 }
 
+template<MemoryDomain domain>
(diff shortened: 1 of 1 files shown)


mawww:
This was fixed with a more generic system.

PR #2236, opened 2018-07-20
Fix a problem with scrolling after search

yshui:
If you scroll right after searching, the scroll should start from the
search result, but instead the scroll just starts as if the search never
happened.

This commit fixes that.


mawww:
This commit seems to remove the restoring of window position while searching, this code exist to place the window back at its original position when using incsearch (incsearch will jump around the buffer while we are writing the search pattern, without this, those jump around have an influence on window position). 

I dont think just removing that code is a good solution, could you describe the problem a bit more, I am not sure what you are actually trying to solve.

yshui:
@mawww I thought I described the problem quite clearly. But I will list steps to reproduce:

* Open some file in kakoune
* Press '/' to start searching
* Search for a word, preferable a word that's further down the file
* Press enter
* Scroll down with mouse

Expected behavior:

File is scrolled down from current position

Actual behavior:

File is scrolled down from the beginning of the file

yshui:
@mawww If you are talking about returning to previous position after delete the search pattern, that still happens after this patch.

mawww:
I cannot reproduce that here, I open main.cc, search for `\btry`, hit enter then hit the mouse wheel down once. And things behave as expected.

Could you capture an asciicast/gif/video of the bug as it happens on your machine ?

yshui:
@mawww https://asciinema.org/a/QYSIOsqQREGGAKTuGpQ74cFcs

Note that the file jump back to the top because I scrolled.

maximbaz:
I reproduce it too, with exactly these steps, in kitty, with `kak -n`.

> I open main.cc, search for \btry, hit enter then hit the mouse wheel down once

aeosynth:
I reproduced without mouse, using `pagedown` and `<c-f>`, on v2018.09.04, kitty terminal

PR #2240, opened 2018-07-21
Recognize systemd timer syntax

maximbaz:
image


Clyybber:
Unrelated, but may I ask what font you are using? I've lately seen quite a few of your screenshots and it looks beautiful.

maximbaz:
@Clyybber it is [Fira Code](https://github.com/tonsky/FiraCode) 🙂

lenormf:
I think it'd be better to handle all of `systemd`'s configuration files (which use the INI syntax) in a separate dedicated support script, rather than adding unrelated extensions to `ini.kak`.

Extensions used by `systemd` files are `.service`, `.mount`, `.timer`, `.target`, `.path`…

mawww:
Yep, moving systemd file detection to an `extra/systemd.kak` file would be nice

PR #2249, opened 2018-07-25
Make ncurses title configurable

yshui:
When ncurses_set_title is set to a string other than "yes" and "true", it will be used to set the title line. This give the user the ability to configure the title line.

### DIFF
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -458,7 +458,7 @@ void NCursesUI::draw_status(const DisplayLine& status_line,
     {
         char buf[4 + 511 + 2] = "\033]2;";
         auto buf_it = &buf[4];
-        if (m_title_line != "")
+        if (not m_title_line.empty())
         {
             auto buf_end = &buf[511 + 1];
             for (auto ch: m_title_line)
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -463,7 +463,7 @@ void NCursesUI::draw_status(const DisplayLine& status_line,
             auto buf_end = &buf[511 + 1];
             for (auto ch: m_title_line)
             {
-                if (buf_it == buf_end)
+                if (buf_it + 1 == buf_end)
                     break;
                 *buf_it++ = ch;
             }
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -456,19 +456,34 @@ void NCursesUI::draw_status(const DisplayLine& status_line,
 
     if (m_set_title)
     {
-        constexpr char suffix[] = " - Kakoune\007";
         char buf[4 + 511 + 2] = "\033]2;";
-        // Fill title escape sequence buffer, removing non ascii characters
-        auto buf_it = &buf[4], buf_end = &buf[4 + 511 - (sizeof(suffix) - 2)];
-        for (auto& atom : mode_line)
+        auto buf_it = &buf[4];
+        if (m_title_line != "")
         {
-            const auto str = atom.content();
-            for (auto it = str.begin(), end = str.end();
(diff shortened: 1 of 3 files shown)


maximbaz:
Does this allow using variables, e.g. can I put a buffer name in the title? If not, this is not very useful.

lenormf:
I don't think the string is being expanded, and also this is probably not the most elegant way to allow users to configure an option (`bool`/`str` hybrid type).

The title of the terminal also doesn't seem to be that important that users should be able to configure it manually, maybe it should just be a static format string in the code (e.g. `Kakoune - %val{bufname}` which would be expanded)?

maximbaz:
I agree that the way of configuring title in PR is weird and useless if we can't have strings expanded in it.

However I don't want to see the link between modeline and title go away completely in favor of static string, I personally do see the title important and I rely on having there the path to the current buffer, but with `~` instead of `$HOME` to save space. In other words, I want there to be `~/.config/kak/kakrc`, and right now I adapted my modeline to have this info. 

I usually have many terminal windows open and use this both to visually distinguish them and with `rofi -show window` to quickly focus a particular terminal.

My ideal title would be `~/.config/kak/kakrc - Kakoune` (with `Kakoune` in the end in case the string doesn't fit the screen and needs to be trimmed), but of course right now I have there more info (things that I need in modeline).

yshui:
@maximbaz @lenormf This string not being expanded is intentional. The intended way to use it is to setup a BufCreate hook, and set the title there.

yshui:
@lenormf The title of the terminal can be pretty important, esp. for people using tmux.

yshui:
@maximbaz I assume that user who cares about title line probably don't want it to change all the time, and do it this way means the change to the code would be minimal.

maximbaz:
It does feel correct to leave it to the user to choose how often to refresh the title, with what I initially expected (automatically expanding variables) kakoune would have refreshed the title all the time.

So after you explained the intended usage and posted a working example, I retract my earlier comments and I support this PR — definitely gonna use this feature myself.

hacktivista:
Changing the title on each key I press is very annoying for me, since I track my activity based on window titles D: having a title unchanged for every open kakoune session is equally useless for me. lol

Hope this gets merged someday

src/ncurses_ui.cc
```diff
@@ -456,19 +456,34 @@ void NCursesUI::draw_status(const DisplayLine& status_line,
 
     if (m_set_title)
     {
-        constexpr char suffix[] = " - Kakoune\007";
         char buf[4 + 511 + 2] = "\033]2;";
-        // Fill title escape sequence buffer, removing non ascii characters
-        auto buf_it = &buf[4], buf_end = &buf[4 + 511 - (sizeof(suffix) - 2)];
-        for (auto& atom : mode_line)
+        auto buf_it = &buf[4];
+        if (m_title_line != "")
```
lenormf:
```suggestion
        if (not m_title_line.empty())
```

src/ncurses_ui.cc
```diff
@@ -456,19 +456,34 @@ void NCursesUI::draw_status(const DisplayLine& status_line,
...
-                 it != end and buf_it != buf_end; utf8::to_next(it, end))
-                *buf_it++ = (*it >= 0x20 and *it <= 0x7e) ? *it : '?';
+            auto buf_end = &buf[511 + 1];
+            for (auto ch: m_title_line)
+            {
+                if (buf_it == buf_end)
+                    break;
+                *buf_it++ = ch;
+            }
+            *buf_it = '\007';
```
lenormf: If the loop above reaches `buf_it == buf_end`, then this line will cause a bug.

hacktivista:
ping @mawww

yshui:
Rebased and revived at #3829

Issue #2175, opened 2018-06-30
Always unconditionally remove trailing white spaces


maximbaz:
I noticed that every language in `rc` has to deal with trailing white spaces on its own, and `ag '(remove|clean).*trailing.*space' --stats` shows that there are currently 38 files that clean white spaces. This is a lot of duplicated logic! 

With the same motivation "kak is a code editor, and not arbitrary data editor" as mentioned in #2147, I propose to consider making kak always remove trailing white spaces. A simple snippet below should simplify those 38 files and also remove the handling of trailing white spaces from editorconfig parsing:

```
define-command trim-whitespaces -docstring 'trim trailing whitespaces' %{
    try %{ execute-keys -draft \%s\h+$<ret>d }
}

hook global BufWritePre .* %{ trim-whitespaces }
```

danr:
I think this is blocked because there is no reliable way to track dependencies between scripts, see mawww's comment here: https://github.com/mawww/kakoune/pull/1811#issuecomment-360634833

lenormf:
Actually the language support scripts I've written do not handle trailing whitespace, as I think it's not the indentation hooks' purpose to handle that.

Secondly, removing all trailing whitespace by default without the user's consent is also bad, the editor shouldn't assume how the user wants a buffer's contents to be formatted.

Example: I regularly have to fix code that wasn't written by me, that has trailing whitespace everywhere (and even LF and CRLF line endings mixed together), if the editor were to arbitrarily fix those issues, I would end up with patches that are 99% cosmetic changes. If the user wants proper formatting, they should use `:format` or declare hooks in their configuration.

Finally it's not an issue that some code is duplicated across a few files, as the user is free to load whatever script they want, and factorizing redundant code would make things harder to cherry-pick scripts in `autoload` (c.f. @danr's answer regarding dependencies).

maximbaz:
There is probably no source code that requires trailing whitespaces to be preserved, `git diff -w` and modern code review tools are able to hide whitespace differences, so I don't think it's too bad if kakoune would slightly clean code files that you touch. It's a good practice to not have trailing whitespaces in the code, and cleaning them up is not the same as reformatting the file.

[...]

PR #2253, opened 2018-07-25
Added google protobuf syntax highlighting

laelath:


rc/extra/protobuf.kak
```diff
@@ -0,0 +1,89 @@
...
+    try %[ execute-keys -itersel -draft <a-h><a-k>^\h+\}$<ret>hms\A|.\z<ret>1<a-&> ]
+]
+
+# Initialization
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+hook -group protobuf-highlight global WinSetOption filetype=protobuf %{ add-highlighter window/protobuf ref protobuf }
+
+hook global WinSetOption filetype=protobuf %[
+    # cleanup trailing whitespaces when exiting insert mode
```
lenormf: Are trailing whitespace strictly forbidden by the protobuf syntax? If not, then this hook should be removed, as it's not the role of syntax support scripts to decide how the buffer should be formatted.

laelath: It's not strictly forbidden by c/c++ syntax but the c-family syntax files still strip trailing whitespace on empty lines.

lenormf: I know some language support scripts do it, but that is a mistake that was fixed in other scripts, while it stayed in the older ones.

One example why that hook is annoying from a practical point of view is when you want to paste something: <kbd>i</kbd> `int<space>nvar<space>=<space>` <kbd>esc</kbd><kbd>p</kbd> would remove the space after the equal sign.

maximbaz: I raised this as well in IRC and #2175, I think the consensus is that no syntax file should ever touch whitespaces, this should be the job of formatters.

laelath: That sounds like a sensible change, but if I did that now for the protobuf syntax then it's behavior would be inconsistent from the other syntax files until that change gets made.

maximbaz: Yep, but that would be the problem of the other syntax files, not protobuf syntax. In other words, if someone is annoyed by inconsistency they would submit a PR for the other syntax files.

laelath: Should the cleanup of whitespace following a newline also be removed from the syntax files?

laelath: Also I feel like something like that should be done all at once for every syntax file with that behavior if this change is going to be made to avoid inconsistency.

maximbaz: > Should the cleanup of whitespace following a newline also be removed from the syntax files?

In my mind we should only keep the hooks that help you _type_, things like increase indent after typing `{` and decrease indent after typing `}`. But there should be no hooks that post factum reformat my code — if I added whitespaces after newline, so let them be.

> Also I feel like something like that should be done all at once for every syntax file with that behavior if this change is going to be made to avoid inconsistency.

I will leave this to you and others to decide, I already expressed my opinion 🙂

laelath:
Any reason this hasn't been merged? It's just a syntax support script.

PR #2255, opened 2018-07-25
Showcase lazy-loading for c-family.kak

maximbaz:
<details>
<summary> Append <code>?w=1</code> to the URL before you review the code for your own sanity. </summary>

image

</details>


lenormf:
"Lazy loading" could be as easy as calling a wrapper to the `source` command, I don't think splitting up code in functions is the best way to implement this.

maximbaz:
Closing, will do this for all files if I get some support in https://github.com/mawww/kakoune/issues/2152#issuecomment-408599318

PR #2256, opened 2018-07-25
Only show trailing spaces in show_whitespace

divarvel:
This PR adds a switch to the `show_whitespace` to only show trailing whitespaces.
- [x] only display trailing spaces when instructed
- [x] make it configurable by the user

This is my first PR on kakoune, don't hesitate to tell me if I'm doing suboptimal things:

Also, if this has no business being in kakoune core, no problem :-)
I know something approaching can be done with custom highlighters by blending the non-trailing space chars in the background, but I felt this is a common enough use case to be directly supported without the need of a brittle workaround.

### DIFF
--- a/src/highlighters.cc
+++ b/src/highlighters.cc
@@ -958,9 +958,9 @@ struct TabulationHighlighter : Highlighter
 
 struct ShowWhitespacesHighlighter : Highlighter
 {
-    ShowWhitespacesHighlighter(String tab, String tabpad, String spc, String lf, String nbsp)
+    ShowWhitespacesHighlighter(String tab, String tabpad, String spc, String lf, String nbsp, bool only_trailing)
       : Highlighter{HighlightPass::Move}, m_tab{std::move(tab)}, m_tabpad{std::move(tabpad)},
-        m_spc{std::move(spc)}, m_lf{std::move(lf)}, m_nbsp{std::move(nbsp)}
+        m_spc{std::move(spc)}, m_lf{std::move(lf)}, m_nbsp{std::move(nbsp)}, m_only_trailing{std::move(only_trailing)}
     {}
 
     static std::unique_ptr<Highlighter> create(HighlighterParameters params, Highlighter*)
@@ -970,11 +970,13 @@ struct ShowWhitespacesHighlighter : Highlighter
               { "tabpad", { true, "" } },
               { "spc", { true, "" } },
               { "lf", { true, "" } },
-              { "nbsp", { true, "" } } },
+              { "nbsp", { true, "" } },
+              { "only-trailing", { false, "" } } },
             ParameterDesc::Flags::None, 0, 0
         };
         ParametersParser parser(params, param_desc);
 
+        bool only_trailing = (bool) parser.get_switch("only-trailing");
         auto get_param = [&](StringView param,  StringView fallback) {
             StringView value = parser.get_switch(param).value_or(fallback);
             if (value.char_length() != 1)
@@ -984,7 +986,7 @@ struct ShowWhitespacesHighlighter : Highlighter
 
         return std::make_unique<ShowWhitespacesHighlighter>(
             get_param("tab", "→"), get_param("tabpad", " "), get_param("spc", "·"),
-            get_param("lf", "¬"), get_param("nbsp", "⍽"));
+            get_param("lf", "¬"), get_param("nbsp", "⍽"), only_trailing);
     }
 
 private:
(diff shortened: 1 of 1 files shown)


lenormf:
If this PR is to be merged, you'll need a copyright waiver!

divarvel:
Right, I forgot. Should be okay now

src/highlighters.cc
```diff
@@ -992,7 +992,7 @@ void show_whitespaces(HighlightContext context, DisplayBuffer& display_buffer, B
                                                   std::max(win_column - column, 0_col);
                         atom_it->replace(tab + String(tabpad[(CharCount)0], count - tab.column_length()));
                     }
-                    else if (cp == ' ')
+                    else if (cp == ' ' && (!only_trailing || it == end - 1))
```
Delapouite: Small nitpicking: if I remember correctly, mawww uses English operators like `not`, `and`, `or`…

maximbaz:
Nice job @divarvel! It doesn't work as expected just yet, it shows only one space, not all trailing spaces:

image

divarvel:
I've implemented this way for now as it doesn't require backtracking in / traversing twice the line iterator (also I find the output to be less crowded while still showing that there are trailing spaces).

I can give a try at traversing the line iterator once to get the last non-space char position.

src/highlighters.cc
```diff
@@ -1013,11 +1013,14 @@ std::unique_ptr<Highlighter> show_whitespaces_factory(HighlighterParameters para
...
           { "lf", { true, "" } },
-          { "nbsp", { true, "" } } },
+          { "nbsp", { true, "" } },
+          { "only-trailing", { false, "" } } },
         ParameterDesc::Flags::None, 0, 0
     };
     ParametersParser parser(params, param_desc);
 
+    // ToDo: I think there must be a better way to transform it into a boolean
+    bool only_trailing = parser.get_switch("only-trailing") ? true : false;
```
maximbaz: try this: `bool only_trailing = (bool)parser.get_switch("only-trailing")`

maximbaz: and I'd get rid of the variable and just move the `(bool)parser.get_switch("only-trailing")` directly to the function call on line 1037

maximbaz:
Output should be crowded, so you feel the need to cleanup these trailing whitespaces 😉

src/highlighters.cc
divarvel: :ok_hand:

divarvel:
If the double traversal is no issue, I'll have a go at it later today

divarvel:
I've fixed the description as well

divarvel:
@maximbaz done :-)

maximbaz:
Good stuff! But why does this logic only works for spaces, and not for other characters? I basically want to see only trailing whitespace, no other spaces, and I was hoping to achieve it with this command, but right now I always see tabs, even non-trailing ones:

```
:add-highlighter global/ show-whitespaces -lf ' ' -only-trailing 
```

trailing-tab

divarvel:
I could extend it to spaces and nbsp. I went with the assumption that if
spaces are to be displayed, it's always, whereas I just want trailing
spaces to be displayed.
I can add a switch to give tabs the same treatment

maximbaz:
I think one switch should be enough, if it says "only trailing", it should really show only trailing whitespaces and nothing else.

divarvel:
The thing is, I do want to have all tabs displayed, but only trailing spaces.
You want to only display trailing hidden chars (what about nbps?).
Is it possible to have both behaviours with a single flag?

maximbaz:
You should use something like this:

```
# show trailing whitespaces of any kind
:add-highlighter global/ show-whitespaces -only-trailing

:add-highlighter global/ show-whitespaces -spc ' ' -nbsp ' ' -lf ' '
```

divarvel:
I see, that would indeed solve the issue

divarvel:
It works well this way :-)

maximbaz:
It's just perfect now, thanks a lot!

divarvel:
(I can squash commits if needed)

src/highlighters.cc
```diff
@@ -971,29 +971,43 @@ void show_whitespaces(HighlightContext context, DisplayBuffer& display_buffer, B
...
                 if (cp == '\t' or cp == ' ' or cp == '\n' or cp == 0xA0)
                 {
+                    auto should_highlight = !only_trailing or it.coord() > last_non_space;
                     if (coord != begin.coord())
                         atom_it = ++line.split(atom_it, coord);
                     if (it != end)
                         atom_it = line.split(atom_it, it.coord());
 
-                    if (cp == '\t')
+                    if (cp == '\t' and should_highlight)
```
maximbaz: Maybe wrap this whole `if - else if - else if` block in another `if (should_highlight)` to avoid duplicating this condition all the time

divarvel:
Okay, if that's the case I'll update the issue and this PR to point to a how-to with replace-regex.

utdemir:
Hey. I want to use this features, can you point me to how to do this with `replace-regex`?


Thank you :).

divarvel:
Closing this PR, but i'd like to know how to achieve that with `replace-regex`

CharlesGueunet:
Hello, any new on this PR ?

shrit:
Exactly the same question here is it good to merge?

ilserd:
@mawww same as above.

mawww:
Well, that was a pretty long wait, but thanks for the PR @divarvel

divarvel:
Nice! I'm glad it's merged, thanks for handling it!

PR #2265, opened 2018-07-28
Do not enforce autoformat in ocaml

### LINKED ISSUE #2264: Newlines get deleted in .ml files
Create a file, ending in .ml, then try to insert newlines -- all lines, including first get deleted

For instance:
execute-keys `":e a.ml<ret>ix<ret>y<ret>z<ret>"`

maximbaz:
This is what `:format` is for, let the user decide when and how they want to format their code.

Fixes #2264


Delapouite:
Thanks. I've updated the wiki page to add ocp-indent: 
https://github.com/mawww/kakoune/wiki/Languages

PR #2268, opened 2018-07-30
rc lint: Create temporary file with the same buffer extension

### LINKED ISSUE #2189: Problem with ruby lint with rubocop
This is my first issue in the project. I like a lot Kakune, an awesome job!

### Steps
In the `kakrc` file I configure the ruby lint with rubocop in this way:

```
hook global WinSetOption filetype=ruby %{
  set window lintcmd "rubocop"
  lint-enable
}
````

Then I open the file `app/models/user.rb` (that has a lot of offenses) and execute the command `:lint`

### Outcome

The output format is not valid to view the offenses in the file buffer, but when I open the `*lint-output*` buffer I can see:

```
app/models/user.rb:0 files inspected, no offenses detected
app/models/user.rb:Inspecting 0 files
```

This message is like if rubocop doesn't find any file.

### Expected

I expected to get the errors in the `*lint-output*` buffer like I see when I execute the same command in the terminal for that file.

```
rubocop app/models/user.rb
```

### PR DESCRIPTION
**Pull request by lenormf**
Fixes #2189

### DIFF
--- a/rc/base/lint.kak
+++ b/rc/base/lint.kak
@@ -15,9 +15,14 @@ define-command lint -docstring 'Parse the current buffer with a linter' %{
             exit 1
         fi
 
+        extension=""
+        if printf %s "${kak_buffile}" | grep -qE '[^/.]\.[[:alnum:]]+$'; then
+            extension=".${kak_buffile##*.}"
+        fi
+
         dir=$(mktemp -d "${TMPDIR:-/tmp}"/kak-lint.XXXXXXXX)
         mkfifo "$dir"/fifo
-        printf '%s\n' "evaluate-commands -no-hooks write -sync $dir/buf"
+        printf '%s\n' "evaluate-commands -no-hooks write -sync $dir/buf${extension}"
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: rc/base/lint.kak [hunk 1]
```diff
@@ -15,9 +15,15 @@ define-command lint -docstring 'Parse the current buffer with a linter' %{
             exit 1
         fi
 
+        basename="${kak_buffile##*/}"
```
mawww: Do we need this basename ? I was hoping for a oneliner to get the extension.

lenormf: You thought of simply using `${kak_buffile##*.}`? So did I, but that doesn't work well with dotfiles (e.g. `.config`), is there a simpler way of handling that?

Clyybber:
Thank you!
This also makes linting nim files with this: https://gist.github.com/Clyybber/8861110a55f63ef1c255b26dbabfec1c much more simpler since nim refuses to compile files without an extension, so I had to copy the temporary buffer before linting.

maximbaz:
And `go vet`, had to use the same workaround 😉

PR #2269, opened 2018-07-30
Remove whitespace stripping from all rc files

laelath:
This pull request implements the changes proposed by @lenormf and @maximbaz to remove the stripping of whitespace from the syntax scripts. I'm not sure if this should get merged as is, this is mostly to see the proposed new default behavior of kakoune.

### DIFF
--- a/rc/base/clojure.kak
+++ b/rc/base/clojure.kak
@@ -22,7 +22,6 @@ add-highlighter shared/clojure/ regex \b(clojure.core/['/\w]+)\b 0:keyword
 # Commands
 # ‾‾‾‾‾‾‾‾
 
-define-command -hidden clojure-filter-around-selections lisp-filter-around-selections
 define-command -hidden clojure-indent-on-new-line       lisp-indent-on-new-line
 
 # Initialization
@@ -30,7 +29,6 @@ define-command -hidden clojure-indent-on-new-line       lisp-indent-on-new-line
 hook -group clojure-highlight global WinSetOption filetype=clojure %{ add-highlighter window/clojure ref clojure }
 
 hook global WinSetOption filetype=clojure %[
-    hook window ModeChange insert:.* -group clojure-hooks  clojure-filter-around-selections
     hook window InsertChar \n -group clojure-indent clojure-indent-on-new-line
 ]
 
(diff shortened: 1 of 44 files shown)


maximbaz:
Here's why I like this PR:

1. No logic duplication all over the codebase (and notice the code itself is slightly different in every file, which makes it harder to track).
1. If you are writing code, you are probably using `:format`, which probably removes trailing whitespaces anyway.
1. Even if you don't use `:format`, configuring a command that strips trailing whitespaces is super easy, and you can choose to execute it on a key press, on buffer save or even when entering normal mode.
1. No more annoying inconsistency in cursor position, like in this example:
    1. Open a python file.
    1. Enter insert mode, type `def fn():<ret>`
    1. Exit to normal mode because you need to think what to write and your muscle memory is forcing your hand to press Esc
    1. Press `i` to continue typing — in `master` you are typing in a wrong position, but in this PR the cursor position is correct.

lenormf:
It looks like the hooks removed in this PR get rid of extra whitespace that would stay behind as a result of the indentation hook (e.g. after hitting the return key twice in a row) - which is something we would want to keep.

What we don't want is scripts formatting the *entire* buffer (using a formater or `%s\h+<ret>d`), but the hunks removed in this patch only modify the previous line to remove artifacts.

laelath:
Ahh it seems like I was confused about what you were proposing.

lenormf:
The problem with your original protobuf PR is that there was a hook that unconditionally removed trailing whitespace on the current line upon leaving insert mode, whereas the hooks removed here compensate for an artifact created after the indentation hook is run.

laelath:
Should that behavior be removed from the scripts that currently do that? I can work on those changes in another pull request.

maximbaz:
Yeah, I think we have different ideas, and I'm in the minority here.

If not remove, I wish there was at least a way to disable all these whitespace hooks, the example I described above with python is annoying 😞

laelath:
I may be wrong but I believe that that is one of the things that @lenormf wants to remove, as that would fall under "unconditionally removed trailing whitespace on the current line upon leaving insert mode".

maximbaz:
Huh, I thought all of them do that 🙂 In that case @lenormf I would like to know which syntax file currently shows this correct behavior that you want to keep.

PR #2273, opened 2018-07-31
Removed whitespace stripping when leaving insert mode

laelath:
Second attempt at implementing @lenormf's proposed changes across all rc files.


maximbaz:
Now I see what you meant, my example of an annoying behavior with `python` no longer reproduces on this branch (i.e. exiting insert mode doesn't clear my precious indentation), but `<ret><ret><ret>` still doesn't leave two lines full of trailing spaces as you guys wanted. Nice!

laelath:
@lenormf, what are your thoughts on these changes?

PR #2275, opened 2018-08-01
rc systemd: Make sure files are in the `systemd` directory

### LINKED ISSUE #2272: Fix comment char in *.conf files
Since https://github.com/mawww/kakoune/pull/2246/ all my *.conf files started having a wrong comment char when I use `:comment-line`: `;` instead of `#`. This happened because `.conf` is now considered a systemd file bound to `ini` filetype.

There are no rules as to how `*.conf` file is formatted, some use `ini` style, some don't. But one thing I know is that all of them use `#` as a comment char.

I'm not sure how this should be handled.

`nvim` has a dedicated `conf` syntax, maybe that's what we should do.

We could also just force `#` for all `ini` files by removing `|ini` from this line, since we bound so many different file types to `ini`:

https://github.com/mawww/kakoune/blob/1c0cc61ccc1e9bdbfbe3bafb1d2470f38f2fe21b/rc/core/comment.kak#L52-L54

This is what I did locally.

@lenormf what do you think?

-----------

Some of `.conf` files from my own dotfiles:

| file | looks like `ini` | comment char | contents | syntax detected by my `nvim` |
|-|-|-|-|-|
| `.tmux.conf` | no | `#` | full-blown tmux commands | `tmux` |
| `compton.conf` | yes | `#` | equals-separated entries with semicolon | `conf` |
| `.i3status.conf` | no | `#` | kinda like ini, but has `{}` blocks and `+=` | plain text |
| `gpg.conf` | no | `#` | gpg commands | `gpg` |
| `pacman.conf` | yes | `#` | normal `ini` file | `conf`|

lenormf:
Fixes #2272

### DIFF
--- a/rc/extra/systemd.kak
+++ b/rc/extra/systemd.kak
@@ -6,4 +6,9 @@
 
 hook global BufCreate .*/systemd/.+\.(automount|conf|link|mount|network|path|service|slice|socket|target|timer) %{
     set-option buffer filetype ini
+
+    # NOTE: INI files define the commenting character to be `;`, which won't work in `systemd` files
+    hook -once buffer BufSetOption comment_line=.+ %{
+        set-option buffer comment_line "#"
+    }
 }


maximbaz:
Not exactly fixes, have a look in /etc/systemd files, they all use `#` as comment char...

lenormf:
But `ini.kak` highlights comments when a line starts with `[;#]`, shouldn't that work?

maximbaz:
Oh sorry, I realized that I didn't include the most valuable piece of information in my bug description: #2272 is first and foremost about `:comment-line` helper, I'm not talking about highlighting, I'm talking about which comment char gets inserted when I comment a line.

lenormf:
That should work, but that could be made sexier with a `-oneshot` flag to `hook`, to allow the hook to run once (and avoid declaring a group).

lenormf:
Modified to use `-once` on the `hook` command, rebased on master.

PR #2279, opened 2018-08-02
rust.kak: Support Rust's raw-string syntax.

Screwtapello:
According to the [Rust language reference][1], a raw string starts with an 'r', zero or more '#' characters, and a '"', and doesn't close until a '"' is immediately followed by the matching number of '#' characters.

[1]: https://doc.rust-lang.org/reference/tokens.html#raw-string-literals


lenormf:
> followed by the matching number of '#' characters

Could you use the `-match-capture` flag in your code, to match the exact same number of `#` characters that were used at the beginning and end of the raw string? That's how here-docs are handled in [`sh.kak`](https://github.com/mawww/kakoune/blob/master/rc/core/sh.kak#L10).

PR #2280, opened 2018-08-02
 src: Stay in user-mode when an unmapped key is hit in lock-mode

lenormf:


lenormf:
Bump!

lenormf:
PING

Rosuavio:
What is holding up this PR, this is something I want, also some would consider it a bug that the this is not implemented as the docs says `stay in mode until <esc> is pressed`.

Is there something I can do to help move this along?

Rosuavio:
If this were fixed I believe I could use a user mode instead of recursive command. It maybe be similar to what happened in https://github.com/mawww/kakoune/issues/4735. It seems like @theowenyoung as going to use a user mode too, but could not because `-lock` was not working as expected.

mawww:
Thanks for bumping that, not sure why I did not merge it back in the days.

PR #2281, opened 2018-08-02
src: Allow hooks to be run only once

### LINKED ISSUE #2277: Implement a `-oneshot` flag for the `hook` command
Hi,

It'd be convenient to turn the following pattern:

```
hook -group mygroup global … %{
    # do something
    rmhooks global mygroup
}
```

into:

```
hook -oneshot global … %{
    # do something
}
```

HTH.

lenormf:
This commit implements the -once flag on the `:hook` command, which
automatically removes a hook after it was run, to avoid having to
declare a group and remove it in the hook implementation.

Closes #2277

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -842,7 +842,8 @@ const CommandDesc add_hook_cmd = {
     "  * window: hook is executed only for the current window\n",
     ParameterDesc{
         { { "group", { true, "set hook group, see remove-hooks" } },
-          { "always", { false, "run hook even if hooks are disabled" } }},
+          { "always", { false, "run hook even if hooks are disabled" } },
+          { "once", { false, "run the hook only once" } } },
         ParameterDesc::Flags::None, 4, 4
     },
     CommandFlags::None,
@@ -860,9 +861,9 @@ const CommandDesc add_hook_cmd = {
         Regex regex{parser[2], RegexCompileFlags::Optimize};
         const String& command = parser[3];
         auto group = parser.get_switch("group").value_or(StringView{});
-        get_scope(parser[0], context).hooks().add_hook(parser[1], group.str(),
-                                                       parser.get_switch("always") ?
-                                                       HookFlags::Always : HookFlags::None,
+        const auto flags = (parser.get_switch("always") ? HookFlags::Always : HookFlags::None) \
+                            | (parser.get_switch("once") ? HookFlags::Once : HookFlags::None);
+        get_scope(parser[0], context).hooks().add_hook(parser[1], group.str(), flags,
                                                        std::move(regex), command);
     }
 };
--- a/src/hook_manager.cc
+++ b/src/hook_manager.cc
@@ -63,7 +63,7 @@ CandidateList HookManager::complete_hook_group(StringView prefix, ByteCount pos_
     return res;
 }
 
-void HookManager::run_hook(StringView hook_name, StringView param, Context& context) const
+void HookManager::run_hook(StringView hook_name, StringView param, Context& context)
 {
     const bool only_always = context.hooks_disabled();
 
@@ -99,7 +99,7 @@ void HookManager::run_hook(StringView hook_name, StringView param, Context& cont
     for (auto& hook : hook_list->value)
     {
         MatchResults<const char*> captures;
-            if ((not only_always or (hook->flags & HookFlags::Always)) and
+        if ((not only_always or (hook->flags & HookFlags::Always)) and
             (hook->group.empty() or disabled_hooks.empty() or
              not regex_match(hook->group.begin(), hook->group.end(), disabled_hooks))
             and regex_match(param.begin(), param.end(), captures, hook->filter))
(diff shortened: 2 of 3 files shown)


src/hook_manager.cc
```diff
@@ -123,6 +123,17 @@ void HookManager::run_hook(StringView hook_name, StringView param, Context& cont
...
+            if (to_run.hook->flags & HookFlags::Once)
+            {
+                auto it = std::remove_if(hook_list->value.begin(), hook_list->value.end(),
+                                       [&](const std::unique_ptr<Hook>& h)
+                                       { return h.get() == to_run.hook; });
+
+                m_hooks_trash.insert(m_hooks_trash.end(), std::make_move_iterator(it),
+                                     std::make_move_iterator(hook_list->value.end()));
+                hook_list->value.erase(it);
+            }
```
lenormf: Please pay extra attention to this hunk, it seems to work but I'm not confident it's what's needed to properly remove a hook from the hooks list.

mawww: This is a strange way to do it, it seems correct, but is more complex than necessary.

* `std::remove_if` will move all the hooks to remove at the end of the container and return an iterator to the first one of those "removed" elements. It is strange to use it because we should have exactly one element to remove, so just using `find_if(hook_list->value, ...);` should be enough. 
* m_hooks_trash.insert is unnecessary, use push_back or emplace_back, as we have a single element, and just use `std::move(*it)` to move the hook out.

src/commands.cc
```diff
@@ -857,12 +858,17 @@ const CommandDesc add_hook_cmd = {
         if (not contains(hooks, parser[1]))
             throw runtime_error{format("no such hook: '{}'", parser[1])};
 
+        const bool flag_always = (bool)parser.get_switch("always");
+        const bool flag_once = (bool)parser.get_switch("once");
+
+        if (flag_always and flag_once)
+            throw runtime_error("The -once and -always flags are mutually exclusive");
```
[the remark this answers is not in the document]
lenormf: I got confused by the naming of the `-always` flag, didn't really think about it when I added this check. Should the `-once` flag be renamed?

src/hook_manager.cc
lenormf: I originally did use `find_if()` and a simple `push_back(std::move(*it));`, but then got weird compilation errors (because I didn't notice straight away this function was `const`!), so I based this bit on `remove_hooks()`.

I'll revert the changes back.

lenormf:
I renamed the `-once` flag into `-oneshot` as originally suggested, and simplified the removal of the hook from the run-list.

mawww:
I think `-once` was pretty good actually, if anything is `-always` that was a confusing choice.

src/commands.cc
```diff
@@ -857,12 +858,14 @@ const CommandDesc add_hook_cmd = {
...
+        const bool flag_always = (bool)parser.get_switch("always");
+        const bool flag_oneshot = (bool)parser.get_switch("oneshot");
+
         Regex regex{parser[2], RegexCompileFlags::Optimize};
         const String& command = parser[3];
         auto group = parser.get_switch("group").value_or(StringView{});
-        get_scope(parser[0], context).hooks().add_hook(parser[1], group.str(),
-                                                       parser.get_switch("always") ?
-                                                       HookFlags::Always : HookFlags::None,
+        const auto flags = flag_always ? HookFlags::Always : (flag_oneshot ? HookFlags::Once : HookFlags::None);
```
mawww:
```
const auto flags = (parser.get_switch("always") ? HookFlags::Always : HookFlags::None) |
                             (parser.get_switch("once") ? HookFlags::Once : HookFlags::None);
```

And no need for flag_always and flag_once variables.

src/commands.cc
```diff
@@ -842,7 +842,8 @@ const CommandDesc add_hook_cmd = {
     "  * window: hook is executed only for the current window\n",
     ParameterDesc{
         { { "group", { true, "set hook group, see remove-hooks" } },
-          { "always", { false, "run hook even if hooks are disabled" } }},
+          { "always", { false, "run hook even if hooks are disabled" } },
+          { "oneshot", { false, "run the hook only once" } } },
```
mawww: I liked `once` better, `oneshot` is less clear to me (whats a shot ?)

lenormf: The term was suggested to me by `systemd` services, whose [type](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Type=) can be `oneshot`.

lenormf:
I thought about renaming `-always`, but after a few seconds of not being able to come up with a replacement, I decided to be lazy and rename `-once` :p

Here are some ideas:

* `-constant`
* `-invariant`
* `-persistent`

maximbaz:
Just to clarify, if you and I are both talking about renaming `-always` to `-ignore-disabled`, then I guess I just don't really understand what `-always` currently does, I came up with this name based on its description: `run hook even if hooks are disabled`

Screwtapello:
In CSS, "this applies even if something else tries to turn it off" is called [important](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#The_!important_exception).

I agree that `-ignore-disabled` doesn't really fit since the hook *manager* is the thing that respects or ignores whether hooks are disabled, not the hook itself. However, `-no-disable` or `-cannot-disable` might work.

maximbaz:
Now I see what you guys mean

lenormf:
I was thinking of single words, but `-no-disable` could work.

Issue #2122, opened 2018-06-13
Insert mode maps always trigger ModeChange hooks


shachaf:
It would be useful to be able to write insert mode maps that use normal mode commands without triggering insert:normal ModeChange hooks (particularly with <a-;>?). For example, the c-family-trim-autoindent hook unavoidably whitespace at end of line.

lenormf:
Use `\`.

shachaf:
`\` is fine for specific uses but I want to do things like `map global insert key '<a-;>: foo<ret>'` that are always in effect when inserting, without e.g. clearing trailing whitespace.

mawww:
My idea regarding this problem is that ModeChange should make it possible to know if the change is temporary (as in `<a-;>` in insert mode) or definitive (as in `<esc>` in insert mode). In other words, we should know if we are changing mode A to B by pushing B on top of A in the mode stack, or by poping A and returning to B. The obvious place to get such an information is in the filter text of the ModeChange hook.

That said, I have been thinking that ModeChange is a failure, and that we should go back to <Mode>Begin/<Mode>End hooks. My reasoning is that:

1. In rc/ files, ModeChange is never used caring both about starting and end mode, only one of them.
2. The moment ModeChange runs in is not always the one we would like (It runs *after* the new mode was activated, so it makes it imposible to execute keys in previous mode context).
3. In general, filtering hooks by regex is more expensive than directly by hook name.

PR #2286, opened 2018-08-05
Added an option for vertical menu layout

laelath:
This is the same pull request as #2212, but from a different branch so the branch name reflects the changes made.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -396,6 +396,7 @@ void register_options()
                        "The ncurses ui supports the following options:\n"
                        "    <key>:                        <value>:\n"
                        "    ncurses_assistant             clippy|cat|dilbert|none|off\n"
+                       "    ncurses_menu_layout           horizontal|vertical\n"
                        "    ncurses_status_on_top         bool\n"
                        "    ncurses_set_title             bool\n"
                        "    ncurses_enable_mouse          bool\n"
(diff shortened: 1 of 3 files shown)


mawww:
Hi, still did not get to review that and trying it, I'd like to experiment a bit with it to see if I think its an improvement or not, I'd be interested in feedback from other users as well. As said on the previous PR, I try to avoid adding options to control such behaviours, I'd rather have a single way menu are displayed and avoid the combinatorial complexity that an option would add (like bugs only happening with vertical layout menu combined with status on top and mouse disabled...)

maximbaz:
#2199 and #2122 both have some upvotes, those should be counted as supporting this idea 😉 

There wasn't any explicit comment yet against this proposal — to everyone, if you cannot stand vertical layout, now is the time to raise your concerns.

lenormf:
The direction of the menus isn't (probably?) such a big deal that it warrants creating an option for it. If it's an improvement, or if it's more consistent with what users expect (shell completion menus), or if nobody really cares in which direction menus are rendered, let's just switch over completely.

The issue with that being that if the majority of users who don't actively participate in discussions about development doesn't like it or finds an issue that we didn't think of, we'll have to revert the changes back, which is always confusing for users who don't re-compile the development version regularly.

laelath:
There is a slight inconsistency in that the scroll bar is still vertical, but the columns appear to scroll sideways. Not sure how important this is though.

laelath:
I've removed the option and switching logic now, so this branch can get merged if it gets decided to go with vertical menus.

maximbaz:
@mawww have you had a chance to try this PR out?

laelath:
I've rebased against the latest master, seems to still work perfectly.

maximbaz:
Woohoo, thanks for merging!

Issue #2276, opened 2018-08-01
Folder completion should remove the trailing slash


maximbaz:
Let's say I want to type `~/bin/demo.sh`.

I type `~/` and can navigate the completions list with `<C-n>`. I scroll until `~/bin/` entry and now in my buffer I have typed `~/bin/`. I want to continue typing the path, but I don't know it exactly. Currently I have to type `/<Backspace` or `<a-Space>` to refresh the list of candidates.

I propose the same approach as neovim (with deoplete plugin) does. The candidates popup has entry `~/bin/`, but when I select it with `<C-n>`, the trailing slash should be removed from my buffer, i.e. I should have inserted just `~/bin`. That way, I can press `/` and get a refreshed list of child files and folders.

Same should apply for completion in prompt.

laelath:
Because once a completion is added to the list it's just a string, the only way to do this without significant changes to the completion system would be to put the directories into the completions list without the slash (so all files in a directory would appear the same). This might be better as the current system assumes that you're always inputting a path to a file rather than a directory, but also isn't as clear as you can't directly tell if the completed path is for a directory or a file.

laelath:
Though I guess that could be done with the docstring for the completion candidate. I'll try and make a pull request to test it out.

PR #2287, opened 2018-08-06
Do not insert slash when selecting a directory filename completion

laelath:
Attempt to implement #2276, though only the behaviour in insert mode could be implemented due to how the completions work.

### DIFF
--- a/src/insert_completer.cc
+++ b/src/insert_completer.cc
@@ -211,7 +211,13 @@ InsertCompletion complete_filename(const SelectionList& sels,
     {
         for (auto& filename : Kakoune::complete_filename(prefix,
                                                          options["ignored_files"].get<Regex>()))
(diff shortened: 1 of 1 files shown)


maximbaz:
Just tried, good job!

maximbaz:
Why closed? :'(

laelath:
I felt like the inconsistent behaviour between insert completion menus and prompt completion menus was not desirable, and I couldn't think of a way to make the prompt completion work without majorly refactoring the way those completions work. Though I guess there's no harm in leaving it open as a future todo.

lenormf:
The insert-mode completion engine already works "correctly", it's just that you'd prefer it worked a different way.

Closing the issue was premature, but breaking consistency of completion across all modes in one of them is not a good idea.

PR #2296, opened 2018-08-10
Update README.asciidoc

Meai1:
much easier to find this when it's under "navigate" like every other IDE calls it


mawww:
I am not sure navigate is a common term for that, the jump list concept comes from vim. Which IDE's use navigate to mean jumping between different position we visited in the same/different files ?

PR #2298, opened 2018-08-11
rc c-family: Show a menu to list alternative files

lenormf:
Hi,

This is a partial re-work of #1483, the `:alt` command will now show a menu if several alternative files are available when the current buffer is a C-family one.

It's still unclear how this functionality should be migrated to other modules (e.g. the LUA one), if it was accepted.


lenormf:
Bump!

PR #2300, opened 2018-08-11
rc lisp: Fix variable highlighting

### LINKED ISSUE #2289: regex problems with add-highlighter .. regex
I have noticed many regex do not catch highlighting with add-highlighter .. regex

### Steps
One example is in base/lisp.kak :

line 21 :  add-highlighter shared/lisp/code/ regex \b(([':]\w+)|([*]\H+[*]))\b 0:variable

### Outcome
the quote symbol is never highlighted as a variable in lisp code.

lenormf:
Fixes #2289


cduret:
not bad but regexp should work also for dash separated names like 'hello-world

cduret:
a reasonable goal would be to have the same highlights as vim

lenormf:
I've added more characters to the "identifier" highlighter, as per the [documentation](http://www.r6rs.org/final/html/r6rs/r6rs-Z-H-7.html#node_sec_4.2.4).

I've also added a highlighter for numbers.

Could you please verify that your code is highlighted properly?

PR #2303, opened 2018-08-14
Add Dart rc file

haarts:
Based heavily on `go.kak`.

### DIFF
--- a/rc/extra/dart.kak
+++ b/rc/extra/dart.kak
@@ -1,0 +1,108 @@
+# https://dartlang.org/
+#
+
+# Detection
+# ‾‾‾‾‾‾‾‾‾
+
+hook global BufCreate .*\.dart %{
+    set-option buffer filetype dart
+}
+
+# Highlighters
+# ‾‾‾‾‾‾‾‾‾‾‾‾
+
+add-highlighter shared/dart regions
+add-highlighter shared/dart/code default-region group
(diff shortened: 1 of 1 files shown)


rc/extra/dart.kak
```diff
@@ -0,0 +1,106 @@
...
+
+    # Highlight keywords
+    printf %s "
+        add-highlighter shared/dart/code/ regex \b(${keywords})\b 0:keyword
+        add-highlighter shared/dart/code/ regex \b(${attributes})\b 0:attribute
+        add-highlighter shared/dart/code/ regex \b(${types})\b 0:type
+        add-highlighter shared/dart/code/ regex \b(${values})\b 0:value
+        add-highlighter shared/dart/code/ regex \b(${functions}) 2:function
+        add-highlighter shared/dart/code/ regex \b(${annotations})\b 0:meta
+        add-highlighter shared/dart/code/ regex \b(${classes})\b 0:module
```
lenormf: Some symbols that start/end with a punctuation sign won't be highlighted because of the `\b` escape sequence in the regex.

Example: `\bsync\*\b`, the second `\b` sequence will trigger a match-miss.

haarts: Great catch! I've amended my commit.

lenormf: Actually you'll have the same issue with annotations (start with `@`), and functions that start with `_` or end with `(`/`>`.

PR #2308, opened 2018-08-16
docs: add -always and -once switch on hooks page

Delapouite:


lenormf:
We should probably rename `-always` first, as we determined that it was confusing.

Delapouite:
Right. I'll rebase after this PR is merged: https://github.com/mawww/kakoune/pull/2308

mawww:
Too late, we'll fix it once/if we switch to the new name.

PR #2309, opened 2018-08-17
 src: Rename the `hook` -always flag into -no-disabled

lenormf:


lenormf:
Rebased on `master`.

mawww:
I am willing to merge that, but I think `-no-disabled` is not really self-explanatory, I would prefer `-not-disableable` which seems much clearer to me, would you agree ?

lenormf:
C.f. #2281, I think we settled on `no-disabled` as it's fairly short and explanatory at the same time.

Screwtapello:
From #2281, apparently I'm the one that suggested "no disable". I guess I'm not the best at remembering things. :/

I think "no disable" is better than "no disabled", though: in the documentation the act is generally referred to as "disable hooks" rather than "disabled hooks". The actual `eval` flag is `-no-hooks`, though, and `-no-no-hooks` would be a bit silly. Perhaps `eval`'s flag could be renamed to `-disable-hooks` for consistency? ;)

PR #2322, opened 2018-08-21
Support user map options

Screwtapello:


Screwtapello:
I want this feature so I can use it with ul/kak-lsp#86.

lenormf:
I think `str-map` is a less cumbersome type name, even though it's not as specific as `str-to-str-map` can be. There's no immediate need for any other `str-to-*-map` option type, so we can afford to take this shortcut.

Screwtapello:
I picked `str-to-str-map` because that's what the existing documentation suggested. I don't mind *too* much that it's cumbersome, because I don't intend to type it very often, but I'd be more than happy to change it to something else if there's general agreement.

PR #2336, opened 2018-08-25
rc lint formatter: Let users handle shell commands fully

lenormf:
Hi.

It looks like users are more comfortable handling the commands that format/lint code directly in their configuration file, rather than drop a wrapper in their `$PATH` to work around limitations set by the lint/formatter scripts.

This PR exports variables in the environment of the command that is run during linting/formatting, which makes the `lintcmd`/`formatcmd` options uglier, but at the same time easier to use.

Comments?

### DIFF
--- a/rc/core/formatter.kak
+++ b/rc/core/formatter.kak
@@ -1,4 +1,10 @@
-declare-option -docstring "shell command to which the contents of the current buffer is piped" \
+declare-option -docstring %{
+    shell command running an external formatting tool to format the current buffer
+
+The following environment variables are available:
+  - format_file_in: path to a copy of the current buffer
+  - format_file_out: path to the file that should be filled with the buffer's formatted data
+    } \
     str formatcmd
 
 define-command format -docstring "Format the contents of the current buffer" %{ evaluate-commands -draft -no-hooks %{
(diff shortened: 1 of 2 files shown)


Clyybber:
I think is great, much more robust then the current system. Since we now output to a file, are multiline errors/warnings possible?

lenormf:
The only thing that changed is how the formatter/linter is being invoked, the output was always redirected to a file.

Multi-line diagnostics would have to be handled in a separate issue (you can create one if it doesn't exist already).

lenormf:
Bump!

lenormf:
With [https://github.com/hadolint/hadolint](hadolint), users have the possibility to invoke the linter from a Docker container:

```
$ docker run --rm -i hadolint/hadolint < Dockerfile
```

To get the above to work, `lintcmd` would have to be set to `docker run --rm -i hadolint/hadolint <`, which partially leaks shell syntax into a Kakoune option.

lenormf:
PING

rc/core/formatter.kak
```diff
@@ -11,7 +17,10 @@ define-command format -docstring "Format the contents of the current buffer" %{
                 evaluate-commands %sh{
                     readonly path_file_out=\$(mktemp \"${TMPDIR:-/tmp}\"/kak-formatter-XXXXXX)
 
-                    if cat \"${path_file_tmp}\" | eval \"${kak_opt_formatcmd}\" > \"\${path_file_out}\"; then
```
krobelus, line 14: Not sure about `lintcmd`, but I think `formatcmd` should stay. I can't think of a formatter that can't use stdin/stdout, it's the obvious interface.

Sometimes we need to pass additional information to the formatter, like

```kak
hook global WinSetOption filetype=(c|cpp|java) %{
	set-option window formatcmd "clang-format -style=file -assume-filename=%val{buffile}"
}
```

but that's independent of this change

lenormf: The point of this PR is to avoid having the editor assume that whatever tool the user might pick will support being called in the way prescribed in `.kak` scripts.

Even though your remark is accurate, it also applies to linters (most of them certainly support input on the standard stream), and it's the same assumption (that is already being made by the editor at the moment) that led to my making the PR.

I think it's better to be consistent.

lenormf:
Rebased on master.

I also re-worked the PR to be more lenient with regards to how the standard output and error streams are handled.

Users would still need to add some `"$lint_in"` to the configuration's `lintcmd`, but in some cases it also allows them to drop the `run() {…} && run` hack.

lenormf:
I also dropped the modifications to `format.kak`, it doesn't look like there is much to be gained by changing that file, but this paradigm could be ported over for consistency later on.

PR #2340, opened 2018-08-28
Speed up wrapping at word boundaries.

### LINKED ISSUE #2339: "wrap -word" highlighter is slow on very long lines.
### Steps 

1. I'm using the following Kakoune:

         version: v2018.04.13-469-g373858f9
         pid: 15167
         session: 15167
         build: release

1. Create a file with a very long line

         python3 -c 'f = open("longline.txt", "w"); all(f.write("a") for _ in range(1_000_000)); f.write("\n")'

1. Open the file in Kakoune, with the `wrap -word` highlighter enabled.

        kak -n -e "addhl global/ wrap -word" longline.txt

### Outcome 

Kakoune hangs for a very long time while drawing the screen; on my machine, it's about 20 seconds.

Running `sudo perf top` at the same time suggests most of the time is being spent in `Kakoune::WrapHighlighter::next_split_pos` and `Kakoune::utf8::read_codepoint<Kakoune::utf8::InvalidPolicy::Pass, char const*>`.

### Expected 

Editing this file, even with line wrapping, should be fast, as with Vim:

    vim -u NONE -U NONE -c "set linebreak" longline.txt

Screwtapello:
Previously, when wrapping lines at word boundaries, we would iterate forwards for "wrap-width" characters, then iterate backwards until we found a word-break, which was horribly slow.

Now we record the last word-boundary we saw as we iterate forwards, getting a result in one pass.

Fixes #2339.

### DIFF
--- a/src/highlighters.cc
+++ b/src/highlighters.cc
@@ -822,6 +822,8 @@ struct WrapHighlighter : Highlighter
         StringView content = buffer[line];
 
         SplitPos pos = current;
+        SplitPos last_boundary = {0, 0};
+
         while (pos.byte < content.length() and pos.column < target_column)
         {
             if (content[pos.byte] == '\t')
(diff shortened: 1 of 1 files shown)


aver-d:
Thank you. Another nice thing is that this would fix the left margin occasionally being indented by one space character.  

That is, depending on the width, currently a wrapped line could be displayed as the second line here:

```
Lorem ipsum dolor sit amet,
 consectetur adipiscing elit, 
sed do eiusmod tempor incididunt…
```

mawww:
@aver-d The problem there is where to put that space ? Shall kakoune split before `amet,`, or should that space end up hidden past the window end ?

aver-d:
I think the split before `amet` (leaving its preceding space at the end of the current display line) would seem right for Kakoune, primarily because this trailing space character is still easily selectable. Screwtappello looks to have used that approach here.

Besides, having a slight one-char gap on the right side is perhaps a good thing anyway given most texts are written with some sort of padding for ease of reading.

Screwtapello:
I confess I didn't really study how the old code chose break-points, I just played around until things lined up correctly in `-indent` mode. In particular, that means whitespace following a word will always be on the same line, even if it's a long span of whitespace. For example, let's say we have a line of text with a span of whitespace characters, and the wrap column happens to fall inside that span:

    some text with a          lot of spaces

With this patch, `a          ` doesn't fit on the first line, so it gets wrapped to the second:

    some text with
    a          lot of
    spaces

Ideally, the extra spaces after `a` would be chopped off so that `a` could stay on the first line and `lot` would get wrapped to the second.

    some text with a
    lot of spaces

Unfortunately, the wrap highlighter doesn't seem to provide a way to skip characters in the input line... or at least, not in the `next_split_pos` method that I was looking at.

PR #2343, opened 2018-08-28
Source with args

occivink:
patch by @shachaf (thanks!), I'm mostly interested in the source-with-args part but having scoped colorscheme seems useful too.

### DIFF
--- a/doc/pages/commands.asciidoc
+++ b/doc/pages/commands.asciidoc
@@ -67,8 +67,9 @@ command *q!* has to be used). Aliases are mentionned below each commands.
 *rename-buffer* <name>::
     set current buffer name
 
-*source* <filename>::
+*source* <filename> <param>...::
     execute commands in <filename>
+    parameters are avilable in the sourced script through the `arg` values
 
 == Clients and Sessions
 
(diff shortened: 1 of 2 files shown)


occivink:
My usecase for it is this [snippet](https://github.com/occivink/config/blob/master/.config/kak/kakrc#L11-L22) that searches parent directory for `init.kak` files, it's how I define project-specific things. However, the script being source doesn't what is its path, and it's harder to reference relative paths in it.

lenormf:
Why not just create a command that wraps `source`, and updates an option that contains the path of the current file being `source`'d ? It won't work for recursive sourcing, but it's cleaner.

occivink:
I actually think that'd be less clean, but it's a solution too I suppose.

shachaf:
I'm not so sure at this point that this is the right way to do colorscheme scoping -- that change should probably be separate at any rate.

mawww:
I think both changes are worthwhile, but I'd like them split, passing argument to source is pretty simple and can open loads of interesting capabilities, so I'd be happy to merge that.

occivink:
Ok, I only kept the core change in the branch and added some doc. I'll create another PR for scoped colorschemes if this is merged (or let schachaf do it).

PR #2349, opened 2018-08-31
Clojure improvements

eraserhd:
* Supports .cljc .cljx as Clojure
* Syntax is no longer `ref lisp` (before, it was highlighting all symbols as builtin or something weird).
* Highlights everything in `clojure.core/`, some as keywords, other as functions, some as variables
* Adds everything to static_words for completion.
* Sets `extra_word_chars` (there's probably more to add here, but this works pretty nicely).
* Includes copyright waver

It does use `sed` in two places to escape special characters.  Let me know if that's a problem.

screen shot 2018-08-31 at 4 02 01 pm

### DIFF
--- a/rc/base/clojure.kak
+++ b/rc/base/clojure.kak
@@ -6,18 +6,151 @@
 # Detection
 # ‾‾‾‾‾‾‾‾‾
 
-hook global BufCreate .*[.](cljs?) %{
+hook global BufCreate .*[.](clj|cljc|cljs|cljx) %{
     set-option buffer filetype clojure
 }
 
 # Highlighters
 # ‾‾‾‾‾‾‾‾‾‾‾‾
 
-add-highlighter shared/clojure group
+add-highlighter shared/clojure regions
+add-highlighter shared/clojure/code default-region group
+add-highlighter shared/clojure/comment region '(?<!\\)(?:\\\\)*\K;' '$'                 fill comment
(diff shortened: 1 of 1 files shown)


eraserhd:
Oh, additional note.  I'm not sure how to make the region openings for comments and quotes perfect.  These should be mostly good enough.  For example:

`; hello` is a comment (works)
`\; hello` is a character (works)
`\\; hello` is a character followed by comment (doesn't work)
`\\\; hello` is two characters and no comment (doesn't work)

It seems like it needs quantifiers in negative look-behinds to make this work.  If there's a way, let me know.

mawww:
@eraserhd For comments and quotes, wouldnt `(?<!\\)(\\\\)*;` work (`;` optionally preceeded by a **pair** amount of backslashes) ?

eraserhd:
@mawww It would kind of work, but since the doubled backslashes with the quantifier aren't a look-behind, they would be highlighted as comment and not characters.

It's better though, since it can't get out of sync.  I'll update it.

Wait, can I add "preceding" characters to the comment region?  I'm not sure how to match "only at the beginning"...  I'll play with it some more this weekend.

mawww:
`\K` might be of use there.

mawww:
Could you squash those commits together (except for the copyright waiver one that should be a separate commit) ?

eraserhd:
@mawww `\K` works, commits squashed and rebased!

PR #2363, opened 2018-09-06
Update ncurses menu line-by-line when scrolling down

### LINKED ISSUE #2362: Unify scrolling of autocompletion menus
Currently when you scroll through the prompt suggestions, the behavior is slightly different when going up or down. If you go down, the suggestions are changed page by page as you reach the bottom, but when going up it happens line by line. 
It'd be nice if the behavior was the same in both directions, and I think it's nicer when going line by line (which is also how you scroll in the buffer).

The same applies to autocompletion suggestions in insert mode.

Here is an asciicast to show what I'm talking about: https://asciinema.org/a/jlABLLnBdEo6tsM8TZRhfobmZ

occivink:
Fixes #2362

### DIFF
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -891,7 +891,7 @@ void NCursesUI::menu_select(int selected)
         if (selected_line < top_line)
             m_menu.first_item = (int)selected_line * m_menu.columns;
         if (selected_line >= top_line + win_height)
-            m_menu.first_item = (int)min(selected_line, menu_lines - win_height) * m_menu.columns;
+            m_menu.first_item = (int)(selected_line + 1 - win_height) * m_menu.columns;
     }
     draw_menu();
 }


lenormf:
Nice bug fix, can't believe I've lived with this bug for so many years.

aver-d:
I gave this patch a try since you were asking for feedback. Actually, my own view is that the current way (going forwards by page) is better.

As @mawww https://github.com/mawww/kakoune/issues/2362#issuecomment-420215243 mentions, moving to the second page is normally when the first is not useful. If this second page all appears in one go, then I have three advantages: (1) immediate conceptual awareness of not one but many possibilities, (2) non-relevant results are cleared, and (3) typing a corresponding prefix is faster than moving down one by one in the hope that the next candidate will finally be something relevant.

I agree that the line-by-line movement feels somewhat smoother, so I think I it works well for going backwards. In this case, the role of the autocomplete is no longer to introduce new ideas, so it seems fine to restrict the concept of going back to mean simply I've gone too far, so back up by one.

So, essentially, I've no problem with making a change, but I'd hope any change would only be optional because I think the current system works well.

maximbaz:
I've just realized that with #2286 merged this is now even more weird and definitely needs fixing, could you rebase?

mawww:
I dont think this would be very useful, getting a full page of new completion candidates seems like a better behaviour than getting them one by one in the last column.

occivink:
I think it's a shame, but I understand the reasons against

PR #2383, opened 2018-09-12
Added deindenting of single line statements

laelath:


mawww:
Could you add a test case for that in `test/indent/c-family` ?

PR #2384, opened 2018-09-12
Formatted the makefile output to be clearer

laelath:
My attempt at making the makefile output clearer. Also removed some warning suppressions that weren't triggering.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -70,23 +70,29 @@ else
     LDFLAGS += -rdynamic
 endif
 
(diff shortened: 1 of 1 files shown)


Screwtapello:
In this configuration, `make` won't print out the commands it's about to run before it runs them. If a command fails, you won't have the command there to copy/paste for diagnosis.

However, running `make --trace` should attempt exactly the same steps and *will* print out the commands, so this PR should make things prettier in the common case while keeping problem diagnosis possible.

lenormf:
Not sure this kind of cosmetics is really worth cluttering the Makefile for.

Also it's better to show exactly what's being run by default, from a CI perspective.

laelath:
My justification was mainly that we don't need to see the same g++/clang flags for every file, and as stated running `make --trace` will print out all the commands that are being run

PR #2389, opened 2018-09-13
Add completion for rename-buffer, rename-client and rename-session

Delapouite:
Hi

Often, only a slight change to an original name is needed to get a new one during a renaming.
Like `month-08` → `month-09` or `project-customer-fr` → `project-customer-es`…

So with this PR you can just do a `:rename-session <tab>` it auto-completes with the current session name and there are just a few letters to change.

I know it can somehow be achieved with `:rename-buffer <c-r>%` but the new approach here encompasses the 3 `rename-*` commands and is friendlier to use.

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -157,6 +157,15 @@ static Completions complete_buffer_name(const Context& context, CompletionFlags 
     return { 0, cursor_pos, res };
 }
 
+auto make_single_word_completer(std::function<String (const Context&)> func)
+{
+    return make_completer(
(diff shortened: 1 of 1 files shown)


Delapouite:
Thanks. I've used `make_completer` now.

mawww:
There is still quite a bit of duplication there, I suggest adding a helper make_single_word completer taking a Context& -> String lambda.

Delapouite:
Code has been factorized with `make_single_word_completer`.

Issue #2388, opened 2018-09-13
Syntax Highlighting Overlapping


andreyorst:
I feel like syntax highlighting is not working as it should.

Consider neovim and kakoune in comparison:
This is how I tend to configure my kakoune - I'm iterating over string with filenames and sourcing them one by one. This allows me to separate configurations to different meaningful files, and avoid using autoload directory that breaks the whole loading process for some reason:
![image](https://user-images.githubusercontent.com/19470159/45503092-85a19b80-b78e-11e8-9462-d8b18b3f671e.png)

As you can see on this screenshot, i'm using highlighter `show-whitespaces` set to `-tab "▏" -lf " " -nbsp "⋅" -spc " "` Whitespaces has it's own highlighting color which is gray in my case, and those should stay gray, because those are whitespaces, and are not belong to the actual file contents. So for example, if I add this `▏` symbol to my string, I'll won't be able to tell if it is a whitespace highlighted by kakoune, or a part of a string:
![image](https://user-images.githubusercontent.com/19470159/45503391-648d7a80-b78f-11e8-8ada-961506f48539.png)
So in this example a whitespace is being highlighted as a part of a string.

Another example:

A c filetype file:
![image](https://user-images.githubusercontent.com/19470159/45503889-e3cf7e00-b790-11e8-9550-eed41e6f5272.png)
I use this user added highlighters:
```kakoune
hook global WinSetOption filetype=(c|cpp) %{                                                                               
    add-highlighter buffer/ regex \w+(\h+)?(?=\() 0:function                                                               
    add-highlighter buffer/ regex (?<=\.)\w+(?!\()(?!>)(?!") 0:Child                                                       
    add-highlighter buffer/ regex (?<=->)\w+(?!\()(?!>)(?!") 0:Child                                                       
    add-highlighter buffer/ regex \b(v|u|vu)\w+(8|16|32|64)(_t)?\b 0:type                                                  
    add-highlighter buffer/ regex \b(v|u|vu)?(_|__)?(s|u)(8|16|32|64)(_t)?\b 0:type                                        
    add-highlighter buffer/ regex \b(v|u|vu)?(_|__)?(int|short|char|long)(_t)?\b 0:type                                    
    add-highlighter buffer/ regex \b\w+_t\b 0:type                                                                         
}                                                                                                                          
add-highlighter global/ regex (\+|-|\*|=|\\|\?|%|\|-|!|\||->|\.|,|<|>) 0:operator                                          
[...]

PR #2390, opened 2018-09-15
Tweaked some of the c-family highlighting of defines

laelath:
Trying to address some of the issues with highlighting in #2388. This pull request adds recognition of comments and line comments to the \#(define,include, etc.) syntax, though it doesn't work perfectly as I could not find out how to get it to ignore line breaks within /* ... */ comments.

### DIFF
--- a/rc/core/c-family.kak
+++ b/rc/core/c-family.kak
@@ -145,7 +145,7 @@ evaluate-commands %sh{
             add-highlighter shared/FT/comment region /\* \*/ fill comment
             add-highlighter shared/FT/line_comment region // (?<!\\)(?=\n) fill comment
             add-highlighter shared/FT/disabled region -recurse "#\h*if(?:def)?" ^\h*?#\h*if\h+(?:0|FALSE)\b "#\h*(?:else|elif|endif)" fill rgb:666666
-            add-highlighter shared/FT/macro region %{^\h*?\K#} %{(?<!\\)(?=\n)} group
+            add-highlighter shared/FT/macro region %{^\h*?\K#} %{(?<!\\)(?=\n)|(?=//)} group
 
             add-highlighter shared/FT/macro/ fill meta
             add-highlighter shared/FT/macro/ regex ^\h*#include\h+(\S*) 1:module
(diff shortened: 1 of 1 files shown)


laelath:
screenshot_20180914-231328
Here's a test C file highlighting the differences between these changes and the current. Changes are on lines 6,7,18, and 21. Note this still fails on lines 16 and 17 as they should be highlighted as part of the macro.

rc/core/c-family.kak
```diff
@@ -143,12 +143,13 @@ evaluate-commands %sh{
...
             add-highlighter shared/FT/comment region /\* \*/ fill comment
-            add-highlighter shared/FT/line_comment region // $ fill comment
+            add-highlighter shared/FT/line_comment region // (?<!\\)(?=\n) fill comment
             add-highlighter shared/FT/disabled region -recurse "#\h*if(?:def)?" ^\h*?#\h*if\h+(?:0|FALSE)\b "#\h*(?:else|elif|endif)" fill rgb:666666
-            add-highlighter shared/FT/macro region %{^\h*?\K#} %{(?<!\\)\n} group
+            add-highlighter shared/FT/macro region %{^\h*?\K#} %{(?<!\\)(?=\n)} group
 
             add-highlighter shared/FT/macro/ fill meta
             add-highlighter shared/FT/macro/ regex ^\h*#include\h+(\S*) 1:module
+            add-highlighter shared/FT/macro/ regex /\*.*?\*/ 0:comment
```
lenormf: I don't think comments should be highlighted within macros, as they are supposed to be interpreted by the compiler, whereas macros are a pre-processor thing (two separate layers). It's more consistent to highlight the entire macro as such, and be agnostic of its content.

andreyorst: macroses can be huge, I once saw a macro that is 87 line long. It was something like implementation of template for plain C. It was filled with meaningful comments, explaining some complex preprocessor tricks. It is just easier to read such macro, when comments are comments.

andreyorst: I understand, but isn't it common practice in most popular editors? In Vim and Neovim for sure comments are highlighted as comments in defines.

laelath: After some experimenting, I found that the order is `line joining -> comment stripping -> macro processing` so comments preempt macros, which i think should be reflected in the highlighting. I think it would be best to show as closely as possible how the compiler is gonna read the code.

PR #2404, opened 2018-09-19
Scope clang output with characters less likely to appear

laelath:


lenormf:
Might as well use the format here!

```sh
printf 'info -anchor %d.%d "%s"\n' "${kak_cursor_line}" "${kak_cursor_column}" "$(printf %s "${desc}" | sed 's/"/""/g')"
```

PR #2407, opened 2018-09-19
Fix highlighting of unsigned long long in C filetype

andreyorst:
Values like `0xffffffffffffffffull` were highlighted only till first `u` leaving `ll` without highlighting. This change addresses this issue. It also adds uppercase `ULL` highlighting in values.
Old:
image
New:
image

### DIFF
--- a/rc/core/c-family.kak
+++ b/rc/core/c-family.kak
@@ -154,7 +154,7 @@ evaluate-commands %sh{
 }
 
 # c specific
-add-highlighter shared/c/code/numbers regex %{\b-?(0x[0-9a-fA-F]+|\d+)[fdiu]?|'((\\.)?|[^'\\])'} 0:value
+add-highlighter shared/c/code/numbers regex %{\b-?(0x[0-9a-fA-F]+|\d+)([fdiuUlL]+)?|'((\\.)?|[^'\\])'} 0:value
 evaluate-commands %sh{
     # Grammar
     keywords="asm break case continue default do else for goto if return
--- a/rc/core/c-family.kak
+++ b/rc/core/c-family.kak
@@ -154,7 +154,7 @@ evaluate-commands %sh{
 }
 
 # c specific
-add-highlighter shared/c/code/numbers regex %{\b-?(0x[0-9a-fA-F]+|\d+)([fdiuUlL]+)?|'((\\.)?|[^'\\])'} 0:value
+add-highlighter shared/c/code/numbers regex %{\b-?(0x[0-9a-fA-F]+|\d+)([fFdDiIuUlL]+)?|'((\\.)?|[^'\\])'} 0:value
 evaluate-commands %sh{
     # Grammar
     keywords="asm break case continue default do else for goto if return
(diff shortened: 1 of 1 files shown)


rc/core/c-family.kak
```diff
@@ -154,7 +154,7 @@ evaluate-commands %sh{
 }
 
 # c specific
-add-highlighter shared/c/code/numbers regex %{\b-?(0x[0-9a-fA-F]+|\d+)[fdiu]?|'((\\.)?|[^'\\])'} 0:value
+add-highlighter shared/c/code/numbers regex %{\b-?(0x[0-9a-fA-F]+|\d+)([fFdDiIuUlL]+)?|'((\\.)?|[^'\\])'} 0:value
```
mawww: `([fFdDiIuUlL]+)?` == `[fFdDiIuUlL]*`, right ?

andreyorst: Yes. This is better. Thanks, I'll fix it

andreyorst: I think then we should use `[fFdDiIuUlL]{,3}`

andreyorst: because something like `[fFdDiIuU]?[lL]{,2}` requires ordering in suffixes, in terms of the fact, that `ull` is possible, yet `llu` isn't, while it is still valid.

lenormf: I think it's fine to have this kind of false positive, implementing static checks with regex is close to impossible (ever read the regex that validates an email address?). In this particular case, `[fFdDiIuUlL]*` is good enough.

andreyorst: with this regexp `([uU][lL]{1,2}|[uU]|[lL]{1,2}|[lL]{1,2}[uU]|[fFdDiI])?` I've managed to make kakoune work similar to vim in term of highlighting suffixes (except i'm not sure about `i` and `d`):
image
Should I use it in this PR?

andreyorst: > ever read the regex that validates an email address?

@lenormf Yes........ :cry:

andreyorst: I've pushed new regexp. Waiting for your judgement.

rc/core/c-family.kak
```diff
@@ -154,7 +154,7 @@ evaluate-commands %sh{
 }
 
 # c specific
-add-highlighter shared/c/code/numbers regex %{\b-?(0x[0-9a-fA-F]+|\d+)[fdiu]?|'((\\.)?|[^'\\])'} 0:value
+add-highlighter shared/c/code/numbers regex %{\b-?(0x[0-9a-fA-F]+|\d+)([uU][lL]{1,2}|[uU]|[lL]{1,2}|[lL]{1,2}[uU]|[fFdDiI])?|'((\\.)?|[^'\\])'} 0:value
```
mawww: I think this can be compacted into `[uU][lL]{0,2}|[lL]{1,2}[uU]?|[fFdDiI]`, is that right ?

andreyorst: Yes. Seem's like I've gone too far at some point. I've tested your proposed regex, it works fine. I'll commit it to my branch. Thanks.

PR #2410, opened 2018-09-20
Fixed several clang formatting errors caused by string escaping

laelath:


rc/extra/clang.kak
```diff
@@ -136,11 +136,11 @@ define-command -hidden clang-show-error-info %{
         shift # skip timestamp
         for error in "$@"; do
             if [ "${error%%|*}" = "$kak_cursor_line" ]; then
-                desc=$(printf '%s%s\n' "$desc" "${error##*|}")
+                desc=$(printf '%s%s\n ' "${desc::-1}" "${error##*|}")
```
mawww: I dont know this `${var::-1}` syntax, cannot see it in the posix shell spec.

laelath:
This is ready to merge, I fixed the newlines being eaten by the subshell by adding a character after it ('x'), which is removed when the variable is expanded.

rc/extra/clang.kak
```diff
@@ -136,11 +136,11 @@ define-command -hidden clang-show-error-info %{
         shift # skip timestamp
         for error in "$@"; do
```
mawww: How about

```
desc=$(for error in "$@"; do ...)
```

I belive this would only eat the last newline, which we dont really care about.

rc/extra/clang.kak
```diff
@@ -134,13 +134,13 @@ define-command -hidden clang-show-error-info %{
...
+        desc=$(for error in "$@"; do
             if [ "${error%%|*}" = "$kak_cursor_line" ]; then
-                desc=$(printf '%s%s\n' "$desc" "${error##*|}")
+                printf '%s\n' "${error##*|}"
             fi
-        done
+        done)
         if [ -n "$desc" ]; then
-            printf %s\\n "info -anchor ${kak_cursor_line}.${kak_cursor_column} \"$desc\""
+            printf %s "${desc%x}" | sed "s/'/''/g" | xargs -0 printf "info -anchor %d.%d '%s'\n" "${kak_cursor_line}" "${kak_cursor_column}"
```
mawww: I think that `%x` in `${desc%x}` is not required anymore, is it ?

lenormf: I'm also not sure `xargs -0` is needed here, this should probably be turned into a `while read -r;` loop.

laelath: Is there any reason to avoid xargs? I thought it would be the clearest way to show we wanted the uninterpreted value as the argument.

lenormf: In this particular instance, the `-0` flag is not defined by POSIX so `xargs -0` is most likely not portable. Otherwise no strong bias against it!

laelath: Oh, I didn't know the `-0` flag wasn't POSIX. I'll try and make a `while read` loop work.

PR #2411, opened 2018-09-20
Document that `-f` starts with whole file selected

eraserhd:


lenormf:
I'm wondering if this description isn't a bit misleading, as the `%` is actually never executed - the entire buffer is rather selected before-hand. Maybe users are going to think that their custom mapping of `%` is going to be executed every time they use `-f`.

eraserhd:
@lenormf Good point.

PR #2415, opened 2018-09-21
Added more kitty integration

laelath:


rc/extra/kitty.kak
```diff
@@ -22,7 +35,27 @@ If no client is passed then the current one is used} \
...
+define-command -docstring %{kitty-repl [<arguments>]: create a new window for repl interaction
+All optional parameters are forwarded to the new window} \
+    -params .. \
+    -command-completion \
+    kitty-repl %{ evaluate-commands %sh{
+        if [ -z "${kak_opt_termcmd}" ]; then
+           echo "echo -markup '{Error}termcmd option is not set'"
+           exit
+        fi
+        if [ $# -eq 0 ]; then cmd="${SHELL:-sh}"; else cmd="$@"; fi
```
lenormf: I think that the `$SHELL` variable should contain a fullpath, so the default value here should probably be `/bin/sh`.

As for the `else` statement, I think you want `cmd="$*"` here, which is guaranteed to be portable.

laelath:
Would it be possible to make the shell command completion used in insert-output, pipe-to, etc. available from declare-command?

rc/extra/kitty.kak
```diff
@@ -1,8 +1,13 @@
-hook global KakBegin .* %sh{
+declare-option -docstring %{window type that kitty creates on new and repl calls (kitty|os)} str kitty_window_type kitty
```
mawww: I dont really get the interest of `os` window type, doesnt it boils down to doing the same as x11.kak and x11-repl.kak ? Once kitty does not do window management anymore, why go through kitty.kak instead of x11.kak (so a user can opt to use x11.kak by manually overriding the aliases).

maximbaz: I believe it is faster to spawn, kitty reuses some resources when doing so

laelath: Also, if someone is using kitty on Wayland or macOS they couldn't go the X11 route, but kitty could offer that behaviour.

PR #2419, opened 2018-09-22
Added shell command completion support to define-command

laelath:
This option was introduced so `repl` commands could have proper completion suggestions.

This commit also introduces a regression in that I decided that the best way to avoid overly long and confusing switch names was to rename the current `shell-*` switches to `script-*`, and have the shell command completion switch be `shell-completion`. I am open to name suggestions if this regression is a bad idea.

### DIFF
--- a/doc/pages/commands.asciidoc
+++ b/doc/pages/commands.asciidoc
@@ -291,6 +291,9 @@ New commands can be defined using the *define-command* command:
     try command completion on any parameter passed to this command
 
 *-shell-completion*:::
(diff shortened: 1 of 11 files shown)


mawww:
I like this change, I am wondering about the `-script-...` name, I think we should have shell in it, as `script` could be confusing (a `.kak` file is a script as well, so scripts are not necessarily shell scripts in Kakoune). Maybe `-shell-script-{candidate,completion}` ?

PR #2424, opened 2018-09-23
src: Switch to the new file when saving a scratch buffer

### LINKED ISSUE #2360: Rename buffer to file when saving
Rename buffer to file when saving, when the buffer is not associated to a file yet, and when saving with a name different from the buffer name.

The buffer flags would need to be updated accordingly.

For example, when opening Kakoune with no file argument, its entry-point is a scratch buffer; so when saving with `:write {file-name}`, the buffer will be renamed and the scratch flag be removed.

lenormf:
Fixes #2360 

Selections won't be preserved, not sure if we want that.

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -343,9 +343,23 @@ void do_write_buffer(Context& context, Optional<String> filename, bool sync_file
 template<bool force = false>
 void write_buffer(const ParametersParser& parser, Context& context, const ShellContext&)
 {
-    return do_write_buffer(context,
-                           parser.positional_count() > 0 ? parser[0] : Optional<String>{},
-                           (bool)parser.get_switch("sync"), force);
+    do_write_buffer(context,
+                    parser.positional_count() > 0 ? parser[0] : Optional<String>{},
+                    (bool)parser.get_switch("sync"), force);
+
+    if (parser.positional_count() > 0
+        and not (context.buffer().flags() & (Buffer::Flags::File | Buffer::Flags::Debug)))
+    {
+        auto& name = parser[0];
+        Buffer* buffer = BufferManager::instance().get_buffer_ifp(name);
+
+        if (not buffer)
+            buffer = open_file_buffer(name, Buffer::Flags::File);
+        else
+            reload_file_buffer(*buffer);
+
+        context.change_buffer(*buffer);
+    }
 }
 
 const CommandDesc write_cmd = {


src/commands.cc
```diff
@@ -343,9 +343,23 @@ void do_write_buffer(Context& context, Optional<String> filename, bool sync_file
...
 void write_buffer(const ParametersParser& parser, Context& context, const ShellContext&)
 {
-    return do_write_buffer(context,
-                           parser.positional_count() > 0 ? parser[0] : Optional<String>{},
-                           (bool)parser.get_switch("sync"), force);
+    do_write_buffer(context,
+                    parser.positional_count() > 0 ? parser[0] : Optional<String>{},
+                    (bool)parser.get_switch("sync"), force);
+
+    if (parser.positional_count() > 0
```
mawww: That should be done inside do_write_buffer, so that `:wq` does rename the buffer as well (it is relevant in the case `q` fails). Additionally, it should be done by changing the buffer flags and name rather than by reloading the file itself.

lenormf: That's what my initial implementation did, it got ugly quickly. I'll give it another shot.

lenormf: Actually I don't understand: why change the name? I'm using `reload_file_buffer()` to lazily have the timestamp refreshed, so that we don't get the "do you want to refresh the buffer?" prompt. In theory, all I need to do is change the timestamp, not the name/flags?

mawww: Change the buffer flags, name, then actually write it so that the timestamp gets updated as a side effect.

lenormf: I think there's a confusion around how the feature is supposed to be implemented.

It looks like you'd have the scratch buffer _moved_ to a file buffer.

I'd like a new file buffer to be created, containing whatever was in the scratch buffer at the time of writing.

Right?

mawww: Right, we do not agree on the semantics of the change, to me it does not make sense for `:write` to open a new buffer for *scratch* buffers. If it did that, why wouldn't it do it for regular buffers ?

Frankly my preference is for `:write` to do what it always did, which is just write the buffer to a file. But if we decide it needs to do more for scratch buffers, changing their scratch nature to convert them to full "File" buffers is what I'd prefer, duplicating the buffer into a File one, keeping the scratch one away feels really wrong.

lenormf: I wanted to trigger this code only on scratch buffers, I must have messed up the copying/pasting here.

As for turning all scratch buffers into actual files, it looks weird to apply that behaviour to scratch buffers that are user interfaces for commands (`*lint-output*` etc).

I'll rework this hunk and update the PR.

mawww: The more I think about this, the more I think the current behaviour is just better, its simple, and `:write` does exactly what the name says, not more, not less. I think we would be better of with a way to convert a scratch buffer to a file buffer (say `rename-buffer -file <filename>`).

lenormf: We probably need to go back to the drawing board on that one, I'll close this so we can decide what we want in the issue.

PR #2430, opened 2018-09-24
Handlebars highlighter refactor

schickm:
I've been working a lot with handlebars files recently and noticed a number of deficiencies in the current highlighter.  In this PR I've done the following:

- Updated a number of the regexs to better match the handlebars spec
- Restructured the highlighter so that it injects itself into the html highlighter.  As handlebars code can live inside of html tags, this is necessary to highlight it correctly.  I've made to to remove any of the injected highlighters when the user transitions to a non-handlebars window (as highlighters are shared globally)
- Enabled the html indentation hooks
- Fix the existing handlebars indentation hooks


mawww:
Could you squash those commits together ?

mawww:
Thank you very much.

PR #2433, opened 2018-09-24
src: Drop selections that only contain whitespace on `_`

### LINKED ISSUE #2387: Use _ command to filter blank selections

lenormf:
Fixes #2387


src/normal.cc
```diff
@@ -1710,18 +1710,32 @@ void spaces_to_tabs(Context& context, NormalParams params)
 
 void trim_selections(Context& context, NormalParams)
 {
+    const size_t nb_sels = context.selections().size();
     auto& buffer = context.buffer();
-    for (auto& sel : context.selections())
+    auto& selections = context.selections();
+    Vector<size_t> drop_idx;
```
mawww: Avoid that vector, you can do
```
for (auto it = selections.begin(); it != selections.end(); )
{
    ...
    if (beg == end and is_blank(*beg))
        it = selections.erase(it);
    else
        ++it;
}
```

lenormf: I was worrying that removing elements within a loop that iterates over the container would not be safe.

lenormf: Actually `SelectionList::erase()` doesn't take any argument, and I don't really understand what it's doing?

mawww: Sorry I got confused, I assumed selections was a vector, but its a SelectionList, which has a different interface. In that case, use an index:

```
for (size_t i = 0; i < selections.size(); )
{
    ...
    if (...)
        selections.remove(i);
    else
        ++i;
}
```

There remains one problem which is what to do when every selections are removed. The convention is to throw an error which means we should ensure the selections are restored. So in the end your solution is actually better, you should tweak it to look more like what the `select` function in normal.cc:68 does, in particular just push_back and then iterate on the reversed view.

lenormf:
I made the function be more similar to `void select(Context& context, T func)`, rebased on `master`.

lenormf:
Bump!

PR #2379, opened 2018-09-10
Documentation refactoring

lenormf:
Hi,

I removed the documentation from the README that was redundant (and outdated) with the pages in `doc/`.

Some bits have been re-written, spelling was checked, items re-ordered alphabetically…

HTH.

PR #2434, opened 2018-09-25
expansions.asciidoc: Rewrite to make Kakoune syntax the primary reference

Screwtapello:
Also, moved the "Markup strings" section to faces.asciidoc, since it wasn't anything to do with expansions, and updated various hyperlinks to point at the new location.

As proposed [on the forum](https://discuss.kakoune.com/t/proposal-reorganising-the-expansions-documentation/59).

[Rendered](https://github.com/Screwtapello/kakoune/blob/rewrite-expansions-doc/doc/pages/expansions.asciidoc)

Things that particularly need fact-checking:

- the rules in "Shell Expansions" for when an environment variable is shell-quoted versus when it's raw
- the valid context listed for each value expansion

Tag @lenormf, since this conflicts with #2379.

### DIFF
--- a/doc/pages/expansions.asciidoc
+++ b/doc/pages/expansions.asciidoc
@@ -1,199 +1,294 @@
 = Expansions
 
-== Strings
+While parsing a command (see <<command-parsing#,`:doc command-parsing`>>),
+Kakoune recognises certain patterns and will replace them with their
+associated value before executing the command. These patterns are called
+expansions.
 
-\'strings'::
-    uninterpreted strings, use a backslash (\') to escape the separator
+Every expansion consists of a `%`, followed by the expansion _type_ (one
+or more alphabetic characters), a nestable punctuation character (`(`, `[`,
+`{`, or `<`), and then all the text up to and including its matching opposite
+(`)`, `]`, `}`, or `>`). It doesn't matter which character is used, but
+`{}` are most common. For example, the command `echo %val{session}` will
+echo the ID of the current Kakoune session.
 
-"strings"::
-    expanded strings, % strings (c.f. next section) contained are expended,
-    use a backslash (\\%) to escape the separator
+Expansions are processed when unquoted and anywhere inside double-quoted or
+%-strings, but not inside unquoted words, inside single-quoted strings, or
+inside other expansions. So:
 
-%\{strings\}::
-    these strings are very useful when entering commands
+* `echo %val{session}` echoes the current session ID
 
-    * the '{' and '}' delimiters are configurable, any non alphanumeric
-      character can be used
+* `echo x%val{session}x` echoes the literal text `x%val{session}x`
 
-----------------------------------------------------------
-e.g. %[string], %<string>, %(string), %~string~, %!string!
-----------------------------------------------------------
+* `echo '%val{session}'` echoes the literal text `%val{session}`
 
-    * if the character following '%' is one of '{[(<', then the closing
-      one is the matching '}])>' and the delimiters are not escapable but
-      are nestable
+* `echo "x%val{session}x"` echoes the current session ID, surrounded by `x`
 
------------------------------------------------------------
-e.g. %{ roger {}; } is a valid string, %{ marcel \} as well
------------------------------------------------------------
+* `echo %{x%val{session}x}"` echoes the current session ID, surrounded by `x`
 
-== Typed expansions
+* `echo %sh{ echo %val{session} }"` echoes the literal text `%val{session}`
 
-%\{strings\} can have an expansion type between the *%* and the opening
-character. They will be written *%<type>\{<content>\}*. They will be
-expanded according to the given *<type>* using *<content>* as its
-parameter:
+When processing an expansion, Kakoune ensures the result is properly quoted so
+that any special characters do not interfere with the rest of the command. If
+an expansion expands to multiple values, each value is individually quoted.
 
-*sh*::
-    shell expansion, similar to posix shell '$(...)' construct (c.f. next
-    section)
+== Argument expansions
 
-*reg*::
-    register expansion, will expand to the strings stored in the register
-    named by *<content>*. See <<registers#,`:doc registers`>>
+Expansions with the type `arg` can only be used inside the "commands" parameter
+of the `define-command` command (See <<commands#declaring-new-commands,`:doc
+commands declaring-new-commands`>>).
 
-*opt*::
-    option expansion, will expand to the value of the option named by
-    *<content>*. See <<options#,`:doc options`>>
+The following expansions are available:
 
-*val*::
-    value expansion, will expand to the value of the environment variables
-    available to shell expansion. *<content>* shall be the name of that
-    variable without the *kak_* prefix.
+*%arg{n}*::
+     (where _n_ is a decimal number) +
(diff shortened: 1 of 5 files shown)


doc/pages/expansions.asciidoc
```diff
@@ -1,199 +1,294 @@
...
+`{`, or `<`), and then all the text up to and including its matching opposite
+(`)`, `]`, `}`, or `>`). It doesn't matter which character is used, but
+`{}` are most common. For example, the command `echo %val{session}` will
+echo the ID of the current Kakoune session.
 
-"strings"::
-    expanded strings, % strings (c.f. next section) contained are expended,
-    use a backslash (\\%) to escape the separator
+Expansions are processed when unquoted and anywhere inside double-quoted or
+%-strings, but not inside unquoted words, inside single-quoted strings, or
```
mawww: Thats incorrect, %-strings do not expands nested % strings.

doc/pages/expansions.asciidoc
```diff
@@ -1,199 +1,294 @@
...
 
-    * if the character following '%' is one of '{[(<', then the closing
-      one is the matching '}])>' and the delimiters are not escapable but
-      are nestable
+* `echo "x%val{session}x"` echoes the current session ID, surrounded by `x`
 
------------------------------------------------------------
-e.g. %{ roger {}; } is a valid string, %{ marcel \} as well
------------------------------------------------------------
+* `echo %{x%val{session}x}"` echoes the current session ID, surrounded by `x`
```
mawww: This is false.

doc/pages/expansions.asciidoc
```diff
@@ -1,199 +1,294 @@
...
-== Typed expansions
+* `echo %sh{ echo %val{session} }"` echoes the literal text `%val{session}`
 
-%\{strings\} can have an expansion type between the *%* and the opening
-character. They will be written *%<type>\{<content>\}*. They will be
-expanded according to the given *<type>* using *<content>* as its
-parameter:
+When processing an expansion, Kakoune ensures the result is properly quoted so
+that any special characters do not interfere with the rest of the command. If
+an expansion expands to multiple values, each value is individually quoted.
```
mawww: I am not sure `quoted` is correct here, expansions can expand to multiple words, but they are not reparsed at all.

doc/pages/expansions.asciidoc
```diff
@@ -1,199 +1,294 @@
...
-    variable without the *kak_* prefix.
+*%arg{n}*::
+     (where _n_ is a decimal number) +
+     expands to argument number _n_ of the current command
 
-*arg*::
-    argument expansion, expand to the arguments of the current
-    command, *<content>* can be a number, or @ for all arguments
+*%arg{@}*::
+    expands to all the arguments of the current command, individually quoted
```
mawww: As individual "words" I think a "word" should be clearly defined before.

doc/pages/expansions.asciidoc
```diff
@@ -1,199 +1,294 @@
...
+----
 
-*kak_version*::
-    version of the current Kakoune server (git hash or release name)
+...will find none of Kakoune's special environment variables, but:
 
-*kak_count*::
-    count parameter passed to the command, defaults to 0 if no count given
+----
+echo %sh{ env | grep ^kak_ # $kak_session }
```
mawww: the `$` is not necessar here.

doc/pages/expansions.asciidoc
```diff
@@ -1,199 +1,294 @@
...
+Kakoune uses a shell command, such as the `|`, `!` or `$` normal mode commands
+(See <<keys#,`:doc keys`>>).
 
-*kak_reg_<r>*::
-    quoted list of value of register *r*
+== Value expansions
 
-*kak_main_reg_<r>*::
-    content of register *r* associated with the main selection.
+Expansions with the type `val` can be used in various places, depending on the
```
mawww: Expansions of type `val` give access to Kakoune internal data that is not stored in an option or a register.

doc/pages/expansions.asciidoc
Screwtapello: Oops! Was that ever true, or am I just confused?

doc/pages/expansions.asciidoc
Screwtapello: Oh, good to know!

doc/pages/expansions.asciidoc
Screwtapello: Well put, I'll use that.

Screwtapello:
Thanks for the thorough review! I've pushed a new commit that addresses your comments.

doc/pages/expansions.asciidoc
lenormf: They never did.

doc/pages/expansions.asciidoc
```diff
@@ -1,199 +1,300 @@
...
 
-*kak_config*::
-    directory containing the user configuration
+----
+echo %sh{ env | grep ^kak_ }
+----
 
-*kak_version*::
-    version of the current Kakoune server (git hash or release name)
+...will find none of Kakoune's special environment variables, but:
```
mawww: I am surprised by `...` not being followed by a space, but it might well be the usual way to write it, is that correct ?

doc/pages/expansions.asciidoc
```diff
@@ -1,199 +1,300 @@
...
-*kak_source*::
-    path of the file currently getting executed (through the source
-    command)
+*%val{bufname}*::
+    _in buffer scope_ +
+    name of the current buffer
 
-*kak_modified*::
-    buffer has modifications not saved
+*%val{client_env_X}*::
```
mawww: I wonder if repeating `%val{name}` instead of just `name` is better or not.

doc/pages/expansions.asciidoc
```diff
@@ -1,199 +1,300 @@
...
-    unquoted list of the names of clients connected to the current session
+*%val{buflist}*::
+    quoted list of the names of currently-open buffers (as seen in
+    `%val{bufname}`)
 
-*kak_source*::
-    path of the file currently getting executed (through the source
-    command)
+*%val{bufname}*::
+    _in buffer scope_ +
```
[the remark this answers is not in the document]
Screwtapello: There's a paragraph at the end that mentions anything "in buffer scope" is also available in window scope. Maybe it would be better to just say "in buffer, window scope" for such expansions?

doc/pages/expansions.asciidoc
Screwtapello: My personal habit is that an ellipsis gets space on one side or the other depending on whether it's trailing off ...

... or resuming a previous thought. However, now that I check Wikipedia, it has [any number of contradictory guidelines](https://en.wikipedia.org/wiki/Ellipsis#In_American_English), none of which really cover this specific circumstance. 

I'll add a space.

doc/pages/expansions.asciidoc
Screwtapello: It's a bit redundant, but I like the idea that somebody can grep for `%val{` and get a list of all of them, right there in the grep-results window. If Kakoune someday gets a documentation index,  `%val{selection}` will be a much more useful and specific index-entry than `selection`.

PR #2440, opened 2018-09-27
Fix/improve html-indent-on-new-line hook in html.kak highlighter

schickm:
* Fixes erroring execute-keys statement by using `<lt>` rather then `<`
* Reworks hook to not indent after self closing tags

### DIFF
--- a/rc/base/html.kak
+++ b/rc/base/html.kak
@@ -51,8 +51,7 @@ define-command -hidden html-indent-on-new-line %{
         # filter previous line
         try %{ execute-keys -draft k : html-filter-around-selections <ret> }
         # indent after lines ending with opening tag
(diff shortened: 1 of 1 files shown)


lenormf:
I'm not sure using a shell for that is necessary, you might have gotten away with embedding the keywords in the function directly.

schickm:
@lenormf Mind sharing with me an example of how to do that?  I only did it in shell because I wanted to have a nice list of self closing tags, and then reference them in the regex.  I didn't want a giant long regex. 

I picked it up from a pattern I saw in [scheme.kak](https://github.com/mawww/kakoune/blob/master/rc/extra/scheme.kak#L45)

schickm:
@lenormf Alright, it's updated to remove the shell wrapper.  I wasn't able to find any better way of doing without shell then just dropping all of those self-closing-tag regexes into the main regex.  But, now that I did it, I don't think it looks all that bad.

Issue #2436, opened 2018-09-25
Use the matching_pairs option for the show-matching highlighter


Delapouite:
Hi

The `m` / `M` commands honors the `matching_pairs` option as defined [here](https://github.com/mawww/kakoune/blob/master/src/selectors.cc#L227):
```c++
auto& matching_pairs = context.options()["matching_pairs"].get<Vector<Codepoint, MemoryDomain::Options>>();
```

Meanwhile the `show-matching` highlighter uses a [hardcoded list](https://github.com/mawww/kakoune/blob/master/src/highlighters.cc#L1134):
```c++
static const CodepointPair matching_chars[] = { { '(', ')' }, { '{', '}' }, { '[', ']' }, { '<', '>' } };`
```
It would be great if the `show-matching` also refers to the same `matching_pairs` option. This way when new chars pairs are added to this option, the highlighter works for them as well.

PR #2443, opened 2018-09-28
Made the show_matching_char highlighter use the matching_pairs option

laelath:
Implement the idea in #2436

### DIFF
--- a/src/highlighters.cc
+++ b/src/highlighters.cc
@@ -26,6 +26,8 @@
 namespace Kakoune
 {
 
+using Utf8Iterator = utf8::iterator<BufferIterator>;
+
 template<typename Func>
 std::unique_ptr<Highlighter> make_highlighter(Func func, HighlightPass pass = HighlightPass::Colorize)
 {
@@ -1130,8 +1132,7 @@ constexpr StringView LineNumbersHighlighter::ms_id;
 void show_matching_char(HighlightContext context, DisplayBuffer& display_buffer, BufferRange)
 {
     const Face face = context.context.faces()["MatchingChar"];
-    using CodepointPair = std::pair<Codepoint, Codepoint>;
-    static const CodepointPair matching_chars[] = { { '(', ')' }, { '{', '}' }, { '[', ']' }, { '<', '>' } };
+    const auto& matching_pairs = context.context.options()["matching_pairs"].get<Vector<Codepoint, MemoryDomain::Options>>();
     const auto range = display_buffer.range();
     const auto& buffer = context.context.buffer();
     for (auto& sel : context.context.selections())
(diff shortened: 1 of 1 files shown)


src/highlighters.cc
```diff
@@ -1130,52 +1130,56 @@ constexpr StringView LineNumbersHighlighter::ms_id;
 void show_matching_char(HighlightContext context, DisplayBuffer& display_buffer, BufferRange)
 {
     const Face face = context.context.faces()["MatchingChar"];
-    using CodepointPair = std::pair<Codepoint, Codepoint>;
-    static const CodepointPair matching_chars[] = { { '(', ')' }, { '{', '}' }, { '[', ']' }, { '<', '>' } };
+    // static const CodepointPair matching_chars[] = { { '(', ')' }, { '{', '}' }, { '[', ']' }, { '<', '>' } };
```
Screwtapello: You should remove this line, rather than just commenting it out.

src/highlighters.cc
```diff
@@ -1130,52 +1130,56 @@ constexpr StringView LineNumbersHighlighter::ms_id;
...
     const auto& buffer = context.context.buffer();
     for (auto& sel : context.context.selections())
     {
         auto pos = sel.cursor();
         if (pos < range.begin or pos >= range.end)
             continue;
-        auto c = buffer.byte_at(pos);
-        for (auto& pair : matching_chars)
+        auto match = matching_pairs.end();
+        match = find(matching_pairs, buffer.byte_at(sel.cursor()));
```
mawww: Look for the codepoint, not the byte, so that arbitrary matching pairs can be used

src/highlighters.cc
laelath: Oh that's embarrassing.

src/highlighters.cc
laelath: Am I right that dereferencing a buffer iterator at the cursor position would do that? I admit that I have not looked too closely at the internal buffer manipulation.

src/highlighters.cc
```diff
@@ -1130,52 +1130,55 @@ constexpr StringView LineNumbersHighlighter::ms_id;
...
     const auto& buffer = context.context.buffer();
     for (auto& sel : context.context.selections())
     {
         auto pos = sel.cursor();
         if (pos < range.begin or pos >= range.end)
             continue;
-        auto c = buffer.byte_at(pos);
-        for (auto& pair : matching_chars)
+        auto match = matching_pairs.end();
+        match = find(matching_pairs, *get_iterator(buffer, pos));
```
lenormf: Why not just `auto match = find(…)` ?

src/highlighters.cc
```diff
@@ -1130,52 +1130,54 @@ constexpr StringView LineNumbersHighlighter::ms_id;
...
     const auto range = display_buffer.range();
     const auto& buffer = context.context.buffer();
     for (auto& sel : context.context.selections())
     {
         auto pos = sel.cursor();
         if (pos < range.begin or pos >= range.end)
             continue;
-        auto c = buffer.byte_at(pos);
-        for (auto& pair : matching_chars)
+        auto match = find(matching_pairs, *get_iterator(buffer, pos));
```
mawww: utf8::codepoint(get_iterator(buffer, pos), buffer.end());

PR #2447, opened 2018-09-28
Better Lisp and Clojure indent

eraserhd:
Algol-like indenting doesn't work for Lisp as we tend to open (and
close) many forms on one line.  This generally puts the cursor
where it should be, though it is probably confused by strings
with brackets and things.

### DIFF
--- a/rc/base/clojure.kak
+++ b/rc/base/clojure.kak
@@ -158,7 +158,32 @@ evaluate-commands %sh{
 # ‾‾‾‾‾‾‾‾
 
 define-command -hidden clojure-filter-around-selections lisp-filter-around-selections
-define-command -hidden clojure-indent-on-new-line       lisp-indent-on-new-line
+
+declare-option \
+    -docstring 'regex matching the head of forms which have options *and* indented bodies' \
+    regex clojure_special_indent_forms \
+    '(?:def.*|doseq|for|fn\*?|if(-.*|)|let.*|loop|ns|testing|with-.*|when(-.*|))'
+
+define-command -hidden clojure-indent-on-new-line %{
+    # registers: i = best align point so far; w = start of first word of form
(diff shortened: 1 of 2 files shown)


rc/base/clojure.kak
```diff
@@ -167,6 +167,7 @@ hook -group clojure-highlight global WinSetOption filetype=clojure %{ add-highli
 hook global WinSetOption filetype=clojure %[
     hook window ModeChange insert:.* -group clojure-hooks  clojure-filter-around-selections
     hook window InsertChar \n -group clojure-indent clojure-indent-on-new-line
+    set-option window lisp_special_indent_forms '(?:def.*|fn\*?|if(-.*|)|let.*|loop|with-.*|when(-.*|))'
```
lenormf: Don't be afraid to duplicate code, it's preferable to creating inter-module dependencies with options.

rc/base/lisp.kak
```diff
@@ -26,17 +26,31 @@ add-highlighter shared/lisp/code/ regex (\b\d+)?\.\d+([eEsSfFdDlL]\d+)?\b 0:valu
 # Commands
 # ‾‾‾‾‾‾‾‾
 
-define-command -hidden lisp-filter-around-selections %{
+define-command -override -hidden lisp-filter-around-selections %{
```
lenormf: I think we'd rather rename this function for use in the `lisp.kak` module, rather than try to override a command from another module.

eraserhd: Oops.  This wasn't actually overridden, I just had this here for development.  Removed.

rc/base/clojure.kak
eraserhd: Ahh, OK.  The option is there because people often make macros that need to be added here, but I can copy all the lisp stuff to the Clojure and rename.

PR #2448, opened 2018-09-28
improve editorconfig support

TeddyDD:
- ~~support for insert_final_newline option~~ already enforced by Kakoune
- highlighting (ini) and static_words

### DIFF
--- a/rc/extra/editorconfig.kak
+++ b/rc/extra/editorconfig.kak
@@ -1,6 +1,16 @@
 # http://editorconfig.org/#file-format-details
 # ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
 
+# Detection
+# ‾‾‾‾‾‾‾‾‾
+
+hook global BufCreate .*[.](editorconfig) %{
+    set-option buffer filetype ini
+    set-option buffer static_words indent_style indent_size tab_width \
+    end_of_line charset insert_final_newline trim_trailing_whitespace root \
+    latin1 utf-8 utf-8-bom utf-16be utf-16le lf cr crlf unset space tab 
+}
+
 declare-option -hidden bool editorconfig_trim_trailing_whitespace false
 
 define-command editorconfig-load -params ..1 -docstring "editorconfig-load [file]: set formatting behavior according to editorconfig" %{


rc/extra/editorconfig.kak
```diff
@@ -1,9 +1,20 @@
...
+    set-option buffer static_words indent_style indent_size tab_width \
+    end_of_line charset insert_final_newline trim_trailing_whitespace root \
+    latin1 utf-8 utf-8-bom utf-16be utf-16le lf cr crlf unset space tab 
+}
+
 declare-option -hidden bool editorconfig_trim_trailing_whitespace false
+declare-option -hidden bool editorconfig_insert_final_newline false
 
-define-command editorconfig-load -params ..1 -docstring "editorconfig-load [file]: set formatting behavior according to editorconfig" %{
+define-command -override editorconfig-load -params ..1 -docstring "editorconfig-load [file]: set formatting behavior according to editorconfig" %{
```
lenormf: Why do we need this flag here?

rc/extra/editorconfig.kak
```diff
@@ -33,13 +45,19 @@ define-command editorconfig-load -params ..1 -docstring "editorconfig-load [file
...
+                    print "set-option buffer editorconfig_insert_final_newline true"
+                }
                 if (trim_trailing_whitespace == "true") {
                     print "set-option buffer editorconfig_trim_trailing_whitespace true"
                 }
             }
         '
     }
     hook buffer BufWritePre %val{buffile} -group editorconfig-hooks %{ evaluate-commands %sh{
+        if [ ${kak_opt_editorconfig_insert_final_newline} = "true" ]; then
```
lenormf: I think you need to enclose this expansions in double quotes.

rc/extra/editorconfig.kak
TeddyDD: that was leftover of debug, removed

rc/extra/editorconfig.kak
TeddyDD: not sure, line below is:
 `if [ ${kak_opt_editorconfig_trim_trailing_whitespace} = "true" ]; then`

TeddyDD: also shellcheck does not complain

lenormf: I guess it's fine because this option can only expand in a single alphabetical word, but usually arbitrary expansions have to be enclosed in double quotes.

TeddyDD: I know, I can change it for both lines to be consistent, wdyt?

lenormf: Leave it, it's not an issue for now.

casimir:
The `insert_final_newline` option was left out because kakoune assumes that files always end with a newline, adding one if absent. Unless it changed since that time this option would always be `true`.

TeddyDD:
That's not true and can be easily tested:

`kak -n /tmp/foo` and write single line of text: `ifoo<esc>:wq`. Open file with any editor. There is no empty line at the end of the file.

TeddyDD:
Okay, since there is no way to disable this, I'll drop the commit.

PR #2449, opened 2018-09-28
rc git: Support notes

lenormf:


eraserhd:
Isn't the notes format completely arbitrary (as in, the user puts whatever they want)?

lenormf:
When `git` starts the editor to edit notes on an object, there are lines commented with a hash sign `#` in the buffer already that describe the object.

I don't know more than that, and this PR simply highlights the comments.

rc/base/git.kak
```diff
@@ -14,8 +18,16 @@ hook -group git-commit-highlight global WinSetOption filetype=git-commit %{
     add-highlighter window/git-commit-highlight/comments/ regex "\b(?:(modified)|(deleted)|(new file)|(renamed|copied)):([^\n]*)$" 1:yellow 2:red 3:green 4:blue 5:magenta
 }
 
+hook -group git-commit-highlight global WinSetOption filetype=git-notes %{
+    add-highlighter window/git-notes-highlight regions
+    add-highlighter window/git-notes-highlight/comments region '^\h*#' '$' group
```
mawww: why make that a group if you only put a fill highlighter inside ?

lenormf: Because there's more data within the comments that could be highlighted similarly to how the commit highlighter does it, I thought somebody would mention it. If we only care about highlighting full comments, I guess we don't need to leave that door open.

lenormf:
Turned the region highlighter into a simple regex one.

PR #2452, opened 2018-09-29
src: Allow single-word expansions to have surrounding whitespace

### LINKED ISSUE #2451: Trim surrounding whitespace when expanding %opt{} or %val{} scopes
Some users put spaces before/after braces/brackets, for stylistic reasons, I think it's fine to support that since the values/options are always single words.

### Steps 

* `:echo %val{ version }`

### Outcome 

```
no such env var:  version 
```

### Expected 

The version is being printed.

lenormf:
Fixes #2451


src/command_manager.cc
```diff
@@ -108,8 +108,18 @@ QuotedResult parse_quoted(Reader& reader, Codepoint delimiter)
 QuotedResult parse_quoted_balanced(Reader& reader, char opening_delimiter,
```
mawww: I think this is the wrong place to do that, just use `trim_whitespace(content)` instead of `content` in the `expand_token` function for relevant types.

lenormf: We're losing completion if we do that.

mawww: I think we can fix that easily in CommandManager::complete by adding a `const bool skip_whitespace = contains({...}, token.type);` then offseting `start` by the amount of whitespaces at the begining of the token, something like:

```
const bool skip_whitespace = contains({...}, token.type);
const ByteCount start = std::min(token.pos + skip_whitespaces ? whitespace_count(token.content) : 0, cursor_pos);
```

lenormf: Why isn't the patch correct? I modified the function that return whatever is between delimiters, that sounds correct and completion still works.

Instead you propose I modify the function that calls the one I modified originally, and also the function that handles completion, it sounds more complicated and forces checking the token type in two separate place.

mawww: I dont like doing this in the parsing code because it moves knowledge of the semantics of specific expansions inside the lexer. I think parse_percent_token should have no knowledge of what each specific token types do and what is/isnt relevant in the text.

Completion and execution both need to know about the semantics of the token type, so it makes more sense to me that they would take care of triming whitespace if needed.

lenormf:
Made the changes requested.

lenormf:
Rebased on master. Fixed a bug with how whitespace were trimmed.

PR #2455, opened 2018-09-30
 src: Implement <a-m> and <a-M>

### LINKED ISSUE #2425: Backward `m`/`M`
Hi,

I use <kbd>m</kbd> to move around a lot, but I have to use <kbd>t</kbd> / <kbd>f</kbd> to move backwards. Now that EOL characters are not treated differently than other characters, <kbd>a-m</kbd> could be freed up (it's basically <kbd>H</kbd><kbd>L</kbd>), and assigned to backwards <kbd>m</kbd>.

<kbd>a-M</kbd> is already free, for backwards <kbd>M</kbd>.

HTH.

lenormf:
<kbd>a-m</kbd> was moved to <kbd>a-_</kbd>, to allow implementing a backward <kbd>m</kbd>.

<kbd>a-M</kbd> was implemented to keep the mappings consistent.

Closes #2425

### DIFF
--- a/src/selectors.cc
+++ b/src/selectors.cc
@@ -220,6 +220,7 @@ select_to_first_non_blank(const Context& context, const Selection& selection)
     return {it.coord()};
 }
 
+template<bool forward>
 Optional<Selection>
 select_matching(const Context& context, const Selection& selection)
 {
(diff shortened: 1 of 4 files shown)


src/selectors.cc
```diff
@@ -220,20 +220,31 @@ select_to_first_non_blank(const Context& context, const Selection& selection)
...
         match = find(matching_pairs, *it);
         if (match != matching_pairs.end())
             break;
         ++it;
     }
+    else while (true)
+    {
+        match = find(matching_pairs, *it);
+        if (match != matching_pairs.end()
+            or it == buffer.begin())
```
Screwtapello: I guess it doesn't matter either way, but why did you put the `it == buffer.begin()` inside the `if`, instead of in the `while` condition to match the other case?

lenormf: Because otherwise any opening delimiter at the very beginning of the buffer would not be considered for forward matching (the loop would quit as soon as the iterator reached the beginning of the buffer, without matching it with `matching_pairs`).

PR #2456, opened 2018-09-30
src: Replace include guards with #pragma once

lenormf:
Hi,

This prama is widely supported by compilers, and [presents a few advantages](https://en.wikipedia.org/wiki/Pragma_once#Advantages) in large codebases, among which better readability, potentially fast compilation times, and less error prone.

HTH.

### DIFF
--- a/src/alias_registry.hh
+++ b/src/alias_registry.hh
@@ -1,5 +1,4 @@
-#ifndef alias_registry_hh_INCLUDED
-#define alias_registry_hh_INCLUDED
+#pragma once
 
 #include "safe_ptr.hh"
 #include "string.hh"
@@ -41,5 +40,3 @@ private:
 };
 
 }
-
-#endif // alias_registry_hh_INCLUDED
--- a/src/array_view.hh
+++ b/src/array_view.hh
@@ -1,5 +1,4 @@
-#ifndef array_view_hh_INCLUDED
-#define array_view_hh_INCLUDED
+#pragma once
 
 #include <vector>
 #include <initializer_list>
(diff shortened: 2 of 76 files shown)


lenormf:
I made this PR mostly because it improves code readability, and also because I'm not sure how to accurately measure improvements (dealing with caching etc).

But here is compilation (`make clean all`) before and after the PR:

```
make clean all  119.55s user 8.24s system 99% cpu 2:08.94 total
make clean all  146.17s user 9.94s system 98% cpu 2:38.57 total
```

I also had a look at the Travis linux/`gcc` worker (it might be even less accurate/relevant than compilation times on my system), and the times are as follows:

* before [4 min 8 sec](https://travis-ci.org/mawww/kakoune/jobs/433191840)
* after [3 min 36 sec](https://travis-ci.org/mawww/kakoune/jobs/435183231)

mawww:
This pragma is non-standard, and the include guard pattern is idiomatic enough so that I dont believe it is hindering readability.

I am very surprised that the performance would be different, as both gcc and clang have a special mechanism to detect header guards and treat them specially.

Testing here I get

```
real	0m24.126s
user	2m30.333s
sys	0m6.905s

and 

real	0m24.633s
user	2m30.091s
sys	0m6.904s
```

lenormf:
It's not standard, but it's implemented by both compilers the project supports, and most other popular compilers as well (for a while at that). Why use three lines when a single one does the same job, without any pre-processing hack?

I tried building the project on an older system, the compilation times were identical this time, but like I said performance wasn't the main focus of the PR. I suppose there's not much point to keeping this open if you consider that `#ifndef`s are the same thing.

mawww:
Compilation performance is the only thing that would really make me go against the standard, we avoid using non-standard extensions in Kakoune, I allow non-standard attributes because they are guaranteed to be ignored if unknown.

While I agree they are ugly, I dont find header guards bothering enough to replace them with a non-standard alternative.

PR #2458, opened 2018-09-30
rc git-tools: Allow jumping between hunks

### LINKED ISSUE #2423: Implement `git` commands to jump to the previous/next modification in the buffer
Hi,

It'd be nice to have commands that jump from one hunk to the other, whose lines can already be highlighted with the `:git show-diff` command.

However I think it'd also be better if users didn't have to use `show-diff` in order to be able to jump to a hunk (to make use of the flag option that stores line coordinates), so some kind of refactoring would potentially have to take place.

HTH.

lenormf:
Fixes #2423

### DIFF
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -23,10 +23,10 @@ declare-option -hidden line-specs git_diff_flags
 define-command -params 1.. \
   -docstring %sh{printf 'git [<arguments>]: git wrapping helper
 All the optional arguments are forwarded to the git utility
-Available commands:\n  add\n  rm\n  blame\n  commit\n  checkout\n  diff\n  hide-blame\n  hide-diff\n init\n log\n  show\n  show-diff\n  status\n  update-diff'} \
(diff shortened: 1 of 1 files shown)


lenormf:
Thanks, rebased on eba4002.

lenormf:
Bump!

rc/extra/git-tools.kak
```diff
@@ -146,6 +146,79 @@ Available commands:\n  add\n  rm\n  blame\n  commit\n  checkout\n  diff\n  hide-
         '
     }
 
+    jump_next_hunk() {
```
mawww: Thats around 80 new lines for next/prev, I would be more confortable with less, I think there should be a bit of room for simplification, and hopefully sharing code between the two implementation.

lenormf:
The two functions don't have the same logic even though they both use a loop, I don't think factorizing for the sake of saving on the few lines that could be factorized is really beneficial.

Rebased on master.

PR #2461, opened 2018-09-30
Add mark highlighter

Delapouite:
Hi

Here's a pull request to add a new convenient *mark* highlighter.

## Usage

```
:add-highlighter buffer/ mark reg face
```

## Example

```
:add-highlighter buffer/ mark ^ red
```

Open a file and save interesting selections with `Z` (which default to the `^` register). Move elsewhere in the buffer. You still have a visual clue (here in red) of the saved selections, so later you'll know what will be the effect of pressing `z` to restore them.

## Rationale

_"Out of sight, out of mind"_

In my daily workflow, I found the `Z`, `z`, `<a-z>` and `<a-Z>` are great commands but can be unfortunately neglected because once selections have been stored in a register, it is hard to remember their precise range after a few minutes of doing something else.

This new highlighter offers a nice way to always have a visual indicator of them. In my example the red color is obviously too aggressive, so it's up to each user to find a more subtle face.

I'm convince that the addition of this interactive feedback fits nicely with Kakoune philosophy.

## Precedent

Some great plugins already offers related helpers around the topic "visual feedback for selections":
- https://github.com/occivink/kakoune-phantom-selection
- https://gitlab.com/fsub/kakoune-mark

So I guess the community has an interest in these functionalities. 

## Technicalities

Why not reusing the `ranges` highlighter instead of introducing a new one? The ranges highlighter accepts an option name which must contain a timestamped list of RangeAndString tuples.

The `Z` command saves selections in the form of `buffer_name@timestamp@main_sel_index list_of_ranges`. The two formats are somehow similar but different enough that converting from one to another involved doing cumbersome string parsing/formatting in a shell context. Also this dance needs to be performed each time the targeted register is updated with hooks that do not exist yet.

I strongly believed that a builtin one-liner `add-highlighter buffer/ mark reg face` like the one proposed here is way more convenient and beneficial for everyone.

Thanks

### DIFF
--- a/doc/pages/highlighters.asciidoc
+++ b/doc/pages/highlighters.asciidoc
@@ -89,6 +89,9 @@ from the remaining parameters.
 *line* <number> <face>::
     highlight line *number* with face *face*
 
+*mark* <register> <face>::
+    highlight selections saved in register *register* with face *face*
+
 *regex* <regex> <capture_id>:<face> ...::
     highlight a regex, takes the regex as first parameter, followed by
     any number of face parameters.
--- a/src/highlighters.cc
+++ b/src/highlighters.cc
@@ -1426,6 +1426,9 @@ InclusiveBufferRange option_from_string(Meta::Type<InclusiveBufferRange>, String
 BufferCoord& get_first(RangeAndString& r) { return std::get<0>(r).first; }
 BufferCoord& get_last(RangeAndString& r) { return std::get<0>(r).last; }
 
+BufferCoord& get_first(InclusiveBufferRange& r) { return r.first; }
+BufferCoord& get_last(InclusiveBufferRange& r) { return r.last; }
+
 void option_list_postprocess(Vector<RangeAndString, MemoryDomain::Options>& opt)
 {
     std::sort(opt.begin(), opt.end(),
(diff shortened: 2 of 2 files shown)


Delapouite:
>  In my example the red color is obviously too aggressive, so it's up to each user to find a more subtle face.

On this [forum topic](https://discuss.kakoune.com/t/rgb-blending-for-primaryselection-secondaryselection-face-colors/753/3), alex mentioned this great tool to manipulate colors: https://github.com/sharkdp/pastel

This way, the face chosen for the mark highlighter can piggy-back on an existing one, but be more dimmed.

danr:
I wrote a script to see the navigation options in kak-tree using the range highlighter: https://github.com/ul/kak-tree/issues/12 This would be easier and more convenient to do with a mark highlighter.

One thing that struck me is that the mark has to be highlighted in one and the same colour. There seems to be no way highlight the anchor ends and secondary selections differently. This might be useful to have.

Delapouite:
@mawww I've noticed in your last updated commits that you're currently addressing the "highlighter" problem space. Do you have some time to give a shot at this PR while your thoughts are on this matter?

Note that I'm not really attached to the code it contains, but rather the end functionality. So if you have any idea on how to design it better and more efficiently, feel free to merge your own implementation instead. Thanks.

mawww:
I agree this is a worthwhile feature, and I cannot find a nice way to merge ranges/mark highlighters. It does however look like it would be nice to be able to support the mark style syntax in options as well (as registers are a bit transient and I imagine it would be usefull to save a mark register into an option, that actually leads to something I'd like to explore later on, which is merging registers and options into a single concept)

Screwtapello:
Now that we have the `RegisterModified` hook, I figured I'd have a go at implementing something like this as a plugin rather than in C++. It [turns out](https://gitlab.com/Screwtapello/kakoune-mark-show) to be pretty straightforward.

The biggest difference between my plugin and the functionality of this PR (so far as I can tell) is that my plugin only supports the caret register, not custom registers. The biggest obstacle to that feature is that the `RegisterModified` hook requires both the symbol (`^`) for the hook parameter regex, and the alphabetic name (`caret`) for the hook group. That means the plugin would have to convert register names in both directions, which I think would be a pain.

Kratacoa:
Also interested in this feature!

mawww:
We could change RegisterModified to never use `^` but use `caret` instead. This would be a slight breaking change but I think it would both simplify that use case, and make things generally more consistent.

Because hooks just use a regex for filtering, supporting both ^ and caret means providing multiple filter strings for that hook which seems like a pretty complex change to support this.

PR #2465, opened 2018-10-01
Support bracketed paste mode in Kakoune

Screwtapello:
Many terminals support "bracketed paste" mode, where the terminal sends a special CSI sequence before and after pasted text, so the application can distinguish it from typed text. This PR teaches Kakoune to request this mode, to recognise the sequences, and to handle them as the new keys `<paste:begin>` and `<paste:end>`.

It also adds default mappings for those keys, so that pasting text will automatically disable hooks, preventing auto-indent messes.

Open questions:

- Should these sequences be handled as keys (like most ncurses input) or as hooks (like `FocusIn`/`FocusOut`)?
- I picked key-names with a colon, vaguely after the pattern of the mouse event "key" syntax, which doesn't actually show up anywhere except the parameter string of various hooks. Perhaps a different naming scheme would be better?
- The state machine for parsing CSI sequences that ncurses doesn't support is a bit of a hack: unrecognised sequences are discarded instead of treated as text, it can't support keys with modifiers (like `CSI 15 ; 2 ~` for `<s-f5>`), and we don't always reset `wtimeout()` if we find a valid sequence. I'd be more than happy for somebody with more C++ skill than I to sort it out.
- Should there be default mappings for `<paste:begin>` and `<paste:end>`? If so, should they be built-in like normal-mode commands, or ordinary mappings? If mappings, should they be in `kakrc`, or a separate `.kak` file?

### DIFF
--- a/share/kak/kakrc
+++ b/share/kak/kakrc
@@ -25,6 +25,10 @@ def -params 1 -docstring "colorscheme <name>: enable named colorscheme" \
     fi
 }}
 
+map global normal <paste:begin> '\i'
+map global insert <paste:begin> '<a-;>\i'
+map global insert <paste:end>   <esc>
+
 evaluate-commands %sh{
     autoload_directory() {
         find -L "$1" -type f -name '*\.kak' \
--- a/share/kak/kakrc
+++ b/share/kak/kakrc
@@ -27,7 +27,6 @@ def -params 1 -docstring "colorscheme <name>: enable named colorscheme" \
 
 map global normal <paste:begin> '\i'
 map global insert <paste:begin> '<a-;>\i'
-map global insert <paste:end>   <esc>
 
 evaluate-commands %sh{
     autoload_directory() {
--- a/share/kak/kakrc
+++ b/share/kak/kakrc
@@ -25,8 +25,6 @@ def -params 1 -docstring "colorscheme <name>: enable named colorscheme" \
     fi
 }}
 
-map global insert <paste:begin> '<a-;>\i'
-
 evaluate-commands %sh{
     autoload_directory() {
         find -L "$1" -type f -name '*\.kak' \
--- a/share/kak/kakrc
+++ b/share/kak/kakrc
@@ -25,7 +25,6 @@ def -params 1 -docstring "colorscheme <name>: enable named colorscheme" \
     fi
 }}
 
-map global normal <paste:begin> '\i'
 map global insert <paste:begin> '<a-;>\i'
 
 evaluate-commands %sh{
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -665,7 +665,7 @@ Optional<Key> NCursesUI::get_next_key()
             Codepoint csi_val;
 
             while ((csi_val = wgetch(m_window)) != ERR) {
-                if ('0' <= csi_val && csi_val <= '9')
+                if (is_basic_digit(csi_val))
                 {
                     keycode *= 10;
                     keycode += (csi_val - '0');
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -661,12 +661,27 @@ Optional<Key> NCursesUI::get_next_key()
         const int new_c = wgetch(m_window);
         if (new_c == '[') // potential CSI
         {
-            const Codepoint csi_val = wgetch(m_window);
-            switch (csi_val)
-            {
-                case 'I': return {Key::FocusIn};
-                case 'O': return {Key::FocusOut};
-                default: break; // nothing
+            int keycode = 0;
+            Codepoint csi_val;
+
+            while ((csi_val = wgetch(m_window)) != ERR) {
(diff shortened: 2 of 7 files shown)


src/ncurses_ui.cc
```diff
@@ -661,12 +661,27 @@ Optional<Key> NCursesUI::get_next_key()
...
-            switch (csi_val)
-            {
-                case 'I': return {Key::FocusIn};
-                case 'O': return {Key::FocusOut};
-                default: break; // nothing
+            int keycode = 0;
+            Codepoint csi_val;
+
+            while ((csi_val = wgetch(m_window)) != ERR) {
+                if ('0' <= csi_val && csi_val <= '9')
```
lenormf: `if ('0' >= csi_val && csi_val <= '9')` ?

Screwtapello: That condition would be the same as `csi_val <= '0' && csi_val <= '9'`, and since `'0' < '9'`, that would be the same as `csi_val < '0'`, which isn't strict enough to match ASCII digits.

eraserhd: `std::isdigit()` seems to accept only ASCII digits, anyway.  Unless in Microsoft 1252 codepage on Microsoft compilers, but we only support UTF-8.

eraserhd: (Link https://en.cppreference.com/w/cpp/string/byte/isdigit)

eraserhd: And... ignore all that, because it's a Codepoint.  Ne'ermind.

lenormf: I misread the line, I've made the mistake so many times that I see it everywhere now. But the fact that the operands are switched makes it harder to read somehow.

Screwtapello: As a Python programmer, I'm used to saying "value X is between A and B" with `A <= X <= B` but of course that doesn't work in C++.

Now that I look, I see `unicode.hh` has an `is_basic_digit()` function that's basically that, so I guess I should use it.

mawww:
While I do like adding support for braketed paste as a quality of life improvement, I dont really like exposing terminal implementation details to the user like that. I am wondering if another approach would be better:

Add support for "pasting" at the UserInterface level, where the UI can send to Kakoune core a "Paste" command, instead of a "Key" command, with the actual pasted string given. A UI paste command would just insert the pasted text before current selection/cursor with no regard to the current mode.

That might be overly complex compared to your proposal where we just expose the bracketed paste to user level and map it. What do you think ?

share/kak/kakrc
```diff
@@ -25,6 +25,10 @@ def -params 1 -docstring "colorscheme <name>: enable named colorscheme" \
     fi
 }}
 
+map global normal <paste:begin> '\i'
+map global insert <paste:begin> '<a-;>\i'
+map global insert <paste:end>   <esc>
```
mawww: Those should not be mappings, but built-ins in the C++ code, the user can override the mappings as we do for any other commands.

Screwtapello:
I have rebased on master, and moved the default behaviour of `<paste:begin>` and `<paste:end>` from mappings to C++. Unfortunately, it doesn't work quite right: pasting in insert mode *should* perform the equivalent of `<a-;>\i`, and it *mostly* does, except that when the paste is over no-hooks mode remains in effect forever.

When I handle `<paste:begin>` from normal mode, I can disable hooks just the way that the `\` key does so they get automatically re-enabled afterward. However, when I handle `<paste:begin>` from insert mode, the first thing I do is create a new normal mode, and I don't know how to "reach inside" it and tell it to re-enable hooks afterward.

Screwtapello:
Thank you again for reviewing my code, and giving me a suggestion on how to proceed.

However, I confess that I only got this far by copy/pasting chunks of code that seemed to do more or less the right thing; I don't really know where to put an `Optional<ScopedSetBool> m_disable_hooks_for_pasting` or how to ensure it gets emplaced and reset at the correct times.

mawww:
`Optional<ScopedSetBool> m_disable_hooks_for_pasting` should be a field of InputModes::Insert, call `emplace(context.hooks_disabled())` on it on <paste:begin> and reset on it on <paste:end>. The optionally contained ScopedSetBool will disable the NestedBool given to it (here context.hooks_disabled()) for its lifetime, so from emplace until either a call to reset, or the destruction of the Optional itself, which will happen when the insert mode is destructed. That should give the behaviour we want for pastes while in insert mode.

But thinking more about it, it will be complex to make paste:end work nice when pasting from normal mode this way, as we need to tell the insert mode to exit on paste:end. I need to think a bit more about how to implement that nicely.

pickfire:
Any news on this?

Screwtapello:
Just to be clear: I think I've done as much work as I can here. Perhaps I could make more progress if I spent some concerted time and effort on it, but originally I just wanted to see how difficult it would be to support bracketed-paste, and when I got something that worked I made a PR just for the heck of it. Personally, I don't mind having to use `\i` occasionally, or even just manually fixing indentation after pasting.

If somebody else wants to push this forward, be my guest.

PR #2467, opened 2018-10-01
Made share/kak/kakrc look for kakrc.local in current directory

laelath:
I'm not too sure why it would always source kakrc.local from the runtime directory as that would have the same effect as the current share/kak/kakrc file does. I assume the intended behavior is sourcing this file in the current directory if it exists, and this commit it restoring that.

### DIFF
--- a/share/kak/kakrc
+++ b/share/kak/kakrc
@@ -39,8 +39,8 @@ evaluate-commands %sh{
         autoload_directory ${kak_runtime}/autoload
     fi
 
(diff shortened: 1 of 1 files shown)


occivink:
I think the intent is to have some system-wide configuration, which might be useful in a multi-user environment. Since it's not part of the kakoune install, it can be modified without touching the package files.

Regardless, sourcing an arbitrary file is a potential security risk so we don't want to do it by default, the user can set it themselves if they want to (I do).

laelath:
Ah, good point about it being a system wide configuration file

PR #2468, opened 2018-10-01
Added ansi keymap file

laelath:


mawww:
I am not sure of the value of this keymap, keeping one keymap file in sync with the actual bindings is already quite error prone, and it seems to me the conversion iso/ansi is pretty simple to do. Is there something I missed that makes having a second keymap file important ?

Note that I dont really mind which layout the keymap file is in, if the ansi layout is more popular, I am happy to replace the existing keymap with this one, its mostly maintaining two almost equal files that I am reserved about.

laelath:
Makes sense. I don't have strong opinions about whether it should be ISO or ANSI, I mainly just made this because I was bored. On a related note, it might be helpful to add a way to view the keymap from within kakoune for new users.

PR #2469, opened 2018-10-02
Add '<' and '>' to argument text-object selector

Delapouite:
Hi

This is handy for situation involving template, like in C++. Example from Kakoune source code:

```c++
template<typename Iterator, typename Container>
Optional<std::pair<Iterator, Iterator>>
```

### DIFF
--- a/src/selectors.cc
+++ b/src/selectors.cc
@@ -744,8 +744,8 @@ select_argument(const Context& context, const Selection& selection,
     auto classify = [](Codepoint c) {
         switch (c)
         {
-        case '(': case '[': case '{': return Opening;
-        case ')': case ']': case '}': return Closing;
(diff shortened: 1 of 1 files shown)


Screwtapello:
This is definitely useful for C++ templates, but the edge-case is when the argument is an expression involving comparison:

    some_func(a < b, c > (d + e) * f);

Put the cursor on `a` and hit `<a-i>u`. Will it select `a < b`, or the whole contents?

I guess if it turns out to have a too-high error-rate in practice, it can always be removed again.

Delapouite:
In light of https://github.com/mawww/kakoune/issues/3046 and https://github.com/mawww/kakoune/issues/2505 , closing this PR since the argument text-object is already brittle enough as is.

PR #2477, opened 2018-10-05
jump command

eraserhd:
Rationale:

I would like jump to behave differently for me, and there's no single command
to override, so I've made one to wrap up the existing behavior.

For reference, my personal override will be somewhat like Vim's
`switchbuf=usetab`: if the file is open in any client, make that client
jump and switch to it, and fall back on jumpclient otherwise.

### DIFF
--- a/rc/core/grep.kak
+++ b/rc/core/grep.kak
@@ -40,16 +40,12 @@ hook global WinSetOption filetype=(?!grep).* %{
     remove-hooks buffer grep-hooks
 }
 
-declare-option -docstring "name of the client in which all source code jumps will be executed" \
-    str jumpclient
-
 define-command -hidden grep-jump %{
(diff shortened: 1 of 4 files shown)


lenormf:
I understand why this was done, but I don't think it's worth merging.

The idiomatic way of distributing a feature across files is to duplicate the code, not to factorize it and share an API. That way users are generally more or less guaranteed the single file they load from their `autoload` will work without having to load another.

The other thing is that this change impacts only three files, the modifications are so small and specific to your workflow that it's probably best you implement them in your user configuration.

eraserhd:
OK, I'll keep this in my fork for now until the dependency stuff is resolved.

eraserhd:
Referencing issue #1767 for, well, reference.

PR #2479, opened 2018-10-06
Simplify surround selection

eraserhd:
This has the same effect with fewer conditions, and I think it also
specifies the intent more closely this way.


mawww:
Doing a quick test on that, it breaks repeating `{{` for example (try hitting `{{` a few times, without this change it will select the parent ones, with it it will not)

Delapouite:
I've added a test this use case: https://github.com/mawww/kakoune/pull/2531

eraserhd:
@Bruno Thanks!  I haven’t forgotten about this and I’m hoping the refactor
is salvageable.

eraserhd:
OK, this works now.  The commit was squashed, and this also merges @Delapouite's test, but below is the fix to the original PR.

```diff
--- a/src/selectors.cc
+++ b/src/selectors.cc
@@ -393,9 +393,10 @@ select_surrounding(const Context& context, const Selection& selection,
 
     auto res = find_surrounding(buffer, pos, opening, closing, flags, level);
 
-    // When the selection did not change, select its parent
+    // If the ends we're changing didn't move, find the parent
     if (res and not (flags & ObjectFlags::Inner) and
-        res->first.coord() == selection.min() and res->second.coord() == selection.max())
+        (res->first.coord() == selection.min() or not (flags & ObjectFlags::ToBegin)) and
+        (res->second.coord() == selection.max() or not (flags & ObjectFlags::ToEnd)))
         res = find_surrounding(buffer, pos, opening, closing, flags, level+1);
 
     if (res)
```

PR #2484, opened 2018-10-09
Be specific about the units of ranges and cursor positions.

Screwtapello:
This adds things I learned from ul/kak-lsp#98 to the documentation.

### DIFF
--- a/doc/pages/options.asciidoc
+++ b/doc/pages/options.asciidoc
@@ -86,13 +86,19 @@ are exclusively available to built-in options.
     `declare-option`.
 
 *range-specs*::
-    a list of a pair of a buffer range (`<begin line>.<begin column>,
-    <end line>.<end column>` or `<begin line>.<begin column>+<length>`)
-    and a string (separated by `|`), except for the first element which
-    is just the timestamp of the buffer. When the `update-option` is
-    used on an option of this type, its ranges gets updated according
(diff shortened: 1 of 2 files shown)


doc/pages/options.asciidoc
```diff
@@ -86,13 +86,19 @@ are exclusively available to built-in options.
...
-    to all the buffer modifications that happened since its timestamp.
-    See <<highlighters#specs-highlighters,`:doc highlighters specs-highlighters`>>)
+    a timestamp (like `%val{timestamp}`,
+    see <<expansions#value-expansions,`:doc expansions value-expansions`>>)
+    followed by a list of range descriptors. Each range descriptor must use
+    the syntax `a.b,c.d|string` or `a.b+length|string`, where _a_ is the line
+    containing the first character, _b_ is the number of bytes from the start
+    of the line to the first byte of the first character, _c_ is the line
+    containing the last character, _d_ is the number of bytes from the start of
+    the line to the first byte of the last character, _length_ is the length of
```
Screwtapello: Actually, now that I think about it, I wonder if ranges *must* be specified in forward (`<a-:>`) direction.

mawww: They dont, but its still the preferred way to specify them. So I think the doc you wrote is good enough.

PR #2490, opened 2018-10-12
Update spec file to the new release

jkonecny12:
Thanks spacekookie for pointing this out :)

### DIFF
--- a/contrib/kakoune.spec
+++ b/contrib/kakoune.spec
@@ -1,5 +1,5 @@
 Name:           kakoune
-Version:        2018.04.13
+Version:        2018.09.04
 Release:        1%{?dist}
 Summary:        Vim inspired editor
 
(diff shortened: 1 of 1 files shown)


jkonecny12:
Maybe we should think about a script (Makefile) to make a new release and update all these files automatically on a release.

lenormf:
Isn't there an official package by now? The project can't maintain this kind of files forever now that several official releases have been produced, users will almost constantly be out of date.

jkonecny12:
Sadly it is not. It is on my TODO list but I have it a little bit bigger... :(

jkonecny12:
I will try to make it a priority and keep this file up to date.

lenormf:
Thanks, it would be really convenient to distribute the work load!

PR #2493, opened 2018-10-14
better tmux-repl: support multiple repl window

co-dh:
You can have 1 repl window per session.

### DIFF
--- a/rc/extra/tmux-repl.kak
+++ b/rc/extra/tmux-repl.kak
@@ -1,6 +1,8 @@
 # http://tmux.github.io/
 # ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
 
+declare-option str repl_pane_id
+
 hook global KakBegin .* %sh{
     if [ -n "$TMUX" ]; then
         VERSION_TMUX=$(tmux -V | cut -d' ' -f2)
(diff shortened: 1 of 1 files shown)


rc/extra/tmux-repl.kak
```diff
@@ -1,6 +1,8 @@
 # http://tmux.github.io/
 # ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
 
+declare-option str repl_pane_id
```
eraserhd: I'd suggest adding `-docstring` and prefix with `tmux_`, since this pane id doen't make sense for other repl mechanisms.

eraserhd:
This seems pretty close to what I was going to attempt for the same reason.  Looks good to me.

rc/extra/tmux-repl.kak
```diff
@@ -22,7 +24,7 @@ hook global KakBegin .* %sh{
 }
 
 define-command -hidden -params 1..2 tmux-repl-impl %{
-    evaluate-commands %sh{
+    set-option global repl_pane_id %sh{
```
Delapouite: In `$TMUX` is not defined, it results in 
```
set-option global repl_pane_id echo -markup '{Error}This command is only available in a tmux session
```
which would fail.

rc/extra/tmux-repl.kak
eraserhd: This reminds me... one lesson I learned in my kakrc for using fzf-tmux, is that kak should test $kak_client_TMUX and run tmux like `TMUX=$kak_client_TMUX tmux display-message ...`.  Otherwise, if you connect to a tmux daemonized from outside the tmux, you don't have access to window commands.

co-dh:
The idea is already merged in #2695

PR #2500, opened 2018-10-17
[src] Added ClientCreate and ClientClose hooks

topisani:
I know these to some extend seem a bit weird, since there is no `client` scope, but i think they still have their place, for example to remove tools/doc clients when exitting the main client, which is especially useful when those clients are hidden.


Delapouite:
Hooks have been refactored from string to enum in https://github.com/mawww/kakoune/commit/dfc11d1c43b898394df8fda52a8079b34058de27 and this PR is unfortunately in conflict.

topisani:
Conflicts should be resolved. What is the opinion on this?

mawww:
Could you rebase and resolve the conflict directly in the commit ?

Issue #2070, opened 2018-05-22
[Suggestion] modelinefmt changes


occivink:
Currently the modeline is defined with the `modelinefmt` string but there are issues with spacing potentially empty elements. Even with the default value, if `{{mode_info}}` is empty there is a pesky double space. And any plugin that defines an option to be shown there also experiences the same issue, and has to work around it by handling spacing inside the option.

I suggest changing its type to `str-list` and when generating the modeline, join all non-empty elements (after expansion) with a space. This could also be used to split `{{mode_info}}` into separate atoms for mode, parameters and selections to be more flexible.

Delapouite:
Talking about the content of `{{mode_info}}`, here's a non submitted PR that I've been using for a while:
https://github.com/Delapouite/hakoune/commit/978b9b9ddc3093558409f7a0c892359c7aece6fe

What it does is removing "selections info" from `{{mode_info}}` and introduce a new `%val{selection_count}`.
This way the use is free to use something along `{StatusLineInfo}#%reg{#}/%val{selection_count}{Default}` in `modelinefmt`. This is really useful while performing an interactive search to know how many selections are being created.

Pantoofle:
Can't you use something like this : 
```     
items=(
    	"%val{bufname}"
    	"%opt{filetype}"
    	"%val{cursor_line}:%val{cursor_char_column}"
    	"{{context_info}} {{mode_info}}"
    	"%val{client}@[%val{session}]"
    	)
	
    line=""

    for t in "${item[@]}"; do
       # test if the item is empty. Add more complex tests if needed
        test -z "${t}" && continue
        # add it to the line. Special case for first element 
        test -n "${line}" && line="${line} | ${t}" || line="${t}"
    done

    echo set global modelinefmt "'${line}'" 
```

PR #2503, opened 2018-10-20
Expose selection(s) length with a value

occivink:
I have a couple of plugins ([expand](https://github.com/occivink/kakoune-expand), [vertical-selection](https://github.com/occivink/kakoune-vertical-selection)) where I need the selections' length. I currently use `exec s.<ret>; echo %reg{#}` as a workaround, but that can get slow with large selections.

### DIFF
--- a/doc/pages/expansions.asciidoc
+++ b/doc/pages/expansions.asciidoc
@@ -265,6 +265,14 @@ The following expansions are supported (with required context _in italics_):
     unquoted list of the ranges of all selections, in the same format as
     `%val{selection_desc}`
 
+*%val{selection_length}*::
+    _in window scope_ +
+    length (in codepoints) of the main selection
+
+*%val{selections_length}*::
+    _in window scope_ +
+    unquoted list of the lengths (in codepoints) of the selections
+
 *%val{session}*::
     name of the current session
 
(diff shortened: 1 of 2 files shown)


doc/pages/expansions.asciidoc
```diff
@@ -252,6 +252,14 @@ The following expansions are supported (with required context _in italics_):
...
     quoted list of the contents of all selections
 
+*%val{selection_length}*::
+    _in window scope_ +
+    length (in codepoints) of the main selection
+
+*%val{selections_length}*::
+    _in window scope_ +
+    comma-separated list of the lengths (in codepoints) of the selections
+
```
Screwtapello: Currently, the expansions are listed in approximately alphabetical order, so you probably want to put these below `selection_desc` and `selections_desc`.

occivink: Good points

Delapouite:
If I may suggest a third useful related value. Example (https://github.com/mawww/kakoune/issues/2070)

```cpp
{
        "selection_count", false,
        [](StringView name, const Context& context, Quoting quoting) -> String
        { return to_string(context.selections().size()); }
}

occivink:
Sounds reasonable, and more explicit than using `"%reg{#}"`

PR #2506, opened 2018-10-21
gruvbox: show comments in italic

maximbaz:
Many editors follow a convention of showing comments in italic, would be great if gruvbox theme also did so by default.

I didn't touch the rest of the themes because I wasn't sure it is desirable, let me know if you want me to change all of them. Desertex theme for example already uses italic font.

/cc @laelath because you recently edited this theme too 😉

### DIFF
--- a/colors/gruvbox.kak
+++ b/colors/gruvbox.kak
@@ -33,7 +33,7 @@ evaluate-commands %sh{
         face global keyword   ${red}
         face global operator  ${fg}
         face global attribute ${orange}
-        face global comment   ${gray}
+        face global comment   ${gray}+i
         face global meta      ${aqua}
         face global builtin   ${fg}+b
 


lenormf:
The [original theme](https://github.com/morhetz/gruvbox) doesn't highlight comments in italic though. It's better to have a port that's as close to the original as possible, as users can easily modify faces locally.

maximbaz:
I think making comments italic it is often a property of the editor, and not theme itself (take VSCode for example, not only it shows comments in italic, it even _fakes_ italic font if such font doesn't exist). I agree I can modify the face locally, but I've seen italic comments so much recently that I thought it would make sense to include it in the kakoune's distribution as well.

net:
I am 👎🏼 on this as a standard for kakoune color schemes. Many monospace fonts are harder to read in oblique, and setting the italic property locally is trivial.

I believe italic comments are defined by the theme in VSCode, not the editor.

maximbaz:
Unsetting it is just as trivial 😉 

So are you saying that gruvbox in vscode does come with italic font, like kakoune now?

net:
I don't know what VSCode gruvbox you use, but if it has italic text then most likely yes.

The original vim gruvbox has a settings for italic (enabled for comments by default).

> Unsetting it is just as trivial

It is, but I do not think it is a good default—at least not as a rule.

PR #2509, opened 2018-10-21
Improve ubuntu installation instructions

FranzSkuffka:
`pkg-config` was required on 18.10

--depth=1 should accelerate the clone process a lot.


Screwtapello:
`--depth=1` can accellerate the initial clone, but makes every subsequent clone slower, so it might not be a good thing to recommend in general.

+1 for the pkg-config thing, though.

PR #2521, opened 2018-10-24
rc: Replace occurrences of `setsid`

### LINKED ISSUE #2520: x11-new broken on FreeBSD
setsid is used in x11.kak which is not available on FreeBSD.

Removing setsid seems to work, but this is probably not right.

Is there a portable solution to this or do I need to provide a setsid package?

### Steps 

Run `:x11-new` in Kakoune on FreeBSD.

### Outcome 

Nothing happens

### Expected 

Open a new terminal window with current Kakoune session

lenormf:
The `setsid` isn't typically installed on BSD systems, so we replace
it with the usual shell pattern using braces.

Fixes #2520

### DIFF
--- a/rc/base/x11.kak
+++ b/rc/base/x11.kak
@@ -32,7 +32,7 @@ The optional arguments will be passed as arguments to the new client} \
            exit
         fi
         if [ $# -ne 0 ]; then kakoune_params="-e '$@'"; fi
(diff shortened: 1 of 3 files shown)


mawww:
Those setsid are here for a reason, we need to detach those processes from the process group of Kakoune server, because we want them to opt-out of `<c-c>` in Kakoune killing them (well, sending SIGINT to them). I do not think the proposed replacement will do that, and unfortunately I am not aware of any portable way to do it.

lenormf:
Right, closing then.

PR #2533, opened 2018-10-27
doc: complete list of normal commands using default registers

Delapouite:
I tried to be as exhaustive as possible. For instance `n` commands were not explicitly listed, hiding the fact that you can do cool sequences like that:

```
"c/foo<ret>
/bar<ret>
"cN
```


Delapouite:
Thanks for the feedback @Screwtapello . I included your suggestions.

PR #2541, opened 2018-10-30
Truncate incomplete debug command docstring

Delapouite:
I noticed that the `regex` command was missing in the info.
I was about to add it, but then I decided to simplify the message since it's just repeating what's already provided by the auto-completion.
This will prevent future discrepancy.


lenormf:
The completion will only show the current value, not all the ones that it's possible to use, which the docstring should document.

Delapouite:
When I type `:debug<space>` all the completions are provided. I'm not quite sure I understand what you mean by "current value".

lenormf:
I was thinking of the option, nervermind.

Issue #2513, opened 2018-10-22
PushMode and PopMode hooks for fully-general mode-tracking.


Screwtapello:
Originally, Kakoune had four hooks for mode-change events: `InsertBegin`, `InsertEnd`, `NormalBegin` and `NormalEnd`. However, Kakoune does have other modes, and adding "begin" and "end" hooks for all of them would be prohibitive. In fact, it would be impossible, since Kakoune has user-defined modes as well.

In #1772, a new `ModeChange` hook was introduced, aiming to be a general replacement for the previous mode-change hooks. For example, `InsertEnd` could be replaced with a `ModeChange insert:.*` hook. Unfortunately, that didn't quite work out because Kakoune maintains a *stack* of current modes. For example, from normal mode `i` pushes an insert mode, and from the insert mode `<a-;>` pushes a second normal mode. That transition from insert mode to the second normal mode counts as a `ModeChange insert:normal` because the active mode changes, but it does *not* count as an `InsertEnd` because the insert mode doesn't *end*, it just gets temporarily replaced. See also #2122 

To fully generally replace the original events, we should have `PushMode` and `PopMode` events with the same parameter string as `ModeChange`. That is:

- `InsertBegin` can be replaced with `PushMode .*:insert`
- `InsertEnd` can be replaced with `PopMode insert:.*`
- `NormalBegin` can be replaced with `PushMode .*:normal`
- `NormalEnd` can be replaced with `PopMode normal:.*`
- `ModeChange X:Y` can be replaced with hooking both `PushMode X:Y` and `PopMode X:Y`

In the style of the existing hooks documentation:

**PushMode** `X:Y`: triggered when a command in mode X starts a new Y mode; for example, when `<a-;>` temporarily enters normal mode from insert mode, `PushMode insert:normal` will trigger

**PopMode** `X:Y`: triggered when a command in mode X exits that mode, returning to mode Y; for example, when `<esc>` exits insert mode, `PopMode insert:normal` will trigger

Delapouite:
Yes, mawww stated multiple times that the `ModeChange` hook was not that good in the end. `PushMode` and `PopMode` are indeed nice improvements! Thanks.

But I feel that the same ambiguity remain:

```
normal → PushMode → insert
```
Will the commands done during the `PushMode` hook be treated as `normal` or `insert` commands?
The hook seats right in the middle and there's no way to know to which context it belongs.

So maybe we need 4 hooks:

```
[...]

PR #2545, opened 2018-11-01
Added ModePush and ModePop hooks

laelath:
Implemented ModePush and ModePop as proposed in #2513.

### DIFF
--- a/src/hook_manager.hh
+++ b/src/hook_manager.hh
@@ -50,6 +50,8 @@ enum class Hook
     NormalIdle,
     NormalKey,
     ModeChange,
+    ModePop,
+    ModePush,
     RawKey,
     WinClose,
     WinCreate,
@@ -60,7 +62,7 @@ enum class Hook
 
 constexpr auto enum_desc(Meta::Type<Hook>)
 {
-    return make_array<EnumDesc<Hook>, 39>({
+    return make_array<EnumDesc<Hook>, 41>({
         {Hook::BufCreate, "BufCreate"},
         {Hook::BufNewFile, "BufNewFile"},
         {Hook::BufOpenFile, "BufOpenFile"},
@@ -94,6 +96,8 @@ constexpr auto enum_desc(Meta::Type<Hook>)
         {Hook::NormalIdle, "NormalIdle"},
         {Hook::NormalKey, "NormalKey"},
         {Hook::ModeChange, "ModeChange"},
+        {Hook::ModePop, "ModePop"},
+        {Hook::ModePush, "ModePush"},
         {Hook::RawKey, "RawKey"},
         {Hook::WinClose, "WinClose"},
         {Hook::WinCreate, "WinCreate"},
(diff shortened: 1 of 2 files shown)


mawww:
Yeah, I am wondering about that as well, and it seems safer to do it so that hooks can rely on PushMode always being matched by a PopMode.

On the other hand, we do not have a lot of tools to manage nested states: say you have a `[..., normal, insert, normal]` stack to pop, for the two normal modes that are going to be popped, if we want to do something that requires to run on two normal pops instead of one, we would need to somehow distinguish those two pops, which would likely require some kind of stack accessible from the hook to get some state associated with those pops. It seems to me that without those tools, we cannot easily do anything interesting with two pops that we could not do with a single one.

Screwtapello:
I'm trying to think of a way to emulate a stack by means of one-shot PopMode hooks, but if triggering a PopMode hook triggers *all* the PopMode hooks, instead of just the ones added in the most recent PushMode, then yeah, not much use.

On the other hand, if there were a PopThisMode hook that fired when a specific mode was popped from the stack (instead of when any mode of a given type was popped), that'd do it.

At the end of the day, though, I think it's just less surprising for everybody (both scripters, and people working on the Kakoune code-base) if every PushMode is always matched by a PopMode.

mawww:
@laelath That could be an idea, the ModePush/ModePop filter string could be "<prev_mode>:<new_mode>:<stack_level>". Then `hook global ModePush \w+:insert:(\d+) %{ hook -once global ModePop "insert:\w+:%val{hook_param_capture_1}" 'echo matching pop' }` could do the trick.

mawww:
I think we should think a bit about the mode in which the hook is run.

In the proposed implementation, ModePush runs in the pushed mode, while ModePop runs in the parent mode of the popped mode (If I am in mode A, and push then pop mode B, I get a ModePush A:B that runs in mode B, and a ModePop B:A that runs in ModeA).

I am not sure this is what we want, because it breaks symmetry, we cannot undo something we did on ModePush in ModePop, because we are in another mode already.

I think running ModePush in the pushed mode context is correct, and hence I think ModePop should run in the pushed mode just before it gets disabled.

This leads me to another question: the filtering string format. Do we want to have the parent mode there ? I am not sure there is much value in knowing that information, It would be nice to have an explicit use case for that. Actually it probably would be nice to have a real use case for the stack level as well. Maybe just putting the pushed/popped mode name is good enough ?

Opinions ?

Screwtapello:
> I think we should think a bit about the mode in which the hook is run.

I talked about this in https://github.com/mawww/kakoune/issues/2513#issuecomment-431795566. Long story short, it's safer to always run the hook in the "after" mode.

> This leads me to another question: the filtering string format. Do we want to have the parent mode there?

I'm not actually sure what the Kakoune mode state diagram looks like. If it's a star with "normal" in the centre, so you always have to go via "normal" to get from one non-normal mode to another... then yeah, there's not much point in knowing the parent mode. On the other hand, if there's multiple ways to get from one mode to another, it might be important to know which route the user took.

> Actually it probably would be nice to have a real use case for the stack level as well.

It might be useful for a snippet plugin, where inserting a snippet begins a new insert mode and adds a bunch of hooks and mappings to jump between the snippet's fields... hooks and mappings that need to be removed when that specific insert-mode is popped, not just when any old insert-mode is popped.

mawww:
Sorry, I had forgotten about that PR, revisited that and in the end went with a push/pop prefix for the ModeChange hook as ModePush/ModePop would have required to duplicate hooks for simply running commands when a mode gets enabled (regardless on how we got to it).

PR #2547, opened 2018-11-01
Add 'shell' option

eraserhd:
Rationale: On OS X, I don't have a decent way to replace the shell (I get
"Operation not permitted", even as root though I'd rather not much with
the system shell anyhow).  But I'd like to use dash to speed things up,
and the current logic doesn't allow me to put it in the path or
anything.

### DIFF
--- a/src/shell_manager.hh
+++ b/src/shell_manager.hh
@@ -51,10 +51,10 @@ public:
     CandidateList complete_env_var(StringView prefix, ByteCount cursor_pos) const;
 
 private:
-    String m_shell;
-
     ConstArrayView<EnvVarDesc> m_env_vars;
 };
+
+String find_shell();
 
 }
 
--- a/doc/pages/options.asciidoc
+++ b/doc/pages/options.asciidoc
@@ -307,3 +307,8 @@ are exclusively available to built-in options.
     Controls which messages will be displayed in the startup info box, only messages
     relating to a Kakoune version greater than this value will be displayed. Versions
     are written as a single number: Like `20180413` for version `2018.04.13`
+
+*shell* `str`::
+    path to a POSIX shell to use for shell expansion.  On startup, Kakoune sets this
+    to an appropriately executable `sh` found in a minimal path where "all standard
+    POSIX utilities can be found" -- _not_ the user's `$PATH`.
--- a/src/main.cc
+++ b/src/main.cc
@@ -418,6 +418,7 @@ void register_options()
         "set of pair of characters to be considered as matching pairs",
         { '(', ')', '{', '}', '[', ']', '<', '>' });
     reg.declare_option<int>("startup_info_version", "version up to which startup info changes should be hidden", 0);
+    reg.declare_option("shell", "path to a POSIX shell used for shell expansion", find_shell());
 }
 
 static Client* local_client = nullptr;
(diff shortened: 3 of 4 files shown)


src/shell_manager.hh
```diff
@@ -34,6 +34,8 @@ class ShellManager : public Singleton<ShellManager>
 public:
     ShellManager(ConstArrayView<EnvVarDesc> builtin_env_vars);
 
+    static String find_shell();
```
lenormf: I don't think this function should be static, or even part of the `ShellManager` class. You could just leave the implementation in `shell_manager.cc`, and write down a prototype in `shell_manager.hh`.

mawww:
I am afraid by that option, sure we can document it to say it must be a posix shell, but I am afraid its going to be abused to run with fish, oilshell or  something else, and that will lead to subtle (or not so subtle) breakage of scripts.

While I sympathize with the use case, I wonder if we cannot find another way to make overriding the shell  possible without making it easier. (Keeping it "hard" is a good way to ensure people understand what they are doing).

alexherbo2:
How about using `sh` in `PATH` and `/bin/sh` as the falling case?

Screwtapello:
The issue is that the `sh` in `$PATH` may not be a POSIX shell. In particular, Solaris has a pre-POSIX `/bin/sh`. The reason Kakoune does the `confstr(_CS_PATH)` dance is to get a path that's guaranteed to have POSIX-compatible stuff in it.

How about, if `~/.config/kak/custom_shell` exists at startup, that script is used as the shell for that session. It's still pretty easy to do, but if it's not a standard option it will be harder to find. On the other hand, if somebody forgets they've configured Kakoune to use a custom shell and they uninstall that shell, breaking Kakoune, they're probably going to look in `~/.config/kak` and see the `custom_shell` file there. It might even be a broken symlink, which usually gets highlighted in a scary colour, if you're using GNU coreutils.

Bonus points: if the `custom_shell` exists but is not executable, put a warning in `*debug*`.

maximbaz:
I don't think you should introduce weird alternative ways to configure kakoune, name the option `posix_shell` instead of `shell` to make it _very clear_ that it needs to be POSIX-compliant, and let people handle the rest. It's not your fault if someone wants to mess with options, and besides, nobody will be replacing shell with `fish` anyway, people just want to use dash to workaround the absence of lazy loading.

eraserhd:
This doesn't really help lazy loading, since the user's kakrc (which is presumably where you'd have to set your `shell` option) loads after all the autoload scripts.

I'm trying to make my parinfer plugin work faster, though I can't even test this because dash on Mac OS X is super broken.  Let me try installing an older version today.

maximbaz:
Hopefully some kind of this alias will help you to speed up kakoune startup while we wait for lazy loading: 
`alias kak='kak -e "set global posix_shell /bin/dash"'`

maximbaz:
But on the other hand, having an option that changes shell _in the middle of execution_ (so that some scripts are executed with one shell and then others are executed with another shell) sounds a bit dangerous. Because of this, maybe kakoune instead should support some `KAKOUNE_POSIX_SHELL` environment variable that would contain path to the shell to use.

lenormf:
Make the shell configurable through a pre-processing option.

PR #2548, opened 2018-11-02
Add set-face/unset-face CommandHelpers

Delapouite:
Useful to peek the description of a face with a similar name before adding a new one or to confirm that its the right one before removing a face.


src/commands.cc
```diff
@@ -2091,6 +2091,19 @@ static Completions complete_face(const Context& context, CompletionFlags flags,
                      transform([](auto& entry) -> const String& { return entry.key; }))};
 }
 
+static String face_doc_helper(const Context& context, CommandParameters params)
+{
+    if (params.size() < 2)
+        return "";
```
mawww:
```suggestion
        return {};
```

Creates an empty string without having to parse a const char*.

src/commands.cc
```diff
@@ -2091,6 +2091,19 @@ static Completions complete_face(const Context& context, CompletionFlags flags,
...
+    try
+    {
+        auto face = context.faces()[params[1]];
+        return format("{}:\n{}", params[1], indent(to_string(face)));
+    } catch (runtime_error&) {
```
mawww:
```suggestion
    }
    catch (runtime_error&)
    {
```

Delapouite:
Thanks, I've applied your corrections.

PR #2560, opened 2018-11-08
rc c-family: Update C grammar

Muffindrake:
This updates the existing C grammar (and static completion) to include previously-missing types and macros, particularly those part of C11, as well as optional features of the language.

The incentive to do this was the missing highlighting for the types defined in `stdint.h`.

Functions still aren't highlighted, but some function-like macros are set to be highlighted the same as regular values, which include object-like macros.

Brace expansion is heavily used, which can be replaced by the expanded strings if that's desired.


lenormf:
Using subshells to avoid writing words with the same prefix/suffix is pretty brutal, the loading time on this file is going to explode.

Also I'm not sure the syntax is going to be supported by all shells, so it's probably best we stick to hardcoded, as unelegant it may be.

Muffindrake:
Fair enough.

PR #2561, opened 2018-11-08
Fix markdown list highlighter

### LINKED ISSUE #2546: Region highlighter breaks regex highlighter
markdown issue

### Steps 
Open a markdown file with a list (either using '*'s or '-'s) and backticks in a list element.

### Outcome 
List highlighting stops after the backticks.

### Expected 
The entire list element is highlighted (highlighting stops after multiple new lines or a non-whitespace character at the first column).

The relevant highlighters are:
```
add-highlighter shared/markdown/codespan region -match-capture (`+) (`+) fill mono
# This one is modified from what is in rc/base to fix multi-line highlighting
add-highlighter shared/markdown/content/ regex ^\h*([-\*])\h+[^\n]*(\n\h+[^-\*][^\n]*)* 0:list 1:bullet
```

Disabling the backtick highlighter results in the correct list highlighting.

lePerdu:
This changes the markdown highlighter to make list items (along with code blocks) match on the top level so that they are not broken by other syntax elements (in particular, inline code).

Fixes #2546


lePerdu:
After some testing, I found a bunch of weird problems:
kakonue_markdown_problems

I feel like I'm missing something about how highlighters in kakoune work. Could someone well-versed in them tell me if any of these are incorrect?
1. `region` highlighters split up code so that the delegated highlighter can only match inside the region
2. `region` highlighters end at the first occurrence of the end marker (when `-recurse` is not used)
3. `\A` and `\z` mean the beginning and end of the current region, regardless of what `regex` and other highlighters match
4. `^` and `$` match the beginning and end of the region in addition to the beginning and of the lines

JJK96:
The first 3 points you stated are correct as far as I can see.

For point 4 and 5 I conducted some tests:

test.kak:
```
add-highlighter shared/test regex ^a.*z$ 0:Error
add-highlighter shared/test1 regex \Aa.*z\z 0:Error
add-highlighter shared/test2 regex \A\(a.*z\)\z 0:Error
add-highlighter window/test regions
add-highlighter window/test/test region _ _ ref test
add-highlighter window/test/test1 region \[ \] ref test1
add-highlighter window/test/test2 region \( \) ref test2
```

Buffer:

```
_
az
_
_az_

# so ^ an $ only match beginning and end of line, not beginning and end of region.

[az]
[
az
]

(az)

# so \A and \z do not mean the beginning and end of the region.
```

Source `test.kak` with above buffer. You will see only the first `az` highlighted.

JJK96:
Your PR does more than stated in the description, it seems you also change the way headers are highlighted.

lePerdu:
Didn't mean to have the header changes in there. I took them out now.

lePerdu:
It seems that `\A` and `\z` have different semantics when matching in regions than when matching in regex highlighters.

---

With this highlighter:
```
add-highlighter window/test regions
add-highlighter window/test/toend region start \z fill Error
```

Only the line containing `start` is highlighted (the `\n` at the end of the line is also highlighted) here:
```
[not highlighted]
start [highlighted]
[not highlighted]
```

If `$` is used instead of `\z`, the only difference is that the new-line at the end of the line containing `start` is not highlighted.

---

With this highlighter:
```
add-highlighter window/test regions
add-highlighter window/test/fromstart region \A end fill Error
```

The only part of this buffer not highlighted is the text after `end` in the same line
```
[highlighted]
end [not highlighted]
[highlighted]
```

---

With this highlighter (it should be equivalent to the first one):
```
add-highlighter window/test regex 'start.*?\z' 0:Error
```

Everything until the first `start` is highlighted (as expected)
```
[not highlighted]
start [highlighted]
[highlighted]
```

---

With this highlighter (it should be equivalent to the second one):
```
add-highlighter window/test regex '\A.*end' 0:Error
```

Everything after the first `end` is not highlighted (as expected)
```
[highlighted]
end [not highlighted
[not highlighted]
```

---

This difference in behavior between the region and regex highlighters that should otherwise be equivalent seems like a bug to me, and if it's a feature, it should be better documented (or changed). The current regex documentation says that `\A` and `\z` match at the "subject" start and end, but it doesn't really specify what "subject" means.

lePerdu:
@mawww Wait, sorry, this PR was not ready for merging; it still has the problems outlined in the first comment. I guess I should have marked it as such, but this is a WIP. Same goes for #2546, which is not really fixed.

PR #2563, opened 2018-11-09
doc: fix misc typos for source and region highlighter

Delapouite:


src/commands.cc
```diff
@@ -1304,7 +1304,8 @@ const CommandDesc debug_cmd = {
 const CommandDesc source_cmd = {
     "source",
     nullptr,
-    "source <filename> <params>...: execute commands contained in <filename>",
+    "source <filename> <params>...: execute commands contained in <filename>\n"
+    "parameters are available in the sourced script through the `arg` values",
```
Screwtapello: It may not be obvious what "the arg values" are here; how about we be more explicit:

> ...in the sourced script as `%arg{0}`, `%arg{1}`, ...

PR #2566, opened 2018-11-10
website: add link to discuss.kakoune.com in navbar

Delapouite:
Despite the small size of our community, I think that the initiative launched by @robertmeta has proven successful with already many interesting bits of shared knowledge about kakoune.

Those links will further increase the visibility of this discussion board for new comers seeking help.

### DIFF
--- a/gallery.html
+++ b/gallery.html
@@ -68,6 +68,12 @@
                             <span tid="navbar:plugins">Plugins</span>
                         </a>
                     </li>
+                    <li class="nav-item">
+                        <a class="nav-link text-nowrap" href="https://discuss.kakoune.com">
+                            <i class="fa fa-fw fa-comments"></i>
+                            <span tid="navbar:forum">Forum</span>
+                        </a>
+                    </li>
                     <li class="nav-item dropdown">
                         <a class="nav-link dropdown-toggle" href="javascript:void(0)" id="navbarLanguages" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
                             <i class="fa fa-flag"></i>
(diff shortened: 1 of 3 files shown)


lenormf:
As this is not an official media, I'd rather we put the link into a "Community" dropdown. We could also put the Wiki link into that same dropdown menu.

Delapouite:
As this is not an official media

mawww is participating on it. By hiding it behind a dropdown, it complicates discoverability

lenormf:
It's being maintained by a member of the community, making it a community site.

Also when more sites ever pop-up (btw there's a sub-reddit that could be added to the list as well), we'll need to group them together, as we can't have all of them occupy a full item in the navbar.

robertmeta:
@lenormf what changes would in your mind transition it from community maintained to a more official place?  Mawww is already an admin on the site (I would welcome you as one as well if interested).  I can setup the domain and any resources that are applicable to fall to @mawww in a pinch.  

I think it is important to pick one among the many sites to be the heart of the **user** community.  I think this is important for the growth of Kakoune among users.  I have been a moderator of the #vim channel on freenode for going on two decades, as well recently (last few years) the vim subreddit.  While I love both of those mediums, I think both make a poor heart of a community due to various things they lack, from topic splitting to mailing list mode to time-shifting.

robertmeta:
@lenormf thank you for your clarification, I don't find it particularly compelling but I appreciate you making your POV clear.

mawww:
I do not have a strong opinion on that menu, both options are fine by me, and I'd rather not have to make an arbitrary decision.

In general I would tend to favor a "self organizing community", in the sense that I dont think I should bless anything in particular, I try to follow discuss, reddit, irc... Thinking of it, I might have a preference for the community menu as we can add links to all of those (irc in particular).

PR #2570, opened 2018-11-12
python: add async/await keywords highlighting

Roger:


maximbaz:
Since this list of keywords is generated by `python3 -c 'import keyword; print("\n".join(keyword.kwlist))'`, can you pls add them after `assert` to maintain the order?

Roger:
> Since this list of keywords is generated by `python3 -c 'import keyword; print("\n".join(keyword.kwlist))'`, can you pls add them after `assert` to maintain the order?

yeah, done

PR #2573, opened 2018-11-13
Prototype lazy loading implementation

laelath:
This is a test implementation of lazy loading. For this implementation I tried to keep as much of the existing structure as possible, and keep everything in kak scripts. I added a new file called `filetypes.kak` that includes all the hooks for changing filetype, as well as defining a `str-to-str-map` that contains a mapping of filetypes to support files.

Testing on my raspberry pi, the previous start up time for opening a C++ file was around 2.5 seconds, dropping to 1.2 with lazy loading.


laelath:
Sub highlighters with generic code highlighting are currently broken, and I'm not sure what the best way to fix them would be. ie. Markdown code regions like \`\`\`rust don't show highlighting unless you manually lazy-load rust.

laelath:
Another possible issue is that there's no idea of dependencies, so all dependencies have to be manually loaded. For example, `hbs` files need `html`, which needs `css` and `javascript`, so currently the `hbs` detection hook needs to lazy-load all three of these dependencies.

mawww:
I dont like the fact this introduces a central file that coordinates loading, I believe this should not be in the base distribution. If we go with lazy loading by default, I'd prefer @maximbaz approach, which is less intrusive and keeps each filetype support self-contained.

laelath:
However, when it comes to lazy-loading we still need some place to put the hooks that detect the file type, I think the only two options are to put all the hooks in a single file (as I did for this pull request, mainly because it was easier) or have each language have a separate file that is auto loaded containing only the detection hooks.

mawww:
Closing this PR as I do not plan to go with this centralized approach for lazy loading (if we ever get to lazy loading by default). Thanks @laelath for taking the time to experiment with this.

Issue #1038, opened 2016-12-19
equivalent of vim's expandtab


jtmarmon:
I'd like to be able to indent `indentwidth` when I hit the tab option. right now my successful workaround is to add the following to `kakrc`:
`map global insert <tab> <space><space>`

However, considering this seems like a relatively popular feature in tabless codebases, it may be worth considering as an officially supported setting.

leemachin:
This might be indirect, but there is built-in [editorconfig](http://editorconfig.org) support, which allows you to configure indentation on a per-file/per-project basis.

[This wiki page](https://github.com/mawww/kakoune/wiki/EditorConfig) explains how to enable it automatically.

casimir:
Personally, I have this in my `kakrc`:
```
map global insert <tab> '<a-;><gt>'
map global insert <backtab> '<a-;><lt>'
```
It does the same but considers `%opt{indentwidth}`, inserting the right number of `<space>`'s or a `<tab>` in insert mode. What @leemachin proposed is a good complement too.

PS: for whoever is wondering (I did) it's not a problem even if you use `<tab>` for completion canditate selection, kakoune is smart enough to do both

doppioandante:
@casimir I will put those in the wiki if you don't mind.

doppioandante:
Just for the record, by pressing @ on a selection tabs will be converted to the number of spaces specified by the `tabstop` option.

casimir:
@doppioandante go ahead, IIRC I snatched these on IRC

mawww:
The canonical way to do that would be to do `hook window InsertKey \t %{ exec -draft h@ }` that is, add a hook that after inserting a tab in insert mode will use the `@` command to convert it to spaces.

kballard:
@mawww That hook doesn't work for me. It doesn't ever seem to fire. In fact, I can't seem to get an `InsertKey` hook to fire on tab, space, return, backspace, forward delete, home, end, etc. I don't know why none of the whitespace keys do it, and as for the others, I would have assumed the difference between `InsertKey` and `InsertChar` is the former gets the keys that don't insert characters (such as delete).

I tested using `hook window InsertKey .? %{ echo "key: '%val{hook_param}'" }` (the regex `.?` is because I can't figure out how to provide an empty regex, so I wanted something that would match anything, including the empty string).

kballard:
[...]

PR #2583, opened 2018-11-19
Added expandtab, noexpandtab and smarttab features

andreyorst:
This code addresses already closed issue #1038, however I don't think that this issue was fixed.

The `expandtab.kak` file adds three commands to use:
- `noexpandtab` - use `tab` character for everything.  
  <kbd>Tab</kbd> will insert `\t` character, and <kbd>></kbd> will use `\t` character when indenting. Aligning cursors with <kbd>&</kbd> uses `\t` character.
- `expandtab` - use `space` character for everything.  
  <kbd>Tab</kbd> will insert `%opt{tabstop}` amount of spaces, and <kbd>></kbd> will indent with spaces.
- `smarttab` - indent with `tab` character, align with `space` character.  
  <kbd>Tab</kbd> will insert `\t` character if your cursor is inside indentation area, e.g. before any non-whitespace character, and insert spaces if cursor is after any non-whitespace character. Aligning cursors with <kbd>&</kbd> uses `space`.
The demo can be found in [issue comment](https://github.com/mawww/kakoune/issues/1038#issuecomment-439716724).

I've added this to `rc/core` because it contains most of default things, while `rc/extra` and `rc/base` provide mostly language support. This is really basic functionality that is shipped with most editors, and since it is implemented with hooks it works like most other builtin features.

### DIFF
--- a/rc/core/expandtab.kak
+++ b/rc/core/expandtab.kak
@@ -1,0 +1,38 @@
+define-command -docstring "noexpandtab: use tab character to indent and align" \
+noexpandtab %{
+    remove-hooks global noexpandtab
+    hook -group noexpandtab global NormalKey <gt> %{ try %{
+        execute-keys -draft "<a-x>s^\h+<ret><a-@>"
+    }}
+    set-option global aligntab true
+    remove-hooks global expandtab
+    remove-hooks global smarttab
+}
+
+define-command -docstring "expandtab: use space character to indent and align" \
+expandtab %{
+    remove-hooks global expandtab
+    hook -group expandtab global InsertChar '\t' %{ execute-keys -draft h@ }
+    hook -group expandtab global InsertKey <backspace> %{ try %{
+        execute-keys -draft <a-h><a-k> "^\h+.\z" <ret>I<space><esc><lt>
+    }}
+    set-option global aligntab false
+    remove-hooks global noexpandtab
+    remove-hooks global smarttab
+}
+
+define-command -docstring "smarttab: use tab character for indentation and space character for alignment" \
+smarttab %{
+    remove-hooks global smarttab
+    hook -group smarttab global InsertKey <tab> %{ try %{
+        execute-keys -draft <a-h><a-k> "^\h*.\z" <ret>
+    } catch %{
+        execute-keys -draft h@
+    }}
+    hook -group smarttab global NormalKey <gt> %{ try %{
+        execute-keys -draft "<a-x>s^\h+<ret><a-@>"
+    }}
+    set-option global aligntab false
+    remove-hooks global expandtab
+    remove-hooks global noexpandtab
+}
--- a/rc/core/expandtab.kak
+++ b/rc/core/expandtab.kak
@@ -1,33 +1,68 @@
+declare-option -docstring "amount of spaces that should be treated as single tab character when deleting spaces" \
+int softtabstop 0
+
+declare-option -docstring "displays current tab handling mode" \
+str smarttab_mode ''
+
+declare-option -hidden int oldindentwidth %opt{indentwidth}
+define-command -hidden smarttab-set %{ evaluate-commands %sh{
+    if [ $kak_opt_indentwidth -eq 0 ]; then
+        echo "set-option window indentwidth $kak_opt_oldindentwidth"
+    else
+        echo "set-option window oldindentwidth $kak_opt_indentwidth"
+    fi
+}}
+
 define-command -docstring "noexpandtab: use tab character to indent and align" \
 noexpandtab %{
-    remove-hooks window tabmode
-    hook -group noexpandtab window NormalKey <gt> %{ try %{
-        execute-keys -draft "<a-x>s^\h+<ret><a-@>"
-    }}
+    set-option window smarttab_mode 'noexpandtab'
+    remove-hooks window smarttab-mode
+    smarttab-set
+    set-option window indentwidth 0
     set-option window aligntab true
+    hook -group smarttab-mode window InsertDelete ' ' %{ try %sh{
+        if [ $kak_opt_softtabstop -gt 1 ]; then
+            echo 'execute-keys -draft <a-h><a-k> "^\h+.\z" <ret>I<space><esc><lt>'
+        fi
+    } catch %{
+        try %{ execute-keys -draft h %opt{softtabstop}<s-h> 2<s-l> s "\h+\z" <ret>d }
+    }}
 }
 
 define-command -docstring "expandtab: use space character to indent and align" \
 expandtab %{
-    remove-hooks window tabmode
-    hook -group tabmode window InsertChar '\t' %{ execute-keys -draft h@ }
-    hook -group tabmode window InsertDelete ' ' %{ try %{
-        execute-keys -draft <a-h><a-k> "^\h+.\z" <ret>I<space><esc><lt>
-    }}
+    set-option window smarttab_mode 'expandtab'
+    remove-hooks window smarttab-mode
+    smarttab-set
     set-option window aligntab false
+    hook -group smarttab-mode window InsertChar '\t' %{ execute-keys -draft h@ }
+    hook -group smarttab-mode window InsertDelete ' ' %{ try %sh{
+        if [ $kak_opt_softtabstop -gt 1 ]; then
+            echo 'execute-keys -draft <a-h><a-k> "^\h+.\z" <ret>I<space><esc><lt>'
+        fi
+    } catch %{
+        try %{ execute-keys -draft h %opt{softtabstop}<s-h> 2<s-l> s "\h+\z" <ret>d }
+    }}
 }
 
 define-command -docstring "smarttab: use tab character for indentation and space character for alignment" \
 smarttab %{
-    remove-hooks window tabmode
-    hook -group tabmode window InsertKey <tab> %{ try %{
+    set-option window smarttab_mode 'smarttab'
+    remove-hooks window smarttab-mode
+    smarttab-set
+    set-option window indentwidth 0
+    set-option window aligntab false
+    hook -group smarttab-mode window InsertChar '\t' %{ try %{
         execute-keys -draft <a-h><a-k> "^\h*.\z" <ret>
     } catch %{
         execute-keys -draft h@
     }}
-    hook -group tabmode window NormalKey <gt> %{ try %{
-        execute-keys -draft "<a-x>s^\h+<ret><a-@>"
+    hook -group smarttab-mode window InsertDelete ' ' %{ try %sh{
+        if [ $kak_opt_softtabstop -gt 1 ]; then
+            echo 'execute-keys -draft <a-h><a-k> "^\h+.\z" <ret>I<space><esc><lt>'
+        fi
+    } catch %{
+        try %{ execute-keys -draft h %opt{softtabstop}<s-h> 2<s-l> s "\h+\z" <ret>d }
     }}
-    set-option window aligntab false
 }
 
--- a/rc/core/expandtab.kak
+++ b/rc/core/expandtab.kak
@@ -1,38 +1,33 @@
 define-command -docstring "noexpandtab: use tab character to indent and align" \
 noexpandtab %{
-    remove-hooks global noexpandtab
-    hook -group noexpandtab global NormalKey <gt> %{ try %{
+    remove-hooks window tabmode
+    hook -group noexpandtab window NormalKey <gt> %{ try %{
         execute-keys -draft "<a-x>s^\h+<ret><a-@>"
     }}
-    set-option global aligntab true
-    remove-hooks global expandtab
-    remove-hooks global smarttab
+    set-option window aligntab true
 }
 
 define-command -docstring "expandtab: use space character to indent and align" \
 expandtab %{
-    remove-hooks global expandtab
-    hook -group expandtab global InsertChar '\t' %{ execute-keys -draft h@ }
-    hook -group expandtab global InsertKey <backspace> %{ try %{
+    remove-hooks window tabmode
+    hook -group tabmode window InsertChar '\t' %{ execute-keys -draft h@ }
+    hook -group tabmode window InsertDelete ' ' %{ try %{
         execute-keys -draft <a-h><a-k> "^\h+.\z" <ret>I<space><esc><lt>
     }}
-    set-option global aligntab false
-    remove-hooks global noexpandtab
-    remove-hooks global smarttab
+    set-option window aligntab false
 }
 
 define-command -docstring "smarttab: use tab character for indentation and space character for alignment" \
 smarttab %{
-    remove-hooks global smarttab
-    hook -group smarttab global InsertKey <tab> %{ try %{
+    remove-hooks window tabmode
+    hook -group tabmode window InsertKey <tab> %{ try %{
         execute-keys -draft <a-h><a-k> "^\h*.\z" <ret>
     } catch %{
         execute-keys -draft h@
     }}
-    hook -group smarttab global NormalKey <gt> %{ try %{
+    hook -group tabmode window NormalKey <gt> %{ try %{
         execute-keys -draft "<a-x>s^\h+<ret><a-@>"
     }}
-    set-option global aligntab false
-    remove-hooks global expandtab
-    remove-hooks global noexpandtab
+    set-option window aligntab false
 }
+


Screwtapello:
Some thoughts:

- since all these modes are mutually exclusive, it doesn't make much sense for them to have their own hook groups. Instead, each command should start with `remove-hooks global tabmode` and then add all its hooks in the `tabmode` group.
- I understand that these commands are named after the Vim options they model, but for Kakoune we can probably come up with more consistent, understandable names. How about `:tabmode-spaces`, `:tabmode-tabs` and `:tabmode-smart`?
- These commands always configure themselves globally, while it's quite likely that people will want different behaviour in different files (I always want space-indents in Markdown and modern Python, but I have a bunch of older Python projects that use tab-indents, and of course Makefiles require tabs).
- What about Vim's `softtabstop` option?

mawww:
I am not sure about this one, this provides some convenience wrapping to make it easier to set-up a certain style of indent behaviour. While I agree it can be a pain point when starting with Kakoune (we just have to look at the regularity at which new users ask about having the <tab> key insert spaces), I wonder if bundling it as a core feature is the way to go.

Usually, I prefer not to have those convenience features bundled for a few reasons:

* I think there is some value in exposing users early to Kakoune concepts such as hooks, exec/eval... Having those introduced for simple (and relatively common) use cases showcases how Kakoune tries to provide the building blocks. This is why expanding tabulations is not bundled, and I usually suggest `hook global InsertChar \t %{ exec -draft h@ }`. I prefer it to suggesting `my-magic-command` because it exposes the user to those concepts (and is still a one liner to put in the kakrc).

* I fear that hiding complexity leads to more complexity (because complexity is now less taxing). Exposed complexity has to be kept in check; I think that has worked relatively well for Kakoune.

* This provides 3 styles of indenting, mapping to vim's behaviour, but I am not convinced this set of styles is exhaustive, and I see no reason to refuse adding other users preferred style of indentation once we bundle those.

* In general, it bothers the minimalist in me.

That said, I understand that there are some users who prefer having the complexity hidden and a nice out of the box experience, for those, I think I'd be more comfortable with a Kakoune distribution that would bundle convenience scripts (like a plugin manager, easier settings...), kind of like what space$editor do.

(This is a kind of thinking out loud post, I am happy to get feedback on those points).

rc/core/expandtab.kak
```diff
@@ -0,0 +1,38 @@
+define-command -docstring "noexpandtab: use tab character to indent and align" \
+noexpandtab %{
+    remove-hooks global noexpandtab
```
mawww: As suggested by @Screwtapello, a single hook group would make things smaller.

rc/core/expandtab.kak
```diff
@@ -0,0 +1,38 @@
...
+    hook -group noexpandtab global NormalKey <gt> %{ try %{
+        execute-keys -draft "<a-x>s^\h+<ret><a-@>"
```
mawww: Why do we need this hook ? Is setting `indentwidth` to 0 not working ? (Maybe its not that well documented, I see that the info box does not mention `indentwidth` of 0 means using a tab character for indentation).

andreyorst: > Is setting indentwidth to 0 not working ?

I didn't knew that. In this case this certainly can be used instead of hook.

> Why do we need this hook ? 

However how to know what `indentwidth` to set when we turn it back by calling another command, like `expandtab`? I surely can store the old value in a hidden option, however it'll be more complex compared to this hook.

andreyorst:
> since all these modes are mutually exclusive, it doesn't make much sense for them to have their own hook groups. Instead, each command should start with remove-hooks global tabmode and then add all its hooks in the tabmode group.

This is nice addition. I'll make it this way.

> I understand that these commands are named after the Vim options they model, but for Kakoune we can probably come up with more consistent, understandable names. How about :tabmode-spaces, :tabmode-tabs and :tabmode-smart?

Well, I think that most of features are easier to find if they named like features that are already known by community. Many Mouilleurs are coming from Vim world, and I suppose that this will be easier to figure out. For my first couple of days I was trying different commands from Vim to find out that they are in Kakoune or not. 

I'm not against the rename, just pointing my thoughts.

> These commands always configure themselves globally, while it's quite likely that people will want different behaviour in different files (I always want space-indents in Markdown and modern Python, but I have a bunch of older Python projects that use tab-indents, and of course Makefiles require tabs).

Oops. I made them window specific at some time, but swithced back because of something. I mostly edit files that use same idiom, so I've missed this fact. `window` makes more sense.

rc/core/expandtab.kak
andreyorst: Yeah. I'll change it.

andreyorst:
> This is a kind of thinking out loud post, I am happy to get feedback on those points

Here's my thoughts on that:

> I think there is some value in exposing users early to Kakoune concepts such as hooks, exec/eval... Having those introduced for simple (and relatively common) use cases showcases how Kakoune tries to provide the building blocks. This is why expanding tabulations is not bundled, and I usually suggest hook global InsertChar \t %{ exec -draft h@ }. I prefer it to suggesting my-magic-command because it exposes the user to those concepts (and is still a one liner to put in the kakrc).

If you don't mind I'll disagree with this point. It is only partly true. The true part is that this is great way to teach people to extend the editor. This fails only when user needs an editor, and not a constructor.

That's exactly why 4 out of 5 people who I've suggested to try Kakoune dropped it. Instead of working they had to make basic things for themselves (this one included). For some people this idiom is fine, since it allows you to build editor of your own, however it doesn't mean that editor shouldn't ship basic functionality. Emacs if full of features, yet Emacs users are changing Emacs to work how they need, since it is customable, but it not forces you to do the customization by not providing basic features.

Why people choose Vim over different text editors? Because of keybindings? Well, most of beginners only know that there's a different style of operating the editor, but is it the most important part? I doubt, because in that case both Vim and Kakoune would finish the developing process on the keyboard shortcuts. Vim has so much more than that. It has many things that allow operate the main thing in the editor, and it is text itself. I'm not talking about <kbd>d</kbd><kbd>w</kbd> here. I'm talking about everything beyond default movements. It is both helper tools that change text for you (like justify macro that comes with Vim), tools that allow you to concentrate on editing, and do all boring stuff for you automatically. Kakoune has lots of such features. Less than Vim, but it's due the age of it. It will extend, and expand, as users will come. However will they come if basic things that exist in other editors are non existing in Kakoune? Well, that will bring enthusiasts like me, who thinks of it as a some kind of a challenge, but many other users will be lost. In Vim the adaptation  period is accepting the keybindings. In Emacs, it is accepting the functions for everything (Emacs allows you edit and navigate like in any basic editor, most of the time). In Kakoune it is both keybindings, and functions, many of which user needs to define himself.

(Well this is only my point of view, I'm not trying to insult anyone or bring a holy-war. I'm not forcing anyone to agree with me.)

> I fear that hiding complexity leads to more complexity (because complexity is now less taxing). Exposed complexity has to be kept in check; I think that has worked relatively well for Kakoune.

I doubt that this is complex. Maybe I see the complexity in a different way.

> This provides 3 styles of indenting, mapping to vim's behaviour, but I am not convinced this set of styles is exhaustive, and I see no reason to refuse adding other users preferred style of indentation once we bundle those.

This provides three the most basic styles of indenting. I don't know any other. It may be language dependent, like in Lisp spaces are superior, but idk.

> That said, I understand that there are some users who prefer having the complexity hidden and a nice out of the box experience, for those, I think I'd be more comfortable with a Kakoune distribution that would bundle convenience scripts (like a plugin manager, easier settings...), kind of like what space$editor do.

No, please, no SpaceKakoune. This is the absolute ultimate way to break the (small) community into pieces. I understand that such distributions will appear at some point, but I'd not praise them. Distribution means distinguishing form origin and providing a certain way of this distribution only. This means that at some point the project is mature enough to have nice features but it refuses them, and users are forced to make the changes in a sort of a fork, that in the end will go absolutely different path.

Well, maybe I'm wrong and everything is a huge mistake. If so, shame on me.

andreyorst:
> What about Vim's softtabstop option?

Softtab is difficult to implement. I've addressed this in [this comment](https://github.com/mawww/kakoune/issues/1038#issuecomment-422288357), and tried to do, however had no luck. I'll try it again, I have some idea, but it fells complex

lenormf:
How about making this a plugin first, and if it's really popular, consider merging it in `extra`?

andreyorst:
Ok.

andreyorst:
I've made a plugin: https://github.com/andreyorst/smarttab.kak
Code review is welcome.

andreyorst:
I've didn't change the plugin code in a while and not planning to, so I've decided to open this PR back, since the code is stable and I had no issues with it so far.

aldanor:
👍 for the plugin, just leaving some feedback. Would it ever make sense to merge it into the master?

endersstocker:
It’s been said that this tab-handling feature isn’t oft-requested, therefore Kakoune users (or potential Kakoune users) don’t need it. The nature of the feature is that (a.) it is ubiquitous, you’d be hard-pressed to find a text editor used by professionals that doesn’t have this feature (b.) it is set-it-and-forget-it, once configured one forgets that is even a bit of configuration. Because of these two facts, after committing some code in version control, that a linter in CI caught the mix of space and tab characters in my code. Only _then_ did it dawn on me that I needed to configure tab-handling with `expandtab`. Because of fact _b_, I had incorrectly assumed that my tab-handling was the same as it’s been for nearly two decades. When I went to “flip the switch,” I discovered it didn’t exist in Kakoune.

I’m dipping my toes into Kakoune. As such, I’d like to set up the “text editor basics” to make it useful for my daily work. The thing that makes Kakoune the next step in the evolution from _ed_ to _vi_ to _Vim_ to _Kakoune_ is the object-verb reversal and concomitant merging of visual and normal modes, not the minimal core. Kakoune has to operate as professional tool in order to successfully revolutionize the interaction pattern (i.e., object-verb vi). Dropping tab-handling and treating the tab key as the tab character is equally revolutionary, but is that the revolution Kakoune is going for? Because that would certainly relegate it to a toy.

When I endeavoured to see what it would take to enable `expandtab` in Kakoune, I was met with a cacophony. I had to read through the entirety of https://github.com/mawww/kakoune/issues/1038 and this one to be sure that https://github.com/andreyorst/smarttab.kak was the right choice. Given that nearly 60% of professionals reported following the `expandtab` or `smarttab` approach in the [Stack Overflow Survey](https://insights.stackoverflow.com/survey/2017), over half of Kakoune’s eventual users need this plugin. And that’s only current users. Who’s to say that a `noexpandtab` user won’t work on an `expandtab` or `smarttab` codebase in the future?  If over half the user-base needs a plugin, and every user _might_ need that plugin at some point, doesn’t that reveal a fundamental shortcoming of the tool? Can EditorConfig support be claimed to be complete without full support for `expandtab`?

Combining the fact that so many developers (admittedly not the only users of a text editor, though a significant portion) rely on `expandtab` with facts _a_ and _b_ above, I can only conclude that a significant number of those who try Kakoune don’t realize the basic text editor config they need will have to come from a plugin. If those users are in the “vanilla all the things!” camp, they will move on from Kakoune as soon as they realize this. For every individual person that spends the time and energy to request functionality, a multitude moves on to another tool.

Tab-handling isn’t an oft-requested feature because its absence isn’t immediately noticed and when it is, it’s a deal-breaker that many don’t have the time to address in lengthy debates.

endersstocker:
It’s also worth nothing that Kakoune’s own [C++ Coding Style](https://github.com/mawww/kakoune/blob/master/doc/coding-style.asciidoc) prescribes:
> 4 spaces for indentation, no tabs

Kakoune contributors would benefit directly from `expandtab` functionality.

mawww:
@endersstocker Hello,

I'll try to explain my reasoning on why this is not bundled a bit further:

There are different possible behaviours when tab is pressed, insert a tab characters, expand it, smart tabs... In any case Kakoune needs to default to one of those and the user has to override the default if it does not match their desired behaviour. When I had to decide on this default behaviour, I decided to go with what seemed to be the simplest behaviour as default: tab is not special, it generates a tab character, and we just insert it.

Once this decision is done, any user that wants another behaviour has to add some configuration somewhere to override it, the most common alternate behaviour requested is "tab insert the number of spaces a tab character would visually occupy". Kakoune already provides all the necessary tools to do that, namely the `@` command that replaces tabs with spaces and a simple hook can do that automatically: `hook global InsertChar \t %{ exec -draft h@ }`. I decided that adding a built-in option to control that was hence unecessary.

I would agree that `set global expandtab true` is a bit more self explanatory that `hook global InsertChar \t %{ exec -draft h@ }`, but I decided it was a reasonable trade-off.

Now smarttab.kak came into existence thanks for @andreyorst work, and the question became whether to bundle it or not. I decided against bundling it as I thought users would need to look up how to enable smarttab anyway, being directed to a plugin did not seem that bad.

To be honest another reason is a matter of personal preference, I think indenting / deindenting should be done with the indent/deindent commands (`<` and `>`), it seems not only unnecessary but also a mode responsibility violation to put so much magic under the **insert mode** `<tab>` key,. I will conceed this is a pretty weak argument though.

lenormf:
While @mawww's reply addressed the rational parts of the complaint, I will address the irrational ones.

>(a.) it is ubiquitous, you’d be hard-pressed to find a text editor used by professionals that doesn’t have this feature

If you want a feature to be provided out of the box, you can't imply that Kakoune has to catch up with "editors used by professionals" to stay relevant. The point of this project is to create a good tool, decisions are made according to what makes sense, not what the most popular editors of the moment already implement.

"Used by professionals" is a vague label that could refer to any arbitrary group of tools of your choosing, so the comparison is also irrelevant. Other uses of this hypocritical terminology litter your reply, I won't address them.

Finally, you can see that the overwhelming majority of users who took the survey you linked to report using IDEs, which Kakoune is not, so discussing the features that Kakoune _should_ provide cannot be framed in this type of comparison.

>(b.) it is set-it-and-forget-it, once configured one forgets that is even a bit of configuration

The original thread mentions a single line of code to add to your user configuration in order to enable the behaviour that you want, not sure what setting an option would do than setting a hook wouldn't.

>nearly 60% of professionals reported following the expandtab or smarttab approach in the Stack Overflow Survey, over half of Kakoune’s eventual users need this plugin

Assuming you're referring to the following bit of data:



I don't see how you're getting a sense of what developers prefer in terms of behaviour of their editor whenever tabs are being inserted, nor where the number of 60% comes from.

Even if your assumptions were accurate, there's no way to correlate what the population who took the survey with what Kakoune users want (we don't even know how many Kakoune users we have at the moment).

By the way, you linked to data from 2017, this particular question was removed from the subsequent yearly surveys, which hints that maybe tabs/spaces indentation isn't that big of a topic to users.

>I had incorrectly assumed that my tab-handling was the same as it’s been for nearly two decades

You assumed Kakoune would be a drop-in replacement for whatever you were using before. But you also refer to the feature as being a "deal-breaker", why haven't you made sure the editor supported it before using it? Note that the feature is fully supported, you have several documented solutions to pick from already.

andreyorst:
>I would agree that `set global expandtab true` is a bit more self explanatory that `hook global InsertChar \t %{ exec -draft h@ }`, but I decided it was a reasonable trade-off.

this is a bit harder than that hook, hence the plugin was created to support `softtabstop` thing.

on a side note, every time I have a nice discussion of text editing with some Emacs user at my work, they often  praise the fact that in Emacs single tab press sets correct indentation based on syntax. While kakoune relies on formatters, some sort of syntax parsing still done for `<ret>` key to provide correct indentation when inserting line breaks. Taking this into account:

>I think indenting / deindenting should be done with the indent/deindent commands (`<` and `>`), it seems not only unnecessary but also a mode responsibility violation to **put so much magic under the insert mode** `<tab>` key

What I wanna to say is that it is fine for `<ret>` to be *magical* key in insert mode but not fine for `<tab>`. Why?

mawww:
> > I would agree that `set global expandtab true` is a bit more self explanatory that `hook global InsertChar \t %{ exec -draft h@ }`, but I decided it was a reasonable trade-off.
> 
> this is a bit harder than that hook, hence the plugin was created to support `softtabstop` thing.

Yep, I was trying to retrace how we got to the current state, soft tabs were not even a concern at that point.

> What I wanna to say is that it is fine for `<ret>` to be _magical_ key in insert mode but not fine for `<tab>`. Why?

As I said, this is a pretty weak argument anyway. I think I am more comfortable with `<ret>` indentation hooks because there is no reasonable alternative provided by Kakoune.

andreyorst:
I'm not pushing on you, I suppose plugin is just fine for most people, but I'm also interested in which direction Kakoune will go in terms of indentation handling as tools that provide syntax tree (parinfer, treesitter) to parse evolve and becoming more popular and widespread. Because I see some weak points in current semantics of indentation handling that may conflict with future implementations of this feature.

mawww:
@andreyorst No worries, I am happy to get decisions like that challenged.

Regarding the current state of auto-indentation, it is brittle, arcane and enforce an arbitrary coding style. I'd be tempted to make them opt-in. On the other hand, I am quite happy that Kakoune built-in features make it possible to implement those, but if we are to keep things as-is, I'd like to see much more test cases.

PR #2596, opened 2018-11-25
src: Build options to compile with sanitizers

lenormf:
This commit adds options to the Makefile that make the compiler
enable sanitizers.

The flags and option names where based on the official upstream
documentation (for `clang` v8 as of writing this):

https://clang.llvm.org/docs/index.html

Note that the `CXX` variable should be forced to `clang++` in theory,
but in practice `g++` seems to support some sanitizers (ASAN for
example), so it's the user's responsibility to pick the right compiler.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -11,6 +11,17 @@ endif
 ifeq ($(debug),yes)
     CPPFLAGS += -DKAK_DEBUG
     suffix := .debug
+
+    ifeq ($(debug_asan),yes)
+        CXXFLAGS += -fsanitize=address -fno-omit-frame-pointer
+    else ifeq ($(debug_msan),yes)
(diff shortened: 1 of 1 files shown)


Screwtapello:
Would it be a better UI to have for example `make debug=asan` rather than `make debug=yes debug_asan=yes`? That way we don't have to care about conflicting configurations like `make debug=no debug_asan=yes`.

lenormf:
I don't think we need to care about conflicts, it seems obvious enough (to me) that enabling any `debug_*` option will be a no-op if `debug` is set to `no`.

I'd rather keep the `debug` and other debugging options separated (it has to benefit of not making package maintainers fix their scripts), but since I made it so sanitizers are mutually exclusive (which isn't a technical limitation of some sanitizers), I guess I could use a `debug_sanitizer=.+` option?

PR #2597, opened 2018-11-25
[WIP] Implement a scrollbar highlighter

occivink:
This highlighter adds a scrollbar that visualizes the position of the view within the buffer. I find it useful because we currently don't have a way to get a sense of how big a buffer is, nor where we are relatively.

It also allows adding 'indicator' sub-highlighters that can be used to modify the color of portions of the scrollbar. This is currently done via `line-specs` options because the main use-case I can think of is to get a rough idea of where you have {errors,warnings,diff,breakpoints,...} in a buffer. But it would be easy to change it to an `int-list` that simply specify the lines.

Multiple 'indicators' can be added, in which case they will compete for setting the face of the scrollbar cell. The first indicator added has precedence.

[Screenshot](https://0x0.st/s01u.jpg) (without indicators) for people who don't want to build this.
This can be tested like so:
```
addhl global/ scrollbar
addhl global/scrollbar/ scrollbar-indicator red,default git_diff_flags
git show-diff
```
(on a buffer that has diff, in order to see the colors changing).

### Issues

Broken when the view is at the end of the buffer. Since there are no DisplayLine that cover the empty part, the scrollbar does not extend over the entire view. I tried pushing new DisplayLine but this messes with other highlighters. 

The padding is currently done in a hacky way (at least I find it): the highlighter shouldn't have to consider whether an atom has a buffer range, especially since we simply want to add one cell. But I could not find a way to consistently display at the right edge with just the value of `context.setup.window_range`.

### DIFF
--- a/src/face_registry.cc
+++ b/src/face_registry.cc
@@ -157,6 +157,7 @@ FaceRegistry::FaceRegistry()
         { "MatchingChar", {Face{ Color::Default, Color::Default, Attribute::Bold }} },
         { "BufferPadding", {Face{ Color::Blue, Color::Default }} },
         { "Whitespace", {Face{ Color::Default, Color::Default, Attribute::FinalFg }} },
+        { "Scrollbar", {Face{ Color::White, Color::Default }} },
       }
 {}
 
--- a/src/highlighters.cc
+++ b/src/highlighters.cc
@@ -23,6 +23,7 @@
 
 #include <locale>
 #include <cstdio>
+#include <cmath>
 
 namespace Kakoune
 {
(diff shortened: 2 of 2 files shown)


andreyorst:
> This highlighter adds a scrollbar that visualizes the position of the view within the buffer. I find it useful because we currently don't have a way to get a sense of how big a buffer is, nor where we are relatively.

Small thought. While this is kind of true, and scrollbar is cool feature, I've never though that positioning in a file means anything really. I've liked that in Emacs with powerline you can have like a mini-map in the statusline, that shows position in the file, but never found it really useful. Unless I'm comparing files with diff, and I need to see if there's more to see below. That's why I've made a `position` module in my [powerline.kak](https://github.com/andreyorst/powerline.kak) plugin (the last one):
image

By the way sublime text style minimap is way more informative compared to scrollbar, but I understand that we can't do such thing in TUI. So percent position in a file usually enough for me in most cases

> It also allows adding 'indicator' sub-highlighters that can be used to modify the color of portions of the scrollbar. This is currently done via line-specs options because the main use-case I can think of is to get a rough idea of where you have {errors,warnings,diff,breakpoints,...} in a buffer. But it would be easy to change it to an int-list that simply specify the lines.

That's actually a great feature in such editors as VS Code, and IIRC Intellij Idea. If we could utilize it with language server, or builtin linters this actually will make whole life easier, but I see a small problem here.

Since such scrollbar is limited in "resolution" to terminal height, and one char width, we can't display such marks with precision. For really long files this would mean that seeing error mark on a scrollbar isn't correspond with actual positon of the mark, since limited resolution of scrollbar would include hundreds of lines in one vertical character. Tha'ts not a problem for GUI, since we can create scrollbar of any pixel height, and display marks on it with pixel precision, but that's not the case for TUI. So I feel like this feature is questionable, and may lead to inconsistency of movement within the file for newcomers.

occivink:
Thanks for the feedback.

>If we could utilize it with language server, or builtin linters this actually will make whole life easier

If you checkout the branch, you can try it on lint flags or lsp errors, it reuses the same options.
```
addhl global/ scrollbar
addhl global/scrollbar/ scrollbar-indicator red,default lsp_error_lines
addhl global/scrollbar/ scrollbar-indicator yellow,default lint_flags
```
You're right that the lack of resolution is a concern, I think I need to test the PR on "real" usecases. It might be possible to double the resolution by using half-blocks with different foreground/background, but that would hard-code the characters, which I don't think is worth it.

FlyingWombat:
Do you know if it would be feasible to make the scroll bar something like [Oni](https://github.com/onivim/oni) did? That is to have a small line within the view block to indicate the cursor position in the view.  
To do this, maybe we could make use of unicode braile characters to make a "high resolution" scroll bar, like how [gtop](https://github.com/aksakalli/gtop) does its stuff.

This, of course, wouldn't be necessary. But it would be cool.

occivink:
It would be possible but we would have to hardcode the characters used for the scrollbar. Currently they are configurable and by default use the same characters as the completion scrollbar, which has this nice dithered pattern for the "background". 
If we were to hardcode some characters, I think I would rather use the half-block ones so that we would have twice the scrollbar resolution.
I'm also not too sure about the feature itself; since the cursor is already in the view, I think it makes more sense to look at the view itself to find the cursors rather than at the scrollbar

andreyorst:
> Currently they are configurable and by default use the same characters as the completion scrollbar, which has this nice dithered pattern for the "background".

wait, so I can configure those scrollbar symbols?

occivink:
Not the ones that already exist (completions, prompt suggestions), only the one from this PR's highlighter.

EPonyA:
> By the way sublime text style minimap is way more informative compared to scrollbar, but I understand that we can't do such thing in TUI. So percent position in a file usually enough for me in most cases

It's also a lot more expensive in terms of screen real estate. It effectively prohibits a two-pane vertical spilt on a portrait oriented window, and the additional information a minimap provides is worth very little by comparison imho. A scroll bar does not have this drawback.

Screwtapello:
For anyone wanting a scrollbar that works with official Kakoune builds, there's a scrollbar plugin at https://github.com/sawdust-and-diamonds/scrollbar.kak

PR #2601, opened 2018-11-27
src: Remember the selection count when jumping across search matches

### LINKED ISSUE #807: Search buffer not matching selection when using a count
First let me see if I understand this ok. Correct me if I am wrong. I think of count + selection as a lookahed/lookbehind mechanism.

Examples:
- `1sfoo(bar)`: selects bar only if preceded by foo
- `1s(foo)bar`: selects foo only if followed by bar

The problem is when using `n` after these selections. It highlights all of `foobar` instead of only the selected bits.

Also related, `^r1` does paste the selected portion but only when in search mode, so after `/` but not in the buffer in insert mode. Trying in the buffer in insert mode inserts nothing + after it doesn't insert anymore in search mode too.

Can you please give us a few useful examples of count selection use + `^r` + count use.

Thanks

lenormf:
Fixes #807


src/selectors.cc
```diff
@@ -906,22 +906,22 @@ Selection find_next_match(const Context& context, const Selection& sel, const Re
...
       : find_prev(buffer, pos, matches, regex, wrapped);
 
-    if (not found or matches[0].first == buffer.end())
+    if (not found or matches[capture_count].first == buffer.end())
         throw runtime_error(format("no matches found: '{}'", regex.str()));
 
     CaptureList captures;
-    for (const auto& match : matches)
-        captures.push_back(buffer.string(match.first.coord(), match.second.coord()));
+    captures.push_back(buffer.string(matches[capture_count].first.coord(),
```
mawww: I think this will break registers 1-9, as we do not store the captures anymore.

lenormf: Yea I wasn't sure this change was correct, but it seemed that we only cared about `matches[0]` when doing simple a search. I'll revert it.

lenormf:
A search will reset the capture count to 0, I don't think there's anything unexpected in your example, with regards to how I implemented the feature. Do you mean you want the capture count to be stored in a client-specific context, as opposed to the shared one?

lenormf:
In that case that's the behaviour I wanted, I want the count passed to `s` to be remembered only in the client that used it. The others will have 0 by default.

But you'd like the count to be considered a metadata of the select pattern, and be shared among all clients? Have any idea where that count could be stored then? Another register?

PR #2603, opened 2018-11-27
Setup language cleanup hooks only during initialization

occivink:
With all plugins loaded, this lowers the numbers of hooks executed when doing `set buffer filetype kak` from 130 to 3. 
I tried to make the setup consistent in all languages.

I manually tested all the languages in core/, some in base/ (javascript in particular, which has an unusual setup), and only a few in extra/.

### DIFF
--- a/rc/base/d.kak
+++ b/rc/base/d.kak
@@ -116,7 +116,10 @@ define-command -hidden d-indent-on-closing-curly-brace %[
 # Initialization
 # ‾‾‾‾‾‾‾‾‾‾‾‾‾‾
 
-hook -group d-highlight global WinSetOption filetype=d %{ add-highlighter window/d ref d }
+hook -group d-highlight global WinSetOption filetype=d %{
+    add-highlighter window/d ref d
+    hook -once -always window WinSetOption filetype=(?!d).* %{ remove-highlighter window/d }
+}
 
 hook global WinSetOption filetype=d %{
     # cleanup trailing whitespaces when exiting insert mode
@@ -124,10 +127,6 @@ hook global WinSetOption filetype=d %{
     hook window InsertChar \n -group d-indent d-indent-on-new-line
     hook window InsertChar \{ -group d-indent d-indent-on-opening-curly-brace
     hook window InsertChar \} -group d-indent d-indent-on-closing-curly-brace
-}
 
-hook -group d-highlight global WinSetOption filetype=(?!d).* %{ remove-highlighter window/d }
-
-hook global WinSetOption filetype=(?!d).* %{
-    remove-hooks window d-.+
+    hook -once -always window WinSetOptionfiletype=(?!d).* %{ remove-hooks window d-.+ }
 }
--- a/rc/base/d.kak
+++ b/rc/base/d.kak
@@ -118,7 +118,7 @@ define-command -hidden d-indent-on-closing-curly-brace %[
 
 hook -group d-highlight global WinSetOption filetype=d %{
     add-highlighter window/d ref d
-    hook -once -always window WinSetOption filetype=(?!d).* %{ remove-highlighter window/d }
+    hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/d }
 }
 
 hook global WinSetOption filetype=d %{
@@ -128,5 +128,5 @@ hook global WinSetOption filetype=d %{
     hook window InsertChar \{ -group d-indent d-indent-on-opening-curly-brace
     hook window InsertChar \} -group d-indent d-indent-on-closing-curly-brace
 
-    hook -once -always window WinSetOption filetype=(?!d).* %{ remove-hooks window d-.+ }
+    hook -once -always window WinSetOption filetype=.* %{ remove-hooks window d-.+ }
 }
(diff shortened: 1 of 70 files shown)


rc/core/c-family.kak
```diff
@@ -318,34 +318,37 @@ evaluate-commands %sh{
...
+    hook -once -always window WinSetOption filetype=(?!c)(?!cpp)(?!objc).* %{
+        unalias window alt c-family-alternative-file
+    }
 ]
 
-hook -group c-highlight global WinSetOption filetype=c %[ add-highlighter window/c ref c ]
-hook -group c-highlight global WinSetOption filetype=(?!c).* %[ remove-highlighter window/c ]
+hook -group c-highlight global WinSetOption filetype=c %{
+    add-highlighter window/c ref c
+    hook -once -always window WinSetOption filetype=(?!c).* %{ remove-highlighter window/c }
```
occivink: Using `(?!c).*` is less than ideal, because any filetype that begins with c will not trigger this hook

rc/base/d.kak
```diff
@@ -116,18 +116,17 @@ define-command -hidden d-indent-on-closing-curly-brace %[
 # Initialization
 # ‾‾‾‾‾‾‾‾‾‾‾‾‾‾
 
-hook -group d-highlight global WinSetOption filetype=d %{ add-highlighter window/d ref d }
+hook -group d-highlight global WinSetOption filetype=d %{
+    add-highlighter window/d ref d
+    hook -once -always window WinSetOption filetype=(?!d).* %{ remove-highlighter window/d }
```
occivink: Same remark as with c, this hook may not be triggered depending on the new filetype

rc/core/c-family.kak
```diff
@@ -3,7 +3,7 @@ hook global BufCreate .*\.(cc|cpp|cxx|C|hh|hpp|hxx|H)$ %{
 }
 
 hook global BufSetOption filetype=c\+\+ %{
-    set-option buffer filetype cpp
+    hook -once buffer NormalIdle '' "set-option buffer filetype cpp"
```
occivink: I changed this hook because the previous was prone to weird/racy behavior, with the filetype=cpp hook getting triggered twice.

rc/extra/just.kak
```diff
@@ -28,25 +28,21 @@ add-highlighter shared/justfile/comment  region '#' '$'  fill comment
...
 add-highlighter shared/justfile/content/ regex '^([\w-]+)\s=' 1:value
 
-hook global WinSetOption filetype=justfile %{
-    hook window InsertChar \n -group justfile-indent just-indent-on-new-line
+hook -group justfile-highlight global WinSetOption filetype=justfile %{
     add-highlighter window/justfile ref justfile
-
-    # variable substititution highlighting
-    add-highlighter shared/justfile/shell regex '(\{{2})([\w-]+)(\}{2})' 1:operator 2:variable 3:operator
-
```
occivink: @TeddyDD I think this used to be incorrect, can you check that it works as expected?

occivink:
Also maybe the cleanup hooks should be changed from `filetype=(?!language)` to `filetype=.*`? Since the only possible change is from language to not-language.

mawww:
> Also maybe the cleanup hooks should be changed from `filetype=(?!language)` to `filetype=.*`? Since the only possible change is from language to not-language. Incidentally this should fix the issues with languages like C or D

Yep, I think this should work, just pushed a commit to fix a corner case where the hook would trigger on unset-option even if the parent had the same value. Can you update the PR to do that ?

occivink:
There still seems to be an issue. If you try this snippet
```
hook -group kak-highlight global WinSetOption filetype=kak %{
    echo -debug "bbb"
    hook -once -always window WinSetOption filetype=.* %{ echo -debug "aaa" }
}
```
you'll see that both hooks are triggered at once. I guess the second hook is added during hook execution, and triggered right away since it matches?

rc/extra/just.kak
TeddyDD: This is a hack that highlight `{{variable}}` substitutions in shell regions. I'm not sure if this can be done other way. I'll check it after Wednesday, too busy right now :(

mawww:
> you'll see that both hooks are triggered at once. I guess the second hook is added during hook execution, and triggered right away since it matches?

It is also due to adding hooks in the window scope from the global scope, hooks are executed in order global -> buffer -> window, so if a hook in global scope adds new hooks in the window scope that matches the currently executed hook, they will get triggered. I am not sure if thats desired or not, I can see arguments both ways. (If you were adding the hooks to the global scope, I thing they would not get executed).

mawww:
> I find it strange, if you expect a hook to be executed right away it should rather be done directly.

Yeah, I guess that makes sense, when you write the hook, you should be in a good enough position to decide if you want the new hook to run or not, and if you want it, you can just run it manually. I'll see how I can disable that behaviour.

rc/extra/just.kak
TeddyDD: I fixed it here https://github.com/mawww/kakoune/pull/2622 Feel free to pull those changes here or I rebase after this PR will be merged.

occivink: They are the changes I had made as well

TeddyDD: I only saw part of the diff or I'm tired :| sorry :)

mawww:
The behaviour has been changed in Kakoune, can you update this PR ?

occivink:
I hadn't noticed, done.

occivink:
Actually nevermind, that broke c-family's `alt` alias let me fix that up

occivink:
Ok it's fixed. I had to do some minor changes ot c-family to keep it in a correct state throughout filetype changes.

Issue #2554, opened 2018-11-04
Switch away from ncurses to terminfo-based input handling


maximbaz:
This was brought up many times, and is indirectly implied by a few existing issues asking to distinguish certain keyboard mappings regardless of the taken approach. Contrary to all those issues, this ticket is proposing a specific approach: switching away from ncurses to terminfo-based input handling.

Input handling is a key aspect of code editor, I strongly believe this should be done directly using terminfo and not via some third-party wrapper libraries such as ncurses or libtickit. The current limitations of ncurses are simply embarrassing. Terminals evolve, if a new solution / xterm protocol extension gets adopted, we must be able to enable it quickly, not wait for months or years until it's adopted by a library of choice.

If this is implemented, this fixes issues like #2553, #2209, #1527, #1270, #997, #970 and #1012.

maximbaz:
If possible, I suggest to tackle that separately, primarily to simplify the transition.

Screwtapello:
Abandoning ncurses' input-handling and using raw terminfo would make it easier to handle keys like `<c-left>` (which are listed in the terminfo database), but it wouldn't help with most of the issues you list:

* #2553 is about mapping ASCII NUL; glancing at the code I *think* this is just because Kakoune says `> 0` where it means `>= 0`.
* #2209, #1527, #1270, #970 are about the fact that terminals use control-characters for some special keys. That's a fundamental truth of terminals, regardless of whether you use ncurses, terminfo, or write custom code yourself. #1012 (using libtickit) helps if you're using a terminal that implements libtickit's extended-input mode, which (as far as I know) no common terminals actually do.
* For #997, I cannot fathom how `<a-right>` became bound to `<a-f>` (it's not a standard terminal binding at all) and the comments suggest it's a weird Mac thing. It would remain a weird Mac thing regardless of what terminal input library Kakoune uses.

maximbaz:
Well somehow nvim is able to distinguish certain mappings that kakoune cannot (such as `c-j` and `ret`), so I'm sure it's possible. Furthermore, if we use terminfo-based input handling, it is very simple to support modern terminals - for example kitty supports something called [full keyboard mode](https://sw.kovidgoyal.net/kitty/protocol-extensions.html#keyboard-handling) that allows to distinguish each and every key combination, and in order to support that the application needs to simply send out a single escape code that some terminals will respect and others will silently ignore. This is impossible with ncurses or libtickit.

[...]

PR #2609, opened 2018-11-28
Parse xterm-keys for motion directly

jeapostrophe:
Ideally, something better should be done (re #2554) but this is a decent
intermediate step for some useful keys.

Note: NCurses supports parsing these keys when shifted (KEY_SR,
_SLEFT, S_RIGHT, etc), but it does not do the same thing for the other
modifiers.

### DIFF
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -26,6 +26,7 @@ namespace Kakoune
 
 using std::min;
 using std::max;
+using std::function;
 
 struct NCursesWin : WINDOW {};
 
@@ -671,11 +672,12 @@ Optional<Key> NCursesUI::get_next_key()
                   const Codepoint c2 = wgetch(m_window);
                   if ( c2 != ';' )
                   {
-                      ungetch(c2); ungetch(c1); break;
+                      ungetch(c2); ungetch(c1);
+                      break;
                   }
 
                   const Codepoint c3 = wgetch(m_window);
-                  Key (*f)(Key) = NULL;
+                  function<Key(Key)> f  = nullptr;
                   switch (c3)
                   {
                       case '2': f = shift; break;
@@ -686,9 +688,10 @@ Optional<Key> NCursesUI::get_next_key()
                       case '7': f = alt_ctrl; break;
                       case '8': f = shift_alt_ctrl; break;
                   }
-                  if ( f == NULL )
+                  if ( f == nullptr )
                   {
-                      ungetch(c3); ungetch(c2); ungetch(c1); break;
+                      ungetch(c3); ungetch(c2); ungetch(c1);
+                      break;
                   }
 
                   const Codepoint c4 = wgetch(m_window);
@@ -702,10 +705,12 @@ Optional<Key> NCursesUI::get_next_key()
                       case 'F': return f(Key::End);
                   }
 
-                  ungetch(c4); ungetch(c3); ungetch(c2); ungetch(c1); break;
+                  ungetch(c4); ungetch(c3); ungetch(c2); ungetch(c1);
+                  break;
                 }
             default:
-              ungetch(c1); break;
+              ungetch(c1);
+              break;
             }
         }
         wtimeout(m_window, -1);
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -661,12 +661,51 @@ Optional<Key> NCursesUI::get_next_key()
         const int new_c = wgetch(m_window);
         if (new_c == '[') // potential CSI
         {
-            const Codepoint csi_val = wgetch(m_window);
-            switch (csi_val)
+            const Codepoint c1 = wgetch(m_window);
+            switch (c1)
             {
                 case 'I': return {Key::FocusIn};
                 case 'O': return {Key::FocusOut};
-                default: break; // nothing
+                case '1':
+                {
+                  const Codepoint c2 = wgetch(m_window);
+                  if ( c2 != ';' )
+                  {
+                      ungetch(c2); ungetch(c1); break;
+                  }
+
+                  const Codepoint c3 = wgetch(m_window);
+                  Key (*f)(Key) = NULL;
+                  switch (c3)
+                  {
+                      case '2': f = shift; break;
+                      case '3': f = alt; break;
+                      case '4': f = shift_alt; break;
+                      case '5': f = ctrl; break;
+                      case '6': f = shift_ctrl; break;
+                      case '7': f = alt_ctrl; break;
+                      case '8': f = shift_alt_ctrl; break;
+                  }
+                  if ( f == NULL )
+                  {
+                      ungetch(c3); ungetch(c2); ungetch(c1); break;
+                  }
+
+                  const Codepoint c4 = wgetch(m_window);
+                  switch (c4)
+                  {
+                      case 'A': return f(Key::Up);
+                      case 'B': return f(Key::Down);
+                      case 'C': return f(Key::Right);
+                      case 'D': return f(Key::Left);
+                      case 'H': return f(Key::Home);
+                      case 'F': return f(Key::End);
+                  }
+
+                  ungetch(c4); ungetch(c3); ungetch(c2); ungetch(c1); break;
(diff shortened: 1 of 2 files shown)


jeapostrophe:
I am particular interested in Alt+Arrows, because I am building something that matches my muscle memory from other editors.

Screwtapello:
The "proper" way to handle these keys (according to `user_caps(5)`) is:

- start with the list of modifiable keys: `kDC`, `kDN`, `kEND`, `kHOM`, `kLFT`, `kNXT`, `kPRV`, `kRIT`, `kUP`
- and the list of modifier codes: 2 for <kbd>Shift</kbd>, 3 for <kbd>Alt</kbd>, 4 for <kbd>Shift</kbd>+<kbd>Alt</kbd>, 5 for <kbd>Control</kbd>, etc.
- for each possible combination:
     - concatenate the key and the modifier to produce a potential property name like `kLFT4`
     - call `tigetstr(property_name)` to obtain the byte-string that the terminal will send to signify that key and modifier (on my terminal, `kLFT4` is `^[[1;4D`)
     - call `key_defined(byte_string)` to return the dynamically-allocated key-code ncurses uses to signify that key
     - add the resulting key-code to some dynamically-allocated equivalent of the switch statement at line 584 of `ncurses_ui.cc`

Compared to this patch, such an algorithm would work for all modifiable keys, not just the arrow keys, Home, and End. Also, it would work for any terminal in the terminfo database, not just ones that exactly copy xterm's encoding.

On the other hand, such an algorithm would be a much more invasive change, and nobody's actually done the work to implement it, while this patch is already here and ready.

src/ncurses_ui.cc
```diff
@@ -661,12 +661,51 @@ Optional<Key> NCursesUI::get_next_key()
...
+                      case '4': f = shift_alt; break;
+                      case '5': f = ctrl; break;
+                      case '6': f = shift_ctrl; break;
+                      case '7': f = alt_ctrl; break;
+                      case '8': f = shift_alt_ctrl; break;
+                  }
+                  if ( f == NULL )
+                  {
+                      ungetch(c3); ungetch(c2); ungetch(c1); break;
+                  }
```
lenormf: Check if the functor (from above) has a valid target, instead of checking for the `NULL` value (in C++, you should rather use `std::nullptr`, but it won't matter with a functor).

src/ncurses_ui.cc
```diff
@@ -661,12 +661,51 @@ Optional<Key> NCursesUI::get_next_key()
...
             {
                 case 'I': return {Key::FocusIn};
                 case 'O': return {Key::FocusOut};
-                default: break; // nothing
+                case '1':
+                {
+                  const Codepoint c2 = wgetch(m_window);
+                  if ( c2 != ';' )
+                  {
+                      ungetch(c2); ungetch(c1); break;
```
lenormf: I understand why the calls to `ungetch` are on the same line, but I don't think the `break` should be on that line as well. Applies to the other hunks as well.

src/ncurses_ui.cc
jeapostrophe: I've switched to `nullptr`. I didn't embed this check into the switch, because we don't want the `break` on the inner one, we want it on the outer one.

jeapostrophe:
@lenormf I've updated based on your feedback.

@Screwtapello Good to know. An efficient way to implement the dynamic switch would be a trie from key sequences to key encodings and have the reading code explore that trie and have a default case of letting the bytes through. This is essentially just a tokenization/scanner (lex) problem, so there's lots of efficient techniques to do it well if you know the mapping between strings and values you want. The problem is that doing it smart would be out-of-scope for Kakoune and you don't want to depend on other libraries. Just using std::map as a trie is likely to be much less efficient, but doable.

src/ncurses_ui.cc
mawww: nullptr is a built in, so no need for std:: in front. but you are correct, we use nullptr in Kakoune codebase.

src/ncurses_ui.cc
```diff
@@ -686,9 +688,10 @@ Optional<Key> NCursesUI::get_next_key()
                       case '7': f = alt_ctrl; break;
                       case '8': f = shift_alt_ctrl; break;
                   }
-                  if ( f == NULL )
+                  if ( f == nullptr )
```
lenormf: If you want to simplify this further:

```c++
if (f)
```

There's a [boolean operator](https://en.cppreference.com/w/cpp/utility/functional/function/operator_bool) for that. Also no spaces between the parenthesis.

src/ncurses_ui.cc
```diff
@@ -671,11 +672,12 @@ Optional<Key> NCursesUI::get_next_key()
...
                   if ( c2 != ';' )
                   {
-                      ungetch(c2); ungetch(c1); break;
+                      ungetch(c2); ungetch(c1);
+                      break;
                   }
 
                   const Codepoint c3 = wgetch(m_window);
-                  Key (*f)(Key) = NULL;
+                  function<Key(Key)> f  = nullptr;
```
lenormf: To simplify further, you can drop the initialization (` = nullptr`).

jeapostrophe:
Updated again re: @lenormf

jeapostrophe:
I added my copyright waiver.

src/ncurses_ui.cc
```diff
@@ -661,12 +662,55 @@ Optional<Key> NCursesUI::get_next_key()
...
+                  {
+                      case '2': f = shift; break;
+                      case '3': f = alt; break;
+                      case '4': f = shift_alt; break;
+                      case '5': f = ctrl; break;
+                      case '6': f = shift_ctrl; break;
+                      case '7': f = alt_ctrl; break;
+                      case '8': f = shift_alt_ctrl; break;
+                  }
+                  if (f)
```
mawww: I think this is the wrong way around, you want `if (not f)`

jeapostrophe:
Ya, that was stupid. Fixed.

src/ncurses_ui.cc
jeapostrophe: Ya, that was stupid, fixed.

mawww:
Hi, I tried to test that PR locally, and I could not get it to work proprerly, what settings do you put xterm in ? I see that putting it in 8-bit mode generates those keys, except that the CSI sequence is sent as 8-bits as well, which is not parsed correctly. I have local changes to handle all that, but I was wondering what settings you were using.

jeapostrophe:
I don't use xterm. I am using iterm2/tmux and kitty on OS X. kitty has default settings. iterm2 has the "xterm defaults" and tmux has the xterm keys on.

Screwtapello:
If I start a genuine xterm and run:

    tput smxk; cat

...and then press <kbd>Ctrl</kbd>+<kbd>Up</kbd>, I see the output:

    ^[[1;5A

...which would be parsed as `<c-up>` by this code. `smkx` enables application keypad mode, which Kakoune does in `ncurses_ui.cc` line 507. The only configuration option I've messed with is `metaSendsEscape: true', but that shouldn't be relevant to here since I'm not pressing <kbd>Meta</kbd>.

Screwtapello:
The `keypad()` call determines whether ncurses interprets the key encodings defined in terminfo, or passes them through to the application. It seems like mawww's terminal emulator properly generates the special key encodings for modified cursor keys, and his terminfo database describes those encodings, so ncurses recognises them and `wgetch()` returns a dynamically-allocated key number, as described in [my comment above](https://github.com/mawww/kakoune/pull/2609#issuecomment-442624491).

jeapostrophe's situation is that while their terminal sends those encodings, their terminfo database *doesn't* describe them, so ncurses doesn't recognise them and passes them straight through to the application uninterpreted.

mawww:
I merged this and used it as a base to refactor key parsing slightly, I am still unsure what direction we should take from here, at  lease this adds more support for key combinations while preserving what we had, I am tempted to evolve that in the direction Screwtape described, but maybe we can hope for a common, forward compatible key encoding scheme that we can implement (maybe the one Leonerd suggested, or the kitty one...)

PR #2616, opened 2018-12-01
src: Make `gi` jump on the anchor's line

### LINKED ISSUE #1984: `gi` on a multiline selection
I'm not sure if this is a bug along the same lines of #1964, or if this is how <kbd>g</kbd><kbd>i</kbd>
 behaved like since the beginning.

### Steps 

* Create a multiline selection
* Jump to the beginning of the line with <kbd>g</kbd><kbd>i</kbd>

### Outcome 

The jump is made on the line relative to the cursor.

### Expected 

The jump is made on the line relative to the anchor. I think this is unexpected because <kbd>I</kbd> jumps to the anchor.

### Example

With the cursor on `f`, and the anchor on `a`:

```
%(abc
def)
```

<kbd>g</kbd><kbd>i</kbd> selects `d`, not `a`.

lenormf:
Fixes #1984


src/normal.cc
```diff
@@ -226,8 +226,20 @@ void goto_commands(Context& context, NormalParams params)
                 select<mode, select_to_line_begin<true>>(context, {});
                 break;
             case 'i':
-                select<mode, select_to_first_non_blank>(context, {});
```
mawww: select_to_first_non_blank is only used here, so it would make more sense to modify that function.

lenormf: But my modifications don't select the first whitespace character, so I should probably rename that function as well?

Issue #2602, opened 2018-11-27
Modified options are not properly restored by language scripts


occivink:
On any buffer, do
```
set buffer filetype kak
echo -- %opt{extra_word_chars}
# it's set to '-'
unset buffer filetype
echo -- %opt{extra_word_chars}
# still '-'
```

The problem is that the same option can be manipulated by different language scripts. 
If we decided to reset the value of the option in the cleanup hook, then there might be a race between the cleanup hook of language A and the setup hook of language B.

The same apply to the static_words option.

mawww:
My plan to fix this is the to introduce scope overlays (similar in a way to emacs minor modes), language support would just be isolated into a named scope that could be enabled/disabled by adding it to the window/buffer/global scope. This way, enabling a language is just `hook global WinSetOption filetype=python 'enable-scope window python'`

PR #2617, opened 2018-12-02
Expose a 'terminal' command for the various windowing systems

occivink:
It allows plugins to create generic terminal using the user's preferred windowing system
For example, it can be used to run fzf, gdb or simply a shell.

* 'new' commands are refactored to simply use the 'terminal' one
* style and docstrings has been unified <- i can revert this but imo the previous style was confusing and poorly highlighted
* all windowing systems go through "sh -c" for consistency purposes, even if unnecessary <- this is necessary because 'sh -c' is the lowest common denominator, unfortunately.

It this is accepted, I'll try to port iterm.kak as well but I cannot test it, so I'll need help.

### DIFF
--- a/rc/base/new-client.kak
+++ b/rc/base/new-client.kak
@@ -1,0 +1,16 @@
+define-command new -params .. -command-completion -docstring '
+new [<commands>]: create a new kakoune client
+The ''terminal'' alias is being used to determine the user''s preferred terminal emulator
+The optional arguments are passed as commands to the new client' \
+%{
+    try %{
+        terminal %sh{
+            # double-up single-quotes
+            escaped=$(printf %s "$*" | sed -e "s|'|''|g")
+            printf "kak -c '%s' -e '%s'" "$kak_session" "$escaped"
+        }
+    } catch %{
+        fail "The 'terminal' alias must be defined to use this command"
+    }
+}
+
--- a/rc/base/new-client.kak
+++ b/rc/base/new-client.kak
@@ -4,11 +4,7 @@ The ''terminal'' alias is being used to determine the user''s preferred terminal
 The optional arguments are passed as commands to the new client' \
 %{
     try %{
-        terminal %sh{
-            # double-up single-quotes
-            escaped=$(printf %s "$*" | sed -e "s|'|''|g")
-            printf "kak -c '%s' -e '%s'" "$kak_session" "$escaped"
-        }
+        terminal kak -c %val{session} -e "%arg{@}"
     } catch %{
         fail "The 'terminal' alias must be defined to use this command"
     }
--- a/rc/base/screen.kak
+++ b/rc/base/screen.kak
@@ -6,44 +6,70 @@ hook -group GNUscreen global KakBegin .* %sh{
     echo "
         alias global focus screen-focus
         alias global new screen-new-vertical
+        alias global terminal screen-terminal-vertical
     "
 }
 
-define-command screen-new-vertical -params .. -command-completion -docstring "Split the current pane into two, left and right" %{
-     nop %sh{
-        tty="$(ps -o tty ${kak_client_pid} | tail -n 1)"
-        screen -X eval \
-            'split -h' \
-            'focus down' \
-            "screen sh -c 'kak -c \"${kak_session}\" -e \"$*\" ;
-                screen -X remove'" \
-        < "/dev/$tty"
-    }
-}
-
-define-command screen-new-horizontal -params .. -command-completion -docstring "Split the current pane into two, top and bottom" %{
-     nop %sh{
-        tty="$(ps -o tty ${kak_client_pid} | tail -n 1)"
-        screen -X eval \
-            'split -v' \
-            'focus right' \
-            "screen sh -c 'kak -c \"${kak_session}\" -e \"$*\" ;
-                screen -X remove'" \
-        < "/dev/$tty"
-    }
-}
-
-define-command screen-new-window -params .. -command-completion -docstring "Create a new window" %{
+define-command screen-terminal-impl -hidden -params 3.. -command-completion %{
     nop %sh{
         tty="$(ps -o tty ${kak_client_pid} | tail -n 1)"
-        screen -X screen kak -c "${kak_session}" -e "$*" < "/dev/$tty"
+        screen -X eval "$1" "$2"
+        shift 2
+        cmd=$(printf %s "$*")
+        screen -X screen sh -c "${cmd}; screen -X remove" < "/dev/$tty"
     }
 }
 
-define-command -docstring %{screen-focus [<client>]: focus the given client
-If no client is passed then the current one is used} \
-    -params ..1 -client-completion \
-    screen-focus %{ evaluate-commands %sh{
+define-command screen-terminal-vertical -params 1.. -command-completion -docstring '
+screen-terminal-vertical <program> [<arguments>]: create a new terminal as a screen pane
+The current pane is split into two, left and right
+The program passed as argument will be executed in the new terminal' \
+%{
+    screen-terminal-impl 'split -v' 'focus right' %arg{@}
+}
+define-command screen-terminal-horizontal -params 1.. -command-completion -docstring '
+screen-terminal-horizontal <program> [<arguments>]: create a new terminal as a screen pane
+The current pane is split into two, top and bottom
+The program passed as argument will be executed in the new terminal' \
+%{
+    screen-terminal-impl 'split -h' 'focus down' %arg{@}
+}
+define-command screen-terminal-window -params 1.. -command-completion -docstring '
+screen-terminal-window <program> [<arguments>]: create a new terminal as a screen window
+The program passed as argument will be executed in the new terminal' \
+%{
+    nop %sh{
+        tty="$(ps -o tty ${kak_client_pid} | tail -n 1)"
+        screen -X screen sh -c "$*" < "/dev/$tty"
+    }
+}
+
+define-command screen-new-vertical -params .. -command-completion -docstring '
+screen-new-vertical [<commands>]: create a new kakoune client as a screen pane
+The current pane is split into two, left and right
+The optional arguments are passed as commands to the new client' \
+%{
+    screen-terminal-vertical "kak -c '%val{session}' -e '%arg{@}'"
+}
+define-command screen-new-horizontal -params .. -command-completion -docstring '
+screen-new-horizontal [<commands>]: create a new kakoune client as a screen pane
+The current pane is split into two, top and bottom
+The optional arguments are passed as commands to the new client' \
+%{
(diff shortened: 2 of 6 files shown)


mawww:
Hello,

I quite like this change, I think it makes sense to unify on top of this command.

Regarding the docstrings, I'd like to experiment with implementing the proposed change to `%{...}` strings (if it starts with a eol followed by some blanks, trim that indentation from the string content), this will make the docstring cleaner as they can be nicely indented. It has also the added benefit that it will reduce the general size of the various strings we store in Kakoune (as we will be able to drop the base indentation from them) at the price of a bit more work at parsing time.

Regarding the `sh -c` I believe that is already what x11.kak was doing, it was unifying on a syntax accepting a single parameter containing the shell script to run. Some terminals did take multiple parameters that would generate the command, so they went through `sh -c` while other already expected a single shell command, so we did not need to introduce an additional shell. Is there anything else that would force to always use `sh -c` instead of requiring individual `terminal` command implementation to take a single shell command to execute ?

occivink:
Yes you're right that the `terminal` commands should take a single parameter, it makes more sense like that and there is no second-guessing how they're joined. Some additional thoughts:
* Maybe the `new` commands could be renamed to `new-client` to be more explicit?
* The terminal commands could default to opening `$SHELL` if no argument was passed.

Regarding the docstring I'd be happy with this idea, but one problem that I have is that docstrings are semantically dumb static strings for which the `"..."` string highlighting makes sense, whereas `%{...}` strings end up with keywords seemingly randomly highlighted.

lenormf:
The concept of terminal only makes sense in the X11 and terminal support scripts, why introduce it in scripts where it doesn't (e.g. `screen` or `tmux`)? The `*-terminal-*` commands don't spawn a terminal, they only spawn `kak` in a panes?

If you want a single command to be agnostic of the context in which it's being called to spawn either a terminal, or a pane, or whatever, it's better to create an abstraction layer that wraps all those primitive concepts, rather than make them all aware of each other.

All that's needed at the moment to spawn a terminal is `eval "${kak_opt_termcmd}"`, and detecting which multiplexer you're in is done checking the environment, it should be fairly easy to come up with a script that unifies everything.

occivink:
I think the misunderstanding here is that I say "terminal" to really mean "terminal emulator", which makes as much sense in x11, kitty, tmux or iterm. 

The goal is simply to expose a command to open such a terminal emulator in a common way with the specified program running in it. More specifically in gdb.kak, I want to be able to open a new terminal emulator with gdb running in it without having to care about x11, tmux... I'd simply call `terminal  'gdb ...'` and the alias would defer that to the user's preferred windowing system.

As such, the `new` commands are just a special case of `terminal` in the sense that they create a new terminal emulator with kakoune running in it and connected to the current server.

lenormf:
Like I said, this is a case that would be better handled in a separate script that calls the right command according to the current context.

The terminology is indeed wrong, as users will expect a new X11 window if they ever call that `:terminal` command.

occivink:
I don't think putting the terminal creation in a single command is a better approach for the following reasons:

* windowing systems functionality is not centralized anymore. Currently if a user wants to modify some tmux functionality, they only have to look at tmux.kak. Similarly, if they don't use tmux, they should be able to not load tmux.kak and not have anything tmux-related loaded.
* it would duplicate the terminal creation code (commands such as `screen-terminal-impl`) between the new script and the current one. Like I said, creating a new kakoune client is a subset of creating a terminal emulator (at least with ncurses, but a non-terminal UI should be able to cope with that).
* the current approach is easily extensible. A user on a completely different windowing system can create their own new-windowing-system.kak, alias the commands to `terminal` or `new`, and everything that depends on it should *just work*. There is only minimal competition between the different windowing system: they try to set the same alias according to some auto-detection, and the user can easily hard-overwrite that if they don't like the result of the auto-detection.

As for the terminology, I think it's generally understood that `terminal` refers to `terminal emulator`, but I'm not opposed to changing the name to something better.

occivink:
Ok I've tried to port iterm.kak as well. If a kind soul on macos could test it for me I'd appreciate. In particular check that the following works on the different variants: 
```
# a new client of the server is created
iterm-new-vertical
# the new client is named abc
iterm-new-vertical rename-client abc 
# a new pane with bash running in it is created
iterm-terminal-vertical bash
# an empty pane appears and disappears after 5 seconds
iterm-terminal-vertical 'sleep 5'
```

Unfortunately the iterm.kak looks quite fragile regarding to escaping but an actual user will have to look at that. I think some of it could be mitigated by replacing the `env` call with `export`

mawww:
@lenormf It is pretty clear to me that tmux/screen/etc... are terminal emulators themselves, I would agree that many users do not realise that, but that should not prevent Kakoune from using the correct terminology. `x11-terminal` will open a new x11 window, as `x11-new` does right now, and `terminal` will either use x11, tmux, screen, iterm or kitty depending on what it is aliasing to, the same way `new` does right now.

lenormf:
From what I can tell, this PR only allows calling the right command depending on the current context, which is basically a bunch of `if` statements that check for environment variables, and call the right multiplexer command. That's a wrapper.

Modifying the multiplexer support scripts also prevents conflicts handling (`screen` in `tmux` in `iterm` for example), because a wrapping layer could be aware of what's running underneath, as opposed to having all the multiplexers trying to detect each other.

Finally, `terminal` is not a good term, even if technically multiplexers implement a terminal emulator, they are not terminals (which users consider those to be a new window, not something that runs in an already existing one).

occivink:
Ok so if your issue is with the current way of handling windowing and auto-detection with separate scripts (which this PR preservers and only extends to new commands), please take the discussion to its own issue. Note that I don't agree with your proposal for the reasons I've stated above.

mawww:
In my view, aliasing `new` or `terminal` to one specific command does not need to be perfect (we have no way to know what users wants), just to be relatively convenient. Users can implement more complex behaviour in their `kakrc`, setting the `new/terminal` alias to what they want in there.

I do not want to add a central place that knows about all of those, because now this central decision script has a dependency on all the other scripts, or (even worse), needs to detect which of those scripts have been loaded.

I think the `terminal` name is correct, but I am not married to it, howeveer I do not really see any better term, the only alternative I can think of is `tty` but this will be even more obscure to users not familiar with terminal emulation. Looking at documentation from terminal emulators, tmux says that panes are pseudo terminals, in screen they name them "virtual terminals", kitty seems to only speak of "windows", iterm2 says panes are different "terminal sessions"...

mawww:
I was thinking that the `new` command does not really need to be provided by each of those scripts anymore, it should be implementable on top of the `terminal` alias in a generic way, right ? The question is where to define it, maybe in its own `base/new.kak` file ?

occivink:
Agreed, I wasn't sure if we wanted this kind of implicit dependency until the dependency problem is solved. I've implemented this, and you can decide whether to keep the commit. 

I was also thinking that this one is a good usecase for the overlays you mentioned in #2602, since a user should be able to temporarily override what (for example if they generally prefer spawning an x11 term, but for fzf specifically they want to use a kitty pane).

rc/base/new-client.kak
```diff
@@ -0,0 +1,16 @@
+define-command new -params .. -command-completion -docstring '
+new [<commands>]: create a new kakoune client
+The ''terminal'' alias is being used to determine the user''s preferred terminal emulator
+The optional arguments are passed as commands to the new client' \
+%{
+    try %{
+        terminal %sh{
```
mawww: We should not need to have a shell block here, in the end I think the terminal command should take a list of arguments to make forwarding cleaner.

occivink: I don't think we can really make forwarding "clean" in this case, the `terminal` commands go through `sh -c` so the best they can do is join the arguments with a space. Just taking one argument makes things more explicit. 
As for the `new` command, we could bypass the shell block by doing `terminal "kak -e '%arg{@}'` but this has issues if the arguments to `new` contain single quotes. The current way seems better to me, even if it goes through a shell scope (which is negligible in this context).

mawww: * I'd like `terminal` to use -shell-completion, so that I can just do `:terminal <use shell command completion here directly from kakoune>`, if terminal takes a single string parameter that will not be possible.
* `:new my command as separate args` should forward to `terminal kak -c %val{session} -e "%arg{@}"` which will pass "my commands as separate args" as a single argument to the `terminal` command, this is effectively not fool proof, but thats already the behaviour we have right now with the `:new` command. (try `:new echo %{foo     bar}`, you'll see it output just `foo bar`).
* The shell escaping should be an implementation detail of the `terminal` command implementation, as it depends on how that works, if the `terminal` command knows the argument is passed as-is, there is no need for escaping, on the contrary if for some reasons it has to foward it through `sh -c`, then it needs to escape it.

Regarding the second point, it is still problematic for complex commands. I think its a more general problem that ought to be fixed another way, it seems like we need a foolproof way to do perfect list -> string -> list forwarding in Kakoune. I'd rather keep this terminal change simple and not try to fix it here.

occivink: Ok I think I see what you're getting at. I've changed the logic of the `terminal` commands so that the argument splitting done by kakoune is preserved. For example, in kakoune if you do
`:terminal printf "%s\n" a b 'c d' 'de''f' &` you will get
```
a
b
c d
de'f
&
```
Consequently, all the escaping logic is moved to the support scripts that need it (i.e. not kitty.kak and not tmux.kak). Anything else that goes through `sh -c` has some logic to preserve the arguments splitting. Some x11 terminals don't need it but go through it anyway so I guess these could still be improved in that regard.

If one actually needs to run things in a shell, they can still do 
`:terminal sh -c 'echo abc; sleep 5'`

If you're happy with this iteration, the iterm script will need some testing as its escaping is rather complicated.

mawww: Yep, I am happier with that iteration, can you rebase onto master instead of merging it ?

occivink:
Should I squash the commits? I think it makes some sense to keep the history.

occivink:
Yeah, in your case you have two options: temporarily alias `terminal` to one of `tmux-terminal-horizontal`... and then call `new`,
Or directly call `tmux-terminal-horizontal kak -c %val{session}`

One advantage is that now you can open a new pane with `$SHELL` with minimal work, judging by how vim integrated a terminal emulator that's a desirable feature.

andreyorst:
I've did this piece of code here https://github.com/andreyorst/fzf.kak/blob/d1bccaf3d70eadd6ad7162c69b6cc15211f9c0bb/rc/fzf.kak#L66

In this case it suppots older method too

andreyorst:
I wonder how to make an analog of `new` but which splits the tmux into top and bottom panes. I've used to have an alias to `tmux-new-vertical` for that, but now there's nothing like this. `tmux-terminal-vertical` accepts shell commands, while `tmux-new-vertical` accepted editor commands, as `new` does currently. That's looks like a downgrade to me

PR #2622, opened 2018-12-06
Remove highlighter definition from hook in just.kak

TeddyDD:
That was old hack to highlight `{{variable}}` substitutions in sh regions. This works too and does not require adding highlighter in hook.


TeddyDD:
Resolved by #2603

PR #2628, opened 2018-12-11
doc: Mention that commands requiring a client context do nothing in a draft context.

### LINKED ISSUE #2618: prompt command does nothing inside "eval -draft"
### Steps 

 1. Start a fresh Kakoune with no config:

        kak -n

 1. Type the following command:

        :eval -draft %{ prompt foo nop }<ret>

### Outcome 

Nothing happens.

### Expected 

A prompt "foo" appears, which asks for a string then ignores it.

The same command without `-draft` works fine:

    :eval %{ prompt foo nop }<ret>

Screwtapello:
Fixes #2618.

### DIFF
--- a/doc/pages/commands.asciidoc
+++ b/doc/pages/commands.asciidoc
@@ -284,10 +284,19 @@ but not really useful in that context.
     prompt content will be passed as a single token. In other words,
     word splitting does not take place.
 
+    NOTE: The prompt is displayed in and receives input from the
+    current client context, so inside a draft context like
+    `evaluate-commands -draft`, it is invisible and only responds to
+    an `execute-keys` command in the same context.
+
 *on-key* <command>::
     wait for next key from user, then execute <command>, the key is
     available through the `key` value, accessible through `$kak_key`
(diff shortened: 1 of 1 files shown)


Delapouite:
This PR needs rebasing.

Screwtapello:
Rebased.

mawww:
This is not quite correct, things like `on-key` do work in draft context, they are just not very useful as the key has to come in that same context.

```
eval -draft %{
    on-key %{echo -debug %val{key}}
    exec <tab>
}
```

This does write `<tab>` to the debug buffer. It is hence incorrect to says that `on-key` does nothing in draft contexts.

draft contexts do not have any clients to display things to, but they do have an input handlers, as they can handle keystrokes comming from `execute-keys`...

Screwtapello:
I've rebased again, and made the text more accurate as @mawww described. `prompt`, `on-key` and `menu` all mention that inside a draft context, they only respond to `execute-keys` within that same context. `info` doesn't respond to input at all, so I just mentioned that it's "invisible" rather than "does nothing".

PR #2632, opened 2018-12-13
rc: Don't hide errors produced by shell commands

lenormf:
Errors returned by shell utilities will be printed in the *debug*
buffer, hiding them prevents users from figuring out what's wrong
if commands they're excecuting silently fail.

### DIFF
--- a/rc/base/spell.kak
+++ b/rc/base/spell.kak
@@ -28,7 +28,7 @@ Formats of language supported:
         fi
 
         {
-            sed 's/^/^/' "$kak_opt_spell_tmp_file" | eval "aspell --byte-offsets -a $options" 2>&1 | {
+            sed 's/^/^/' "$kak_opt_spell_tmp_file" | eval "aspell --byte-offsets -a $options" | {
                 line_num=1
                 regions=$kak_timestamp
                 read line # drop the identification message
--- a/rc/base/tmux.kak
+++ b/rc/base/tmux.kak
@@ -22,7 +22,7 @@ define-command -hidden -params 1.. tmux-new-impl %{
         tmux_args="$1"
         shift
         if [ $# -ne 0 ]; then kakoune_params="-e '$@'"; fi
-        TMUX=$tmux tmux $tmux_args "env TMPDIR='${TMPDIR}' kak -c ${kak_session} ${kakoune_params}" < /dev/null > /dev/null 2>&1 &
+        TMUX=$tmux tmux $tmux_args "env TMPDIR='${TMPDIR}' kak -c ${kak_session} ${kakoune_params}" < /dev/null > /dev/null &
     }
 }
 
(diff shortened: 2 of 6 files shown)


rc/base/spell.kak
```diff
@@ -28,7 +28,7 @@ Formats of language supported:
         fi
 
         {
-            sed 's/^/^/' "$kak_opt_spell_tmp_file" | eval "aspell --byte-offsets -a $options" 2>&1 | {
```
mawww: stderr goes to /dev/null anyway

rc/base/tmux.kak
```diff
@@ -22,7 +22,7 @@ define-command -hidden -params 1.. tmux-new-impl %{
         tmux_args="$1"
         shift
         if [ $# -ne 0 ]; then kakoune_params="-e '$@'"; fi
-        TMUX=$tmux tmux $tmux_args "env TMPDIR='${TMPDIR}' kak -c ${kak_session} ${kakoune_params}" < /dev/null > /dev/null 2>&1 &
+        TMUX=$tmux tmux $tmux_args "env TMPDIR='${TMPDIR}' kak -c ${kak_session} ${kakoune_params}" < /dev/null > /dev/null &
```
mawww: This will prevent the command from running detached

rc/extra/kitty.kak
```diff
@@ -46,7 +46,7 @@ All optional parameters are forwarded to the new window} \
     -shell-completion \
     kitty-repl %{ evaluate-commands %sh{
         if [ $# -eq 0 ]; then cmd="${SHELL:-/bin/sh}"; else cmd="$*"; fi
-        kitty @ new-window --no-response --window-type $kak_opt_kitty_window_type --title kak_repl_window --cwd "$PWD" $cmd < /dev/null > /dev/null 2>&1 &
+        kitty @ new-window --no-response --window-type $kak_opt_kitty_window_type --title kak_repl_window --cwd "$PWD" $cmd < /dev/null > /dev/null &
```
mawww: Will prevent the command from running detached.

rc/base/tmux.kak
lenormf: Actually the whole chain of redirection is uncessary, it will work without any of them as well.

lenormf: No it won't, I've tested those changes and they work? I can close the first client, and the second one will remain usable.

Have you tested the PR?

mawww: I dont think we understood each other, the tmux command is going to return relatively quickly, its just sending a command to the tmux server. Yet there is a difference between the previous behaviour, where that tmux command was fired, then forgotten and the shell returned immediately, and the new behaviour would be to wait for that command to return (which, again, is not going to wait for the pane to be closed, just for the command to be acknowledged by the tmux server).

It would be ok not to launch that tmux command in parallel, I merely wanted to point out that this behaviour was changing, as it did not look that was obvious to you (as you did not remove the whole redirection and backgrounding).

PR #2639, opened 2018-12-16
Replaced rawgit.com urls with combinatronics.com urls.

c0mbinatr0nics:
Combinatronics.com is a drop in replacement for rawgit.com which is EOL.


lenormf:
We could also link to http://kakoune.org/img/kakoune_logo_full.png, instead of relying on an external service.

PR #2643, opened 2018-12-18
src: Add support for right click events

### LINKED ISSUE #843: Mouse right click to extend selection
Allow extending a selection or selections (by merging?) with the right click of the mouse same as it is possible in vim/emacs/xterm. `:h'mousemodel` in vim for reference

lenormf:
The current implementation treats left mouse button clicks as a
generic "mouse press" modifier, this commit extends the list of
modifiers by adding a "right mouse click" one.

The proper way to implement this would be to ship the coordinates
of mouse key press events in each `Key` object, and pass whichever
button was clicked as a codepoint value (instead of coordinates
currently), but this would require more work.

This commit allows:

* right clicks to set the cursor of the main selection
* control-right clicks to merge all the selections, and then set
  its cursor

Fixes #843

### DIFF
--- a/src/keys.hh
+++ b/src/keys.hh
@@ -21,16 +21,19 @@ struct Key
         Alt     = 1 << 1,
         Shift   = 1 << 2,
 
-        MousePress   = 1 << 3,
-        MouseRelease = 1 << 4,
-        MousePos     = 1 << 5,
-        MouseWheelDown = 1 << 6,
-        MouseWheelUp = 1 << 7,
-        MouseEvent = MousePress | MouseRelease | MousePos |
-                     MouseWheelDown | MouseWheelUp,
+        MousePressLeft    = 1 << 3,
+        MousePressRight   = 1 << 4,
+        MouseReleaseLeft  = 1 << 5,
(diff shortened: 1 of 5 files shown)


src/keys.hh
```diff
@@ -21,15 +21,16 @@ struct Key
...
 
-        MousePress   = 1 << 3,
-        MouseRelease = 1 << 4,
-        MousePos     = 1 << 5,
-        MouseWheelDown = 1 << 6,
-        MouseWheelUp = 1 << 7,
-        MouseEvent = MousePress | MouseRelease | MousePos |
-                     MouseWheelDown | MouseWheelUp,
-
-        Resize = 1 << 8,
```
lenormf: Note that `Resize` used to be `1 << 8` while the next entry `MenuSelect` is `1 << 10` - was that done on purpose, or was it simply an artifact?

lenormf:
Why do we need to know which mouse button was released? Isn't the handling code valid for both events?

lenormf:
Renamed "one" with "left" and "three" with "right"; added a release event for the right click.

PR #2645, opened 2018-12-19
Put ModeChange indent cleaning hooks into their own group

occivink:
* ModeChange hooks that remove indent now belong to *-trim-indent groups, instead of just -indent
* *-filter-around-selections hooks and commands have been renamed to trim-indent for clarity

This doesn't change anything regarding the status quo of indent removal, but this lets users opt out of the ModeChange hook (using disabled-hooks), which are too aggressive in my opinion.


occivink:
(you merged the pull request without the unit test fix)

PR #2646, opened 2018-12-19
Improve README: clarify the difference between 'alt-j' and 'alt-J' join commands.

jbrains:
I'm not sure how much detail you want to put into the README compared to other documents. I don't know, for example, where to find a guide like "translating vim commands into kakoune commands". I'm trying to add to the README whenever I find something surprising, but I don't want to overdo it. Let me know if I should add these clarifications somewhere else.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -454,9 +454,12 @@ Changes
 
  * `r`: replace each character with the next entered one
 
- * `<a-j>`: join selected lines
- * `<a-J>`: join selected lines and select spaces inserted
-            in place of line breaks
+ * `<a-j>`: join selected lines, replacing the line break with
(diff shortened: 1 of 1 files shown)


occivink:
The detailed README info should be phased out in favor of the pages that are accessible in the editor, so I'd suggest making theses changes to keys.asciidoc as well.

jbrains:
Thank you. Yes. I'll do that... eventually.

mawww:
Thanks for this PR, as @occivink said, we are trying to have the reference documentation be in the bundled pages (which are readable through github in `doc/pages/` as well), so I prefer to avoid having clearer information in the readme than in those. Please feel free to improve the documentation there.

PR #2648, opened 2018-12-21
src: Run and display the time taken by unit tests to run in debug mode

lenormf:
Knowing how much time the editor took to run unit tests gives users
a notion of how fast it's performing on a given system.

### DIFF
--- a/src/clock.hh
+++ b/src/clock.hh
@@ -8,6 +8,7 @@ namespace Kakoune
 
 using Clock = std::chrono::steady_clock;
 using TimePoint = Clock::time_point;
+using DurationMs = std::chrono::milliseconds;
 
 }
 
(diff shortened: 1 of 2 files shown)


src/unit_tests.cc
```diff
@@ -56,10 +56,13 @@ UnitTest test_diff{[]()
 
 UnitTest* UnitTest::list = nullptr;
 
-void UnitTest::run_all_tests()
+DurationMs UnitTest::run_all_tests()
```
mawww: I dont think this function should be measuring the time it takes, it should be done externally.

lenormf:
It's an informative value that I use to grossly compare how well a device can handle the editor, probably similar to the timings made around the hash map module.

Can we have that under the profiling debug value then?

PR #2650, opened 2018-12-21
Add quotes to expansions example

eraserhd:
Without them, newlines are eaten.  I have no idea why, but to demonstrate:

```
[jfelice@C02X421DJHD4 kakoune(quote-eval-set)]$ foo="'hello 
> there                       
> again '"                    
[jfelice@C02X421DJHD4 kakoune(quote-eval-set)]$ eval set -- $foo                                                         
[jfelice@C02X421DJHD4 kakoune(quote-eval-set)]$ echo "$1"   
hello there again             
[jfelice@C02X421DJHD4 kakoune(quote-eval-set)]$ eval set -- "$foo"                                                       
[jfelice@C02X421DJHD4 kakoune(quote-eval-set)]$ echo "$1"   
hello                         
there                         
again                         
```

```
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin17)
Copyright (C) 2007 Free Software Foundation, Inc.
```

### DIFF
--- a/doc/pages/expansions.asciidoc
+++ b/doc/pages/expansions.asciidoc
@@ -109,7 +109,7 @@ back the original values. For example, if you want to process the contents
 of each selection, you can do something like:
 
 ----
-eval set -- $kak_selections
+eval set -- "$kak_selections"
 while [ $# -gt 0 ]; do
     # ... do a thing with $1 ...
     shift


lenormf:
It's a bug from your shell.

lenormf:
Here is the explanation from a knowledgeable `zsh` user:

> okdana | most shells apply field-splitting to parameter expansions, by default zsh doesn't
> ironzorg |       why not?
> ironzorg |       I feel like the behaviour should be predictable in that case
> okdana | because it's a source of constant bugs, and unlike most shells zsh provides lots of ways to tell it exactly how you want to join/split/quote/whatever the result of an expansion
> okdana | it's probably zsh's most notable syntactic difference from other shells
> okdana | other shells with bourne-style grammar, anyway

Issue #2021, opened 2018-04-30
Persistent Undo


jupart:
Vim has a feature called `persistent_undo` that is a nice quality-of-life feature I really like. 

The implementation in nvim is like: It writes some kind of binary file to VIM_UNDO_DIR for each file. The undo files are written with their full paths separated by `%`, like `%home%name%path%to%changed%file.c`. 

Does `kak` plan to support something like this one day? I think I'd like to implement a plugin for it either way, but I'm not sure where to start other than just to dig into other plugins.

danr:
Note: the undo history is not exposed to plugins right now so there is no way to support this as a plugin without changing the c++ source code as well.

mawww:
Exposing the necessary data so that a support script can take care of storing/restoring would be my preferred way. The problem is how to expose the undo data. This is related to the requests for a nice interface to the undo tree, we need to expose the undo tree data to the external/script world and let them do their thing (a graphical UI should be able to get the undo data and provide its own interface).

So, first steps would be to make it possible to get the undo tree from Kakoune. For that a text format must be defined. An undo tree is a tree of history nodes. history nodes can be reduced to a list of modifications and a timestamp. modifications are a tuple (insert/erase, buffer position, text content).

Second phase would be to make it possible to set the undo tree from a script. This is tricky because we need to validate that the undo tree is valid with the current state of the buffer.

yshui:
@mawww A history node should also contain its parent node. It is a undo **tree** after all.

And for validating, we can probably store and verify a checksum of the buffer.

mawww:
@yshui Right, the parent would be an index in the array. (so history node would be a tuple (parent index, timestamp, list of modifications).

Validating with a checksum is not safe enough, this is arbitrary data coming from an external source, Kakoune cannot trust it to be valid, it needs to either check it, or tolerate invalid undo data.

yshui:
@mawww I was thinking more about "how would the script figure out if the undo data is still valid" aspect.

I agree kakoune needs to check the undo data, but that's probably not difficult to do.

mawww:
Its pretty involved because each undo node in the undo tree applies to its parent buffer state, so in order to check that an undo node is valid, we need to compute its parent buffer state. Trying to do that without having to apply the modifications to the buffer each time might be a bit tricky to implement, I need to take a look.

[...]

PR #2670, opened 2018-12-30
Remove unused history timepoint field

eraserhd:
I noticed this while trying to work out a text format for #2021.


eraserhd:
Seems like this field was added for Vim-`:earlier`-like functionality, and keeping the field and presenting it in #2021 will enable external plugins to do that.

PR #2673, opened 2019-01-01
Add 'history' expansion

### LINKED ISSUE #745: Have an interface for the history tree, document it

eraserhd:
Closes #745 and aids #2021.


doc/pages/expansions.asciidoc
```diff
@@ -218,6 +218,23 @@ The following expansions are supported (with required context _in italics_):
     _in window scope_ +
     line of the main cursor, the third component of `%val{selection_desc}`
 
+*%val{history}*::
+    _in buffer, window scope_ +
+    the full change history of the buffer, including undo forks, as a
+    _history_id_ followed by one or more history entries. _history_id_ may
+    be different from `%val{history_id}` to capture uncommitted changes, for
```
Screwtapello: I want to know more about these uncommitted changes, and how they get committed. Are they related to Git commits? Is there any point in the program's lifecycle where there are guaranteed to be no uncommitted changes?

I think this means something like "if the user has entered insert mode and begun typing, but not yet hit Escape, _history_id_ may be different from `%val{history_id}` to represent those uncommitted changes."

eraserhd:
For `parinfer-rust`, I would have to track the last history_id seen and the number of modifications we've processed from that history_id, since it wants to process uncommitted changes as they occur.

We could save a lot of space by merging modifications on commit, but that would make it hard to figure out what to do in parinfer, though, since it is possible we've already applied part of a committed change when it was uncommitted.

doc/pages/expansions.asciidoc
```diff
@@ -218,6 +218,23 @@ The following expansions are supported (with required context _in italics_):
...
+    `parent committed redo_child modification0 modification1 ...`.
+    _parent_ is the index of the entry's predecessor.  Entry 0, which is the
+    root of the history tree, will always have `-` for _parent_. _committed_
+    is the count of seconds from January 1, 1970 to the creation of the
+    history entry, or `-` if the entry represents uncommitted changes.
+    _redo_child_ is the index of the child which will be visited for `U` and
+    will be `-` at the leaves of the history. Each _modification_ is in the
+    format `op|line.column|text`, where _op_ is `+` for insertions or `-`
+    for deletions, _line_ and _column_ are the 1-based position of the start
+    of the change, and _text_ is the content of the insertion or deletion.
```
Screwtapello: "column" gets messy when you have files with code-points that can take two or zero columns on-screen. It would be better to define these as "in the same format used for the anchor and cursor positions in `%val{selection_desc}`" or "in the same format used for the anchor and cursor positions in the `:select` command" or similar.

eraserhd:
Maybe it would be much cleaner to have a separate expansion for uncommitted changes?  Gonna sleep on it.

Screwtapello:
It would be nice if there were a way to say "get all the history items since..." instead of reparsing and rebulding the tree every time.

I guess *importing* history is still an open problem.

doc/pages/expansions.asciidoc
eraserhd: Your last paragraph is correct, it is not related to git.  Maybe the text "changes not yet saved to the history, such typing in insert mode before pressing `<esc>`" would be clearer?

Screwtapello: I like that, yes!

doc/pages/expansions.asciidoc
```diff
@@ -218,6 +218,23 @@ The following expansions are supported (with required context _in italics_):
...
+*%val{history}*::
+    _in buffer, window scope_ +
+    the full change history of the buffer, including undo forks, as a
+    _history_id_ followed by one or more history entries. _history_id_ may
+    be different from `%val{history_id}` to capture uncommitted changes, for
+    example from an InsertChar hook. Each entry is
+    `parent committed redo_child modification0 modification1 ...`.
+    _parent_ is the index of the entry's predecessor.  Entry 0, which is the
+    root of the history tree, will always have `-` for _parent_. _committed_
+    is the count of seconds from January 1, 1970 to the creation of the
```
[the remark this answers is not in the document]
eraserhd: Hmm, in the case of parinfer, whenever it fixes parens it will create a peer node with almost certainly the same timestamp.  For things like `:earlier`, we can prefer the node with the highest index.

In this case, it doesn't make sense to select the old node, since parinfer would re-fix it anyway.  I suppose unless we've turned parinfer off because it got confused and broke our parens.

I propose that if we need higher resolution, we add a decimal portion.  This seems compatible enough to do later if needed?

doc/pages/expansions.asciidoc
eraserhd: done.

src/buffer.cc
```diff
@@ -728,6 +728,46 @@ String Buffer::debug_description() const
...
+    auto format_history_node = [&](const HistoryNode& node) {
+        return format("{} {} {}{}{}",
+                      node.parent,
+                      WallClock::to_time_t(node.committed),
+                      node.redo_child,
+                      node.undo_group.empty() ? "" : " ",
+                      join(node.undo_group |transform(format_modification), ' ', false));
+    };
+    auto history_id = m_history_id;
+    auto history = m_history;
```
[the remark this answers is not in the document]
eraserhd: I split apart the committed and uncommitted change expansions, which now avoids copying the history.

eraserhd:
@Screwtapello I've been keeping importing history in mind.  We'll need to add a built-in command which validates the history, but this format should work.

src/buffer.cc
```diff
@@ -728,6 +728,50 @@ String Buffer::debug_description() const
...
+
+String Buffer::format_modification(const Modification& modification, Quoting quoting) {
+    auto quote = quoter(quoting);
+    return quote(format("{}|{}.{}|{}",
+                        modification.type == Buffer::Modification::Type::Insert ? '+' : '-',
+                        modification.coord.line, modification.coord.column,
+                        modification.content->strview()));
+}
+
+String Buffer::history_as_string(Quoting quoting) const
```
mawww: I dont think those should be in Buffer itself, I would prefer for buffer to expose a getter for history and uncommited modifications, and have free functions in buffer_utils.hh that format those to strings.

(Its okay to make Modification a public type in Buffer in order to achieve that).

src/buffer.hh
```diff
@@ -261,14 +265,16 @@ private:
     Flags  m_flags;
 
     using UndoGroup = Vector<Modification, MemoryDomain::BufferMeta>;
+    using WallClock = std::chrono::system_clock;
+    using WallTime = WallClock::time_point;
```
mawww: Not sure using the system_clock is better than the steady_clock we use by default, as that means we now potentially have children with timestamp earlier than their parent (if the clock was changed in the mean time). I think its safer to just use the existing Clock and TimePoint.

ganwell:
@eraserhd 

Paste creates a multi-character modifications:

```
{"kak_history":"- 1575757783 -","kak_buffile":"*scratch*"}
{"kak_buffile":"*scratch*","kak_history":"- 1575757783 1 0 1575757799 - +|0.1| +|"}
{"kak_buffile":"*scratch*","kak_history":"- 1575757783 1 0 1575757799 2 +|0.1| +| 1 1575757811 - +|0.1| +|0.-1|x"}
```
I pasted ` +|` then ` +|0.-1|x`.

Because of that the serialization format is not sound, at least I couldn't figure out a reliable way to parse it. I think 

1. either the multi-character modifications have to be split into single-character modifications

2. a length argument must be added

3. some kind of escaping must be introduced

4. or json as serialization format must be used

I think 1. would be nice, because it is easy to parse.

I currently use this programm:

https://github.com/ganwell/hitch/commit/9bc0b6a904ad2183e5a78d97650bec395ee33010#diff-a84ae920b8f85f66f7c23d48829582ad

with these hooks

```
hook global InsertIdle .* %{ nop %sh{ ./target/debug/kak-env-to-fifo /tmp/out kak_buffile kak_uncommitted_modifications } }
hook global NormalIdle .* %{ nop %sh{ ./target/debug/kak-env-to-fifo /tmp/out kak_buffile kak_history } }
```

and

```
tail -f /tmp/out
```

to test the history expansion.

ganwell:
In the proposal the format is `'+|1.0|'`, would you then escape `'` with `\'`?

eraserhd:
@ganwell This PR is old enough that it was before quoting was selectable by the user.  If you applied this to a more recent Kakoune, you should use `$kak_quoted_history`.

In any case, the format is designed to allow multi-character changes unambiguously.  The quotes allow parsing elements, and with shell quoting, `'` should become `'\''`.

ganwell:
> @ganwell This PR is old enough that it was before quoting was selectable by the user. If you applied this to a more recent Kakoune, you should use `$kak_quoted_history`.
> 
> In any case, the format is designed to allow multi-character changes unambiguously. The quotes allow parsing elements, and with shell quoting, `'` should become `'\''`.

Hmm, I changed the format to

```cpp
diff --git a/src/buffer.cc b/src/buffer.cc
index 3a3d9837..173dbe46 100644
--- a/src/buffer.cc
+++ b/src/buffer.cc
@@ -743,9 +743,10 @@ InplaceString<23> to_string(Buffer::HistoryId id)

 String Buffer::format_modification(const Modification& modification, Quoting quoting) {
     auto quote = quoter(quoting);
-    return quote(format("{}|{}.{}|{}",
+    return quote(format("{}|{}.{}.{}|{}",
                         modification.type == Buffer::Modification::Type::Insert ? '+' : '-',
                         modification.coord.line, modification.coord.column,
+                        modification.content->strview().length(),
                         modification.content->strview()));
 }
```
I of course will use what every gets merged in this PR, but with the length, parsing is a lot easier.

I am currently a bit puzzled that: 🧑 makes coord.column increase by four. But the library I am using client-side expects just that. I thought it would be better to use character count for column (it is called column).

ganwell:
Here is the parser I currently use: https://github.com/ganwell/hitch/blob/025b8c8eb4761ec1f2c90d7f43aa078d93fc0c6e/kak-buffer-sync/src/lib.rs#L133

Screwtapello:
Generally Kakoune's internal coordinates use "bytes" as the unit (for example, `%val{selection_desc}`), so emoji like that are expected to have width 4.

eraserhd:
Rebased the branch.  I'll work to address feedback over the next few days.

One thing I have to think about is that I was motivated to do this by parinfer, which is now using display coordinates instead of byte coordinates (and therefore handling Unicode properly).  There's enough information here, but I have to figure out how to do it.

ganwell:
> One thing I have to think about is that I was motivated to do this by parinfer, which is now using display coordinates instead of byte coordinates (and therefore handling Unicode properly). There's enough information here, but I have to figure out how to do it.

I'd be glad if you don't change to grapheme-coordinates in the history expansion. I think it would be better to convert to grapheme-coordinates on client side. All my use-cases work with byte-coordinates, so I can avoid any potentially buggy conversion between coordinates. Since kakoune uses byte-coordinates, it would be strange and confusing if the history expansion was different.

eraserhd:
@mawww All feedback addressed.

doc/pages/expansions.asciidoc
```diff
@@ -353,6 +365,15 @@ The following expansions are supported (with required context _in italics_):
     buffer is modified, including undoing and redoing previous modifications
     (see also `%val{history_id}`)
 
+*%val{uncommitted_modifications}*::
```
mawww: I wonder if this should be exposed directly as part of `history` as a special last element ? Do you see cases where we would want to get history, but not uncommitted_modifications ?

src/buffer_utils.cc
```diff
@@ -213,4 +213,47 @@ void write_to_debug_buffer(StringView str)
     }
 }
 
+InplaceString<23> to_string(Buffer::HistoryId id)
+{
```
mawww:
```
    if (id == Buffer::HistoryId::Invalid)
        return InplaceString<23>{1, "-"};
    return to_string(static_cast<size_t>(id));
```

src/buffer_utils.cc
```diff
@@ -213,4 +213,47 @@ void write_to_debug_buffer(StringView str)
...
+        InplaceString<23> res;
+        res.m_data[0] = '-';
+        res.m_length = 1;
+        return res;
+    } else {
+        return to_string(static_cast<size_t>(id));
+    }
+}
+
+String format_modification(const Buffer::Modification& modification, Quoting quoting)
```
mawww: This does not need to be exposed, and could just be the body of `modification_as_string`

src/buffer_utils.hh
```diff
@@ -84,6 +84,11 @@ void reload_file_buffer(Buffer& buffer);
 
 void write_to_debug_buffer(StringView str);
 
+InplaceString<23> to_string(Buffer::HistoryId id);
+String format_modification(const Buffer::Modification& modification, Quoting quoting);
```
mawww: No need to expose this one.

test/compose/history/rc
```diff
@@ -0,0 +1,7 @@
+# Make our expansion have a predictable timestamp
+hook global ClientClose .* %{
+    evaluate-commands %sh{
```
mawww: We dont have any kakoune commands we want to evaluate from this output, so `nop` would be better.

test/compose/history/rc
```diff
@@ -0,0 +1,7 @@
...
+        kak -f 'ghf<space>ec$timestamp<esc>2f<space>ec$timestamp<esc>' <kak_quoted_history >tmp
+        mv tmp kak_quoted_history
```
mawww: just pass kak_quoted_history as an argument to `kak -f` to edit the file in place, no need for the tmp dance.

mawww:
Thinking a bit more about this, I think there is one missing issue we need to solve:

The state of the buffer is `history id` + `current uncommitted modifications`, so to move from one state to another, we need to keep the current list of uncommitted modifications, revert it, walk our way in the tree to ne new `history id` (reverting until common ancestors, then applying until the new `history id`).

That means we need to store the current ucommitted modification list, as it is a mutable data structure in Kakoune (it gets cleared on undo group commit, it can grow, we could even add a bit of compression logic to it (join adjacent modifications, so that typing `w` `o` `r` `d` gets compressed to a single modification inserting `word` instead of 4 adjacent ones)). Kakoune cannot really help there, the client script has to keep track of that list along with the history id.

I suppose Kakoune could still provide a way to walk the history tree between history ids (say `kak_modifications_from_$id_to_$id`) which means a script would be able to update its view of the buffer with

```
for each modifications in
  concat_lists(
    reverse($saved_kak_uncommitted_modifications),
    $kak_modifications_from_${kak_saved_history_id}_to_${kak_history_id},
    $kak_uncommitted_modification)) {
  apply modification;
}
```

`reverse` is trivial (reverse the list and swap Insert/Erase in each element). getting `$kak_modifcation_from_...` could be done with `kak -p 'eval -buffer=... echo -to-file blah %val{modifications_from_%d_to_%d}`, so the only non-trivial logic would be implementing `apply modification` which should not the that hard (although it could be slow if the underlying buffer data structure is not optimized for line/column based access).

So in the end, after this thinking out-loud comment, I think the current design is fine, in particular it does make sense to keep uncommitted_modification separate as it does require specialized handling and trying to merge it inside $kak_history would likely just complexify parsing.

mawww:
Applied the suggested modifications myself as I had done them locally anyway.

eraserhd:
Thanks!  The `kak_history_from_$id_to_$id` thing has some interesting wrinkles.  It can't work with our shell expansions as it is unless the ids are hard-coded into the shell script, which isn't useful.  The from and to would have to be variable names, e.g. `${kak_history_from_${kak_opt_fooplugin_lasthistory}_${kak_timestamp}}`.  Except that doesn't work in shell -- it would require `eval` and some token pasting that might be hard to parse.

The file-descriptor thing for requesting expansions would help here.

ganwell:
@eraserhd Big thanks! :tada: also to @mawww 

About the `$kak_modifications_from_${kak_saved_history_id}_to_${kak_history_id}` it has to be double evaluated, so the actual IDs are in the shell code kakoune sees. I tried this and failed.

Could it be a command that stores the history in a register. I currently use `%` and the dot register when I have to pass the whole buffer.

https://git.io/JepXe

PR #2677, opened 2019-01-06
c-family: Regroup C++ keywords and consider C++20

fsub:
Besides adding [keywords introduced in C++20](https://en.cppreference.com/w/cpp/keyword) I moved a few identifiers into seemingly more appropriate categories. Although it's not perfect it hopefully is an improvement.


occivink:
What about moving `auto` to types as well? In modern C++ you'd see it much more often as a type than an attribute

fsub:
@occivink I thought about that but I wasn't sure. Although `auto` is often used instead of a concrete type it's not a type. I'm also unsure about whether `template` or `concept` should be in the same category as `class`. Finally, things like `delete` have multiple roles. On the one hand it is a counterpart to `new` in another context it falls into the same category as `default`.

Issue #2037, opened 2018-05-10
formatter.kak: Cursor position lost after executing the format command


fsub:
### Steps 

- :set-option buffer formatcmd "clang-format --assume-filename ${kak_buffile}"
- open some C++ source file
- move the cursor to line x, column y
- :format

### Outcome 

- the cursor is placed at the end of the current buffer

### Expected 

- the cursor should be located at its original position, i.e. line x, column y (see #1492)

Screwtapello:
This happened to me the other day with `rustfmt` set up as `formatcmd`. But I think that was in a long-running Kakoune instance so maybe the running Kakoune binary was out of sync with the `.kak` files on disk... I didn't have the patience to look deeper. Perhaps I was wrong, though.

fsub:
@Screwtapello Thank you for the hint. In my case, the problem can be reproduced even with a fresh kakoune session and with `cat` as `formatcmd`. But the issue seems to be related to my config.

Screwtapello:
If you figure out what the problem was, let me know!

fsub:
`apply_diff()` in `normal.cc` has a fixed internal size limit. If this limit is exceeded, instead of performing a detailed diff, the buffer is cleared and rewritten.

mawww:
This is actually related to pipe and size, we apply a diff algorithm to determine what changed after a pipe, but as this can get slow, Kakoune limits it to 100KB, if the selection is bigger, it will just erase and replace, leading to the behaviour you experience.

That 100KB limit was added to avoid those slow downs. I guess we could raise that slightly.

lenormf:
The `ko.cpp` file is only 62KB, even if bumping up the hard-coded limit works for this file, it's not unusual to have source files as big as a few hundred KBs. Is there a way to make this algorithm more efficient, or to use one that crunches more data?

mawww:
We can probably make the algorithm much more efficient by diffing lines first, then diffing characters (or maybe just at the word level ?). Right now we are diffing at the character level, which is what makes this much slower than say reloading a file (where we do diff the buffer state and the filesystem state at the line level).

fsub:
I originally experienced the problem when trying to format about 10k lines of C++ code. Bumping the limit from 100KB to 5000KB seems to work for my expected use cases. For me, a small amount of lag is acceptable. It seems hard to tell if there is a good choice that serves most users. If there is an improved algorithm it could relax the situation.

Screwtapello:
[...]

PR #2678, opened 2019-01-06
Increase buffer size limit in apply_diff() in order to mitigate issue…

fsub:
… #2037

### DIFF
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -493,7 +493,7 @@ void apply_diff(Buffer& buffer, BufferCoord pos, StringView before, StringView a
 {
     // The diff algorithm is O(ND) with N the sum of string len, and D the diff count
     // do not use it if our data is too big
-    constexpr ByteCount size_limit = 100 * 1024;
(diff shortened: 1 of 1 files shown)


lenormf:
Is that the smallest value you would need to be able to format the file mentioned in #2037?

Jumping from 100k to 5M doesn't look very appealing, considering the original value was already supposed to be enough for most cases.

fsub:
@lenormf `ko.txt` mentioned in #2037 is just a synthetic example that allows reproducing the issue. It consists of about 13K (short) lines. IIRC this is just big enough such that a buffer size of 100K is _not_ sufficient. I'm regularly working on files that are reasonably larger. 5M is the result of a rough estimate of what can realistically happen in my space. I've been using a correspondingly patched version of kakoune for several months. So far, I haven't observed any drawbacks. The one case that might be problematic is editing big files in severely restricted environments. People working with small files won't be affected. After all, we are talking about a buffer size _limit_.

PR #2688, opened 2019-01-14
Cache clojure highlighting info

eraserhd:
Related to #2683, I thought, "What if shell expansions handled their own caching manually?  Would that be too bad?"

This is the result.  I think it isn't too bad, actually.  (`clojure.kak` was the second worst offender on startup script timings on some list I saw a while back.)

Before:
```
[jfelice@C02X421DJHD4 kakoune(master)]$ time for ((i = 0; i < 1000; i++)); do kak -n -e 'source rc/base/clojure.kak; quit'; done                                    

real    1m12.413s                        
user    0m37.475s                        
sys     0m35.105s                        
```

After:
```
[jfelice@C02X421DJHD4 kakoune(cache-clojure-highlighting)]$ time for ((i = 0; i < 1000; i++)); do kak -n -e 'source rc/base/clojure.kak; quit'; done                

real    0m20.221s                        
user    0m5.919s                         
sys     0m10.725s                        
```

### DIFF
--- a/rc/base/clojure.kak
+++ b/rc/base/clojure.kak
@@ -28,6 +28,14 @@ hook global WinSetOption filetype=clojure %{
 }
 
 evaluate-commands %sh{
+    cache_version=1 # Change when below code is changed
+    cache_dir="${XDG_CACHE_HOME:-$HOME/.cache}/kak"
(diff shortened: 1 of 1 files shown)


TeddyDD:
Wow this is nice since it works with current system.

PR #2691, opened 2019-01-15
Add Ubuntu apt install instructions

Gabgab2003:
Add explaination to installing kak from the officicial repos


Screwtapello:
It's worth pointing out that Kakoune is still under development and the version in the repositories is likely to be somewhat out of date.

Gabgab2003:
Yeah I just noticed, it's the version from Dec 2016

Issue #2250, opened 2018-07-25
ctags.kak can't open absolute paths


alyssais:
### Steps 

Given a tags file containing absolute paths:

```
UserRefreshCredentials	/nix/store/b075lxb0bx6af95ycfhjv25w3h7ndj4v-dev/lib/ruby/gems/2.3.0/gems/googleauth-0.5.3/lib/googleauth/user_refresh.rb	/^    class UserRefreshCredentials < Signet::OAuth2::Client$/;"	c	module:Google.Auth
```

### Outcome 

Attempting to jump to this definition using `ctags-search` will fail, because it will be appended to the directory of the tags file.

### Expected 

`ctags-search` jumps to the appropriate definition.

PR #2703, opened 2019-01-22
Fix ctags plugin

stevenyvr987:
1. As noted in #2250, if the tags file contains absolute paths, the ctags-search command can't find the destination of a function definition. This will happen if the tags file is generated externally (not by the ctags-generate command, eg, the command line 'ctags -R $HOME/project' is used). The plugin presumed the destination is a relative path and prefixed it with a tagroot to make it absolute. The fix is to check for an absolute path, and if found then omit the prefix.

1. The command ctags-funcinfo is supposed to show the signature of a selected function. The relevant sed regex uses a mandatory 'signature:' pattern and expects to find the signature value as part of the extended fields. However, the default ctagscmd ('ctags -R') doesn't include signature as one of the default extended fields; the consequence is that the sed regex fails and the whole ctags line is shown instead. The fix is to ensure that the default ctags command generates the signature field ('ctags -R --fields=+S'). With the fix, just the function name and function signature is shown instead of the whole ctags line. Tested with c and c++ source code.


mawww:
Thank you for those fixes !

PR #2718, opened 2019-02-01
Push selection to jump list for cursor movements with count

### LINKED ISSUE #2709: [Feature request] j/k with count push to jump list
I often use something like <kbd>20k</kbd> to jump to a certain position in the file. Currently, to go back to where I came from, I then need to do <kbd>20j</kbd>. 

It would be nice if j/k when executed with a count could be pushed to the jump list as it is effectively a jump. This way, I could use `<c-o>` to jump back.

JJK96:
Fixes #2709


lenormf:
Isn't this going to make RAM use huge? Pushing coordinates onto a container every time the cursor moves is bound to create a huge jump list over time.

JJK96:
It's not on every move, only every move with a count, which reduces the number greatly (in my case)

aver-d:
I see your reasoning for why this could be convenient for occasional movements with large counts (removing the need to press `<c-s>` beforehand – which is easily forgotten). But I think this PR may have unintended consequences for some users depending on their kakrc and plugins. For example, for faster up/down movement, I have `5k` and `5j` mapped to their own keys. Storing that each time would result in a massive jump list of hundreds or thousands of five-line cursor changes.

Perhaps a higher minimum count could be hardcoded or made an option, or alternatively, if it's of any help, perhaps something similar to the function below could also work for you. It makes use of Kakoune's existing tools, but also has the advantage of being more flexible, allowing the user to set a threshold for when to apply the jumplist save (using 10 here as an example).

    def jk-save-jump -hidden -params 1 %{
      eval %sh{
        save=''
        if test "$kak_count" -ge 10; then
          save='<c-s>'
        fi
        printf 'exec %s%d%s' "$save" "$kak_count" "$1"
      }
    }

    map global normal j ': jk-save-jump j<ret>'
    map global normal k ': jk-save-jump k<ret>'

JJK96:
That's a valid point, it might still get big, though I don't expect memory usage will be a problem ( hundreds is not that much ) it might get in the way of other jumps.

Thanks for thinking of a solution for me!
However, I have been using this on my fork for the past 2 months without problems, so I will probably keep using that

Delapouite:
> Pushing coordinates onto a container every time the cursor moves is bound to create a huge jump list over time.

Maybe a (configurable) limit on the size of jump list could be added, with old jumps being pruned to let room for newer ones?

JJK96:
That might be a good idea, but I think it should be added somewhere outside of this pull request, since the jump list may grow for any other reason than this PR

JJK96:
I've been using this for more than two years now, no memory issues whatsoever

lenormf:
How did you measure the memory overhead and what are the numbers?

JJK96:
I never noticed kakoune using much memory. I'll try to gather some statistics on actual memory usage. But I can't really compare with the normal usage.

mawww:
I would not expect memory usage to become such a big issue in general cases, I guess it could be if we have many selections and use lots of count based movements but I'd expect this to be pretty rare.

I am more concerned about pushing a lot of elements that are not really usefull to the user in the jump list, forcing much more jumps than necessary. Its a hard design question because it really depends on user preferences, and as you probably know now, I am usually relunctant to add options to control editing behaviours as they tend to complexify scripting.

I actually would like to take the time to redesign the jump list functionality at some point and introduce the regularly requested selection undo feature. My current idea would be that selection undo would be cleared once we modify the buffer, making it just an easy way to recover from a selection mistake. Because those two features are very similar I think it would make sense to think how they would interact and what should go in each list.

All that to say I still dont really have an opinion on whether I want to merge this or not, which tends to favor the status quo.

PR #2719, opened 2019-02-01
Add explanation of count for s and S commands

JJK96:


occivink:
Maybe put `count` as monospace so that the sentence sound less weird?

JJK96:
I Followed the same format as in other places in the file (like the @ command) But maybe it's a good idea to make all those cases monospace

PR #2723, opened 2019-02-04
ctags: search tags file in parent directories

fsub:
If a tags file is not found at a given path relative to the current working
directory see whether it can be found relative to some parent directory.
Stop on the first match.
Precedent: lookup of .clang-format or .editorconfig.

### DIFF
--- a/rc/base/ctags.kak
+++ b/rc/base/ctags.kak
@@ -9,9 +9,10 @@ declare-option -docstring "list of paths to tag files to parse when looking up a
 define-command -params ..1 \
     -shell-script-candidates %{
         realpath() { ( path=$(readlink "$1"); cd "$(dirname "$1")"; printf "%s/%s\n" "$(pwd -P)" "$(basename "$1")" ) }
+        search() { [ -f "$1" ] && realpath "$1" || [ "$1" != "${1#/}" ] || ( while [ "$(pwd)" != "/" ]; do cd ..; [ -f "$1" ] && realpath "$1" && break; done ) }
         eval "set -- $kak_opt_ctagsfiles"
         for candidate in "$@"; do
-            [ -f "$candidate" ] && realpath "$candidate"
+            search "$candidate"
         done | awk '!x[$0]++' | # remove duplicates
         while read -r tags; do
             namecache="${tags%/*}/.kak.${tags##*/}.namecache"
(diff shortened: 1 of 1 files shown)


rc/base/ctags.kak
```diff
@@ -9,9 +9,10 @@ declare-option -docstring "list of paths to tag files to parse when looking up a
 define-command -params ..1 \
     -shell-script-candidates %{
         realpath() { ( path=$(readlink "$1"); cd "$(dirname "$1")"; printf "%s/%s\n" "$(pwd -P)" "$(basename "$1")" ) }
+        search() { [ -f "$1" ] && realpath "$1" || [ "$1" != "${1#/}" ] || ( while [ "$(pwd)" != "/" ]; do cd ..; [ -f "$1" ] && realpath "$1" && break; done ) }
```
occivink: I don't have any stake in the maintenance of this script, but I find this style of doing lots of things on one line rather hard to read.

fsub: @occivink I totally agree. When changing someone else's code I always try to follow the prevailing style. If the file should be reformatted I can do it.

andreyorst: should one search past `#$HOME`? This may cause permission error on some machines, where users can't access files beyond their home directory.

lenormf:
The `ctagsfiles` option exists so that the script doesn't need to execute logic that the user doesn't expect.

If you want to auto-detect tag files, write a shell scope that populates `ctagsfiles` when a new buffer is opened, in your user configuration.

rc/base/ctags.kak
fsub: @andreyorst In such an enviroment, what does `cd ..` do when called in `$HOME`?

fsub:
@lenormf Thank you for providing an alternative. How does it detect a new tags file that is created after the buffer was opened?

lenormf:
It doesn't, fortunately you could just wrap the auto-detection code into a command and call it on the rare occasion you would need it. Read the documentation about how to do that in `:doc commands declaring-new-commands`.

It's also better than making the editor poke through the user's filesystem every time a tag is being searched for. This process would prove very slow e.g. when the current buffer is located deep into a network mount-point, and the user would have to revert the hunk of this PR, which is more cumbersome than letting people who want that behaviour implement it in their user configuration.

mawww:
I agree with @lenormf on this one, the point of `ctagsfile` is to make it possible to look at different places in the the filesystem already, I would prefer to keep a relatively simple logic on how those paths are interpretted.

fsub:
@mawww C'est tout à fait bien. As always, I have only offered what has proven helpful in my environment. I see that the proposed change might not be helpful for everyone.
Thank you very much for all the work you put into kakoune!

mawww:
@fsub Thanks for opening the PR, happy Kakoune proves useful to you.

PR #2732, opened 2019-02-09
Include more details in test/README.asciidoc

stevenyvr987:
The test/README asciidoc file includes an ascii diagram showing the layout of files for a test suite. I've included a section to describe the role and expected content of each file.


lenormf:
Could you add some markup to the text, for instance directory names that are enclosed in `+` signs in the rest of the document.

Also replace artifacts such as "ie" or "eg" with the abbreviated text: `i.e.`, `e.g.`.

test/README.asciidoc
```diff
@@ -12,12 +12,14 @@ Source structure
...
 └── compose
     └── …
+        ├── [rc]         → configuration
         ├── cmd          → command
         ├── [in]         → start file
         ├── [out]        → end file
         ├── [selections] → selection contents
         ├── [state]      → selection states
-        └── [rc]         → configuration
+        ├── [display]    → out display in JSON format
```
mawww: That is out of date, and should be named `ui-out`, there is also a `ui-in` that can contain json commands as they would be sent by a json-ui.

test/README.asciidoc
```diff
@@ -12,12 +12,14 @@ Source structure
 │   └── …
 └── compose
     └── …
+        ├── [rc]         → configuration
         ├── cmd          → command
         ├── [in]         → start file
         ├── [out]        → end file
         ├── [selections] → selection contents
```
mawww: I think we should rename those files so that they match the variable name (rename state to selections_desc)

stevenyvr987:
I've just forced pushed an amendment to the PR. It adds more markup, changes ie and eg to _i.e._ and _e.g._, and updates the file names wrt `ui-in` and `ui-out`.

I didn't do anything about changing `state` to `selections_desc`, because it would need changes to the run shell script and to some existing test suites. I can do that in another PR.

test/README.asciidoc
```diff
@@ -26,3 +29,48 @@ Usage
...
+and should contain a key sequence that will edit the input buffer.
++cmd+ is executed after the +rc+ command sequence is sourced.
+
++in+ is optional
+and should contain the initial text loaded into the input buffer
+for editing by the +cmd+ key sequence.
+
++ui-in+ is optional
+and should contain the initial json-rpc commands loaded from the UI
+at startup to set a specific UI state.
```
mawww: Not really, ui-in simulates the json-rpc commands sent by the ui, its not an "initial" set of commands in the sense that its the only set of ui commands coming in. For example, we could get rid of cmd and encode all the keystrokes directly as UI key commands inside the ui-in file, but that would be much more cumbersome to write.

stevenyvr987:
The original wording was a paraphrase from commit a64afd7, which introduced `ui-in`. I've just force-pushed the PR to slightly change the wording.

Issue #2669, opened 2018-12-29
Add a signature to map command


alexherbo2:
### Prototype

```
map {scope} {mode} [signature] {key} {keys}
```

With _signature_ being a regular expression.

### Signatures

``` yaml
object:
- name: select-whole
  key: '<a-a>'
- name: select-whole-start
  key: '['
- name: select-whole-end
  key: ']'
- name: extend-whole-start
  key: '{'
- name: extend-whole-end
  key: '}'
- name: select-inner
  key: '<a-i>'
- name: select-inner-start
  key: '<a-[>'
- name: select-inner-end
  key: '<a-]>'
- name: extend-inner-start
  key: '<a-{>'
- name: extend-inner-end
  key: '<a-}>'
prompt:
- name: command
  key: ':'
- name: insert-output
  key: '!'
- name: append-output
  key: '<a-!>'
- name: pipe
  key: '|'
- name: pipe-to
  key: '<a-|>'
- name: pipe-keep
  key: '$'
- name: select
  key: 's'
- name: split
  key: 'S'
- name: keep
  key: '<a-k>'
- name: clear
  key: '<a-K>'
- name: select-next-match
  key: '/'
- name: select-previous-match
  key: '<a-/>'
- name: extend-next-match
  key: '?'
- name: extend-previous-match
  key: '<a-?>'
go-to:
- name: select
  key: 'g'
- name: extend
  key: 'G'
view:
- name: default
  key: 'v'
- name: lock
  key: 'V'
```

#### Objects

Selecting nodes with [kak-tree]:

``` kak
map global object select-(whole|inner) n -docstring 'Node' '<esc>:tree-select-node<ret>'
map global object (select|extend)-(whole|inner)-start n -docstring 'Node' '<esc>:tree-select-prev-node<ret>'
map global object (select|extend)-(whole|inner)-end n -docstring 'Node' '<esc>:tree-select-next-node<ret>'
```

#### Search

Jump to the next occurence while searching:

``` kak
map global prompt (select|extend)-(next|previous)-match <tab> <ret>/<up>
```

Issue: #1663

#### User prompts

Add a mapping to a prompt defined by user:

``` kak
define-command prompt-echo %{
  prompt -signature echo '❯ ' 'echo %val(text)'
}

map global prompt echo <c-n> 'Tchou'
```

[kak-tree]: https://github.com/ul/kak-tree

Issue #1299, opened 2017-03-27
Cannot remap keys in "extend-goto" mode


occivink:
Or whatever the mode triggered by `G` is called.

Also the infobox could probably differentiate both goto and extend-goto.

Delapouite:
> Also the infobox could probably differentiate both goto and extend-goto.

Yes, there could also be a difference between `v` and `V` in the info box. Something mentioning *locking* for `V`.

Delapouite:
More generally there are also no easy ways to make distinct mapping for the various *flavors* of `object` mode: `<a-i>`, `<a-a>`, `[`, `{`, `]`, `}`, `[`, `<a-[>`, `<a-{>`, `<a-]>`, `<a-}>`…

Those are not really distinct mode, it's just the `SelectMode` switching from `Replace` to `Extend`, and `ObjectFlags` being a combination of `ToBegin`, `ToEnd`, `Inner`.

One solution could be to have access to this values in `%val{}` expansions. (i.e. `%val{select_mode}` and `%val{object_flags}`) Then it would be up to the user to implement custom logic in scripts accordingly.

eraserhd:
This was resolved by #2736, right?

Issue #1362, opened 2017-05-12
Rename the `:` object delimiter into `c`


lenormf:
Hi,

Since we support arbitrary punctuation sign delimiters for objects, it's become impractical to easily select a colon-delimited object, because the key for custom objects is a colon.

Renaming the custom object delimiter into `c` (or another letter) would allow using `<a-i> :` to do an inner selection of a colon delimited object, whereas right now we need to use `<a-i> : :,:`.

Thanks.

Delapouite:
Your suggestion makes sense for the colon char.
Another improvement could be to also support the underscore char `_` better, which is heavily present in snake_case languages.
Currently, the the `is_punctuation` is used to trigger the fallback behavior:

https://github.com/mawww/kakoune/blob/master/src/normal.cc#L1149

```cpp
        if (is_punctuation(cp))
        {
            auto utf8cp = to_string(cp);
            return select_and_set_last<mode>(
                context, std::bind(select_surrounding, _1, _2,
                                   utf8cp, utf8cp, count, flags));
        }
```

but the `_` char is not considered a punctuation:

https://github.com/mawww/kakoune/blob/master/src/unicode.hh#L32-L47
```cpp
template<WordType word_type = Word>
inline bool is_word(Codepoint c) noexcept
{
    return c == '_' or iswalnum((wchar_t)c);
}

inline bool is_punctuation(Codepoint c) noexcept
{
    return not (is_word(c) or is_blank(c));
}
```

PR #2736, opened 2019-02-13
Add object command submode

### LINKED ISSUE #2735: No way to access `%val{count}` from an object mode mapping to a command
### Steps 

I've tried the following:
```
map window object e '<esc>: clojure-text-object element %val{count}<ret>'
```

And also accessing `$kak_count` in a shell expansion inside clojure-text-object.

I invoke this with `2<a-a>e`.

### Outcome 

The value is always `0`.

### Expected 

The value should be `2`.

eraserhd:
Closes #2735

In addition, we *could* add additional expansions to tell the command what
kind of object it is (e.g. `<a-a>` vs. `]`), resolving part of #1299 and part
of #2669.

### DIFF
--- a/doc/pages/keys.asciidoc
+++ b/doc/pages/keys.asciidoc
@@ -681,6 +681,10 @@ in order to specify the wanted object:
 *c*::
     select user defined object, will prompt for open and close text
 
+*<a-;>*::
+    run a command in object context.  The expansions `%val{count}` and
+    `%val{register}` are available here.
+
 == Prompt commands
 
 When pressing `:` in normal mode, Kakoune will open a prompt to enter
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -1252,12 +1252,12 @@ void select_object(Context& context, NormalParams params)
                       whole ? "" : (flags & ObjectFlags::ToBegin ? " begin" : " end"));
     };
 
-    const int count = params.count <= 0 ? 0 : params.count - 1;
     on_next_key_with_autoinfo(context, KeymapMode::Object,
-                             [count](Key key, Context& context) {
+                             [params](Key key, Context& context) {
         if (key == Key::Escape)
             return;
 
+        const int count = params.count <= 0 ? 0 : params.count - 1;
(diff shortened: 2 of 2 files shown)


src/normal.cc
```diff
@@ -1254,7 +1254,7 @@ void select_object(Context& context, NormalParams params)
 
     const int count = params.count <= 0 ? 0 : params.count - 1;
     on_next_key_with_autoinfo(context, KeymapMode::Object,
-                             [count](Key key, Context& context) {
+                             [count,params](Key key, Context& context) {
```
mawww: Remove the count capture, and declare the count constant directly inside this lambda.

doc/pages/keys.asciidoc
```diff
@@ -681,6 +681,9 @@ in order to specify the wanted object:
 *c*::
     select user defined object, will prompt for open and close text
 
+*<a-;>*::
+    run a command
```
Delapouite: I find this description a bit too much straight to the point. By reading this, I would think "why would I need to run a command here?"

src/normal.cc
eraserhd: done

doc/pages/keys.asciidoc
eraserhd: updated

eraserhd:
After thinking about it, I changed the key from `<a-;>` to `:`.  I was kind of on the fence about this, so I'm happy to change it back.

My thinking is that `:` can still be remapped to some object, and it won't affect other mappings.  It isn't very likely to be mapped for an object, and it is more directly analogous to `:` than `<a-;>` anyhow, which, even though both normal-mode `:` and insert-mode `<a-;>` push a mode and return to the previous mode in the same way, it still "feels" more to me like `<a-;>` would return to object mode after the command.

Delapouite:
Here are some previous discussions about `:` : https://github.com/mawww/kakoune/issues/1362

eraserhd:
@Delapouite alrighty, reverted back to `<a-;>`

PR #2738, opened 2019-02-14
Add `select_mode` and `object_flags` expansions

eraserhd:
This is built on top of #2736, but is here for separate consideration.

Adds `select_mode` and `object_flags` expansions for the command-line
invoked from the object menu's `<a-;>`.

### DIFF
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -9,6 +9,7 @@
 #include "commands.hh"
 #include "context.hh"
 #include "diff.hh"
+#include "enum.hh"
 #include "face_registry.hh"
 #include "file.hh"
 #include "flags.hh"
@@ -37,6 +38,14 @@ enum class SelectMode
     Append,
 };
 
+constexpr auto enum_desc(Meta::Type<SelectMode>)
+{
+    return make_array<EnumDesc<SelectMode>, 3>({
+        { SelectMode::Replace, "replace" },
+        { SelectMode::Extend, "extend" },
+        { SelectMode::Append, "append" },
+    });
+}
 void merge_selections(Selection& sel, const Selection& new_sel)
 {
     const bool forward = sel.cursor() >= sel.anchor();
@@ -437,7 +446,7 @@ void for_each_codepoint(Context& context, NormalParams)
     selections.insert(strings, InsertMode::Replace);
 }
 
-void command(Context& context, NormalParams params)
+void command(Context& context, EnvVarMap env_vars)
 {
     if (not CommandManager::has_instance())
         throw runtime_error{"commands are not supported"};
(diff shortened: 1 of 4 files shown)


src/normal.cc
```diff
@@ -1311,6 +1316,25 @@ void select_object(Context& context, NormalParams params)
...
+                object_flags = "inner";
+            if (flags & ObjectFlags::ToBegin)
+                object_flags += format("{}to_begin", object_flags == "" ? "" : ",");
+            if (flags & ObjectFlags::ToEnd)
+                object_flags += format("{}to_end", object_flags == "" ? "" : ",");
+            EnvVarMap env_vars = {
+                { "count", to_string(params.count) },
+                { "register", String{&params.reg, 1} },
+                { "select_mode", mode == SelectMode::Extend ? "extend" : "replace" },
+                { "object_flags", object_flags }
```
mawww: I was wondering between exposing those values like you did, or exposing the keystroke that led to them. The keystroke seems simpler, but I suspect in most use cases where you'd use that keystroke, you could just get it with a non-command based key mapping in object mode.

src/normal.cc
```diff
@@ -437,7 +437,7 @@ void for_each_codepoint(Context& context, NormalParams)
     selections.insert(strings, InsertMode::Replace);
 }
 
-void command(Context& context, NormalParams params)
+void command(Context& context, NormalParams params, EnvVarMap env_vars)
```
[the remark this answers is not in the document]
eraserhd: done

src/normal.cc
```diff
@@ -1311,6 +1316,25 @@ void select_object(Context& context, NormalParams params)
             return;
         }
 
+        if (key == alt(';'))
+        {
+            String object_flags;
```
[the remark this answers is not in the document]
eraserhd: done

eraserhd:
Capturing by reference causes an `abort()`, so I assume that `prompt` returns immediately, so I used `std::move()` in the capture.   I think this is right, though my billion years of C++ were before C++11 was finalized.

I added `enum_desc` for `SelectMode` as well, on the reasoning that if `SelectMode` gets new options, we don't want to go searching for all of its uses.

mawww:
Correct, prompt returns immediately (we cannot be blocking, we might have other clients to service), in general all user interaction related function take a callback on what to do once we get the input and return immediately.

src/normal.cc
```diff
@@ -37,6 +38,16 @@ enum class SelectMode
     Append,
 };
 
+constexpr bool with_bit_ops(Meta::Type<SelectMode>) { return false; }
```
mawww: No need for that, false is the default.

src/normal.cc
```diff
@@ -479,16 +490,21 @@ void command(Context& context, NormalParams params)
...
         });
 }
 
+void command(Context& context, NormalParams params)
+{
+    EnvVarMap env_vars = {
+        { "count", to_string(params.count) },
+        { "register", String{&params.reg, 1} }
+    };
+    command(context, env_vars);
```
mawww: either write env_vars directly in the parameter, or std::move it.

src/normal.cc
eraserhd: done

src/normal.cc
eraserhd: Used `std::move`

Issue #2495, opened 2018-10-15
Add hyperlink support to :man


Delapouite:
When using the `:doc` commands, jumping from topic to topic by hitting `<ret>` on a keyword is a very handy way to navigate. The same behavior would be neat for `:man`.

References to other man pages are already highlighted. Also, this PR provides a way to have multiple `*man*` buffers opened at the same time: https://github.com/mawww/kakoune/pull/1927

PR #2746, opened 2019-02-20
Add hyperlink support to man plugin

stevenyvr987:
This is a fix for #2495 (Add hyperlink support to :man).
* The first commit adds code to `man.kak` to enable a user to follow man page links. The new code is guided by the code in `doc.kak` to follow doc links.

  - A hidden command `man-parse-links` is defined to parse the current man page for links.
  - Link details are stored in a new hidden global option `man_links`.
  - A command `man-follow-link` is defined to follow a selected link.

  Further details are described in the `doc.kak` source code.
* The second commit provides a unit test for the new code. The unit test doesn't have to be merged; I used it to convince myself that the new code works well enough. The test strategy is described in `test/man/rc`.

### DIFF
--- a/contrib/kakoune.spec
+++ b/contrib/kakoune.spec
@@ -1,5 +1,5 @@
 Name:           kakoune
-Version:        2018.09.04
+Version:        2019.01.20
 Release:        1%{?dist}
 Summary:        Vim inspired editor
 
@@ -10,6 +10,7 @@ Source0:        %{name}-%{version}.tar.bz2
 BuildRequires:  ncurses-devel >= 5.3
 BuildRequires:  asciidoc
 BuildRequires:  gcc-c++
+BuildRequires:  glibc-langpack-en
 Requires:       ncurses-libs >= 5.3
 
 %description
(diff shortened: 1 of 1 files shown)


lenormf:
Are the `-override` flags necessary?

stevenyvr987:
`-override` is not necessary but helpful; it makes for easier development because i can re-source the plugin after re-editing. Also, why not?

`travis` is failing because of a complaint with `awk` using `--assign` (long version of `-v`); `travis` thinks `--assign` is a file, but it works on my laptop.

Screwtapello:
Only GNU awk supports long options like `--assign`. POSIX [does not mention it](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/awk.html), it's apparently not available in the BSD awk used by macOS, and it's [not available in mawk](https://invisible-island.net/mawk/manpage/mawk.txt), the default awk implementation for many Linux distributions including Debian.

Generally Kakoune tries to stick to basic POSIX features where possible.

stevenyvr987:
@Screwtapello, thanks for the note; I'm not a regular `awk` user. I'll update the PR to use `-v`.

I've also changed the newly-introduced `man-parse-links` command to accept more patterns for man links; links for some man pages, especially for those in section 3 are quite diverse.

mawww:
Sorry for the delay in reviewing that,

I am surprised with the amount of necessary code to achieve that, I dont think it is necessary to pre-parse the buffer as we can just check if we are onto a link immediately when we try to follow it:

extend to previous word start: `<a-?>\b\w<ret>`, extend to next closing parenthesis: `<a-;>f)`, validate its a link `<a-k>\w+\(\w+\)<ret>`, follow it: `:man %val{selection}<ret>`.

`doc.kak` needs to parse links because it removes the markup from the buffer, but we do not need to do that in `man.kak`.

Does that make sense ?

stevenyvr987:
@mawww I just saw your comment; let me think about what you've suggested; it sounds viable.

PR #2756, opened 2019-02-27
Notes on tmux true color and Mac OS

eraserhd:
I needed this to make it work on Mac OS.

@lenormf do you have a screen-256color terminal on your system?  If so, maybe instead of merging this PR, we can just change it?

Also, I added support for italics to my terminfo.  Would that be a useful thing to add here?


Screwtapello:
The `screen-256color` and `tmux-256color` terminfo entries are quite different - `tmux` supports all kinds of things (modifiers on cursor movement keys, modifiers on function keys, italic mode, strike-through) that screen does not, so defaulting to `screen-256color` would be an annoying downgrade for some people.

eraserhd:
OK, I've updated the documentation with a better solution for Mac OS, renamed the old terminfo to `tmux-256color.ncurses61.terminfo`, and added `tmux-256color.ncurses57.terminfo`.  This should supports Mac OS nicely and should be usable on other systems with older ncurses.

doc/pages/faq.asciidoc
```diff
@@ -53,23 +53,47 @@ regular color palette by Kakoune. Fortunately, on recent versions of `tmux`
...
-```
+----
 
 Note: `\*col*` is a regular expression that matches your terminal's `$TERM`
 value, modify it if necessary, e.g. `xterm-termite`.
 
-Next, run the following command to create a local `terminfo` override:
+Next, run the following command to create a local `terminfo` override.  If
+your system has a modern ncurses installed, this will be sufficient (see
+notes below if you have an older ncurses or are runing Mac OS X):
```
Screwtapello: "runing" → "running"

doc/pages/faq.asciidoc
```diff
@@ -53,23 +53,47 @@ regular color palette by Kakoune. Fortunately, on recent versions of `tmux`
...
+=== Mac OS
+
+Mac OS's system ncurses is too old, but Kakoune will automatically use a
+version of ncurses installed with Homebrew, though we need to configure both
+versions of ncurses to support the `tmux-256color` terminal type:
+
+----
+$ brew install ncurses
+$ tic /path/to/kakoune/contrib/tmux-256color.ncurses57.terminfo
+$ /usr/local/opt/ncurses/bin/tic /path/to/kakoune/contrib/tmux-256color.ncurses61.terminfo
```
Screwtapello: Are you sure it's necessary to run both? I worry about a sequence of events like:

- read system-wide `screen-256color` entry
- add modifications
- write `~/.terminfo/t/tmux-256color`
- read `~/.terminfo/t/tmux-256color`
- add modifications that are already there, so effectively no change
- overwrite `~/.terminfo/t/tmux-256color`

If that's how you've configured your system, it may be worth running `/usr/local/opt/ncurses/bin/infocmp tmux-256color` and see if it contains the tmux-specific entries like `kLFT` and `kRIT`.

eraserhd:
OK, I did find out that this doesn't work.  :(

The ncurses 6.1 `tic` does not write to `~/.terminfo`.  It updates the terminal definitions in `/usr/local/opt/ncurses/share`.

The system `tic` writes to `~/.terminfo`.

Both ncurses, however, will read from `~/.terminfo`.

Both kinds of programs need to see the same value for `$TERM`, so it seems the only option (aside from overriding the system ncurses, which doesn't seem like a good idea), is to come up with terminal description that works for both.  Which I've done, and apparently without losing features (although there is some newly supported syntax in the terminfo file which I needed to convert).

So I've pushed a *full* terminal description that works on ncurses 5.7 and ncurses 6.1, has all the features.  The only drawback is that it will need to be updated for new tmux features occasionally.

Screwtapello:
It's worth pointing out that `tic` has the `-o PATH` option to tell it where to write the compiled entry. Manually specifying `tic -o ~/.terminfo` might be easier than keeping a duplicate `tmux-256color` terminfo entry up-to-date.

PR #2762, opened 2019-03-03
unset-option: throw if option is not set in the scope

occivink:
This makes it possible to reset an option (in order to rerun hooks for example), by doing

```
try "
    unset buffer filetype
    set buffer %opt{filetype}
"
```

### DIFF
--- a/src/option_manager.cc
+++ b/src/option_manager.cc
@@ -91,6 +91,8 @@ void OptionManager::unset_option(StringView name)
         if (changed)
             on_option_changed(parent_option);
     }
+    else
+        throw runtime_error{format("option '{}' is not set in this scope", name)};
 }
 
 void OptionManager::on_option_changed(const Option& option)


mawww:
I dont know about this one, it is not clear to me if unset-option would be better if it failed when the value is not set, I would like to see a good motivating use case to add that.

occivink:
I'm not sure if it's such a strong usecase, but in [kakoune-snippets](https://github.com/occivink/kakoune-snippets), it is possible to edit the available snippets "live", and the filetype hooks will be regenerated. For example
```
set buffer filetype kak
# snippets of filetype 'kak' are available
# user modifies snippet files of type 'kak'
snippets-directory-reload
# from now on, the next time filetype=kak is set, the new snippets will be available
# but only old snippets are available on existing filetype=kak 
```
If we had the possibility to reset the filetype, the hooks would be re-run and we'd already have the new snippets in existing buffers.

But regardless it seems relatively in line with other commands to show an error if the operation cannot be carried out meaningfully, no?

mawww:
> But regardless it seems relatively in line with other commands to show an error if the operation cannot be carried out meaningfully, no?

This is what I am wondering about, I am not sure it applies to this case, which seems to me more similar to say `:buffer <name-of-current-buffer>`. I wonder if it really make sense to error on unset-option if it does not have anything to do, it is not clear to me that the command is failing to achieve its purpose in that case.

PR #2767, opened 2019-03-05
Add palenight colorscheme

eraserhd:
Interesting note: Several scripts use hard-coded colors.  For example, git
has hard-coded "red" and "green" when showing a diff.  I've configured my
terminal colors to match, for now, but it seems like we might want to either
allow overriding color names, or require face names in highlighters.

screen shot 2019-03-06 at 9 30 41 am

### DIFF
--- a/colors/palenight.kak
+++ b/colors/palenight.kak
@@ -1,0 +1,96 @@
+# palenight theme
+
+# This was ported from https://github.com/drewtempelmeyer/palenight.vim
+
+evaluate-commands %sh{
+    red=rgb:ff5370
+    light_red=rgb:ff869a
+    dark_red=rgb:be5046
+    green=rgb:c3e88d
+    yellow=rgb:ffcb6b
+    dark_yellow=rgb:f78c6c
+    blue=rgb:82b1ff
+    purple=rgb:c792ea
+    cyan=rgb:89ddff
+    white=rgb:bfc7d5
+    black=rgb:292d3e
+    comment_grey=rgb:697098
+    gutter_fg_grey=rgb:4b5263
+    cursor_grey=rgb:2c323c
+    visual_grey=rgb:3e4452
+    menu_grey=rgb:3e4452
+    special_grey=rgb:3b4048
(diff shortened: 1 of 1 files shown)


lenormf:
#1560 

Do you have a screenshot?

colors/palenight.kak
```diff
@@ -0,0 +1,96 @@
+# palenight theme
+
+# This was ported from https://github.com/drewtempelmeyer/palenight.vim
```
mawww: How was this ported ? I am worried this means that this palenight.kak file is now a derivative work, and hence cannot be published inside Kakoune codebase as a UNLICENSED file.

lenormf:
I don't see any screenshot :(

eraserhd:
@lenormf oops, forgot to hit save.  It's there now.

colors/palenight.kak
eraserhd: Hmm, I didn't think about that.  My exact process:

* I copied desertex.kak to palenight.kak
* Extracted the color numbers from https://github.com/drewtempelmeyer/palenight.vim/blob/master/autoload/palenight.vim using  Kakoune multiple cursors, and pasted them into palenight.kak as a comment.
* Updated various face commands to use rgb:xxxx, usually retyping the color number from the comment.  Read various bits of palenight.kak for reference, but not copied.
* Decided that was dumb, and reformatted the comment into shell variables.

So, everything aside from the colors would be a derivative of `desertex.kak`.  If you think that's too much of a risk, I'll split this out to another repo and make it MIT.  Or we could just make palenight.kak MIT with a copyright header and attribution in comments.

lenormf: That's how most of the builtin colorschemes were written I reckon, mines definitely so.

mawww: I am not a lawyer, but seems good enough to me, if you started with palenight.vim and progressively massaged it into palenight.kak I would have been a bit more worried, but here seems you used palenight.vim to read the color values, which as far as I know is not something you can claim copyright on.

PR #2768, opened 2019-03-05
Various fixes to iterm.kak

occivink:
The 'terminal' alias wasn't automatically setup
Missing `\` in sed calls
Descriptions of vertical and horizontal split was swapped around

Thanks @grtlr for testing

### DIFF
--- a/rc/extra/iterm.kak
+++ b/rc/extra/iterm.kak
@@ -1,12 +1,13 @@
 # https://www.iterm2.com
 # ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
 
-## The default behaviour for the `new` command is to open an vertical pane in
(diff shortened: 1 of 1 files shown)


grtlr:
Thanks @occivink for the quick fix!

I just stumbled on another quirk: the `termcmd` option is not set for iTerm. This prevents plugins such as `fzf.kak` to open a new buffer, for example for file selection. Is there a way to tackle this as well?

grtlr:
`tmux.kak` and `kitty.kak` both define a `*-repl` command. Now that we have the necessary infrastructure for splitting in iTerm, would it be possible to add such a command in `iterm.kak` too?

occivink:
This is mostly up to iterm, and whether it supports sending text to another pane/window. You'll find that the repl (tmux/kitty) stuff is not standardized (nor is there a good reason to), so you're mostly free to implement it however you see fit

PR #2774, opened 2019-03-07
Refactor colorize function away from the test script

alexherbo2:


mawww:
Does not work here, it seems the change relies on `echo` expanding escape sequences, which is not portable, those echo calls should be replaced with printfs.

alexherbo2:
Done

test/run
```diff
@@ -50,7 +50,7 @@ main() {
       echo "$indent$name"
       continue
     elif [ -x enabled ] && ! ./enabled; then
-      echo "$indent$name (disabled)" | colorize $yellow $none
+      printf "${yellow}$indent$name (disabled)${none}\n"
```
mawww: `$name` could be pretty arbitrary as it comes from the filesystem, so we should probably not put it in the `printf` format string.

alexherbo2:
Done

Issue #2772, opened 2019-03-07
[Proposal] Provides-Requires Lazy Loading Model


laelath:
I wanted to get feedback on this lazy-loading idea before attempting to implementing it.

## Requirements given by mawww

* No central place that knows of all filetype/scripts
* One way dependencies (sh.kak should not know that kak.kak depends on it)
* Simple file layout and behaviour, easy to work without a plugin manager
* Fast

## Provides-Requires Model

This model handles lazy-loading by adding two new commands, `provides` and
`requires`. The `provides` command specificies a module that is being provided,
and the commands for that module. The `requires` command looks up the modules
that have been specified by `provides`, runs the commands for the given module,
and marks it as loaded. If the module has already been loaded, it returns
immediately. It would fail if the requested module has not been provided. This
method would allow for overriding default modules with a `-override` switch for
the `provides` command, which would fail if the module has already been loaded.

### Pros
 * Follows kakoune's current file-less structure (excepting sourcing)
 * Easy to override specific highlighters in kakrc

### Cons
 * Needs the entire environment to be loaded before `requires` are guaranteed to work

### Example

```kak
## sh.kak
hook global BufCreate .*\.(z|ba|c|k|mk)?sh(rc|_profile)? %{
    requires sh
    set-option buffer filetype sh
}

provides sh %{
    # highlighters, etc for sh
}



## kak.kak
hook global BufCreate (.*/)?(kakrc|.*.kak) %{
    requires sh
    requires kak
    set-option buffer filetype kak
}

provides kak %{
    # highlighters, etc for kak
}



## kakrc
provides -override sh %{
    # Overriding sh
}
```

mawww:
I think this is a promising direction, its simple, fast, and very flexible, in particular I like the fact that it opens the door towards vim-like autoloading with:

```
provides sh %{ source "%val{runtime}/rc/sh.kak" }
```

(I dont think we should do that right now, but in case some scripts become so big that reading and keeping them in memory is deemed too slow, we can go this way).

It shares the nice property @maximbaz's solution had that its just a few lines to add to each .kak files (just wrap the specific parts into `provides`).

Another thing I was wondering is if we can provide fallback: Say we have `kakrc.kak` in our autoload but not `sh.kak`, it would be nice to have that work out of the box. That means `kakrc.kak` should be able to specify that `sh` can be provided by sourcing `%val{runtime}/rc/core/sh.kak`.

Something that will work without extra features would be:
```
[...]

PR #2782, opened 2019-03-13
Provides requires

laelath:
Implemented `provide-module` and `require-module` commands from #2772.
Also added a `ModuleLoad` hook which is run when a module is required for the first time. This is needed for setting options that are no longer immediately declared, such as `termcmd`.

I'm seeing an ~2x speed up in load time on my laptop and ~3x speedup on a raspberry pi.

As this required editing a lot of files, there's a high likelihood that I've broken some functionality, but everything I've tested except for markdown/restructured text works, which don't highlight code sections unless you manually load the relevant module.


Screwtapello:
Should these new commands be added to `doc/pages/commands.asciidoc`, and the new hook to `hooks.asciidoc`? Probably also something in `changelog.asciidoc`.

mawww:
Thanks for working on this,

I was initially suprised by the addional hook on BufSetOption, but I understand this makes the module work for files where we manually set the filetype.

I was trying to understand how this managed to work, as I expected the following problem:

1. User/hook sets filetype to cpp
2. The added hooks `requires-module cpp`
3. Somehow magically, hooks added by the cpp module are run as well for this filetype change.

This 3. was surprising to me, because Kakoune is supposed not to execute those newly added hooks until the hook itself gets triggered again. And yet, it does seems to work...

After a bit of debugging, I understood what was going on: The require-module command is triggered by `BufSetOption` while most filetype support hooks added by that module are triggered by `WinSetOption` which happens to run afterwards, making all that work.

I dont think this is good enough, it works when we load based on filetype as long as modules do add their hooks at window scope, but if we had a module adding BufSetOption hooks, I think it would be reasonable to expect that to work.

So we either need a way to re-trigger the BufSetOption hook after running `require-module`, or modify the design so that it works by construction. One solution could be to change how Kakoune handle hooks to guarantee that if we add new hook commands for the currently executing hook, they will be executed.

laelath:
Yeah, I originally didn't plan to create the extra `BufSetOption` hook, but when implementing it I thought that we would want to maintain the ability to manually set the filetype without also requiring a manual `require-module`.

As for the hook situation, I agree that the way I implemented it is not ideal. I'm not sure about guaranteeing that new hooks are run for the currently executing hooks. It seems like it could be really powerful, but also easily infinitely loop (though maybe not really in practice?). The best way to change the design that I can think of would be to run `unset-option buffer filetype`, `set-option buffer filetype <name>` after the `requires-module <name`.

doc/pages/changelog.asciidoc
```diff
@@ -3,13 +3,22 @@
 This changelog contains major and/or breaking changes to Kakoune between
 released versions.
 
+== Development version
+
+* Introduced a module system using the `provides-module` and
+  `requires-module` commands that allows for lazily loading language
```
Screwtapello: I think the commands are `provide-module` and `require-module` (without the S).

doc/pages/changelog.asciidoc
```diff
@@ -3,13 +3,22 @@
...
+  support files with dependency resolution.
+
+* Added a new hook `ModuleLoad` which is run when a module is loaded, as
+  now options needed by modules aren't defined until the module is loaded.
```
Screwtapello: It's not immediately clear why options are relevant here; how about:

> Added a new hook `ModuleLoad` which is run when module is loaded, so one module can integrate with another if it gets loaded, without adding a hard dependency.

doc/pages/commands.asciidoc
```diff
@@ -299,6 +299,15 @@ but not really useful in that context.
 *debug* {info,buffers,options,memory,shared-strings,profile-hash-maps,faces,mappings}::
     print some debug information in the *\*debug** buffer
 
+== Module commands
+
+*provide-module* [<switches>] <name> <commands>::
+    declares a new module that is provided by the given commands. The commands
+    passed into the module are guaranteed to be run only once.
```
Screwtapello: This would be the primary documentation for this feature, so it's worth going into a bit of detail. How about:

> Declares that the module <name> is defined by <commands>. <commands> will be evaluated as if by `evaluate-commands` the first time `require-module <name>` is run.

Actually, *is* it evaluated as if by `evaluate-commands`? Does it execute in a draft context? Does it execute in a particular client context? Does it have access to buffer or window scopes, or does that depend on where the `require-module` command is executed? If so, should it be restricted to global scope, so that we don't get weird bugs where behaviour depends on which `require-module` statement executes first?

doc/pages/commands.asciidoc
```diff
@@ -299,6 +299,15 @@ but not really useful in that context.
...
+
+*require-module* <name>::
+    guarantees the <name> module is loaded before continuing command execution
```
Screwtapello: "loading" a module isn't really a concept mentioned in the `provide-module` docs; how about:

> guarantees the commands associated with `provide-module <name>` have been evaluated before continuing command execution. Fails if no `provide-module <name>` command has been evaluated so far.

doc/pages/commands.asciidoc
Screwtapello: Also, this command fails if the same module is defined twice, yes? That should be mentioned too.

doc/pages/hooks.asciidoc
```diff
@@ -173,6 +173,9 @@ name. Hooks with no description will always use an empty string.
 *RawKey* `key`::
     Triggered whenever a key is pressed by the user
 
+*ModuleLoad* `module`::
+    Triggered when a module is loaded
```
Screwtapello: Again, "loading" isn't well-defined for Kakoune; how about:

> Triggered when a `require-module` command evaluates a module's commands.

occivink:
>One solution could be to change how Kakoune handle hooks to guarantee that if we add new hook commands for the currently executing hook, they will be executed.

This was [changed recently](https://github.com/mawww/kakoune/pull/2603), I don't think we absolutely need to keep the current behavior but we have this usecase to think about as well.

laelath:
After thinking about it, I'm not too sure the `WinSetOption`/`BufSetOption` is that big of an issue. Filetype modules are responsible for defining how they're loaded, so if a module used `BufSetOption` hooks to setup, that module would be responsible for making sure those hooks are run after the first load.

That being said, this solution is still kinda messy, and I would be happy to implement a more robust solution.

mawww:
In my view it is problematic because module `kakrc` will require module `sh`, so `sh` cannot assume it is being loaded by a `BufSetOption filetype=sh` hook and hence would be wrong if it started to enable `sh` highlighting and indentation on that buffer.

The current implementation works by chance, because most filetype support hook on WinSetOption which gets triggered after BufSetOption. If a filetype was, as it was suggested a few times, to be added at buffer scope (so using BufSetOption), the module pattern that works in all other filetypes would break on this filetype, for reasons that are far from obvious...

One solution would be to change the behaviour of the `hook` command itself to eagerly execute the hooks commands if we are already running that hook. Might be surprising though as it means sometimes `hook ...` would work as `evaluate-commands`.

laelath:
A lot of hooks use the pattern:
```
hook -group kak-highlight global WinSetOption filetype=kak %{
    add-highlighter window/kakrc ref kakrc
    hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/kakrc }
}
```
Which would be broken by always eagerly executing the hooks.

Maybe it could be a switch in the `hook` command? Something like `hook -eager ...` it would reduce the amount of unexpected behavior at the cost of an esoteric switch. But this would mean the modules would have to anticipate that they could be loaded from those hooks.

Another option would be to require loading the module before setting filetype, since the module would be loaded before the execution of any of these hooks, like how I wrote it in the proposal. However this could be confusing to people who aren't aware of modules, which ideally would be invisible to users.

Finally, we could pull the initialization hooks out of the module, and have them require their own module, i.e.:
```
hook -group kak-highlight global WinSetOption filetype=kak %{
    require-module kak
    ...
}

hook global WinSetOption filetype=kak %~
    require-module kak
    ...
~
```
Thoughts?

Screwtapello:
I like the idiom of having all the hooks outside the module. As I understand it, hook statements generally aren't what slows down Kakoune startup, and I think it's easier to understand and reason about a system where all the hooks are set up at the start, instead of some of them appearing part-way through.

laelath:
Turns out many highlighters use a shell script to create a hook that adds the language's keywords to the `static_words` list, so moving the hooks outside the modules won't work. Having shell scripts outside of the modules would defeat one of the main goals of a module system.

laelath:
I've come up with this potential pattern. After experimenting I've found that changing options inside of `*SetOption` hooks results in unexpected behavior.
```
hook -once global BufSetOption filetype=(c|cpp|objc) %{
    require-module c-family
    hook -once buffer NormalIdle '' %sh{
        printf "
            unset-option buffer filetype
            set-option buffer filetype %s
        " $kak_hook_param_capture_1
    }
}
```

mawww:
@laelath The `static_words` problem can be fixed another way, through I feature I have wanted to tackle for while: user scopes.

The idea is to introduce a way to create named scopes (say a `cpp` scope), add all the aliases, hooks, option values, highlighters that are specific to those directly inside that scope, and provide a way to add those scopes to the built-in ones:

`enable-scope window cpp` would add the `cpp` scope as a "delegate scope" of `window`, and would act as an additional "parent" of the window scope (so that say `options` in the window scope do override the value they might have in the `cpp` scope).

With that, the `cpp` module would mostly consist of setting up the `cpp` scope, and there would likely only be one hook that on filetype set to `cpp` would enable the `cpp` scope in the current window. The `static_words` would hence only be set once, eagerly during the `cpp` module execution, and there should be no need for hooks inside modules.

Of course that means we still have to implement all that... And it also means we are introducing two brand new concepts into Kakoune, which adds a bit of complexity.

As an aside, I have been wondering if we can get the module system implemented directly as a pattern using hooks (potentially adding resonable features to the hook system) as they are pretty close to each other, but did not find any satisfying design so far.

laelath:
Something similar to the pattern I was doing here would be to (ab)use the command overriding to make a command that only meaningfully runs once.
```
define-command load-c-family %{
    define-command -override load-c-family ""

    # c-family commands
}
```
Though this is basically just a replacement for provides/requires, and misses some nice things like erroring if you try and override a module that's already been loaded.

Named scopes sound really interesting! It sounds like a lot of the use cases of modules would be very related to named scopes. Maybe something like
```
# detection hooks

hook -once global ScopeEnable cpp %{
    # indention hooks, aliases, keywords
    require-module c-family
}

provide-module c-family %{
    # highlighters, commands
}
```

mawww:
As a temporary stop-gap for `static_words`, we could make support script introduce their own `cpp_static_words` that gets filled at require time with the correct list of words, then have the hook just
do `set buffer static_words %opt{cpp_static_words}` after `require`.

This way we should have a robust solution to all the problems we identified so far, and we could go ahead with provide/requires support.

laelath:
@mawww I've finished updating all the files. If it looks good I can create a squash commit for merging.

laelath:
We're out of strictly ascii characters, but we could use some characters from the Latin-1 block, like §, or ¶.

maximbaz:
Nice job everyone 🎉 

I have one question on the usage, previously I had this line in `kakrc`:

```
set-option global kitty_window_type 'os'
```

In the current `master` it fails with `'set-option' option not found: 'kitty_window_type'. Use declare-option first`.

Adding `require-module kitty` before that line helps, but is that the intended usage pattern to require some stuff in kakrc, or there's another approach I'm missing?

laelath:
I put the windowing support into modules, so those options aren't defined until the module is first loaded, which is usually after the user's kakrc is loaded. The intended approach would be
```
hook ModuleLoad kitty %{
    set-option global kitty_window_type 'os'
}
```
The `ModuleLoad` hook is run when the module is first sourced by a `require-module`. Ideally, `require-module` should only be used when implementing support files and invisible to users. (except that hook)

PR #2784, opened 2019-03-14
rc: Use builtin faces

eraserhd:
I took about five commits from PR #1560, rebased and squashed them.  I aimed
for the pieces that seemed "uncontroversial" to me.  I think I may have made
a minor change, but I kept @lenormf's authorship on the commit.

I didn't check to ensure this was _complete_.  New uses of non-semantic faces
may have been introduced since the PR was made.

(I like the diff stuff, but I think the face names warrant being defined in
`colors/default.kak`.)


mawww:
I will merge that, but I think it might make sense at some point to introduce a separate "disabled" face that would be separate from comment.

PR #2785, opened 2019-03-15
Highlight every Ruby symbol, global variable, class variable, and instance variable correctly

alyssais:
`:_foo`, `@_foo`, `$_foo`, and `{ _foo: 1 }` are all valid Ruby syntax, but were not recognized by the highlighter.


alyssais:
The second commit makes Kakoune highlight every (AFAIK – I checked with `ruby -e puts 'Symbol.all_symbols.map(&:inspect)' | kak`) symbol, global, class, and instance variable correctly, including all the weird special cases like `$/`. It doesn't seem to affect anything else – other Ruby syntax is still highlighted as normal.

It also fixes `=begin/=end` comments.

PR #2797, opened 2019-03-22
call find in BSD compatible way

vbauerster:
`make install` fails on macOS, because of `find`.
```
find -type f -exec chmod 0644 {} +
find: illegal option -- t
usage: find [-H | -L | -P] [-EXdsx] [-f path] path ... [expression]
       find [-H | -L | -P] [-EXdsx] -f path [path ...] [expression]
```
This PR fixes this issue.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -140,7 +140,7 @@ dist:
 
 distclean: clean
 	rm -f kak kak$(suffix)
-	find . ../doc -type f \( -name \*\\.gz -o -name \*\\.1 \) -exec rm -f '{}' +
+	find ../doc -type f \( -name \*\\.gz -o -name \*\\.1 \) -exec rm -f '{}' +
 
 installdirs:
 	install -d $(bindir) \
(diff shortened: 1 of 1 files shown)


src/Makefile
```diff
@@ -140,7 +140,7 @@ dist:
 
 distclean: clean
 	rm -f kak kak$(suffix)
-	find ../doc -type f \( -name \*\\.gz -o -name \*\\.1 \) -exec rm -f '{}' +
+	find . ../doc -type f \( -name \*\\.gz -o -name \*\\.1 \) -exec rm -f '{}' +
```
casimir: This changes the meaning of the command, it will search from the current directory in addition to the doc directory which is not the intention here.

vbauerster: You're right, thanks. I've pushed the update to undo this.

casimir: 👍

laelath:
Oh that's my bad, the call should actually be `find $(sharedir)/rc -type f -exec chmod 0644 {} +`

vbauerster:
@laelath updated

PR #2800, opened 2019-03-22
Add connect.kak

alexherbo2:
Continuing from #1450 and #2799

You can now `connect ranger` to edit a file from [ranger] where the [connect][connect.kak] command was executed.

[connect.kak]: https://github.com/alexherbo2/connect.kak
[ranger]: https://ranger.github.io


mawww:
While I think connect is a real nice tool, I dont think it should be part of the bundled scripts, and if I was to merge it, it would not be as a submodule anyway.

I think connect.kak should remain an external plugin, I believe it is very easy to opt-in for users.

Issue #2801, opened 2019-03-22
Add rustfmt to filetype


pickfire:
### Steps 

Need to manually add to config from what is shown in wiki https://github.com/mawww/kakoune/wiki/Languages.
```
hook global WinSetOption filetype=rust %{
  set window formatcmd 'rustfmt'
}
```

### Outcome 

Nothing but need to manually add `formatcmd`

### Expected 

Rust filetype are expected to use `rustfmt` by default. Possible to add it to https://github.com/mawww/kakoune?

Delapouite:
Same could be done for `gofmt` I think. These 2 tools are standards in their respective ecosystem.

andreyorst:
I think that `c-family` can test if either `clang-format` or another supported formatter is installed and set `formatcmd` accordingly.

maximbaz:
@Delapouite not sure about `gofmt` being standard, there's also `goimports` 😉

mawww:
This is now fixed.

pickfire:
@Delapouite @andreyorst I had already added for rust, feel free to add one for go and c. (I need to fix the wiki as well)

PR #2802, opened 2019-03-22
Default to rustfmt for formatcmd in rust

pickfire:
`rustfmt` is the de facto standard for formatting in rust so setting it by default would be nice.

https://github.com/mawww/kakoune/issues/2801


occivink:
You'll want to put this line inside the `hook global WinSetOption filetype=rust %[` block, otherwise it will get executed at initialization when the script is sourced (at which point there is no window). 
You should also put it inside a `try %{ .. }` block, since it might fail if `format.kak` was not sourced for some reason (and therefore formatcmd not defined).

pickfire:
@occivink Thanks a lot for noticing that issue.

PR #2803, opened 2019-03-23
doc: The faces doc should list all the supported color names.

Screwtapello:


andreyorst:
I didn't even knew those are available.

Delapouite:
I wonder what a "bright-black" could look like. A "dark-white" maybe?

Screwtapello:
It sounds weird, but it's the natural result of adding eight "bright" colours to the original eight regular colours. The exact shades depend on what palette you're using, but with the traditional Windows 16 colour palette:

- black: 0% brightness
- white: 75% brightness
- bright black: 50% brightness
- bright white: 100% brightness

Issue #2790, opened 2019-03-17
Better syntax highlighting for rust code


nikhilmitrax:
PROBLEM:

The syntax highlighting in kakoune for rust seems to be lacking some highlighters.

Attaching screenshot of Kakoune and Vim running the `gruvbox` theme showing rust code from alacritty.

There seems to be a lot more highlighting on the vim side, especially around module imports and function calls.

The syntax highlighting I'm using for vim : [https://github.com/rust-lang/rust.vim/blob/master/syntax/rust.vim](https://github.com/rust-lang/rust.vim/blob/master/syntax/rust.vim)

I apologize for just creating this issue instead of going in myself to fix it, but I tried doing that and quickly realised I'd need some amount of help understanding highlighters in kakoune, which may not be worth the time of a contributor here.

EDIT :

I started a fork with at least one improvement that adds highlighting for the syntax methods.
[https://github.com/nikhilmitrax/kakoune/commit/86be04ffc7f596bb8a687a9bcd49650f906bb382](https://github.com/nikhilmitrax/kakoune/commit/86be04ffc7f596bb8a687a9bcd49650f906bb382)

<img width="1552" alt="Screen Shot 2019-03-17 at 3 28 54 PM" src="https://user-images.githubusercontent.com/3902132/54498889-8d6f0380-48c9-11e9-8a21-2256622d6c23.png">

Screwtapello:
The Rust syntax highlighter is less *detailed* than the Vim one, but different people have different opinions about how much colour is helpful and how much is distracting. I'm sure @mawww would take PRs for any improvements you wanted to submit, but probably nobody is going to add extra things just for feature-parity with Vim.

If you'd like help understanding highlighters, feel free to ask questions on the forum or the IRC channel (both linked under "Community" at the top of [the official homepage](https://kakoune.org/)).

andreyorst:
I've added custom highlighting in my [kakrc](https://github.com/andreyorst/dotfiles/blob/30c1708e30754f41ed78fd61e8ef447bb3228af5/.config/kak/languages.kak#L52-L69). Now it looks like so:
![image](https://user-images.githubusercontent.com/19470159/54532107-06d52780-4998-11e9-98c2-3c54ba344cdc.png)

nikhilmitrax:
@Screwtapello 
Thanks! Yes I think I have a better understanding of how syntax highlighters work.

[...]

PR #2804, opened 2019-03-23
add namespace and function highlighting to rust and c-family

andreyorst:
Inspired by #2790


andreyorst:
These constructions are highlighted:
```rust
function() // as function face
something.field // as meta face
something.method() // as function face
something->field // as meta face
something->method() // as function face
something::something // as module face
```

andreyorst:
I'm taking this down in favor of separate requests, like #3027

Issue #2712, opened 2019-01-29
Aliases should be available in prompt mode only


andreyorst:
Aliases are needed to simplify writing commands in prompt mode, where horisontal space is limited by screen width and vertical space is limited by single line. Aliases also make writing commands in prompt mode a lot faster, because we don't have to write long names like `define-command`. 

In scripts however, we don't limited with single line, and can have really long lines just fine.  So the only meaning of aliases there is to write shorter commands. But
- this is error prone
- aliases doesn't have syntax highliting
- aliases can be misunderstood in shell expansions when reading, like `set` in shell and `set-option` aliased to `set`.

And user can break all scripts that use aliases by simply changing or removing any alias.
For example I've broken my whole setup just by placing `unalias global def` at the beginning of kakrc. I also use plugin [edit-or-dir](https://github.com/TeddyDD/kakoune-edit-or-dir) by @TeddyDD, and I've aliased it to `e`, to call ot from command line. What will happen if some plugins decide to call `e` instead of `edit` to open files? Hopefully they won't break because `edit-or-dir` works fine, but that's doesn't guarantee anything.

In my opinion aliases should be available only in prompt mode, and nowhere else. Vim also have aliases (abbreviations), but the are not allowed to use inside vimscript, and everyone is fine with it. Sure this change will break hell lot of plugins, but it is better to do it now, and don't have problems later.

Screwtapello:
Aliases are also a way for users to configure how things work. For example, `terminal` is an alias for `x11-terminal`, `tmux-terminal-window`, or `screen-terminal-window` depending on what environment Kakoune is running in. It's specifically an alias so that user configuration can override what it does, and have all their terminal-using plugins automatically respect that decision without having to update them individually.

On the other hand `define-command` has an `-override` switch that could be used for the same thing, and there's precedent in that `execute-keys` does not respect mappings.

On the other, other hand, since `define-command -override` is a thing, preventing scripts from using aliases doesn't make them much less fragile, since a user could override important commands like `terminal` anyway.

andreyorst:
my main complain is more about Kakoune's aliases, that are not commands on the first place, like:
```sh
def # python has def for functions
set # sh has set
rmhl
addhl
eval # sh has eval command
```

I don't see why they are either mutable, or even exist

andreyorst:
[...]

PR #2805, opened 2019-03-24
kakrc: Add aliases for scripting

alexherbo2:


Delapouite:
I disagree about this additions, for reasons detailed by @andreyorst in https://github.com/mawww/kakoune/issues/2712 Alias should be used interactively, not for writing scripts.

mawww:
Yep, I dont think should have those aliases in the completion words, because aliases are expected for interactive use, and also because they are short, completion is not very helpfull for typing `cd` or `def`.

PR #2810, opened 2019-03-25
doc: add precision about -buffer switch on eval/exec

Delapouite:
Hello

Relevant filter in the code:
```
if (*bufnames == "*")
{
    for (auto&& buffer : BufferManager::instance()
                       | transform(&std::unique_ptr<Buffer>::get)
                       | filter([](Buffer* buf) { return not (buf->flags() & Buffer::Flags::Debug); })
                       | gather<Vector<SafePtr<Buffer>>>()) // gather as we might be mutating the buffer list in the loop.
        context_wrap_for_buffer(*buffer);
}
```

### DIFF
--- a/doc/pages/execeval.asciidoc
+++ b/doc/pages/execeval.asciidoc
@@ -42,7 +42,8 @@ when the commands have been executed: */*, *"*, *|*, *^*, *@*.
 
 *-buffer* <names>::
     execute in the context of each buffers in the comma separated list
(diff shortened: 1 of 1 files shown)


Screwtapello:
Maybe add a link to the "Debug buffers" section in `buffers.asciidoc` for people who don't already know what a debug buffer is?

Delapouite:
Absolutely. Done.

Issue #2168, opened 2018-06-29
Autoindentation isn't always done with tabs only with indentwidth set to 0


Laaas:
### Example 1
Attempt to write the following in a .c file:
```c
if (true &&
        false) {
    ...
```
### Example 2
```c
func(a, b, c,
    d, e, f);
```

### Outcome
The autoindentation will use spaces instead of tabs, so the first level of indentation on the second line of the first example will be spaces, while the second level will use one tab.

In the second example when you attempt to write it, kakoune will align the beginning of the second line with the first parameter of the first line with spaces, when it should not align them at all when using tabs.

### Expected

All indentation should be done using tabs only when indentwidth is 0, and when spaces are used to align two lines, the lines should not be aligned at all instead, since that is not possible with tabs (you can not assume the width of a tab).

Laaas:
It is a bug IMO that & assumes the amount of spaces to use for each tab, since that can easily change from one user to another.

When attempting to align across lines with different indentation while using tabs, kakoune should err. You could perhaps have a lenient mode that makes kakoune instead make a best guess, that makes it either the same indentation level as the other line or one indentation level above it.

This fits the indent with tabs align with spaces behavior, since you should not align with tabs or align based on indentation.

Laaas:
Furthermore, when & creates indentation, it should always consider indentwidth, regardless of aligntab.

Laaas:
@mawww do you think making a new key for this kind of indentation is a good idea?

It would only support a single column, since it's for indentation.
If using spaces, the key should act just as & would act.
If using tabs, it would indent the selections underneath the highest selection so that they are one indentation level higher than the highest selection.

I am not convinced that this is the best behavior myself, but I want something which I can just drop in instead of &, where it's used for indentation.

mawww:
I dont think we require a key for that, combining `<a-&>` which copies indentation and `>` should work.

Laaas:
But how would it switch between the two styles of indentation? The one where you align to the first parenthesis of the block and the one where you just increase indentation. Your method would hard code it to the second and I don't see how c-family.kak would switch between the two styles without an extra key/command.

PR #2811, opened 2019-03-27
Preserve tabs for autoindent

GrantMoyer:
Related to Issue #2168.

Addresses two problems in c-family autoindentation:

1. Preserves tabs during auto indentation and alignment with '(' followed by letters

    New behavior (`>...` represents a tab character):

    ```
    >...if (true &&
    >...    false) {
    ```

    Old behavior:

    ```
    >...if (true &&
            false) {
    ```

2. Correctly aligns a closing brace to the block following a case like in 1.

    New behavior:

    ```
        if (true &&
            false) {

        }
    ```

    Old behavior:

    ```
        if (true &&
            false) {

            }
    ```


lenormf:
Thanks, that's useful. Could you please squash the commits?

GrantMoyer:
@lenormf I thought GitHub could automatically squash the commits when merging a pull request. If that's not the case, I can squash them. I should squash the copyright waiver too, correct?

lenormf:
Keep the waiver commit as it is now, and squash all the other commits.

Issue #2036, opened 2018-05-08
While saving a file, the inode number doesn't change


lord-re:
### Steps
launch **watch -d stat you-file**

Open this file with kak and save it with :w 

### Outcome 
The *Modify* and *Change* are updated but the *inode number* doesn't change

### Expected 
While doing the exact same thing with vim, the inode number is changed so external programs see the file has changed (for example *hugo server --navigateToChanged*)

Actually i don't think it's really a bug as discussed on irc, it's just a different way of doing things. But this way can be quite cumbersome while using other software which watch the edits only relying on the inode number.

lenormf:
The logs of the conversation about this issue on IRC:

```
2018-04-19 13:14:44     Screwtape       Lord: If you can confirm that backupcopy=yes makes Vim exactly as broken as Kakoune with respect to Hugo, then we'll know exactly what Kakoune has to change to behave like Vim.
2018-04-19 13:14:56     martanne        I actually recently looked into how a file should ideally be saved and unless I'm missing something all options have drawbacks (e.g. the "correct" approach on Linux requires a kernel patch) ...
2018-04-19 13:15:27     Screwtape       The documentation for Vim's 'backupcopy' option makes it pretty clear that yeah, everything is terrible.
2018-04-19 13:16:42     Screwtape       As just one data-point: some people want their editor to break hard-links when saving, as a cheap form of copy-on-write. Other people want their editor to keep hard-links, in order to keep all "copies" of a file in sync.
2018-04-19 13:18:33     ironzorg        martanne: what patch ?
2018-04-19 13:21:57     martanne        Screwtape: well that is just a question of semantics. But there are problems which apply even to regular files.
2018-04-19 13:22:16     martanne        ironzorg: it adds new flags to linkat(2)
```

eraserhd:
Leaving this here for informational purposes: Editing long-running shell scripts in Kakoune makes them fail with weird errors.

dmerejkowsky:
I've just tried this:

My vim *does not* change the inode number, but hugo sees the changes made by vim.

Using the `watch` trick from @lord-re I found that the *access time* is not changed by kakoune, but is changed by vim. 

So maybe there's something something to fix in kakoune after all?

Screwtapello:
> Using the watch trick from @lord-re I found that the access time is not changed by kakoune, but is changed by vim.

[...]

PR #2812, opened 2019-03-27
write_buffer_to_fd(): write the whole file at once

nonumeros:
As mentioned here on https://github.com/mawww/kakoune/issues/2036

commit [4dae2c8](https://github.com/mawww/kakoune/commit/4dae2c875bf70d87ace53d7b03c4a9289b2bb241) 

did not fix the prevalent problem of a delay that occurs while writing the file and the use of static site generators, e.g., hugo, jekyll, etc

### DIFF
--- a/src/file.cc
+++ b/src/file.cc
@@ -270,14 +270,10 @@ void write_buffer_to_fd(Buffer& buffer, int fd, bool sync)
         if (::write(fd, "\xEF\xBB\xBF", 3) < 0)
             throw runtime_error(format("unable to write data to the buffer (fd: {}; errno: {})", fd, ::strerror(errno)));
 
-    for (LineCount i = 0; i < buffer.line_count(); ++i)
-    {
-        // end of lines are written according to eolformat but always
-        // stored as \n
-        StringView linedata = buffer[i];
-        write(fd, linedata.substr(0, linedata.length()-1));
-        write(fd, eoldata);
-    }
+    BufferCoord start { 0, 0 };
+    auto end = buffer.end_coord();
+    String contents = buffer.string(start, end);
+    write(fd, contents);
 
     if (sync)
         ::fsync(fd);


mawww:
Hello,

This will break writing files that use non `\n` end of lines. It also requires to allocate a huge buffer to write the buffer data. I would be more interested in some investigation on why using the `set buffer writemethod replace` does not work with Hugo.

nonumeros:
Hello @mawww 
> why using the set buffer writemethod replace does not work with Hugo
> 

I couldn't follow the code but if `eoldata` on file.cc already defined `\r` and `\n` why does it need to go around with the `\n` again. Therein may lie the problem, or so I naively think. I can provide you with some feedback about hugo, although there's not much to it, other than it doesn't detect any changes afterwards - even though there were changes from kak's side. What's even more weird at this point, is that as soon as it's nearing line one hundredth of the file (plain text only) hugo stops detecting the changes. 

Perhaps it's mere coincidence that this happennig, but I could see hugo stopping detecting any changes.
:)

nonumeros:
So to sum it up, anything less of (give or take one or two at the most) one hundred lines followed by a `write` operation from kak will be 'detected' by the hugo engine. Anything higher will face a countdown before that same text is finally rendered. With the countdown there's no 'message' from hugo.

Screwtapello:
Kakoune does a sensible, reasonable thing: writing out a file in parts. It would nice if Hugo had some documentation about how it expects files to be written in order to work correctly with its automatic reloading system. Does it expect the new file data to be fully available as soon as the first modification is seen? Does it expect the file to be written to a temporary file? Does it expect the file to *not* be written to a temporary file? If it wants a temporary file, does it matter if the temporary file is inside or outside the directory containing the target file? Does it matter if the temporary file's name is based on the target file's name?

nonumeros:
> sensible, reasonable thing: writing out a file in parts

@Screwtapello thanks for pitching in. I do agree with you there. 

On your https://github.com/mawww/kakoune/issues/2036#issuecomment-460418581 you broke it down nicely, except 
>hugo decides the file content is identical, does nothing

I'm not so sure this is the case, but the `write` operation from kak's side that never gets there...or better put that the `write` operation is never 'interpreted` by hugo, and when it does, it's on a timer. But this last part of either a timer or a countdown is irrelevant, because by then hugo no longer 'detects' the changes

> Does it expect the file to not be written to a temporary file?

Correct. That's a no no. And I noticed it when I created the file with kak. 

>If it wants a temporary file, does it matter if the temporary file is inside or outside the directory containing the target file?

outside. 

>Does it matter if the temporary file's name is based on the target file's name?

Can't answer that. That's beyond my knowledge. If I had to guess, I think it would have to be based on the filename.

perhaps @bep from hugo can give a better answer :)

bep:
If this is an issue with Hugo's "live reload" on changes, we use https://github.com/fsnotify/fsnotify -- which is highly OS dependent.

nonumeros:
> 

> If this is an issue with Hugo's "live reload" on changes, we use https://github.com/fsnotify/fsnotify -- which is highly OS dependent.

@bep thanks for the quick reply. But I don't think so. If it were I'd have the same issue with GNU/Emacs, neovim, to name a few.

nonumeros:
@Screwtapello you were asking about the reloading on hugo documentation. 

Hugo reloading comes from file https://github.com/gohugoio/hugo/blob/master/livereload/livereload.go from the livereload directory

It seems it also imports the websocket implementation from https://github.com/gorilla/websocket 

the website of the toolkit is at http://www.gorillatoolkit.org/pkg/websocket

nonumeros:
With the current implementation of both **file.cc** and **buffer.cc** `eoldata`  and `EolFormat` default values, respectively. , the hugo engine (live reloading) simply chokes by not detecting any changes' until it eventually crashes with
     
     panic: runtime error: invalid memory address or nil pointer dereference
     [signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0xd411ae]
     
     goroutine 169 [running]

Even if I were to modify the false/true values on buffer.cc, makes no difference. 

Of course, not detecting any changes by hugo is a reversible condition, as long as the number of lines on the given file is kept to a minimum, until it [hugo] restarts detecting the changes again. 

I wanted to recreate the same crash with other editors, but it's not possible. Needless to say, this was by running the same hugo version, same templates, same configuration, and same file. 

@mawww  @Screwtapello let me know if I can be of any help with this issue.

mawww:
@nonumeros Thanks for investigating this issue.

It would seem that there is a bug hugo itself, if it crashes after a hundred updates to the file it tracks.

All those tests were done with the default `overwrite` write method, right ? What happens when you use the `replace` write method (set global writemethod replace) ?

nonumeros:
> All those tests were done with the default overwrite write method, right ?

yes. If it's the default, yes.

> What happens when you use the replace write method (set global writemethod replace) ?

~~same behavior~~. 

it seems to be working. :)  Does it have to be `global`?

nonumeros:
@mawww  I could see one of the outstanding problems 

you're telling me that the default is `overwrite` but thtat's not what hugo is seeing 

this is definitely one hundred percent on kak's side, not hugo. (I'm not being cocksure here and or blaming it on kak, as my main objective is to have it working no matter what), but I can assure you that if `overwrrite` is supposed to do that, is not behaving as such. Being default, that is.

mawww:
Hello,

`overwrite` will write open the file truncated and write each lines one by one, which I believe is what you described initially. Hugo will hence see a series of writes and my understanding is that at some point it crashes after seing around 200 writes (each lines takes two writes, one for the line, one for end of line).

What Kakoune might be doing differently here (compared to emacs or vim) is that its actually very naive in its write implementation. It does not use buffering at all, which is very innefficient, it might make sense to introduce some buffering there to reduce the number of system calls.

if the `replace` method does not work, it would suggest hugo does not track a file by filename, but by some kind of file descriptor.

alexherbo2:
How to determine if the file being edited is managed by Hugo?

Currently, I have something like:

``` kak
hook global BufCreate '.*/(site-1|site-2|…)/.*' %{
  set-option buffer writemethod replace
}
```

mawww:
I just pushed a change that will buffer writes and might fix this issue, could you try the latest git and see if the issue is fixed ?

dmerejkowsky:
Looks like it's working :)

Thanks

nonumeros:
@mawww beautiful code.
@dmerejkowsky thank you.

PR #2816, opened 2019-03-27
Rust block indent

### LINKED ISSUE #2439: Rust syntax should always use block indenting
### Steps 

1. Open a file with the `.rs` extension:

        $ kak /tmp/foo.rs

2. Type something that looks like the beginning of a function call

        some_long_method_name(something,

3. Press `<ret>`

### Outcome 

Kakoune indents the next line so the cursor starts just underneath the `(` character.

### Expected 

Kakoune indents the next line by one `%opt{indentwidth}`, like if you hit `<ret>` immediately after the bracket.

### Notes

Once upon a time, it was fairly common in the Rust community to use "visual" indenting; that is, to indent wrapped lines so they line up with the opening bracket of the function call. However, since then there was a community-wide consultation and [the official formatting guidelines](https://github.com/rust-lang-nursery/fmt-rfcs/blob/master/guide/expressions.md#multi-line-calls) now recommend "block" indenting everywhere; that is, always indenting by exactly one `%opt{indentwidth}`.

The `rustfmt` tool that comes with the compiler cleans this up nicely, but I'm not always working in a codebase that uses fully-automatic formatting, and it's always nicer to avoid creating a problem than to clean it up.

I believe that [these lines](https://github.com/mawww/kakoune/blob/93f913705abafbd7341e4fdd897bef5f6ef4c0dd/rc/base/rust.kak#L55-L56) are responsible for the behaviour, but I don't have time to decipher how that works right now, so I'm filing this ticket in case somebody else wants to fix it before I get around to it.

GrantMoyer:
Fixes #2439.

Implements two changes to rust autoindent behavior:

1. Uses block indent for parameter lists. Also moves first element to it's own line.

    Given the line `long_function_name(param1,`, inserting a newline at the end results in the new behavior (where `█` represents where the cursor ends up after hitting enter):

    ```
    long_function_name(
        param1,
        █
    ```

    Old behavior

    ```
    long_function_name(param1,
                      █
    ```

2. Aligns closing parenthesis correctly with opening

    New behavior:

    ```
    long_function_name(
        param1,
        param2,
    )
    ```

    Old behavior:

    ```
    long_function_name(
        param1,
        param2,
        )
    ```

I've also added tests for this behavior.

If moving the first parameter to it's own line is too surprising, I can remove that behavior.


GrantMoyer:
Squashed commits as in #2811

mawww:
I do not write rust myself, so happy to change the indentation behaviour. Thanks a lot for this thorough pull request with tests, much appreciated.

PR #2817, opened 2019-03-27
Fix autoindent issues with paren in literal

### LINKED ISSUE #2565: [c-language] Parens in character literals cause incorrect indentation.
1. Open a new file w/ C file-type (`kak foo.c` will do the trick)
2. Insert the following:

```c
void foo(void) {
	if ('a' == '(') {}
}
```

3. Note how pressing enter after the closing brace of the if-statement causes spurious indentation

This is a pretty minimal test-case; the bug appears to be that kakoune treats the paren inside the character literal as part of the syntax.

GrantMoyer:
Fixes  #2565, and adds test case.

Conflicts with #2811.

New behavior (`█` represents cursor after newline):

```
if(c == '(') {}
█
```

Old behavior:

```
if(c == '(') {}
   █
```


GrantMoyer:
Rebased to resolve conflict with #2811

PR #2820, opened 2019-03-28
Don't bother wrapping before word if it's longer than a full line anyway

### LINKED ISSUE #2245: Wrapped indented line is pushed down
### Steps

`kak -n`

Type this, indented with two spaces:

```
  short line

  loooooooooooooooooooooooooooooooooooooooooooooong line
```

Enable wrapping: `:addhl global/ wrap -word -indent`

### Outcome

Notice that the line is pushed down:

kak-line-push

As a user I don't see any reason why it shouldn't stay on the line above:

image

### Expected

The line is not pushed down in this case.

GrantMoyer:
Fixes #2245.

Doesn't bother wrapping before a word if it is longer than a full line and would be broken in the word anyway. Instead just break inside the long word on the first line.

### DIFF
--- a/src/highlighters.cc
+++ b/src/highlighters.cc
@@ -662,8 +662,9 @@ struct WrapHighlighter : Highlighter
             const LineCount buf_line = it->range().begin.line;
             const ByteCount line_length = buffer[buf_line].length();
             const ColumnCount indent = m_preserve_indent ? line_indent(buffer, tabstop, buf_line) : 0_col;
+            const ColumnCount line_column_length = wrap_column - indent;
 
-            auto pos = next_split_pos(buffer, wrap_column, tabstop, buf_line, {0, 0});
+            auto pos = next_split_pos(buffer, wrap_column, tabstop, buf_line, {0, 0}, line_column_length);
             if (pos.byte == line_length)
                 continue;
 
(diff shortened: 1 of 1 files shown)


maximbaz:
Oh thank you soooo much for this! Works well :)

GrantMoyer:
Squashed commits as in #2811

mawww:
Hi, just to let you know, I think this is a good change, but I need to take the time to review it. It looks a bit bigger than I would have expected, so I need to take the time to see if there is no way around adding that much code to fix this issue (if you come up with a smaller, cleaner change in the mean time, I'll be extra happy).

GrantMoyer:
I figured something like that was the case, since changing C++ code is potentially a lot more dangerous than changing kakscript, escpecially when it comes to unicode.

The reason for the size of the change is basically that the current loop only iterates until the end of the current line. The new loop need to iterate past the end of the current line to see if the last word is longer than a whole line. It's possible these loops could be combined, but then we'd need to track whether or not we'd already reached of full line length within the combined loop. I'm not sure if this would lead to a smaller change overall.

mawww:
Hello,

Thanks for this PR, I gave a quick shot at implementing this myself initially to understand your own implementation, but I believe I arrived to a simpler change. I might have overlooked some corner case so please tell me if the pushed version has some drawbacks. Thanks again for proposing this change and for the initial implementation.

Cheers,

Maxime.

mawww:
Ahh, just understood the corner case that required you to pass the line length parameter. I'll add a few test cases and fix that.

GrantMoyer:
Your implementation looks correct to me.

PR #2826, opened 2019-03-30
Run default shell if terminal is given no arguments

laelath:
Modify the `terminal` alias to run the user's shell if `terminal` is called with no arguments.

This is incomplete as I am unfamiliar with macos, so I don't know how to change or test the iterm support.

### DIFF
--- a/rc/windowing/kitty.kak
+++ b/rc/windowing/kitty.kak
@@ -12,7 +12,7 @@ hook -group kitty-hooks global KakBegin .* %sh{
     fi
 }
 
-define-command kitty-terminal -params 1.. -shell-completion -docstring '
+define-command kitty-terminal -params .. -shell-completion -docstring '
 kitty-terminal <program> [<arguments>]: create a new terminal as a kitty window
 The program passed as argument will be executed in the new terminal' \
 %{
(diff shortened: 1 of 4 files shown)


mawww:
While I am sympathetic to the idea of making terminal more "helpful" by default, I am not really keen on having every implementation of terminal required to duplicate almost the same logic to open the default shell.

What you want can be implemented as a separate command in a trivial way: `def shell %{ terminal %val{client_env_SHELL} }`. I am not even sure it would really make sense to provide this by default, it is very easy to add to your kakrc if you want it.

I think `terminal` should stay focused on providing a common interface to open a terminal, possibly with additional windowing control (`-horizontal, -vertical, -tab...` as discussed in a separate issue).

Does that make sense ?

alexherbo2:
The changes involved are reasonable though.  Plus it would allow to run the default program of the terminal which is not necessarily the `SHELL`.

**Example** – [Alacritty]

``` yaml
shell:
  program: sh
  args:
  - --login
```

**Example** – [tmux]

```
set-option -g default-command sh
```

[Alacritty]: https://github.com/jwilm/alacritty
[tmux]: https://github.com/tmux/tmux

mawww:
Closing, as I dont think we should have this default behaviour.

Issue #2536, opened 2018-10-29
[Feature request] Conceal


andreyorst:
Conceal is somewhat arcane feature of Vim, that allows to display text differently from how it actually tstored in text document. It is widely used in Vim documentation, and another filetypes, to hide formatting symbols from eyesight. Concealing also can be used by various plugins to display information on a some point of line by dynamically generating conceal rules.

In short, concealing works this way: display "some text" from the file as "some other text"
Usage cases:
1. We can [add different ligatures](https://maximewack.com/post/emulating_ligatures/) for different filetypes, even if our font doesn't support ligatures.
2. Plugins such as [kak-lsp](https://github.com/ul/kak-lsp) would be able to display additional info about the code, like Intellij Idea does:
  ![image](https://user-images.githubusercontent.com/19470159/47630904-80d55300-db54-11e8-8936-e20197279837.png)
  (`:Unit` is displayed as concealed text)
3. Filetypes like `markdown` would be able to display such things like **\*bold text\*** as **bold text**.

I think that conceal should be a regexp highlighter, that acts like show whitespaces but for any regexp matched text:
```kak
add-highlighter global/div conceal "/" "÷"
add-highlighter window/arrow conceal "->" "→"
```
And so on.

While we're in insert mode on the same line with some concealed text, conceal maybe should be disabled, so we could see real contents of the line.

Screwtapello:
See also #453, in which code folding would be implemented by exactly the same mechanism.

mawww:
This should be handled by the (existing but limited) `replace-ranges` highglighter, and the (not existing yet) `replace-regex` highlighter.

An additional feature we need is a way to specify empty ranges (so that we can introduce the `:Unit` example), which we cannot now because we want to share the range syntax and the selection syntax (to make it easy to generate ranges from selections), and the selection syntax is necessarily inclusive (as selections are inclusive *and* oriented). As ranges support the <pos>+<len> syntax as well, we could change it to be non-inclusive, so that specifing an empty range would be <pos>+0.

andreyorst:
I've changed the issue name to represent the feature @mawww described in the previous comment.

I'm currently in process of updating my `kaktree` plugin, and we decided to store path info in the buffer. I have a question: will we be able to use replaced info in its original state for other highlighters?

What I mean is, if we have this buffer:
```
item [[some additional info]]
other_item [[some other info]]
```
And we replace ranges with nothing by using `replace-regex` highlighter.
What we actually see now:
```
item
[...]

Issue #453, opened 2015-11-01
code folding


b21e:
Arbitrary text folding like in vim zf.
Made this an issue so that I am notified of any progress.

lenormf:
Has the integration of soft wrapping made things easier, to support folding?

net:
Code folding would be **very** useful.

lenormf:
My question still stands.

mawww:
Could be easier yeah, the main problem with folding is similar to what we have with wrapping: we dont know how things are outside of the displayed window, so any movement primitive that would rely on how lines are displayed (necessary with folding if we dont want to open folds everytime the cursor moves on top of them) will behave strangely for non displayed selections.

In general, it seems tricky to have folding and multiple selections work together. Also, I have been trying to move away from display dependent movements, to make interactive use and script use of the editing language as close as possible.

Iiridayn:
I personally don't use code folding, though I've sincerely tried it (a team member loves it). I feel it makes it too easy to hide code smells which should be cleaned up anyway. Ie - file/method/if-then-else/loop body too long? Split it out (into a file or another method/function - especially long loop bodies). Interdependent code sections too far apart? Move them closer. Even so, I've found decent movement commands (usually `*`) work pretty well for single file navigation in vim w/o folds for even the worst files I've seen.

So, while I can see why folding would be difficult to do well, I don't mind the lack.

jeanm:
I also don't use code folding, although I know plenty of people who do. However, one good reason I see for supporting folding in general it is that it would enable the development of plugins similar to VimOutliner and org-mode. KakOutliner would rock.

laelath:
One way you could get around the problem of inconsistent syntax is to automatically expand a fold if a selection enters it.

mawww:
@andreyorst The expected difficulty was that it was not clear what say `j` in normal mode would do on undisplayed text that could be potentially folded.

I believe this difficulty has been solved by deciding that:

* normal mode keys should never take display information into account (with the exception of mouse clicks, which are fundamentally visual)
* Any highligthing that hides buffer text should disable itself when the hidden text is partially selected (i.e. neither fully selected, nor not selected at all).

With that said, I believe we can implement folding by extending the replace-ranges highlighter, currently it assumes (but does not enforce IIRC) that we replace a range with text the same length, we need to extend it so that it can replace arbitrary text, and then its just a matter having an external script setting a range-specs option to `<range to fold>|<text to display>`.

[...]

PR #2835, opened 2019-04-04
Org-mode language support

andreyorst:
I think basic support of [Org-mode][1] should be included by default. This will provide syntax highlighting and indentation rules much like [markdown.kak][2]

[1]: https://orgmode.org
[2]: https://github.com/mawww/kakoune/blob/master/rc/filetype/markdown.kak


rc/filetype/org.kak
```diff
@@ -0,0 +1,78 @@
...
+add-highlighter shared/org/inline/text/ordered-lists regex ^(?:\h*)(\d+[.)])\h+ 1:bullet
+
+# Headings
+add-highlighter shared/org/inline/text/heading    regex "^[*]+\h+[^\n]+"                          0:header
+add-highlighter shared/org/inline/text/italic     regex "\s([/][^\s/].*?[^\s/]*?[/])\W" 1:default,default+i
+add-highlighter shared/org/inline/text/verbatim   regex "\s([=][^\s=].*?[^\s=]*?[=])\W" 1:meta
+add-highlighter shared/org/inline/text/code       regex "\s([~][^\s~].*?[^\s~]*?[~])\W" 1:mono
+add-highlighter shared/org/inline/text/math       regex "\s([$][^\s].*?[^\s]*?[$])\W"   1:mono
+add-highlighter shared/org/inline/text/underlined regex "\s([_][^\s_].*?[^\s_]*?[_])\W" 1:default,default+u
+add-highlighter shared/org/inline/text/bold       regex "\s([*][^\s*].*?[^\s*]*?[*])\W" 1:default,default+b
```
Screwtapello: There are existing, standard faces named `italic` and `bold` you can use instead of `default,default+i` and `default,default+b`. I'm surprised there isn't already one for `underlined`, but it shouldn't be hard to add.

rc/filetype/org.kak
```diff
@@ -0,0 +1,78 @@
...
+# But `*' list must be indented with at least single space, if not it is treated as a heading
+add-highlighter shared/org/inline/text/star-list regex ^(?:\h+)([*])\h+ 1:bullet
+
+# Ordered list items start with a numeral followed by either a period or a right parenthesis, such as `1.' or `1)'
+add-highlighter shared/org/inline/text/ordered-lists regex ^(?:\h*)(\d+[.)])\h+ 1:bullet
+
+# Headings
+add-highlighter shared/org/inline/text/heading    regex "^[*]+\h+[^\n]+"                          0:header
+add-highlighter shared/org/inline/text/italic     regex "\s([/][^\s/].*?[^\s/]*?[/])\W" 1:default,default+i
+add-highlighter shared/org/inline/text/verbatim   regex "\s([=][^\s=].*?[^\s=]*?[=])\W" 1:meta
```
Screwtapello: What is "verbatim" and how does it differ from "code"?

rc/filetype/org.kak
```diff
@@ -0,0 +1,78 @@
...
+add-highlighter shared/org/inline/text/code       regex "\s([~][^\s~].*?[^\s~]*?[~])\W" 1:mono
+add-highlighter shared/org/inline/text/math       regex "\s([$][^\s].*?[^\s]*?[$])\W"   1:mono
+add-highlighter shared/org/inline/text/underlined regex "\s([_][^\s_].*?[^\s_]*?[_])\W" 1:default,default+u
+add-highlighter shared/org/inline/text/bold       regex "\s([*][^\s*].*?[^\s*]*?[*])\W" 1:default,default+b
+add-highlighter shared/org/inline/text/link       regex "\[[^\n]+\]\]"                            0:link
+add-highlighter shared/org/inline/text/option     regex "(?i)#\+[a-z]\w*\b"                       0:meta
+
+# Strikethrough highlighter is disabled because Kakoune doesn't support strikethrough attribute
+# add-highlighter shared/org/inline/text/strikethrough regex "\W*\+[^\s][^\n+]+[^\s]\+\W" 0:default,default+s
```
Screwtapello: You should still define the regex so it can be highlighted with a colour, even if Kakoune can't display the actual strikethrough yet.

Maybe even add a standard "strikethrough" face beside the bold, italic, and underlined ones.

rc/filetype/org.kak
andreyorst: I didn't use those faces because org mode in Emacs doesn't colors those, but adds attribute to the text. This way nesting is also possible, like bold links that still use link face.

rc/filetype/org.kak
andreyorst: Idk. It's what Org-mode specifies in documentation. On practice those are very similar, but have different faces in Emacs. When exporting Org-mode to PDF via LaTeX they both use `\texttt{}` though.

rc/filetype/org.kak
andreyorst: >You should still define the regex so it can be highlighted with a colour, even if Kakoune can't display the actual strikethrough yet.

Agree, I've commented it out to decide later.

>Maybe even add a standard "strikethrough" face beside the bold, italic, and underlined ones.

I want to extend all current themes with new standard faces for markup languages, like `heading`, `subheading`, `subsubheading`, `subsection`, `subsubsection`, `subsubsubsection`, and others, becouse in current state highlighting is limited too strictly.

rc/filetype/org.kak
Screwtapello: I get the impression that many/most Emacs users run it in a GUI mode so they can get all the fancy formatting features, so they can assume that those attributes are available without any kind of fallback. Since Kakoune is primarily a terminal-based editor, it can't safely assume those attributes will always be available, so it's good to have the fallback of Kakoune's standard colors.

That said, nesting *should* be possible in Kakoune, since attributes aren't mutually exclusive. It doesn't seem to work in the Kakoune ncurses UI, though: if I create a Markdown document with:

```markdown
A test of **some bold _with italics_ text**
```

the text "with italics" is red and italic but not bold in regular Kakoune, but red and bold-italic in my experimental `kakoune-gtk`. How strange.

andreyorst: > I get the impression that many/most Emacs users run it in a GUI mode so they can get all the fancy formatting features, so they can assume that those attributes are available without any kind of fallback.

The _markup symbols_ itself are fallback, e.g. /*italic text*/ looks nice, but if attribute isn't available one can still distinguish /italic text/ (kinda old convention for italic formatting) from non-italic. Same goes for \***bold text**\*. I mean, it's not Markdown, where we really need to track the count of `*` or `_` to find out if it is providing any attribute to the text.

andreyorst: > That said, nesting should be possible in Kakoune, since attributes aren't mutually exclusive. It doesn't seem to work in the Kakoune ncurses UI, though

My main concern is that bold links and must be colored as links with bold attribute. I guess this can be specified in the theme though.

Delapouite:
org-mode is a behemoth of the emacs ecosystem. I'm afraid that by providing a half-assed version of the gazillions stuffs the original can do, it will open a constant flow of users complaining about "this feature is missing", "the behavior is slightly different" etc…

One of the obvious functionality missing from kakoune right now is folding support, which is omnipresent while working with org-mode.

andreyorst:
> org-mode is a behemoth of the emacs ecosystem. I'm afraid that by providing a half-assed version of the gazillions stuffs the original can do, it will open a constant flow of users complaining about "this feature is missing", "the behavior is slightly different" etc…

I think that Org-mode initial support is crucial, but I also think that if anyone wants to work with Org, he will juts use Emacs. I mean, If I would like to write some Vim documentation (with their markup) I would use Vim for that, but being available to read and do quick edits is still a good thing. I don't know any Vim user who uses Org-mode for real. But I know a lot of people who uses Emacs + Evil to write Org-mode.
Also org-mode plugins exist for other editors and they only provide syntax highlighting, and pretty much everyone if fine with that.

After using Emacs for Markdown, I can say that Kakoune laks lots of markdown related features too, but I'm good with that.

>One of the obvious functionality missing from Kakoune right now is folding support, which is omnipresent while working with org-mode.

so such mode would boost priority on this feature, I guess?

rc/filetype/org.kak
Screwtapello: Yeah, the markup symbols themselves are fallbacks, but I think an Org-mode plugin that ships with Kakoune should follow the same conventions as all the other plugins that ship with Kakoune. If it were a third-party plugin, it wouldn't matter so much (although if it were me, I'd still follow the Kakoune conventions rather than the Org-mode conventions).

Screwtapello:
As a single data-point, I got interested in Org-mode when I found [a neat TODO manager for Android](https://f-droid.org/en/packages/com.orgzly/) that stores data in .org files. It would be nice to have syntax-highlighting if I edit the files in Kakoune, but I can't imagine it being worth my time to learn Emacs just to add milk to my shopping list.

TeddyDD:
Thanks for working on this, I'v got a lot of org files from time when I was using Emacs, it will be nice to have syntax highlighting for them

rc/filetype/org.kak
andreyorst: Got it. Well, I guess the perfect solution would be to use faces, and override those faces in user configuration then.

andreyorst:
I've added indentation handling for lists, and by now this script covers pretty much everything that I use write Org-mode. @TeddyDD Can you test it on your files, to see if there are unhandled cases popping out?

TeddyDD:
All my files are looking okay

andreyorst:
I'm currently having troubles with defining a regexp for markup that will find that:
```
*text
text*
```
But not that:
```
*text

text*
```

andreyorst:
I've added faces to distinguish `header` from `section`, `subsection`, and `subsubsection`

andreyorst:
I think that it's ready in terms of syntax support. I've thrown some pretty big and complex org files into it and it displays them correctly.

andreyorst:
Oops I forgot to handle TODOs and Priorities

andreyorst:
Found [this document](https://orgmode.org/worg/dev/org-syntax.html). Going to re-check if I've done everything right

andreyorst:
I'm satisfied with provided highlighting and all test files look exacly the same between Emacs and Kakoune. This got pretty complex in various places, but I've followed [Org Syntax Document](https://orgmode.org/worg/dev/org-syntax.html) and tried my best to provide optimal regular expressions for the syntax elements described in it.

colors/default.kak
```diff
@@ -17,8 +17,13 @@ face global builtin default+b
 # For markup
 face global title blue
 face global header cyan
-face global bold red
-face global italic yellow
+face global section blue
+face global subsection magenta
+face global subsubsection red
+face global bold red+b
```
mawww: I do not like those 4 new faces (bold, itallic, underline and strikethrough), we already have attributes to express those (even though strikethrough is not supported pending replacing ncurses).

Regarding section/subsections/subsubsection, we already have header and title which are pretty close, I'd like to see if we can merge those.

andreyorst: But `bold`, `italic`, and `underline` are not new faces. I only added `strikethrough` because it was missing and it's good to have a face for it along with those even if it only can be expressed by color.

`Title` face is used to highlight the title, and I think that it is different from `header` (although I'd like to rename `header` to `heading`), because title can be only one per document, and there may be any amount of headings.

mawww: Oh, yeah, you are right, `bold` and `italic` already exist, however I think I'd rather see them go and just use `default+b` and `default+i`. What we have in the `default` colorscheme is the minimum set of faces a colorscheme must support, and I would like to keep that amount of colors in check.

I think having a single `heading` face shared by all section/subsection/subsubsection is probably enough, eventually we could use `title` for top level sections if we really need a different highlighting for those.

andreyorst: Having `bold`, `italic`, `underline`, and `strikethrough` faces is great when you don't have support for those faces in the terminal, and you can code those attributes with colors. Especially with `strikethrough` which isn't supported with `ncurses`. 

In Org Mode, different heading levels represent nodes of a tree structure. Highlighting them in different colors makes semantic meaning in the Org language. And It helps organize TODOs.
Here's an example of agenda in Org mode, whith `(sub){1,2}section` and `header` highlighters using different faces:
image
Here's how in looks like when all headings use the same face:
image

Emacs screenshots:
image
Also [concealing](https://github.com/mawww/kakoune/issues/2536) can hide bold, italics, and link markers to make it look more like text.

When [folding](https://github.com/mawww/kakoune/issues/453) will be implemented, I'll add it to this script in order to support this kind of thing:
image
Notice that headings in fzf.kak entry are hidden, as well as text in the four level section.

I think I can move those faces to the Org mode highlighter, and set them by default to default `heading` face. What do you think about it?

andreyorst: >  eventually we could use title for top level sections if we really need a different highlighting for those.

`title` has another semantic meaning in Org mode.
Emacs:
image

Kakoune:
image

Again, I can set all these faces only for Org within the script.

andreyorst:
> What we have in the default colorscheme is the minimum set of faces a colorscheme must support, and I would like to keep that amount of colors in check.

@mawww I've thought about it, and opened an issue that I think can solve this problem and will also allow scripts to be more unique in terms of faces: https://github.com/mawww/kakoune/issues/2862

This also means that this script could be reworked to provide faces that are needed by Org mode, but not needed by another scripts. And with [lazy loading model proposed](https://github.com/mawww/kakoune/pull/2782) this will not cost too much.

mawww:
@andreyorst I think that is my preferred way for faces: have a very general and rather small set of faces provided as default (in the default color scheme) and expected to be provided by every colorscheme. Then let various scripts add their own faces if they do need them with default values based off the default color scheme (while encouraging re-use of the default ones when possible). Other colorschemes can then decide to add support for those faces as well.


4e24ba8 should help with that, as it makes it possible for scripts to base faces off other faces with slight changes (additional attributes, changing only fg or bg color...)

andreyorst:
> @andreyorst I think that is my preferred way for faces: have a very general and rather small set of faces provided as default (in the default color scheme) and expected to be provided by every colorscheme. Then let various scripts add their own faces if they do need them with default values based off the default color scheme (while encouraging re-use of the default ones when possible). Other colorschemes can then decide to add support for those faces as well.
> 
> [4e24ba8](https://github.com/mawww/kakoune/commit/4e24ba86cc95c51f78ebe9e77f1848b6760dbeb5) should help with that, as it makes it possible for scripts to base faces off other faces with slight changes (additional attributes, changing only fg or bg color...)

This is great change but I see the major problem here. Faces are faces, not colors. If I need to display something in `red`, that matches current colorscheme, I can't use face, because in another color scheme this face may not be `red`. What I mean is that `red` in `tomorrow-night` is `#cc6666`, and `red` in `solarized-dark` is `#dc322f` and if I want something to be `red` I need the color code itself. I could use Error face as a base, but Error faces (which may actually be any color) in these color schemes are different:

    face global Error ${red},default+b # solarized dark
    face global Error white,${red} # tomorrow night 

That's why I think that colorschemes should also provide colors and not only faces.

mawww:
That would be quite a big change, currently colors are fully immutable, the named ones are just a way to access the terminal provided colors.

I can see the additional flexibility of having named colors as well as named faces, but at the same time, it seems they are very close in purpose, and having a single unified way to specify a colorscheme would be better.

Basically it seems the question boils down to whether we want a colorscheme just to provide a palette of colors, or a mapping from semantic names to faces. I would prefer to find a solution that does not requires to define both.

andreyorst:
Currently most of color schemes define colors as shell variables. Why not define them as options instead? It will be a set of main 16 colors, and some colors for `fg` `bg` and cursors defined as options. Not really a big change, since the only thing that really changes in addition of several options. Like here: https://github.com/andreyorst/powerline.kak/blob/master/rc/themes/github.kak

andreyorst:
@mawww  I understand that this is kind of duplication of the same thing, but it allows granular control over the colors, **and** currently some themes and scripts fallback to terminal colors, but some users like me use different colorscheme in Kakoune and terminal, and this causes bad colors like this:

image

I use light colors cheme in terminal but dark inside kakoune, and this is the result of using plain colors.

mawww:
named colors currently refer to the terminal colors, making them an option seems confusing to me... Maybe letting user tweak their terminal color palette would be a simpler and better way to solve that (and having more colorscheme using the terminal colors instead of rgb colors).

Faces are pairs of colors, I am not sure we can guarantee that every pair are going to be legible, so I am not even sure specifying colors individually is going to be that useful. It might be better to be able to express more things directly on faces, such as "Error face with brighter fg" or "Error face with dimmer fg"...

Lets get back at what we want (as I understand it):

1. We want a way to add new faces in scripts for use cases where no default faces are good enough.
2. We want those faces to be settable by some color scheme if their authors decide to add support for this script
3. We want default values for those new faces not to look out of place in colorschemes that did not add explicit support for them.

1. and 2. are already handled, what we are trying to solve is 3.

Having a color palette, possibly handled through options could help, but could be tricky as we need to guarantee that whatever colors are chosen as default do look reasonably readable (fg against bg, both of those possibly implicitely defined by the colorscheme, although we should be able to assume any color scheme provided fg color is readable on its default bg).

Another option is to take a known to exist face, and add tweak it slightly, it should be harder to end-up with something unreadable, but there is not always a face semantically close to what we want, and sometimes we really want a specific color (diff highlighting for example).

None of those is really satisfying.

Screwtapello:
One issue with things like "powerline" (I haven't looked at @andreyorst's plugin specifically, but I've seen powerline plugins for Vim and bash) is that it tries to have overlapping blocks. For example, if you have a white-on-green block next to a white-on-blue block, there's a special transition character between them that needs to be drawn in green-on-blue. If a theme provides a white-on-green face and a white-on-blue face, there's no way for `powerline.kak` to combine them to produce the green-on-blue face it needs.

I *think* @andreyorst has worked around this by defining options for all the individual foreground and background colours so that he can synthesise whatever combinations he needs, but perhaps that's the kind of thing that should have dedicated syntax. For example, maybe something like `SomeFace.fg` to refer to the foreground colour of the `SomeFace` face:

    set-face global InfoJoinMenu Info.bg,MenuBackground.bg

andreyorst:
The main problem that I see in current implementation is when I need a face which has for example purple (magenta) foreground and black background, that will stay purple/black across different color schemes, but will respect purple and black colors defined by color scheme. How can I achieve it with current approach?

andreyorst:
> One issue with things like "powerline" (I haven't looked at @andreyorst's plugin specifically, but I've seen powerline plugins for Vim and bash) is that it tries to have overlapping blocks. For example, if you have a white-on-green block next to a white-on-blue block, there's a special transition character between them that needs to be drawn in green-on-blue. If a theme provides a white-on-green face and a white-on-blue face, there's no way for powerline.kak to combine them to produce the green-on-blue face it needs.

Correct. To workaround this in my powerline plugin, I declare colors and use them to dynamically create faces.

> I think @andreyorst has worked around this by defining options for all the individual foreground and background colours so that he can synthesise whatever combinations he needs, but perhaps that's the kind of thing that should have dedicated syntax. For example, maybe something like SomeFace.fg to refer to the foreground colour of the SomeFace face

This would be great but for powerline I need colors, not faces. I try to avoid separate block faces, because it will make harder to create new blocks, as they will also require a new face.

Aslo extending your syntax:

    set-face global InfoJoinMenu Info.bg,MenuBackground.bg.Error.attributes

Maybe this discussion should continue in https://github.com/mawww/kakoune/issues/2862?

Screwtapello:
I don't think it makes sense to say "I want a face with purple foreground and black background" and "I want a face that looks natural with the user's current colour scheme". If the user's current colour scheme is green, orange and white, purple and black will *never* look right, no matter what RGB values you happen to pick.

As another example, Solarized includes a `green` colour, but not a `bright-green`. If you ask for `bright-green` and the user is using Solarized, what would you expect to get? If you get ordinary green, then bright-green-on-green is invisible for Solarized users. If you get the colour that replaces bright-green in the Solarized terminal palette, that's actually a pretty dim background colour, so if you wanted something contrasting against a black background, that won't work.

andreyorst:
> I don't think it makes sense to say both "I want a face with purple foreground and black background" and "I want a face that looks natural with the user's current colour scheme". 

I didn't mean natural. I meant consistent. To me this is far less consistent:
image

andreyorst:
@mawww Now, when pretty much most of the languages provide modules do I need to require them all when opening org file in order to highlight source blocks? It seems that without requiring a module highlighting doesn't happen. It affects markdown.kak too:

markdown.kak:
image

org.kak:
image

If module is required highlighting is visible:
image
image

Both scripts define those highlighters for all languages at once.
markdown.kak:
https://github.com/mawww/kakoune/blob/9ffde03ca74ce691882c47776fe8d014792dff4c/rc/filetype/markdown.kak#L36-L49

[org.kak](https://github.com/andreyorst/kakoune/blob/6434ce9f5020583c5f0fc116771e68889f7e7f46/rc/filetype/org.kak#L71-L94)

But I feel that requiring all modules kinda violates lazy loading model.

andreyorst:
I've updated the script to detect and load language modules as it is done currently in markdown.kak

This PR is probably ready to be reviewed.

rc/filetype/org.kak
```diff
@@ -0,0 +1,262 @@
...
+set-face global org_done          default,default+b@string
+set-face global org_priority      default,default@value
+set-face global org_heading       default,default+b@header
+set-face global org_section       default,default@function
+set-face global org_subsection    default,default@variable
+set-face global org_subsubsection default,default@module
+set-face global org_bold          default,default+b
+set-face global org_italic        default,default+i
+set-face global org_underline     default,default+u
+set-face global org_strikethrough default,default
```
lenormf: There are already global faces for markup (`header`, `section` etc), why not let the user override them whenever the filetype is `org`, instead of creating duplicate faces?

rc/filetype/org.kak
```diff
@@ -0,0 +1,262 @@
...
+        esac
+        printf "%s\n" "add-highlighter shared/org/${lang}/inner region \A#\+(?i)BEGIN_SRC(?I)\h+[^\n]+\K '(?i)(?=#\+END_SRC)' ref ${ref}"
+    done
+}
+
+# No-language SRC block
+add-highlighter shared/org/src_block region ^(\h*)(?i)#\+BEGIN_SRC(\h+|\n) ^(\h*)(?i)#\+END_SRC\h*$ fill meta
+
+# Comment
+add-highlighter shared/org/comment region (^|\h)\K#[^+] $ fill comment
```
lenormf: Maybe a `regex` would be clearer here, e.g.

```
regex (^|\h)(#[^\n]+) 1:comment
```

rc/filetype/org.kak
```diff
@@ -0,0 +1,262 @@
...
+
+# Table
+add-highlighter shared/org/table region ^\h*[|][^+] $ fill string
```
lenormf: Same as above.

rc/filetype/org.kak
```diff
@@ -0,0 +1,262 @@
...
+# Headings. Also includes highlighting groups for TODO and PRIORITIES
+# format: STARS TODO PRIORITY TEXT TAGS
+add-highlighter shared/org/inline/text/heading          regex   '^(?:[*]{1}|[*]{5}|[*]{9})\h+[^\n]*?(:[^:\n]*?:)?\n' 0:org_heading       1:module
+add-highlighter shared/org/inline/text/section          regex  '^(?:[*]{2}|[*]{6}|[*]{10})\h+[^\n]*?(:[^:\n]*?:)?\n' 0:org_section       1:module
+add-highlighter shared/org/inline/text/subsection       regex  '^(?:[*]{3}|[*]{7}|[*]{11})\h+[^\n]*?(:[^:\n]*?:)?\n' 0:org_subsection    1:module
+add-highlighter shared/org/inline/text/subsubsection    regex '^(?:[*]{4}|[*]{8}|[*]{12,})\h+[^\n]*?(:[^:\n]*?:)?\n' 0:org_subsubsection 1:module
+add-highlighter shared/org/inline/text/heading_comment  regex '^[*]+\h+[^\n]*((?<=\h)COMMENT)\s'                     1:comment
+add-highlighter shared/org/inline/text/heading_todo     dynregex '^[*]+\h+(?:%opt{org_todo})'                        1:org_todo 2:org_done
+add-highlighter shared/org/inline/text/heading_priority dynregex '^[*]+\h+[^\n\[]*(\[#(?:%opt{org_priority})\])'     1:org_priority
+# (?:(?:%opt{org_todo})\h+)?(\[#(?:%opt{org_priority})\])?
```
lenormf: Was that line written for debugging, or as a test? It should probably be an empty line.

rc/filetype/org.kak
```diff
@@ -0,0 +1,262 @@
...
+# Options
+add-highlighter shared/org/inline/text/option     regex "(?i)#\+[a-z]\w*\b[^\n]*"          0:module
+add-highlighter shared/org/inline/text/title      regex "(?i)#\+title:([^\n]+)"            0:module 1:title
+add-highlighter shared/org/inline/text/requisites regex "(?i)#\+(?:author|email):([^\n]+)" 0:module 1:keyword
+
+# Drawer
+add-highlighter shared/org/inline/text/drawer   regex "^\h*([:][^\s][^\n]*?[^\s]*?[:])\W" 1:keyword
```
lenormf: `[:]` could probably be written simply as `:`, and `[^\s]` as `\S`. Note that `\n` matches `\s` too.

rc/filetype/org.kak
```diff
@@ -0,0 +1,262 @@
...
+
+# Timestamps
+declare-option -docstring "Org date
+          format: YYYY-MM-DD DAYNAME" \
+regex org_date '\d\d\d\d-\d\d-\d\d\h+[^\s-+>\]\d]+'
```
lenormf: Using the `\d{n}` notation is more elegant than repeating the escape sequence, and the last word could also simply be `\w+`.

rc/filetype/org.kak
```diff
@@ -0,0 +1,262 @@
...
+add-highlighter shared/org/inline/text/verbatim       regex "(^|[\h({'""])([=][^\h,'""][^\n]*?(\n[^\n]*?[^,'""\s])?[=])[\s.,:!?')}]" 2:meta
+add-highlighter shared/org/inline/text/code           regex "(^|[\h({'""])([~][^\h,'""][^\n]*?(\n[^\n]*?[^,'""\s])?[~])[\s.,:!?')}]" 2:mono
+add-highlighter shared/org/inline/text/underlined     regex "(^|[\h({'""])([_][^\h,'""][^\n]*?(\n[^\n]*?[^,'""\s])?[_])[\s.,:!?')}]" 2:org_underline
+## will be deprecated in future releases of Org but currently is supported
+add-highlighter shared/org/inline/text/inline-math    regex "(^|[\h({'""])([$][^\h,'""][^\n]*?(\n[^\n]*?[^,'""\s])?[$])[\s.,:!?')}]" 2:mono
+## bold is kinda tricky because we need to HL everything but headings, so it's split up on several regexps
+add-highlighter shared/org/inline/text/bold regex "(?:^|[\h({'""])([*][^\h,'""][^\n]*?(\n[^\n]*?[^,'""\s])?[*])[\s.,:!?')}]|([*]{3,})\n|\h([*]{3})[\s.,:!?')}]" 1:org_bold
+
+# LaTeX
+add-highlighter shared/org/LaTeX region -match-capture '\\begin\{([A-Za-z0-9*]+)\}' '\\end\{([A-Za-z0-9*]+)\}' fill string
```
lenormf: Maybe you could just reference the LaTeX highlighter directly here (`ref latex`)?

rc/filetype/org.kak
```diff
@@ -0,0 +1,262 @@
...
+                                                                                             $_ =~ s/(.*) ([^\s]+$)/($1)|($2)/;
+                                                                                             $_ =~ s/\(\s+/(/g;
+                                                                                             $_ =~ s/\s+\)/)/g;
+                                                                                             $_ =~ s/[\t ]+/|/g;
+                                                                                         }' }
+    }
+    # Parse priority items
+    try %{
+        # this `execute-keys' selects first three letters in `#+PRIORITIES' line
+        execute-keys -draft '/(?i)^\h*#\+PRIORITIES:[^\n]+<ret><a-h>f:l<a-l>s\h*\w(\s+)?(\w)?(\s+)?(\w)?\s<ret>y: set-option buffer org_priority %reg{dquote}<ret>'
```
lenormf: Maybe this line would be clearer written as an `eval` with `exec`s in it.

rc/filetype/org.kak
```diff
@@ -0,0 +1,262 @@
...
+    set-option buffer filetype org
+}
+
+# Initialization
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+hook global WinSetOption filetype=org %{
+    require-module orgmode
+    hook window InsertChar \n -group org-indent org-indent-on-new-line
+    hook -once -always window WinSetOption filetype=.* %{ remove-hooks window org-.+ }
```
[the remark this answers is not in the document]
andreyorst: I've took these hooks from markdown.kak in their original state: https://github.com/mawww/kakoune/blob/master/rc/filetype/markdown.kak#L18

rc/filetype/org.kak
andreyorst: Sorry, I didn't get what you meant. I've created these faces in order to let users override settings specifically for `org` if they want, without affecting other markup language highlighting. 

There's only `header` face for markup right now, and `heading` and `section` are different things in org by means of LaTeX

rc/filetype/org.kak
andreyorst: I followed the org spec here, and I thought that there may be difference in actual `\w` and this regex, since user can add `extra_word_chars` so the data format will be broken.

rc/filetype/org.kak
andreyorst: that's nice idea, I'll try

rc/filetype/org.kak
andreyorst: not sure what's in your mind

rc/filetype/org.kak
lenormf: Overloading faces for a specific filetype is trivial now that faces are scoped:

```
hook global WinSetOption filetype=org %{
    set-face window header red
}
```

Use global faces (i.e. the ones declared in colorschemes) as much as possible, users can overload them in their own configuration if needed: `bold`, `italic`, `header`…

I see you've already had a discussion about faces earlier with @mawww, maybe I'm just repeating what he/you said.

rc/filetype/org.kak
lenormf: Something like the following (untested), to make things more readable:

```
evaluate-commands -draft %{
    execute-keys /(?i)^\h*#\+PRIORITIES:[^\n]+<ret> \
                          <a-h>f:l<a-l> \
                          s\h*\w(\s+)?(\w)?(\s+)?(\w)?\s<ret> \
                          y
    set-option buffer org_priority %reg{dquote}
}
```

andreyorst: Ah, got it, I'll redo this part

mawww:
Hello,

I have been looking at that, and wonder if this should be part of the default distribution, or provided as a separate plugin. Org mode is a quite popular so it would make sense to provide it, but at the same time I would expect such a script to quickly grow with additional features, and it might make more sense to keep it in its own repository.

In any case, would it be possible to clean-up the history a bit, rebase and squash things together ?

andreyorst:
it's not yet ready. I had very little time, and probably will not have enough till next month. The script is 90% ready, of this OK I can rebase an squash

mawww:
No rush then, can you just add [WIP] to the PR title, and remove it whenever it is ready.

andreyorst:
Oh, sorry for that, I thought that it's there, it shows up in my inbox as:
image
I do not reply through email though, but it moves me automatically to the end of discussion so I do not look at the title much.

andreyorst:
@mawww Sorry for such long time not working on this. I think that this feature is ready now, but I have troubles with squashing everything. I can go and create new branch with this file as a single commit, or if it is possible to squash when merging, go for it. What do you think?

billksun:
@andreyorst, do you still plan to merge in org-mode support? This would be a very attractive feature for me to stick with kakoune.

andreyorst:
no, its pretty useless, and the spec is fragile, it's better to use real parser, which is currently in the works.

PR #2849, opened 2019-04-11
Add ClientClosePre hook

JJK96:
This extra hook can be used to handle stuff in the client that is about to be closed, such as closing the current buffer

I will update the documentation if this is going to be merged
Edit: updated it already

### DIFF
--- a/doc/pages/hooks.asciidoc
+++ b/doc/pages/hooks.asciidoc
@@ -143,7 +143,10 @@ name. Hooks with no description will always use an empty string.
 *ClientCreate* `client name`::
     executed when a new client is created.
 
-*ClientClose* `client name`::
+*ClientClosePre* `client name`::
+    executed before a client is closed
+
+*ClientClosePost* `client name`::
     executed when a client is closed, after it was removed from the client
     list.
 
--- a/src/client_manager.cc
+++ b/src/client_manager.cc
@@ -111,6 +111,8 @@ bool ClientManager::has_pending_inputs() const
 
 void ClientManager::remove_client(Client& client, bool graceful, int status)
 {
+    auto& context = client.context();
+    context.hooks().run_hook(Hook::ClientClosePre, context.name(), context);
     auto it = find(m_clients, &client);
     if (it == m_clients.end())
     {
@@ -122,8 +124,7 @@ void ClientManager::remove_client(Client& client, bool graceful, int status)
     m_client_trash.push_back(std::move(*it));
     m_clients.erase(it);
 
-    auto& context = client.context();
-    context.hooks().run_hook(Hook::ClientClose, context.name(), context);
+    context.hooks().run_hook(Hook::ClientClosePost, context.name(), context);
 
     if (not graceful and m_clients.empty())
         BufferManager::instance().backup_modified_buffers();
--- a/src/hook_manager.hh
+++ b/src/hook_manager.hh
@@ -29,7 +29,8 @@ enum class Hook
     BufReadFifo,
     BufSetOption,
     ClientCreate,
-    ClientClose,
+    ClientClosePre,
+    ClientClosePost,
     InsertBegin,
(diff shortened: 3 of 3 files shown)


Delapouite:
It looks like this branch contains many unrelated commits. A rebase problem?

JJK96:
I noticed, fixed it

andreyorst:
such hook will suit me even better than `ClientClose` to handle tagbar.kak with unsaved buffers

mawww:
I changed ClientClose to run at a slightly different time, hopefully this is enough for most use case. If not we can explore having both ClosePre and ClosePost, but I'd like to avoid that if possible.

andreyorst:
> I changed ClientClose to run at a slightly different time, hopefully this is enough for most use case. If not we can explore having both ClosePre and ClosePost, but I'd like to avoid that if possible.

The main idea I had when I thought about `ClientClosePre` was to handle unsaved buffers in current closing client instead of `toolsclient` which in my case should be closed before main client. Currently with `ClientClose` Hook executed after client closes, so I need to make `toolsclient` an ordinary client and user may get confused by this.

mawww:
Is the new behaviour of ac73f7c sufficient for your use case ?

andreyorst:
ac73f7c had not changed behavior for me. I'm calling a command in this hook, that removes other hooks, deletes tool buffer and closes tool client if it can, but if there are unsaved buffers, we can't close client

https://asciinema.org/a/SAryjwn0Whzmbg3oWABX6RAnE

mawww:
So, if I understand correctly, the problem is not really the point at which ClientClose runs, but more that the current way we validate if a client can close or not (if its not the last or if all the buffers are saved) does not fit well with "helper" clients such as tagbar.

I am not sure I understand how having two hooks would help there, as that check is done before any of those would run. I do agree it would be nice to have a solution for that, not sure what it would look like though.

andreyorst:
> I am not sure I understand how having two hooks would help there, as that check is done before any of those would run. I do agree it would be nice to have a solution for that, not sure what it would look like though.

For example we have two clients `client0`, and `client1` with one modified buffer.

```kak
hook global ClientClosePre .* %{
    evaluate-commands -client 'client1' %{ quit }
}
```
now we quit `client0`:
```sh
#user executes `:quit' command
:quit
#client is about to close, server needs to do everything before closing it
evaluate-commands -client 'client1' %{ quit }
#client1 was closed
#client0 should close now, but it can't, because buffer has unsaved changes
```
At this point we have `client0` with unsaved buffer and no other clients.

---

The same thing with `ClientClosePost`
```kak
hook global ClientClosePost .* %{
    evaluate-commands -client 'client1' %{ quit }
}
```

Again we quit `client0`:
```sh
#user executes `:quit' command
:quit
#client closes and server need to run all hooks after it:
evaluate-commands -client 'client1' %{ quit }
#client1 can't be closed bevause buffer has unsaved changes
```
At this point we have `client1` with unsaved buffer and no other clients.

---

`ClientClosePost` is useful when you have to do something when you exit client, but you don't really need any active client to exist, so even if you close last client the hook will be executed before server shuts down. This is current behavior of `ClientClose` hook.

`ClientClosePre` is useful when you want to execute something before client is closed and abort if it can't be closed.

mawww:
I see, however I dont think the first case would have worked as you described with the proposed change because the check for closing is done inside the `quit` command itself:

```
#user executes `:quit' command
:quit
# validation that the client can quit happens now, it can, so we proceed with closing client0
evaluate-commands -client 'client1' %{ quit }
#client1 is closed, as there are two clients in the client list, so the second quit commands goes through
#client0 closes now, as there is no re-validation that it can, acting as-if a `:quit!` command had been run.
```

In order to achieve the desired behaviour, we would need to move ClientClosePre much earlier, or move the no-modified-buffers validation if last client later.

PR #2860, opened 2019-04-17
Replace egrep with grep

alexherbo2:


Screwtapello:
Is egrep non-POSIX or something?

alexherbo2:
@Screwtapello I was thinking it was not. :confused:

nonumeros:
https://www.gnu.org/software/findutils/manual/html_node/find_html/posix_002degrep-regular-expression-syntax.html 

I only have use for egrep.

lenormf:
Should be fine to keep `{e,f}grep` around a little longer, even Busybox has them. But in the -very- long term, it's probably best not to rely on them.

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/grep.html#tag_20_55_18

>This grep has been enhanced in an upwards-compatible way to provide the exact functionality of the historical egrep and fgrep commands as well. It was the clear intention of the standard developers to > consolidate the three greps into a single command.
>
>The old egrep and fgrep commands are likely to be supported for many years to come as implementation extensions, allowing historical applications to operate unmodified.

Issue #2640, opened 2018-12-17
Deprecate the documentation in the README, redirect to the doc pages


lenormf:
Hi,

I think the documentation in the README has fallen out of sync with the documentation pages accessible from the editor, and bringing it up to speed might not be worth it, considering how redundant it's been.

I made some changes in a previous PR that was ultimately closed, maybe we could dig them up?

https://github.com/lenormf/kakoune/blob/a659d1f2b48f2cf153a778cc593d6f8f56c58486/README.asciidoc

HTH.

lenormf:
Mostly implemented, closing.

PR #2866, opened 2019-04-19
doc: move aliases explanation from README to commands.asciidoc

Delapouite:
Hi

The paragraph about `alias` occupies quite a larger place than more core topics on the README.
Also it deserves to be more easily accessible through the `:doc`.

I try to add meaningful example to illustrate their benefit complementary to regular commands in light of recent discussion like https://github.com/mawww/kakoune/pull/2805 or https://github.com/mawww/kakoune/issues/2712.


lenormf:
C.f. #2640, I would just remove all the editor-specific documentation from the file.

PR #2877, opened 2019-04-24
Add a new build dependency required for Fedora 30

jkonecny12:
This will fix Fedora 30 & Fedora 31 daily builds.


jkonecny12:
Updated to a new version.

PR #2880, opened 2019-04-25
Add hyperlink support to man plugin

stevenyvr987:
This is a 2nd attempt at it. I've somehow made the first attempt disappear, when I meant to keep it and its discussion. Sorry.

I've organized the changes into three commits.

1. First, I re-implement PR #1927. The original PR lowered the arity of man-impl command, but I do not change it. But it achieves the same thing, which is to create a new buffer for each new man page. A buffer name will look like, for example, _\*man 1 apropos\*_.

1. The original `man` command did not show the stderr text, but I've exposed it as the error message, because sometimes the specific text, for example, _No manual entry for..._, is more informative than generic text.

1. The final commit is the one defining new commands for navigating between man page links and jumping to a man page. See the commit message for details.

### DIFF
--- a/rc/tools/man.kak
+++ b/rc/tools/man.kak
@@ -26,8 +26,9 @@ define-command -hidden -params 2..3 man-impl %{ evaluate-commands %sh{
     buffer_name="$1"
     shift
     manout=$(mktemp "${TMPDIR:-/tmp}"/kak-man-XXXXXX)
+    manerr=$(mktemp "${TMPDIR:-/tmp}"/kak-man-XXXXXX)
     colout=$(mktemp "${TMPDIR:-/tmp}"/kak-man-XXXXXX)
-    MANWIDTH=${kak_window_width} man "$@" > $manout 2>/dev/null
+    MANWIDTH=${kak_window_width} man "$@" > "$manout" 2> "$manerr"
     retval=$?
     col -b -x > ${colout} < ${manout}
     rm ${manout}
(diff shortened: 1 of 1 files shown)


rc/tools/man.kak
```diff
@@ -64,5 +77,71 @@ The page can be a word, or a word directly followed by a section number between
...
+
+# Define a useful command sequence for searching a given regex
+# and a given sequence of search keys.
+define-command man-search -params 2 %{
+  set-register / %arg[1]
+  execute-keys %arg[2]
+}
+
+define-command -docstring 'Go to next man page link' \
+man-link-next %{ man-search %opt[man_link2] n }
```
mawww: I am a bit sceptical that we need all those features in man.kak. Does anybody really iterates on links inside a man page ? I would expect that to be extremely rare, and users only to follow links they voluntarily select.

rc/tools/man.kak
```diff
@@ -64,5 +77,71 @@ The page can be a word, or a word directly followed by a section number between
...
+define-command man-link %{ evaluate-commands -save-regs / %{
+  man-search %opt[man_link1] s<ret>
+}} -hidden
+
+define-command -docstring 'Try to jump to a man page' \
+man-jump %{
+  try %{ man-link } catch %{
+  try %{ man-link-here } catch %{
+    fail 'Not a valid man page link'
+  }}
```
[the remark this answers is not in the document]
stevenyvr987: I've force-pushed a change that uses the `try...catch...catch` pattern. (Wasn't clear from reading the docs that `try` could do this.)

rc/tools/man.kak
stevenyvr987: Some man pages list a lot of useful man page links in their SEE ALSO section that one would like to follow. `man-link-next` is convenient for quickly navigating to the link one wants to focus on, because the command preloads register / and then one can use `n` or `<alt-n>`. Otherwise, one needs to use arrow keys or do an ad-hoc search.

In my recent force-push, I've made `man-link-here` a hidden command and omitted it from the suggested key maps. It was mainly exposed for testing, but it's probably not directly useful.

PR #2882, opened 2019-04-26
Automatic insert completion using ctags

mogenson:
I noticed that the `ctags-complete` command wasn't working for me because the output didn't conform to the [completions type](https://github.com/mawww/kakoune/blob/master/doc/pages/options.asciidoc#types) and it wasn't using the tags files set in the `ctagsfiles` option.

This PR attempts to fixup `ctags-complete` and add some autocomplete features found in other insert completer scripts. It provides the following changes:

1. Fixup ctags-complete command to search for partial matches for the current selection in all tags files listed in `ctagsfiles` option. Format the results to fit the Kakoune completions type.

2. Add ctags-enable-autocomplete and ctags-disable-autocomplete commands to add and remove an `InsertIdle` hook to select the previous word and call `ctags-complete`.

3. Use a new `ctags_min_chars` option to limit the noise of returning many completions for very short selections. Since Kakoune runs insert completers in order and stops after one returns a result, having a buffer word completion for one character selection is probably more helpful then every symbol that begins with that character. The default value is 3.

4. Check for 'ctags' and 'readtags' dependencies on load and print a warning to the debug buffer like the other tools scripts do. Recommend the installation of universal-ctags. Check for the presence of tags files in `ctagsfiles` list when enabling autocomplete.

Please try out these changes and leave any feedback you have about this implementation. Best regards.

### DIFF
--- a/rc/tools/ctags.kak
+++ b/rc/tools/ctags.kak
@@ -1,10 +1,14 @@
-# Kakoune Exuberant CTags support script
+# Kakoune CTags support script
 #
-# This script requires the readtags command available in ctags source but
-# not installed by default
+# This script requires the readtags command available in universal-ctags
+
+declare-option -docstring "minimum characters before triggering autocomplete" \
+    int ctags_min_chars 3
 
 declare-option -docstring "list of paths to tag files to parse when looking up a symbol" \
     str-list ctagsfiles 'tags'
+
+declare-option -hidden completions ctags_completions
 
 define-command -params ..1 \
     -shell-script-candidates %{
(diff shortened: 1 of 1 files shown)


mogenson:
Hello.

I'm just following up. Please let me know if these changes are suitable to be merged.

Thanks.

rc/tools/ctags.kak
```diff
@@ -121,3 +140,28 @@ define-command ctags-update-tags -docstring 'Update tags for the given file' %{
         printf %s\\n "evaluate-commands -client $kak_client echo -markup '{Information}${msg}'" | kak -p ${kak_session}
     ) > /dev/null 2>&1 < /dev/null & }
 }
+
+define-command ctags-enable-autocomplete -docstring "Enable automatic ctags completion" %{
+    evaluate-commands %sh{
```
mawww: Not sure that we need to check that, I think ctagsfiles should be tolerant to absent files, its more a set of files to look for, missing ones should not be an error.

mogenson:
Here's an amended PR without dependency checks or empty $kak_opt_ctagsfiles check.

The thought was to mirror the functionality of the other completion scripts and to provide feedback to the user as to why their completions might not be working. The two most common problems are missing readtags/ctags dependencies and a missing tags file or a tags file not in the location specified by $kak_opt_ctagsfiles.

However, if other scripts are moving towards the direction of being less verbose and not polluting the debug buffer with warnings, then these checks can be removed from ctags.kak.

Thank you for the review.

mogenson:
@mawww Is there anything else this PR needs?

PR #2886, opened 2019-04-30
rc/filetype/elm.kak: Automatically configure the :format command.

Screwtapello:
"elm-format" is the defacto auto-formatter for Elm, advertised [in the official Elm tutorial](https://github.com/evancz/guide.elm-lang.org/blob/master/book/install.md#configure-your-editor).

I'm not sure how opinionated Kakoune should be about setting up default configuration for languages and tools, but in this instance it seems there's only one formatting tool for Elm, and it doesn't appear to have any configuration options, so it should be uncontroversial.

### DIFF
--- a/rc/filetype/elm.kak
+++ b/rc/filetype/elm.kak
@@ -16,8 +16,12 @@ hook global WinSetOption filetype=elm %{
 
     hook window ModeChange insert:.* -group elm-trim-indent  elm-trim-indent
     hook window InsertChar \n -group elm-indent elm-indent-on-new-line
+    set-option window formatcmd %opt{elm_formatcmd}
 
-    hook -once -always window WinSetOption filetype=.* %{ remove-hooks window elm-.+ }
+    hook -once -always window WinSetOption filetype=.* %{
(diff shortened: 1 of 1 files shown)


mawww:
I think this introduces a dependency onto format.kak, as the formatcmd option would not have been declared if that was not loaded. It should either require it (but is that reasonable), or add the option setting hook inside a ModuleLoad hook for format.kak.

Screwtapello:
I guessed it might, but `format.kak` doesn't actually declare a module. It's just an option and a command, so I guess there was no point converting it to module format, and hence it's always loaded.

laelath:
It might be a bit inconsistent, but yeah, I decided to leave the default tools out of modules in the initial implementation. Though maybe something like
```
hook -once global KakBegin .* %{
    require-module format
}
```
would be reasonable.

andreyorst:
What about empty module in tool scripts? Just so other could detect that module was loaded?

Screwtapello:
In #3193 I think the consensus was that we don't want to set `formatcmd` by default, even if there's a sensible default we could use.

PR #2910, opened 2019-05-16
Fix Rust autocomplete

### LINKED ISSUE #2909: racer.kak autocomplete broken
image

Creating standard project with cargo and trying to create String variable produces this output in debug buffer:

~~~
error running command 'evaluate-commands -client 'client0' %{
                set-option buffer=main.rs racer_completions %@3.9@@18:vec!|info -style menu %!Creates a [`Vec`] containing the arguments.

`vec!!` allows `Vec`s to be defined with the same syntax as array expressions.
There are two forms of this macro\:

- Create a [`Vec`] containing a given list of elements\:

```
let v = vec!![1, 2, 3];
assert_eq!!(v[0], 1);
assert_eq!!(v[1], 2);
assert_eq!!(v[2], 3);
```

- Create a [`Vec`] from a given element and size\:

```
let v = vec!![1; 3];
assert_eq!!(v, [1, 1, 1]);
```

Note that unlike array expressions this syntax supports all elements
which implement [`Clone`] and the number of elements doesn\'t have to be
a constant.

This will use `clone` to duplicate an expression, so one should be careful
using this with types having a nonstandard `Clone` implementation. For
example, `vec!![Rc\:\:new(1); 5]` will create a vector of five references
to the same boxed integer value, not five references pointing to independently
boxed integers.

[`Vec`]\: ../std/vec/struct.Vec.html
[`Clone`]\: ../std/clone/trait.Clone.html!|{default+F}vec!{default+d} macro_rules! vec {@
            }
': parse error: 0:37: unterminated string '%{...}'
~~~

andreyorst:
Fix #2909

### DIFF
--- a/rc/tools/rust/racer.kak
+++ b/rc/tools/rust/racer.kak
@@ -26,55 +26,15 @@ define-command racer-complete -docstring "Complete the current selection with ra
                 }
                 /^MATCH/ {
                     word = $2
-                    type = $7
                     desc = substr($9, 2, length($9) - 2)
                     gsub(/\|/, "\\|", desc)
                     gsub(/\\n/, "\n", desc)
                     gsub(/!/, "!!", desc)
-                    menu = $8
-                    sub(/^pub /, "", menu)
-                    gsub(/\|/, "\\|", menu)
-                    if (type == "Function") {
-                        sub(word, "{default+F}" word "{default+d}", menu)
-                        gsub(/^fn /, "    fn ", menu)                            # The extra spaces are there to vertically align
-                                                                                 # the type of element on the menu to make it easy
-                                                                                 # to read
-                        menu = "{default+d}" menu
-                    } else if (type == "Enum") {
-                        menu = substr(menu, 0, length(menu) - 2)
-                        sub(word, "{default+F}" word "{default+d}", menu)
-                        gsub(/^enum /, "  enum ", menu)                          # The extra spaces are there to vertically align
-                                                                                 # the type of element on the menu to make it easy
-                                                                                 # to read
-                    } else if (type == "Module") {
-                        if (length(menu) > 30) {                                 # The "menu" bit (as returned by racer),
-                                                                                 # contains the path to the source file
-                                                                                 # containing the module...
+                    info = $8
+                    gsub(/\|/, "\\|", info)
 
-                          menu = substr(menu, length(menu) - 29, 30)             # ... trimming it, so the completion menu
-                                                                                 # doesn''t get distorted if it''s too long
-                        }
-                        menu = "   mod {default+F}" word "{default+d}   .." menu # The extra spaces are there to vertically align
-                                                                                 # the type of element on the menu to make it easy
-                                                                                 # to read
-                    } else if (type == "Trait") {
-                        sub(word, "{default+F}" word "{default+d}", menu)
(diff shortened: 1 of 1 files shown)


andreyorst:
Also, I kinda understand that such completion looks fancy:

image

But to me, it is not really convenient. I can't see from the first glance what candidate I may actually want to match, and this way I constantly need to keep in mind that only bold part is inserted. What if I simplify it to look like more classic completion:

image

It will also hugely simplify the `awk` part of the script:

```diff
diff --git a/rc/tools/rust/racer.kak b/rc/tools/rust/racer.kak
index 3df0b418..d784dadd 100644
--- a/rc/tools/rust/racer.kak
+++ b/rc/tools/rust/racer.kak
@@ -34,46 +34,9 @@ define-command racer-complete -docstring "Complete the current selection with ra
                     menu = $8
                     sub(/^pub /, "", menu)
                     gsub(/\|/, "\\|", menu)
-                    if (type == "Function") {
-                        sub(word, "{default+F}" word "{default+d}", menu)
-                        gsub(/^fn /, "    fn ", menu)                            # The extra spaces are there to vertically align
-                                                                                 # the type of element on the menu to make it easy
-                                                                                 # to read
-                        menu = "{default+d}" menu
-                    } else if (type == "Enum") {
-                        menu = substr(menu, 0, length(menu) - 2)
-                        sub(word, "{default+F}" word "{default+d}", menu)
-                        gsub(/^enum /, "  enum ", menu)                          # The extra spaces are there to vertically align
-                                                                                 # the type of element on the menu to make it easy
-                                                                                 # to read
-                    } else if (type == "Module") {
-                        if (length(menu) > 30) {                                 # The "menu" bit (as returned by racer),
-                                                                                 # contains the path to the source file
-                                                                                 # containing the module...
 
-                          menu = substr(menu, length(menu) - 29, 30)             # ... trimming it, so the completion menu
-                                                                                 # doesn''t get distorted if it''s too long
-                        }
-                        menu = "   mod {default+F}" word "{default+d}   .." menu # The extra spaces are there to vertically align
-                                                                                 # the type of element on the menu to make it easy
-                                                                                 # to read
-                    } else if (type == "Trait") {
-                        sub(word, "{default+F}" word "{default+d}", menu)
-                        gsub(/^trait /, " trait ", menu)                         # The extra spaces are there to vertically align
-                                                                                 # the type of element on the menu to make it easy
-                                                                                 # to read
-                    } else if (type == "Type") {
-                        sub(word, "{default+F}" word "{default+d}", menu)
-                        gsub(/^type /, "  type ", menu)                          # The extra spaces are there to vertically align
-                                                                                 # the type of element on the menu to make it easy
-                                                                                 # to read
-                    } else if (type == "Struct") {
-                        sub(word, "{default+F}" word "{default+d}", menu)        # Struct doesn''t have extra spaces because it''s
-                                                                                 # the longest keyword
-                    } else {
-                        menu = "{default+F}" word "{default+d} " menu
-                    }
-                    candidate = word "|info -style menu %!" desc "!|" menu
+                    candidate = word "|info -style menu %!" desc "!|" word " {MenuInfo}" menu
+
                     gsub(/:/, "\\:", candidate)
                     gsub(/@/, "@@", candidate)
                     gsub(/~/, "~~", candidate)
```

PR #2915, opened 2019-05-21
Add -Og for debug builds

eraserhd:
On my system, some optimizations are on by default (NixOS), resulting in
variables being optimized out on debug builds.  It *seems to be*
something about a "_FORTIFY_SOURCE" feature?  In any case, `-Og` is
documented as "Optimize debugging experience".

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -9,7 +9,7 @@ ifneq ($(gzip_man),yes)
 endif
 
 ifeq ($(debug),yes)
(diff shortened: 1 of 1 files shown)


mawww:
I decided to give that a go, if its too painful to work with we will switch back to -O0 (but explicitely this time)

lenormf:
`-Og` is not a pre-processing flag, it should be appended to `CXXFLAGS`.

mawww:
Correct, will fix that.

PR #2917, opened 2019-05-22
Protocol fix and a refactor

eraserhd:
The refactor is extracting a variadic send_message() helper in RemoteUI
that avoids some repetition.

EDIT: The below is a non-issue.  I failed to remember that `read_availble()` won't read more than a packet, so none of the below is necessary and the commit was removed from the PR.

The fix is about MsgReader segmentation.  I haven't directly observed a
problem with it, but it seemed like it might give us trouble later.
Basically, if more than one message is in MsgReader's buffer when
reset() is called, the additional messages (or parts of messages) were
discarded.  This could lead to a corrupted stream.

### DIFF
--- a/src/remote.cc
+++ b/src/remote.cc
@@ -50,8 +50,8 @@ public:
     MsgWriter(RemoteBuffer& buffer, MessageType type)
         : m_buffer{buffer}, m_start{(uint32_t)buffer.size()}
     {
-        write(type);
-        write((uint32_t)0); // message size, to be patched on write
+        write_field(type);
+        write_field((uint32_t)0); // message size, to be patched on write
     }
 
     ~MsgWriter()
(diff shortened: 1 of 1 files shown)


eraserhd:
It occurs to me that it's also possible for the receiver to get stuck if two messages are buffered, since at most one message is processed for a read event.

eraserhd:
(The receiver getting stuck was actually a non-issue, too.)

src/remote.cc
```diff
@@ -143,6 +143,15 @@ class MsgWriter
...
 
+    inline void write_all() {}
+
+    template<typename Arg1, typename ...Args>
+    void write_all(Arg1&& arg1, Args&&... args)
+    {
+        write(std::forward<Arg1>(arg1));
+        write_all(std::forward<Args>(args)...);
+    }
+
```
mawww: We have c++17 fold expressions !
```
template<typename... Args>
void write_all(Args&&... args)
{
    (write(std::forward<Args>(args)), ...);
}
```

Also, maybe we could make this method just called `write`, and make all the others private with another name (write_impl or so).

eraserhd: Oh, that's neat.  I'm apparently still not yet caught up :D 

I made the internal methods `write_field` and `write_raw`.

PR #2918, opened 2019-05-23
Highlight python code in bash

JJK96:
When using something like 

```
python -c "
print(5+3)
"
```

In bash, the python code should be highlighted accordingly, this pull request accomplishes this.

I'm not sure how to solve the dependency on python yet


andreyorst:
Will that work if user uses `"` in the code? I mean that when shell expansions highlighted as `sh` in kakscript, they use reqursive matching on `{` in order to have consistent highlighting even if user puts bunch of balanced `{` `}` in it. This isn't possible to have recursive matching on non pair character like double quote, e.g. this will fail:

```
python -c "print(\"abc\")"
                  ^
                  python highlighting ends here.

```

Using your code shows what I mean:
image

> (I know this because I've tried to do the same thing for perl)

occivink:
@andreyorst I think the end marker can be found with `(?<!\\)(\\\\)*"` or so ( == `"`, optionally preceded by an even number of `\\`)

andreyorst:
It seems that my example was wrong. Because region is highlighted properly, it's strings that have problems here. If I escape quotes with `\\` the highlighting is correct. I wonder why I have to escape them with double slash though, since escaping with single works out correctly.

JJK96:
It might have to do with how the python highlighting works, maybe it does not consider `\"` as the start of a string, while it does consider `\\"` as one.

andreyorst:
> It might have to do with how the python highlighting works, maybe it does not consider \" as the start of a string, while it does consider \\" as one.

@JJK96 It seems so. Sorry, I was mislead with the example as shown below with Perl, but it is completely different case, so my initial comment has nothing to do with your implementation. 

> I think the end marker can be found with `(?<!\\)(\\\\)*"` or so ( == `"`, optionally preceded by an even number of `\\`)

@occivink Maybe you know how to handle this Perl case? The problem is here:
image

when `perl -e '` meets first `'` (red one) the region effectively ends, but then it continues when we meet another `'` however from that quote it is treated as shell string (which is correct, since we pass shell string to `perl`, but not correct in term of highlighting).

rc/filetype/sh.kak
```diff
@@ -16,6 +16,15 @@ provide-module sh %[
 
 add-highlighter shared/sh regions
 add-highlighter shared/sh/code default-region group
+
+# Languages
+try %{
+require-module python
+add-highlighter shared/sh/python region 'python.*?-c *"' '[^\\]"' regions
```
occivink: the `python.*-c` could potentially yield an undesired match, maybe `python[^\n]* -c"` instead?

JJK96: That does not really solve the problem of undesired matches, only limits it. 
It also limits the normal usage, i.e. you can no longer do
```
python \
-c "
...
"
```

andreyorst: -c flag also accepts string in single quotes, so maybe use `['"]` here too?

occivink:
@andreyorst thinking about it, I'm not sure if we can ever make this work properly. The code within `perl -e '...'` is only valid perl *after* it's been parsed by the shell. If we try to highlight it as-is with the perl highlighter it will quickly be confused by the shell syntax, such as `'\''` and the string concatenation.  For the particular problem of snippets.kak I think we can safely extract the perl code to a separate file 

The same issue applies to python, as soon as there is a `\"` inside the string the python highlighter will be confused, and string concatenation will make things worse.
I'm not sure if this can of worm is worth opening.

mawww:
It would work for here documents with no escaping (the `<<'foo'` kind), as in those we know the data is passed verbatim, but I am not sure it is a good enough use case to warrant "opening the can of worms" as @occivink said.

PR #2919, opened 2019-05-24
Sh indent on new line

sjjf:
This doesn't attempt to be optimal or even efficient, but it makes
kakoune considerably more usable for me editing shell scripts.

### DIFF
--- a/rc/filetype/sh.kak
+++ b/rc/filetype/sh.kak
@@ -89,10 +89,15 @@ define-command -hidden sh-indent-on-new-line %{
         # deindent after ;;
         try %{ execute-keys -draft <space> k <a-x> <a-k> \;\;$ <ret> j <a-lt> }
 
-        # function indent
-        try %= execute-keys -draft <space> k <a-x> <a-k> \{$ <ret> j <a-gt> =
-        # deindent at end of function
-        try %= execute-keys -draft <space> k <a-x> <a-k> \}$ <ret> <a-lt> j K <a-&> =
+        # '{}' delimited compound command
+        # Note that in this context the '{' and '}' characters are reserved
+        # words, and hence must be surrounded by white space (including a
+        # newline). Without this, any of the various forms of parameter
+        # expansion that use { and } delimiters will match, for example ${foo},
+        # or brace expansion e.g. thing{1,2}
+        try %= execute-keys -draft <space> k <a-x> <a-k> (\s|^)\{$ <ret> j <a-gt> =
+        # deindent closing } when in reserved word context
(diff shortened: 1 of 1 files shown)


rc/filetype/sh.kak
```diff
@@ -48,4 +52,49 @@ add-highlighter shared/sh/code/function regex ^\h*(\w+)\h*\(\) 1:function
...
+    evaluate-commands -draft -itersel %{
+        # copy '#' comment prefix and following white spaces
+        try %{ execute-keys -draft k <a-x> s ^\h*\K#\h* <ret> y gh j P }
+        # preserve previous line indent
+        try %{ execute-keys -draft \; K <a-&> }
+        # filter previous line
+        try %{ execute-keys -draft k : sh-trim-indent <ret> }
+
+        # indent after do
+        try %{ execute-keys -draft <space> k x <a-k> do$ <ret> j <a-gt> }
```
mawww: it should probably be `<a-x>` here as `x` will jump to next line if the previous line is fully selected (which will be the case if it is empty). If the next line (the one after the inserted newline) ends with `do`, we will end-up wrongly indenting the following line.

sjjf: I'm struggling to visualise what you're talking about - I can't see how the `<ret> k x` combination can ever select the line /after/ the inserted newline? If I understand correctly `<ret>` inserts a newline and moves the selection to that line, `k` moves up to the previous line, and then `x` selects that line. If the previous line (moved to using `k`) is empty `x` will skip to the next line and select that, whereas `<a-x>`won't, but since we've just gone back up a line the next line that `x` can skip to has to be the brand new empty line created by the `<ret>` keystroke, at which point the `<a-k> do$` will fail with an empty result, which is what we want and expect.

Unless I'm just missing something, or misunderstanding something - I'm obviously not the expert on Kakoune ;-)

I can definitely see why it's more correct to use `<a-x>`, though, since it explicitly does what we want rather than effectively relying on the side effect of the other keystrokes, and I'll change the code to use it in all these cases.

mawww: Imagine the following buffer:
```
do
```
Now imagine your cursor is on the `d` of `do`. You do `i<ret>`, now your buffer state is:
```

do
```
with the cursor on `d`, `kx` ends up selecting the `do\n` line, and `<a-k>` does not fail.

sjjf: Ahhh, yes, thank you for explaining.

sjjf:
Closing this because I've created a (hopefully) better alternative.

Issue #2920, opened 2019-05-24
markdown.kak code highlighting broken


andreyorst:
Because languages now provide modules to allow lazy loading, highlighing is broken for markdown:
![image](https://user-images.githubusercontent.com/19470159/58331712-5fb2b780-7e42-11e9-8029-c5a9c425f330.png)

Manually requiring `c-family` module fixes this issue:

![image](https://user-images.githubusercontent.com/19470159/58331729-6e996a00-7e42-11e9-9b1b-d4f281ae6416.png)

Since [markdown.kak defines highlighters for many languages](https://github.com/mawww/kakoune/blob/9ffde03ca74ce691882c47776fe8d014792dff4c/rc/filetype/markdown.kak#L36-L49) we need a solution to require those modules when needed, or provide separate modules for highlighting. Maybe there's another solution.

PR #2924, opened 2019-05-24
[WIP] Markdown code highlighting fix

laelath:
WIP fix of #2920. reStructuredText is also broken, but it's syntax is more complicated.

Added a hook that detects if the buffer contains a code block, then attempts to load the highlighting for that code block. Currently works for most languages, but not the ones where the module is named differently from the markdown name. Notably C, C++, ObjC, and Typescript. 

I can think of three ways to fix this. Create a mapping of language names to module names in Markdown/reStructuredText/Org-mode, ensure there's only one language defined per module (though we wouldn't be able to do things like accept both 'c++' and 'cpp'), or add a kind of module alias system.

### DIFF
--- a/rc/filetype/c-family.kak
+++ b/rc/filetype/c-family.kak
@@ -1,3 +1,4 @@
+# Detection
 hook global BufCreate .*\.(cc|cpp|cxx|C|hh|hpp|hxx|H)$ %{
     set-option buffer filetype cpp
 }
@@ -57,6 +58,11 @@ hook -group objc-highlight global WinSetOption filetype=objc %{
     add-highlighter window/objc ref objc
     hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/objc }
 }
+
+# Module aliases
+provide-module c %{ require-module c-family }
+provide-module cpp %{ require-module c-family }
+provide-module objc %{ require-module c-family }
 
 
 provide-module c-family %§
--- a/rc/filetype/c-family.kak
+++ b/rc/filetype/c-family.kak
@@ -59,11 +59,6 @@ hook -group objc-highlight global WinSetOption filetype=objc %{
     hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/objc }
 }
 
-# Module aliases
-provide-module c %{ require-module c-family }
-provide-module cpp %{ require-module c-family }
-provide-module objc %{ require-module c-family }
-
 
 provide-module c-family %§
 
@@ -450,3 +445,8 @@ define-command objc-alternative-file -docstring "Jump to the alternate objc file
 }
 
 §
+
+# Module aliases
+provide-module c %{ require-module c-family }
+provide-module cpp %{ require-module c-family }
+provide-module objc %{ require-module c-family }
--- a/rc/filetype/javascript.kak
+++ b/rc/filetype/javascript.kak
@@ -34,6 +34,9 @@ hook -group typescript-highlight global WinSetOption filetype=typescript %{
     hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/typescript }
 }
 
+# Aliases
+# ‾‾‾‾‾‾‾
+provide-module typescript %{ require-module javascript }
 
 provide-module javascript %§
 
--- a/rc/filetype/javascript.kak
+++ b/rc/filetype/javascript.kak
@@ -34,9 +34,6 @@ hook -group typescript-highlight global WinSetOption filetype=typescript %{
     hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/typescript }
 }
 
-# Aliases
-# ‾‾‾‾‾‾‾
-provide-module typescript %{ require-module javascript }
 
 provide-module javascript %§
 
@@ -133,3 +130,7 @@ add-highlighter shared/typescript/code/ regex \b(array|boolean|date|number|objec
 add-highlighter shared/typescript/code/ regex \b(as|constructor|declare|enum|from|implements|interface|module|namespace|package|private|protected|public|readonly|static|type)\b 0:keyword
 
 §
+
+# Aliases
+# ‾‾‾‾‾‾‾
+provide-module typescript %{ require-module javascript }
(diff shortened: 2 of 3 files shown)


andreyorst:
we can pass the `$kak_selection` to shellscript which will use `case $kak_selection in` statement, where we can `c|cpp|c++|obj-c|objc) printf "%s\n" "require-module c-family"`.

But the problem is that this hook is being run only at file open and if empty markdown file is opened/created or user inserted code block for the language that wasn't mentioned before this solution will fail.

Proper way could be to detect whether user inserts thre backticks and a language, then call this hook but it is a complex solution that requires a way to detect when actually do parsing. One possible way could be to detect if user inserted ``` then add a hook to run when user hits enter, but I already see a problem if user creates newline with <kbd>o</kbd>.

I guess there's currently no way to properly detect all edge cases.

laelath:
Also keep in mind cases like pasting code in, which may be uncommon, but should still work. I guess ideally the hook should be on both `NormalIdle` and `InsertIdle`.

andreyorst:
Consider using asynchronous shell call for such hook. Something like this:

    define-command -override load-languages %{ nop %sh{ (
        printf "%s\n" "execute-keys -draft -client $kak_client \%s\`\`\`\h*\K[^\s]+<ret><dquote>ay" | kak -p $kak_session
        eval "set -- $kak_reg_a"
        while [ $# -gt 0 ]; do
            case $1 in
                c|cpp|c++|objc) printf "%s\n" "evaluate-commands -client $kak_client %{ require-module c-family }" ;;
                lisp|emacs-lisp) printf "%s\n" "evaluate-commands -client $kak_client %{ require-module lisp }" ;;
            esac
            printf "%s\n" "evaluate-commands -client $kak_client %{ echo -debug %{$1} }"
            shift
        done | kak -p $kak_session
    ) > /dev/null 2>&1 < /dev/null & }}

    hook global NormalIdle .* load-languages
    hook global InsertIdle .* load-languages

It overrides contents of register `a` which isn't good though.

andreyorst:
This shoud do it:

    define-command load-languages %{ evaluate-commands -save-regs 'a' %{
        try %{
            execute-keys -draft '%s```\h*\K[^\s]+<ret>"ay'
            nop %sh{ (
                eval "set -- $kak_reg_a"
                while [ $# -gt 0 ]; do
                    case $1 in
                        c|cpp|c++|objc) module="c-family" ;;
                        lisp|emacs-lisp) module="lisp" ;;
                    esac
                    [ -n "$module" ] && printf "%s\n" "evaluate-commands -client $kak_client %{ require-module $module }"
                    module=
                    shift
                done | kak -p $kak_session
            ) > /dev/null 2>&1 < /dev/null & }}
        }
    }

    hook buffer NormalIdle .* load-languages
    hook buffer InsertIdle .* load-languages

laelath:
I feel like there should be a way to do this without a shell invocation, I think there could be something using a `str-to-str-map`. I'll look into it.

mawww:
I was thinking, maybe it would make sense to go the other way around, add a `markdown-add-filetype-highlighting <filetype> <highlighter>` command in markdown.kak, then have a `ModuleLoaded` hook in each filetype.kak that would register themselves ?

And maybe to fix the full issue, implement user hooks, so that when a language reference appears markdown.kak can trigger `UserHook markdown-nested-filetype=<lang>` and relevant support script can `require-module` itself and add its filetype highlighting.

What I dont like about that direction is that we disseminate knowledge of markdown in every support file.

An alternate direction could be to add at the end of c-family.kak (and similar modules where the module name does not match the filetype) the following:

```
provide-module c %{ require-module c-family }
provide-module cpp %{ require-module c-family }
provide-module objc %{ require-module c-family }
```

This would be pretty nice because it makes c-family just an implementation detail, every external code should just use the c/cpp/objc module, and it would allow to keep markdown support only inside markdown.kak.

laelath:
I like the second approach, especially since it doesn't require any kind of universal data structure.

mawww:
Yeah, I think that would actually be a clean and nice solution, seems we only need to add the 3 c-family modules, and a typescript module inside javascript.kak.

rc/filetype/c-family.kak
```diff
@@ -58,6 +59,11 @@ hook -group objc-highlight global WinSetOption filetype=objc %{
     hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/objc }
 }
 
+# Module aliases
+provide-module c %{ require-module c-family }
+provide-module cpp %{ require-module c-family }
+provide-module objc %{ require-module c-family }
+
```
mawww: I think I'd rather have those *after* the module they reference, so that they dont get declared if declaring the c-family module fails for some reason.

andreyorst:
I guess I missed something. How exactly would a module be loaded when user inserts

    ```cpp

into their markdown document?

Like this?

```kak
define-command markdown-require-highlighters %{ evaluate-commands -save-regs 'a' %{ try %{
    execute-keys -draft 'gtGbGls```\h*\K[^\s]+<ret>"ay'
    evaluate-commands "try %{ require-module %reg{a} } catch %{ fail %{unknown language '%reg{a}' }"
}
hook buffer NormalIdle .* load-languages
hook buffer InsertIdle .* load-languages
```

The only thing I think isn't great here is that if someone wants to add language to Kakoune via external package, they need to put such "aliases" for most of markup languages, that may also be supported via external packages, in order to be consistent. Maybe I'm too paranoid here though.

laelath:
@andreyorst I already added it! I don't think it's that much of a problem, since it's just the module defining some level of its internal representation.

andreyorst:
> @andreyorst I already added it! I don't think it's that much of a problem, since it's just the module defining some level of its internal representation.

I mean, for example, let's imagine a hypothetical markup language that has it's own names for the languages. For example:

```c
#source lang-C
int foo(void);
#end-source
```

```rust
#source lang-Rust
fn foo() {}
#end-source
```
And we decide to add this markup language to Kakoune. What should we do? Should we add new module aliases to all scripts to represent such source blocks, thus making Kakoune know all supported markup languages? Or those aliases should be added to the markup script, thus making it know about all Kakoune supported languages?

mawww:
@andreyorst Nope we would not add additional language modules for those markup languages, the existing problem we had here is that we had a c-family modules providing both c/cpp/objc languages. it makes sense to change that as an implementation detail and introduce a c, cpp, and objc module. This happens to fix the markdown issue, but is a reasonable change in its own right (as is adding a typescript modules that is only forwarding to the javascript one).

If a markup language happens to agree on language naming (and its usually the case, except for a few language like c++/cpp and d/dlang, the name of the language is pretty much the same everywhere), then things are simple, if it uses something else, then it will need to maintain some kind of mapping, but we dont have to worry about that for now.

PR #2925, opened 2019-05-25
Add test/ ruby-alternative-file implementation

greenfork:
Minitest is the default testing framework for Ruby which is also shipped by default with Ruby on Rails. This PR adds support for Minitest along with already implemented RSpec.


mawww:
Thanks !

PR #2927, opened 2019-05-27
Filesystem-based registers

eraserhd:
This is a an experiment in exposing (global, static) registers via the filesystem.  I've left it as two commits:

### Make session a directory

This makes the paths at `/tmp/kakoune/$USER/*` into directories instead of plain sockets, and moves the sockets to `/tmp/kakount/$USER/$kak_session/socket`, setting the stage for storing more session-based state on the filesystem.

### Store registers on the filesystem

This stores registers which are not context-dependent (e.g. not dynamic) on the filesystem.  The files are written when set and read when accessed.  There isn't any noticeable slowness.

## Many random, disorganized thoughts

The idea here was to see if Kakoune could lean more heavily on POSIX without the complexity of implementing 9P, and it seems to be mostly true.  Shell expansions (and integrations outside of Kakoune) can use this to read and update registers without bumping into environment-variable limits and without forking, and without any subtle shell quoting issues.

To me, this seems more reminiscent of how `git` or `mh` work.

A significant portion of Kakoune's state could be handled in this way: Options, highlighters, selections, current window offsets, hooks, and keymaps.  Unfortunately, buffer contents would probably not perform well.

Each register is a directory, and each value is one file in that directory.  I chose this layout so that it could be possible to symlink "main" to the main value in the future, and this wouldn't require shell scripts to keep "reg" and "main_reg" values in sync.  This property, that the layout of the files makes it difficult to create an incoherent state, seems important for this approach (not so much for the 9P approach, though, since we can validate writes and dynamically generate file contents).

If necessary, it should be possible to cache file contents in memory for efficiency, only reading them from disk when directory or file mtimes change.  I've heard that `git` has a number of tricks like this to make things fast, though the only one I remember is using a directory's hard-link count, which doesn't seem useful here.

This approach does not preclude the 9P approach.  A 9P or FUSE filesystem could potentially be mounted inside the session directory, say at `mnt`, and symlinks added to the session tree to make things like buffer contents appear in the right place.

Perhaps options and dynamic registers can be broken into scope directories: "global", "buffers", and "windows", with symlinks back to "more global" values, so that shell scripts only have to read from "windows/client_id/opt/tabstops", and would get the global value if it had not been set on the buffer or window?  This might make setting values a bit more complicated.   Managing scopes like this would make declaring a new option an `O(buffers + windows)` operation, which might be slow.

Possibly the real test of this approach would be large `range-specs` options?


Screwtapello:
Some random, disorganised thoughts of my own:

- This will definitely need changes to the registers documentation
- Perhaps there could/should be some expansion that expands to the filesystem path of a register? Something like `%reg{x,path}` or `$kak_reg_x_path`?
- This should make it easier to preserve state between sessions! This isn't as useful as it might sound, though, because when a register is used as a mark, it identifies the mark's buffer by bufname. That means that in another session, jumping to that mark might jump to the wrong file, or might not work at all if we can't figure out what real file corresponds to a give bufname.

lenormf:
It seems to be conflicting/redundant with the way internals would be exposed by the editor (can't find the issue).

Writing to the filesystem every time a register changes doesn't sound good at all, not only when the editor executes scripts at startup but also during an editing session.

You should rather maintain your own fork that makes use of FUSE properly, I'm afraid this compromise is not a good idea.

Screwtapello:
Actually, that's another interesting question - what happens with things like `exec -save-regs`? Do we get a new directory with symlinks to the existing register files? Do the old register contents get stored in memory and written back out when the context is over?

> Writing to the filesystem every time a register changes doesn't sound good at all

If it's fast enough, why not do the more flexible thing?

eraserhd:
@Screwtapello I would hope that, instead of adding expansions for each file, we chose predictable file locations and document *those*.

@lenormf The idea that writing files to the filesystem is "too slow" is a very familiar idea to me.  It's been one of those heuristics I've internalized over a long time, and I'm worried that this idea is actually an antique relic, no longer supported today - hence this experiment.  I had 48K of RAM when I started programming, and I still get stuck often making "optimizations" which aren't warranted and which make software more complex.

A story:  In like 2010, I wrote some code that loaded an XML file in Java.  A buddy of mine happened to work at the same company, and I showed him the code for some reason.  And he was like, "Why did you use the SAX parser?"  And I was like, "Uh, well gosh, the file could be *megabytes*..."

I have no problem ditching this if it *is* too slow, but my 9P branch has about 900 lines of new code without handling write operations, while this branch has about 250.  FUSE implementations, while they seem to all speak the same protocol, have operating-system specific ways of mounting.  plan9port's `9pfuse` driver handles all that nicely, and allows me to implement this with a much simpler protocol, but at the cost of a needing a new software dependency.  Up until now, a POSIX-compliant system was sufficient for running Kakoune, and this would require POSIX+FUSE+plan9port.  Perhaps @mawww is OK with that, but I've been assuming that he'd not make that decision lightly.

I'm going to dogfood this branch today during work and see if I notice *anything* (so far I haven't).

@Screwtapello I haven't changed anything with `-save-regs`, so the values are read, temporarily stored in memory, then unconditionally rewritten to the filesystem right now.  I can think of two ways to make this better (again, *if necessary*): renaming register contents files and renaming them back to avoid a write, and avoiding rewriting if the mtime or size hasn't changed.

eraserhd:
I'm closing this and focusing on maintaining the 9P branch.

PR #2928, opened 2019-05-27
Basic language support for Awk

dgmulf:
Some syntax highlighting, static completion, and auto-indentation.


Screwtapello:
Man, how did we go this long without having an Awk highlighter?

One thing you might want to do is put the code that actually creates the highlighters into a `provide-module awk` command, and have the hooks that install those highlighters into the current buffer use `require-module awk` — a good example to follow is [the CSS highlighter](https://github.com/mawww/kakoune/blob/master/rc/filetype/css.kak). This means Kakoune can avoid executing any of that code until it's actually needed, which helps Kakoune start up quickly.

Also, some file formats can embed snippets of other file formats, like Markdown's fenced code blocks. You might want to add awk to [the list of embeddable formats](https://github.com/mawww/kakoune/blob/master/rc/filetype/markdown.kak#L36) so that awk content is highlighted there too. ReStructuredText also has [the same kind of thing](https://github.com/mawww/kakoune/blob/master/rc/filetype/markdown.kak#L36).

laelath:
Adding modules kind of broke embedded snippets (the relevant module needs to get manually loaded). We've been working on a fix with #2924.

andreyorst:
> You might want to add awk to [Markdown's list of embeddable formats](https://github.com/mawww/kakoune/blob/master/rc/filetype/markdown.kak#L36) so that awk content is highlighted there too. ReStructuredText also has [the same kind of thing](https://github.com/mawww/kakoune/blob/master/rc/filetype/markdown.kak#L36).

My Org Mode PR also has this. And in current state the support is broken (https://github.com/mawww/kakoune/issues/2920). Currently we developed asynchronous call with huge case statement, but that's not portable, and external language plugins can't benefit from it. I think we need some system to make this thing automatic even for external language scripts. Maybe language scripts should modify global option which holds all and every available language modules with respective language mapped to it. But there's another problem. We have to invoke shell in order to import highlighting on basically every buffer change. If something like this were available:

>It would be nice if there were a way to expand an element of a hash map rather than the whole set. Something like %opt{str_to_str_map_option|map_key}.

We could create a universal system for all language scripts and, all markup scripts will be able to support all languages automatically.

mawww:
Hello,

could you rebase your branch on master, without merge commits ?

Cheers.

dgmulf:
I'll close this and open a new pull request, since I moved the changes to a separate feature branch. Sorry, still learning how to use git! :)

PR #2939, opened 2019-06-04
insert_final_newline support for editorconfig

ath3:
Added insert_final_newline support to editorconfig.kak


TeddyDD:
I already tried that in #2448

Kakoune always adds newline at the end of the file and there is no way to disable that so this setting is useless for us.

ath3:
Yes, i can see it now when opening files in another editor, thanks for the clarification!

PR #2942, opened 2019-06-05
Avoid using the 'setsid' tool

codesoap:
I think I have [found a portable solution](https://www.reddit.com/r/openbsd/comments/b191w3/differences_in_shell_scripting_between_openbsd/eikc8mt?utm_source=share&utm_medium=web2x) for #2520: Using perl's `setsid`.

Tested on OpenBSD 6.5 and Xubuntu 18.04.

I'm not sure how the licensing works here. Should I ask [u/kernigh](https://www.reddit.com/user/Kernigh/) for permission to use his idea for Kakoune?


eraserhd:
On Mac OS, I don't have `setsid`.  On NixOS, I don't have `perl`.  And since we want to retain POSIX compatibility, we should probably avoid `perl`.

POSIX does have the `setsid()` library call, though.  We could, perhaps, ship a `kak-setsid.c` utility?

codesoap:
`kak-setsid` sounds like a good idea to me. I'll investigate how I could implement such a tool, later. Thanks for your input!

codesoap:
I have added the `kak-setsid` tool, which is now used as a replacement for the `setsid` tool.

To avoid licensing problems I've written the tool from scratch, but my experience in writing C Unix tools is very limited. Someone should probably proofread `kak-setsid.c`.

Tested on OpenBSD 6.5 and Xubuntu 18.04.

src/kak-setsid.c
```diff
@@ -0,0 +1,37 @@
...
+            return 3;  /* execute_command_in_arguments() should not return. */
+        default:
+            break;
+    }
+}
+
+void execute_command_in_arguments(int argc, char *argv[])
+{
+    int i;
+    char *command[argc];
```
eraserhd: huh, variable-sized automatic arrays got into C99?  Neat.

eraserhd: According to [this](https://retrocomputing.stackexchange.com/questions/5179/why-historically-include-the-number-of-arguments-argc-as-a-parameter-of-main), ANSI C already NULL-terminates the argv array.  This means that, instead of creating `command` and copying to it, you can just call `execvp(argv[1], argv + 1)`.

src/kak-setsid.c
```diff
@@ -0,0 +1,37 @@
...
+{
+    if(argc < 2) {
+        fprintf(stderr, "Give a command to execute as argument to %s.\n",
+                argv[0]);
+        return 1;
+    }
+    
+    switch(fork()) {
+        case -1:
+            fprintf(stderr, "Failed to create child process.\n");
```
eraserhd: Use `perror("kak-setsid: fork");` here, as that will explain why fork failed.

src/kak-setsid.c
```diff
@@ -0,0 +1,37 @@
...
+void execute_command_in_arguments(int argc, char *argv[])
+{
+    int i;
+    char *command[argc];
+    command[argc-1] = NULL;
+    for(i=0; i<(argc-1); i++) {
+        command[i] = argv[i+1];
+    }
+    execvp(command[0], command);
+    fprintf(stderr, "The given command could not be executed.\n");
```
eraserhd: `perror("kak-setsid: execvp");`

src/kak-setsid.c
```diff
@@ -0,0 +1,37 @@
...
+    switch(fork()) {
+        case -1:
+            fprintf(stderr, "Failed to create child process.\n");
+            return 2;
+        case 0:
+            setsid();
+            execute_command_in_arguments(argc, argv);
+            return 3;  /* execute_command_in_arguments() should not return. */
+        default:
+            break;
```
eraserhd: So the second and third things I've learned about C99 today is that you don't need a return statement in main() (it will return 0) and that you can rely on return statements in main() instead of needing to call exit()!  I have apparently been out of touch.

However, I think it's still a good idea to `return 0;` here to be explicit.

src/kak-setsid.c
codesoap: Nice, that let me make the code much more compact.

src/kak-setsid.c
codesoap: I didn't know about `perror()`. Thanks for help!

src/kak-setsid.c
codesoap: Done.

lenormf:
Shipping small binaries that could help work around portability issues has been considered in the past, but decided against in the end.

We don't want the project to deploy random programs that are redundant with binaries already implemented in well-known packages (for instance, `setsid` is available in `util-linux`), and we also don't want to have to maintain them if they already are, by the community. Just use homebrew to install the `util-linux` package to fulfil the dependency (I think OpenBSD has a port as well).

As for Perl, it's not much better since some distributions don't ship it by default (assuming your one-liner is portable across versions of Perl), so users would end up installing that dependency anyway. The `setsid` binary is installed by default in ArchLinux, so it seems to be the same problem, just with another package.

codesoap:
I couldn't find a port of `setsid` or `util-linux` on OpenBSD, but I guess I can just put the perl oneliner into `$HOME/bin/setsid`.

I still think the current behaviour is irritating, because there is no indication of a problem when trying to use the `:new` command, when `setsid` is missing. What do you think about adding a hint about this dependency to `README.asciidoc`? Alternatively I could try to make sure something is printed to `*debug*`, if setsid is missing.

lenormf:
I don't think we can remove the `>/dev/null 2>&1` from the calls to `setsid` because otherwise the process might block its parent (could be wrong, but I think it's [this shell pattern](https://github.com/mawww/kakoune/blob/master/doc/writing_scripts.asciidoc#running-a-process-in-the-background) we're relying on).

I agree that all programs that strictly do not rely on muting the standard error stream should leave it alone to allow errors to be printed in `*debug*`. Several scripts in the codebase currently do not follow this guideline, which isn't great ([example](https://github.com/mawww/kakoune/blob/master/rc/tools/git.kak#L191,L206)).

I've tried making dependency detection a default for built-in scripts, c.f. #1726, but that effort was denied, so if OpenBSD really doesn't have any package providing `setsid` (I thought a user told me it was in the ports), you're probably stuck with using the Perl one-liner or the C file you wrote, locally.

src/Makefile
```diff
@@ -88,8 +88,9 @@ else
 endif
 
 CXXFLAGS += -pedantic -std=c++17 -g -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare -Wno-address
+CCFLAGS += -pedantic -std=c99 -g -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare -Wno-address
```
lenormf: I think the standard variable name is `CFLAGS` (no doubling of the `C`).

lenormf: You probably don't need all those flags, I reckon `-g -Wall -Wextra` is enough.

src/Makefile
```diff
@@ -171,13 +176,15 @@ endif
...
 	rm -rf $(bindir)/kak \
+		$(bindir)/kak-setsid
 		$(sharedir) \
 		$(docdir) \
 		$(mandir)/kak.1.gz \
 		$(mandir)/kak.1
 
 .PHONY: check TAGS clean dist distclean installdirs install install-strip uninstall
-.PHONY: tags test man kak FORCE
+.PHONY: tags test man kak kak-setsid FORCE
```
lenormf: The `kak-setsid` is an actual file-target, so it probably shouldn't be declared as being phony.

codesoap:
Thanks for the additional review!

> I thought a user told me it was in the ports

You may be referring to [this comment](https://github.com/mawww/kakoune/issues/2520#issuecomment-433044665), but @t6 was using FreeBSD there.

You are correct, removing the `2>&1` blocks the parent. I don't think there is anything more I can do here. I'm closing the PR.

codesoap:
I've added https://github.com/mawww/kakoune/wiki/OpenBSD .

PR #2948, opened 2019-06-06
Fix emission of BufReadFifo events

### LINKED ISSUE #2946: BufReadFifo hook parameter is incorrect on first read
##### `foo.kak`
```kak
hook global BufCreate '\*stdin\*' %{
    hook buffer BufReadFifo .* %{echo -debug BufReadFifo: %val{hook_param}}
    evaluate-commands -draft %{
        execute-keys '%'
        echo -debug BufCreate: %val{selection_desc}
    }
}
```

```sh
$ ( echo hello; sleep 10; echo goodbye ) | kak -n -E foo.kak -debug hooks
```

### Outcome

`*debug*`:
```
hook BufCreate(*stdin*)/
BufCreate: 1.1,1.1
hook BufReadFifo(2.1,2.1)/
BufReadFifo: 2.1,2.1
hook BufReadFifo(2.1,3.1)/
BufReadFifo: 2.1,3.1
```

### Expected

The first `BufReadFifo` should start at 1.1.

eraserhd:
The hook parameter should not be adjusted for the prevention of scrolling.  Also, ensure that the last BufReadFifo is triggered if we encounter an error or EOF after appending some data to the buffer.

Closes #2946

### DIFF
--- a/src/buffer_utils.cc
+++ b/src/buffer_utils.cc
@@ -129,17 +129,18 @@ Buffer* create_fifo_buffer(String name, int fd, Buffer::Flags flags, bool scroll
         // iteration number allows us to go back go back to the event loop and
         // handle other events sources (such as input)
         constexpr size_t max_loop = 16;
+        bool closed = false;
         size_t loop = 0;
         char data[buffer_size];
-        BufferCoord insert_coord;
+        BufferCoord insert_coord = buffer->back_coord();
         const int fifo = watcher.fd();
         do
         {
             const ssize_t count = ::read(fifo, data, buffer_size);
             if (count <= 0)
             {
-                buffer->values().erase(fifo_watcher_id); // will delete this
-                return;
+                closed = true;
+                break;
             }
 
             auto pos = buffer->back_coord();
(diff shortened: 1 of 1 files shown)


eraserhd:
Since I made this PR, I've been wondering why not just multiply `buffer_size` by 16, remove `loop` and `max_loop`, and iterate just once?

src/buffer_utils.cc
```diff
@@ -131,24 +131,22 @@ Buffer* create_fifo_buffer(String name, int fd, Buffer::Flags flags, bool scroll
...
+        BufferCoord insert_coord = buffer->back_coord();
         const int fifo = watcher.fd();
         do
         {
             const ssize_t count = ::read(fifo, data, buffer_size);
             if (count <= 0)
             {
                 buffer->values().erase(fifo_watcher_id); // will delete this
-                return;
+                break;
```
mawww: This is a use after free, the comment means that the object pointer by this is deleted after the erase call, so in order to do that, this erase call needs to be postponed to the end of the function, as accessing 'buffer` will be reading into a dead lambda object.

mawww:
Allocating 32kb on the stack might be a bit much, although that might be pessimistic.

src/buffer_utils.cc
eraserhd: Oh, I see... the captures are part of the lambda object, not on the stack... right.

Updated.

PR #2951, opened 2019-06-07
Explain shell

ath3:
New command explain-shell for shell files

It uses online api to get the explanation to the selected commands,
if possible, showing the results in docsclient.


ath3:
Done!

The plugin can now be found at https://github.com/ath3/explain-shell.kak

PR #2954, opened 2019-06-07
Add support for autoformatting after do keyword in Ruby

greenfork:
In Ruby `do` keyword is at the end of line contrary to `while`, `if` etc. which are at the start.

Currently this PR implements inserting spaces and inserting an `end` keyword. But it fails to insert `end` keyword when below there's a structure with `do`. Example:

```ruby
setup do<cursor here>

setup do
 
end
```

after hitting `<ret>` it inserts spaces but does not insert `end` keyword.

This has to do with this line:

`execute-keys -draft Ge <a-K>\A(^\n|^<c-r>x(?!begin)(?!case)(?!class)(?!def)(?!for)(?!if)(?!module)(?!unless)(?!until)(?!while)[^\n]*\n)+<c-r>xend$<ret>`

I tried adding `(?<!do)` but miserably failed:

`execute-keys -draft Ge <a-K>\A(^\n|^<c-r>x(?!begin)(?!case)(?!class)(?!def)(?!for)(?!if)(?!module)(?!unless)(?!until)(?!while)[^\n]*(?<!do)\n)+<c-r>xend$<ret>`

I don't understand it and would love to get some help with this.


greenfork:
This PR should be ready.

I borrowed this approach `execute-keys -draft }i J <a-x> <a-K> ^<c-r>xend$<ret>` from `lua.kak`. It makes `ruby-insert-on-new-line` dependent on `ruby-indent-on-new-line` but it seems reasonable provided how much simpler and faster it is.

I will test it at work for a couple of days and report if any issues arise.

greenfork:
Tested at work, works fine.

@mawww please review.

PR #2955, opened 2019-06-09
Fix tools using fifos for OpenBSD

codesoap:
Without these changes, kak would hang on the corresponding commands, displaying a 'waiting for shell command to finish' message.

I guess OpenBSD's `sh` behaves differently with the `&` control operator. The man pages for `sh` on OpenBSD and Linux don't really talk about putting whole command lists into the background, so I'd say there is no right or wrong here.

I tested the proposed solution on OpenBSD 6.5 and Xubuntu 18.04. I did not test the `clang`, `lint` and `jedi` commands in depth, because I did't yet set up the tools for it, but the commands don't block kak anymore.

I'm not sure if the `(...)  > /dev/null 2>&1 < /dev/null &` pattern should be adapted in all other places as well. The tools using fifos were just the first time I experienced problems with it.

### DIFF
--- a/rc/tools/clang.kak
+++ b/rc/tools/clang.kak
@@ -36,7 +36,7 @@ The syntaxic errors detected during parsing are shown when auto-diagnostics are 
         # not hang while clang is running. As completions references a cursor
         # position and a buffer timestamp, only valid completions should be
         # displayed.
-        {(
+        ((
             case ${kak_opt_filetype} in
                 c) ft=c ;;
                 cpp) ft=c++ ;;
@@ -100,7 +100,7 @@ The syntaxic errors detected during parsing are shown when auto-diagnostics are 
 
             printf %s\\n "set-option 'buffer=${kak_buffile}' clang_flags ${kak_timestamp} ${flags}
                   set-option 'buffer=${kak_buffile}' clang_errors ${kak_timestamp} ${errors}" | kak -p ${kak_session}
-        ) & } > /dev/null 2>&1 < /dev/null
+        ) & ) > /dev/null 2>&1 < /dev/null
     }
 }
 
--- a/rc/tools/clang.kak
+++ b/rc/tools/clang.kak
@@ -36,7 +36,7 @@ The syntaxic errors detected during parsing are shown when auto-diagnostics are 
         # not hang while clang is running. As completions references a cursor
         # position and a buffer timestamp, only valid completions should be
         # displayed.
-        {(
+        {{
             case ${kak_opt_filetype} in
                 c) ft=c ;;
                 cpp) ft=c++ ;;
@@ -100,7 +100,7 @@ The syntaxic errors detected during parsing are shown when auto-diagnostics are 
 
             printf %s\\n "set-option 'buffer=${kak_buffile}' clang_flags ${kak_timestamp} ${flags}
                   set-option 'buffer=${kak_buffile}' clang_errors ${kak_timestamp} ${errors}" | kak -p ${kak_session}
-        ) & } > /dev/null 2>&1 < /dev/null
+        } & } > /dev/null 2>&1 < /dev/null
     }
 }
 
--- a/rc/tools/clang.kak
+++ b/rc/tools/clang.kak
@@ -36,7 +36,7 @@ The syntaxic errors detected during parsing are shown when auto-diagnostics are 
         # not hang while clang is running. As completions references a cursor
         # position and a buffer timestamp, only valid completions should be
         # displayed.
-        (
+        {(
             case ${kak_opt_filetype} in
                 c) ft=c ;;
                 cpp) ft=c++ ;;
@@ -100,7 +100,7 @@ The syntaxic errors detected during parsing are shown when auto-diagnostics are 
 
             printf %s\\n "set-option 'buffer=${kak_buffile}' clang_flags ${kak_timestamp} ${flags}
                   set-option 'buffer=${kak_buffile}' clang_errors ${kak_timestamp} ${errors}" | kak -p ${kak_session}
-        ) > /dev/null 2>&1 < /dev/null &
+        ) & } > /dev/null 2>&1 < /dev/null
     }
 }
 
--- a/rc/tools/clang.kak
+++ b/rc/tools/clang.kak
@@ -36,7 +36,7 @@ The syntaxic errors detected during parsing are shown when auto-diagnostics are 
         # not hang while clang is running. As completions references a cursor
         # position and a buffer timestamp, only valid completions should be
         # displayed.
-        {{
+        {(
             case ${kak_opt_filetype} in
                 c) ft=c ;;
                 cpp) ft=c++ ;;
@@ -100,7 +100,7 @@ The syntaxic errors detected during parsing are shown when auto-diagnostics are 
 
             printf %s\\n "set-option 'buffer=${kak_buffile}' clang_flags ${kak_timestamp} ${flags}
                   set-option 'buffer=${kak_buffile}' clang_errors ${kak_timestamp} ${errors}" | kak -p ${kak_session}
-        } & } > /dev/null 2>&1 < /dev/null
+        ) & } > /dev/null 2>&1 < /dev/null
     }
 }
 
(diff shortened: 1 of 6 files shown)


lenormf:
Weird, what shell is the default on OpenBSD?

Screwtapello:
https://man.openbsd.org/sh says:

> This version of sh is actually ksh in disguise. As such, it also supports the features described in ksh(1). This manual page describes only the parts relevant to a POSIX compliant sh. If portability is a concern, use only those features described in this page.

https://man.openbsd.org/ksh.1 says:

> This shell is based on the public domain 7th edition Bourne shell clone by Charles Forsyth and parts of the BRL shell by Doug A. Gwyn, Doug Kingston, Ron Natalie, Arnold Robbins, Lou Salkind, and others.

So, `pdksh`, I guess.

As to the original question, [the POSIX Shell Command Language section 2.9](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_09) defines a "command" as a simple command, a pipeline, a "List compound-list" (commands joined with `&&` or `||`), a compound command (a subshell, or things like `if`/`fi`), or a function definition. Later, in [section 2.9.3](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_09_03_02) it says:

> If a command is terminated by the control operator <ampersand> ( '`&`' ), the shell shall execute the command asynchronously in a subshell.

...so I'd expect that applying `&` to any kind of command should work. Indeed, `bash` and `dash` both seem quite happy for me to apply `&` to a function definition, never mind the other more sensible command types.

That said, even if this is a POSIX compatibility bug in OpenBSD, I don't know how much the OpenBSD folk care about such things. And even if they do care, it'll be a while before a fix makes it into the next official release, and another while before that release is widely deployed, so the practical solution is just to patch Kakoune, I guess.

If we had  a "portability guidelines for Kakoune scripts document", this would be a good thing to put into it.

codesoap:
Thanks for the research, @Screwtapello! I didn't know about the 'POSIX Shell Command Language' documentation.

[Chapter 2.9.4](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_09_04) made me realize, that this pull request isn't even POSIX conform right now: I need to put a delimiting operator before `}`.

Besides that, reading the documentation lead me to further investigations and I got some interesting results. This shell script works on OpenBSD:
```bash
#!/usr/bin/env sh

output=$(mktemp -d "${TMPDIR:-/tmp}"/test-git.XXXXXXXX)/fifo
mkfifo ${output}
( git status > ${output} 2>&1 ) > /dev/null 2>&1 < /dev/null &
kak -E "evaluate-commands %{edit! -fifo ${output} *git*}"
```

Another interesting thing I found is, that this patch makes `:git status` non blocking. Of course I get to see an empty `*git*` buffer in `kak`, but the blocking is gone, and the fifo seems to work:
```diff
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -46,6 +46,7 @@ Available commands:\n  add\n  rm\n  blame\n  commit\n  checkout\n
diff\n  hide-
         output=$(mktemp -d "${TMPDIR:-/tmp}"/kak-git.XXXXXXXX)/fifo
         mkfifo ${output}
         ( git "$@" > ${output} 2>&1 ) > /dev/null 2>&1 < /dev/null &
+        cat ${output} > /dev/null 2>&1

         printf %s "evaluate-commands -try-client '$kak_opt_docsclient' %{
                   edit! -fifo ${output} *git*
```
 
Unfortunately these findings just make me more confused. I'll try to investigate further...

codesoap:
I think I've gotten closer to the problem now. This works on Linux, but not OpenBSD:
```
#!/usr/bin/env sh

fill_fifo() {
    mkfifo testfifo
    ( git status > testfifo 2>&1 ) > /dev/null 2>&1 < /dev/null &
    printf %s "evaluate-commands %{edit! -fifo testfifo *git*}"
}
kak -e "$(fill_fifo)"
```

Screwtapello:
> Another interesting thing I found is, that this patch makes `:git status` non blocking. 

The new `cat` instance drains all the data that `git` writes into the fifo, so `git` exits, `cat` gets EOF and exits too, and there's nothing for Kakoune to wait for.

codesoap:
> > Another interesting thing I found is, that this patch makes `:git status` non blocking.
> 
> The new `cat` instance drains all the data that `git` writes into the fifo, so `git` exits, `cat` gets EOF and exits too, and there's nothing for Kakoune to wait for.

The thing I wanted to point out there was, that apparently the `(...) > /dev/null 2>&1 < /dev/null &` pattern is not the problem, since `cat` can be reached and even read the fifo. As stated in the later comment, there may be a problem with fifo and shell scopes present in stead.

codesoap:
Debugging tip: The `mksh` I installed on Xubuntu 18.04 seems to behave similarly to the OpenBSD `sh`.

codesoap:
I [wrote to an OpenBSD mailing list](https://marc.info/?t=156033336000003&r=1&w=2) about this issue and [Philip Guenther reassured me](https://marc.info/?l=openbsd-misc&m=156039950105624&w=2), that the behaviour of the OpenBSD `sh` is buggy. Thus I think there is not much more I can do. I think my proposed fix should solve the issue until the OpenBSD `sh` is repaired.

lenormf:
Was the bug fixed upstream? [Our documentation](https://github.com/mawww/kakoune/blob/master/doc/writing_scripts.asciidoc#running-a-process-in-the-background) refers to the code as it was pre-PR, should it be changed, or the PR's changed reverted?

codesoap:
I tested again and found that the bug still persists in OpenBSD 6.7. Back when I asked on the OpenBSD mailing list, [Philip Guenther wrote](https://marc.info/?l=openbsd-misc&m=156039950105624&w=2):

> Sorry, I don't have a fix in my back pocket.

I think the OpenBSD community often goes by the policy "If it annoys you, feel free to submit a patch.", but I'm not familiar enough with the code of `sh` to do this. I guess this means the bug will be there until someone more knowledgeable than me is annoyed by it.

lenormf:
Thanks!

Does the pattern that works on OpenBSD have to use parenthesis, or braces work fine as well?

```
{ foo & } >/dev/null 2>&1 </dev/null
# or does it have to be
( foo & ) >/dev/null 2>&1 </dev/null
```

And it looks like we might need to group commands in a second subshell in some cases?

```
{ { foo; bar; baz; } & } >/dev/null 2>&1 </dev/null
```

I'm trying to figure out what pattern is guaranteed to be working in any case.

codesoap:
All the above patterns seem to work fine on OpenBSD's `sh`. If I remember correctly, the problem really only occurs when FIFOs are involved. Take a look at [my mail](https://marc.info/?l=openbsd-misc&m=156033323516331&w=2) to find a minimal working example:

```
mkfifo 'testfifo'
cat "$(
    ( printf 'foo\n' > testfifo 2>&1 ) > /dev/null 2>&1 &
    printf 'testfifo'
)"
```

It seems like `mksh` has the same bug as the OpenBSD `sh`, so you can use it on Linux to test for yourself.

lenormf:
Sorry for asking you directly, my IP was range banned by marc.info :)

So the only thing that changes is that commands that write to a named pipe using a shell redirection must have their dedicated subshell?

codesoap:
> Sorry for asking you directly, my IP was range banned by marc.info :)

No problem.

> So the only thing that changes is that commands that write to a named pipe using a shell redirection must have their dedicated subshell?

I'm not 100% sure. I think it's more like: When you have a process group, that writes to an "outer" FIFO and you put that in the background, then you must use another subshell that ends after the `&`.

Note that this must be a subshell. `{}` won't work because there seems to be a bug with those in `bash`, `dash`, ... E.g. this does work with OpenBSD's `sh`, but not with `bash`:

```
mkfifo 'testfifo'
cat "$(
    { ( printf 'foo\n' > testfifo 2>&1 ) & } > /dev/null 2>&1
    printf 'testfifo'
)"
```

codesoap:
Oh and since you can't view the OpenBSD mail list, this is the script that seems to work everywhere:

```
mkfifo 'testfifo'
cat "$(
    ( ( printf 'foo\n' > testfifo 2>&1 ) & ) > /dev/null 2>&1
    printf 'testfifo'
)"
```

PR #2963, opened 2019-06-12
Editorconfig max line length

ath3:
Support for max_line_length property
https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#max_line_length

If the property is set, then a highlighter will be added after the max_line_length column, and autowrapping gets enabled.

I tried to make it unobtrusive, but still apparent that a rule had been set for the line length, so it wont do potentially breaking rewrapping of files or paragraphs.


mawww:
This introduces a dependency on autowrap.kak, so the commands will fail if it is not loaded, but as those are the last commands run, I guess it is ok.

Issue #2973, opened 2019-06-15
X11 REPL broken


alexherbo2:
The window title does not change to `kak_repl_window`.

https://github.com/mawww/kakoune/blob/master/rc/windowing/repl/x11.kak

Screwtapello:
On line 18, the plugin calls `%opt{termcmd}` with `-t kak_repl_window`, presumably to set the window title.

On my system:

- `gnome-terminal` doesn't support setting the window title
- `xterm` wants `-title` or `-T` for setting the window title
- `urxvt` also wants `-title` or `-T`
- `pterm` wants `-title`
- `qterminal` doesn't support setting the window title
- `terminology` wants `-T` or `--title`
- `kitty` wants `-T` or `--title`
- `alacritty` [wants `-t` or `-title`](https://github.com/jwilm/alacritty/blob/6cd0f12efb7664bb99a4b79732b8f4f63e42b033/alacritty/src/cli.rs#L119)

It doesn't look like any single option is universally understood.

alexherbo2:
Even with Alacritty it doesn’t work.

mawww:
It is probably more compatible to emit an escape sequence to set the title than to try to find a common switch.

codesoap:
This works for me (`bspwm` and `st`):

```bash
setsid ${kak_opt_termcmd} ${cmd} < /dev/null > /dev/null 2>&1 &
sleep 0.5
xdotool getactivewindow set_window --name kak_repl_window
```

Seems rather hacky, though, because it creates a race condition with the `sleep` and relies on the window manager to 'activate' the newly created window. Maybe someone can manage to improve this...

alexherbo2:
@codesoap Does it work to put `xdotool` in the `${cmd}`?

codesoap:
Apparently I was too tired yesterday to see this option. Thanks for pushing me towards a solution, @alexherbo2 ! I think #2981 should be considerably more portable than the current `x11-repl`.

lenormf:
This issue was fixed in merged PR #2981, we can now close it.

mreppen:
Would it make sense to append `$kak_session` to `kak_repl_window`? I can't test the X11 code itself (on Wayland), but it seems to be limited to one repl per X11 server. One per kak session would be more natural to me. Alternatively, there could be an option like `repl_id` (with default `kak_repl_window_$kak_session`) keeping track of associated repl names, allowing e.g. per-buffer repls.

PR #2981, opened 2019-06-18
Set the REPL window title using xdotool

codesoap:
As mentioned in issue #2973, the -t argument for setting the terminal title was not very portable.

This relies on the window manager to 'activate' the newly created window, but I haven't met one that doesn't.

### DIFF
--- a/rc/windowing/repl/x11.kak
+++ b/rc/windowing/repl/x11.kak
@@ -15,9 +15,8 @@ All optional parameters are forwarded to the new window} \
            exit
         fi
         if [ $# -eq 0 ]; then cmd="${SHELL:-sh}"; else cmd="$@"; fi
-        setsid ${kak_opt_termcmd} \
-               "xdotool getactivewindow set_window --name kak_repl_window \
-                && exec \"${cmd}\"" < /dev/null > /dev/null 2>&1 &
+        setsid ${kak_opt_termcmd} "printf '\e]2;kak_repl_window\a' \
+                && ${cmd}" < /dev/null > /dev/null 2>&1 &
 }}
 
 define-command x11-send-text -docstring "send the selected text to the repl window" %{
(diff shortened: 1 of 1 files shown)


rc/windowing/repl/x11.kak
```diff
@@ -15,7 +15,9 @@ All optional parameters are forwarded to the new window} \
            exit
         fi
         if [ $# -eq 0 ]; then cmd="${SHELL:-sh}"; else cmd="$@"; fi
-        setsid ${kak_opt_termcmd} ${cmd} -t kak_repl_window < /dev/null > /dev/null 2>&1 &
+        setsid ${kak_opt_termcmd} \
+               "xdotool getactivewindow set_window --name kak_repl_window \
+                && exec \"${cmd}\"" < /dev/null > /dev/null 2>&1 &
```
Screwtapello: How about instead of using xdotool, changing `${cmd}` to something like:

    "${SHELL:-sh}" -c "printf '\e]2;kak_repl_window\a'; \"${cmd}\""

codesoap: I didn'd know about this, TIL. This would be the same as `printf "$(tput tsl)kak_repl_window\a; \"${cmd}\""`, right? [This serverfault thread](https://serverfault.com/questions/23978/how-can-one-set-a-terminals-title-with-the-tput-command/23998#23998) would suggest, that your solution would rely on the terminal to support these capabilities. It worked for me on OpenBSD using the `st` terminal, though.

Im not sure which solution is more portable.

Screwtapello: The `tsl` terminfo string is supposed to take the cursor to a particular column of the status line. For example, `tput tsl 5` should take you to the sixth column, so any text you print would leave the first five columns undisturbed. Using `tsl` and `fsl` to represent the window title is common enough to be provided as the `xterm+sl` terminal component, but it's kind of a hack, so no official terminal descriptions use it - it's just there for end-users to use if they want.

That string has been the "set the window title string" sequence for decades, though, so I'd be surprised if there were any GUI terminal emulators that didn't support it. Even tmux supports it, these days.

codesoap: Alright, since the original command you suggested works even with my somewhat unusual `st` terminal, I'll believe you.

This solution seems cleaner then, since it doesn't rely on the window manager to activate the new window. I'll update the pull request.

PR #2983, opened 2019-06-20
Add support for basic auto-indenting of shell code.

sjjf:
This attempts to support a simple formatting and intentation style for
plain sh syntax (and other sh-compatible code which doesn't stray too
far from portable sh).

The complexity of sh syntax means that we have to be opinionated -
attempting to be more flexible would require extensive context
awareness, and would require something more akin to a proper
autoformatting tool or a language server.

The formatting style used here makes use of vertical whitespace as the
primary delimiter, so that code ends up looking like this:

```
if [ $foo = "bar" ]; then
        thing1
else
        thing2
fi

for i in foo bar baz; do
        thing1
        thing2
done

case "$foo" in
        bar) thing1;;
        baz)
                thing1
                thing2
                ;;
esac
```

Since the formatting style used is very opinionated the `sh_auto_indent`
option can be used to enable and disable auto-indentation, with the default set to
`no`.


mawww:
I dont think we need the sh_auto_indent option, the usual way to disable autoindentation is to set the `disabled_hooks` option to a value that will match the hook group.

sjjf:
I thought I was being so clever . . . ;-)

I've pushed an additional commit that removes the `sh_auto_indent` option and simply sets up the hook unconditionally.

mawww:
This is fine as it is, but I would be even happier if you could provide a few tests for that, take a look at `test/indent/...` for examples.

If you do not have the time to do that, just tell me and I'll merge that as-is.

Thanks for that work

sjjf:
I'm sure it counts as overkill, but I've added a very comprehensive set of tests - it's a nice simple test system once you figure out the wrinkles (like how to get consistent whitespace).

PR #2987, opened 2019-06-24
Make sure hooked function keys are normalized

lenormf:


src/commands.cc
```diff
@@ -988,12 +988,25 @@ const CommandDesc add_hook_cmd = {
...
+        const Vector<Hook> key_hooks = {Hook::InsertKey, Hook::NormalKey, Hook::RawKey};
+        if (contains(key_hooks, it->value))
+        {
+            const auto keys = parse_keys(keydesc);// throws a runtime_error on parsing errors
+
+            if (keys.size() != 1)
+                throw runtime_error{format("a single key description is expected, got {}", keys.size())};
+
+            if (keys[0].key >= Key::F1 and keys[0].key <= Key::F12)
+                keydesc = key_to_str(keys[0]);
```
lenormf: Note: I realize this is overkill, as technically all we need is a call to `uppercase()`, however I felt calling `key_to_str()` was more accurate since that's the same function called to normalize keys when they're hit in normal mode.

lenormf:
Actually this won't even work, as things like the following won't work anymore (c.f. failed CI tests):

```
hook global InsertKey '<(?<name>\w+)>' %{ exec "<backspace>%val(hook_param_capture_name)" }
```

lenormf:
Closed in favor of #2988.

Issue #2991, opened 2019-06-25
Rename ModuleLoad hook to ModuleLoaded


codesoap:
When learning about the `ModuleLoad` hook I was uncertain whether the hook runs before or after the specified module is loaded. The documentation says:

> Triggered when a module is evaluated by the first require-module call

which still sounds somewhat ambiguous to me.

After trying it out and thinking for a second it became clear, that the hook is run after the module has been sourced, but maybe renaming the hook to `ModuleLoaded` would prevent any confusion for future readers.

What do you think? I could prepare a pull request if there is a consensus, that this change would be a good idea.

lenormf:
Make a PR anyway!

Delapouite:
To alleviate this ambiguity, we could follow this existing pattern: `BufWritePre` vs `BufWritePost`.
So in this case `ModuleLoad` would become `ModuleLoadPost`

laelath:
@Delapouite I'm not sure having a hook run before a module loads would be too helpful. It was added to have a point where options declared by modules could be known to exist and be changed, which only makes sense after the module is loaded. Unless there's a use case I'm not seeing.

codesoap:
@Delapouite I see, that it would be more consistent to use `ModuleLoadPost`, but after thinking about it for a bit I found that it introduces new ambiguity. If I didn't see `BufWritePre` and `BufWritePost` before, I could imagine the `Post` to be irritating, since it has several meanings in the englisch language.

I have created #2992.

codesoap:
#2992 is merged.

PR #2992, opened 2019-06-25
Rename ModuleLoad hook

codesoap:
With this pull request I intend to clarify the exact time, when the hook is executed. When I first encountered the hook it was unclear to me, whether it's run before or after a module is loaded.

See #2991.

### DIFF
--- a/doc/pages/changelog.asciidoc
+++ b/doc/pages/changelog.asciidoc
@@ -28,8 +28,8 @@ released versions.
   `require-module` commands that allows for lazily loading language
   support files with dependency resolution.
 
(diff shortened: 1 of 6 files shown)


lenormf:
Could you please remove all those empty merge commits, to have just your last one commit on top of `master`?

codesoap:
@lenormf Done. Not sure where they all came from...

mawww:
I dont think I have a copyright waiver commit from you yet, could you add one (as this change is almost non-trivial).

codesoap:
I hadn't come across the `CONTRIBUTING` file yet. I need to read documentation more...

Issue #1733, opened 2017-11-29
Rename to :color-scheme command


alexherbo2:

mawww:
To what, why ?

Delapouite:
I think what Alex had in mind is going from the current `colorscheme` to `color-scheme` (with a dash) which seems to be the proper spelling in english : https://en.wikipedia.org/wiki/Color_scheme 

But if we ever decide to go this road, we should also do this for real homogeneity :
- call it `change-color-scheme` to have a verb in the command
- rename the `/colors` dir to `/color-schemes`

lenormf:
I would rename it to `set-colorscheme`, to keep the name similar to `set-face`.

AprilArcus:
Against this: `color-scheme` is harder to type than `colorscheme` and the matching `vim` command is called `colorscheme` (and uses the `colors` directory). It would be okay to have both spellings, but requiring a hyphen seems like a bad choice.

PR #2995, opened 2019-06-29
rename colorscheme to set-color-scheme and use color-schemes dir

TeddyDD:
`:set-color-scheme` instead `:colorscheme` (alias `color-scheme`)
`color-schemes` directory instead `colors`

Closes: #1733


README.asciidoc
```diff
@@ -679,7 +679,7 @@ Color Schemes
 
 Kakoune ships with some color schemes that are installed to
 `../share/kak/colors/`. If `$XDG_CONFIG_HOME/kak/colors/` is present
```
Screwtapello: You need to change these paths too!

TeddyDD: Good catch

TeddyDD:
umm, failing tests are not good sign :thinking:

TeddyDD:
Fixed

mawww:
I am wondering if `set` is the correct verb here, I guess it makes sense as color schemes are mostly going to run `set-face` commands. Any feedback from the community on this change ?

alexherbo2:
`apply`?

lenormf:
I've commented otherwise in the related issue, but now I'm thinking there's no real benefit to renaming that command.

PR #2997, opened 2019-07-01
src: Use all selections to generate a pattern upon hitting `*`

### LINKED ISSUE #2994: [Suggestion] Make '*' generate an alternation of the selections
Currently `*` and `<a-*>` only consider the main selection, but I think it'd be useful if it created an OR of all the selections, like that `sel1_escaped|sel2_escaped|sel3_escaped|...`
This would enable certain use-cases that are currently hard to perform. such as selecting all instances of multiple identifiers at the same time.

Getting the old main-selection-only behavior would also be trivial to map.

lenormf:
This commit makes the `*` and <a-*> primitives compose a search pattern
comprised of all the current selections, as opposed to only the main one.

All selections are OR'd into the default search register, which makes it
convenient to search for several identifiers already selected.

To retain the old behaviour, the following mappings can be used:

```
map global normal * ': exec -draft -save-regs "" %{<space>*}<ret>'
map global normal <a-*> ': exec -draft -save-regs "" %{<space><a-*>}<ret>'
```

Fixes #2994

### DIFF
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -939,12 +939,20 @@ template<bool smart>
 void use_selection_as_search_pattern(Context& context, NormalParams params)
 {
     const auto& buffer = context.buffer();
-    auto& sel = context.selections().main();
-    const auto beg = sel.min(), end = buffer.char_next(sel.max());
-    String pattern = format("{}{}{}",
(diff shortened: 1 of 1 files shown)


lenormf:
Simplified the pattern generated after @occivink's advice.

PR #3001, opened 2019-07-01
Add ledger filetype

codesoap:
This pull request adds basic support for the [ledger](https://www.ledger-cli.org/) file format. The format is described [here](https://www.ledger-cli.org/3.0/doc/ledger3.html#Journal-Format). I think I covered most of the syntax, but I don't use most of ledgers advanced features myself. This means I didn't implement syntax highlighting for the features I didn't grasp yet.

I left out the syntax, that is described as deprecated or only exists for backwards compatibility.

Since this is the first Kakoune filetype handler I wrote, I copied a lot of the structure from other filetype handlers (mainly `awk.kak`).

Here is an example ledger file and a screenshot to show the syntax highlighting:
```
account Assets
account Expenses
    note This is where my money goes
account Expenses:Food
account Expenses:Indoor Plants

; Comment example

comment
This is a block comment.
end comment

alias food=Expenses:Food

2019-07-01 small cactus
    Expenses:Indoor Plants  8€
    Assets

2019-07-01 bread  ; this is a transaction note
    food  3.50€
    Assets
```

highlighted ledger file

### DIFF
--- a/rc/filetype/ledger.kak
+++ b/rc/filetype/ledger.kak
@@ -130,6 +130,8 @@ define-command -hidden ledger-indent-on-new-line %[
         try %[ execute-keys -draft \; K <a-&> ]
         # cleanup trailing whitespaces from previous line
         try %[ execute-keys -draft k <a-x> s \h+$ <ret> d ]
+        # indent after the first line of a transaction
+        try %[ execute-keys -draft k<a-x> <a-k>^[0-9]<ret> j<a-gt> ]
     ]
 ]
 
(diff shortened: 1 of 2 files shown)


greenfork:
Maybe add

```
set-option window comment_line ';'
```

to window hook? It appears to be a standard comment https://www.ledger-cli.org/3.0/doc/ledger3.html#Commenting-on-your-Journal

codesoap:
Good point, @greenfork! For completeness I would then also add the config for `comment-block`. Block comments wouldn't work right now, so I opened #3017. When it is merged (or denied) I'll add a commit to this pull request.

PR #3006, opened 2019-07-03
fix a few typos

jo-he:


src/input_handler.cc
```diff
@@ -514,7 +514,7 @@ class LineEditor
         else if (key == ctrl('w') or key == alt(Key::Backspace))
             erase_move(&to_prev_word_begin<Word>);
```
jo-he: I suggest to change the default of these bindings:
- `<C-w>` → `<a-h>` so that are `<C-h>` and `<a-h>` are symmetric to `<C-d>` and `<a-d>`
- `<C-W>` → `<C-w>` like in Bash

lenormf: You could create another PR that implements those changes, mentioning that it depends on this one, so has not to mix fixes and changes in the same PR.

Those discrepancies you're seeing stem from the fact that readline-type shortcuts came on top of custom bindings that don't follow the same logic (as far as the bindings themselves).

jo-he:
further fixed added

PR #3012, opened 2019-07-07
allow for mapping the Insert key

jo-he:
enable the use of `<ins>` for custom bindings, e. g.:
```
:map global normal <ins> ';i'
```


mawww:
Thanks !

PR #3014, opened 2019-07-08
Fix *-trim-indent commands showing up in history

### LINKED ISSUE #3013: *-trim-indent commands appear in history after entering insert mode
### Steps 
Create a new Rust file with kakoune, say with `kak azdef.rs`.

Enter insert mode.

Type in `fn () {` followed by a newline.

Then leave insert mode with escape.

Enter command mode with `:`.

### Outcome 

The command `rust-trim-indent` is show as the last executed command in history.

### Expected 

I expect no commands to be shown in history, as I haven't typed any.

As far as I can tell, this happens for all files with similar indent hooks, using `Kakoune v2019.07.01`.

daboross:
This adds an explicit space when executing all `*-trim-indent` commands in `execute-keys`, so that the commands don't show up in history.

I chose using quotes here of `<space>` since it seemed simpler. Perhaps it's less obvious what the purpose is though?

I'm assuming that it's unintentional that these show up in history. If it is, then this could be closed.

Fixes #3013.

### DIFF
--- a/rc/filetype/cabal.kak
+++ b/rc/filetype/cabal.kak
@@ -57,7 +57,7 @@ define-command -hidden cabal-indent-on-new-line %[
         # preserve previous line indent
         try %[ execute-keys -draft \; K <a-&> ]
         # filter previous line
-        try %[ execute-keys -draft k : cabal-trim-indent <ret> ]
+        try %[ execute-keys -draft k : ' cabal-trim-indent' <ret> ]
         # indent after lines ending with { or :
         try %[ execute-keys -draft <space> k <a-x> <a-k> [:{]$ <ret> j <a-gt> ]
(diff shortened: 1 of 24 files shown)


Screwtapello:
I was expecting something like:

    execute-keys -draft k ": cabal-trim-indent<ret>"

...which is usually the pattern I see suggested when people ask on IRC about preventing commands from showing up in history.

Your way works too, though.

daboross:
That's... better, I think. To be honest I really wasn't sure about the best way to do this.

daboross:
Fixed after https://github.com/mawww/kakoune/commit/39445f9acb93b62a57ef1c1f31cd1c290a12a4b3.

PR #3017, opened 2019-07-11
Add newlines with comment-block

codesoap:
I think this looks cleaner and it is needed for l[edger block comments](https://www.ledger-cli.org/3.0/doc/ledger3.html#Commenting-on-your-Journal) to work.

Example block to be commented in c:
```c
int main(void) {
    return 0;
}

void foo(void);
```

Result of running comment-block on the first three lines with the old behaviour:
```c
/*int main(void) {
        return 0;
}
*/
void foo(void);
```

Result of running comment-block on the first three lines with the new behaviour:
```c
/*
int main(void) {
        return 0;
}
*/

void foo(void);
```

### DIFF
--- a/rc/tools/comment.kak
+++ b/rc/tools/comment.kak
@@ -125,15 +125,15 @@ define-command comment-block -docstring '(un)comment selections using block comm
             set-register / "\A\Q%opt{comment_block_begin}\E.*\Q%opt{comment_block_end}\E\n*\z"
             execute-keys "s<ret>"
             # Uncomment it
-            set-register / "\A\Q%opt{comment_block_begin}\E|\Q%opt{comment_block_end}\E\n*\z"
+            set-register / "\A\Q%opt{comment_block_begin}\E\n*|\Q%opt{comment_block_end}\E\n*\z"
             execute-keys s<ret>d
-        } catch %{
+        } catch %{ evaluate-commands %sh{
             # Comment the selection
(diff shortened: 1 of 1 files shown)


lenormf:
Does `:command-block` insert an empty line after the commented data, or is it a discrepancy in your examples?

The command should not assume that entire lines are being commented, the user should still be able to e.g. comment a single argument in a function prototype using block comment characters (`int foo(/*int a*/);`).

codesoap:
> Does :command-block insert an empty line after the commented data, or is it a discrepancy in your examples?

The original `:comment-block` didn't insert any newline. The closing `*/` is on a new line in my example, because I had included the newline character of the third line in my selection.

  > The command should not assume that entire lines are being commented, [...]

Oh, right, I didn't think about that. Maybe then I should rather adapt `comments.kak`, such that special characters like `\n` are interpreted for `comment_block_begin` and `comment_block_end`. This way I could still get the extra newlines for ledger's `comment` and `end comment` lines.

codesoap:
In dae59a18cbefe6041ad4cb6b8be00b8d2eb664ff I reverted the addition of newlines for all filetypes. Instead escape characters are now interpreted through `printf`. The escaping for the ragel filetype is getting somewhat funky, but I think overall it won't change the current behaviour and enables adding block comments, where the opening and / or closing token must be on their own lines.

PR #3018, opened 2019-07-11
Check if file exists before loading editorconfig settings

mogenson:
I noticed when using the recommended [editorconfig hook](https://github.com/mawww/kakoune/wiki/EditorConfig) from the Wiki, that the editorconfig.kak script failed for buffers that were not real files. For example `*scratch*`, `*debug*`, `*make*`, `*grep*`, etc.

This PR adds a check in the `editorcofig-load` command to see if a file exists before calling the external `editorconfig` utility.

It also moves the `remove-hooks` line to after the valid file check and directly prints the command to add a remove trailing whitespace hook from awk (if necessary). This prevents an invalid file from clearing a hook and removes the need to store an option and reference it from a spawned shell in the later hook.

### DIFF
--- a/rc/detection/editorconfig.kak
+++ b/rc/detection/editorconfig.kak
@@ -11,53 +11,51 @@ hook global BufCreate .*[.](editorconfig) %{
     latin1 utf-8 utf-8-bom utf-16be utf-16le lf cr crlf unset space tab max_line_length
 }
 
-declare-option -hidden bool editorconfig_trim_trailing_whitespace false
-
 define-command editorconfig-load -params ..1 -docstring "editorconfig-load [file]: set formatting behavior according to editorconfig" %{
-    remove-hooks buffer editorconfig-hooks
     evaluate-commands %sh{
-        command -v editorconfig >/dev/null 2>&1 || { echo 'echo -markup "{Error}editorconfig could not be found"'; exit 1; }
-        editorconfig "${1:-$kak_buffile}" | awk -F= -- '
-            /indent_style=/             { indent_style = $2 }
-            /indent_size=/              { indent_size = $2 == "tab" ? 4 : $2  }
-            /tab_width=/                { tab_width = $2 }
-            /end_of_line=/              { end_of_line = $2 }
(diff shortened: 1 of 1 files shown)


mogenson:
Hello,  does this PR seem reasonable?

rc/detection/editorconfig.kak
```diff
@@ -11,53 +11,50 @@ hook global BufCreate .*[.](editorconfig) %{
...
-                }
-                if (trim_trailing_whitespace == "true") {
-                    print "set-option buffer editorconfig_trim_trailing_whitespace true"
-                }
-                if (max_line_length && max_line_length != "off") {
-                    print "set window autowrap_column " max_line_length
-                    print "autowrap-enable"
-                    print "add-highlighter window/ column %sh{ echo $((" max_line_length "+1)) } default,bright-black"
+        file="${1:-$kak_buffile}"
+        if [ -f "$file" ]; then
```
lenormf:
```suggestion
        if [ -e "$file" ]; then
```

To handle symlinks.

mawww:
Hello,

I dont think the logic should be to disable editorconfig for non-existing files, what we want is to disable it for non-file buffers (scratch buffers in Kakoune lingo). One distinction between file and non-file buffers is that $kak_buffile will be an absolute path for any file buffer, but just match the buffer name for a non-file buffer, so I suspect ignoring all buffers where $kak_buffile does not start with a slash would do the trick.

mogenson:
I've updated this PR to check if `$kak_buffile` begins with a `/` character.

As mawww suggested, this should work for all unix environments. On Cygwin, it appears you can specify a path that begins with a drive letter (ex: `cd c:/some/path`) but Cygwin will report back a path that begins with `/` (ex: `/cygdrive/c/some/path`).

Lastly, if you want to keep the editorconfig.kak script as-is. I've been using a WinCreate hook with regex to filter out buffer names that start with a `*`. This could be added to the wiki.

```
hook global WinCreate ^[^*]+$ %{ editorconfig-load } # no *buffers*
```

PR #3021, opened 2019-07-15
Be explicit about ordering when autoloading

winged:
Some autoloads may depend on another. Right now, the autoload ordering depends on the filesystem, so the ordering may change depending on what happens on the file system (`find` does not list the files in any given order).

This sorts the files before `source`ing them, thus making the ordering explicit.

### DIFF
--- a/share/kak/kakrc
+++ b/share/kak/kakrc
@@ -28,6 +28,7 @@ def -params 1 -docstring "colorscheme <name>: enable named colorscheme" \
 evaluate-commands %sh{
     autoload_directory() {
         find -L "$1" -type f -name '*\.kak' \
+            | sort \
             | sed 's/.*/try %{ source "&" } catch %{ echo -debug Autoload: could not load "&" }/'
     }
 


Screwtapello:
If Kakoune didn't have any way to declare that one module depends on another, this would be a good way to make autoloads more consistent and reliable.

As it is, though, we have the `provide-module` and `require-module` commands so modules can guarantee they load in the correct order, and the `ModuleLoaded` hook so modules can be loaded as lazily as possible. Kakoune will be faster the more people use those features; if `shuf` were a standard POSIX command, I'd propose we shuffle the output of `find` so we could find implicit dependencies more quickly and convert them to explicit dependencies.

winged:
Fair enough, I even kind of like the explicit shuffling idea :)

I was using the autoload folder to load my configs, and as such was depending on a certain order to get things right (They're not modules, so the provide/require-module mechanism can't be used).

After your comment I've now changed my config to a custom loader that /does/ the ordering withing my `.config/kak`, so I guess this can be dropped (unless someone steps in and wants to merge despite your objections :) )

andreyorst:
simple sorting makes little sense anyway. Breadth-first sorting would be more appropriate way. But unfortunately there's no builtin way to do such sorting with `find`. I wrote this piece of code to do this kind of sorting when sourcing plugins before `require-module` was introduced: https://github.com/andreyorst/plug.kak/blob/e7582a5a9c39b8b7f9d699400492889a39e1b6be/rc/plug.kak#L214-L218

mawww:
Yes, we should try to make module dependencies explicit rather than trying to enforce a sourcing order.

PR #3027, opened 2019-07-22
Extended Rust highlighting

andreyorst:
I've decided to extend Rust syntax highlighting based on how Emacs highlights Rust (without using actual Emacs regular expressions if that matters).

I've added function declaration highlighting, variable definition highligting, user types highlighting, and tweaked some other bits. Here's comparison, old on the left, new on the right:
sidy_by_side

To make it easier to view the difference, here's separate images to toggle between: [old.png](https://user-images.githubusercontent.com/19470159/61660006-96397100-acd1-11e9-8b00-8d6de26a70c5.png), [new.png](https://user-images.githubusercontent.com/19470159/61660009-99346180-acd1-11e9-8d5a-99a7ade9d6af.png), and here's Emacs screenshot for comparison: [emacs.png](https://user-images.githubusercontent.com/19470159/61634579-42f7fc00-ac9a-11e9-8f9f-60d4eae2eeea.png).

### DIFF
--- a/rc/filetype/rust.kak
+++ b/rc/filetype/rust.kak
@@ -6,13 +6,13 @@
 
 hook global BufCreate .*[.](rust|rs) %{
     set-option buffer filetype rust
-    require-module rust
 }
 
 # Initialization
 # ‾‾‾‾‾‾‾‾‾‾‾‾‾‾
 
 hook global WinSetOption filetype=rust %[
+    require-module rust
     hook window InsertEnd .* -group rust-trim-indent rust-trim-indent
     hook window InsertChar \n -group rust-indent rust-indent-on-new-line
     hook window InsertChar \{ -group rust-indent rust-indent-on-opening-curly-brace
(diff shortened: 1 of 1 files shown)


rc/filetype/rust.kak
```diff
@@ -6,15 +6,14 @@
 
 hook global BufCreate .*[.](rust|rs) %{
     set-option buffer filetype rust
+    require-module rust
```
lenormf: Why move this line to here? Other language support scripts have it in the `filetype` hook.

andreyorst: Why not? We set current filetype, which means we will load the module. There are several hooks for this filetype and I think it's safer this way.

lenormf: Moving this line does not fix a bug, and creates a discrepancy with how other filetype scripts work. If you think it's safer, better modify all the scripts at once in a bug-fix commit instead of having unrelated commits doing it randomly.

PR #3029, opened 2019-07-22
Allow / in target when highlighting in makefile.kak

codesoap:
Example (snippet):
```makefile
bin/mytool: src/mytool.c
    ...
```

### DIFF
--- a/rc/filetype/makefile.kak
+++ b/rc/filetype/makefile.kak
@@ -55,7 +55,7 @@ define-command -hidden makefile-indent-on-new-line %{
         # preserve previous line indent
         try %{ execute-keys -draft \;K<a-&> }
         ## If the line above is a target indent with a tab
-        try %{ execute-keys -draft Z k<a-x> <a-k>^[^\t].*?(::|:|!)<ret> z i<tab> }
+        try %{ execute-keys -draft Z k<a-x> <a-k>^\S.*?(::|:|!)<ret> z i<tab> }
         # cleanup trailing white space son previous line
         try %{ execute-keys -draft k<a-x> s \h+$ <ret>d }
         # indent after some keywords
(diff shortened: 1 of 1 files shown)


rc/filetype/makefile.kak
```diff
@@ -33,7 +33,7 @@ add-highlighter shared/makefile/content default-region group
 add-highlighter shared/makefile/comment region '#' '$' fill comment
 add-highlighter shared/makefile/evaluate-commands region -recurse '\(' '\$\(' '\)' fill value
 
-add-highlighter shared/makefile/content/ regex ^[\w.%-]+\h*:\s 0:variable
+add-highlighter shared/makefile/content/ regex ^[\w.%-/]+\h*:\s 0:variable
```
lenormf: What about windows style paths, e.g. `abc\\def`? I don't think we need to whitelist allowed characters here, we can simplify the regex with a simple-stupid:

```suggestion
add-highlighter shared/makefile/content/ regex ^\S+\h*:\s 0:variable
```

And if we want to be more accurate and not highlight trailing whitespace:

```suggestion
add-highlighter shared/makefile/content/ regex ^(\S+\h*:)\s 1:variable
```

codesoap: I initially didn't look into the syntax specification for make, but your suggestion made me research a bit more. I looked at the [OpenBSD make](https://man.openbsd.org/make) and [GNU make](https://www.gnu.org/software/make/manual/make.html) manuals.

Apparently there can be [multiple whitespace separated targets](https://www.gnu.org/software/make/manual/make.html#Multiple-Targets) and different target separators (OpenBSD's make has `:`, `::` and `!`). I'll think of a solution and update the pull request.

Thanks for making me think :-)

rc/filetype/makefile.kak
```diff
@@ -33,7 +33,7 @@ add-highlighter shared/makefile/content default-region group
 add-highlighter shared/makefile/comment region '#' '$' fill comment
 add-highlighter shared/makefile/evaluate-commands region -recurse '\(' '\$\(' '\)' fill value
 
-add-highlighter shared/makefile/content/ regex ^[\w.%-]+\h*:\s 0:variable
+add-highlighter shared/makefile/content/ regex ^\S.*?(::|:|!) 0:variable
```
lenormf: I think this would make a non-target line e.g. `ABC:=1`, hence the trailing `\s` in the previous regex.

PR #3030, opened 2019-07-22
ocaml.kak fix '"' and char literals

aecepoglu:
http://caml.inria.fr/pub/docs/manual-ocaml/lex.html#char-literal

Before
before
After
after


aecepoglu:
The ocaml.kak file is quite lacking. So if you like we can keep this PR open the following few weeks while I keep adding new syntax rules

Issue #2405, opened 2018-09-19
Add a syntax for indented strings


alexherbo2:
https://nixos.org/nix/manual#ssec-values

lenormf:
You want a special delimiter that removes whitespace in a uniform manner (not only surrounding)?

mawww:
As discussed on IRC, my proposal for that would be to tweak the behaviour of `%{...}` strings by adding the following rule:

"If a %{...} string contents starts with and end-of-line, then the immediately following indentation will be removed from all lines in the string, and it is an error for a line in the string not to contain that identation".

This would avoid adding yet another string type, and would help having %sh{...} expands nicely formatted, as the Kakoune side indentation would be stripped before passing that string to the shell.

alexherbo2:
@mawww `{…}` strings means all string pairs?

- `(…)`
- `{…}`
- `[…]`
- `<…>`

Or `{…}` specifically. :thinking:

mawww:
All, possibly even including non balanced % strings...

eraserhd:
Working through corner cases in my head:

1. How to quote a literal newline?  Easy, make a newline followed by no indentation (and end of string?) literal.  In fact, the rule should not apply unless there's a non-\s, in case we want to quote newline followed by whitespace for some reason (?)
2. Blank lines in indented region?  Should allow them to have no (or less than) the required indentation.  I think Python does this.

net:
@mawww How would you represent a string who's first line is indented further than a subsequent line?

```
  indented first line
second line
```

I propose that the indentation is derrived from the closing delimiter:

```
... %{
    indented first line
  second line
  }
```

without trailing newline:

```
... %{
    indented first line
  second line\
  }
```

alexherbo2:
Why do not handle the same way as Nix does?

PR #3031, opened 2019-07-23
src: De-indent docstrings passed to command/option/mapping definitions

lenormf:
This commit implements formatting behaviour when the first character of a
docstring is a newline. In that case, the exact indentation level of the
next line will be removed from that line and all subsequent non-empty lines.

An error will be returned if a subsequent non-empty line does not have the
same indentation level.

The docstrings are always trimmed (surrounding whitespaces) whether the
first character is a newline or not, as was the case prior to this commit.

Example: the following declaration

```
define-command test -docstring %{
    test: do something

    Nothing really.
        More
            indented
                lines.
} nop
```

would be rendered as

```
test: do something

Nothing really.
    More
        indented
            lines.
```

Related to #2405

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -1163,9 +1163,9 @@ void define_command(const ParametersParser& parser, Context& context, const Shel
         };
     }
 
-    auto docstring = trim_whitespaces(parser.get_switch("docstring").value_or(StringView{}));
+    auto docstring = trim_indent(parser.get_switch("docstring").value_or(StringView{}));
 
-    cm.register_command(cmd_name, cmd, docstring.str(), desc, flags, CommandHelper{}, completer);
+    cm.register_command(cmd_name, cmd, docstring, desc, flags, CommandHelper{}, completer);
 }
 
 const CommandDesc define_command_cmd = {
@@ -1590,7 +1590,7 @@ const CommandDesc declare_option_cmd = {
         if (parser.get_switch("hidden"))
             flags = OptionFlags::Hidden;
 
-        auto docstring = trim_whitespaces(parser.get_switch("docstring").value_or(StringView{})).str();
+        auto docstring = trim_indent(parser.get_switch("docstring").value_or(StringView{}));
         OptionsRegistry& reg = GlobalScope::instance().option_registry();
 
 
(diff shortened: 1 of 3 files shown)


src/commands.cc
```diff
@@ -1045,6 +1046,37 @@ Vector<String> params_to_shell(const ParametersParser& parser)
     return vars;
 }
 
+String deindent_docstring(StringView docstring)
```
mawww: I'd rather have this in string_utils.cc, and not referecing docstrings explicitely. Maybe called 'trim_indent' ?

src/commands.cc
```diff
@@ -1045,6 +1046,37 @@ Vector<String> params_to_shell(const ParametersParser& parser)
...
+{
+    if (docstring.empty())
+        return {};
+    else if (docstring[0_byte] != '\n')
+        return trim_whitespaces(docstring).str();
+
+    String ds = docstring.substr(1_byte).str();
```
mawww: `ds` is not very expressive, I'd rather maybe reuse the parameter (str = str.substr(1_byte)) ? Also, why a String instead of a StringView here ?

src/string.hh
```diff
@@ -83,6 +83,8 @@ public:
     StringView substr(CharCount from, CharCount length = INT_MAX) const;
     StringView substr(ColumnCount from, ColumnCount length = INT_MAX) const;
 
+    bool starts_with(StringView prefix) const;
```
mawww: I dont like adding that to the string interface, it should be a free function in string_utils.hh.

src/commands.cc
lenormf: I don't think "trim" is specific enough, and this function is only used on docstrings hence the "docstring" in the name.

src/commands.cc
lenormf: It's short for "docstring", I didn't really think hard on the variable name.

I created a string because I thought I would modify it directly at first, but then I used `transform` and didn't change the type back.

src/string.hh
lenormf: Actually this function has already been implemented, it's called `prefix_match()`!

lenormf:
Ready for further review!

PR #3034, opened 2019-07-24
Use KAK_BIN_PATH as portable fallback.

niacat:
Instead of just using KAK_BIN_PATH on OpenBSD, use it on any OS that doesn't have platform-specific code, instead of erroring.

Fixes build on NetBSD, probably illumos too.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -84,6 +84,7 @@ else
     endif
 
     LIBS += $(shell $(PKG_CONFIG) $(PKG_CONFIG_FLAGS) --libs ncursesw)
+    CPPFLAGS += -D'KAK_BIN_PATH="$(bindir)/kak"'
     NCURSES_CFLAGS += $(shell $(PKG_CONFIG) $(PKG_CONFIG_FLAGS) --cflags ncursesw)
     LDFLAGS += -rdynamic
 endif
--- a/src/file.cc
+++ b/src/file.cc
@@ -642,10 +642,8 @@ String get_kak_binary_path()
     kak_assert(res != -1);
     buffer[res] = '\0';
     return buffer;
-#elif defined(__OpenBSD__)
-    return KAK_BIN_PATH;
 #else
-# error "finding executable path is not implemented on this platform"
+    return KAK_BIN_PATH;
 #endif
 }
 


alexherbo2:
@niacat Maybe rename `KAK_BIN_PATH` → `KAKOUNE_BIN_PATH`

d127483

src/Makefile
```diff
@@ -84,6 +84,7 @@ else
     endif
 
     LIBS += $(shell $(PKG_CONFIG) $(PKG_CONFIG_FLAGS) --libs ncursesw)
+    CPPFLAGS += -D'KAK_BIN_PATH="$(bindir)/kak"'
```
lenormf: Are the simple quotes necessary here?

niacat: The single quotes can be replaced with escaped double quotes: -DKAK_BIN_PATH=\"$(bindir)/kak\". It's a matter of preference.

I used single quotes because that's how it was already done. It can be changed to use escaped double quotes for both occurrences if wanted.

lenormf:
```
-DKAK_BIN_DIR="$(bindir)/kak"
```

The above works on my system, does it not with BSD/make as well? I can see now that the same syntax (with both styles of quotes) is being used for OpenBSD targets, it's not clear to me why you would need to escape the expansion twice.

niacat: Escaping is necessary because the quotes are part of the define value, and are not intended to be parsed by the shell.

Without escaping the path, the preprocessor will output invalid C++: `return /usr/pkg/bin/kak`.

unattributed: > ```
> -DKAK_BIN_DIR="$(bindir)/kak"
> ```
> 
> The above works on my system, does it not with BSD/make as well? I can see now that the same syntax (with both styles of quotes) is being used for OpenBSD targets, it's not clear to me why you would need to escape the expansion twice.

The only reason it "works on my system" is that you're not running NetBSD and so the code isn't actually getting compiled.

lenormf: Running NetBSD is irrelevant, it was a matter of not letting the shell expand the string as @niacat explained.

unattributed: It's actually very relevant; anything inside an #if block isn't going to be seen by the C compiler if the conditional evaluates to 0. And if the code doesn't get seen by the C compiler, it can't see the syntax error you've just introduced by stripping out the quotes.

lenormf: I'm running Linux, I could test this PR without having to use NetBSD - it's irrelevant.

Thanks @niacat for the explanation, sorry for the noise!

mawww:
Hello,

I am not very keen on this change, the preferred way is to have runtime lookup of the binary path, so for new platforms I would rather have a compilation failure that would lead to checking if that can be provided, and only in the case where it is not possible shall we fallback on KAK_BIN_PATH.

Moreover KAK_BIN_PATH handling is not really fool proof, changing it will not trigger a recompilation of the relevant sources for example.

Regarding NetBsd in particular, it seems it does support KERN_PROC_PATHNAME sysctl, so could share the FreeBsd implementation.

Issue #3015, opened 2019-07-09
Option for sorting punctuation/non-punctuation in completion?


ganwell:
I use kak-lsp with python. In python there are a lot `__*__()` methods, these are hidden methods. Because kakoune sorts punctuation < non-punctuation, I get all the `__*__()` methods first. I am not interested in these. (Update typo: > should be <)

Changing the sort-order is a breaking change, therefore I propose an option to reverse the order, defaulting in the current behavior. As I just started using kakoune, chances are high that there are better solutions for the problem. Are there any workarounds? I didn't find a way to suppress `__*__()` methods in kak-lsp nor jedi.

One can test the reversed order with this patch:

```cpp
diff --git a/src/ranked_match.cc b/src/ranked_match.cc
index 423fc814..810aa873 100644
--- a/src/ranked_match.cc
+++ b/src/ranked_match.cc
@@ -214,7 +214,7 @@ bool RankedMatch::operator<(const RankedMatch& other) const
             const bool punct1 = iswpunct((wchar_t)cp1);
             const bool punct2 = iswpunct((wchar_t)cp2);
             if (punct1 != punct2)
-                return punct1;
+                return punct2;

             const bool low1 = iswlower((wchar_t)cp1);
             const bool low2 = iswlower((wchar_t)cp2);
```

mawww:
Hello,

There is no way to tweak the ordering besides the C++ sorting code.

From looking at the code, it seems this punctionation ordering was introduced to ensure the following:
```
    kak_assert(RankedMatch{"find(1p)", "find"} < RankedMatch{"findfs(8)", "find"});
```
That is, so that `find(1p)` is considered a better match than `findfs(8)` when the search candidate is `find`. I guess we also want that to be the case if the search candidates was say `fin`. However I agree that in general, punctuation should be ranked after alphanumeric, so what we need is to find a tweak to the sorting that fulfills those requirements.

I suspect what we really want for the `find` case is that shorter words are ranked first, so maybe it would be better to detect word boundaries instead of punctuation.

ganwell:
Closeing because it is merged

PR #3038, opened 2019-07-29
Rank a word-boundary after a non-word-boundary

ganwell:
Update: removed ideas that turned out to be wrong.

### DIFF
--- a/src/ranked_match.cc
+++ b/src/ranked_match.cc
@@ -171,6 +171,12 @@ RankedMatch::RankedMatch(StringView candidate, StringView query)
 {
 }
 
+static bool is_word_boundary(Codepoint prev, Codepoint c)
+{
+    return (iswalnum((wchar_t)prev)) != iswalnum((wchar_t)c) or
+           (iswlower((wchar_t)prev) != islower((wchar_t)c));
+}
+
 bool RankedMatch::operator<(const RankedMatch& other) const
 {
     kak_assert((bool)*this and (bool)other);
@@ -193,6 +199,7 @@ bool RankedMatch::operator<(const RankedMatch& other) const
     auto order = [](Codepoint cp) { return cp == '/' ? 0 : cp; };
 
     auto it1 = m_candidate.begin(), it2 = other.m_candidate.begin();
+    const auto begin1 = it1, begin2 = it2;
     const auto end1 = m_candidate.end(), end2 = other.m_candidate.end();
     auto last1 = it1, last2 = it2;
     while (true)
@@ -207,14 +214,17 @@ bool RankedMatch::operator<(const RankedMatch& other) const
         // compare codepoints
         it1 = utf8::character_start(it1, last1);
         it2 = utf8::character_start(it2, last2);
+        const auto itsave1 = it1, itsave2 = it2;
         const auto cp1 = utf8::read_codepoint(it1, end1);
         const auto cp2 = utf8::read_codepoint(it2, end2);
         if (cp1 != cp2)
         {
-            const bool punct1 = iswpunct((wchar_t)cp1);
-            const bool punct2 = iswpunct((wchar_t)cp2);
-            if (punct1 != punct2)
-                return punct1;
+            const auto cplast1 = utf8::prev_codepoint(itsave1, begin1).value_or(Codepoint{0});
+            const auto cplast2 = utf8::prev_codepoint(itsave2, begin2).value_or(Codepoint{0});
+            const bool is_wb1 = is_word_boundary(cplast1, cp1);
+            const bool is_wb2 = is_word_boundary(cplast2, cp2);
+            if (is_wb1 != is_wb2)
+                return is_wb1;
 
             const bool low1 = iswlower((wchar_t)cp1);
             const bool low2 = iswlower((wchar_t)cp2);
@@ -243,6 +253,11 @@ UnitTest test_ranked_match{[] {
     kak_assert(RankedMatch{"class", "cla"} < RankedMatch{"class::attr", "cla"});
     kak_assert(RankedMatch{"meta/", "meta"} < RankedMatch{"meta-a/", "meta"});
     kak_assert(RankedMatch{"find(1p)", "find"} < RankedMatch{"findfs(8)", "find"});
+    kak_assert(RankedMatch{"find(1p)", "fin"} < RankedMatch{"findfs(8)", "fin"});
+    kak_assert(RankedMatch{"sys_find(1p)", "sys_find"} < RankedMatch{"sys_findfs(8)", "sys_find"});
+    kak_assert(RankedMatch{"init", ""} < RankedMatch{"__init__", ""});
+    kak_assert(RankedMatch{"init", "ini"} < RankedMatch{"__init__", "ini"});
+    kak_assert(RankedMatch{"a", ""} < RankedMatch{"b", ""});
     kak_assert(RankedMatch{"expresions", "expresins"} < RankedMatch{"expressionism's", "expresins"});
 }};
 
--- a/src/utf8.hh
+++ b/src/utf8.hh
@@ -4,6 +4,7 @@
 #include "assert.hh"
 #include "unicode.hh"
 #include "units.hh"
+#include "optional.hh"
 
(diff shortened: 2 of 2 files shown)


ganwell:
Refs #3015

src/ranked_match.cc
```diff
@@ -36,9 +36,11 @@ bool matches(UsedLetters query, UsedLetters letters)
 
 using Utf8It = utf8::iterator<const char*>;
 
-static int count_word_boundaries_match(StringView candidate, StringView query)
+static int count_word_boundaries_match_first(StringView candidate, StringView query, int &first_word_len)
```
mawww: declare a struct WordBoundaries { int count; int first_word_len; } and return that, instead of an out parameter.

src/ranked_match.cc
```diff
@@ -68,6 +75,12 @@ static int count_word_boundaries_match(StringView candidate, StringView query)
     }
     return count;
 }
+static int count_word_boundaries_match(StringView candidate, StringView query)
```
mawww: No need for this function.

src/ranked_match.cc
```diff
@@ -228,6 +238,9 @@ bool RankedMatch::operator<(const RankedMatch& other) const
 }
 
 UnitTest test_ranked_match{[] {
+    int first_word_len = 0;
+    count_word_boundaries_match_first("run(test)", "rat", first_word_len);
```
mawww:
```
auto eq = [](WordBoundaries lhs, WordBoundaries rhs) { return lhs.count == rhs.count and lhs.first_word_len == rhs.first_word_len; };
kak_assert(eq(count_word_boundaries_match(...), {2, 1}))
...
```

ganwell:
Thanks a lot for the tips, I really like what C++ looks like in the kakoune-source.

I first thought that Flags::SingleWord is related to the whole candidate, not only the match. So I had to count word-boundaries for whole candidate, too. If we don't check for the boundary-count, the preference for short words is too strong.

```cpp
    if (first_word_len != other_first_word_len)
        return first_word_len < other_first_word_len;
```

```
hal
Hal                     *s:en_GB*█
hale                    *s:en_GB*░
half                    *s:en_GB*░
hall                    *s:en_GB*░
halo                    *s:en_GB*░
halt                    *s:en_GB*░
Hale                    *s:en_GB*░
Hall                    *s:en_GB*░
Hals                    *s:en_GB*░
halal                   *s:en_GB*░
```

If we only use the first word length, if we match a multi-word, the single-word case stays the same (of course).

```cpp
    if ((count > 1 or other_count > 1) and first_word_len != other_first_word_len)
        return first_word_len < other_first_word_len;
```

```
hal
Hal                     *s:en_GB*█
halal                   *s:en_GB*░
halberd                 *s:en_GB*░
halcyon                 *s:en_GB*░
hale                    *s:en_GB*░
half                    *s:en_GB*░
halfback                *s:en_GB*░
halfhearted             *s:en_GB*░
halfheartedness         *s:en_GB*░
halfpence               *s:en_GB*░
```

I am starting to understand the RankedMatch better, but I still don't get the `find(1p)`/`findfs(8)` use-case completely. Especially because one can complete camelCase and snake_case, but not `find(1p)` the candidate is truncated at `(`. I more or less worked till the tests stopped failing.

So I like the code much better than before, but I am still not sure about the logic.

ganwell:
TLDR: I will write a better solution, using actual boundaries instead of word length.

I discussed the PR with a friend and I now understand the use-case and I know why the current logic isn't right.

It works for a basic case:

```
3│find(1p)
2│findfs(8)
1│
4│fin
~ find(1p)  *scratch*█
~ findfs(8) *scratch*█
```

But not for a more general case:

```
3│bla_find(1p)
2│bla_findfs(8)
1│bla_
4│bla_fin
~ bla_findfs(8) *scratch*█
~ bla_find(1p)  *scratch*█
```

I think, I also found a pretty straight forward way to implement this in a much smaller patch. I replace

```if (punct1 != punct2)```

with

```if (is_boundary1 != is_boundary2)```

Where I have to write code to set is_boundary by using the last codepoint too.

**Comment by mawww** [comment]
I had a WIP implementation for this that implemented the new behaviour you propose, it seemed to be working nicely at the time, I can either get it back, or let you implement that. It is a bit trickier than it appears due to utf8 decoding, as we first look at byte mismatches then switch to decoded codepoint at the mismatch position.

**Comment by ganwell** [comment]
@mawww you say your implementation is WIP? If its easy to finish, I'd just push it. It would be interesting to add the "bla_find(1)"-case to the asserts, though. If there is still work to do, I could try to finish it, just point to the brach/commit.

**Comment by ganwell** [comment]
@mawww I updated the PR, see the description on top https://github.com/mawww/kakoune/pull/3038#issue-302156615

File: src/ranked_match.cc [hunk 4]
```diff
@@ -207,14 +223,19 @@ bool RankedMatch::operator<(const RankedMatch& other) const
...
         if (cp1 != cp2)
         {
-            const bool punct1 = iswpunct((wchar_t)cp1);
-            const bool punct2 = iswpunct((wchar_t)cp2);
-            if (punct1 != punct2)
-                return punct1;
+            auto itlast1 = character_start_safe(itsave1 - 1, begin1);
+            auto itlast2 = character_start_safe(itsave2 - 1, begin2);
+            const auto cplast1 = utf8::read_codepoint(itlast1, end1);
+            const auto cplast2 = utf8::read_codepoint(itlast2, end2);
```
mawww: I think this will never consider the first char as a word boundary, as character_safe_start will trigger comparison of this character against itself. I think a better alternative would be to make `cplastX = prev_codepoint(itlastX, beginX, endX).value_or(0)` with prev_codepoint returning an Optional<Codepoint> when it exists.

ganwell: * I implemented prev_codepoint and put it into the utf8 namespace, since it seems reusable

* I added cp_zero so my code doesn't add a warning or need a `(Codepoint) 0` cast. But let me know what you prefer.

* I added

```cpp
kak_assert(RankedMatch{"b(fa", "a"} < RankedMatch{"(bfa", "a"} );
```

Because one might think, following the logic of `find(1p)` this should be sorted the opposite way, but it is essentially the same as:

```cpp
kak_assert(RankedMatch{"init", ""} < RankedMatch{"__init__", ""});
```

Which is my initial case (#3015) and it is important for completing something like `sys.` in python. At least after trying for a while I didn't find a useful criteria to distinguish these cases.

* And of course I was wrong about combining the two is_word_boundary, I stashed away `debug =? yes`  The tests would fail and its also clear that the count needs directed word-boundaries.

src/utf8.hh
```diff
@@ -258,6 +259,18 @@ Iterator character_start(Iterator it, const Sentinel& begin) noexcept
...
 
+// returns an optional iterator to the first byte of the previous character
+// or no value if it is at begin
+template<typename Iterator, typename Sentinel>
+static Optional<Codepoint> prev_codepoint(Iterator it, const Sentinel& begin) noexcept
+{
+    if (it <= begin)
+        return {};
+    auto itlast = utf8::character_start(it - 1, begin);
+    return utf8::read_codepoint(itlast, it);
```
mawww: no need for utf8:: here as we already are in that namespace. Also, you can just do

```
return codepoint(character_start(it -1, begin), it);
```

ganwell: @mawww Thank you for the patience, I should have seen these things.

src/unicode.hh
```diff
@@ -12,6 +12,7 @@ namespace Kakoune
 {
 
 using Codepoint = char32_t;
+const static Codepoint cp_zero = 0;
```
mawww: using `0` or `Codepoint{0}` would be as good. I'd rather not add magic constants like that.

ganwell: I removed the constant. If there is any issue left - also formatting or style - let me know. I am not up to date with C++, it was 2005 when I last used it full-time.

ganwell: Clang on travis threw an error because of `0`, I switched to `Codepoint{0}`

mawww:
Thanks for working on that !

PR #3039, opened 2019-07-29
Use $XDG_RUNTIME_DIR

### LINKED ISSUE #3019: Use $XDG_RUNTIME_DIR
### Steps

1. Open a Kakoune session (perhaps with a wrapper that chooses the session name based on tmux window name).
2. Experience power failure
3. Power server back on and log back into server
4. Open a Kakoune session

### Outcome

"Socket already in use"

`kak -clear` will clean it up.

### Expected

Socket doesn't exist, because it was stored in `$XDG_RUNTIME_DIR`, which intentionally does not persist across reboots.  If the variable is set... probably fine to default to the current mechanism when not set (it isn't set on Mac OS).

eraserhd:
Falls back on old mechanism if `XDG_RUNTIME_DIR` is not set.

The ability to specify a session as "<user>/<name>" was removed, since
it isn't possible to compute the value of `XDG_RUNTIME_DIR` for another
user, we wouldn't have access to it if we could, and it would be awkward
to support this feature only when `XDG_RUNTIME_DIR` is unset.  Also,
`rename-session` did not work when another user's session was specified.

Closes #3019

### DIFF
--- a/src/remote.cc
+++ b/src/remote.cc
@@ -537,18 +537,35 @@ String get_user_name()
     return getenv("USER");
 }
 
+void make_session_directory()
+{
+    StringView xdg_runtime_dir = getenv("XDG_RUNTIME_DIR");
+    if (xdg_runtime_dir.empty())
+    {
+        // set sticky bit on the shared kakoune directory
+        make_directory(format("{}/kakoune", tmpdir()), 01777);
+        make_directory(format("{}/kakoune/{}", tmpdir(), get_user_name()), 0711);
+    }
+    else
+        make_directory(format("{}/kakoune", xdg_runtime_dir), 0711);
+}
+
+String session_path(StringView session)
+{
+    if (contains(session, '/'))
+        throw runtime_error{"session names cannot have slashes"};
+    StringView xdg_runtime_dir = getenv("XDG_RUNTIME_DIR");
+    if (xdg_runtime_dir.empty())
+        return format("{}/kakoune/{}/{}", tmpdir(), get_user_name(), session);
+    else
+        return format("{}/kakoune/{}", xdg_runtime_dir, session);
+}
+
 static sockaddr_un session_addr(StringView session)
 {
     sockaddr_un addr;
     addr.sun_family = AF_UNIX;
-    auto slash_count = std::count(session.begin(), session.end(), '/');
-    if (slash_count > 1)
-        throw runtime_error{"session names are either <user>/<name> or <name>"};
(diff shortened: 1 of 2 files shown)


src/remote.cc
```diff
@@ -537,18 +537,31 @@ String get_user_name()
     return getenv("USER");
 }
 
+String session_path(StringView session)
+{
+    auto slash_count = std::count(session.begin(), session.end(), '/');
+    if (slash_count > 0)
+        throw runtime_error{"session names cannot have slashes"};
```
mawww: `if (contains(session, '/'))` ?

src/remote.cc
```diff
@@ -537,18 +537,31 @@ String get_user_name()
...
+    StringView xdg_runtime_dir = getenv("XDG_RUNTIME_DIR");
+    if (xdg_runtime_dir.empty())
+    {
+        // set sticky bit on the shared kakoune directory
+        make_directory(format("{}/kakoune", tmpdir()), 01777);
+        make_directory(format("{}/kakoune/{}", tmpdir(), get_user_name()), 0711);
```
mawww: I find it surprising that the `session_path` function would create directories as a side effect, why is the sticky bit needed here ?

src/remote.cc
```diff
@@ -537,18 +537,31 @@ String get_user_name()
...
     addr.sun_family = AF_UNIX;
-    auto slash_count = std::count(session.begin(), session.end(), '/');
-    if (slash_count > 1)
-        throw runtime_error{"session names are either <user>/<name> or <name>"};
-    else if (slash_count == 1)
-        format_to(addr.sun_path, "{}/kakoune/{}", tmpdir(), session);
-    else
-        format_to(addr.sun_path, "{}/kakoune/{}/{}", tmpdir(),
-                  get_user_name(), session);
+    strcpy(addr.sun_path, session_path(session).c_str());
```
mawww: its a shame we have to go through an allocation here now, but I guess its not that bad.

src/remote.cc
eraserhd: The sticky bit was already there previously.  I looked it up, and it prevents users from renaming and deleting each others' sockets in `/tmp/kakoune`, which normally only requires directory write permission.

src/remote.cc
eraserhd: Done

eraserhd:
I updated the PR to resolve everything (except the sticky bit, which I think is right).

PR #3040, opened 2019-07-30
Insert Clojure ns directive for new files

eraserhd:


rc/filetype/clojure.kak
```diff
@@ -212,4 +217,19 @@ define-command -hidden clojure-indent-on-new-line %{
     }
 }
 
+declare-option -docstring %{top-level directories which can contain clojure files
+e.g. '(src|test|dev)'} regex clojure_source_directories '(src|test|dev)'
+
+define-command -override -docstring %{clojure-insert-ns: Insert namespace directive at top of Clojure source file} \
```
laelath: I don't think you need `-override` here?

eraserhd: Oops, removed.   🙇‍♀

PR #3047, opened 2019-08-05
nim.kak: General improvements

tawsi-melek:
- Add autoindentation for enum, tuple and object:

Before:
```nim
type
  MyType = tuple
  myint: int     # This line needs to be indented manually
```

After:
```nim
type
  MyType = tuple
    myint: int     # Lines after 'tuple' are now indented automatically
```

- Prevent keywords from being highlighted when inside of a comment (e.g. "if" was being highlighted when written in a comment)

- ~~Highlight Nim's escape sequences inside strings (\p,\n,\r,\t, etc.)~~ removed

- Add on/off to the boolean values understood by the highlighter

- Remove redundant regex highlighter for comments

- Fix autoindent indenting more than it should

mawww, as this is my first "proper" PR, I've included the Copyright Waiver.

### DIFF
--- a/rc/filetype/nim.kak
+++ b/rc/filetype/nim.kak
@@ -37,8 +37,18 @@ add-highlighter shared/nim regions
 add-highlighter shared/nim/code default-region group
 add-highlighter shared/nim/triple_string region '([A-Za-z](_?\w)*)?"""' '"""(?!")' fill string
 add-highlighter shared/nim/raw_string region [A-Za-z](_?[A-Za-z])*" (?<!")"(?!") fill string
-add-highlighter shared/nim/string region (?<!'\\)"(?!') (?<!\\)(\\\\)*" fill string
-add-highlighter shared/nim/comment region '#?#\[' '\]##?' fill comment
+add-highlighter shared/nim/string region (?<!'\\)"(?!') (?<!\\)(\\\\)*" group
+add-highlighter shared/nim/comment region '#?#\[' '\]##?' group
+add-highlighter shared/nim/comment_line region '#?#[^\[]' $ group
+
+add-highlighter shared/nim/string/fill fill string
+add-highlighter shared/nim/comment/fill fill comment
+add-highlighter shared/nim/comment_line/fill fill comment
+# Escape sequences in string
+add-highlighter shared/nim/string/escape regex \\[prcnlftv] 0:value
+# Comment tags
+add-highlighter shared/nim/comment/tags regex \b(TODO|FIXME|XXX|NOTE|BUG)\b 0:default+rb
+add-highlighter shared/nim/comment_line/tags regex \b(TODO|FIXME|XXX|NOTE|BUG)\b 0:default+rb
 
 evaluate-commands %sh{
     # Grammar
@@ -66,7 +76,7 @@ evaluate-commands %sh{
     types="int int8 int16 int32 int64 uint uint8 uint16 uint32 uint64 float
     float32 float64 bool char object seq array cstring string tuple varargs
     typedesc pointer byte set typed untyped void auto"
-    values="false true"
+    values="false true on off"
 
     join() { sep=$2; set -- $1; IFS="$sep"; echo "$*"; }
 
@@ -96,7 +106,6 @@ evaluate-commands %sh{
 }
 
 add-highlighter shared/nim/code/ regex '(,|;|`|\(\.?|\.?\)|\[[.:]?|\.?\]|\{\.?|\.?\})' 0:meta
-add-highlighter shared/nim/code/ regex '#[^\n]+' 0:comment
 add-highlighter shared/nim/code/ regex %{'(\\([rcnlftvabe\\"']|0*[12]?\d?\d|x[0-9a-fA-F]{2})|[^'\n])'} 0:string
 
 # Commands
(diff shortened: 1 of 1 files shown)


SolitudeSF:
`enum`?

tawsi-melek:
Nice catch. Updated.

<details><summary>ot</summary>
<p>
It goes deeper than you think. I've been using Void Linux for years. Then I discovered nim two years ago (I'm amalek on the forum btw). And when some months ago I switched to kakoune from neovim I noticed you wrote the linter hook for nim and I said: "SolitudeSF? Wait a moment, I heard that name before..." Not that having Lain-chan as an avatar is that rare for programmers, mind you
</p></details>

rc/filetype/nim.kak
```diff
@@ -37,8 +37,18 @@ add-highlighter shared/nim regions
...
+add-highlighter shared/nim/comment region '#?#\[' '\]##?' group
+add-highlighter shared/nim/comment_line region '#?#[^\[]' $ group
+
+add-highlighter shared/nim/string/fill fill string
+add-highlighter shared/nim/comment/fill fill comment
+add-highlighter shared/nim/comment_line/fill fill comment
+# Escape sequences in string
+add-highlighter shared/nim/string/escape regex \\[prcnlftv] 0:value
+# Comment tags
+add-highlighter shared/nim/comment/tags regex \b(TODO|FIXME|XXX|NOTE|BUG)\b 0:default+rb
```
SolitudeSF: This shouldn't be done by language plugin.

rc/filetype/nim.kak
```diff
@@ -66,7 +76,7 @@ evaluate-commands %sh{
     types="int int8 int16 int32 int64 uint uint8 uint16 uint32 uint64 float
     float32 float64 bool char object seq array cstring string tuple varargs
     typedesc pointer byte set typed untyped void auto"
-    values="false true"
+    values="false true on off"
```
SolitudeSF: only place where on/off are used are pragmas for compiler options.

tawsi-melek:
I've fixed a few additional things and added some more. Updated to OP.

rc/filetype/nim.kak
tawsi-melek: No, they actually are aliases for true and false. I use them all the time. Try this on play.nim-lang.org ;-)
```nim
let is_on_an_alias_for_true = on

if is_on_an_alias_for_true != off:
  echo "YES!!!"
```

SolitudeSF: thats cool that they are not even mentioned in manual or grammar. i guess that was a hack made specifically for pragmas.

SolitudeSF:
you've actually fixed some string/comment edgecases i struggled with. i have one left that i remember. `'#'` gets highlighted incorrectly.

rc/filetype/nim.kak
tawsi-melek: Are you referring to both the escape sequences and the comment tags or just the tags? I know that there is [this](https://github.com/lenormf/kakoune-extra/blob/master/comnotes.kak) script for the tags, but it has a global scope and I couldn't get it to activate only on nim buffers (for some reason changing global/ to buffer/ completely disables ALL highlighters, even the ones in nim.kak).

Of course, if it's a kakoune policy I'll just remove it.

tawsi-melek:
If you're referring to the fact that a single `#` wasn't enough to trigger the highlighter, that should be solved now.

I actually studied how python.kak and sh.kak did things and used their approach. One thing that would be cool to do is have formatted strings like `fmt"{myvar} is my var!"` highlight the sequence inside `{}`, but I'm still thinking about how to go about it.

rc/filetype/nim.kak
SolitudeSF: these should be opt-in, since they are not related to language in any way. its a one liner, you can just add it to your own `hook global WinSetOption filetype=nim`.

SolitudeSF:
again, maybe overdoing is not the best. format strings are not builtin in nim, so you cant guarantee that `fmt"{myvar}"` in your code has the same meaning in someone else's code (even if its unlikely).

SolitudeSF:
also, this is how the `'#'` is displayed with your current version.
image

rc/filetype/nim.kak
tawsi-melek: Yeah, Nim's manual is perennially lagging behind the actual specification.

tawsi-melek:
I've managed to fix that with the last commit, but there is still an edge case that I can't fix:

```nim
let
  c1 = '1' # This is highlighted properly
  c2 = '2'# This is NOT highlighted properly (notice the lack of space between ' and #)
```

rc/filetype/nim.kak
tawsi-melek: Ok. I removed them.

SolitudeSF:
that probably pretty uncommon pattern. maybe you could take a look at another one:
image

this one is too, result of fixing another edgecase, but its not that uncommon.

SolitudeSF:
im not sure if this is even possible to fix.

i changed `nim/string` region highlighter to `add-highlighter shared/nim/string region '"' ((?<!\\)(\\\\)*"|$) group`, so `'"'` breaks only to the end of the line, but the edgecase i posted above is fixed then. fair tradeoff.
this goes out of my initial philosophy, that incorrect syntax shouldn't be highlighted, but i guess we're just trying to make it not break.

tawsi-melek:
I guess then we need a generalized char literal highlighter which overrides the string/comment hl. The c-family.kak seems to have one, I'll see if I can adapt it to work for Nim.

tawsi-melek:
Aaaand... it's done. I've modified the string hl after your input: it now works.

I've also ported the char hl from c-family.kak and made sure that the issue at https://github.com/mawww/kakoune/pull/3047#issuecomment-519683833 doesn't happen anymore.

If no more issues pop up, I think the PR can be merged.

SolitudeSF:
i guess a nitpick, but i made that only correct char literals are highlighted, now you can write incorrect char constanst and they will be highlighted. i guess more specification is more expensive, but was that really an issue?

tawsi-melek:
I reverted it. I changed it in the first place because I thought that it would fix https://github.com/mawww/kakoune/pull/3047#issuecomment-519683833, but upon close inspection all it was needed was to change the `comment_line`.

PR #3049, opened 2019-08-06
[Rust] Highlight strings in meta attributes

### LINKED ISSUE #3043: New rust highlighting doesn't handle braces in strings in meta attributes
### Steps 

Compile and install the latest kakoune from git, from the commit https://github.com/mawww/kakoune/commit/550feb1706e248edd4fe0cf526c12cdefbfab7ce.

Create a new file, `test.rs`, and enter the following code:

```rust
#[doc = "See: [std::result::Result]."]
fn function() {

}
```

### Outcome 

The code is highlighted as though the inner `]` ends the meta block, and then the following `"` starts a new string.

screenshot-20190802-134422

### Expected 

I would expect that the meta-color highlighting continues until the final `]`, and then the function is highlighted as normal. Something like what happens if I have a string in a meta block which doesn't contain `[]`:

screenshot-20190802-134446

Running kakoune compiled from https://github.com/mawww/kakoune/commit/e840140b38f5e690b61016f5f982c1765165328e, directly before #3027 was merged, also results in correct (but slightly less nice) highlighting:

screenshot-20190802-135053

---

I should say that I do like the new highlighting better overall, and this is definitely an edge case. But this is correct rust code, so it'd be nice if it was highlighted as such.

andreyorst:
fix #3043

Before:
image

After:
image

@daboross does this syntax supports raw strings?

---

Added raw strings:
image

### DIFF
--- a/rc/filetype/rust.kak
+++ b/rc/filetype/rust.kak
@@ -47,7 +47,8 @@ add-highlighter shared/rust/line_comment     region "//" "$"                    
 
 add-highlighter shared/rust/macro_attributes region -recurse "\[" "#!?\[" "\]" regions
 add-highlighter shared/rust/macro_attributes/ default-region fill meta
-add-highlighter shared/rust/macro_attributes/inner region %{(?<!')"} (?<!\\)(\\\\)*" fill string
+add-highlighter shared/rust/macro_attributes/string region %{(?<!')"} (?<!\\)(\\\\)*" fill string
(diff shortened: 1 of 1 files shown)


daboross:
@andreyorst Thanks for this!

This syntax does support raw strings - the following is a valid rust file:

```rust
#[doc = r#"
Hey, some "documentation".

For this item. "
"#]
pub fn x() {

}
```

I think this is probably an even less common case than using `#[doc = ""]` at all, though.

andreyorst:
ill add this

Issue #3051, opened 2019-08-09
Makefile highlighting issue/improvement list


andreyorst:
This issue is for tracking notable highlighting issues for Makefile filetype. I've provided screenshot for two more editors to compare.

- [x] - Targets with curly braces are not highlighted: [Kakoune](https://user-images.githubusercontent.com/19470159/62757073-ac418280-ba82-11e9-8602-9192d31b716b.png), [Emacs](https://user-images.githubusercontent.com/19470159/62757082-b82d4480-ba82-11e9-8db2-37ebbb517318.png), [Atom](https://user-images.githubusercontent.com/19470159/62757112-d430e600-ba82-11e9-9cd4-d361e50ecdca.png) (fixed by #3029)
- [x] - Targets with `/` are not highlighted: [Kakoune](https://user-images.githubusercontent.com/19470159/62757217-37bb1380-ba83-11e9-82d0-857bcbf9673c.png), [Emacs](https://user-images.githubusercontent.com/19470159/62757187-1b1edb80-ba83-11e9-9dc8-704ac29fb2d4.png), [Atom](https://user-images.githubusercontent.com/19470159/62757162-080c0b80-ba83-11e9-8781-c6f0064cb697.png) (fixed by #3029)
- [x] - Multiple targets separated with spaces are not highlighted: [Kakoune](https://user-images.githubusercontent.com/19470159/62852752-0e082380-bcf3-11e9-8cb2-ce295834846d.png), [Emacs](https://user-images.githubusercontent.com/19470159/62852740-021c6180-bcf3-11e9-8047-4d09770ec058.png), [Atom](https://user-images.githubusercontent.com/19470159/62852809-2ed07900-bcf3-11e9-8aa2-3a54a4d3b596.png) (fixed by #3029)
- [ ] - Strings are not highlighted: [Kakoune](https://user-images.githubusercontent.com/19470159/62757272-7224b080-ba83-11e9-8efc-deb0cef2d971.png), [Emacs](https://user-images.githubusercontent.com/19470159/62757282-7fda3600-ba83-11e9-9570-eaa314c6576f.png), [Atom](https://user-images.githubusercontent.com/19470159/62757296-908aac00-ba83-11e9-97cb-24889b3b8fe8.png) (tracked by #3052)
[...]

PR #3052, opened 2019-08-10
rc makefile: Various improvements

lenormf:
Hi,

I made changes to the highlighters that handle `Makefile` files in order to tick some boxes in #3051.

Here's what the project's `Makefile` looks like with my theme:

HTH.

### DIFF
--- a/rc/filetype/makefile.kak
+++ b/rc/filetype/makefile.kak
@@ -34,6 +34,7 @@ add-highlighter shared/makefile/comment region '#' '$' fill comment
 add-highlighter shared/makefile/evaluate-commands region -recurse '\(' '\$\(' '\)' fill value
 
 add-highlighter shared/makefile/content/ regex ^[\w.%-]+\h*:\s 0:variable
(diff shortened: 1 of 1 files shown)


andreyorst:
It seems that spaces in targets aren't supported:
image
Also, note orange `)`, might be some variable regex match? Here as well:
image

Other than that everything looks fine so far

lenormf:
Could you please paste the `Makefile` you use for testing?

andreyorst:
```Makefile
clean distclean:
	@echo "Cleaning common libraries.. ($(MAKECMDGOALS))"
	@for i in $(SUBDIRS); do \
	$(MAKE) -C $$PWD/$$i $(MAKECMDGOALS); \
	done
	$(MAKE) -C $$PWD/../tools $(MAKECMDGOALS); \
	rm -rf build *.lock
```

lenormf:
Thanks.

I've identified the problem, I'm just not sure how to handle that cleanly. Stay tuned.

lenormf:
I've created #3055 and #3056, which are the problems we're facing here. I'm not sure how to make highlighting accurate enough otherwise.

PR #3066, opened 2019-08-24
rc latex: Highlight .cls files as LaTeX

lenormf:
When .cls files are opened, the local `filetype` option is automatically
set to `tex`, which isn't supported.


nonumeros:
Hi @lenormf so this highlights `cls` files. It'd be nice If `dtx`, `cfg`, and `sty` can be included.

lenormf:
Since `cfg` is so generic, I didn't add it to the hook.

PR #3075, opened 2019-09-04
Shell cache proof of concept

JasonBrownDeveloper:
See #2683

### DIFF
--- a/rc/filetype/c-family.kak
+++ b/rc/filetype/c-family.kak
@@ -185,7 +185,7 @@ define-command -hidden c-family-insert-on-newline %[ evaluate-commands -itersel 
 ] ]
 
 # Regions definition are the same between c++ and objective-c
-evaluate-commands %sh{
+evaluate-commands %cached{
     for ft in c cpp objc; do
         if [ "${ft}" = "objc" ]; then
             maybe_at='@?'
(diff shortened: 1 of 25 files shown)


lenormf:
This would probably be better handled through a custom shell set with `$KAKOUNE_POSIX_SHELL`.

JasonBrownDeveloper:
I'm not familiar with how that would work.  Could you provide an example of how to set the `$KAKOUNE_POSIX_SHELL` to a shell that would already have the script result cached.

Screwtapello:
When `$KAKOUNE_POSIX_SHELL` is set to the absolute path of an executable, Kakoune will use that executable for all shell-related operations instead of falling back to `/bin/sh`. The executable is called with two arguments, `-c` and the text of the script to execute.

You could write a script that memoizes such invocations - hash the argument to the `-c` option, look it up in some database, if it exists pull the cached result from the database and use it, otherwise invoke the real shell yourself and cache a copy of the output.

That said, it's probably not possible to replicate the behaviour of this PR with a shell-wrapper. This PR lets the user decide which shell-scripts are cacheable, instead of letting the wrapper try to guess what possible parts of the environment might affect the output (environment variables, the contents of stdin, the current system time, what network hosts happen to be responding to PINGs).

Perhaps instead of *always* caching, the wrapper could be made to only cache scripts beginning with `#pragma cachable\n` or something?

JasonBrownDeveloper:
We're so far away from the original issue that I'm just cleaning this up.

PR #3076, opened 2019-09-04
Goto next and previous buffer

JasonBrownDeveloper:
I know this can be done with scripting, but it just seem like it was missing from the goto command.  I'm sure there are better ways to do this.  I often work with a handful of files and this seems more natural than :bn and :bp.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -184,7 +184,7 @@ sudo zypper install kakoune
 [TIP]
 .Ubuntu
 ====
-Building on Ubuntu 16.04.
+Building on Ubuntu 16.04 and 18.04.
 Make sure you have .local/bin in your path to make the kak binary available from your shell.
 
 ----------------------------------------------------------------
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -46,6 +46,41 @@ namespace Kakoune
 
 extern const char* version;
 
+template<bool next>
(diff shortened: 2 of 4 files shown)


lenormf:
No need for scripting:

```
map global goto n '<esc>: bn<ret>'
map global goto p '<esc>: bp<ret>'
```

JasonBrownDeveloper:
By scripting, I meant the map command in an auto loaded script.  I was just lamenting that its not part of the standard package.  Maybe adding those map commands to an existing script that comes with Kakoune?  None of the other goto commands are mapped though.

Delapouite:
I'm not in favor of adding buffer commands to the `goto` mode. I much prefer the strategy described here: https://github.com/delapouite/kakoune-buffers

TL;DR: having a dedicated `buffers` buffers mode bound on the `b` key, to do `bn` (buffer next), `bp` (buffer previous), `bd (buffer delete)`

What could make more sense is to migrate `<c-o>` and `<c-i>`. Adding in C++ the equivalent of:

```
map global goto n <esc><c-o> -docstring 'jump backward'
map global goto p <esc><tab> -docstring 'jump forward'
```

The fact that the current mapping use the `ctrl` modifier is a legacy from vim. Also, it would avoid the conflict between `<c-i>` and `<tab>`

PR #3081, opened 2019-09-09
add '-u' cmdline flag to specify config directory

### LINKED ISSUE #3072: Specify a kakrc to use.
Would it be possible to specify a config file instead of using nothing or the one at `~/.config/kak/kakrc`? I'd like to do this to increase startup speed when using it to edit commit messages etc.

nrdxp:
This resolves #3072. The user can now specify the directory containing their kakrc and other resource files using the `-u` command line switch.

### DIFF
--- a/doc/kak.1
+++ b/doc/kak.1
@@ -40,6 +40,7 @@ kak \- a vim inspired, selection oriented code editor
 .PP
 .B kak
 [\fB\-c\fR \fIsession_id\fR|\fB\-s\fR \fIsession_id\fR]
+[\fB\-u\fR \fIconfig_dir\fR]
 [\fB\-n\fR] [\fB\-ro\fR]
 [\fB\-ui\fR \fIui_type\fR] [\fB\-e\fR \fIcommand\fR]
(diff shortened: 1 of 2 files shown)


mawww:
It seems to me that `kak -u "$dir"` is equivalent to `XDG_CONFIG_HOME="$dir" kak`, I am not sure adding this alternate flag is really worth it, especially as it seems this would mostly be used through shell aliases, rarely interactively.

nrdxp:
The main benefit to the switch is not having to overwrite `XDG_CONFIG_HOME`. E.g. On NixOS, other processes kakoune might call in a script or shell expansion might expect their config files in a stateful location such as the users home directory, while `kakoune` itself can have its config living in the nix store.

It's a may be an obscure scenario but seeing as common programs like `git` also use this variable, perhaps the user doesn't wish to overwrite it for the entire session. For what it's worth, conceptually it just seems less hacky and more succinct to me. Overwriting a variable that is meant to determine a global configuration directory for a single process just doesn't 'feel' right.

That said, whatever we all decide is best, I am more than happy to abide by.

Issue #2985, opened 2019-06-24
Compilation warning in definition of the ArrayView type


lenormf:
The following warning is being printed when compiling with `g++` v9 (`clang++` v8 doesn't show it):

```
g++  -O3 -pedantic -std=c++17 -g -Wall -Wextra -Werror -Wno-unused-parameter -Wno-sign-compare -Wno-address -MD -MP -MF .alias_registry.opt.d -c -o .alias_registry.opt.o alias_registry.cc
In file included from unicode.hh:7,
                 from utf8.hh:5,
                 from string.hh:7,
                 from alias_registry.hh:5,
                 from alias_registry.cc:1:
array_view.hh: In instantiation of ‘constexpr Kakoune::ArrayView<T>::ArrayView(const std::initializer_list<_Tp>&) [with T = const Kakoune::StringView]’:
string_utils.hh:117:18:   required from ‘Kakoune::String Kakoune::format(Kakoune::StringView, Types&& ...) [with Types = {Kakoune::String}]’
string_utils.hh:132:44:   required from here
array_view.hh:39:48: error: initializing ‘Kakoune::ArrayView<const Kakoune::StringView>::m_pointer’ from ‘std::initializer_list<const Kakoune::StringView>::begin’ does not extend the lifetime of the underlying array [-Werror=init-list-lifetime]
   39 |         : m_pointer(v.begin()), m_size(v.size()) {}
      | 
```

I think the following documentation paragraph is relevant ("since C++14"):

[cppref](https://en.cppreference.com/w/cpp/utility/initializer_list)

>The underlying array is a temporary array of type const T[N], in which each element is copy-initialized (except that narrowing conversions are invalid) from the corresponding element of the original initializer list. […]

The problem might be fixed by turning the internal pointer in question into an rvalue reference?

[cppref](https://en.cppreference.com/w/cpp/language/reference_initialization#Lifetime_of_a_temporary)

HTH.

mawww:
This warning looks spurious to me, in C++ temporary object lifetime ends at the end of the full expression that triggers their creation, in other words, 'at the semicolon'. That means the temporary initializer_list (and backing array) are alive until the end of the format call, and do outlive the ArrayView.

Seems g++-9 gets confused about that somehow, or their might a C++ subtlety being triggered here.

mawww:
[...]

PR #3088, opened 2019-09-17
Added -Wno-init-list-lifetime to CXXFLAGS

laelath:


mawww:
With which compiler does that warning appear ?

lenormf:
Info in #2985, not sure adding a compilation flag is the way to go forward (a standard-compliant code patch is better than this workaround).

PR #3100, opened 2019-09-21
Make autoscroll behavior of grep buffer configurable via option

schickm:
This declares a new option named 'grepscroll'.  It defaults to the existing
behavior of always scrolling the grep fifo buffer to the end when new content
comes in.  However, when disabled (set to no/false), the selection will stay on
the first line.

I personally prefer to read my grep results starting at the top and working my way down.  [@alexherbo2 suggested this method of disabling scrolling of the grep buffer.](https://discuss.kakoune.com/t/goto-top-of-grep-buffer-after-each-run-of-grep/727/2)

### DIFF
--- a/rc/tools/grep.kak
+++ b/rc/tools/grep.kak
@@ -2,6 +2,8 @@ declare-option -docstring "shell command run to search for subtext in a file/dir
     str grepcmd 'grep -RHn'
 declare-option -docstring "name of the client in which utilities display information" \
     str toolsclient
+declare-option -docstring "scroll to bottom of results buffer each time grep is run" \
+	bool grepscroll true
 declare-option -hidden int grep_current_line 0
 
 define-command -params .. -file-completion \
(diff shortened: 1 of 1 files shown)


mawww:
I am not sure this needs to be configured, I think grep should not be scrolling at all. It makes sense for make buffers, as we usually want to follow the output and it tends to stop whenever we hit an error, but for grep buffers all parts of the buffer are equally important, so scrolling does not add any value.

Anybody thinks it really ought to be configurable ? Or would just removing scrolling from grep support be enough ?

lenormf:
I think the usecase is for grep utilities that search in the entire directory tree by default (e.g. `ag`), in which case searching can take a while.

mawww:
> I think the usecase is for grep utilities that search in the entire directory tree by default (e.g. `ag`), in which case searching can take a while.

Is that an argument in favor of scrolling ? Thats what I usually do, but I dont think scrolling is really helping there.

I think I'll disable scrolling locally to see if I miss anything.

lenormf:
Your first comment led me to believe the PR added the `-scroll` flag to `:edit` (I don't use that feature often), so I suggested users would probably want the command to work like in a shell where scrolling occurs automatically as matches are returned.

I think the current behaviour of the command is the result of a copy/paste of the implementation of `make.kak`, and not a conscious decision.

I can see why both behaviours could be asked for by users, but I don't have a dog in the fight (although sticking to what makes sense instead of allowing either behaviour arbitrarily at runtime seems best - as usual).

schickm:
Not scrolling makes the most sense to me😉

With the machine and codebases that I work with, grep results are rarely delayed to the point where terminal style scrolling would be desirable. 

Maybe leave the option in place, but default it to false?

PR #3105, opened 2019-09-23
Paste youngest instead of oldest value of read-only history registers 

krobelus:
The docs say that the register `:` contains the *last* entered command.
However, when typing `":p`, the *first* entered command is inserted.

This patch fixes this by always using the last entry for `:` and other history registers, for all selections.

### DIFF
--- a/src/input_handler.cc
+++ b/src/input_handler.cc
@@ -1353,7 +1353,7 @@ public:
                     auto cp = key.codepoint();
                     if (not cp or key == Key::Escape)
                         return;
-                    insert(RegisterManager::instance()[*cp].get(context()));
+                    insert(RegisterManager::instance()[*cp].get_for_pasting(context()));
                 }, "enter register name", register_doc.str());
             update_completions = false;
         }
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -708,7 +708,7 @@ template<InsertMode mode>
 void paste(Context& context, NormalParams params)
 {
     const char reg = params.reg ? params.reg : '"';
-    auto strings = RegisterManager::instance()[reg].get(context);
+    auto strings = RegisterManager::instance()[reg].get_for_pasting(context);
     const bool linewise = any_of(strings, [](StringView str) {
         return not str.empty() and str.back() == '\n';
     });
--- a/src/register_manager.cc
+++ b/src/register_manager.cc
@@ -53,6 +53,16 @@ void HistoryRegister::set(Context& context, ConstArrayView<String> values, bool 
 
 const String& HistoryRegister::get_main(const Context&, size_t)
 {
+    return last_entry();
+}
+
+ConstArrayView<String> HistoryRegister::get_for_pasting(Context&)
+{
+    return last_entry();
+}
+
+const String& HistoryRegister::last_entry() const
+{
     return m_content.empty() ? String::ms_empty : m_content.back();
 }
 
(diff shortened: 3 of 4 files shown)


caksoylar:
I think pasting the last entry as would be expected for a single selection makes sense, but AFAICT this breaks the use case of getting the full history with `":<a-p>`. Would it be possible to instead reverse the order when pasting history registers so that the last entry would still be pasted with a single selection, but the last two would be with two selections, etc.?

krobelus:
> AFAICT this breaks the use case of getting the full history with ":<a-p>

Wow, I've been running this for a while and never noticed.

Reversing history registers would work but of course it's a compatibilty break.
Probably it's also possible to fix this patch in another way. We just need to detect
if we are pasting with `<a-p>` or `p`, and only with `p` use the last value.

It looks that we can already use `%val{kak_main_reg_:}` to access the last vaue from history.
I don't think we can override the behavior of `":p` from kakscript but we can for `<c-r>`, which is what I use most of the time.
This is mostly enough for my use case; I think it should be the default.

```kak
map global insert <c-r> %{<a-;>: my-insert-register<ret>}
map global prompt <c-r> %{<a-;>: my-insert-register<ret>}
define-command -override my-insert-register -docstring %{
	Like <c-r>, but inserts the last (not first) value from history registers ":/|"
} %{
	# Copied from src/input_handler.cc::register_doc
	info -title "enter register name" 'Special registers:
	[0-9]: selections capture group
	%:     buffer name
	.:     selection contents
	#:     selection index
	_:     null register
	":     default yank/paste register
	@:     default macro register
	/:     default search register
	^:     default mark register
	|:     default shell command register
	::     last entered command
	'
	on-key %{
		evaluate-commands %sh{
			if expr index ':/|' "$kak_key" >/dev/null; then
			printf %s '
			evaluate-commands -save-regs t %{
				evaluate-commands "set-register t %%val{main_reg_%val{key}}"
				execute-keys <c-r>t
			}
			'
			else
			printf %s "execute-keys <c-r>${kak_key}"
			fi
		}
	}
}
```

TuM0xA-S:
1) why docs saying that colon register contains last command?
2) also `":p` for pasting first command is absolute useless and not makes any sense

Screwtapello:
> why docs saying that colon register contains last command?

Once upon a time, `":` contained *only* the last command. Later, it was changed to contain the command *history*, but since it still contained the last command (plus other, previous commands) I guess the docs weren't updated.

krobelus:
Fixed the problem with `":<a-p>`, this seems better than manually overriding `<c-r>`.
Reversing history registers might be a more elegant solution.

I didn't manage to add tests, I think this one used to work:

```diff
diff --git a/test/normal/prompt-history/cmd b/test/normal/prompt-history/cmd
new file mode 100644
index 00000000..d26b0b95
--- /dev/null
+++ b/test/normal/prompt-history/cmd
@@ -0,0 +1,3 @@
+:echo first  command<ret>
+:echo second command<ret>
+gkC":p
diff --git a/test/normal/prompt-history/in b/test/normal/prompt-history/in
new file mode 100644
index 00000000..1191247b
--- /dev/null
+++ b/test/normal/prompt-history/in
@@ -0,0 +1,2 @@
+1
+2
diff --git a/test/normal/prompt-history/out b/test/normal/prompt-history/out
new file mode 100644
index 00000000..e5e40c1c
--- /dev/null
+++ b/test/normal/prompt-history/out
@@ -0,0 +1,2 @@
+1echo second command
+2echo second command
```

PR #3110, opened 2019-09-28
Improve Elixir highlighting

endersstocker:
- [x] Add function highlighters for Elixir
- [x] Add operator highlighters for Elixir
- [x] Correctly identify Elixir atoms


colors/default.kak
```diff
@@ -6,6 +6,7 @@ face global type yellow
 face global variable green
 face global module green
 face global function cyan
+face global functionPrivate yellow
```
mawww: I am not sure about adding this as part of the default faces, you could introduce an `ElixirPrivateFunction` directly inside elixir.kak that would default to some modification of the function, (say `default+i@function` to add italic to whatever function is in the current colorscheme).

Phlogistique:
Hi!

Had not seen this PR, and created another one that also improves on Elixir hilighting: https://github.com/mawww/kakoune/pull/3462

Luckily they don't seem to collide.

@endersstocker will you be able to update your PR anytime soon?

Issue #1496, opened 2017-07-11
Fill regular buffer as Kakoune entry point


alexherbo2:

Delapouite:
I think I see what you mean.

Sometimes, you start kakoune, write some stuff in the default `*scratch*` buffer and want to save your progression `:w foo.txt`. You keep going writing stuffs and want to save again with a simple `:w`. It raises an error because the current buffer is still not linked to a file. So maybe the first `:w foo.txt` could transform the scratch buffer into a regular one.

danr:
I wish `:w <file>` would set the buffer name to `<file>` instead of retaining it. It is what vim does I think.

mawww:
@danr Would that do it for any buffer ? Say if I edit `a.txt` and do `:w b.txt`, would that rename the buffer to `b.txt` ? Or would that be the behaviour only for scratch buffers ?

lenormf:
Adding a side effect to this pure command would probably bend its orthogonality, but I suppose that's what most users use it for anyway?

maximbaz:
nvim does this only for `*scratch*` buffer, but to be honest I would prefer this behavior to happen in all cases, i.e. I would prefer to be switched to buffer `b.txt`.

alexherbo2:
How about filling a splash buffer as _New document_?

PR #3115, opened 2019-09-29
Automatically open a file buffer when saving a scratch buffer

codesoap:
I often find myself opening Kakoune to quickly take some notes and then decide to save them in a file.

With this pull request, Kakoune will automatically delete the scratch buffer and open the file in a buffer.  This seems convenient to me and also ensures that I won't accidentally quit Kakoune with unsaved changes.

**EDIT 2019-12-07:** `:rename-buffer -file /tmp/foo.txt` followed by `:write` can now be used, but I think this is still less convenient than simply using `:write /tmp/foo.txt`. Also, path auto-completion will only work with the latter.

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -341,6 +341,16 @@ struct CommandDesc
     void (*func)(const ParametersParser&, Context&, const ShellContext&);
 };
 
+Buffer* get_file_buffer(Context& context, String file, Buffer::Flags flags, bool file_must_exist)
+{
+    Buffer* buffer = file_must_exist ? open_file_buffer(file, flags)
+                                     : open_or_create_file_buffer(file, flags);
+    if (buffer->flags() & Buffer::Flags::New)
+        context.print_status({ format("new file '{}'", file), context.faces()["StatusLine"] });
+    buffer->flags() &= ~Buffer::Flags::NoHooks;
+    return buffer;
+}
+
 template<bool force_reload>
 void edit(const ParametersParser& parser, Context& context, const ShellContext&)
 {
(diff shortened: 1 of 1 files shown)


lenormf:
This change has already been implemented/suggested, and was denied:

https://github.com/mawww/kakoune/pull/2424#discussion_r222522559

codesoap:
I'm sad to hear this feature was denied before. Yet in #2424 it doesn't sound like the decision of @mawww was definitive.

> [...] if we decide it needs to do more for scratch buffers, changing their scratch nature to convert them to full "File" buffers is what I'd prefer [...].

I'm not 100% sure I understand this right, but with this pull request the scratch buffer will be deleted and a "full File buffer" will be opened instead. Is this what @mawww had in mind?

In case of doubt, I'd like to keep this pull request open.

alexherbo2:
https://github.com/mawww/kakoune/issues/1496#issuecomment-375243908

alexherbo2:
To sum up 3 scenarios, after saving a scratch buffer:

1. `write foo.txt`
2. Continue editing (It’s still the scratch buffer)
3. `write` → Error: `'write' cannot write a non file buffer without a filename`

---

1. `write foo.txt`
2. Continue editing
3. Do something else, forget or don’t realize it’s still a scratch buffer
3. `quit` → Ouch!

---

1. `write foo.txt`
2. `edit foo.txt`
3. Continue editing

mawww:
This got superseeded by `rename-buffer -file ...`

codesoap:
> This got superseeded by `rename-buffer -file ...`

I had updated the first comment on this PR, when `rename-buffer -file ...` became available and still think it would be more convenient and less surprising to most if the behaviour of `:write` changed.

mawww:
`rename-buffer -file` could be improved to provide filename completion, I agree its currently less usefull.

You should also be able to opt-in the behaviour you want by aliasing `write` to `def rename-write -params 1 %{ rename-buffer -file %arg{1} ; write; }` although this is not scratch buffer specific.

I understand this is slightly less convenient as a behaviour, but there is a tradeoff to be made between orthogonality of commands (`write` writes to file, `rename-buffer` changes the file a buffer is linked to). I currently prefer having simple and orthogonal (hence easily predictible) behaviours for those features.

codesoap:
> You should also be able to opt-in the behaviour you want by aliasing `write` to `def rename-write -params 1 %{ rename-buffer -file %arg{1} ; write; }` although this is not scratch buffer specific.

Thanks for the instructions, I'll give it a shot!

> I understand this is slightly less convenient as a behaviour, but there is a tradeoff to be made between orthogonality of commands [...]

I understand your point. Maybe I'll be happy with the command you suggested, or maybe I'll write myself some other script later. I'm sure there is a solution I can live with.

PR #3116, opened 2019-09-30
src: Serve a virtual filesystem via 9P to interact with Kakoune programmatically

eraserhd:
I don't really expect this to be accepted (although that would be cool); however, I'm making it a pull request for two reasons: 1. it is of interest to more than one person (maybe); 2. it has a few small improvements and innovations that might be useful to Kakoune.

I'm committing myself to maintaining this branch and rebasing it regularly.  As such, I'm willing to spend a lot of effort to rework any improvements here to contribute to Kakoune.

There are no new build dependencies, all the 9P was implemented in Kakoune itself.   To mount the filesystem, plan9port should be in the path and FUSE (e.g. osxfuse or Linux fuse) installed.  Then the ":mount" and ":unmount" commands will work, and the filesystem will be mounted at `$kak_opt_kakfs` (which is in `$XDG_RUNTIME_DIR`).

In the near term, I intend to add tests, which requires some refactoring of Kakoune's test script, and I'm dogfooding it, so I'll likely be adding new virtual files.


alexherbo2:
How to access Kakoune’s states?

lenormf:
Maintaining a PR is good, because users who want the feature will find it through the search field, and it's also very easy to grab a patch off of Github:

https://github.com/mawww/kakoune/pull/3116.patch

alexherbo2:
@eraserhd The [Makefile] of [plan9port] is a troll x)

[plan9port]: https://9fans.github.io/plan9port/
[Makefile]: https://github.com/9fans/plan9port/blob/master/Makefile

alexherbo2:
```
/run/user/1000/kakfs
└── 7350
   ├── buffers
   │  ├── 0x1c4ab70
   │  │  ├── buf_line_count
   │  │  ├── buffile
   │  │  ├── bufname
   │  │  ├── history_id
   │  │  ├── modified
   │  │  ├── opt_{name}
   │  │  └── timestamp
   │  └── 0x1c5c110
   │     ├── buf_line_count
   │     ├── buffile
   │     ├── bufname
   │     ├── history_id
   │     ├── modified
   │     ├── opt_{name}
   │     └── timestamp
   ├── global
   │  ├── buflist
   │  ├── client_list
   │  ├── config
   │  ├── opt_{name}
   │  ├── runtime
   │  ├── session
   │  ├── user_modes
   │  └── version
   └── windows
      └── client0
         ├── client
         ├── client_pid
         ├── cursor_byte_offset
         ├── cursor_char_column
         ├── cursor_char_value
         ├── cursor_column
         ├── cursor_display_column
         ├── cursor_line
         ├── opt_{name}
         ├── selection
         ├── selection_desc
         ├── selection_length
         ├── selections
         ├── selections_desc
         ├── selections_length
         ├── window_height
         └── window_width
```

alexherbo2:
@eraserhd The window scope is not correct.  Window is buffer + client.

eraserhd:
@alexherbo2 Hrmm. How about this: `windows/$client/buffer` is a directory that is identical to `buffers/$buffer_id`.  In fact, since I think I've just learned there are 9P extensions to serve symlinks even though Plan 9 abhorred them, so it would be a symlink when the 9P client supports it.

alexherbo2:
Yes.

Currently, there is no way to access a buffer state because of the buffer ID in the path.  You should add `%val(buffer_id)` state.

occivink:
`window` is a different scope from `buffer` so the symlink would be incorrect. 
However alex's point is that the `window` scope does not just exist for the buffer currently displayed. There is a `window` for each `{buffer, client}` combination that was ever displayed  so maybe you want to expose them via `windows/$client/$buffer`, and possibly a special `current_window`?

eraserhd:
I have never been able to wrap my head around window vs. client, so I finally figured it out in the source.  I don't want this PR to remap existing abstractions, so, what I'm going to do:

1. Declare that the directories `global`, `buffers/*` and `windows/*` are representations of the scopes, not representations of buffers and windows.
2. Mount buffer-scoped things in `windows/*`.
3. Change from `windows/$client_name/*` to `windows/$window_id/*`, where window id is like the buffer name.
4. Add `clients/$client_name/window` as an alias or symlink.

alexherbo2:
How do you programmatically access a given buffer or window state, since the ID is not exposed.

eraserhd:
@alexherbo2 I decided against using the window ids... Exposing all the windows is too hard, because the ClientManager hides them, so there's no point in not using the client name.

Regarding buffers, if you need to find one, you'll have to do something like:
```
for buffer in $kak_opt_kakfs/buffers/*; do
  if [ "$(cat $buffer/bufname)" = "*debug*" ]; then
    break;
  fi
done
```

(There's no value that can really be used as a short ID.)

I've just pushed commits exposing buffer-scope things under `windows/$client_name`, and also exposing `reg_*` and `main_reg_*` there.

alexherbo2:
You should at least expose the current buffer and window IDs in a variable imo.

Delapouite:
Is this great work on 9P tightly related to https://github.com/mawww/kakoune/pull/2927 ? Or does it even supersede it?

eraserhd:
> Is this great work on 9P tightly related to #2927 ? Or does it even supersede it?

I closed that one.  That was an experiment to see if I could get all the features I wanted without 9P.  The answer was "not quite".

eraserhd:
Closing this...  Thinking about the transactional issue that @mawww brought up.  Interestingly, I think the only way to make that work is to have a command language processed by the server over 9p, but I was hoping to obviate the command language in the first place, which I think would be the only way 9p could've been an improvement.

I'm going to try out [connect.kak](https://github.com/alexherbo2/connect.kak) to replace it.  It isn't quite what I'm looking for, but it is close.

mawww:
This was a great experiment, thanks a lot for exploring this design space.

losnappas:
> Thinking about the transactional issue that[...]

I see no reference to that here, please share? And please share your opinion on connect.kak (or similar) as replacement, I would appreciate.

On paper, I think "files as interface," as done here, is very appealing.

mawww:
I do not remember where the discussion happened, but the issue was that 9p is fundamentally asynchronous, so any information you get from the filesystem might be out-of-date when you try to act on it.

The command based interface blocks out any external mutation of state during its execution, but in order to achieve that with 9p we would need some kind of locking mechanism which would likely end up making 9p more cumbersome than the current approach.

In other words, when you execute a command in the command language, you can assume no state (buffer content, selection positions, ...) is changing behind your back except at well defined place (if you use `prompt` you cannot guarantee the buffer did not change between when `prompt` was called and when its callback gets executed). If you use a filesystem based interface you do not get such guaratees.

eraserhd:
^^ exactly that.  IIRC, @mawww pointed it out first, and I spent a lot of time thinking about how to make the filesystem abstraction _transactional_, and the answer is, it just isn't.  At least not without changing the abstraction in some fundamental way.

It might be possible to make transaction-based subdirectories, for example (it would take a lot of work in Kakoune), but there's still nothing built into the filesystem abstraction for "commit" or "rollback".  This could be done by command-file (which is already a kind of hack of the filesystem abstraction), but then we have to have some way to deal with abandoned transactions, which seem like they would be frequent occurrences, and these would block other actions.

Sorting these things out, you pretty much get back to a "command-pipe", which we already have.

It's interesting, though, because I think this transactional bit is a problem, and it's kind of an indictment of the whole Plan9 way of doing things, which is a sort of continuation of [Worse is Better](https://en.wikipedia.org/wiki/Worse_is_better) from Unix.

losnappas:
Alright yeah, I see the problem.

But do I? How often would I run into issues with transactionality? If I look at the way the [ad-editor](https://github.com/sminez/ad) does it, same issue exists there, but if you assume that plugins only react to specific user actions and you have the user themselves define the actions (the "plumbing rules" in ad), I don't think it'd all so often come up as a problem.

So while theoretically it is a problem, is it one in practice? Did you encounter this issue in this test run you did, already?

eraserhd:
You might not have a problem 99% of the time, but as a personal philosophical rule, I think we need to design software better.  I've paraphrased colleagues, "I don't understand why our system is broken, we designed all 100 components to be 99% reliable!"

That 1% is literally the bane of my existence.  I spend most of my job on that 1%.  Last week was my on call week, and the entire week was spent dealing with one-offs for odd cases.

lobre:
> It's interesting, though, because I think this transactional bit is a problem, and it's kind of an indictment of the whole Plan9 way of doing things, which is a sort of continuation of from Unix.

I never heard about this "limitation" of the plan9 filesystem design. By any chance, does anybody have references for posts that discuss this limitation? I read a few books about plan9 and this was never mentioned.

Also, acme relies only on this 9p filesystem protocol for extension. How can such a thing be considered by plan9 people back in the day? They seem to be really competent in OS intrinsic and I am stunned to see that this could be a problem that was not tackled. Because this issue would arise in many aspects of the the plan9 bits that use the filesystem as an abstraction...

I am referencing [this issue](https://github.com/sminez/ad/issues/11) in the `ad` editor just so that it creates a backlink. Maybe this can help to engage more brains in the discussion.

Screwtapello:
Like a lot of things about Plan 9, the limitations of any individual part don't matter because the rest of the system was designed to take them into account. All the cool things that Acme does with filesystem access are carefully chosen so that the "stale data" issue doesn't arise. There's probably a bunch of things they thought about adding (or at least, approaches they considered taking) that they avoided because of the asynchronous-filesystem-access thing.

The thing is, Kakoune was designed around synchronous access by launching shell-scripts, and so it has a bunch of things like highlighters that refer to specific buffer coordinates. That stuff works really well when you can only access it synchronously, but if you take all that existing stuff and expose it via 9p as-is, you're going to get problems.

It would be possible to make a version of Kakoune that used 9p instead of launching shell-scripts, but it would involve ripping out all the bits and pieces that don't work nicely with that approach. You'd probably wind up with something a lot closer to Acme.

lobre:
Thanks, @Screwtapello for the answer. Though, I don’t understand how Acme could mitigate those issues. From what you say, Kakoune embraces synchronous operations, while 9p is by definition asynchronous.

> All the cool things that Acme does with filesystem access are carefully chosen so that the "stale data" issue doesn't arise.

I don’t understand how those issues cannot arise. Let’s take the example of a plugin that would automatically add missing imports (such as what `goimports` is doing for Go). This would need to trigger a modification of the buffer as soon as a new import is detected to be missing.

Here, how can Acme handle that situation? If this is implemented externally by listening to the control sequence emitted by Acme through the 9p protocol for any modification, it means the plugin will then want to modify the buffer by writing to the corresponding file descriptor. In the meantime, the buffer could already have changed.

Is it the kind of situation where this asynchronous communication would be prone to errors? Meaning that Acme is not fitted for such plugins? If not, do you know how Acme mitigate that?

Screwtapello:
> Let’s take the example of a plugin that would automatically add missing imports (such as what goimports is doing for Go).

Unfortunately, I've never actually used Acme myself, let alone tried to write a plugin for it. However, I watched a video about how Acme works, and read [acme(4)](https://man.cat-v.org/plan_9/4/acme) describing the API Acme presents for other tools to integrate with.

Acme has an "events" pipe. If no process has that open, no process can react to events in a buffer, and the only interactions are presumably clobbering the entire buffer (writing to the `data` pipe) or Acme piping data to or from newly-launched processes, in which case Acme is responsible for merging the changes into the buffer.

On the other hand, if a process *does* read from the "events" pipe, it *can* react to things that happen inside the buffer... but it's not clear to me that more than one process can be processing events at the same time, or how much synthetically-generated events can affect the buffer contents.

Meanwhile, in Kakoune land, there can be a bazillion hooks at global, buffer, and window scopes, they can all launch processes in parallel that do computations and try to manipulate the editor's state. Kakoune is just a whole lot more *interactive* than Acme.

> Meaning that Acme is not fitted for such plugins? If not, do you know how Acme mitigate that?

Again, I haven't used Acme myself, but having watched the careers of the people behind it for a while, I get the impression that they are very happy to avoid problems by dropping features if it would be difficult to implement them reliably.

PR #3120, opened 2019-10-02
Better elm syntax highlighting

chaoky:
this will only highlight elm 0.19 (latest) properly
very close to how https://ellie-app.com/ highlights
- anything that starts with an upper case letter is a type
- anything that starts with a lower case letter is a function
- function declaration is just a function at the start of a line

note: the default color scheme looks too colorful


mawww:
could you add a copyright waiver commit to this ? (See the CONTRIBUTING file).

Issue #2056, opened 2018-05-18
Assertion failure with small kakoune windows


alyssais:
### Steps 

1. Enable line numbers
2. Ensure the current buffer has at least one line containing a character other than a newline
3. Shrink kakoune horizontally until even the line numbers are partially obscured (and the buffer is completely hidden)

### Outcome 

Kakoune crashes because this assertion fails:

https://github.com/mawww/kakoune/blob/57baad4afde8d1bab4040c5a5cbe8a5b367bceba/src/display_buffer.cc#L223

### Expected 

Kakoune gracefully handles this and just doesn't render the buffer.

PR #3122, opened 2019-10-04
test: Improve test runner

eraserhd:
This runs Kakoune asynchronously during tests, connecting the JSON UI via fifos.  If a file named "script" exists in the test directory, it is sourced, and it has available some shell functions:

* `_in` - send a JSON RPC input event.
* `out` - expect the given output event, and fail the test if something different is received.

After the script, Kakoune is quit and output files written using `kak -p`.

It seems to perform about the same as before, surprisingly.

Since `script` is a shell script, it can be extended by adding new functions, or shell can be written for odd cases.

No secret: This enables me to write tests on my 9p branch about the files that Kakoune serves.

Possible drawback: Since POSIX sh does not support non-blocking io, if you use `out` because you expect an event, but Kakoune doesn't produce it, the tests will hang.  (Bash has a `-t` extension for `read`.)


eraserhd:
Welp, it is way past my bedtime, but I found a problem - one test hangs on Mac.  No doubt because Linux has very large pipe buffers and Mac has absurdly small ones.  The test (`2499-html-regions-assert`) probably produces a lot of events.

Will fix tomorrow.

lenormf:
Why is this needed?

eraserhd:
@lenormf There are many sorts of issues which could not be tested before.  e.g.

* Behavior as a consequence of UI events which are not "Urgent".
* Behavior as a consequence of UI input events that happen in a particular sequence (e.g. after Kakoune has processed a screen refresh)
* Behavior as a consequence of commands piped to `kak -p`.
* Behavior as a consequence of the combination of the any of the above, meaning we could add test cases for some past deadlocks.

There was one test which piped UI events into Kakoune and verified that an assertion did not occur.  According to @mawww, it worked only because "Urgent" events are processed, but I'm not sure this is correct... it looks like the test was no longer testing anything.  I'll verify that and ensure the test would fail if a regression occurred.

lenormf:
Would this fix PR #871 ?

eraserhd:
@lenormf It looks like it will.

Also, it looks like the test runner now supports `enabled`, which is like that PR's `dependencies`.

eraserhd:
Alright, all tests pass, and I verified that test for #2056 still fails if I introduce the regression.

eraserhd:
> -ignore support adds a lot of code to only partially solve the problem (if a test emits more output, it will require -ignore as well... it would be better to drain those pipes).

I'm worried that specifying all the UI output in a test means that highlighter file tweaks will cause test failures.  I had two other different ideas wrt `-ignore`:

* Use wildcard or regex to allow the test to specify only what it cares about.
* Make `out` accept parameters like `-method` and `-params`, and only check what was specified. (This one seems complicated.)

A third option, might be to add `-until` option to out to skip until a match?  I think that's better for making the tests robust.

> Do you have use cases where script has to be a shell script, I am wondering if we could use a simple diff like format (>expected output,<input line), the file would then just be ui-script. (I am more tempted by this approach, but at the same time I cannot justify it that well).

In my `9p` branch, [the buffer tests](https://github.com/eraserhd/kakoune/blob/9p/test/protocol/9p/buffer/script), which uses [this new shell function](https://github.com/eraserhd/kakoune/blob/9p/test/run#L204-L229).

This is the worst one, and I hated assigning `buffer`, but I don't have a predictable way to name buffers.  Maybe I can just number them somehow and avoid that?

In any case, if we adopt a diff-like format, I can modify it in the 9p branch.  Your call. (Also `-until` in diff-like would be an ellipsis, which is kind of nice.)

> This is also a good opportunity to rename state to selections_desc and eventually to have a general env var exposition logic, what do you thin about that.

Sounds nice!  I added a general mechanism to this PR.  What do you think?

mawww:
> I'm worried that specifying all the UI output in a test means that highlighter file tweaks will cause test failures. I had two other different ideas wrt `-ignore`:

Oh, I was more thinking about having something that reads the remaining output and discard it as a standard feature of the test, removing the need for -ignore, at least when the `script` file does not exist.

> In my `9p` branch, [the buffer tests](https://github.com/eraserhd/kakoune/blob/9p/test/protocol/9p/buffer/script), which uses [this new shell function](https://github.com/eraserhd/kakoune/blob/9p/test/run#L204-L229).

I think my main issue with the current version is the ugly underscore in front of `_in`, could be renamed `ui_in` and `ui_out` or `ui_call` and `ui_expect`...

Another slight issue is the need for shell quoting the json, which can make it a bit ugly, but I guess thats ok.

> Sounds nice! I added a general mechanism to this PR. What do you think?

Yeah, that's much nicer than what we had.

Thanks a lot for this PR.

eraserhd:
Rename to `ui_in` and `ui_out`.

PR #3124, opened 2019-10-07
adding simple color scheme

nfultz:
This adds a color scheme that makes almost everything use the terminal's defaults. It is intended to be minimal, so that users can override individual settings in their personal configuration.

### DIFF
--- a/colors/plain.kak
+++ b/colors/plain.kak
@@ -1,0 +1,49 @@
+# Kakoune simple colors, mostly default
+
+# For default
+face global value default
+face global type default
+face global identifier default
+face global string blue
+face global keyword default
+face global operator default
+face global attribute default
+face global comment blue
+face global meta default
+face global builtin default
+
+
+# For default
+face global title default
+face global header default
+face global bold default
+face global italic default
+face global mono default
+face global block default
+face global link blue
+face global bullet default
+face global list default
+
+# builtin default
+face global Default default,default
+face global PrimarySelection white,blue
+face global SecondarySelection black,blue
+face global PrimaryCursor black,white
+face global SecondaryCursor white,blue
+face global PrimaryCursorEol default
+face global SecondaryCursorEol default
+face global LineNumbers default
+face global LineNumberCursor default
+face global MenuForeground default
+face global MenuBackground default
+face global MenuInfo default
+face global Information default
+face global Error default
+face global StatusLine default
+face global StatusLineMode default
+face global StatusLineInfo default
+face global StatusLineValue default
+face global StatusCursor default+r
+face global Prompt default
+face global MatchingChar default
+face global BufferPadding default


colors/simple.kak
```diff
@@ -0,0 +1,46 @@
...
+face link blue
+face bullet default
+face list default
+
+# builtin default
+face Default default,default
+face PrimarySelection white,blue
+face SecondarySelection black,blue
+face PrimaryCursor black,white
+face SecondaryCursor black,white
```
danr: Where is `{Primary,Secondary}CursorEol`? https://github.com/mawww/kakoune/blob/master/colors/default.kak#L34

Also: you have the same colour for `PrimaryCursor` and `SecondaryCursor`. Then with multiple selections of length the user won't be able to discern which is the primary cursor.

andreyorst: > Also: you have the same colour for PrimaryCursor and SecondaryCursor. Then with multiple selections of length the user won't be able to discern which is the primary cursor.

On a side note, do you really need to know which cursor is the primary one?

danr: Yes, to use some features reliably you do, some that come to mind are completions in insert mode, `vc`, `*`, `N`, `<space>`, `(`, well look for main in https://github.com/mawww/kakoune/blob/master/doc/pages/keys.asciidoc

andreyorst: Heh, never noticed anything like that, but you're right

nfultz:
> Could you add a copyright waiver commit (see the CONTRIBUTING file) ?

[Done](https://github.com/mawww/kakoune/pull/3124/commits/af504139cd2531db850a32b594309ab396d92f31)

colors/simple.kak
nfultz: [done](https://github.com/mawww/kakoune/pull/3124/commits/e1b81d4ac6e28b43045e8778e18c94b4db11dc29)

lenormf:
How about calling it `dummy.kak`?

nfultz:
> How about calling it `dummy.kak`?

I'm open to that, is there another simple.kak that this collides with? Otherwise I'll defer to @mawww to name it.

lenormf:
There's no collision, it's just that the colorscheme defers the highlighting to the terminal and does nothing, it's customary to call that "dummy".

nfultz:
I don't get any themes when I google for "dummy vim color" or "dummy emacs
color" so I think it is not so common to name themes that. Customary for
other components though.

Screwtapello:
It's not entirely a dummy theme, since it does do *something* (selection colours). In a browser, this might be considered a "reset" or "normalize" stylesheet. "Plain", as in "plain text", would be another good name.

mawww:
Hello,

I am fine with the `simple.kak` name, but I think `plain` is my preferred name.

The colorscheme still has some issues though.

* Its written against an old kakoune version, before faces were scoped. So it does not load on current master (you probably just want to add `global` after all face commands).

* I cannot see my cursor in the prompt, as `StatusCursor` is `default` maybe just reverse it (`default+r`) ?

* I cannot see `SecondaryCursor` as they have the same color as the `SecondarySelection` so it makes it impossible to know for sure the direction of a secondary selection (it usually matches the primary selection direction, but that is not guaranteed).

could you fix those ?

nfultz:
Done!

lenormf:
Please squash all the commits together (except the waiver one), if the branch is ready to be merged.

nfultz:
Done!

PR #3126, opened 2019-10-08
src: Allow renaming the current session with an already used name

lenormf:
This commit allows clients to make sure they are part of a given session
by implementing the `:rename-session!` command.

The `:rename-session` command will error out if the given name is already in
use, the new `:rename-session!` command will merely remove the old session
socket and join the new session seamlessly.

### DIFF
--- a/src/remote.cc
+++ b/src/remote.cc
@@ -832,7 +832,7 @@ Server::Server(String session_name)
     m_listener.reset(new FDWatcher{listen_sock, FdEvents::Read, accepter});
 }
 
-bool Server::rename_session(StringView name)
+bool Server::rename_session(StringView name, bool force)
 {
     if (not all_of(name, is_identifier))
         throw runtime_error{format("invalid session name: '{}'", name)};
@@ -840,10 +840,9 @@ bool Server::rename_session(StringView name)
     String old_socket_file = session_path(m_session);
     String new_socket_file = session_path(name);
 
-    if (file_exists(new_socket_file))
-        return false;
-
-    if (rename(old_socket_file.c_str(), new_socket_file.c_str()) != 0)
+    if ((file_exists(new_socket_file) or
+         rename(old_socket_file.c_str(), new_socket_file.c_str()) != 0)
+         and not force)
         return false;
 
     m_session = name.str();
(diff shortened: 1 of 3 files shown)


lenormf:
My usecase if for the [Kakoune-based Git merge tool](https://github.com/lenormf/kakmerge), which spawns a certain number of clients that we need be within the same session.

My user configuration automatically renames the current session name into a human readable screen, which means that the clients spawned by the merge tool (`kak -c …`) end up being in unpredictable sessions.

With this commit, the merge tool can safely spawn clients that will be guaranteed to be part of a given session (`kak -e "rename-session! …"`).

eraserhd:
I think `rename-session` doesn't do what you think it does.  It causes the server to rename its socket, but doesn't cause a client to join a different session, or sessions to be merged.  Meaning that use of this new command will cause the original server to be orphaned, but still operational, and two Kakoune sessions with the same name will be running, and one will be permanently inaccessible.

src/remote.cc
```diff
@@ -832,18 +832,17 @@ Server::Server(String session_name)
...
     String old_socket_file = session_path(m_session);
     String new_socket_file = session_path(name);
 
-    if (file_exists(new_socket_file))
-        return false;
-
-    if (rename(old_socket_file.c_str(), new_socket_file.c_str()) != 0)
+    if ((file_exists(new_socket_file) or
+         rename(old_socket_file.c_str(), new_socket_file.c_str()) != 0)
+         and not force)
```
mawww: So, if the file exist and we are forcing, we wont rename the scoket file, and but we will still believe that we did. That means we will remove the wrong socket (the one from another session) in close_session, and that any use of $kak_session will refer to another server, that does not look like desired behaviour.

lenormf:
Refer to my comment which contains the goal of this change.

Is there no way to have clients join an already existing session at runtime?

mawww:
I dont even know what would the semantics of a client joining another session at runtime be, all state is stored server side, so the most basic client reconnection feature would be equivalent to quitting this client and starting another on the other session. That makes me think you want to preserve some state that you view as the client one, but its tricky, the other session server might not have the current buffer open, it might not even be running on the same machine if the socket has been mounted through ssh...

`kak -c <session>` guarantees that a client is part of that session.

lenormf:
Alright thanks

PR #3127, opened 2019-10-08
Add unambiguous titles for infobox "enter combining operator"

### LINKED ISSUE #3089: add unambiguous titles for infobox "enter combining operator"
Right now, both `<a-z>` and `<a-Z>` display the same infobox: 
```
enter combining operator

a: append lists
u: union
i: intersection
<: select leftmost cursor
>: select rightmost cursor
+: select longest
-: select shortest
```

It would be a great help to have a different title for each commands. Because of my fat fingers I often mistype `<a-Z>` because it involves pressing `alt` and `shift` modifiers at the same time.
So having different title helps confirming that the right keys have been pressed before proceeding to apply the combining operator.

I suggest:

`<a-z>`
```
combine selections from register

a: append lists
u: union
i: intersection
<: select leftmost cursor
>: select rightmost cursor
+: select longest
-: select shortest
```

`<a-Z>`
```
combine selections to register

a: append lists
u: union
i: intersection
<: select leftmost cursor
>: select rightmost cursor
+: select longest
-: select shortest
```

Delapouite:
fix #3089

### DIFF
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -1849,7 +1849,7 @@ void combine_selection(const Buffer& buffer, Selection& sel, const Selection& ot
 }
 
 template<typename Func>
-void combine_selections(Context& context, SelectionList list, Func func)
+void combine_selections(Context& context, SelectionList list, Func func, StringView title)
 {
     if (&context.buffer() != &list.buffer())
         throw runtime_error{"cannot combine selections from different buffers"};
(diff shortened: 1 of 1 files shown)


src/normal.cc
```diff
@@ -1848,7 +1848,7 @@ void combine_selection(const Buffer& buffer, Selection& sel, const Selection& ot
     }
 }
 
-template<typename Func>
+template<bool to_reg, typename Func>
```
mawww: I'd prefer a StringView for the info box title to be passed there rathen than a bool template parameter, my issue here is that its the fact that we are saving to/from register is mostly controlled by Func, not by to_reg, which only selects the mesage.

Passing a normal StringView parameter would be fine.

Delapouite:
Thanks for the feedback. I've updated the PR accordingly.

PR #3133, opened 2019-10-12
Make clojure-insert-ns detect the project root

eraserhd:
I am moving to one global Kakoune session, and its working directory is now one level up from my projects, so `clojure-insert-ns` started getting confused.

### DIFF
--- a/rc/filetype/clojure.kak
+++ b/rc/filetype/clojure.kak
@@ -222,13 +222,35 @@ e.g. '(src|test|dev)'} regex clojure_source_directories '(src|test|dev)'
 
 define-command -docstring %{clojure-insert-ns: Insert namespace directive at top of Clojure source file} \
     clojure-insert-ns %{
-    evaluate-commands -draft %{
-        execute-keys -save-regs '' 'gk\O' "%val{bufname}" '<esc>giZ'
-        try %{ execute-keys 'z<a-l>s\.clj[csx]?$<ret><a-d>' }
-        try %{ execute-keys 'z<a-l>s^' "%opt{clojure_source_directories}" '/<ret><a-d>' }
-        try %{ execute-keys 'z<a-l>s/<ret>r.' }
-        try %{ execute-keys 'z<a-l>s_<ret>r-' }
-        execute-keys 'z<a-l>\c(ns <c-r>")<ret><esc>'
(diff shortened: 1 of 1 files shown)


eraserhd:
@mawww Any thoughts on this?

mawww:
A bit on the fence about it, it seems like both the current and proposed behaviour are pretty brittle heuristics to determine the actual namespace name, the proposed one likely works better for your use case, but it does complexify the implementation quite a bit, which I am a bit unhappy about.

eraserhd:
That makes sense.

I was not happy with so much code for this purpose.  I think of Kakoune's configuration language as intending to be thin glue, but often can't figure out how to make small utilities when things get complicated, so...

Thought experiment: Remove this functionality from Kakoune and create a new plugin in a new repo, say `clojure-editing-tools`, with a command-line program `clojure-new-file` and one `kak-configure-clojure-new-file`.  A bit of Kakoune code scans the path for `kak-configure-*` and evaluates commands printed by them the way kak-lsp does.

This way, small utilities can be installed in the path with a normal sort of build and install routine.

mawww:
> Thought experiment: Remove this functionality from Kakoune and create a new plugin in a new repo, say `clojure-editing-tools`, with a command-line program `clojure-new-file` and one `kak-configure-clojure-new-file`. A bit of Kakoune code scans the path for `kak-configure-*` and evaluates commands printed by them the way kak-lsp does.

What would `kak-configure-clojure-new-file` provide that a `clojure-editing-tools.kak` could not ? (A `.kak` script can find its path with `$kak_source`, which allows it to find support scripts/binaries there).

eraserhd:
> What would kak-configure-clojure-new-file provide that a clojure-editing-tools.kak could not ?

The only thing is provides is a portable way to install, find and load scripts.  And after some thought, having a program print the config is actually superfluous, and installing a .kak is the way to go.

However... I think the plugins really should install these files into `$prefix/share/kakoune/autoload`, as installing them into `~/.config/kakoune/autoload` could break the user's config, and `~/.config` could be managed by a dotfile manager that wants the user to manipulate a github repo (my old method) or which makes things read-only (my current method).

The only problem is that this requires some code in kakrc to load things.

PR #3134, opened 2019-10-13
Support git -C

eraserhd:
To support one Kakoune session for multiple projects, I want to
run `git -C project log` and `git -C project diff`.  This adds
support for that option, including completions.

### DIFF
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -25,15 +25,25 @@ define-command -params 1.. \
 All the optional arguments are forwarded to the git utility
 Available commands:\n  add\n  rm\n  blame\n  commit\n  checkout\n  diff\n  hide-blame\n  hide-diff\n  init\n  log\n  show\n  show-diff\n  status\n  update-diff'} \
   -shell-script-candidates %{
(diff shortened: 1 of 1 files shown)


mawww:
How about instead of this we introduce a `gitcmd` option, then you can set it to `git -C blah` for some buffers, and `git -C foo` for others ?

eraserhd:
That doesn't really fit my workflow, since I keep Kakoune one directory up from my projects now.  So, I might be looking at a file from project A when I want to see a diff from project B.  (And really, I'm trying to do away with the clean separation between projects.)

I have another idea, though, that might resolve this... I think the only reason I want this is to browse the output in Kakoune, and I'd rather run shell commands in the shell.  So...

I've been thinking about a script (in my head it's called `scratch`, but that's not technically right), such that I can `git diff foo.c |scratch`.  I guess it would make a fifo, though.  I think this is probably generally useful, and I can nicely set an option on the buffer to indicate the working directory, for purposes of plumbing clicks and jumping to files.

The only thing I haven't figured in my head is how to detect common filetypes.

PR #3142, opened 2019-10-17
Support for JSON object key highlighting

cipharius:
Since JSON object keys are just strings in specific context, they are highlighted same as literal JSON strings. This problem was originally brought up by @Delapouite in [kakoune forum thread](https://discuss.kakoune.com/t/json-highlighters-for-keys/758).

Following images are taken from previously mentioned discussion thread to demonstrate the desired effect:

Current JSON highlighter

Desired JSON highlighter

I have tested the proposed highlighter using the following test JSON:
```json
{
    "test1": true,
    "test2\\" : ["a\"123:s true", "a\\\\bc\"defg\\"],
    "t\"est\\"
    :123
}
```

### DIFF
--- a/rc/filetype/json.kak
+++ b/rc/filetype/json.kak
@@ -32,11 +32,10 @@ provide-module json %(
 # Highlighters
 # ‾‾‾‾‾‾‾‾‾‾‾‾
 
-add-highlighter shared/json regions
-add-highlighter shared/json/code default-region group
-add-highlighter shared/json/string region '"' (?<!\\)(\\\\)*" fill string
-
-add-highlighter shared/json/code/ regex \b(true|false|null|\d+(?:\.\d+)?(?:[eE][+-]?\d*)?)\b 0:value
+add-highlighter shared/json group
+add-highlighter shared/json/string regex (?<!\\)("(?:[^"\\]|\\\\|\\")*") 1:string
+add-highlighter shared/json/key regex (?<!\\)("(?:[^"\\]|\\\\|\\")*")\s*: 1:keyword
+add-highlighter shared/json/literal region '"' (?<!\\)(\\\\)*" regex \b(true|false|null|\d+(?:\.\d+)?(?:[eE][+-]?\d*)?)\b 0:value
 
 # Commands
 # ‾‾‾‾‾‾‾‾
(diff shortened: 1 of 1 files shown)


cipharius:
Just noticed some issues with the modified literals, resolving that now.

rc/filetype/json.kak
```diff
@@ -32,11 +32,11 @@ provide-module json %(
...
-add-highlighter shared/json regions
-add-highlighter shared/json/code default-region group
-add-highlighter shared/json/string region '"' (?<!\\)(\\\\)*" fill string
-
-add-highlighter shared/json/code/ regex \b(true|false|null|\d+(?:\.\d+)?(?:[eE][+-]?\d*)?)\b 0:value
+add-highlighter shared/json group
+add-highlighter shared/json/value regions
+add-highlighter shared/json/value/string region '"' (?<!\\)(\\\\)*" fill string
+add-highlighter shared/json/value/ default-region regex \b(true|false|null|\d+(?:\.\d+)?(?:[eE][+-]?\d*)?)\b 0:value
+add-highlighter shared/json/key regex (?<!\\)("(?:[^"\\]|\\\\|\\")*")\s*: 1:keyword
```
mawww: Not fully happy about that overwritting strings approach, I wonder if we could have something that matches language semantics more closely (probably using a recursive `json-object` shared highlighter), but at the same time it seems a proper solution would get really complex.

cipharius: You're right, this isn't a clean solution.

I was thinking about recursive highlighter previously as well. It could be possible splitting buffer in list and object regions, but I wasn't sure whether kakoune can handle recursive highlighters with regions within regions.

I can give the recursion a try, perhaps it won't be as complex as it seems.

mawww: Any success with recursion ? Kakoune should handle recursive highlighting fine, if it does not thats a bug.

cipharius: When I tried to do recursive regions, kakoune refused to recurse once it detected recursive region highlighter definition. Here is the sanity check that trips: https://github.com/mawww/kakoune/blob/7438f23b9beddc42b6561fe8be3f953aff2f73b1/src/highlighters.cc#L1636

mawww: That means your recursive highlighter is running in the same range as a parent one, so it is going to give the same result (resulting in an infinite loop), recursion is only going to work if you apply it on a sub-range of the buffer (in a region highlighter for example).

PR #3145, opened 2019-10-19
Add syntax highlighting for F#

nhey:
Hi,

Currently there is no syntax highlighting for F#. This commit adds such.

I made this a while ago and thought it might be useful for other people. It is largely based on https://github.com/fsharp/vim-fsharp/blob/master/syntax/fsharp.vim, although probably not as extensive.

### DIFF
--- a/rc/filetype/fsharp.kak
+++ b/rc/filetype/fsharp.kak
@@ -1,0 +1,142 @@
+# https://fsharp.org/
+#
+
+# Detection
+# ‾‾‾‾‾‾‾‾‾
+
+hook global BufCreate .*[.](fs|fsx|fsi) %{
+    set-option buffer filetype fsharp
+}
+
+# Initialization
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+hook global WinSetOption filetype=fsharp %{
+    require-module fsharp
+
+    # indent on newline
+    hook window InsertChar \n -group fsharp-indent fsharp-indent-on-new-line
+
+    hook -once -always window WinSetOption filetype=.* %{ remove-hooks window fsharp-.+ }
+}
+
+hook -group fsharp-highlight global WinSetOption filetype=fsharp %{
+    add-highlighter window/fsharp ref fsharp
+    hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/fsharp }
+}
+
+provide-module fsharp %§
+
+# Highlighters & Completion
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+add-highlighter shared/fsharp regions
+add-highlighter shared/fsharp/code default-region group
+add-highlighter shared/fsharp/docstring region \(\*(?!\)) (\*\)) regions
+add-highlighter shared/fsharp/double_string region @?" (?<!\\)(\\\\)*"B? fill string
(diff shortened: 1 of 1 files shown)


alexherbo2:
Looks good.  Maybe single quote the first argument of `printf` and always add `\n`.

nhey:
> Looks good. Maybe single quote the first argument of `printf` and always add `\n`.

Thanks. I force-pushed these changes.

nhey:
I'm not sure why the CI job failed?

alexherbo2:
Can you remove unwanted whitespaces by doing:

```
%<a-s>_s\h{2,}<ret>c<space><esc>
```

rc/filetype/fsharp.kak
```diff
@@ -0,0 +1,126 @@
...
+}
+
+# Initialization
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+hook global WinSetOption filetype=fsharp %{
+    require-module fsharp
+
+    # cleanup trailing whitespaces upon exiting insert mode
+    hook window ModeChange insert:.* -group fsharp-trim-indent %{ try %{ execute-keys -draft \; <a-x> s ^\h+$ <ret> d } }
```
lenormf: We tend to avoid this hook (which might still be lying around in some form or another in the codebase). Consider the following usecase:

* yank some text
* start writing `let x = ` (trailing whitespace)
* leave insert mode, which triggers the hook
* paste

The outcome would be `let x=[text]`, when what we wanted was really `let x = [text]` (whitespace around the operator).

I think a better way of handling whitespace errors is to use a formatter (c.f. `:format`, `formatcmd`); or you could set a highlighter in your configuration that shows trailing whitespace:

```
add-highlighter global/ regex '\h+$' 0:Error
```

nhey: Right, I have wondered about this behaviour. Most filetypes seem to have it -- I have grown used to pasting _after_ (`p`), s.t. the paste lands on a new line, and then concatenating the lines with `<s-J>`.

I will remove this line from the file and keep the formatter option in mind.

pema99:
There is an issue with the implementation of multiline comments using `(*` and `*)`. Consider the following line of code: `let a = (*) 2 5`. The * operator in paranthesis is wrongly interpreted as the start of a multiline comment, and all following lines are highlighted as comments.

pema99:
I also noticed some issues with some other operators, summed up in the following image.

nhey:
@pema99 thanks for pointing this out. Things that are incorrectly highlighted definitely have to be fixed before merging. As for operators lacking I probably won't find the time to add this; my ambition is not that this commit adds full F# syntax highlighting, but rather that it adds some where there is none (not that you implied these must be added)

I'll add a WIP tag for now :-)

pema99:
Cheers, and thanks for the work so far, I will definitely be using this.

nhey:
by the efforts of @pema99 this PR is now up for review :-)

nhey:
@lenormf Should I squash the commits to something appropriate now, or wait for review?

lenormf:
If the file is ready, you can squash all the commits (*except* the waiver commit) into the first one. Then highlight @mawww to have him review/merge the changes.

nhey:
I have squashed the commits adding @pema99 as coauthor while keeping his waiver commit.

@mawww This PR is ready for review

PR #3146, opened 2019-10-19
colors: Implement a colorblind-friendly light colorscheme

lenormf:
Hi,

I came up with a light colorscheme that should hopefully play nice with colorblind users. There's some information loss when going from one type of colorblindness to another (the user has to rely on shades and font style), but there shouldn't be two different kinds of information highlighted with two colors that look identical.

Let me know if things need be fixed!

"Normal" view:

Blue-weak:

Blue-blind:

Green-weak:

Green-blind:

Red-weak:

Red-blind:

Monochromacy:

### DIFF
--- a/colors/kaleidoscope-light.kak
+++ b/colors/kaleidoscope-light.kak
@@ -1,0 +1,130 @@
+# Kaleidoscope: colorblind-friendly light colorscheme
+# https://personal.sron.nl/~pault/
+
+evaluate-commands %sh{
+    # NOTE: tone down black and white for aesthetics,
+    # ideally those should be pure #000 and #FFF
+    black="rgb:1C1C1C"
+    white="rgb:FDFDFD"
+
+    # Regular text
+    bright_blue="rgb:4477AA"
+    bright_cyan="rgb:66CCEE"
+    bright_green="rgb:228833"
+    bright_yellow="rgb:CCBB44"
+    bright_red="rgb:EE6677"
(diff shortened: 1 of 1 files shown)


alexherbo2:
Feat. @dmerejkowsky

vbauerster:
I think `wrap` highlighter shouldn't share its colors with `StatusLineInfo` face, it deserves its own face. See line 9.
[](https://postimg.cc/vxmyC90g)

lenormf:
Thanks for the feedback, I've created #3149 accordingly.

dmerejkowsky:
Hello,

I'm the color-blind kakoune user who was mentioned in this PR.

The default light them is quite readable for me. Good job there :)

Just a question: do you think it would be possible to have a dark color scheme too?

If not, I would loke to get some pointers and try and implement it myself.

lenormf:
Glad to hear it's usable! Note that other faces used by misc. scripts might not play well with it, e.g. the green/red (the two statistically most problematic colors) faces for diffs.

I'd like to trying making a dark theme as well, it might just be much trickier since in the worst case (monochromacy), all colors will be brought down to a grey-scale color, which can make the theme barely more relevant than a pure black/white one.

If you want to give it a try, you can just copy the colors from `kaleidoscope-light.kak` and come up with something you like by changing the values of the faces.

lenormf:
Related #3155

PR #3147, opened 2019-10-19
Accept tmux pre-release versions

eraserhd:
Currently, the master branch of tmux reports `tmux next-3.1`, which fails the version check logic.

This also removes one fork during boot.


lenormf:
Is it possible some users will run a Tmux that returns `X.Y-rc` as well? There's a `3.0-rc` branch in the current Github repository.

eraserhd:
@lenormf It looks like the current version on that branch is `3.0-rc5`.  I'll update.

PR #3156, opened 2019-10-22
src: Create a <semicolon> named key

lenormf:
This commit allows using the <semicolon> expansion in commands, instead
of `\;`.

It makes commands look more elegant, and prevents new-comers from
falling into the trap of using <a-;> without escaping the semicolon.

### DIFF
--- a/rc/filetype/elixir.kak
+++ b/rc/filetype/elixir.kak
@@ -70,13 +70,13 @@ define-command -hidden elixir-indent-on-new-line %{
         # copy -- comments prefix and following white spaces
         try %{ execute-keys -draft k <a-x> s ^\h*\K--\h* <ret> y gh j P }
         # preserve previous line indent
-        try %{ execute-keys -draft \; K <a-&> }
+        try %{ execute-keys -draft <semicolon> K <a-&> }
         # indent after line ending with:
 	# try %{ execute-keys -draft k x <a-k> (do|else|->)$ <ret> & }
 	# filter previous line
         try %{ execute-keys -draft k : elixir-trim-indent <ret> }
         # indent after lines ending with do or ->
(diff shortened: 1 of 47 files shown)


rc/filetype/elixir.kak
```diff
@@ -70,13 +70,13 @@ define-command -hidden elixir-indent-on-new-line %{
...
         # preserve previous line indent
-        try %{ execute-keys -draft \; K <a-&> }
+        try %{ execute-keys -draft <semicolon> K <a-&> }
         # indent after line ending with:
 	# try %{ execute-keys -draft k x <a-k> (do|else|->)$ <ret> & }
 	# filter previous line
         try %{ execute-keys -draft k : elixir-trim-indent <ret> }
         # indent after lines ending with do or ->
-        try %{ execute-keys -draft \\; k x <a-k> ^.+(do|->)$ <ret> j <a-gt> }
+        try %{ execute-keys -draft <semicolon> k x <a-k> ^.+(do|->)$ <ret> j <a-gt> }
```
lenormf: Does anybody know if this is a typo (double backslash), or if the intent was to disable hooks before issuing a semicolon?

This looks like a failed copy-paste over from `elm.kak`, which stores the command into a double quoted string, which would explain the double escaping.

occivink:
I think quoting the parameters of `exec` would be more elegant, `exec ';'` already works without escaping

lenormf:
That works, but then you have to quote the whole key with the alt-based equivalent, `'<a-;>'`, which does not look elegant.

And if you're using those keys in an already quoted string, you have to think about using the right quoting separators as well, which is one more mental obstacle when trying to interactively use keys (e.g. while running `exec` on the current buffer).

PR #3161, opened 2019-10-25
wip: add tools.asciidoc

Delapouite:
Hi

The creation of this page has been discussed briefly on IRC.
I'm starting this doc about tools available in `rc/tools` to get the ball rolling.

It's a **WIP** with only a broad skeleton for now.

For some obvious commands, the description will mostly repeat the content of the associated docstring (if it exists, if not I'll make other PRs).

For more involved integration, larger explanations will be added.

Doc for tools specific to a lang (rust, python) will be added later.

### DIFF
--- a/doc/pages/tools.asciidoc
+++ b/doc/pages/tools.asciidoc
@@ -1,0 +1,145 @@
+= Tools
+
+== Description
+
+Easily making programs cooperate with each others is one of the main strength
+of the Unix environment. Kakoune is designed to integrate nicely with
+tools available on a POSIX system.
+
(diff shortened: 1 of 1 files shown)


lenormf:
I can't find the PR, I'd started working on this a while back. I think we should rather have this documentation auto-generated on release (in C++? with a shell script? with a Kakoune script?). Modifications to the tools come fairly often, which puts a high risk of desync between this hypothetical documentation and the actual implementation.

I also think the documentation should be split in individual files, it would make automating the generation easier.

pickfire:
Is this dead?

By the way, happy chinese new year!

PR #3163, opened 2019-10-25
fix errors on markdown files introduced by ec31d839

bitonic:
ec31d839 introduced a bug whereby opening a markdown file results
in the following error:

    error running hook NormalIdle()/markdown-load-languages: no such command: 'markdown-load-languages'

I'm not an expert in kakoune scripting, but I think the problem is
that `markdown-load-languages` is in its own module, while the hook
command is not.


bitonic:
wrong naming, see https://github.com/mawww/kakoune/pull/3164

PR #3164, opened 2019-10-25
fix errors on markdown files introduced by 89b50daa

bitonic:
89b50daa introduced a bug whereby opening a markdown file results
in the following error:

    error running hook NormalIdle()/markdown-load-languages: no such command: 'markdown-load-languages'

I'm not an expert in kakoune scripting, but I think the problem is
that `markdown-load-languages` is in its own module, while the hook
command is not.


mawww:
It is strange that you get this error, as the hook on line 14 of markdown.kak should ensure that the markdown module is loaded (at line 15) before those other hooks are registered, and hence get a chance to run.

Hooks at line 14 and 21 should be executed in their definition order, so things should work fine. I dont seen any errors here opening a markdown file.

Could you try to get a more precise reproduction case ?

Issue #2724, opened 2019-02-05
Proposal: Select by character or display indices


eraserhd:
Expansions such as `%val{selection_desc}` expose columns in byte index within the line, and commands such as `select` take selection in terms of byte index within line.

This makes integrations hard.  So far as I've experienced, byte-index columns aren't useful external to Kakoune.  So far, I have unicode issues with `parinfer-rust` that I will have to fix by converting coordinates to characters, and I have unicode issues with my new selection tool based on this as well.

The selection tool is written in Clojure, therefore using Java strings, so reading the file produces UTF16 characters anyway, so the byte index of UTF8 can't be used directly for converting, and if there are multiple encodings of a character in UTF8, I'll have to guess which way it occurred to interpret the bytes.

Further, converting these requires extra content from the buffer.  If I have a $kak_selection that start > column 1, I need to know the entire text of the starting line in addition to the text of the selection I probably want.  We don't have access to the text of the buffer outside of selections unless we run some keys in draft mode and save the value to an option.

So, I propose that byte indices are not exposed for scripting.  Instead, all expansions and commands use character indices.  The user can split diacritics and other compound characters, but not multibyte character encodings.

No matter how it is sliced, this would be a bit of work.  There's a couple ways to do it.  One way is to audit everything at user-interface layer (commands and expansions), and ensure they provide the right data.  Another way is to encapsulate this at the buffer layer and not expose byte indices to the rest of Kakoune.

IMHO, both add internal complexity.  Both reduce interface complexity.  The latter way contains the complexity better, but is harder.  It would be nice to figure out how to break it down into smaller steps.

(Note: It would be possible to make parallel expansions that use char indices - some already exist - and add options to commands like `select`.  I'm proposing the big change here because I think it'll be better for Kakoune.)

Screwtapello:
This affects kak-lsp too: https://github.com/ul/kak-lsp/issues/98

"character" is kind of a hazy concept in the world of Unicode, and the official documentation tries to avoid it. Some common alternatives include:

- bytes in a particular encoding, as Kakoune does, so U+0041 LATIN CAPITAL LETTER A is one byte, U+0107 LATIN SMALL LETTER C WITH ACUTE is two, and U+1F404 COW is four.
- UTF-16 code units, as Java, and Win32 do, so U+0041 is one unit, U+0107 is also one unit, and U+1F404 is two
[...]

PR #3167, opened 2019-10-28
Merge selection list parsing into selection_list_from_strings

eraserhd:
Two places built `SelectionList`s from a list of strings.
The one used by `:select` ignored the timestamp when clamping the selections, so I extracted the code from `read_selections_from_register` and added added a check that `main` is in bounds.
(Otherwise, this would have fell through to a `kak_assert` in `check_invariant`.)

I actually didn't experience the clamping issue.  I'm looking at the code to
see about implementing something a la the proposal at the end of #2724.

### DIFF
--- a/src/selection.hh
+++ b/src/selection.hh
@@ -94,9 +94,6 @@ struct SelectionList
     SelectionList(Buffer& buffer, Vector<Selection> s);
     SelectionList(Buffer& buffer, Vector<Selection> s, size_t timestamp);
 
-    struct UnsortedTag {};
-    SelectionList(UnsortedTag, Buffer& buffer, Vector<Selection> s, size_t timestamp, size_t main);
-
     void update(bool merge = true);
 
     void check_invariant() const;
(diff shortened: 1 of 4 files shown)


src/selection.hh
```diff
@@ -163,7 +160,30 @@ Vector<Selection> compute_modified_ranges(Buffer& buffer, size_t timestamp);
 String selection_to_string(const Selection& selection);
 String selection_list_to_string(const SelectionList& selection);
 Selection selection_from_string(StringView desc);
-SelectionList selection_list_from_string(Buffer& buffer, ConstArrayView<String> descs, size_t timestamp);
+
+template<class StringArray>
+SelectionList selection_list_from_strings(Buffer& buffer, StringArray&& descs, size_t timestamp, size_t main)
+{
+    if (timestamp > buffer.timestamp())
+        throw runtime_error{"register '{}' refers to an invalid timestamp"};
```
mawww: this format string seems wrong

eraserhd: Weird.  Updated and verified.

PR #3173, opened 2019-11-02
src: Some recommended fixes from cppcheck

eraserhd:
I ran it to see if there was dead code, and there was only one function, which is quite nice.

Most of the changes were making functions static or const, which I figured could have some mild, nice performance implications, if not just code readability.

It also found one outright bug parsing OSI escape sequences and an unnecessary BufferCoord copy, both of which I fixed.

There are two scary warnings left which I'm not quite sure how to deal with:

```
src/commands.cc:2431:13: note: Calling function 'on_next_key_with_autoinfo', 4th argument 'lock?format("{} (lock)",mode_name):mode_name' value is <Moved>
    }, lock ? format("{} (lock)", mode_name) : mode_name,
            ^
```

I think this is right, and it's a false positive.  I think everything that uses the `StringView` title runs before the call returns, and I'm just weirded out trying to figure out how a C++ can generate the code to conditionally destroy an object created in a ternary for a parameter after the call returns.  It is supposed to, right?

```
src/highlighters.cc:63:66: note: Missing bounds check for extra iterator increment in loop.
        for (auto atom_it = line.begin(); atom_it != line.end(); ++atom_it)
```

This one, there's an extra `++atom_it` on 79, which doesn't check if we've reached past-the-end.  I'm not sure if we know that it won't.

### DIFF
--- a/src/client.cc
+++ b/src/client.cc
@@ -212,7 +212,7 @@ void Client::redraw_ifn()
     if (m_ui_pending == 0)
         return;
 
-    auto& faces = context().faces();
+    const auto& faces = context().faces();
 
     if (m_ui_pending & Draw)
         m_ui->draw(window.update_display_buffer(context()),
--- a/src/highlighters.cc
+++ b/src/highlighters.cc
@@ -899,9 +899,9 @@ struct WrapHighlighter : Highlighter
         }
 
         return pos;
-    };
+    }
 
-    ColumnCount line_indent(const Buffer& buffer, int tabstop, LineCount line) const
+    static ColumnCount line_indent(const Buffer& buffer, int tabstop, LineCount line)
     {
         StringView l = buffer[line];
         auto col = 0_byte;
@@ -947,7 +947,7 @@ struct TabulationHighlighter : Highlighter
     void do_highlight(HighlightContext context, DisplayBuffer& display_buffer, BufferRange) override
     {
         const ColumnCount tabstop = context.context.options()["tabstop"].get<int>();
-        auto& buffer = context.context.buffer();
+        const auto& buffer = context.context.buffer();
         auto win_column = context.setup.window_pos.column;
         for (auto& line : display_buffer.lines())
         {
@@ -1033,7 +1033,7 @@ private:
     {
         const int tabstop = context.context.options()["tabstop"].get<int>();
         auto whitespaceface = context.context.faces()["Whitespace"];
-        auto& buffer = context.context.buffer();
+        const auto& buffer = context.context.buffer();
         auto win_column = context.setup.window_pos.column;
         for (auto& line : display_buffer.lines())
         {
@@ -1112,7 +1112,7 @@ private:
         if (contains(context.disabled_ids, ms_id))
             return;
 
-        auto& faces = context.context.faces();
+        const auto& faces = context.context.faces();
         const Face face = faces["LineNumbers"];
         const Face face_wrapped = faces["LineNumbersWrapped"];
         const Face face_absolute = faces["LineNumberCursor"];
@@ -1153,7 +1153,7 @@ private:
         unique_ids.push_back(ms_id);
     }
 
-    int compute_digit_count(const Context& context) const
+    static int compute_digit_count(const Context& context)
     {
         int digit_count = 0;
         LineCount last_line = context.buffer().line_count();
@@ -1268,7 +1268,7 @@ void highlight_selections(HighlightContext context, DisplayBuffer& display_buffe
 
 void expand_unprintable(HighlightContext context, DisplayBuffer& display_buffer, BufferRange)
 {
-    auto& buffer = context.context.buffer();
+    const auto& buffer = context.context.buffer();
     auto error = context.context.faces()["Error"];
     for (auto& line : display_buffer.lines())
     {
@@ -1367,7 +1367,7 @@ private:
     void do_highlight(HighlightContext context, DisplayBuffer& display_buffer, BufferRange) override
     {
         auto& line_flags = context.context.options()[m_option_name].get_mutable<LineAndSpecList>();
-        auto& buffer = context.context.buffer();
+        const auto& buffer = context.context.buffer();
         update_line_specs_ifn(buffer, line_flags);
 
         auto def_face = context.context.faces()[m_default_face];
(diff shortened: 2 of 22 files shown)


src/client.cc
```diff
@@ -212,7 +212,7 @@ void Client::redraw_ifn()
     if (m_ui_pending == 0)
         return;
 
-    auto& faces = context().faces();
+    auto const& faces = context().faces();
```
lenormf:
```suggestion
    const auto & faces = context().faces();
```

src/constexpr_utils.hh
```diff
@@ -13,7 +13,7 @@ namespace Kakoune
 template<typename T, size_t N>
 struct Array
 {
-    constexpr size_t size() const { return N; }
+    static constexpr size_t size() { return N; }
```
lenormf: Not sure this is very elegant, even if it makes sense in theory.

src/highlighters.cc
```diff
@@ -901,7 +901,7 @@ struct WrapHighlighter : Highlighter
         return pos;
     };
 
-    ColumnCount line_indent(const Buffer& buffer, int tabstop, LineCount line) const
+    static ColumnCount line_indent(const Buffer& buffer, int tabstop, LineCount line)
```
lenormf: Same remark, random static member function.

src/highlighters.cc
```diff
@@ -1153,7 +1153,7 @@ struct LineNumbersHighlighter : Highlighter
         unique_ids.push_back(ms_id);
     }
 
-    int compute_digit_count(const Context& context) const
+    static int compute_digit_count(const Context& context)
```
lenormf: Same remark, random static member function.

src/json_ui.cc
```diff
@@ -19,7 +19,7 @@ namespace Kakoune
 {
 
 struct invalid_rpc_request : runtime_error {
-    invalid_rpc_request(String message)
+    invalid_rpc_request(const String& message)
```
lenormf: I think the idiom that is used in the codebase is to pass a `StringView` rather than a constant reference?

src/regex.hh
```diff
@@ -199,7 +199,7 @@ struct RegexIterator
         : RegexIterator{begin, end, begin, end, vm_arg, flags} {}
 
     It begin() { return {*this}; }
-    Sentinel end() const { return {}; }
+    static Sentinel end() { return {}; }
```
lenormf: Same remark, random static member function.

src/regex_impl.cc
```diff
@@ -425,14 +425,14 @@ struct RegexParser
         parse_error(format("unknown atom escape '{}'", cp));
     }
 
-    void normalize_ranges(Vector<CharacterClass::Range, MemoryDomain::Regex>& ranges)
+    static void normalize_ranges(Vector<CharacterClass::Range, MemoryDomain::Regex>& ranges)
```
lenormf: Same remark, random static member function.

src/regex_impl.hh
```diff
@@ -521,7 +521,7 @@ private:
         }
     }
 
-    void to_next_start(Iterator& start, const ExecConfig& config, const StartDesc& start_desc)
+    static void to_next_start(Iterator& start, const ExecConfig& config, const StartDesc& start_desc)
```
lenormf: Same remark, random static member function.

src/shell_manager.cc
```diff
@@ -97,7 +97,7 @@ struct Pipe
     void close_write_fd() { close_fd(m_fd[1]); }
 
 private:
-    void close_fd(int& fd) { if (fd != -1) { close(fd); fd = -1; } }
+    static void close_fd(int& fd) { if (fd != -1) { close(fd); fd = -1; } }
```
lenormf: Same remark, random static member function.

src/alias_registry.hh
```diff
@@ -22,7 +22,7 @@ public:
     {
         auto merge = [](auto&& first, const AliasMap& second) {
             return concatenated(std::forward<decltype(first)>(first)
-                                | filter([&second](auto& i) { return not second.contains(i.key); }),
+                                | filter([&second](const auto& i) { return not second.contains(i.key); }),
```
lenormf: Making iterators constant in utility functions is nice but it's a lot of work, it clutters the code slightly, and also not really a big deal. Can't really undo it now that you've fixed them all, but it's a practice that will probably not be observed in future code :p

eraserhd: I'm ok with that. :)

src/constexpr_utils.hh
eraserhd: Alright, I buy that.

eraserhd: Well, I guess it doesn't really make sense in theory, either.  Removed.

src/json_ui.cc
eraserhd: I think StringView is about generalizing interfaces to work for `const char*`, `String` and various proxy types.  This only has one caller that passes a String, so using a const reference avoids a copy.  Granted, this copy is super unimportant, since it's for an exception path, and I'm not attached to it.  StringView would construct a temporary object here.

eraserhd:
Regarding making member functions like `normalize_ranges` and `to_next_char` static:  I can't figure out how this is "random", they are not part of a public contract, but implementation details that don't need a `this` pointer.  They could be moved outside the class, work, and the names would even be correct without their scope.  Perhaps in this codebase, it is more normal to move them outside of the class - I guess I didn't consider that - is it?   (Also, the highlighters have been performance critical.)

Regarding `RegExIterator.end()`, this only narrows the interface slightly, in that it isn't possible to take a member pointer, but would allow better optimizations in a performance critical bit of code.  I swear I've seen this in libstdc++ as a matter of course, and as a point of accepted C++ practice, but I can't find a reference to it now.  (Actually, isn't this the whole reason the compiler will allow calling a static method with a `.` or `->` on an instance?)

mawww:
Adding `static` is a contract, we guarantee calling code that it can just call it as `ClassName::function` so I am not in favor of adding just just because we can. It is the same for `const` however in most examples I've seen the `const` is actually mandated.

I am not really keen on adding `const auto&` in place of `auto&` for lambda parameters, as I dont think it adds much value when the lambda body is small (its quite obvious we dont mutate the value), and it makes those lambda even more verbose.

src/context.hh
```diff
@@ -22,7 +22,7 @@ class AliasRegistry;
 struct JumpList
 {
     void push(SelectionList jump, Optional<size_t> index = {});
-    const SelectionList& forward(Context& context, int count);
+    const SelectionList& forward(const Context& context, int count);
```
mawww: Here, its surprising that forward takes a const Context but backward not. I think thats due to backward potentially mutating the jump list, which forward never does, but it still feels curious to put that in the function contract (as a caller, its strange that if you have a const Context, you can jump forward but not backwards)

src/regex.hh
eraserhd: Reverted

PR #3181, opened 2019-11-05
src: Make sure clients connect from a TTY

### LINKED ISSUE #3159: kak -c does not check that stdout is a terminal.
```
$ pacman -Qi kakoune | egrep 'Name|Version'
Name            : kakoune
Version         : 2019.07.01-1
$ uname -s -r -v -m -o
Linux 5.3.6-arch1-1-ARCH #1 SMP PREEMPT Fri Oct 11 18:28:05 UTC 2019 x86_64 GNU/Linux
```

### Steps 
Start a session.
```
$ kak -s foo
...
```
From a different terminal, join the session, but forward stdout to a non-terminal.
```
$ kak -c foo >/dev/null
```
### Outcome 
kak starts up as if nothing is wrong, but of course you can't see anything. You can still type commands and they will still work. For instance, to exit that client, you can still type :wq<ret>, and if you do insertion commands, you can see the result in the first terminal.

### Expected 
kak should fail in the same way it does without the -c:
```
$ kak >/dev/null
Could not start kakoune: stdout is not a tty
```

Though this behaviour could actually be useful under some circumstances. Like if a kak session is running on a networked computer hooked up to a projector and multiple people are sshing into that networked computer and editing the same file, looking at the projector screen to guide their actions. In which case, it would only be useful if the cursors of all clients were rendered on the projector, there might be some scrolling logistics you'd have to work out, and the same behaviour would be possible without having to forward stdout to /dev/null.

lenormf:
Fixes #3159

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -1104,6 +1104,9 @@ int main(int argc, char* argv[])
             for (auto name : files)
                 new_files += format("edit '{}';", escape(real_path(name), "'", '\\'));
 
+            if (not isatty(1))
+                throw startup_error("stdout is not a tty");
+
             return run_client(*server_session, {}, new_files + client_init, init_coord, ui_type, false);
         }
         else


eraserhd:
I think this will fail on JSON UI clients (which don't necessarily have a tty)?

I suspect this should somehow be in `ncurses_ui.cc`.

PR #3182, opened 2019-11-05
src: Allow `:on-key` to display an information box

lenormf:
Key mappings that use `on-key` to emulate single hit prompts (like
the `t` primitive) currently have no way of notifying the user that
an input is pending.

This commit allows passing an -info flag to `on-key`, in order to
display the given string in an information box until a key is hit, e.g.

```
on-key -info "hit a key" 'echo %val{key}'
```

### DIFF
--- a/doc/pages/commands.asciidoc
+++ b/doc/pages/commands.asciidoc
@@ -274,10 +274,11 @@ but not really useful in that context.
     prompt content will be passed as a single token. In other words,
     word splitting does not take place.
 
-*on-key* <command>::
+*on-key* [-info <info>] <command>::
     wait for next key from user, then execute <command>, the key is
     available through the `key` value, accessible through `$kak_key`
-    in shells, or `%val{key}` in commands.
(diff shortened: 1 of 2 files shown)


lenormf:
My usecase is for the following user-mapping, which I would like not to be running without any kind of notification that input is pending:

```
map -docstring "select after the given character" global user 't' %{: on-key 'exec / %val{key} \K.<lt>ret>'<ret>}
```

occivink:
aren't you better served by doing `info 'hit a key'; on-key 'echo %val{key}'`? that way you can use the full options of `info`, title and all.

lenormf:
I just wanted a simple way to show a message on screen, adding a flag to `:on-key` as a shortcut is ideal for mappings.

Also the following will show the information box until the screen needs to be refreshed (as opposed to being hidden as soon as any key is hit):

```
:info test; on-key 'echo -debug %val{key}'
```

lenormf:
Rebased on master.

PR #3184, opened 2019-11-08
rc spell: Allow storing a language in an option

lenormf:
This commit adds a `spell_lang` option that will be used by the
`spell` command when no language is passed to it directly.

Setting a buffer/window local option allows using `:spell` seamlessly
across several buffers, written in different languages.

### DIFF
--- a/rc/detection/modeline.kak
+++ b/rc/detection/modeline.kak
@@ -34,6 +34,7 @@ define-command -hidden modeline-parse-impl %{
                 ft|filetype) tr="filetype ${value}";;
                 bomb) tr="BOM utf8";;
                 nobomb) tr="BOM none";;
+                spelllang|spl) tr="spell_lang ${value%%,*}";;
                 *) printf %s\\n "echo -debug 'Unsupported vim variable: ${key}'";;
             esac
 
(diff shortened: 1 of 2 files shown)


lenormf:
Added modeline support for the option.

mawww:
Why cant we use the existing `spell_lang` for that ? why do we need to split `spell_last_lang` and `spell_lang` ?

lenormf:
`spell_lang` in the current implementation stores the last language that was used by `aspell` (it's a hidden option). It allows not having to specify a language to `:spell-replace` on every cal.

We want a buffer option, but we also want to be able to override it with a parameter to `:spell`, so we need another - separate, visible - option.

mawww:
so, if I understand correctly, the goal is that `spell` with no parameters re-uses the configured language instead of the last language passed to `spell` ? so that `set buffer spell_lang en; spell fr; spell` the last spell command uses english instead of french ?

lenormf:
The command uses the option's value, unless a parameter is given explicitly to the command which takes precedence. In your example `:spell fr` uses french, `:spell` would use english (the value stored in the option.

mawww:
Ok, it seems to me using a single option and having `:spell <language>` update it as a side effect has a simpler implementation and is also a reasonable behaviour, but I dont care very strongly about that. 

Do you think it is better to not update the language when we explicitely pass it to spell (and hence require a second option to store that additional state) ?

lenormf:
I thought of reusing that option, it even had the right name, but there are cases whereby users could want to spell-check a buffer in a language that's not necessarily the "official one" set in the buffer option.

Example: spell-checking all the french inline translations in the Kakoune website source code, when the text is in english by default - overwriting the buffer option every time in those cases would be inconvenient.

The other advantage is the predictability of calling `:spell` when there's a modeline set: if the buffer states that the spelling language is e.g. english, calling `:spell` will use the english dictionary regardless of the commands the user has previously used. The optional parameter to `:spell` becomes a shortcut to override the buffer option, not a way to set it.

PR #3188, opened 2019-11-10
A 'c' column suffix selects by codepoints

### LINKED ISSUE #2724: Proposal: Select by character or display indices

eraserhd:
Closes #2724

This only selects by codepoints right now.  If this approach is acceptable, I'll add support for a `d` suffix for display columns.

This only allows codepoint-based selection for the most recent timestamp on a buffer.  Adding support for historic timestamps will be quite tricky, but I'm willing to do it if we have a use case (or a compelling reason).

The test should be comprehensive, starting and ending on a single-byte char, a multi-byte char, selecting only a combining character, the main character plus a subset of combining characters, having a reverse selection. I was worried about failure on interesting cases - but none of them turned out to need more work, surprisingly.  I'm happy to add test cases if I missed anything.


Screwtapello:
Please also document this feature in `doc/pages/commands.asciidoc` in the section for the `:select` command!

lenormf:
Is the granularity on the feature necessary? Couldn't we pick one best-case-suits-most (codepoints for instance)?

occivink:
@lenormf I don't know about the codepoint usecase, but it's important that `select` can be used with the format used in `$kak_selection` and similar values

mawww:
I dont expect cases where a single `select` gets a mixed codepoint/bytes/columns coordinates, so I would encourage the use of a switch instead of suffixes for that. For the register case, we have a first elements containing some metadata (buffer, timestamp...) so we could also put something there if we wanted to support putting codepoint coordinates in a register (not convinced we have a use case for that).

I am still not really excited about such a change, as making it work with timestamps is going to be quite complex, and I expect most use cases for this comes from tools like `kak-lsp` that are running asynchronously from Kakoune (so they have not guarantee that the timestamp is still the same).

Could you elaborate on why byte offsets do not work in your case ?

eraserhd:
I accidentally [a whole document](https://gist.github.com/eraserhd/f14469a12f379074b8ff5337cb448d34) on the two cases I have and potential solutions.

My partner suggested the possibility of another program, which I hadn't considered before and actually doesn't seem too bad, if done right.

The big reason for a change to Kakoune was about leverage: doing it in the plugins is a lot of work, and these tools are already trying to be editor-agnostic.  While I'd already started to do this for parinfer-rust, I came to the conclusion there should be an easy path to support tools which think in codepoints or display columns, rather than making column translation a concern that many plugin authors would have to deal with.

mawww:
Thanks for that detailed document. I think making Kakoune able to communicate in terms of codepoints and columns makes sense. I still would prefer a switch provided to `select`, along with a way to get codepoints/columns based selection descs.

Regarding timestamps, it seems this would be fine in the parinfer and select-nrepl cases, as they runs synchronously. It would be an acceptable limitation for now to require it to match the current buffer state.

eraserhd:
Cool!  This sounds like a plan.  I'll do the expansion first, since I can use it by itself.

PR #3190, opened 2019-11-10
Miscellaneous highlighting fixes for Elm

lePerdu:


rc/filetype/elm.kak
```diff
@@ -25,24 +25,31 @@ hook -group elm-highlight global WinSetOption filetype=elm %{
     hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/elm }
 }
 
+hook global WinSetOption filetype=elm %{
+    set-option window formatcmd 'elm-format --stdin'
```
lenormf: Since this tool seems to be a third-party utility, it's better to let users set that option in their local configuration rather than arbitrarily set one by default.

mawww: Also, this introduces a dependency on the format.kak module, which needs to be loaded first, we still did not decide on a guideline on how we want those dependencies to be handled. best avoid this hook for now.

lePerdu: That makes sense. I saw the Rust configuration set `formatcmd` to `rustfmt`, but it looks like it's the only one that does so.

lenormf: Thanks for reporting that bug, it's been addressed in #3193.

PR #3193, opened 2019-11-11
rc rust: Don't set an arbitrary formatter

lenormf:
Users who want to use a formatter are free to pick a tool of their
choosing and set `formatcmd` in their user configuration.

### DIFF
--- a/rc/filetype/rust.kak
+++ b/rc/filetype/rust.kak
@@ -25,14 +25,6 @@ hook -group rust-highlight global WinSetOption filetype=rust %{
     hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/rust }
 }
 
-# Configuration
-# ‾‾‾‾‾‾‾‾‾‾‾‾‾
-
-hook global WinSetOption filetype=rust %{
-    set window formatcmd 'rustfmt'
-}
-
-
 provide-module rust %§
 
 # Highlighters


Delapouite:
Why would a user want to use a different tool than `rustfmt` to write `rust`? This would lead to an unnecessary fragmentation of the rust ecosystem.

I'm all for user freedom, but in this specific case, *choice* only results on countless bikesheddings that is harmful for the rust community. Same for `elm-format` that is the blessed way to write `elm`.

lenormf:
It's only a matter of time before somebody forks the formatting tool, and the community ends up fighting over which one should be the default in the editor.

Or the maintainer stops working on the tool, and we end up with bad defaults.

Or some other new tool comes up and becomes the "standard" tool.

In any case, the file doesn't load the `formatter` module dependency, and [it's still not clear whether that's the behaviour that is wanted anyway](https://github.com/mawww/kakoune/pull/3190#discussion_r344657724).

eraserhd:
I think having a default when there is a reasonable one is nice, but it should definitely be overridable.  The only way to override this is with another hook that changes it back.

And also the dependency issue.

This sounds like a good place for the "included option scopes" thing @mawww mentioned a while back.

mawww:
I think `rustfmt` is a sensible default, probably the same for `elm-format`, its the hidden dependency I object to, rust is the only script that sets formatcmd, so for now I'd rather remove it, then once we have solved the dependency issue we can go through language support scripts and re-add formatcmd support.

PR #3194, opened 2019-11-12
Implement %val{selections_char_desc}

eraserhd:
I did this by extracting a `SelectionImpl<>` template that is instantiated for byte and codepoint selection. This part seems right, but I don't understand the "target" part of BufferCoordAndTarget and now CharCoordAndTarget.  Is this just a cache to avoid recomputing the cursor column?


lenormf:
Maybe `SelectionImp` could be named `SelectionOps`? It's the idiom used in `string.hh` to do code duplication through inheritance.

mawww:
I tried to explain how I wanted the implementation to look like, but it ended up easier to just write it myself, pulled your tests and documentation in.

mawww:
And target is the target column the user wants the cursor at, this is not a cache as it cannot be derived from the other data, this target column only changes on explicit horizontal moves, not on vertical moves, and is the reason the cursor gets back to the previous column after being forced on column zero if you go down through and empty line.

PR #3198, opened 2019-11-12
Change markdown comment block

JJK96:


mawww:
Not really familiar with either syntaxes, and this does not seem to be documented on commonmark spec, can you elaborate ?

JJK96:
I found that the old comment syntax didn't work for me with pandoc
(i.e. that text still appeared in the output). So I did some research and found several guides explaining this new syntax. So I tried it and this worked.

https://www.jamestharpe.com/code-comments-markdown/
https://stackoverflow.com/questions/4823468/comments-in-markdown

alexherbo2:
There is no syntax for comments in Markdown.  There is HTML comments (`<!-- … -->`), but they will appear in the HTML document.  The syntax used is a link reference with no reference in the document.

JJK96:
Correct, they are not actually comments, but neither is the old syntax that is used. But in practice this works like a comment.

PR #3199, opened 2019-11-12
Fix test suite freeze

### LINKED ISSUE #3196: The test suite randomly freezes
Hi,

Ever since `e888f434` was merged, the test suite has been randomly freezing, both on Travis (our C.I.) and the [Fedora COPRS](https://copr.fedorainfracloud.org/coprs/jkonecny/kakoune/builds/).

Now and then, a machine on Travis would fail, I've always dismissed those errors as being Travis-related (we're used to that by now), but there might be more to it.

HTH.

eraserhd:
The issue is that the `test/regression/2056-assert-on-small-window`
sent JSON UI input without waiting for for the JSON UI to be
initialized.

Closes #3196

It could be nice to fix the deadlock at a more fundamental level,
but it's probably not a normal use case to send JSON input messages
before reading the first JSON UI output message.


jkonecny12:
Great thanks for fixing this!

PR #3203, opened 2019-11-13
src: Isolate the JSON implementation from the UI code / limit recursion depth

lenormf:
C.f. commit messages for the details.

### DIFF
--- a/src/json.cc
+++ b/src/json.cc
@@ -1,0 +1,175 @@
+#include "json.hh"
+
+#include "exception.hh"
+#include "string_utils.hh"
+#include "unit_tests.hh"
+#include "utils.hh"
+
+#include <cstdio>
+
+namespace Kakoune
+{
+
+String to_json(int i) { return to_string(i); }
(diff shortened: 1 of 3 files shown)


mawww:
I am not sure about that change, as json is currently ony use for the json ui, and the to_json functions only exist for UI related types.

If we were to move json support to its own file, then it should only provide the to_json functions for basic types, it is quite surprising that json.cpp provides a to_json function for DisplayAtom, DisplayLine, Color.... That means that we would have to add those to_json functions next to the type, which would make json support something quite central in Kakoune.

For this reason, I think until we do have additional use cases for json support, I'd rather keep it isolated as part of the json_ui, where it makes perfect sense to only have support for types needed by the UI.

lenormf:
I've moved the functions with UI-types-dependent signatures back into `json_ui.cc`, even though it's arguable whether it's a good idea to have function that generate/parse JSON strings in files other than `json.cc`.

Don't forget about the other commit which adds a maximum recursion depth limit to parsing, you can cherry-pick that if anything.

PR #3204, opened 2019-11-13
Fix for select-display-column test

eraserhd:
This should disable test select-display-column test when UTF8
is not present, and hopefully it will enable UTF8 in Travis CI.

According to the docs, it is already available in Travis CI, but
the man page for wcwidth() only says it depends on LC_CTYPE.  So
if this PR fails to build, it is likely that ncurses is too old to
recognize the emojis in the select-display-column case.

### DIFF
--- a/test/compose/select-display-columns/enabled
+++ b/test/compose/select-display-columns/enabled
@@ -1,0 +1,15 @@
+#!/bin/sh
+if ! locale | grep LC_CTYPE | grep -qi 'utf-*8'; then
+    exit 1
+fi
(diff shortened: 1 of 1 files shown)


eraserhd:
UPDATE: The issue seems to be the glibc version, and I've updated the `enabled` script to detect it.  Hopefully this will pass.

eraserhd:
*sigh* I give up for now.  I think it's a glibc bug, but perhaps it needs a C program to detect the problem?

mawww:
This got fixed by using different codepoints.

PR #3205, opened 2019-11-13
Add git next-hunk and prev-hunk commands

### LINKED ISSUE #2423: Implement `git` commands to jump to the previous/next modification in the buffer

Anfid:
Fixes #2423 
Alternative implementation #2458

### DIFF
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -19,14 +19,15 @@ hook -group git-status-highlight global WinSetOption filetype=git-status %{
 
 declare-option -hidden line-specs git_blame_flags
 declare-option -hidden line-specs git_diff_flags
+declare-option -hidden int-list git_hunk_list
 
 define-command -params 1.. \
   -docstring %sh{printf 'git [<arguments>]: git wrapping helper
 All the optional arguments are forwarded to the git utility
-Available commands:\n  add\n  rm\n  blame\n  commit\n  checkout\n  diff\n  hide-blame\n  hide-diff\n  init\n  log\n  show\n  show-diff\n  status\n  update-diff'} \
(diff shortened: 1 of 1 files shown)


lenormf:
Why reimplement the feature? The code is essentially the same, except you have one function that does two loops and stores data in an option?

Anfid:
That MR was not updated for a long time and my solution provides a more efficient implementation, that also addresses mawww's notes

lenormf:
It's not clear to me how it's more efficient, could you explain?

Anfid:
All hunks are cached into separate list, that contains only single line number for each hunk. This prevents unnecessary iterations on larger hunks. After that, if no changes to the buffer were made, this function only operates on smaller list.

This also corresponds to regular workflow, alternating between edit and navigation.

lenormf:
I see. It's specifically the caching part that I'm sceptical about. It's not uncommon for me to run linting on several long files at the same time, having persistent storage whose size is proportional to the amount of diagnostics just for jumping across them is not appealing.

Anfid:
I'm not sure, what linting on large files has to do with this storage. It only stores one number for each hunk, even if it consists of multiple lines. If you don't use any of these commands, then it won't be generated for this buffer and no extra memory will be used. And considering some changes to file, that is somewhat like 4 bytes per hunk. I feel like ~40 bytes per buffer with medium size diff is worth it to save some cpu cycles. I don't see a problem here. But if you do have a more concrete problem in mind, we could discuss alternative solution

lenormf:
#3321

PR #3206, opened 2019-11-14
rc: Use the standard `fail` command to report errors

lenormf:
Merely using `echo` with markup doesn't log errors in the debug
buffers, and is also less readable.

### DIFF
--- a/rc/tools/man.kak
+++ b/rc/tools/man.kak
@@ -45,16 +45,11 @@ define-command -hidden -params 2..3 man-impl %{ evaluate-commands %sh{
                 set-option buffer filetype man
                 set-option window manpage $buffer_name $*
         "
-    elif [ "${retval}" -eq 16 ]; then
-        printf %s\\n "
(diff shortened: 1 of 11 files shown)


rc/tools/man.kak
```diff
@@ -45,16 +45,11 @@ define-command -hidden -params 2..3 man-impl %{ evaluate-commands %sh{
...
-        "
     else
-        printf %s\\n "
-        fail $(cat $manerr)
-        nop %sh{ rm ${colout}; rm ${manerr} }
-        "
+        printf '
+            fail %%{%s}
+            nop %%sh{ rm "%s"; rm "%s" }
+        ' "$(cat "$manerr")" "${colout}" "${manerr}"
```
lenormf: Weird redundant code here, let me know if it should stay.

mawww: Yeah, not sure why we delay those rm, but could be fixed in another PR.

PR #3221, opened 2019-11-22
src: Allow user modes to display a docstring

### LINKED ISSUE #3099: Add -docstring switch to declare-user-mode
Currently, although the individual commands can have documentation, the only information about the mode itself is relegated its name. Something really cool would be if instead of a static string, kakoune could allow the mode to specify a command to be called for its docstring. This would allow `kak-lsp` to print the results of `lsp-code-actions` or `lsp-hover` above the list of subcommands.

lenormf:
Fixes #3099


Delapouite:
Here's a screenshot of the ending result:

`declare-user-mode -docstring "let's handle those nice buffers!" buffers`

image

src/hash_map.hh
```diff
@@ -274,6 +275,9 @@ struct HashMap
 
     const Item& item(size_t index) const { return m_items[index]; }
 
+    template<typename KeyType, typename = EnableIfHashCompatible<KeyType>>
+    auto keys() const { return m_items | transform([](auto& i) { return i.key; }) | gather<Container<KeyType, domain>>(); }
+
```
mawww: I'd rather avoid adding this to hashmap, also the use of Container there seems wrong, we might not want the same container type for storing the hashmap and gathering the keys.

src/keymap_manager.hh
```diff
@@ -45,13 +45,13 @@ public:
     };
     const KeymapInfo& get_mapping(Key key, KeymapMode mode) const;
 
-    using UserModeList = Vector<String>;
+    using UserModeList = HashMap<String, String, MemoryDomain::UserModes>;
```
mawww: I would prefer to have a Vector<UserMode> with `struct UserMode { String name; String docstring; }`.

src/commands.cc
```diff
@@ -1381,7 +1382,7 @@ const CommandDesc debug_cmd = {
             auto& keymaps = context.keymaps();
             auto user_modes = keymaps.user_modes();
             write_to_debug_buffer("Mappings:");
-            for (auto& mode : concatenated(modes, user_modes) | gather<Vector<String>>())
+            for (auto& mode : concatenated(modes, user_modes | transform([](auto& i) { return i.key; })) | gather<Vector<String>>())
```
mawww: I believe you can use `transform(&UserModeList::Item::key)` or just `transform(&UserMode::name)` if you use the struct suggested later.

src/keymap_manager.hh
lenormf: That was what my original was using, but using a hashmap was so much easier… I'll change it back to a vector and rebase the PR.

lenormf:
Reverted the container for user modes back to a vector, rebased on `master`.

lenormf:
Rebased on master.

src/commands.cc
```diff
@@ -2556,19 +2557,34 @@ const CommandDesc declare_user_mode_cmd = {
     "declare-user-mode",
     nullptr,
     "declare-user-mode <name>: add a new user keymap mode",
-    single_param,
+    ParameterDesc{
+        { { "docstring", { true, "user mode description" } } },
+        ParameterDesc::Flags::SwitchesOnlyAtStart, 1, 2
```
mawww, line 2562: Any reason for forcing switches at the start ? We usually do that only for commands that might get other commands/things that look like switch later on, here it seems like an unnecessary constraint.

src/commands.cc
```diff
@@ -2556,19 +2557,34 @@ const CommandDesc declare_user_mode_cmd = {
...
+    },
     CommandFlags::None,
     CommandHelper{},
     CommandCompleter{},
     [](const ParametersParser& parser, Context& context, const ShellContext&)
     {
-        context.keymaps().add_user_mode(std::move(parser[0]));
+        auto docstring = parser.get_switch("docstring").value_or(StringView{});
+        context.keymaps().add_user_mode(std::move(parser[0]), std::move(docstring.str()));
```
mawww, line 2570: That second std::move is unnecessary, str() already returns a temporary so it will be moved by default.

src/commands.cc
lenormf: All commands show switches as being the first optional list of flags, so I guess I used this flag here.

I reverted back to `None`.

lenormf:
Addressed review comments.

Issue #1641, opened 2017-10-19
Notify the user when the file underlying an open buffer was removed


lenormf:
Hi,

Currently, there's no notification system that would warn a user that the file that they opened/created isn't present on the disk anymore.

HTH.

mawww:
Can you suggest the behaviour you'd like to see ? Detecting the file is gone seems easy enough, but what to do with that is less clear.

lenormf:
A simple popup message similar to the one that appears when a file was changed would be enough for me.

mawww:
What would that popup propose as options ? we cannot really reload the file, so it would just be a notification. In that case wouldn't it be better to use a non modal info box (and maybe to add back the 'new' flag on the file) ?

lenormf:
I just want a way to know I'm not editing a file that is not backed by the filesystem anymore, I thought about a modal.

```
$ rm main.cc
"The file associated with this buffer was removed, do you want to delete it? y/n"
```

If yes, the buffer is removed, otherwise the new flag is assigned back to it.

lenormf:
Bump.

PR #3225, opened 2019-11-23
src: Reload buffers when their contents' hash changes

lenormf:
Instead of triggering a reload event when the timestamp of a buffer's
underlying file changes, do so when its contents are actually modified.

Re-working of #1963.

Note that this doesn't address #1641 but it should be easy to handle.

### DIFF
--- a/src/buffer.hh
+++ b/src/buffer.hh
@@ -5,6 +5,7 @@
 #include "coord.hh"
 #include "constexpr_utils.hh"
 #include "enum.hh"
+#include "file.hh"
 #include "optional.hh"
 #include "safe_ptr.hh"
 #include "scope.hh"
@@ -141,9 +142,9 @@ public:
     BufferCoord erase(BufferCoord begin, BufferCoord end);
     BufferCoord replace(BufferCoord begin, BufferCoord end, StringView content);
 
-    size_t         timestamp() const;
-    timespec       fs_timestamp() const;
-    void           set_fs_timestamp(timespec ts);
+    size_t          timestamp() const;
+    void            set_fs_status(FsStatus);
+    const FsStatus& fs_status() const;
 
     void           commit_undo_group();
     bool           undo(size_t count = 1);
@@ -284,7 +285,7 @@ private:
 
     Vector<Change, MemoryDomain::BufferMeta> m_changes;
 
-    timespec m_fs_timestamp;
+    FsStatus m_fs_status;
 
     // Values are just data holding by the buffer, they are not part of its
     // observable state
--- a/src/client.cc
+++ b/src/client.cc
@@ -287,7 +287,7 @@ void Client::reload_buffer()
     {
         context().print_status({ format("error while reloading buffer: '{}'", error.what()),
                                  context().faces()["Error"] });
-        buffer.set_fs_timestamp(get_fs_timestamp(buffer.name()));
+        buffer.set_fs_status(get_fs_status(buffer.name()));
     }
 }
 
@@ -312,7 +312,7 @@ void Client::on_buffer_reload_key(Key key)
     else if (key == 'n' or key == 'N' or key == Key::Escape)
     {
         // reread timestamp in case the file was modified again
-        buffer.set_fs_timestamp(get_fs_timestamp(buffer.name()));
+        buffer.set_fs_status(get_fs_status(buffer.name()));
         print_status({ format("'{}' kept", buffer.display_name()),
                        context().faces()["Information"] });
         if (key == 'N')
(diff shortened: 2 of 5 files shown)


src/client.cc
```diff
@@ -354,9 +354,18 @@ void Client::check_if_buffer_needs_reloading()
         return;
 
     const String& filename = buffer.name();
-    timespec ts = get_fs_timestamp(filename);
-    if (ts == InvalidTime or ts == buffer.fs_timestamp())
+    const timespec ts = get_fs_timestamp(filename);
+
+    if (ts == InvalidTime)
         return;
+    else
```
mawww: No need for an else here, we already return in the if case.

src/buffer.hh
```diff
@@ -144,6 +144,7 @@ public:
     size_t         timestamp() const;
     timespec       fs_timestamp() const;
     void           set_fs_timestamp(timespec ts);
+    size_t         fs_hash() const;
```
mawww: Expose fs_status itself here.

src/client.cc
mawww: And check the filesize as well, so that we dont need to read the whole file if it obviously changed. (Checking fd.data.size() against the stored value should be good enough, mapping without actually reading should be fine).

lenormf: Good point.

I put `else` because it looked natural, since we're creating an anonymous scope regardless.

src/buffer.hh
lenormf: I implemented this function because the timestamp is also part of `fs_status`, but didn't want to replace `fs_timestamp()` with `fs_status().timestamp`. Should I do that? Or leave `fs_timestamp()` and add `fs_status()` (would look weird)?

mawww: Just expose `fs_status` and use `fs_status().timestamp`

src/client.cc
mawww: You can use a if statement with initializer for that:
```
if (MappedFile fd{filename}; fd.st.st_size == status.size and hash_data(fd.data, fd.st.st_size) == status.hash)
    return;
```

Note that the use of `StringView{fd.data}` is wrong here, as it will look for a terminating 0-byte, which is not guaranteed and might be wrong (the file could contain zero bytes).

src/buffer.hh
```diff
@@ -144,6 +144,7 @@ public:
     size_t         timestamp() const;
     timespec       fs_timestamp() const;
```
mawww: remove

src/buffer.hh
```diff
@@ -144,6 +144,7 @@ public:
...
     void           set_fs_timestamp(timespec ts);
+    FsStatus       fs_status() const;
```
mawww: return by const-ref.

src/client.cc
```diff
@@ -354,9 +354,15 @@ void Client::check_if_buffer_needs_reloading()
...
     const String& filename = buffer.name();
-    timespec ts = get_fs_timestamp(filename);
-    if (ts == InvalidTime or ts == buffer.fs_timestamp())
+    const timespec ts = get_fs_timestamp(filename);
+
+    if (ts == InvalidTime)
+        return;
+    else if (MappedFile fd{filename};
+            fd.st.st_size == status.size and
+            hash_data(fd.data, fd.st.st_size) == buffer.fs_status().hash)
```
mawww: This does not check timetamp anymore, if it did not change there is no need to map the file. And I still would prefer for `else` not to be there as we return in the previous check.

src/client.cc
```diff
@@ -23,6 +23,13 @@
 namespace Kakoune
 {
 
+static Buffer::FsStatus get_fs_status(StringView filename)
```
lenormf: I implemented this function here, instead of `file.cc` because of the `Buffer::FsStatus` type (probably don't want that visible in the generic file-related API).

mawww: Well, FsStatus does not really need to be buffer specific, it could make sense to define it in file.hh along with `get_fs_status` and just use it in the buffer.

krobelus:
I think this changed the behavior in an unintended way.

When a buffer is modified, and then the underlying file is modified
as well, kakoune will not (ask to) reload the file anymore.

Steps to reproduce:

```sh
echo > file.txt
kak file.txt -e 'execute-keys i 1 <esc>'
# Modify the file outside kakoune
echo 2 >> file.txt
```

Observe that the buffer will not be reloaded.

mawww:
@krobelus I cannot reproduce this behaviour here, Kakoune does offer to releod the file. Both with `kak` and `kak -n`.

Can you check with `kak -n` to ensure this does not rely on some of your configuration ?

krobelus:
@mawww Hmm that's weird I cannot reproduce the one I posted anymore.

However, here's another example that reproduces consistently for me. Does that work?
```sh
echo original > file.txt; \
sh -c 'sleep 1; rm file.txt; echo original > file.txt' & \
./kak file.txt -n -e 'execute-keys o edit <esc> : w <ret>'
```

I originally found this for v2019.12.10 because kak didn't reload after I did `git checkout -- .` (which internally deletes files and writes their old content, hence the `rm`).
It bisects to this PR (e1bd076f5e64010578b904ad4bbb25b462517ef3). It seems to only happen when the content of the file is restored to "original" in the second `echo`, so maybe kak is using a stale checksum.

eraserhd:
FYI, I've also been having issues recently (1 or 1.5 weeks?) where switching branches and commits in git does not reload the file.   I have `autoreload` set to `yes`.

mawww:
Should be fixed on latest master, at the cost of re-reading the whole file just after writing it,

PR #3227, opened 2019-11-25
src: Use `begin()` to get iterators

### LINKED ISSUE #3226: Tests not passed with default Fedora build flags
I am trying to push Kakoune in official Fedora repos but if build it with default build flags for Fedora some tests not passed. Here is a [full build log](https://copr-be.cloud.fedoraproject.org/results/atim/kakoune/fedora-31-x86_64/01116987-kakoune/builder-live.log.gz). Maybe we can fix this somehow? Thank you.

lenormf:
When compiling the code with `-Wp,-D_GLIBCXX_ASSERTIONS`, the process
gets aborted, likely because iterators to standard containers are
not obtained in a safe way.

Fixes #3226.

### DIFF
--- a/src/buffer.cc
+++ b/src/buffer.cc
@@ -282,7 +282,7 @@ void Buffer::reload(StringView data, timespec fs_timestamp)
                         parsed_lines.lines[(int)(d.posB + line)]});
 
                 m_changes.push_back({ Change::Insert, cur_line, cur_line + d.len });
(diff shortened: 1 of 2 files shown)


lenormf:
Note that the [documentation](https://en.cppreference.com/w/cpp/algorithm/accumulate) states that since C++20, `std::accumulate()` will call `std::move()` on the initial value - might want to keep that in mind if C++20 is to be used in the future.

PR #3231, opened 2019-11-28
Update scroll behavior

### LINKED ISSUE #3128: Allow scrolling past the main cursor in the view mode

Anfid:
Fixes #3128 

Changes in scroll behavior:
* Scroll only resets selection if main cursor position is updated after the scroll
* Selections except main are always preserved
* Mouse dragging scroll behavior improved

I also would like to hear extra comments on this.

1) Is this behavior acceptable by default?
2) Should I update current view keybindings with this new behavior or introduce new ones? (`e` and `y`, for example)
3) Should similar behavior be also implemented for horizontal scrolling?

### DIFF
--- a/src/input_handler.cc
+++ b/src/input_handler.cc
@@ -138,7 +138,7 @@ struct MouseHandler
             return true;
 
         case Key::Modifiers::Scroll:
-            scroll_window(context, static_cast<int32_t>(key.key), not m_dragging);
+            scroll_window(context, static_cast<int32_t>(key.key), m_dragging);
             return true;
 
         default: return false;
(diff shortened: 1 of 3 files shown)


mawww:
Hello, just to let you know I do plan to give this a review, just did not get to it yet.

Anfid:
For your information: I have been testing this change locally since the creation of this PR with no issues at all. However my workflow is probably not an exhaustive test of all possible cases.

mawww:
I was planning to rework the whole "main selection must be visible" logic, which would make this change irrelevant, but I figured that in the mean time it is still an improvement to merge this.

PR #3246, opened 2019-12-07
Add OpenBSD section to README

bonds:
Thanks to @andreyorst for figuring out the workarounds to get kak working with plugins on OpenBSD: https://gitlab.com/andreyorst/plug.kak/issues/75. I figure others might benefit from the help.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -207,6 +207,27 @@ pkg install kakoune
 ====
 
 [TIP]
+.OpenBSD
+====
+Building on OpenBSD 6.6 amd64.
+
+--------------------------------------------------
+# pkg_add git gmake
+# git clone https://github.com/mawww/kakoune
+# cd kakoune
+# export CXX=clang++
+# gmake install
+--------------------------------------------------
+
+Running with support for plugins.
+--------------------------------------------------
+# pkg_add dash
+$ export KAKOUNE_POSIX_SHELL=/usr/local/bin/dash 
+$ kak
+--------------------------------------------------
+====
+
+[TIP]
 .Solus
 ====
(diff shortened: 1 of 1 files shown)


lenormf:
It seems like the problem comes from `ksh`, do you know exactly why Kakoune shell scopes don't run with it? Have you reported any bug to the `ksh` maintainers?

The commit documents a workaround for OpenBSD in the official documentation of the editor, maybe putting this information in the wiki until the bug is fixed in `ksh` is a better way.

Screwtapello:
Chasing links, it looks like we dealt with `ksh` compatibility in #2955, which applied a workaround to Kakoune's built-in scripts. That same workaround needs to be applied to each affected Kakoune plugin, though, until OpenBSD fixes its shell ([doesn't sound like a high priority](https://marc.info/?l=openbsd-misc&m=156039950105624&w=2)) and enough time has gone by that nobody is using the broken version.

Or we can just tell OpenBSD users to install a POSIX-compatible shell if they want to run Kakoune.

andreyorst:
> until OpenBSD fixes its shell

I don't think it will happen in near future. It seems that shell devs are really conservative for some reason. For example, in `dash` (Debian Alqumirist Shell) there's a [bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=482999) that I had to tackle in order for my custom job control in plug.kak to work. And this bug is 11 years old already and doesn't seem to be fixed any time soon unfortunately.

I think we should both update the scripts (but I don't think it is a high priority), and suggest OpenBSD users to install `dash` or `bash` for Kakoune to use. Who knows what bugs might be in `ksh` if such thing as putting processes in background works differently?

PR #3248, opened 2019-12-10
rc repl tmux: Do not check for the Tmux version

### LINKED ISSUE #3237: tmux error after startup
### Steps 

1. $ git clone https://github.com/mawww/kakoune
1. \# cd kakoune; gmake install
1. $ kak
1. :buffer \*debug\*

### Outcome 

````
shell stderr: <<<
tmux: unknown option -- V
usage: tmux [-2Cluv] [-c shell-command] [-f file] [-L socket-name]
            [-S socket-path] [command [flags]]
>>>
````

````
~ ❯❯❯ uname -a                                                                      Sun 2019-12-01 15:35:41 -0800
OpenBSD maybe.ggr.com 6.6 GENERIC.MP#3 amd64
~ ❯❯❯ kak -version                                                                  Sun 2019-12-01 15:35:43 -0800
Kakoune v2019.07.01-333-g0749ec9b
 
````

### Expected 

No errors at startup.

lenormf:
This commit removes attempts to auto-detect the version of the local
Tmux binary.

The Tmux package that ships with OpenBSD doesn't have a version
number[1], and therefore doesn't provide any way to check if it's
compatible with the REPL module, making the feature disabled by
default in that distribution.

Fixes #3237.

[1] https://github.com/tmux/tmux/issues/1141#issuecomment-340244302

### DIFF
--- a/rc/windowing/repl/tmux.kak
+++ b/rc/windowing/repl/tmux.kak
@@ -1,5 +1,6 @@
 # http://tmux.github.io/
 # ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+# Tmux version >= 2 is required to use this module
 
 hook global ModuleLoaded tmux %{
     require-module tmux-repl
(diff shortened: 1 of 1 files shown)


Screwtapello:
Since this no longer has additional requirements over the regular `tmux.kak` perhaps one day it could be folded in... but on the other hand, if we want to keep REPL stuff separate from windowing stuff, it should stay where it is.

That's a question for another PR, though — this seems strictly better than what's on `master` (because it works on OpenBSD), so I think it's worth merging as-is.

PR #3249, opened 2019-12-10
Fix build on FreeBSD again

t6:
```
json_ui.cc:29:9: error: use of undeclared identifier 'sprintf'
        sprintf(buffer, R"("#%02x%02x%02x")", color.r, color.g, color.b);
        ^
1 error generated.
```

Regressed by 7cdbe1d3d24c1cc13bd7cbc3fe252f1e88747ffb

### DIFF
--- a/src/json_ui.cc
+++ b/src/json_ui.cc
@@ -9,6 +9,7 @@
 #include "ranges.hh"
 #include "string_utils.hh"
 
+#include <cstdio>
 #include <utility>
 
 #include <unistd.h>


t6:
This kind of error mostly happens because there is no FreeBSD CI.  [Cirrus CI](https://cirrus-ci.org/features/) supports FreeBSD and is free for open source projects. Would you in principle accept a PR that adds a basic `.cirrus.yml`?

lenormf:
Something like the following seems to do the job:

```
test_task:
  matrix:
    - name: linux
      container:
        image: gcc:7
    - name: osx
      osx_instance:
        image: mojave-base
    - name: freebsd
      freebsd_instance:
        image_family: freebsd-12-1
  test_script: make test
```

Do you have any experience with Cirrus? It's much faster, and writing `.cirrus.yml` was very easy because the file is so short…

The locale seems to be wrong in the Linux containers, and I don't know how to handle multiple compilers, do you know how to do that?

t6:
> Something like the following seems to do the job:

I think it needs to be a little more complicated. For example, on FreeBSD `make` is not GNU make which needs to be installed first with `pkg install -y gmake` and needs to be invoked as `gmake`. The build at https://cirrus-ci.com/task/6142799119908864 did not even build anything. :wink: 

Something like (untested):
```
    - name: freebsd
      freebsd_instance:
        image_family: freebsd-12-1
        install_script: pkg install -y gmake
        test_script: make test
        env:
          LC_ALL: en_US.UTF-8
```

> Do you have any experience with Cirrus?

Very little. I'm using it for some of my projects with no problems so far. The FreeBSD Wiki (https://wiki.freebsd.org/HostedCI) lists a bunch of projects that use Cirrus too.

https://github.com/newsboat/newsboat/blob/master/.cirrus.yml#L74 handles multiple compilers by setting some environment variables via `env:`.

As for the locale would something like
```
env:
  LC_ALL: en_US.UTF-8
```
work?

Screwtapello:
I've written [a Cirrus CI build script for another project](https://github.com/byuu/bsnes/blob/4ec45a74533a232d29cd8daedb053ef044458dbd/.cirrus.yml) and it worked out pretty well.

lenormf:
Travis has been pretty slow and prone to error-ing out of test builds recently. If someone came up with a `.cirrus.yml` manifest in a fork and let Cirrus do its thing for a while (to verify whether it's more stable than Travis), I think switching over to it would be a good thing.

t6:
I have a working `.cirrus.yml` here: https://github.com/t6/kakoune/blob/bdee5f4ec9ab5090d7c407eee27d53f17d332f6d/.cirrus.yml
https://cirrus-ci.com/build/5012156017278976

lenormf:
Nice! Any idea how to handle compiling with Clang>=5 as well?

t6:
Sure. I've added a build with Clang on Linux and GCC on FreeBSD: https://github.com/t6/kakoune/blob/991583bdcbb98b009f7d3ef0f8ea7958f90eb99e/.cirrus.yml
https://cirrus-ci.com/build/5357709834059776

t6:
Note that there are lots of warnings when building with GCC9...

PR #3251, opened 2019-12-11
hooks.asciidoc: Document how *Key hooks interact with mappings and exec.

Screwtapello:


mawww:
I dont think this is fully correct, all hooks do not run in `execute-keys` by default, and `execute-keys -with-hooks` should trigger those hooks. There is nothing special about `execute-keys` and `NormalKey`/`InsertKey`. Regarding `RawKey` the keys are exclusively the ones coming from the user interface, so `execute-key` wont trigger it even with `-with-hooks`.

Screwtapello:
Aw man, I forgot about `-with-hooks`. Thanks for reminding me!

I've rebased my patch, and updated the text to not mention `execute-keys` for `NormalKey` and `InsertKey` (because they are perfectly normal in that regard) but to describe how `RawKey` is different.

PR #3253, opened 2019-12-11
src/ncurses_ui.cc: Add support for VT220-style Home and End keys.

### LINKED ISSUE #3252: Inside tmux <home> and <end> keys behave differently
### Steps
1. Try using \<home\> and \<end\> keys outside of tmux
2. Try using them inside tmux
### Outcome
Both \<home\> and \<end\> open the `select to inner surrounding object begin` menu
### Expected
The \<home\> key selects to beginning of line, and the \<end\> key to the end of it

Screwtapello:
Because we now support a bunch of different conventions for these keys, let's add some citations for the benefit of future maintainers.

Fixes #3252, tested with tmux 3.0a.

### DIFF
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -682,8 +682,8 @@ Optional<Key> NCursesUI::get_next_key()
         case 'B': return masked_key(Key::Down);
         case 'C': return masked_key(Key::Right);
         case 'D': return masked_key(Key::Left);
-        case 'F': return masked_key(Key::End);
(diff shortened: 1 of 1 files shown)


Screwtapello:
Also, I compared the `vte-256color` and `tmux-256color` terminfo database records, and the only difference in key-encodings was `khome` and `kend`. For whatever reason, tmux uses the same encoding as xterm for modified keys, it's just the unshifted `<home>` and `<end>` that differ.

PR #3258, opened 2019-12-14
cirrus: Implement a YAML control file for Cirrus

lenormf:


lenormf:
The report: https://cirrus-ci.com/build/5666064184115200

.cirrus.yml
```diff
@@ -0,0 +1,41 @@
...
+          gcc_script: pkg install -y gcc
+          env:
+              CXX: g++
+    test_script: gmake test
+    env:
+        LC_ALL: en_US.UTF-8
+
+linux_task:
+    container:
+        image: gcc:7
```
Screwtapello: If I read this correctly, on Linux we're building with GCC 7 specifically, and with whatever random version of Clang happened to come with whatever version of Debian or Ubuntu that container is based on.

Perhaps instead we should start from some particular OS version (as we do for FreeBSD and macOS) like `debian:stable` or `ubuntu:latest` (which despite the name is actually the latest long-term-support release, not the absolute latest version). That way, we can tell people "we support whatever compilers come with the stable release of Debian" and not have to worry about exactly which compiler version it is as long as our CI passes.

PR #3261, opened 2019-12-14
highlighters.cc: Add a '-min-width' flag to the number-lines highlighter

### LINKED ISSUE #3260: Implement option like Vim’s numberwidth
The option [numberwidth](https://vimhelp.org/options.txt.html#%27numberwidth%27) is used in Vim to specify the minimum width of `number-lines`. Its purpose is to prevent the body text from getting pushed over to the right when `number-lines` needs more digits to display the line numbers. Here is the relevant code:

https://github.com/mawww/kakoune/blob/b982371843168d104767434dfa1ead2befa7b6e3/src/highlighters.cc#L1156-L1163

The following pseudo code shows how a new option `number-lines-width` could work:

```cpp
return if digit_count > %opt{number-lines-width} {
    digit_count 
} else { 
    %opt{number-lines-width}
};
``` 
A good default value for `number-lines-width` would be `2` because it offers a nice trade-off between minimizing waste of screen real estate and preventing distraction caused by body text moving one column to the right.

Maybe you could also implement `flag-lines-width` to prevent linter errors from pushing the body text. I have `lsp-diagnostic-lines-disable window` in my config because kak-lsp’s adding and removing of `flag-lines` highlighters when editing Python files was driving me crazy.

*This issue originated from a discussion on https://github.com/mawww/kakoune/issues/3101#issuecomment-565699976*

Screwtapello:
Fixes #3260.

### DIFF
--- a/src/highlighters.cc
+++ b/src/highlighters.cc
@@ -1079,17 +1079,19 @@ private:
 
 struct LineNumbersHighlighter : Highlighter
 {
-    LineNumbersHighlighter(bool relative, bool hl_cursor_line, String separator)
+    LineNumbersHighlighter(bool relative, bool hl_cursor_line, String separator, int min_digits)
       : Highlighter{HighlightPass::Move},
         m_relative{relative},
         m_hl_cursor_line{hl_cursor_line},
-        m_separator{std::move(separator)} {}
+        m_separator{std::move(separator)},
+        m_min_digits{min_digits} {}
 
     static std::unique_ptr<Highlighter> create(HighlighterParameters params, Highlighter*)
     {
         static const ParameterDesc param_desc{
             { { "relative", { false, "" } },
               { "separator", { true, "" } },
+              { "min-digits", { true, "" } },
               { "hlcursor", { false, "" } } },
             ParameterDesc::Flags::None, 0, 0
         };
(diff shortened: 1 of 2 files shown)


src/highlighters.cc
```diff
@@ -1101,7 +1103,14 @@ struct LineNumbersHighlighter : Highlighter
         if (separator.length() > 10)
             throw runtime_error("separator length is limited to 10 bytes");
 
-        return std::make_unique<LineNumbersHighlighter>((bool)parser.get_switch("relative"), (bool)parser.get_switch("hlcursor"), separator.str());
+        StringView str_min_width = parser.get_switch("min-width").value_or("2");
+        int min_width = str_to_int(str_min_width);
```
mawww: `int min_width = parser.get_switch("min-width").map(str_to_int).value_or(2)`

src/highlighters.cc
```diff
@@ -1116,10 +1125,10 @@ struct LineNumbersHighlighter : Highlighter
         const Face face = faces["LineNumbers"];
         const Face face_wrapped = faces["LineNumbersWrapped"];
         const Face face_absolute = faces["LineNumberCursor"];
-        int digit_count = compute_digit_count(context.context);
+        int number_width = std::max(compute_digit_count(context.context), m_min_width);
```
mawww: its compute_digit_counts that should do the std::max operation. And rename min_width to min_digits.

Screwtapello: Editing 'compute_digit_counts` was my first guess, but it was `static` and therefore didn't have access to `m_min_digits`, and I didn't know how important static-ness was.

I've changed it to be not-static, and moved all the `std::max()` calls into it.

lenormf:
Looks like this broke the `column/multi-columns` test.

Screwtapello:
This is because the previous behaviour of `number-lines` was equivalent to `-min-width 0`, but the default is now `-min-width 2`. Basically, in `test/highlight/column/multi-columns/script`, everything like:

    "contents": "1│"

...should be changed to:

    "contents": " 1│"

(note the extra space before the digit).

PR #3262, opened 2019-12-14
Expose environment variable `$kak_binary_path`

fsub:
The usual pattern `kak -p $kak_session` potentially picks up a wrong
`kak` binary. The pattern `$kak_binary_path -p $kak_session` seems to be
superior.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -179,6 +179,10 @@ static const EnvVarDesc builtin_env_vars[] = { {
         [](StringView name, const Context& context, Quoting quoting)
         { return join(context.selections_content() | transform(quoter(quoting)), ' ', false); }
     }, {
+        "binary_path", false,
+        [](StringView name, const Context& context, Quoting quoting)
+        { return get_kak_binary_path(); }
+    }, {
         "runtime", false,
         [](StringView name, const Context& context, Quoting quoting)
         { return runtime_directory(); }


eraserhd:
Is this true?  I thought Kakoune actually prepends its binary directory to $PATH.

fsub:
See  https://github.com/occivink/kakoune-gdb/issues/15#issuecomment-565740243

eraserhd:
[Here](https://github.com/mawww/kakoune/blob/master/src/shell_manager.cc#L71-L77) is the code which ensures the correct `kak` is in the path.

I can't figure out how that issue could happen unless there's a script that is started from outside Kakoune (which there doesn't seem to be?) or a newer Kakoune is installed into the same path while an older Kakoune server is still running or there's a shell expansion in the kakoune-gdb plugin which has a bug messing up `$PATH`.  Can you clarify how this happened?

fsub:
In the current approach there are two obvious flaws:
- The directory that contains the `kak` executable is appended and not prepended to `PATH`. (Extending `PATH` is dangerous business. I'm advising against it.)
- The kakoune executable is not required to be called `kak`.

The funny part is that the bug was exposed by one of your changes: https://github.com/mawww/kakoune/commit/0e4a4acf61dbcd1e025810f8e67aca2eaeb44e19

eraserhd:
> The directory that contains the kak executable is appended and not prepended to PATH. 

Oh wow,  I failed to read closely enough notice that.  Now I see how the problem could have happened.

I think it should prepend.

> (Extending PATH is dangerous business. I'm advising against it.)

Can you think of a plausible breaking case for prepending to the PATH?

> The kakoune executable is not required to be called kak.

I think it is strongly required.  There are 25 instances of `kak -p` in the scripts that ship with Kakoune, and it is recommended in the documentation three times.  The build scripts symlink `kak.debug` or `kak.opt` to `kak` to fulfill this requirement when running from the source directory.

_If_ we conclude that finding `kak` in `$PATH` is not reliable, we should also change these references, the documentation, and remove the attempt to modify the path.

(But I still think this is what PATH is meant for, and that we should prepend.)

fsub:
Yes, I wouldn't have run into the issue if the directory containing the `kak` executable would be prepended to `PATH`. But extending `PATH` can have side effects. In this regard, prepending is even more dangerous than appending.
If the general approach of exposing an environment variable containing the path of the `kak` binary is accepted by @mawww we should replace all invocations of `kak`. The new variable also needs to be documented.
I can do it if desired but I'd better wait until @mawww says something. We also need a final name for the variable.

fsub:
...and the danger of extending `PATH` again is that wrong executables could be invoked.

fsub:
Example: Given
- `/usr/bin/kak`
- `/usr/bin/python`
- `/home/me/bin/python`
- `PATH=/home/me/bin:/usr/bin`

If I call `python` the system invokes `/home/me/bin/python`.
If kakoune prepended `/usr/bin` to `PATH` and itself or (more plausibly) one of its sub processes called `python` the system would invoke `/usr/bin/python`.
That's at least unexpected. I say it's dangerous.

eraserhd:
Oh, good point. :(

What about prepending `$prefix/libexec/kakoune` and symlink `kak` into that.  This has the advantage of allowing us to break out scripts and small, single-purpose utilities.

fsub:
Let's close it before we have to organize a birthday party... ;-)

mawww:
Realized I never commented on that, I agree with the general concerns, but having to write `$kak_binary_path` instead of `kak` is a deal breaker.

I think introducing a `libexec/kak` folder and having the binary there is the best proposed solution so far.

mawww:
Done.

PR #3263, opened 2019-12-14
Support rxvt style s-F3 to s-F12

fsub:
Rxvt emits `\E[23$` and `\E[24$` for `F21` and `F22` (alias `s-F11` and
`s-F12` ~~provided that `ncurses_shift_function_key` is set to `10`~~),
respectively.
_Addendum:_ Added support for `s-F3` to `s-F10` following @Screwtapello's suggestion.

### DIFF
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -712,6 +712,14 @@ Optional<Key> NCursesUI::get_next_key()
                 return masked_key(Key::F6 + params[0] - 17);
             case 23: case 24:
                 return masked_key(Key::F11 + params[0] - 23);
+            case 25: case 26:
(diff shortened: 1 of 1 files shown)


Screwtapello:
It seems that with the new keyboard parsing system, Kakoune doesn't support *any* shifted function keys in rxvt (except `<s-F1>` and `<s-F2>`, because they're identical to `<F11>` and `<F12>`). It also no longer uses the `ncurses_shift_function_key` option (you can set it, but it does nothing).

It'd be really nice if at some point, somebody taught Kakoune about [all rxvt's function keys](http://cvs.schmorp.de/rxvt-unicode/doc/rxvt.7.html#Key_Codes), but this is a good start.

PR #3266, opened 2019-12-16
test: number-lines highlighter now reserves 2 digits by default.

Screwtapello:
Fixes regression caused by #3261.


eraserhd:
I just encountered this in my usual Monday-morning rebasing, and this patch fixes it.

PR #3273, opened 2019-12-24
Add a variant of clippy with a santa hat

occivink:
Appears by default during December instead of the regular clippy, and
can be controlled with the new ui option 'ncurses_assistant_festive".
Adding a hat to the cat and Dilbert is left as an exercise.

### DIFF
--- a/doc/pages/options.asciidoc
+++ b/doc/pages/options.asciidoc
@@ -302,6 +302,10 @@ are exclusively available to built-in options.
             specify the nice assistant displayed in info boxes,
             can be *clippy* (the default), *cat*, *dilbert* or *none*
 
+        *ncurses_assistant_festive*:::
+            if *yes* or *true* (the default), the assistant's clothing
+             might change based on the time of year
+
         *ncurses_enable_mouse*:::
             boolean option that enables mouse support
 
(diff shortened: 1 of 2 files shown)


occivink:
Suggestions on how to improve the hat are welcome too, of course

andreyorst:
I'll just leave [this](https://github.com/microsoft/vscode/issues/87268) here. Not that I'm supporting such reoprts or against those, but it seems that such change can be not a welcomed one.

lenormf:
There's a fairly high chance that #87268 was a (successful) trolling attempt.

But this change singles-out one religious holiday, and nowadays that's potentially a bad thing no matter how innocent the original motivations were ("why Christmas and not XXX?", "this makes me feel unwelcome", "genocide" etc).

The patch also adds an option that has no practical use, it would be a double standard to merge that when we refuse all year round new options to users who want to customize actual features.

andreyorst:
> The patch also adds an option that has no practical use, it would be a double standard to merge that when we refuse all year round new options to users who want to customize actual features.

*This is why we can't have ni.. Oh, wait, that's actually prevents bloating*

occivink:
It's been decided that, while cute, the santa hat is a superfluous addition, so I'm closing this.

PR #3276, opened 2019-12-30
Add aliases to the rename commands

alexherbo2:


Delapouite:
This would be consistent with the current `nc` for `rename-client`

PR #3285, opened 2020-01-08
rc jinja: Highlight Jinja statements/expressions

lenormf:
This commit implements a standalone shared highlighter group that
highlights Jinja statements and expressions.

The traditional way of highlighting file contents is to hook on the
file extension, and assign a custom filetype/highlighter group to
the current buffer. However, since Jinja templates can be based on
any text file format in existence, we do not have a specific file
extension to hook, and consequently, no custom "jinja" filetype.

The user is expected to add the `jinja` highlighter whenever required:

```
require-module jinja
add-highlighter window/ ref jinja
```

Alternatively, file extensions that are known to occasionally pair
with Jinja can be hooked from the user configuration:

```
hook global WinCreate .+\.html %[
    try %[
        execute-keys -draft \%s \{%|\{\{ <ret>
        require-module jinja
        add-highlighter window/ ref jinja
    ]
]
```

The above hook auto-detects statements/expressions (respectively
{%…%} and {{…}} expansions), but will cause false positives
(in terms of highlighting), and therefore isn't part of `jinja.kak`
by default.

### DIFF
--- a/rc/filetype/jinja.kak
+++ b/rc/filetype/jinja.kak
@@ -1,0 +1,35 @@
+# https://palletsprojects.com/p/jinja/
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+provide-module jinja %[
+
+require-module python
+
+# Highlighters
+# ‾‾‾‾‾‾‾‾‾‾‾‾
+
+add-highlighter shared/jinja regions
+add-highlighter shared/jinja/comment region '\{#' '#\}' fill comment
+
+# TODO: line statements # …
+
+add-highlighter shared/jinja/statement region '\{%' '%\}' group
+add-highlighter shared/jinja/statement/ ref python
+add-highlighter shared/jinja/statement/ regex \{%[+-]?|[+-]?%\} 0:value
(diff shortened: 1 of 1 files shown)


lenormf:
A oneshot hook that removes the `jinja` highlighter from the window if the original `filetype` of the buffer changes should probably be added.

mawww:
One issue I see with that approach is that jinja highlighter will apply **on top** of the base filetype one,  so we might get strange highlighting inside jinja regions.

I think a way to solve that would be to reset those regions to the `Default` face with a `fill` highlighter, however I cannot see a way to do that with the existing features, as that was what the `exclusive` flag on face was doing, but it got replaced with `final` which is sticky (it makes the face unchangable by following highlighters). We hence would need to reintroduce the exclusive semantics (that just replace the face instead of merging it with the existing one).

lenormf:
Is it worth the trouble through?

From what I've seen, Jinja statements and expressions consistently remain un-highlighted in the typical files they're being used (because they're basically Python on top of HTML, or more generally markup languages). Which prompted the idea of a "hovering" highlighter.

If you have to re-introduce face attributes that we didn't need any more until now to handle the occasional highlighting glitch, maybe you could merge that commit and wait until a user actually meets such an exception before investing time?

Either way, there are more templating/processing languages that could be highlighted the same way (`SimpleTemplate`, `Django`, `ClearSilver`), so I'm keen to find a good solution for this.

mawww:
fair enough.

PR #3286, opened 2020-01-08
2 new keywords for julia.kak & my Copyright Waiver

YohananDiamond:
This pull request adds two keywords that have been recently  added to the Julia specification: `mutable` and `struct`. They have similar functionality to `type` and `immutable`.

Also, sorry for committing the code before the waiver. I read the [contributing guidelines](https://github.com/mawww/kakoune/blob/master/CONTRIBUTING) after doing the commit, so I didn't know until it was too late.


occivink:
It's probably ok like this, but just as a tip: you can use the interactive rebase feature of git (`git rebase -i`) to reorder independent commits like these.

TeddyDD:
Just don't be surprised if github will show them in old order anyway

PR #3287, opened 2020-01-09
better scheme highlighter for char,string,comment

aecepoglu:
This should address highlighting issues with regards to:

* mistaking char `#\;` for beginning of a comment. eg. `(foo #\; another-thing)`
* mistaking char `#\"` for beginning of a string. eg. `(foo #\" another-thing)`
* not highlighting chars

also @eraserhd since he might be interested

### DIFF
--- a/rc/filetype/scheme.kak
+++ b/rc/filetype/scheme.kak
@@ -38,14 +38,15 @@ require-module lisp
 add-highlighter shared/scheme regions
 add-highlighter shared/scheme/code default-region group
 
-add-highlighter shared/scheme/string region '"' (?<!\\)(\\\\)*" fill string
-add-highlighter shared/scheme/comment region ';' '$' fill comment
+add-highlighter shared/scheme/string region %{(?<!#\\)"} (?<!\\)(\\\\)*" fill string
+add-highlighter shared/scheme/comment region %{(?<!#\\);} '$' fill comment
 add-highlighter shared/scheme/comment-form region -recurse "\(" "#;\(" "\)" fill comment
 add-highlighter shared/scheme/comment-block region "#\|" "\|#" fill comment
 add-highlighter shared/scheme/quoted-form region -recurse "\(" "'\(" "\)" fill variable
 
 add-highlighter shared/scheme/code/ regex (#t|#f) 0:value
 add-highlighter shared/scheme/code/ regex \b[0-9]+\.[0-9]*\b 0:value
+add-highlighter shared/scheme/code/ regex (#\\((\w+)|(.))) 0:value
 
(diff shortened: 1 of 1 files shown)


rc/filetype/scheme.kak
```diff
@@ -38,14 +38,15 @@ require-module lisp
...
-add-highlighter shared/scheme/comment region ';' '$' fill comment
+add-highlighter shared/scheme/string region %{(?<!#\\)"} (?<!\\)(\\\\)*" fill string
+add-highlighter shared/scheme/comment region %{(?<!#\\);} '$' fill comment
 add-highlighter shared/scheme/comment-form region -recurse "\(" "#;\(" "\)" fill comment
 add-highlighter shared/scheme/comment-block region "#\|" "\|#" fill comment
 add-highlighter shared/scheme/quoted-form region -recurse "\(" "'\(" "\)" fill variable
 
 add-highlighter shared/scheme/code/ regex (#t|#f) 0:value
 add-highlighter shared/scheme/code/ regex \b[0-9]+\.[0-9]*\b 0:value
+add-highlighter shared/scheme/code/ regex (#\\((\w+)|(.))) 0:value
```
eraserhd: I think having the character names here would be nice, so that I don't try `#\nl` or something.

According to r6rs,
```
 <character name> → nul ∣ alarm ∣ backspace ∣ tab
    ∣ linefeed ∣ newline ∣ vtab ∣ page ∣ return
    ∣ esc ∣ space ∣ delete
```
(r5rs has just space and newline).

So, `(nul\b|alarm\b|backspace\b| ... |.)`?

aecepoglu: I couldn't get the highlights working with `((nul)|(alarm)|...)` (`...` standing for more items in the list). That's why I took this shortcut.

Also, I'm just noticing that instead of `.` I should use `\S`.

eraserhd: Actually, `#\ ` is a valid way of saying `#\space`, weirdly enough.

Alternation works slightly differently in Kakoune, preferring the left-most alternative rather than the longest.  That's bitten me when making a list of words like '(foo|bar|baz)', especially if you put `.` first.  (The nested parens here should be unnecessary, though I guess shouldn't be a problem.)

mawww:
So I gather from the comment this is not ready to be merged yet ?

eraserhd:
@mawww it would still be an improvement.  If you merge, I'll see about being more specific about characters in a separate PR.

aecepoglu:
I haven't had the time to patch this as @eraserhd , sorry about that.

But yes, this is useful as is. Thanks for merging it

PR #3291, opened 2020-01-11
Update the UNLICENSE template to use HTTPS

alexherbo2:


mawww:
This would make the UNLICENSE file mismatch the wording on unlicense.org. When and if they update it, I'll be happy to merge that.

alexherbo2:
https://github.com/unlicense/unlicense.org/commits/master/static/UNLICENSE

mawww:
Ok, seems they have a different version on the page, but fair enough.

PR #3292, opened 2020-01-12
Improve sh syntax highlighting

Ordoviz:
Previously, the keywords were a mess. They contained the shell’s reserved words and some arbitrarily selected builtins. I generated the word list using bash because it contains all POSIX builtins and is common for scripting. I separated keywords and builtins because it felt weird to have `if` and `cd` highlighted in the same colour.

In variable assignments some characters that are allowed to be in variables used to not be highlighted, e.g. hyphens. With this PR all characters except whitespace are considered to be part of the variable.

### DIFF
--- a/rc/filetype/sh.kak
+++ b/rc/filetype/sh.kak
@@ -34,18 +34,6 @@ evaluate-commands %sh{
     keywords="if then else elif fi case esac for select while until do done in
              function time coproc"
 
-    join() { sep=$2; eval set -- $1; IFS="$sep"; echo "$*"; }
-
-    # Add the language's grammar to the static completion list
-    printf %s\\n "declare-option str-list sh_static_words $(join "${keywords}" ' ')"
-
-    # Highlight keywords
-    printf %s "add-highlighter shared/sh/code/ regex \b($(join "${keywords}" '|'))\b 0:keyword"
-
-}
-
-# For some reasons this needs to be in its own %sh block
-evaluate-commands %sh{
     # Generated with `compgen -b` in bash
     builtins="alias bg bind break builtin caller cd command compgen complete
              compopt continue declare dirs disown echo enable eval exec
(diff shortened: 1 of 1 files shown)


rc/filetype/sh.kak
```diff
@@ -43,10 +41,30 @@ evaluate-commands %sh{
 
     # Highlight keywords
     printf %s "add-highlighter shared/sh/code/ regex \b($(join "${keywords}" '|'))\b 0:keyword"
+
+}
+
+# For some reasons this needs to be in its own %sh block
```
danr: The reason why this "needs" to be in its own %sh block is that this printf from the previous block does not write a newline:
```sh
   printf %s "add-highlighter shared/sh/code/ regex \b($(join "${keywords}" '|'))\b 0:keyword"
```
￼Can you fix this and merge the two %sh blocks please?

Ordoviz: Thanks for pointing me at the cause of the issue!

rc/filetype/sh.kak
```diff
@@ -30,23 +30,32 @@ add-highlighter shared/sh/double_string/fill fill string
...
-    printf %s "add-highlighter shared/sh/code/ regex \b($(join "${keywords}" '|'))\b 0:keyword"
+    printf %s\\n "add-highlighter shared/sh/code/ regex \b($(join "${keywords}" '|'))\b 0:keyword"
+
+    # Highlight builtins
+    printf %s "add-highlighter shared/sh/code/builtin regex \b($(join "${builtins}" '|'))\b 0:builtin"
 }
 
 add-highlighter shared/sh/code/operators regex [\[\]\(\)&|]{1,2} 0:operator
-add-highlighter shared/sh/code/variable regex (\w+)= 1:variable
+add-highlighter shared/sh/code/variable regex (\S+)= 1:variable
```
lenormf:
```suggestion
add-highlighter shared/sh/code/variable regex ([\w_]+)= 1:variable
```

Ordoviz: The underscore is already part of `\w` but the hyphen isn’t for example. Is there any scenario where `\S` doesn’t highlight the variable properly?

lenormf: In that case the following is not a valid variable symbol in shell: `a!b`.

Ordoviz: I use `alias backup-home='some command'` and it would be nice if the whole alias was highlighted, not only the `home` part. I think that’s more important than checking whether variables names are syntactically correct.

lenormf: That was just one example, there are more invalid symbols that would be highlighted with your proposed change: `a#b`, `a*b` etc.

```suggestion
add-highlighter shared/sh/code/variable regex ([\w-]+)= 1:variable
```

rc/filetype/sh.kak
```diff
@@ -30,23 +30,32 @@ add-highlighter shared/sh/double_string/fill fill string
...
-    printf %s "add-highlighter shared/sh/code/ regex \b($(join "${keywords}" '|'))\b 0:keyword"
+    printf %s\\n "add-highlighter shared/sh/code/ regex \b($(join "${keywords}" '|'))\b 0:keyword"
+
+    # Highlight builtins
+    printf %s "add-highlighter shared/sh/code/builtin regex \b($(join "${builtins}" '|'))\b 0:builtin"
 }
 
 add-highlighter shared/sh/code/operators regex [\[\]\(\)&|]{1,2} 0:operator
-add-highlighter shared/sh/code/variable regex (\w+)= 1:variable
+add-highlighter shared/sh/code/variable regex ([\w-]+)= 1:variable
```
mawww: which shell allows minus in identifiers ?

lenormf: This was discussed [above](https://github.com/mawww/kakoune/pull/3292#pullrequestreview-342342126), it seems that the hyphen was added to support highlighting aliases.

Issue #3230, opened 2019-11-28
Handlebars (.hbs) not highlighting correctly


velrest:
On opening a `*.hbs` file i see following error message in `*debug*`:
```
error running hook WinSetOption(filetype=hbs)/hbs-highlight: 2:6: 'maybe-add-hbs-to-html' 1:2: 'evaluate-commands' 2:14: 'add-highlighter' no such id: 'html'
```

This is probably caused by [this line](https://github.com/mawww/kakoune/blob/master/rc/filetype/hbs.kak#L97).

### Steps
- Open a `*.hbs` file

### Outcome
- The file has no syntax highlighting

### Expected
- The file should be highlighted with the html-highlighter and the hbs additions

lenormf:
The file needs a `require-module html`, after line 12. Want to test that and create a PR? :)

velrest:
Sure :+1:

velrest:
The error is now gone but now `recursive call of hook WinSetOption/filetype=hbs, not executing` occurs and still no syntax highlighting. Sorry im new to kakoune and do not yet understand the hooks completely.

lenormf:
The following fixes it for me:

```diff
diff --git a/rc/filetype/hbs.kak b/rc/filetype/hbs.kak
index df854076..eb788b8b 100644
--- a/rc/filetype/hbs.kak
+++ b/rc/filetype/hbs.kak
@@ -10,6 +10,7 @@ hook global BufCreate .*[.](hbs) %{
 
 hook global WinSetOption filetype=hbs %{
     require-module hbs
+    require-module html
 
     hook window ModeChange pop:insert:.* -group hbs-trim-indent hbs-trim-indent
     hook window InsertChar \n -group hbs-indent hbs-indent-on-new-line
```

I'm just concerned now about `maybe-add-hbs-to-html`, `hbs_highlighters_enabled` and the reason for their existence.

The default regions should probably be HTML, and sub-regions highlight handlebars code - as opposed to having the module inject its highlighters into `html`.

velrest:
Ah true for me too, sorry not sure what i did wrong earlier. Thanks! Should i close this or leave it open for this issue?
> I'm just concerned now about `maybe-add-hbs-to-html`, `hbs_highlighters_enabled` and the reason for their existence.
> 
> The default regions should probably be HTML, and sub-regions highlight handlebars code - as opposed to having the module inject its highlighters into `html`.

lenormf:
[...]

PR #3296, opened 2020-01-15
Add html module to hbs filetype to correctly highlight hbs files

velrest:
Previously the handlebars filetype threw an error since the hbs highlighter depends on the `html` module. This resolves part of https://github.com/mawww/kakoune/issues/3230 .

### DIFF
--- a/rc/filetype/hbs.kak
+++ b/rc/filetype/hbs.kak
@@ -29,6 +29,8 @@ hook -group hbs-highlight global WinSetOption filetype=hbs %{
 
 provide-module hbs %[
 
+require-module html
+
 # Highlighters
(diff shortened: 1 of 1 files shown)


rc/filetype/hbs.kak
```diff
@@ -10,6 +10,7 @@ hook global BufCreate .*[.](hbs) %{
 
 hook global WinSetOption filetype=hbs %{
     require-module hbs
+    require-module html
```
mawww: I think this should be one of the first thing the module itself does (so inside provide-module) instead of after running it. `require-module hbs` should not rely on the caller also doing `require-module html`.

PR #3297, opened 2020-01-16
Backport new downstream spec file to upstream

jkonecny12:
The question is if you want to keep this spec file which will be updated from downstream, or we should have only the downstream spec.

What do you think?

### DIFF
--- a/contrib/kakoune.spec
+++ b/contrib/kakoune.spec
@@ -1,11 +1,11 @@
 %bcond_without tests
 
-### Enable LTO. Profit ~8%
+# Enable LTO. Profit ~8%
 %global optflags        %{optflags} -flto
 %global build_ldflags   %{build_ldflags} -flto
 
 Name:           kakoune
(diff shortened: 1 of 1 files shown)


mawww:
If the main version is the downstream one, then maybe we can just remove the upstream. What value does the upstream version provide ?

jkonecny12:
Hmm, when I'm thinking about it, it may be sensible to have it here. In case that we don't want to / can't update the downstream one or there are dependencies which are not yet part of the upstream one.

What would be great to make everyone life easier is to have some script for releases to update even this spec file automatically updated when new release is created. @mawww are you using something like that to make a new release or are you doing everything manually? If there is something I can add this spec file to your solution.

jkonecny12:
Could we please merge this to fix daily builds on Fedora? Or should I solve this differently to remove the burden from you?

jkonecny12:
Related to this https://github.com/mawww/kakoune/issues/3323.

PR #3298, opened 2020-01-16
Fix version notes formatting

maximbaz:
Just a minor thing, I feel like some newlines were forgotten and formatting of the last news was a bit inconsistent... Let me know if I misunderstood and the current formatting is intended.

Before:

image

After: 

image

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -45,11 +45,11 @@ struct {
     StringView notes;
 } constexpr version_notes[] = { {
         20200116,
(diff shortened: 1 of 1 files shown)


mawww:
The use of *Mode*Begin formatting with Mode in itallic was there to indicate it was a place holder (the actual hooks are InsertBegin, NormalBegin, ...).

It was written as a single change because it was related to the ModeChange update, but splitting it in two changes might make it a bit easier to read.

maximbaz:
Ah makes sense, updated 👍

PR #3300, opened 2020-01-17
Give an explicit name to the Crystal highlighter

alexherbo2:


lenormf:
Does that fix any specific issue? I think the auto-generated name for this highlighter would be `ref_crystal`, it seems relevant enough?

PR #3301, opened 2020-01-18
Add a source-recursive command

### LINKED ISSUE #3299: Add a source-recursive command
It could be useful to extract that command from the [`kakrc`].

[`kakrc`]: https://github.com/mawww/kakoune/blob/master/share/kak/kakrc

alexherbo2:
Closes #3299

### DIFF
--- a/doc/pages/commands.asciidoc
+++ b/doc/pages/commands.asciidoc
@@ -114,6 +114,9 @@ of the file onto the filesystem
     execute commands in <filename>
     parameters are available in the sourced script as `%arg{0}`, `%arg{1}`, …
 
+*source-recursive* <directory>...::
+    Source recursively for `.kak` files in the given directories.
+
 == Clients and Sessions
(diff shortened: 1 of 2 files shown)


lenormf:
Wouldn't it be better to have `:source-file` aliases by `:source`, and `:source-directory` implemented as a C++ builtin command? The amount of scripts shipped with the editor by default is ever-growing, maybe we would see some gains at load-time.

Needs to be measured of course.

occivink:
Yes, or maybe a `-recursive` switch to `source`. It would also bypass possible escaping issues, and still be available if you do `kak -n`.

chambln:
Why not make the `source` command recursive anyway? The decision to source recursively can be inferred by the argument:

- `source /path/to/dir/` will source `dir/` recursively
- `source /path/to/script.kak` will just source `script.kak` as normal.

PR #3308, opened 2020-01-20
fix several control key combinations in st

jo-he:
this patch is mostly important for the **Delete** key: it generated **F1** before

### DIFF
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -691,8 +691,25 @@ Optional<Key> NCursesUI::get_next_key()
         case 'D': return masked_key(Key::Left);
         case 'F': return masked_key(Key::End);   // PC/xterm style
         case 'H': return masked_key(Key::Home);  // PC/xterm style
+        case 'J':
+            switch (params[0])
+            {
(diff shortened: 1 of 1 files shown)


mawww:
As this adds several lines, could you add a copyright waiver commit (see CONTRIBUTING) ?

jo-he:
This is not the first time I contribute. The copyright waiver was already added by #3006 (commit ab87d2b2cbf1a66fa2195f6ab0d41e5e001e185a)

mawww:
Right, sorry about that, for some reason I forgot to add you as a remote in my local git (which I usually do once I have a copyright waiver merged).

mawww:
I realized this is probably the wrong way to fix that issue, st seems to send those if it is not in 'application keypad' mode, if it is it will use the more common format that Kakoune already supports, so we need to ensure we set that mode.

jo-he:
Good catch! I was not aware of key modes (just had a look at https://www.tldp.org/HOWTO/Keyboard-and-Console-HOWTO-21.html). When both, "application cursor key mode" and "application keypad key mode", are set, it seems that they would simplify the keycode handling (I did only test in st).
Maybe we would not even need to distinguish between xterm and rxvt anymore?

PR #3310, opened 2020-01-21
Fix version comparison in show_startup_info

omasanori:
The description of startup_info_version in the manual says "only messages relating to a Kakoune version greater than this value will be displayed," but showed messages relating to the version equal to that value.
This change aligns the code with the manual and makes a workaround that set startup_info_version next to the original version (ex. 20200117) unnecessary.


omasanori:
Thank you!

PR #3312, opened 2020-01-21
Use POSIX -E of sed for extended regular expressions

alexherbo2:


lenormf:
I don't see either flag being defined in [the standard](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sed.html), have you met an implementation that supports one but not the other?

Busybox, GNU and BSD `sed` support both, it should be fairly safe?

PR #3317, opened 2020-01-25
src: Show the startiup information in no-load mode

lenormf:
This commit allows the changelog to be shown at startup even when
the editor was run with the -n flag.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -1160,10 +1160,11 @@ int main(int argc, char* argv[])
             StringView session = parser.get_switch("s").value_or(StringView{});
             try
             {
-                auto flags = (parser.get_switch("n") ? ServerFlags::IgnoreKakrc : ServerFlags::None) |
-                             (parser.get_switch("d") ? ServerFlags::Daemon : ServerFlags::None) |
-                             (parser.get_switch("ro") ? ServerFlags::ReadOnly : ServerFlags::None) |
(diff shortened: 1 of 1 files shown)


src/main.cc
```diff
@@ -1160,10 +1160,11 @@ int main(int argc, char* argv[])
...
             {
-                auto flags = (parser.get_switch("n") ? ServerFlags::IgnoreKakrc : ServerFlags::None) |
-                             (parser.get_switch("d") ? ServerFlags::Daemon : ServerFlags::None) |
-                             (parser.get_switch("ro") ? ServerFlags::ReadOnly : ServerFlags::None) |
-                             (argc == 1 and isatty(0) ? ServerFlags::StartupInfo : ServerFlags::None);
+                auto ignore_kakrc = (bool)parser.get_switch("n");
+                auto flags = (ignore_kakrc                              ? ServerFlags::IgnoreKakrc : ServerFlags::None) |
+                             (parser.get_switch("d")                    ? ServerFlags::Daemon      : ServerFlags::None) |
+                             (parser.get_switch("ro")                   ? ServerFlags::ReadOnly    : ServerFlags::None) |
+                             ((argc == 1 or ignore_kakrc) and isatty(0) ? ServerFlags::StartupInfo : ServerFlags::None);
```
mawww: `(ignore_kakrc and argc == 2)` as we do not want to show the startup message if we have been given a file to open.

lenormf: But now the message won't show when no arguments are passed to the editor!

Issue #3091, opened 2019-09-19
Marks do not work when `@` is in the buffer path


BestFriendChris:
### Steps 
```sh
$ mkdir /tmp/@foo
$ echo "quux\nbaz" > /tmp/@foo/bar
$ kak /tmp/@foo/bar -e 'exec eZjz'
```

### Outcome 
The following error message:
```
1.1: 'exec' expected <buffer>@<timestamp>@main_index
```

### Expected 
The word `quux` selected and the following info message:
```
Restored 1 selections from register '^'
```

### Analysis
The `^` register has the following content:
```
/tmp/@foo/bar@1@0
```
which is being split into 4 sections instead of the expected 3 at [read_selections_from_register()](https://github.com/mawww/kakoune/blob/271b1d4f2ff85668be6e1228b3a3fc957a740adf/src/normal.cc#L1761-L1791)

### Proposed Fix
I think the best solution would be to only split on the last 2 `@` signs since the `timestamp` and `main_index` portions are defined internally to be integers.

Alternatively, when saving and restoring a mark, any `@` characters in the buffer name could be escaped in some way.

Delapouite:
@BestFriendChris can you check if your problem is resolved on master and close this issue if that's the case? thanks

mawww:
Should be fixed, closing, re-open if still an issue.

PR #3319, opened 2020-01-26
Add support for @ symbols in selection buffer name

geppettodivacin:
The selection descriptions use the format
`<buffer>@<timestamp>@<main_index>`. This fails when file paths have `@`
symbols in them: the parser splits on `@` symbols and finds more values
than it expects.

We here modify the behavior to require *at least* two `@` symbols, using
the last two for <timestamp> and <main_index> and leaving the remaining
text for the <buffer>. This should work for any number of `@` symbols,
since `<timestamp>` and `<main_index>` are numbers and should never contain
`@` symbols.

### DIFF
--- a/src/ranges.hh
+++ b/src/ranges.hh
@@ -73,6 +73,24 @@ inline auto skip(size_t count)
     });
 }
 
+template<typename Range>
+struct DropView
+{
+    auto begin() const { return std::begin(m_range); }
+    auto end()   const { return std::end(m_range) - m_drop_count; }
+
+    Range m_range;
+    size_t m_drop_count;
+};
+
+inline auto drop(size_t count)
+{
+    return make_view_factory([count](auto&& range) {
+        using Range = decltype(range);
+        return DropView<decay_range<Range>>{std::forward<Range>(range), count};
+    });
+}
+
 template<typename Range, typename Filter>
 struct FilterView
 {
--- a/src/ranges.hh
+++ b/src/ranges.hh
@@ -37,6 +37,12 @@ struct ReverseView
 {
     decltype(auto) begin() { return m_range.rbegin(); }
     decltype(auto) end()   { return m_range.rend(); }
+    decltype(auto) rbegin() { return m_range.begin(); }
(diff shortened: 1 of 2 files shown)


geppettodivacin:
I thought this worked, but it's got a few problems leftover. I'll open a new one after I work out the kinks.

geppettodivacin:
It doesn't actually work if there is more than one @ symbol in the filename. I haven't worked out why yet.

geppettodivacin:
Also, it breaks lisp indentation, and quoted selections? The code is kind of difficult to reason about. I'm probably overdoing the functional style.

geppettodivacin:
Intended to fix #3091

mawww:
i dont think i have a copyright waiver commit from you, could you add one ?

geppettodivacin:
You bet. Can it go in the same PR?

mawww:
Yep

src/ranges.hh
```diff
@@ -37,6 +37,12 @@ struct ReverseView
 {
     decltype(auto) begin() { return m_range.rbegin(); }
     decltype(auto) end()   { return m_range.rend(); }
+    decltype(auto) rbegin() { return m_range.begin(); }
+    decltype(auto) rend()   { return m_range.end(); }
+    decltype(auto) begin() const { return m_range.rbegin(); }
+    decltype(auto) end() const  { return m_range.rend(); }
+    decltype(auto) rbegin() const { return m_range.begin(); }
+    decltype(auto) rend() const  { return m_range.end(); }
```
lenormf: Are these used in the code?

geppettodivacin:
I'm retrospect they should all be const, but yes, the compiler was
complaining about discarding qualifiers quite a bit.

Or at least it was when I was mucking about with the types. The latest
patch (suggested by mawww) may not require the const qualifier.

We do need the rbegin and rend so that we can reverse a reverse and get
back the original.

geppettodivacin:
... Oh, the transform call no longer uses reverse, so I guess not,
actually...

PR #3325, opened 2020-01-29
rc git: Improve git wrapper highlighters

chambln:
This fixes highlighting of

- `--pretty=oneline` and `--graph` formats for the `git-log` filetype
- `--short` and `--branch` formats for the `git-status` filetype


lenormf:
Maybe it's time a proper shared highlighter group is implemented for this filetype, considering the amount of regex.

Issue #3306, opened 2020-01-19
How to mimic the completion of set-option in define-command?


alexherbo2:

Delapouite:
Here's a raw idea about command completion in general.

Currently, the `define-command` command has a gazillion flags related to the concept of completion. Here are a few: `-file-completion`, `-client-completion`, `-buffer-completion`…
They are not very flexible since they apply to all arguments passed to the command.

Maybe one way to offer more versatility / composable, would be to introduce a new `complete` command, akin to the one found in shells like fish: https://fishshell.com/docs/current/commands.html#complete

The goal would be to first define commands, and then "attach" completion capabilities to it. In very pseudo code:

```
define-command -params 2 foo %{ … }
complete foo arg1 buffers
complete foo arg2 file
```

lenormf:
I've wanted that kind of command for a while as well, but I used it for #3327. If that PR gets merged, it would be good to extend `:complete` to follow the Fish API more.

PR #3327, opened 2020-01-30
src: Implement custom insert-mode completers

### LINKED ISSUE #3222: Trigger a specific completion-engine from a mapping
Hi,

I'd like to be able to complete (in insert mode) a prefix using a completion-engine of my choosing, with a mapping. There does not need to be any menu-ing involved, the name can be hardcoded in the mapping.

One use case would be to trigger Clang completion at will, instead of letting the editor go through all the ones stored in `completers`.

Thanks!

lenormf:
This commit allows users to declare custom insert completion methods.

A new command `complete` is introduced that takes a key (to hit in
the <c-x> menu) and an option from which the completion candidates
will be loaded, as argument.

The `InsertUserCompletion` hook, triggered before the option is
read, allows setting candidates dynamically.

Examples:

```
declare-option -hidden str-list my_dictionary Kakoune
complete -docstring "personal dictionary" d my_dictionary
exec i<c-x>d
```

```
declare-option -hidden str-list my_dictionary
complete -docstring "dynamic dictionary" t my_dictionary
hook buffer InsertUserCompletion .* %{ %sh{ printf 'set buffer my_dictionary %s' … } }
exec i<c-x>d
```

Fixes #3222

### DIFF
--- a/doc/pages/commands.asciidoc
+++ b/doc/pages/commands.asciidoc
@@ -193,6 +193,11 @@ of the file onto the filesystem
     *-lock*:::
         stay in mode until `<esc>` is pressed
 
+*complete* [-docstring <docstring>] <key> <option>::
+    declare a user completion engine mapped to `<key>` and taking candidates
+    from `<option>`; if specified, `<docstring>` will be shown in the
+    insert completion menu to describe the mapping (See <<keys#insert-mode-completion,`:doc keys insert-mode-completion`>>)
+
 == Hooks
 
 *hook* [-group <group>] <scope> <hook_name> <filtering_regex> <command>::
--- a/doc/pages/hooks.asciidoc
+++ b/doc/pages/hooks.asciidoc
@@ -178,6 +178,9 @@ name. Hooks with no description will always use an empty string.
 *ModuleLoaded* `module`::
     Triggered after a module is evaluated by the first `require-module` call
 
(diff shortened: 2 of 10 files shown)


lenormf:
Also included in this PR are two commits that address small issues in the current code.

danr:
Hi, cool patch, here are some suggestions:

1. The name `complete` for the command is too generic, it should be more specific like `define-completion-key` or something involving insertion mode completion menu.

2. You write _Triggered when a user completion mapping is hit, before the option associated with it is read to return a list of completion candidates_, what happens if the hook asynchronously sets the option? Will it be updated?

3. You probably want to change `buffer` to `window` in `hook buffer InsertUserCompletion .* %{ %sh{ printf 'set buffer my_dictionary %s' … } }`

HTH

lenormf:
I kept the name generic because ideally we would unify command arguments completion from #3306 with the insert completers.

If the hook starts an asynchronous shell, then the option will be updated when the shell exits. If that's not desirable, the shell has to be made synchronous.

I store highlighting information in the `window` scope, and "content" information in the buffer. Feel free to change it in your own configuration, those snippets are just generic examples to see what the feature does at a high level.

lenormf:
I'd also like to use this to have a custom completer that completes command names when editing a shell script.

lenormf:
Rebased on master.

PR #3341, opened 2020-02-04
contrib describe_sessions: Look for sessions in XDG_RUNTIME_DIR

lenormf:
The sessions are stored in `$XDG_RUNTIME_DIR` when that variable
is defined.


lenormf:
Thanks @alexherbo2 for the bug report.

PR #3342, opened 2020-02-04
make: use 'info' instead of 'echo'

occivink:
The message is much more prominent this way.
In addition, remove leading whitespace from the message.

### DIFF
--- a/rc/tools/make.kak
+++ b/rc/tools/make.kak
@@ -44,7 +44,7 @@ declare-option -docstring "name of the client in which all source code jumps wil
 define-command -hidden make-open-error -params 4 %{
     evaluate-commands -try-client %opt{jumpclient} %{
         edit -existing "%arg{1}" %arg{2} %arg{3}
-        echo -markup "{Information}{\}%arg{4}"
+        info "%arg{4}"
         try %{ focus }
     }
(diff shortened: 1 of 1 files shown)


mawww:
I am wondering about this, as it will display on top of the buffer. I am not sure we gain much by moving from the status line to an info box.

occivink:
I personally prefer it but maybe we can get some other opinions on this?

lenormf:
I don't care for UI messages much, and would send them to the `*debug*` buffer if I could, as this is the first place I would look if the expected side effects for a command haven't visibly occurred.

I could also see a user sending all messages to a `*notifications*` buffer as well, to keep them within eyesight in a pane, but not letting them pollute the main pane.

Since we now have three different preferences regarding those, maybe a hook/option could be added that is in charge of delivering notifications, defaulting to `echo` if undefined?

Delapouite:
> maybe a hook/option could be added that is in charge of delivering notifications, defaulting to echo if undefined

Yes that's would be a nice way to handle it, a hook letting users decide what to do when a message is about to displayed with `echo`, `info`…

occivink:
Closing, as it seems to be a matter of preference

PR #3356, opened 2020-02-09
src: Let `:set-register` complete register names

lenormf:
This commit allows the `set-register` command to suggest candidates
named after punctuated registers, similarly to %reg{…} expansions.


lenormf:
A possible improvement would be to also complete on the letter-registers that are currently set.

Delapouite:
Nice complement to https://github.com/mawww/kakoune/pull/2074

Screwtapello:
I want this feature because I can't always immediately remember whether the `^` register is named "caret" or "hat", and I think most English speakers would not guess "arobase" without a hint like interactive completion.

PR #3362, opened 2020-02-10
src: Handle errors with -f more consistently

lenormf:
This commit modifies the behaviour of the filter flag -f when errors
have occurred:

* prevent the buffer from being printed
* consistently set the exit code to 1

Errors not directly related to the execution of the given keys are
still being printed even in "quiet" mode, as they are likely to be
unrelated to a logical mistake and thus potentially critical.


lenormf:
Arguably we don't really care about the jumping to a label to make sure the buffer trash is cleared, as the process exits immediately after executing the filter.

This commit also doesn't execute the `buffer_manager.delete_buffer(*buffer);` lines after an error has been raised, for the reason above. Could be done though, I suppose.

src/main.cc
```diff
@@ -908,9 +914,8 @@ int run_filter(StringView keystr, ConstArrayView<StringView> files, bool quiet,
             }
             catch (runtime_error& err)
             {
-                if (not quiet)
-                    write_stderr(format("error while applying keys to buffer '{}': {}\n",
-                                        buffer.display_name(), err.what()));
+                throw filter_key_error{format("could not apply keys to buffer '{}': {}",
+                                              buffer.display_name(), err.what())};
```
mawww: When given multiple files, this will stop all processing at the first failing one, which I dont think would be the desired behaviour.

src/main.cc
```diff
@@ -934,13 +939,23 @@ int run_filter(StringView keystr, ConstArrayView<StringView> files, bool quiet,
             buffer_manager.delete_buffer(buffer);
         }
     }
+    catch (filter_key_error& err)
+    {
+        if (not quiet)
+            write_stderr(format("error: {}\n", err.what()));
+        goto error_cleanup;
```
mawww: use on_scope_end to ensure clear_buffer_trash is called, and avoid goto, as it is a pretty dangerous construct in C++ (due to constructor/destructors). Here we can just return 1, which removes the need for the exit_code variable.

lenormf: We can return 1 if you don't care about freeing up memory right before the process exits, c.f. [my original reply](https://github.com/mawww/kakoune/pull/3362#issuecomment-584220469).

If you want to keep Valgrind happy I'll use `on_scope_end`, otherwise `return 1`, but I need to know which.

src/main.cc
lenormf: I did this on purpose, if errors do not interrupt the processing of the subsequent files, then there's no need to have an exit code - there will be no errors, only warnings.

Users who don't care can wrap the macro in `:try "…"<ret>`, or even spawn `kak -f` on each individual file (ignoring the non-zero exit code).

Users who expect their macro to apply will know immediately something is off, it seems inconvenient to keep executing keys further when there has already been an error.

mawww: It seems that using `kak -f 'sfoo<ret>cbar<esc>' directory/*` is a reasonable use case for replacing all foo with bar in files appearing in the directory.

With this logic, you would end up with a set of file already replaced, and all the files following the first one not containing foo not replaced. A sane behaviour would be to either apply to all files, or none, but not an unpredicitible partial set.

src/main.cc
mawww: Its not only valgrind, clear_buffer_trash is when buffers get effectively deleted, and run eventual hooks, when we do it has semantic meaning, not only memory freeing side effects.

Using on_scope_end ensures that gets called when we return, regardless of which return statement we use.

src/main.cc
lenormf: What about the cases whereby users expect their macro to apply perfectly to *all* the files passed as argument, or when the string of primitives errors out in the middle of it, and not at the very beginning?

Why would the editor apply what it knows is a broken macro to all the files, leaving them *all* in a broken state?

We can't assume *all errors* can safely be ignored because of the one case that would allow `-f` to be used as a drop-in replacement for `sed`.

mawww: Regarding `sed` I believe it will just apply to all files, not complaining about files that do not have any match.

Also, in my view, `sfoo<ret>cbar<esc>` on a file that does not contain any foo does not fail, it mearly stops early because it got to an empty selection set. It still did exactly what was asked for, replacing all `foo` with `bar`, wether the user expects all files to contain foo or not is something Kakoune can not guess.

lenormf:
Rebased on `master`, errors don't stop the execution flow any more, and the exit code can be 1 or 2 on error to indicate what problem occurred.

lenormf:
Rebased on master.

PR #3366, opened 2020-02-13
Work in progress: Add CMake build system

CharlesGueunet:
THIS IS A WORK IN PROGRESS AND SHOULD NOT BE MERGED YET.

This PR bring the support for the CMake ecosystem.

* [x] Config and build using cmake (tested on Arch linux and Gentoo)
* [ ] Packaging using cpack (almost done)
* [ ] Testing using ctest
* [ ] CI using cdash

# Build configuration

Using CMake, kakoune can be compiled using gcc / clang and the build system may be make or ninja. Libs can be linked dynamically or statically.
The version of kakoune is by default the current date (keeping the same syntax than the current makefile) but can also be overriden by hand.

# Packaging

Currently, tar.gz and tar.Z archive as well as a .sh file are generated.
I am working on the rpm, deb and PKGBUILD (aur) files.

# Testing

TODO

# CI

TODO


Screwtapello:
Interesting! But... why CMake specifically?

mawww:
It is unclear to me what we gain from using CMake for Kakoune. The current build system is pretty simple, support static and dynamic builds, support different toolchains by just setting CXX to point to the correct compiler, and Kakoune is small enough that I would not expect ninja to gain any significant build time advantage.

Dont get me wrong, I like CMake and use it quite a lot at work, but it is a hard sell to re-introduce an external dependency to Kakoune when the last one (NCurses) is close to going away.

CharlesGueunet:
> Interesting! But... why CMake specifically?

CMake is a versatile environment that is aimed to be OS agnostic. I see several advantages compared to Makefile:
* For the user, it is more convenient to change compilation parameters. Using ccmake or cmake-gui gives a nice interface.
* For developers, I personally find `CMakeLists.txt` easier to understand than Makefiles. 
* For developers, cpack is able to generate *.deb*, *.rpm* and soon *PKGBUILD* with few configurations.
* For developers, the ctest infrastructure let you choose easily which tests to run depending on the build parameters. It can be coupled with CI tools like travis easily. It suppports regex to choose which tests to launch.

I recognize all these points are subjects to discussion, and I don't want to impose CMake if you don't want to use it. I personally like it, but I have to declare a conflict of interest: I work at Kitware, the corporation that maintain CMake.

>  The current build system is pretty simple, support static and dynamic builds, support different toolchains by just setting CXX to point to the correct compiler.

That is true, make is a really good tool and on this point CMake and make are quite equivalent.

> ... Kakoune is small enough that I would not expect ninja to gain any significant build time advantage.

This is absolutely true. Ninja still have the advantage to build in parallel bu default for people that do not add `-j` or haven't customized their `make.conf`. This is a minor advantage though.

> Dont get me wrong, I like CMake and use it quite a lot at work, but it is a hard sell to re-introduce an external dependency to Kakoune when the last one (NCurses) is close to going away.

In my point of view, CMake is only a build dependency, not a dependency of the Kakoune binary. Similarly to make, it is only used for compilation so binary based distribution won't need to have CMake as a dependency for Kakoune. For source-based distribution, I really think CMake will already be installed in most cases ^^. 

I repeat myself, if you don't think Kakoune would benefit from CMake, you can just close this PR. I will continue to maintain it on my side as I compile Kakoune myself on my various devices. You may then re-open the PR if you change your mind.
On the opposite, if you are interested, feel free to review the code and ask any question about it :)

lenormf:
Any way you can keep this contained within the `contrib` director, as was done with `Tupfile`?

mawww:
I'll be ok adding that to the contrib/ directory, similar to what we do for the Tupfile, but I'd like to keep dependencies (build + run ones) as low as possible, so I'd rather keep the main build system Makefile based.

CharlesGueunet:
I understand. There is not only a CMakeLists.txt file at the root but also a file in the src folder. I fear that refactoring to a single file in contrib folder would make it bad code (and so nobody would use it). I will then keep this on my side. Please feel free to reopen if one day you think CMake would be beneficial to kakoune.

PR #3378, opened 2020-02-20
Use filetype ini for XDG desktop entry files

chambln:
See <https://freedesktop.org/wiki/Specifications/desktop-entry-spec/>


chambln:
While we're at it, are there any other file extensions to consider? Maybe `*.conf`?

PR #3380, opened 2020-02-21
Allow to choose the executabe in the run script (for tests)

CharlesGueunet:
Hello,

This PR allows to specify the executable to use when running the tests with the run script.
It does not change to current behavior but only add a new option (-k) to specify the executable used for the tests.
We can use this for example to test the installed executable and not the one in the build folder.

Charles

### DIFF
--- a/test/run
+++ b/test/run
@@ -24,9 +24,21 @@ main() {
       '
 
   root=$PWD
+  kak_exec=$root/../src/kak
   tmpdir="${TMPDIR:-/tmp}"
   work=$(mktemp -d $tmpdir/kak-tests.XXXXXXXX)
   trap "rm -R $work" EXIT
+
+  # We can specify kakoune executable with -k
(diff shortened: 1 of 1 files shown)


mawww:
Hello,

Sorry for the delay in reviewing that,

Could you expand a bit on what use case led you to need this switch ? Its a bit unclear to me why we would want to test another binary, especially as the tests check for the current expected behaviour, which can change from a release to the next.

Cheers,

Maxime.

CharlesGueunet:
Hello,

The goal of these changes is to allow to specify the binary to test wherever it is installed on the machine (and not only the one in the src/ folder). In my case, I use it as I have several version of kakoune installed with different compilers (Clang, GCC) and different options (Release  / Debug, Static / Dynamic) to test.
It is not intended to test a binary of another release indeed.

Charles

PR #3382, opened 2020-02-23
Add a new 'arrange-buffers' to let users change the order of the buflist

occivink:
This implements [this suggestion](https://discuss.kakoune.com/t/brainstorm-for-a-more-flexible-buffer-list/767/4) by @Screwtapello

### DIFF
--- a/doc/pages/commands.asciidoc
+++ b/doc/pages/commands.asciidoc
@@ -14,6 +14,12 @@ command *q!* has to be used). Aliases are mentionned below each commands.
 For the following *write* commands, the *-sync* switch forces the synchronization
 of the file onto the filesystem
 
+*arrange-buffers* <buffer>...::
+    Reorder the buffers in the buffers list.
+    The named buffers will be moved to the front of the buffer list, in the order
+    given. Buffers that do not appear in the parameters will remain at the
+    end of the list, keeping their current order.
+
 *change-directory* [<directory>]::
     *alias* cd +
     change the current directory to *directory*, or the home directory if
@@ -338,7 +344,7 @@ but not really useful in that context.
 
     *-title* <text>:::
         set the title of the message box
-        
+
     *-markup*:::
         parse markup in both title (if provided) and text. (See
         <<faces#markup-strings,`:doc faces markup-strings`>>)
@@ -369,7 +375,7 @@ but not really useful in that context.
 
 *select* [<switches>] <anchor_line>.<anchor_column>,<cursor_line>.<cursor_column>...::
     replace the current selections with the ones described in the arguments
-    
+
     *-timestamp* <timestamp>:::
         specify which buffer timestamp those coordinates apply to. Uses current
         buffer timestamp if not specified.
--- a/src/buffer_manager.cc
+++ b/src/buffer_manager.cc
@@ -105,4 +105,29 @@ void BufferManager::clear_buffer_trash()
     m_buffer_trash.clear();
 }
 
+void BufferManager::arrange_buffers(ConstArrayView<String> first_ones)
+{
+    Vector<size_t> indices;
+    for (const auto& name : first_ones)
+    {
+        auto it = find_if(m_buffers, [&](auto& buf) { return buf->name() == name or buf->display_name() == name; });
+        if (it == m_buffers.end())
+            throw runtime_error{format("no such buffer '{}'", name)};
+        size_t index = it - m_buffers.begin();
+        if (contains(indices, index))
+            throw runtime_error{format("buffer '{}' appears more than once", name)};
+        indices.push_back(index);
+    }
+
+    BufferList res;
+    for (size_t index : indices)
+        res.push_back(std::move(m_buffers[index]));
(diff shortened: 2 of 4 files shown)


doc/pages/commands.asciidoc
```diff
@@ -14,6 +14,12 @@ command *q!* has to be used). Aliases are mentionned below each commands.
 For the following *write* commands, the *-sync* switch forces the synchronization
 of the file onto the filesystem
 
+*arrange-buffers* <buffer>...::
+    reorder the buffers in the buffers list
+    the list will be sorted following the order specified in the parameters
```
Screwtapello: I worry that the word "sorted" will make people expect some kind of comparison function. How about:

> The named buffers will be moved to the front of the buffer list, in the order given

src/buffer_manager.cc
```diff
@@ -105,4 +105,29 @@ void BufferManager::clear_buffer_trash()
     m_buffer_trash.clear();
 }
 
+void BufferManager::arrange_buffers(ConstArrayView<String> first_ones)
+{
+    Vector<size_t> indices;
+    for (const auto& name : first_ones)
+    {
+        auto it = find_if(m_buffers, [&](auto& buf) { return buf->name() == name or buf->display_name() == name; });
```
lenormf: I think you can just compare the display name here.

src/buffer_manager.cc
```diff
@@ -105,4 +105,29 @@ void BufferManager::clear_buffer_trash()
...
+        size_t index = it - m_buffers.begin();
+        if (contains(indices, index))
+            throw runtime_error{format("buffer '{}' appears more than once", name)};
+        indices.push_back(index);
+    }
+
+    BufferList res;
+    for (size_t index : indices)
+        res.push_back(std::move(m_buffers[index]));
+    for (auto& buf : m_buffers)
```
lenormf: Is it safe to iterate over a `m_buffers` when several items in it have been `std::move`d?

src/buffer_manager.cc
```diff
@@ -105,4 +105,29 @@ void BufferManager::clear_buffer_trash()
...
 
+void BufferManager::arrange_buffers(ConstArrayView<String> first_ones)
+{
+    Vector<size_t> indices;
+    for (const auto& name : first_ones)
+    {
+        auto it = find_if(m_buffers, [&](auto& buf) { return buf->name() == name or buf->display_name() == name; });
+        if (it == m_buffers.end())
+            throw runtime_error{format("no such buffer '{}'", name)};
+        size_t index = it - m_buffers.begin();
```
lenormf: `m_buffers` is a vector, why not just move items at its beginning directly instead of storing indexes?

src/buffer_manager.cc
occivink: Yes, the `if (buf)` check will simply evaluate to false.

src/buffer_manager.cc
occivink: Because at this point we can still throw, so we don't want to manipulate `m_buffers` just yet

src/buffer_manager.cc
occivink: Yes that should be sufficient, and `buflist` contains the display names, but on the other hand it's not like there is any risk of ambiguity when comparing the full name either.

src/buffer_manager.cc
lenormf: According to [the documentation](https://en.cppreference.com/w/cpp/utility/move), all `m_buffers[index]` that have been moved "are placed in a valid but unspecified state".

That sounds to me like the `buf` iterator will be valid, but `if (buf)` may or may not be specified (compiler-dependent).

src/buffer_manager.cc
lenormf: Fair enough, note that you should use `std::distance` to calculate the amount of items between two iterators.

src/buffer_manager.cc
occivink: You want to look at the documentation of unique_ptr in this case.

src/buffer_manager.cc
occivink: I don't see a good reason why, and this pattern is used in several places in the codebase so I will keep it as is.

src/buffer_manager.cc
lenormf: Where exactly?

occivink: here: https://en.cppreference.com/w/cpp/memory/unique_ptr/unique_ptr, the overload #5. 
More specifically: 
>Constructs a unique_ptr by transferring ownership from u to *this and stores the null pointer in u.

lenormf: Makes sense, thanks!

lenormf:
Shouldn't `:buffer-{prev,next}` be impacted by this as well? If users sort their buffers alphabetically, maybe they expect that logic to apply to cycling through them as well.

occivink:
Yes, they are. Since both commands rely on the order in the `BufferManager`, this happens automatically.

PR #3385, opened 2020-02-26
Make `on_next_key_with_autoinfo()` respect `idle_timeout`

### LINKED ISSUE #2066: Configurable delay for info prompt
The info prompt is profoundly useful, but when you know exactly what you are trying to do -- it can just be a bit of a flashing annoyance.  I tried to capture a good example using https://asciinema.org but the window was open for so little time it doesn't appear to have been recorded at all!

Basically, when I know I am going to hit like ]p for example -- and I do it quickly, would be great if the info prompt didn't try to display for $x milliseconds.  Possibly a configurable delay?

### LINKED ISSUE #3365: Feature request: Show clippy after a delay
When quickly using a command that shows Clippy (eg. `[p`), Clippy just quickly flickers in the corner of my screen. This is distracting

If we add a small delay before showing Clippy (like 0.5s), then it reduces flickering, but Clippy can still give me hints when I don't know what key I should press next

6112:
The prompt and autocomplete normally wait for `idle_timeout` before showing
suggestions, however commands like `g`, `v`, or the lead-key show Clippy
instantly.

This patch fixes the issue by making `on_next_key_with_autoinfo()` wait for
`idle_timeout` before displaying suggestions.

Fixes #3365
Fixes #2066


danr:
Thanks for looking into this.

I don't fully agree that it fixes #3365, ideally it shouldn't have to be the same delay as idle_timeout. In #3365 you suggest a 0.5s delay for clippy to show up, this is 10x my idle_timeout :upside_down_face:

6112:
Alternatively, we could add a new option (say, `next_key_idle_timeout`) and have its default slightly higher than `idle_timeout`. (maybe 250ms?)

src/input_handler.hh
```diff
@@ -171,20 +173,24 @@ constexpr auto enum_desc(Meta::Type<AutoComplete>)
...
     context.input_handler().on_next_key(mode_name,
-        keymap_mode, [hide,cmd](Key key, Context& context) mutable {
+        keymap_mode, [cmd](Key key, Context& context) mutable {
+            bool hide = should_show_info(AutoInfo::OnKey, context);
             hide_auto_info_ifn(context, hide);
             cmd(key, context);
-    });
+    }, std::bind([&context](Timer&, String title, String info) {
+           show_auto_info_ifn(title, info, AutoInfo::OnKey, context);
+    }, _1, std::move(title), std::move(info)));
```
mawww: std::bind on a lambda seems weird, why not capture title and info ?
```suggestion
    }, [&context, title=std::move(title),info=std::move(info)](Timer&) {
           show_auto_info_ifn(title, info, AutoInfo::OnKey, context);
    });
```

src/main.cc
```diff
@@ -494,6 +494,8 @@ void register_options()
                        WriteMethod::Overwrite);
     reg.declare_option<int, check_timeout>(
         "idle_timeout", "timeout, in milliseconds, before idle hooks are triggered", 50);
+    reg.declare_option<int, check_timeout>(
+        "next_key_idle_timeout", "timeout, in milliseconds, before next-key suggestions are displayed", 250);
```
mawww: I am not sure it makes sense to make the next_key timeout special, we have auto-info displaying for prompt commands on idle timeout as well, I am not sure why we would have a separate timeout for next-key.

src/input_handler.hh
6112: Done

src/main.cc
6112: Done

6112:
Adressed your comments on the PR, mawww. Can you take a second look?

Delapouite:
The new behavior is interesting with a long timeout, like 2 secs, pressing `v` displays immediately `mode view` and 2 seconds later the regular view autoinfo. Same for `g` which immediately displays `go to location` and the 2 seconds later the goto menu.

mawww:
Is that a good "interesting" or a bad one ?

maximbaz:
Hey guys, first of all, thanks for this PR! I have tried it out and I must agree with @danr, it doesn't feel like it solves #3365... I want to have 1-2s delay before clippy shows up when I press the first key in a series (e.g. `<a-i>`) and a second key (e.g. `"`), simply because I don't need to see the popup with all available options if I know what I'm going to press next (clippy flashing for a fraction of a second is simply annoying in this case, but additionally because of my color scheme it makes my screen flash with white color, which is particularly annoying in a dark environment like at night), but if I set `idle_timeout` to a value of 1-2s, it makes everything else in kakoune unbearably slow, from autocomplete to highlighting search results...

Can we please consider @6112's suggestion to have a separate config property just for clippy? @mawww I saw your comment about a proposal of `next_key_idle_timeout`, I'm not sure if the scenario I described above is enough of a reason to treat "next_key" specially, or maybe we should think of a more global way to have even more dedicated idle timeout settings?

PR #3390, opened 2020-02-29
Fix unnecessary lint refreshes

eraserhd:
Incrementally setting the lint variables triggers multiple refreshes,
including the text jumping as the guttter column is removed and re-
added.  This causes the info message to disappear when linting is done
on NormalIdle.


eraserhd:
@Screwtapello ^^^

PR #3391, opened 2020-02-29
Add a -background flag to the edit command

CharlesGueunet:
Hello,

This small PR add a `-background` flag to the edit command.
This flag allows to open a buffer in the background, without changing the current one.

My first use case for it is a helper similar to the `make` command. In most of the case I am not interested in the content of the fifo except when the build fails, so I prefer to create this buffer in the background.

Best,
Charles


occivink:
Hello,
You can use `eval -draft %{ edit file }`, and similarly `eval -draft make`

CharlesGueunet:
Ho, you are right ! Thanks

PR #3397, opened 2020-03-07
src: Store the version number in a persistent file

lenormf:
Since Git isn't a build-dependency, it's possible for users to grab
a tarball of the latest release and compile a binary that will have
"unknown" for version number. Package maintainers who ship binaries in
their respective distributions are prone to falling victim to this bug.

This commit stores the version number of the latest release in the
`.version` file, which will be read if no Git metadata is available
in the source tree (i.e. if there's no way to get the current version
number on-the-fly with `git describe --tags`).

This also allows using a common idiom whenever a new release is ready
for publishing: the creation of a tagged meta-commit that only updates
whatever file containing the version number in the repository.


Screwtapello:
I've been bitten by this so many times on so many projects. I wish there was a way to make git auto-expand a string like `$Version:$` to `$(git describe --tags --dirty --always)` on checkout/archive and automatically undo that expansion on checkin.

lenormf:
NOTE: I looked into this after I installed [the Kakoune package on Alpine](https://pkgs.alpinelinux.org/package/edge/testing/x86_64/kakoune) and noticed the version number was "unknown".

occivink:
It doesn't seem ideal to use a hardcoded version number. Currently if you build from git you'll get an exact commit, but now  `v2020.01.16` could mean a build from any commit starting with this version up to the next.

lenormf:
I considered a few cases:

| # | downloaded version | version number (currently) | version number (thanks to the PR) |
|---|---|---|---|
|1| `git clone` the repository | accurate | accurate |
|2| download a ZIP of a release | "unknown" | accurate |
|3| download a ZIP based on an arbitrary commit | "unknown" | latest release number ("branch") |

The second row is what I think happened with [the Kakoune package on Alpine](https://pkgs.alpinelinux.org/package/edge/testing/x86_64/kakoune), and the reason why I made the PR.

The third row is a rare case, I reckon, and wanted to turn the version number into `vYYYY.MM.DD+` (trailing `+`), as that's a common idiom in software and is fairly recognizable. But I couldn't find a way to get the Makefile to recognize that the current source tree is not entirely based off a release when there's no Git metadata, so I settled for the "branch" version number. It could probably be done with some Git hooks, but it doesn't seem worth it.

mawww:
Not sure about that, `make dist` generates tarball with accurate version information from any git repo and I dont think dowloading zip files from github is a workflow we want to encourage.

lenormf:
Is there any way you could remove the "Source Code" links from the release announcements, so that the `kakoune-VERSION.tar.bz2` package (which contains a `.version` file) would be the only download-able asset?

I think users who just want to compile a binary expect the version number to be stored in the source code no matter how it's being downloaded. And they won't use Git because they don't need the history, only the files as they are for a given version, which Github makes easy to get thanks to the "Download as ZIP/tar.gz" button.

If the official releases only listed the "dist"-ributable archive, we could at least tell package maintainers never to use anything other than what's provided in the announcements.

mawww:
I dont know why package managers would use those zip files. They surely are used to get all kinds of troubles with that (for example all the autoconf based projects that do not version the configure script, and generate it from configure.ac only for releases).

lenormf:
I reckon because they don't expect things such as a version number not to be hardcoded in the source code, as is done commonly in other software (here is [NeoVim](https://github.com/neovim/neovim/blob/v0.4.3/CMakeLists.txt#L113,L118), [Linux](https://github.com/torvalds/linux/blob/master/Makefile#L2,L5)).

I'm not excited at the idea of redundantly maintaining a file that stores meta-data that belongs in the Git tree, but if we're not doing that then it would be nice to have *some way* to let people know that the distributable archives are the best way of grabbing the code.

mawww:
The version number is already hard coded and persistent in releases, I am pretty sure neither neovim nor linux have a file that gets updated on every git commit to store the generated version number there.

What would be nice is to tell github to stop showing zip files on the release page and only provide the tar.gz file.

lenormf:
>The version number is already hard coded and persistent in releases, I am pretty sure neither neovim nor linux have a file that gets updated on every git commit to store the generated version number there.

Yes, that's also what this PR does as well.

>What would be nice is to tell github to stop showing zip files on the release page and only provide the tar.gz file.

That's exactly what I asked [2 messages ago](https://github.com/mawww/kakoune/pull/3397#issuecomment-598042533), does that mean Github doesn't allow you to do that?

lenormf:
Looks like the Ubuntu Disco package fell into the same trap c.f. #3410

PR #3400, opened 2020-03-10
rc spell: Re-implement message processing in Awk

### LINKED ISSUE #3399: spell check: delays with tex files.
### Steps 

open a `tex` file with a few hundred lines.

Invoke `spell en` 

### Outcome 

Nothing happens or There's usually an enormous delay which can last up to a minute and a half before it kicks in.

On longer lines this behavior worsens. 

### Expected 
 
To highlight spelling errors as expected within a reasonable timeframe.

Tested with v2020.01.16 and also with https://github.com/mawww/kakoune/commit/6766297623e9718db5b2b09bd655fe183acdcf67

lenormf:
Plain shell takes too long on large files.

Fixes #3399

### DIFF
--- a/rc/tools/spell.kak
+++ b/rc/tools/spell.kak
@@ -37,33 +37,48 @@ define-command -params ..1 -docstring %{
         fi
 
         {
-            sed 's/^/^/' "$kak_opt_spell_tmp_file" | eval "aspell --byte-offsets -a $options" 2>&1 | {
-                line_num=1
-                regions=$kak_timestamp
-                while read -r line; do
-                    case "$line" in
-                        @\(\#\)*)
-                            # drop the identification message
-                        ;;
-                        [\#\&]*)
-                            if expr "$line" : '^&' >/dev/null; then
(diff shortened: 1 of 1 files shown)


lenormf:
As mentioned in #3399, spell-checking [this 44kB file](http://mirrors.ctan.org/macros/plain/base/plain.tex) takes ~25s on my computer with the current implementation, and ~3s with this PR.

Delapouite:
Oh! The difference is really significant. I had not idea of such a perf gap between shell and awk. Thanks for investigating this lenormf.

codesoap:
The spell command is now broken on OpenBSD. The problem seems to be the `/* */` style comments. (POSIX)[https://pubs.opengroup.org/onlinepubs/009695399/utilities/awk.html] says the following about comments:

> A comment shall consist of any characters beginning with the number sign character and terminated by, but excluding the next occurrence of, a \<newline\>.

Changing the two comments to that style seems to fix it for the OpenBSD awk and also works with GNU awk.

Should I make a new PR, or can this be fixed on this PR (honest question; I don't know how GitHub works here)?

lenormf:
Please make a new PR, this one was merged and Github considers it closed.

PR #3406, opened 2020-03-12
src: Allow shell commands on fast completion

lenormf:
Interactive user commands with parameters that can be completed upon
using shell commands won't auto-complete, because shell commands are
disabled in prompt mode.

This commit removes that restriction to allow the above usecase.

Example:

    def foo \
        -shell-script-completion %{
            if [ "${kak_token_to_complete}" = 0 ]; then
                list_options | awk -v prefix="$1" 'tolower(substr($0, 1, length(prefix))) == tolower(prefix)'
            fi
        } […]

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -216,9 +216,6 @@ struct ShellScriptCompleter
                            CommandParameters params, size_t token_to_complete,
                            ByteCount pos_in_token)
     {
-        if (flags & CompletionFlags::Fast) // no shell on fast completion
-            return Completions{};
-
         ShellContext shell_context{
             params,
             { { "token_to_complete", to_string(token_to_complete) },


mawww:
The fast completion flag was put in place explicitely to avoid running shells in that case, not to lead to a sluggish interface where every keystroke triggers a shell completer (that can be arbitrary slow).

I dont think disabling it is a good idea, but it could make sense for a shell completion to be tagged as fast.

In general we have moved away from shell-script-completions to shell-script-candidates that runs only once per argument and allow kakoune fuzzy matching to run.

lenormf:
The issue is that "fast completion" means "no completion at all", I had to read the code to find out there was nothing wrong with my command, the editor just wouldn't call the completion script.

We could document that using this flag entails running a shell on each keypress and is obviously very slow, but disabling it in prompt-mode pre-emptively prevents users who are ready to take that trade-off from using it.

In the end I can use `-shell-script-candidates` for my specific use-case, so I don't mind if the PR gets rejected, but without my patch, `-shell-script-completion` looks like dead code.

Issue #3410, opened 2020-03-16
Compile form source: `make` didn't worked


joeitu:
### Steps 
Install kakoune from source on ubuntu 19.04 disco 
After installing kakoune from package manager, I realised kakoune was not up to date ( see https://github.com/danr/kakoune-easymotion/issues/23) 
When I run `kakoune -version` I get `Kakoune unknown`
So I tried to install it from source, following the ubuntu tutorial on the github README

### Outcome 
After updating, upgrading and installing the libraries, I run `make` in the `src` directory and got the following errors and warnings : 
```
 src/ git:(master) make
g++  -O3 -pedantic -std=c++17 -g -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare -Wno-address -Wno-init-list-lifetime -MD -MP -MF .main.opt.d -c -o .main.opt.o main.cc
In file included from /usr/include/c++/8/vector:63,
                 from vector.hh:6,
                 from file.hh:9,
                 from buffer.hh:8,
                 from main.cc:3:
/usr/include/c++/8/bits/stl_uninitialized.h: In instantiation of ‘_ForwardIterator std::__uninitialized_copy_a(_InputIterator, _InputIterator, _ForwardIterator, _Allocator&) [with _InputIterator = Kakoune::TransformView<const Kakoune::SelectionList&, Kakoune::<lambda(Kakoune::StringView, const Kakoune::Context&)>::<lambda(const Kakoune::Selection&)> >::Iterator; _ForwardIterator = Kakoune::String*; _Allocator = Kakoune::Allocator<Kakoune::String, (Kakoune::MemoryDomain)0>]’:
/usr/include/c++/8/bits/stl_vector.h:1469:33:   required from ‘void std::vector<_Tp, _Alloc>::_M_range_initialize(_ForwardIterator, _ForwardIterator, std::forward_iterator_tag) [with _ForwardIterator = Kakoune::TransformView<const Kakoune::SelectionList&, Kakoune::<lambda(Kakoune::StringView, const Kakoune::Context&)>::<lambda(const Kakoune::Selection&)> >::Iterator; _Tp = Kakoune::String; _Alloc = Kakoune::Allocator<Kakoune::String, (Kakoune::MemoryDomain)0>]’
[...]

PR #3411, opened 2020-03-16
Disambiguation: specify to build from release

joeitu:
User could build it after doing a git clone, but this seems to be buggy ( see #3410 ) 
So its more stable to build it from release.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -109,7 +109,8 @@ Kakoune dependencies are:
    associated {cpp} standard library (libstdc{pp} or libc{pp})
  * ncurses with wide-characters support (>= 5.3, generally referred to as libncursesw)
 
-To build, just type *make* in the src directory.
+
+To build, download the latest release from link:https://github.com/mawww/kakoune/releases[here], then just type *make* in the src directory.
 To generate man pages, type *make man* in the src directory.
 
 Kakoune can be built on Linux, MacOS, and Cygwin. Due to Kakoune relying heavily


joeitu:
Related to #3410

Screwtapello:
"Releases" are mostly an excuse to ask Linux distros to update their packages, most people who want to build from source should probably build the latest version.

Unfortunately, while the current version of Kakoune builds properly with GCC 7 and GCC 9, it seems GCC 8 has a bug, and the current stable releases of Ubuntu and Debian default to that version.

Working around the bug is relatively easy, at least right now:

    sudo apt install g++-7
    cd path/to/Kakoune
    make -C src CXX=g++-7

joeitu:
Could be nice to put it in the doc, those kind of installation problems can be pretty painfull. Should I edit the README ?

joeitu:
> "Releases" are mostly an excuse to ask Linux distros to update their packages, most people who want to build from source should probably build the latest version.

It's weird because I have installed kak through my package manager ( apt,  ubuntu disco )  and I got an outdated  version( older than the latest release, the `-to-file` was not implemented - see https://github.com/danr/kakoune-easymotion/issues/23 ) 

Unfortunatly I couldn't see which version apt had installed ( see https://github.com/mawww/kakoune/pull/3397 )

Screwtapello:
Yeah, the person who packages Kakoune for Debian (and therefore Ubuntu) does pretty well, but doesn't always get around to updating the package for every Kakoune release. You can ask the package manager about version numbers like this:

```
$ apt policy kakoune
kakoune:
  Installed: (none)
  Candidate: 2019.07.01-1
  Version table:
     2019.07.01-1 990
        990 http://httpredir.debian.org/debian testing/main amd64 Packages
        500 http://httpredir.debian.org/debian unstable/main amd64 Packages
```

Here "installed" is "(none)" because I build Kakoune from source rather than installing it with apt, but if I *did* have it installed, that's where the version number would be.

"candidate" describes which version apt will try to install at the next `apt upgrade`.

"version table" lists all the versions apt knows about, and where they come from. In this case, the same version is available from both `testing/main` and `unstable/main` on the `httpredir.debian.org` mirror.

Alternatively, you can go to https://packages.ubuntu.com and [search for Kakoune](https://packages.ubuntu.com/search?keywords=kakoune&searchon=names&suite=all&section=all) to see the version available in each release.

lenormf:
The bug described by #3410 is that `g++` v8 doesn't want to compile the project - not that you have downloaded the wrong code.

Use another version of `g++` or `clang++` in the meantime, as mentioned in the related ticket.

lenormf:
C.f. #3413 re. #3410.

mawww:
Users that are going to build Kakoune themselves are encouraged to build latest master, which should be stable.

PR #3417, opened 2020-03-19
Better rust auto-indent for comments.

jneem:
Now whitespace will be matched when auto-indenting a comment (which
regressed in #3384).


pickfire:
This is outdated. I had already fixed it.

mawww:
Thanks for the propsed fix, closing as already fixed.

PR #3420, opened 2020-03-20
Fix flicker on window refresh

eraserhd:
It's clearly possible to be a lot more efficient, but this fix is simple and sufficient to stop my flicker issues.

The curword stuff that I stole from @mawww and modified causes multiple lines to be invalidated, and the terminal would refresh before the lines were in the right place.


eraserhd:
Ok, it still flickers a little -- but the flickers are less noticeable and less frequent.

eraserhd:
After I did this, I was working on pairing adds and deletes into changes and came up with an algorithm similar to your last commit on builtin-terminal-ui (well, except it required a little bit more bookkeeping).  Your commit works nicely for me - I haven't seen flicker yet!

PR #3422, opened 2020-03-21
Add syntax highlighting and indent hooks for zig

ifreund:
This commit adds syntax highlighting (and file type detection) for [zig](https://ziglang.org).  It is loosely based of off the official syntax highlighting for vim [here](https://github.com/ziglang/zig.vim/blob/master/syntax/zig.vim) as well as the [grammar specification](https://ziglang.org/documentation/master/#Grammar).

### DIFF
--- a/rc/filetype/zig.kak
+++ b/rc/filetype/zig.kak
@@ -111,9 +111,11 @@ define-command -hidden zig-trim-indent %{
 
 define-command -hidden zig-indent-on-new-line %<
     evaluate-commands -draft -itersel %<
-        # copy // or /// comments prefix and following white spaces
         try %<
+            # copy // or /// comments prefix and following whitespace
             execute-keys -draft k <a-x> s ^\h*\K///?\h* <ret> y gh j P
+            # preserve indent level
+            try %< execute-keys -draft <semicolon> K <a-&> >
         > catch %<
             # preserve indent level
             try %< execute-keys -draft <semicolon> K <a-&> >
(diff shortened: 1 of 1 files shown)


rc/filetype/zig.kak
```diff
@@ -0,0 +1,80 @@
...
+add-highlighter shared/zig/doc_comment      region '///[^/]' '$'       fill documentation
+add-highlighter shared/zig/comment          region '//' '$'            fill comment
+
+# TODO: highlight escape sequences within strings
+add-highlighter shared/zig/string_double    region '"' (?<!\\)(\\\\)*" fill string
+add-highlighter shared/zig/string_single    region "'" (?<!\\)(\\\\)*' fill string
+add-highlighter shared/zig/string_multiline region '\\\\' '$'          fill string
+
+# storage
+add-highlighter shared/zig/code/ regex '\b(?:const|var|extern|packed|export|pub|noalias|inline|noinline|comptime|callconv|volatile|allowzero|align|linksection|threadlocal)\b' 0:keyword
```
lenormf: There are a bunch of attributes here (e.g. `const`) that could be highlighted using the face of the same name. Maybe you could split that list?

ifreund: Hmm, I believe this entire list is attributes in fact. Good catch, I'll fix that.

ifreund:
Took some inspiration from the support for other languages and added some basic indent hooks as well.  Also put everything inside a module since that seems like the proper way to do things.

PR #3424, opened 2020-03-22
rc spell: Escape pipe symbol in parameter expansion

codesoap:
The `:spell-next` command did not work on OpenBSD. With this PR it seems to work on OpenBSD and Linux.


lenormf:
PING

PR #3425, opened 2020-03-22
Fix build on NetBSD

niacat:


src/file.cc
```diff
@@ -21,7 +21,7 @@
 #include <sys/select.h>
 #include <unistd.h>
 
-#if defined(__FreeBSD__)
+#if defined(__FreeBSD__) or defined(__NetBSD__)
```
lenormf: Does the pre-processor support that?

src/file.cc
```diff
@@ -624,7 +624,7 @@ String get_kak_binary_path()
     kak_assert(res != -1);
     buffer[res] = '\0';
     return buffer;
-#elif defined(__FreeBSD__)
+#elif defined(__FreeBSD__) or defined(__NetBSD__)
     int mib[] = {CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME, -1};
```
krytarowski, line 628: NetBSD muse use different arguments to mib[]

krytarowski:
Thanks. Now it would be nice to fix the `mib[]` arguments and pass `CTL_KERN, KERN_PROC_ARGS, -1, KERN_PROC_PATHNAME` for NetBSD.

mawww:
Do you know if that would be compatible with FreeBSD or if we need to split the logic between NetBSD and FreeBSD ?

krytarowski:
We need to ifdef each OS separately.

mawww:
I do not have NetBSD available for testing, could you create a pull request with such a change ?

krytarowski:
https://github.com/mawww/kakoune/pull/3453

PR #3427, opened 2020-03-22
diff.kak: add diff file and diff hunk text objects

eraserhd:
These really only support `<a-a>` mode, but I expect they are super
useful as-is.

### DIFF
--- a/rc/filetype/diff.kak
+++ b/rc/filetype/diff.kak
@@ -11,3 +11,36 @@ hook -group diff-highlight global WinSetOption filetype=diff %{
     add-highlighter window/diff ref diff
     hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/diff }
 }
+
+define-command \
+    -docstring %{diff-select-file: Select surrounding patch file} \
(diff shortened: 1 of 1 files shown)


rc/filetype/diff.kak
```diff
@@ -11,3 +11,45 @@ hook -group diff-highlight global WinSetOption filetype=diff %{
...
+            execute-keys '"ez'
+        } catch %{
+            execute-keys '"oz'
+            fail 'Not in a diff hunk'
+        }
+    }
+}
+
+hook -group diff-text-objects global WinSetOption filetype=diff %{
+    map window -docstring 'diff file' object f '<a-;> diff-select-file<ret>'
```
lenormf: I don't think we should declare mappings in the users' back, providing a command-based API seems good enough.

The docstrings could be better too.

eraserhd: I removed the mappings and slightly tweaked the docstrings to use the same language as the object menu (though I'm not sure it's better or what better would be).

PR #3433, opened 2020-03-27
rc lint: Check the `lintcmd` option is set in `:lint-buffer`

lenormf:
Calling `:lint-buffer` when `lintcmd` is empty results in a temporary
directory being created, but never removed when the underlying linting
code errors out.

### DIFF
--- a/rc/tools/lint.kak
+++ b/rc/tools/lint.kak
@@ -277,6 +277,12 @@ define-command \
     } \
     lint-buffer \
 %{
+    evaluate-commands %sh{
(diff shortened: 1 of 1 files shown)


Screwtapello:
I wonder why `:lint-buffer` calls `:lint-cleaned-selections` directly, instead of just calling `:lint-selections` and re-using that function's error-checking.

lenormf:
@Screwtapello could be to save one process-call with the shell loop that re-orders all the selections in `:lint-selections`.

PR #3438, opened 2020-03-30
rc doc: Implement a `doc-search` command

lenormf:
This commit implements a `doc-search` command that allows searching
for a pattern into all the documentation pages commonly available
through the `:doc` command.

Since the pages are in the Asciidoc format on disk and subsequently
processed for rendering at runtime, we cannot directly search
them. Therefore an intermediary, dummy session is spawned to render
the required pages, and search matches (along with line context)
shared with the "parent" session using a temporary file.

As a consequence, this functionality is slow as hell, and doesn't
provide with caching of the search results. It's possible however to
narrow down which pages will be searched, which substantially limits
the amount of processing.

### DIFF
--- a/rc/tools/doc.kak
+++ b/rc/tools/doc.kak
@@ -165,3 +165,100 @@ define-command -params 1..2 \
 }
 
 alias global help doc
+
+define-command -params 1.. -docstring %{
+        doc-search <pattern> [topic...]: search for a pattern in the documentation
+        Display a menu listing all the search matches in the given topics, or the whole documentation if unspecified
+    } \
+    -shell-script-candidates %{
+        case "${kak_token_to_complete}" in
+            0) ;;
+            *) find "${kak_runtime}/doc/" -type f -name '*\.asciidoc' | sed 's,.*/,,; s,\.[^/]*$,,';;
+        esac
+    } doc-search %{ evaluate-commands %sh{
+        readonly SELECTIONS_FILE=$(mktemp "${TMPDIR:-/tmp}"/kak-doc-search.XXXXXXXX)
+        readonly SEARCH_PATTERN="${1}"
+
+        shift
+        if [ $# -eq 0 ]; then
+            eval set -- $(ls "${kak_runtime}/doc"/*.asciidoc | sed 's,.*/,,; s,\.[^/]*$,,')
+        else
+            for topic; do
+                if [ ! -e "${kak_runtime}/doc/${topic}.asciidoc" ]; then
+                    printf 'fail No such topic %%{%s}' "${topic}"
+                    exit 1
+                fi
+            done
+        fi
+
+        # NOTE: don't rely on the PATH to make sure the `kak` instance we run implements doc-searching
+        env KAK_DOC_SEARCH_TMP="${SELECTIONS_FILE}" \
+            KAK_DOC_SEARCH="${SEARCH_PATTERN}" \
+            "${kak_runtime}"/../../bin/kak -ui dummy -e "doc-search-impl $*"
+
+        awk -F'\a' '
+            BEGIN {
+                Q = "\047"
+            }
+
+            {
+                context = $3
+                gsub(Q, "&&", context)
+                candidates = candidates " " Q "{MenuInfo}" $1 "{default}{\\} " context Q " " Q "doc " $1 "; select " $2 Q
+            }
+
+            END {
+                if (candidates)
+                    print "menu -markup" candidates
+            }
+        ' "${SELECTIONS_FILE}"
+
+        rm -f "${SELECTIONS_FILE}"
+} }
+
+define-command -hidden -params .. doc-search-impl %{
+    evaluate-commands %sh{
+        readonly SEARCH_PATTERN_ESC=$(printf %s "${KAK_DOC_SEARCH}" | sed "s/'/&&/g")
+
+        for topic; do
+            printf "doc-search-impl-helper %s '%s'\n" "${topic}" "${SEARCH_PATTERN_ESC}"
+        done
+    }
+
+    kill!
+}
+
+define-command -hidden -params 2 doc-search-impl-helper %{ try %{
+    doc %arg{1}
+
+    execute-keys \% s "%arg{2}" <ret> )
+
+    evaluate-commands -save-regs 'mn' %{
+        set-register m %val{selections_desc}
+        set-register n ''
+
+        # For each search match, grab the line on which it lies for context
+        evaluate-commands -draft -itersel %{
+            execute-keys <a-x>H
+            set-register n %reg{n} %val{selection}
+        }
+
+        nop %sh{
+            readonly TOPIC="${1}"
+
+            {
+                desc="${kak_reg_m}"
+
+                eval "set -- ${kak_quoted_reg_n}"
+                # XXX: the first item is an empty string
+                shift
+                for sel; do
+                    printf '%s\a%s\a%s\n' "${TOPIC}" "${desc%% *}" "${sel}"
+                    desc="${desc#* }"
+                done
+            } >> "${KAK_DOC_SEARCH_TMP}"
+        }
+    }
+
+    delete-buffer!
+} }


lenormf:
`:doc-search "you're"` shows a menu with contextual information:



Selecting an item takes us to the match in the appropriate documentation page:

Delapouite:
Thanks! Related work: https://github.com/jbomanson/search-doc.kak

rc/tools/doc.kak
```diff
@@ -165,3 +165,100 @@ define-command -params 1..2 \
...
+                    printf 'fail No such topic %%{%s}' "${topic}"
+                    exit 1
+                fi
+            done
+        fi
+
+        # NOTE: don't rely on the PATH to make sure the `kak` instance we run implements doc-searching
+        env KAK_DOC_SEARCH_TMP="${SELECTIONS_FILE}" \
+            KAK_DOC_SEARCH="${SEARCH_PATTERN}" \
+            "${kak_runtime}"/../../bin/kak -ui dummy -e "doc-search-impl $*"
```
lenormf: I might be able to optimize startup time here by adding `-n -E 'source …/doc.kak'`. Highlighters in the user configuration wouldn't make previously saved selection descriptions invalid, would they?

mawww:
How about changing `doc-render` to not create the buffer itself (basically moving doc.kak line 101 to line 160) so that it renders a page inside the current buffer, then just use `edit -scratch; doc-render <page>` to get a temporary scratch buffer containing the rendered page and extracting matches from there. This would avoid the complex dance that is done with forking a helper session and communicating with it.

lenormf:
It's a good idea. I implemented it, and it does cut the size of the code in half, but it's still so slooow.

It looks like `eval -draft -itersel` is partially to blame for that:

```
# The following two scopes perform the same task, differently:

        evaluate-commands -draft -itersel %{
            # Topic and selection description of the search match
            set-option -add global doc_search_matches %arg{1} %val{selection_desc}

            # Search match context
            execute-keys <a-x>H
            set-option -add global doc_search_matches %val{selection}
        }

        # The following takes ~30% less time on my system!
        evaluate-commands %sh{
            for selection_desc in ${kak_selections_desc}; do
                printf '
                    set-option -add global doc_search_matches %s %s
                    select %s
                    execute-keys <a-x>H
                    set-option -add global doc_search_matches %%val{selection}
                ' "$1" "${selection_desc}" "${selection_desc}"
            done
        }
```

Any idea what's going on?

lenormf:
I worked around the problem by limiting the amount of search results (c.f. reasons in the commit message).

On my humble system, it takes ~10s to get results back from a stupid command like `:doc-search .` (all characters, all topics).

rc/tools/doc.kak
```diff
@@ -154,14 +157,94 @@ define-command -params 1..2 \
...
+        esac
+    } doc-search %{
+    set-option global doc_search_matches
+
+    evaluate-commands %sh{
+        readonly SEARCH_PATTERN="${1}"
+
+        shift
+        if [ $# -eq 0 ]; then
+            eval set -- $(ls "${kak_runtime}/doc"/*.asciidoc | sed 's,.*/,,; s,\.[^/]*$,,')
```
krobelus, line 193: eval looks unnecessary here.

Also maybe use the same find command as for the completions above. But then it will be even slower.

rc/tools/doc.kak
```diff
@@ -154,14 +157,94 @@ define-command -params 1..2 \
...
+            ' "${candidates}"
+        fi
+    }
+}
+
+define-command -hidden -params 2 doc-search-impl %{
+    edit -scratch
+
+    try %{
+        doc-render "%val{runtime}/doc/%arg{1}.asciidoc"
```
krobelus, line 232: It really is a bit slow. I wouldn't mind using a version that does not render the asciidoc. The rendering shouldn't make a difference for most searches' results.

Maybe just render the results of grep -R doc/. Not sure if that would be robust enough.

rc/tools/doc.kak
lenormf: Correct. I'd just copy/pasted the command used by `:doc`, then modified into *this*.

I've simplified and unified how the topics are listed in both commands, in a new commit.

rc/tools/doc.kak
lenormf: We need to render the markup because we don't want to search through text that is part of the markup language, e.g. link.

Some markup is also replaced in the output, which invalidates coordinates if we want to jump to a specific location after a call to `:doc`.

mawww:
I think the performance issue is a common one we get with *SetOption hooks, each time you touch your option, we trigger those hooks for all buffer for all windows, and in particular that means we have to generate the filtering test `<option name>=<value>` with value getting bigger and bigger.

I explored remove the =<value> part of that filter, but unfortunately we rely on the pretty extensively for filetype hooks (and seems its exclusively for that).

Another possible approach is to say that `=<value>` is only there for options of primitive type (int, bool, str) and for other options we just put `<option name>`. It adds a bit of inconsistency, but at the same time it seems very unlikely we would filter on value for any non primitive option types.

Any opinion on that ?

lenormf:
Shouldn't this be handled with `:eval -no-hooks`, then? It doesn't seem to change anything when I use the flag in the code.

mawww:
Can you try with this diff:

```
diff --git a/src/buffer.cc b/src/buffer.cc
index d84586cd..ddbad72c 100644
--- a/src/buffer.cc
+++ b/src/buffer.cc
@@ -685,7 +685,7 @@ void Buffer::on_option_changed(const Option& option)
             m_flags &= ~Flags::ReadOnly;
     }
     run_hook_in_own_context(Hook::BufSetOption,
-                            format("{}={}", option.name(), option.get_as_string(Quoting::Raw)));
+                            format("{}={}", option.name(), option.get_desc_string()));
 }
 
 void Buffer::run_hook_in_own_context(Hook hook, StringView param, String client_name)
diff --git a/src/option_manager.hh b/src/option_manager.hh
index 81f3b1b9..739ec6bd 100644
--- a/src/option_manager.hh
+++ b/src/option_manager.hh
@@ -55,6 +55,7 @@ public:
 
     virtual String get_as_string(Quoting quoting) const = 0;
     virtual Vector<String> get_as_strings() const = 0;
+    virtual String get_desc_string() const = 0;
     virtual void set_from_strings(ConstArrayView<String> strs) = 0;
     virtual void add_from_strings(ConstArrayView<String> strs) = 0;
     virtual void update(const Context& context) = 0;
@@ -154,6 +155,14 @@ public:
         return option_to_string(m_value, quoting);
     }
 
+    String get_desc_string() const override
+    {
+        if constexpr (std::is_same_v<int, T> or std::is_same_v<bool, T> or std::is_same_v<String, T>)
+            return option_to_string(m_value, Quoting::Raw);
+        else
+            return "...";
+    }
+
     void set_from_strings(ConstArrayView<String> strs) override
     {
         set(option_from_strings(Meta::Type<T>{}, strs));
diff --git a/src/scope.cc b/src/scope.cc
index 79da998c..64bcd445 100644
--- a/src/scope.cc
+++ b/src/scope.cc
@@ -19,7 +19,7 @@ void GlobalScope::on_option_changed(const Option& option)
 {
     Context empty_context{Context::EmptyContextFlag{}};
     hooks().run_hook(Hook::GlobalSetOption,
-                     format("{}={}", option.name(), option.get_as_string(Quoting::Raw)),
+                     format("{}={}", option.name(), option.get_desc_string()),
                      empty_context);
 }
 
diff --git a/src/window.cc b/src/window.cc
index 8acc7a03..ab885ca9 100644
--- a/src/window.cc
+++ b/src/window.cc
@@ -346,8 +346,7 @@ void Window::clear_display_buffer()
 
 void Window::on_option_changed(const Option& option)
 {
-    run_hook_in_own_context(Hook::WinSetOption, format("{}={}", option.name(),
-                                                   option.get_as_string(Quoting::Raw)));
+    run_hook_in_own_context(Hook::WinSetOption, format("{}={}", option.name(), option.get_desc_string()));
     // an highlighter might depend on the option, so we need to redraw
     force_redraw();
 }
```

`-no-hooks` would not alleviate the problem that much if its what I think it is because the filter string generation is done regardless (we have -always hooks that might still run).

lenormf:
It'd be nice to have that command included in the next release!

lenormf:
Rebase on master. Added support for other supported directories that may contain documentation. Reported #4273 in the process. It's still slow but usable.

PR #3440, opened 2020-04-03
Allow zero-length ranges for replace-ranges

tadeokondrak:
https://github.com/mawww/kakoune/issues/2536#issuecomment-433797851

Enables inlay hints like:

### DIFF
--- a/src/highlighters.cc
+++ b/src/highlighters.cc
@@ -1448,14 +1448,17 @@ private:
     String m_default_face;
 };
 
-String option_to_string(InclusiveBufferRange range)
+String option_to_string(BufferRange range)
 {
-    return format("{}.{},{}.{}",
-                  range.first.line+1, range.first.column+1,
-                  range.last.line+1, range.last.column+1);
+    if (range.begin == range.end)
+        return format("{}.{}+0", range.begin.line+1, range.begin.column+1);
+    else
+        return format("{}.{},{}.{}",
+                  range.begin.line+1, range.begin.column+1,
+                  range.end.line+1, range.end.column);
 }
 
-InclusiveBufferRange option_from_string(Meta::Type<InclusiveBufferRange>, StringView str)
+BufferRange option_from_string(Meta::Type<BufferRange>, StringView str)
 {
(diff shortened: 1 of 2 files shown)


tadeokondrak:
Hmm, this currently doesn't work for replacing the newline (needed for [chaining hints](https://rust-analyzer.github.io/thisweek/2020/03/30/changelog-18.html)), and I'm not sure how to fix that.

mawww:
f7a2ecfacb566a219d9deef3a5cfaa2d1c5bd404 implemented that, I tried going with a BufferRange initially, but it would break with multi-byte characters as setting a range on a previous timestamp would not allow us to know where the end of the last character is, which could lead to BufferRange.end to point inside a utf-8 sequence.

Seems my implementation does not allow adding ranges at the end of lines, nor at the begining of them.

Adding at the end of line is tricky because in Kakoune models there is no difference between 'after last character on a line' and 'before first chracter on next line'. Maybe it would be better to bite the bullet and have a specific highlighter to add annotations at certain characters, which would allow to specify if annotations should be after or before their anchoring character.

tadeokondrak:
> f7a2ecf implemented that, I tried going with a BufferRange initially, but it would break with multi-byte characters as setting a range on a previous timestamp would not allow us to know where the end of the last character is, which could lead to BufferRange.end to point inside a utf-8 sequence.

Ah. Unfortunately it seems like in that one, if the replacement falls on a space, like in the `opts: Options = ` (where the replacement is adding `: Options`), it doesn't show up.

> Seems my implementation does not allow adding ranges at the end of lines, nor at the begining of them.
> Adding at the end of line is tricky because in Kakoune models there is no difference between 'after last character on a line' and 'before first chracter on next line'. Maybe it would be better to bite the bullet and have a specific highlighter to add annotations at certain characters, which would allow to specify if annotations should be after or before their anchoring character.

I think this would be useful for more than just type hints, an issue I have with the diagnostics feedback is that it's usually more effort to show the diagnostic than to figure out what the error is, with this there could be an option in kak-lsp to show them inline, like I had setup in Neovim before:

tadeokondrak:
Works now 🎉

PR #3446, opened 2020-04-15
rc spell: Add spell-add command

JJK96:
Adds the selected word to the dictionary

### DIFF
--- a/rc/tools/spell.kak
+++ b/rc/tools/spell.kak
@@ -148,3 +148,21 @@ define-command \
     ')
     printf 'try %%{ menu -auto-single %s }' "${menu}"
 } }
+
+define-command -params 0.. \
+    -docstring "Add the current selection to the dictionary" \
+    spell-add %{ nop %sh{
+    if [ -n "$kak_opt_spell_last_lang" ]; then
+        options="-l '$kak_opt_spell_last_lang'"
+    fi
+    if [ $# -eq 0 ]; then
+        # use selections
+        eval set -- "$kak_quoted_selections"
+    fi
+    words=""
+    while [ $# -gt 0 ]; do
+        words="$words"$(printf '*%s\\n#\\n' "$1")
+        shift
+    done
+    printf %b\\n "$words" | eval "aspell -a $options" 
+}}
--- a/rc/tools/spell.kak
+++ b/rc/tools/spell.kak
@@ -151,7 +151,7 @@ define-command \
 
 define-command -params 0.. \
     -docstring "Add the current selection to the dictionary" \
-    spell-add %{ nop %sh{
+    spell-add %{ evaluate-commands %sh{
     if [ -n "$kak_opt_spell_last_lang" ]; then
         options="-l '$kak_opt_spell_last_lang'"
     fi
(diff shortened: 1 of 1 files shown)


rc/tools/spell.kak
```diff
@@ -148,3 +148,12 @@ define-command \
...
     printf 'try %%{ menu -auto-single %s }' "${menu}"
 } }
+
+define-command \
+    -docstring "Add the current selection to the dictionary" \
+    spell-add %{ nop %sh{
+    if [ -n "$kak_opt_spell_last_lang" ]; then
+        options="-l '$kak_opt_spell_last_lang'"
+    fi
+    echo "*$kak_selection\n#" | eval "aspell -a $options"
```
lenormf, line 158: Please use `printf` with the appropriate format here.

rc/tools/spell.kak
```diff
@@ -148,3 +148,12 @@ define-command \
     ')
     printf 'try %%{ menu -auto-single %s }' "${menu}"
 } }
+
+define-command \
+    -docstring "Add the current selection to the dictionary" \
+    spell-add %{ nop %sh{
```
lenormf, line 154: Any way you could make the function 1) use the word passed as argument, if specified 2) or otherwise add all the words in the current multiple selection?

rc/tools/spell.kak
```diff
@@ -148,3 +148,21 @@ define-command \
...
+    if [ -n "$kak_opt_spell_last_lang" ]; then
+        options="-l '$kak_opt_spell_last_lang'"
+    fi
+    if [ $# -eq 0 ]; then
+        # use selections
+        eval set -- "$kak_quoted_selections"
+    fi
+    words=""
+    while [ $# -gt 0 ]; do
+        words="$words$(printf '*%s\\n#\\n' $1)"
```
lenormf, line 164: Shouldn't the format be `'*%s\n#\n'`?

You should probably also quote `$1` as `"$1"`.

rc/tools/spell.kak
```diff
@@ -148,3 +148,21 @@ define-command \
...
+        shift
+    done
+    printf %b\\n "$words" | tee /tmp/words | eval "aspell -a $options"
```
lenormf, line 167: Why is there `tee /tmp/words` here, was that for debugging?

How about printing an error if the command fails?

JJK96: Oh oops, that was for debugging indeed

rc/tools/spell.kak
JJK96: I had to do the format string like this, otherwise the newline at the end would be removed.
If you can get it to work with that format string, be my guest

JJK96: Added the quoting

lenormf: Right, I appreciate the effort, but I don't expect performance to be much of a problem here. Maybe something like the following would make things easier (untested)?

```
    words=""
    while [ $# -gt 0 ]; do
        words="$words"$(printf '*%s\\n#\\n' "$1")
        shift
    done
    printf %b\\n "$words" | eval "aspell -a $options" 
```

↓

```
    for word; do
        if ! printf '*%s\n#\n' "${word}" | eval "aspell -a $options"; then
           printf 'fail "Unable to add word: %s"' "$(printf %s "${word}" | sed 's/"/&&/g')"
           exit 1
        fi
    done
```

JJK96: That works fine, thanks for the suggestion

JJK96:
Tell me if you want to merge this, I will squash the commits.

PR #3448, opened 2020-04-16
Split modules and allow kitty-send-text to take a parameter

eraserhd:


rc/windowing/repl/kitty.kak
```diff
@@ -0,0 +1,39 @@
...
+            cmd="$*"
+        fi
+        kitty @ new-window --no-response --window-type $kak_opt_kitty_window_type --title kak_repl_window --cwd "$PWD" $cmd < /dev/null > /dev/null 2>&1 &
+    }
+}
+
+define-command -hidden -params 0..1 \
+    -docstring %{kitty-send-text [text]: Send text to the REPL window.
+
+If no text is passed, the selection is used.} \
```
lenormf, line 24: Could you please prettify this docstring, as well as the one for `:kitty-repl`? Something like:

```
-docstring %{
    kitty-send-text [text]: Send text to the REPL window.

    If no text is passed, the selection is used.
} \
```

eraserhd: done.

PR #3457, opened 2020-04-28
Implement smarter detection of windowing environments.

sjjf:
This fixes an issue I was having where the hooks to load *both* the x11 and kitty modules were matching, and the x11 module was being loaded on top of the kitty module - this meant that the terminal commands pointed at the x11 variants rather than the kitty variants. Merging the hooks provided by each of the windowing modules into a single hook that only loads one module avoids this issue.

Tested with kitty, gnome-terminal (x11 module), tmux and screen, all on Ubuntu Bionic.

### DIFF
--- a/rc/windowing/detection.kak
+++ b/rc/windowing/detection.kak
@@ -1,0 +1,41 @@
+# Attempt to detect the windowing environment we're operating in
+#
+# We try to load modules from the windowing_modules str-list option in order,
+# stopping when one of the modules loads successfully. This ensures that only
+# a single module is loaded by default.
+#
+# On load each module must attempt to detect the environment it's appropriate
+# for, and if the environment isn't appropriate it must fail with an error.
+# In addition, each module must check for the length of the windowing_modules
+# str-list option defined below, and must /not/ check for an appropriate
+# environment if the list is empty. An example of this test:
+#
+# evaluate-commands %sh{
+#     [ -z "${kak_opt_windowing_modules}" ] || [ -n "$TMUX" ] || echo 'fail tmux not detected'
+# }
+#
+# Each module is expected to define at least two aliases:
+#  * terminal - create a new terminal with sensible defaults
+#  * focus - focus the specified client, defaulting to the current client
+#
+
+declare-option -docstring \
+"Ordered list of windowing modules to try and load. An empty list disables
+both automatic module loading and environment detection, enabling complete
+manual control of the module loading." \
+str-list windowing_modules 'tmux' 'screen' 'kitty' 'iterm' 'x11'
+
+hook -group windowing global KakBegin .* %{
+
+    evaluate-commands %sh{
+        set -- ${kak_opt_windowing_modules}
+        if [ $# -gt 0 ]; then
+            echo 'try %{ '
+            while [ $# -gt 1 ]; do
+                echo "require-module ${1} } catch %{ "
+                shift
+            done
+            echo "require-module ${1} }"
+        fi
+    }
+}
--- a/rc/windowing/kitty.kak
+++ b/rc/windowing/kitty.kak
@@ -1,11 +1,12 @@
-
-hook -group kitty-hooks global KakBegin .* %sh{
-    if [ "$TERM" = "xterm-kitty" ] && [ -z "$TMUX" ]; then
-        echo "require-module kitty"
-    fi
-}
+# https://sw.kovidgoyal.net/kitty/index.html
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
 
 provide-module kitty %{
+
+# ensure that we're running on kitty
+evaluate-commands %sh{
+    [-z "${kak_opt_windowing_modules}" ] || [ "$TERM" = "xterm-kitty" ] || echo 'fail Kitty not detected'
+}
 
 declare-option -docstring %{window type that kitty creates on new and repl calls (kitty|os)} str kitty_window_type kitty
 
--- a/rc/windowing/iterm.kak
+++ b/rc/windowing/iterm.kak
@@ -1,15 +1,12 @@
 # https://www.iterm2.com
 # ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
 
-## The default behaviour for the `terminal` command is to open a vertical pane in
-## an iTerm session if not in a tmux session.
-hook global KakBegin .* %sh{
-    if [ "$TERM_PROGRAM" = "iTerm.app" ] && [ -z "$TMUX" ]; then
-        echo "require-module iterm"
-    fi
-}
-
 provide-module iterm %{
+
+# ensure that we're running on iTerm
+evaluate-commands %sh{
+    [-z "${kak_opt_windowing_modules}" ] || [ "$TERM_PROGRAM" = "iTerm.app" ] || echo 'fail iTerm not detected'
+}
 
 define-command -hidden -params 2.. iterm-terminal-split-impl %{
     nop %sh{
--- a/rc/windowing/screen.kak
+++ b/rc/windowing/screen.kak
@@ -1,12 +1,13 @@
 # http://gnu.org/software/screen/
 # ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
 
-hook -group GNUscreen global KakBegin .* %sh{
-    [ -z "${STY}" ] && exit
-    echo "require-module screen"
-}
 
 provide-module screen %{
+
+# ensure that we're running under screen
+evaluate-commands %sh{
+    [-z "${kak_opt_windowing_modules}" ] || [ -n "$STY" ] || echo 'fail screen not detected'
+}
 
 define-command screen-terminal-impl -hidden -params 3.. %{
     nop %sh{
--- a/rc/windowing/tmux.kak
+++ b/rc/windowing/tmux.kak
@@ -1,13 +1,12 @@
 # http://tmux.github.io/
 # ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
 
-hook global KakBegin .* %sh{
-    if [ -n "$TMUX" ]; then
-        echo "require-module tmux"
-    fi
-}
-
 provide-module tmux %{
+
+# ensure we're running under tmux
+evaluate-commands %sh{
+    [ -z "${kak_opt_windowing_modules}" ] || [ -n "$TMUX" ] || echo 'fail tmux not detected'
+}
 
 define-command -hidden -params 2.. tmux-terminal-impl %{
     evaluate-commands %sh{
--- a/rc/windowing/x11.kak
+++ b/rc/windowing/x11.kak
@@ -1,12 +1,11 @@
 # x11
 
-hook global KakBegin .* %sh{
-    if [ -n "$DISPLAY" ] && [ -z "$TMUX" ]; then
-        echo "require-module x11"
-    fi
-}
-
 provide-module x11 %{
+
+# ensure that we're running in the right environment
+evaluate-commands %sh{
+    [-z "${kak_opt_windowing_modules}" ] || [ -n "$DISPLAY" ] || echo 'fail DISPLAY is not set'
+}
 
 # termcmd should be set such as the next argument is the whole
 # command line to execute


sjjf:
Rebased against master to pick up fixes for the failed tests.

mawww:
Seems this was not correctly rebased and contains a merge commit, could you clean that up ?

sjjf:
Sorry, my git-fu isn't particularly strong . . .

I hope that's fixed up now.

sjjf:
Just wondering if this is being considered for merging - it addresses a real issue that I'm having. I didn't raise an issue report because this fix was pretty simple and I figured a fix was better than an issue report . . .

mawww:
Hey, sorry it is taking a while to go through that. I am on the fence about the tradeoff we need to make between having a clear precedence between the different windowing systems and making each support script standalone.

With that change, if you dont have the x11 module in your config, detection will still try to load it. Maybe we would need some kind of fallback in the case the module is not available.

On the other hand, it does simplify things quite a bit, and users that go out of their way to just include the module they care about could just not use detection and require-module themselves in their kakrc.

Any additional opinions on that ?

lenormf:
I'm also conservative about the individual script loading feature via `autoload`, especially now that there is a substantial amount of scripts loaded by default.

But there might be a way to tweak this patch to get to a compromise: how about an `str-list` option that instructs this detection hook what configurations to detect, and in what order?

In @sjjf's case, the option would be set (in their user configuration) to `"kitty" "x11"`.

We could even expand the concept to `termcmd`.

sjjf:
Hmm - I was assuming that anything under `rc/` was allowed to assume *everything* under `rc/` would be there, and hence it would be safe to unconditionally pull them in. If we're *not* working on that assumption then it gets a lot more complicated.

I think there's a little bit of an internal philosophical conflict here, though. If all the scripts are supposed to be standalone (and implement their own detection and loading) then it's always going to be problematic having an autoload feature, because each script is doing its own thing and could easily step on the toes of other scripts. From the other direction, if we're going to build a system that supports autoloading and automated detection of things (particularly when there are different implementations available, like the `rc/windowing/` scripts) then we're always going to have problems with getting coherent behaviour - things will be determined by implementation details like module autoload ordering, which can only be resolved by something like this patch, explicitly *breaking* the standalone nature of each script.

It seems to me like the most coherent approach would be to take all the autoloading and detection code out of the various modules and put it in its own set of scripts, which would be explicitly designed to be internally consistent (thereby fixing issues like mine), and probably configurable (via things like @lenormf detection ordering/limiting option). The rest of the current `rc/` directory would be a library that could be used by either the standard autoload code or by anyone who wanted to build their own system without using autoload.

Actually, it may be better to wrap each module's detection code in a function that the autoload code could call to make its decisions - that would probably be the best approach to making the scripts reusable in both autoload and no-autoload use cases. I don't know if that's feasible without patching the autoloading code, though.

That would be a very intrusive change, though, and it's more a question of design philosophy than anything else - @mawww is the only one who can make that sort of judgement.

I can certainly tweak this to support specifying the detection options and ordering, though, if that's the preferred solution right now.

lenormf:
The editor will only load whatever scripts are in `${XDG_CONFIG_HOME}/kak/autoload/`, to allow cherry-picking functionalities.

If there are inter-dependencies, they will be handled via `:require-module`, and missing dependencies in `autoload` would be reported to the user in the form of errors.

(FYI)

mawww:
I think it does make sense to centralize the detection and selection logic in a script, as currently each windowing support script has to be aware of others and try not to step on each other too much. It also would enable users to drop the detection script and implement their own if they disagree on what the priority should be (and the detection script could eventually respect an option that specifies that).

That said, the detection script should tolerate missing modules and correctly ignore them, the issue is that the only way to know if a module is available it to try to load it. One solution that come to mind would be to have that ordering option ("kitty" "tmux" "x11") then if Kitty is detected, do `try %{ require-module kitty } catch %{ set current -remove window-ordering kitty; detect-window }`

(this still require a set -remove functionality that is missing).

sjjf:
I've just pushed a draft/WIP implementation that searches for available modules from a str-list option and loads the first module that was found from that list. It's not intended to be the final implementation, but it demonstrates what I think is a workable approach to this particular problem. It may also provide a model for dealing with any similar cases where multiple alternative options are available.

I haven't implemented any support for overriding the hard-coded default list, but I imagine it'll be pretty simple to do so.

If this implementation approach is acceptable then I can clean it up and make it user-configurable, and I'll squash the changes down to give a clean commit history.

rc/windowing/detection.kak
```diff
@@ -0,0 +1,55 @@
...
+
+define-command -hidden detect-windowing-modules -docstring '
+detect-windowing-module: run the detection functions for the windowing modules.
+This creates the detected_modules str-list, containing only the modules
+relevant to the current runtime environment.' \
+%{
+    evaluate-commands %sh{
+        wmods=${kak_opt_windowing_modules}
+        for mod in ${wmods}; do
+            echo "try %{ detect-module-${mod} }"
```
lenormf, line 34: In order to be more conservative with the amount of options declared, I think you could move that scope into the `KakBegin` hook below, and just try to load the module directly.

We don't really need a list of all the modules that are detected because only the first one will be loaded anyway.

rc/windowing/detection.kak
```diff
@@ -0,0 +1,55 @@
...
+    # load the first module on the list.
+    #
+    # At this point we know that these modules exist and are relevant. The list
+    # is ordered by preference, so we can simply load the first module on the
+    # list.
+    evaluate-commands %sh{
+        for i in ${kak_opt_windowing_modules_detected}; do
+            echo "require-module ${i}"
+            break
+        done
```
lenormf, line 53: We probably don't need a loop here:

```sh
printf 'require-module %s' "${kak_opt_windowing_modules_detected%% *}"
```

sjjf: Nice - I was stumped by the lack of arrays (I'm totally a bash person), but I didn't think of that approach.

rc/windowing/detection.kak
sjjf: So there are two reasons I wanted to do it this way. The first is because of the larger design philosophy question I raised - separating out the various different elements of detection, `require-module`, and the definition of the actual module code. Conceptually at least the `windowing_modules_detected` option could be pushed up to a higher level and collect more modules to load (and the order in which to load them). Conceptually at least you could also move the creation of the `windowing_modules` list into each of the modules' `provide-module` code - effectively registering each of the modules that implement windowing functionality, which would then allow the user to implement their own or pull in an external plugin or similar and have it fit into the broader system. So the idea was to provide an initial proof of concept for the design approach I was suggesting.

The second is because I couldn't see a clean way to handle attempting to load a module, having it fail, and then falling through to the next one, programmatically. I'm probably just missing something in the way that try/catch works, or something in the integration between sh%{} code blocks and the kak script level, but short of programmatically building a long set of nested try/catch blocks I can't see a clean way to handle it. Separating out the detection like this makes it very simple.

rc/windowing/detection.kak
krobelus: To also support the scenario where a module is not included by the user, we can try the next one:

```sh
define-command require-windowing-module %{ evaluate-commands %sh{
	set -- $kak_opt_windowing_modules_detected
	module="$1"
	test -n "$module" || exit # No windowing module available.
	shift
	printf %s "try %{
		# Try to load the first module.
		require-module $1
	} catch {
		# If the module failed to load, remove it from the list of modules and call
		# ourselves recursively.
		set-option global windowing_modules_detected $@
		require-windowing-module
	}"
} }
```

This is just like described in https://github.com/mawww/kakoune/pull/3457#issuecomment-636394625, but without the `set-option -remove` feature, which would make this a bit nicer.

sjjf: Is that kind of recursion supported in this context? I'm not sure exactly how the %sh{} construct is implemented, but if each one is its own /bin/sh interpreter there's scope for this to blow up pretty seriously.

Also, is that a better approach than using a single additional option to collect state for further processing? It seems like a far more complex solution in many ways, even aside from the recursion question.

sjjf: Also, reimplemented so it does the recursion within a single shell invocation (untested, just trying this idea on for size):

```
define-command require-windowing-module %{ evaluate-commands %sh{
    try_require_module () {
        set -- $kak_opt_windowing_modules_detected
        if test -n "$1"; do
            module=$1
            shift
            printf %s "try %{
                # Try to load the first module.
                require-module $module
                set current windowing_modules_detected
            } catch {
                set current windowing_modules_detected $@
            }"
            try_require_module
        fi
    }
    try_require_module
} }
```
I'm still not sure it's a better approach than my first one, but it avoids recursing with each level being a whole new shell.

sjjf:
I've just spent some time trying to make different implementations work, and at this point I'm thinking this two-pass approach really is the best model. I've tried doing it with a single pass over the list of modules and I just can't figure out how to make it work while /also/ running the detection code (which is pretty fundamental).

I'm trying to achieve these things:
- only load modules that are relevant to the current runtime environment (which means we need to run the detection logic somewhere)
- only load /one/ module
- allow the user to specify the list of modules to try and load, and the order they'll be tried
- allow the user to add their own modules or plugins (without needing to patch the core `rc/` scripts)

The way I see it there's three approaches:
1) hard-code the list of modules and detection logic in a central location (my initial implementation)
2) a two-pass model that builds a list of present and relevant modules and then loads one of them
3) a single pass model that runs the detection code before deciding whether to try and load the module

Option 1 is out, because it's hard-coded, it centralises the detection logic (meaning the user can't just drop in their own module or plugin or whatever), and it makes it difficult to allow the user to adjust the ordering. Option 3 I can't get to work at all - this may be my lack of skill with kak scripting, but even taking that into account the logic just keeps getting convoluted (mostly because I need shell to implement the logic, but the state information is only available through kak script functions, and I can't figure out how to stitch it all together).

Option 2 (the current state of the code, modulo any further cleanup or bug fixes) feels clean and simple to me, works without any difficulty, and supports user extension very easily. Unless there's a better way to do this which I'm just not seeing can you please consider this version of the code for merging?

If you'd prefer the full merge squashed down to a single commit I'm happy to oblige.

rc/windowing/detection.kak
sjjf: @krobelus this recursive model works for loading the first module in the list, but that's only useful if we already have a list of modules that's been filtered to remove anything not relevant to the current environment. And once you've got that filtered list you don't need to do this kind of recursive jiggery-pokery, you can just load the first module in the list.

I've tried implementing something that replaces the `require-module` with a call to a function provided by each module that checks its environment and calls `require-module` if it's relevant, but it just doesn't work. I'm not sure why it's failing - the functions are very simple, and the only difference between this version (which runs fine, loading the first module in the list) and the failing version is replacing `require-module $module` with a call to `try-require-module-$module`, defined something like this:
```
define-command -hidden try-require-module-tmux %{
    evaluate-commands %sh{
        if [ -n "$TMUX" ]; then
            printf %s "require-module tmux"
        else
            printf %s "fail 'Could not find tmux environment'"
        fi
    }
}
```
I'm really not sure why this isn't working, but even if it did I think it's a lot more convoluted than the two-step approach in the current patch set.

lenormf:
Why try to load only a single module? We want all the relevant ones to be loaded, so that users can use `*-terminal` commands individually.

We just need to load them in the right order, so that the `focus` and `terminal` aliases are populated correctly.

sjjf:
@lenormf I hadn't actually thought that might be an option . . . I still don't think it's a /good/ option, though.

Right now we /can't/ load all modules that are relevant: most of the modules make at least some attempt to not load when they're not in an exclusive environment (most of them explicitly check for `$TMUX` and won't load if it's defined). It's a bit half-hearted - there's no attempt to check for `$STY`, and no attempt to restrict any other combination of environments, so as it stands you can have both tmux and screen (which makes no sense!), but not tmux and kitty or tmux and x11; you can also have screen and kitty or screen and x11, which is exactly the opposite of what you'd expect from the tmux behaviour. None of this is particularly coherent design. The coherent approach would be to make /everything/ mutually exclusive, or ensure that /everything/ can load at the same time.

Also, not all the modules define the same set of aliases and functions, or they define the same functions differently, or they define the same functionality with different functions. Kitty defines a `terminal-tab` alias, which is currently the only additional alias any of these modules define, but most of the modules define their own functions with similar behaviour - iTerm provides an `iterm-terminal-tab` function, and both tmux and screen support the same functionality (via the `*-terminal-window` functions), so there's no reason they /couldn't/ provide that alias. tmux, screen and iterm provide `*-terminal-vertical` and `*-terminal-horizontal`, but kitty doesn't (though it could technically support it). iterm provides an `iterm-terminal-window` which is different from the tmux and screen functions with equivalent names. The modules are all different and distinct enough that a) they're not mutually compatible in any meaningful way (beyond the basic `terminal` and `focus` aliases), and b) there's no reason to believe that they won't interfere with each other in unexpected ways.

If we try and load all relevant modules it could become very messy - depending on the modules loaded and the loading order you might end up with a `terminal` alias provided by tmux, two `*-terminal-window` functions provided by iterm and tmux that do completely different things, and two each of the `*-terminal-(vertical|horizontal)` functions that are pretty much functionally identical. Yes, the user could figure out how to do stuff, but as default behaviour I think this is just confusing and incoherent.

Ensuring we only load one windowing module at a time ensures that we're presenting a coherent and predictable interface that probably matches what the user expects (e.g. if you're running tmux on top of kitty you probably expect to be using tmux windowing functionality rather than kitty windowing functionality). With this implementation, if you want to do things differently you can override everything by setting the `windowing_modules` option to an empty list and manually loading the modules you want and setting up whatever aliases you want, but the default behaviour is simple, predictable and easily configurable.

The alternative would be to rewrite all of this code to be coherent and consistent, with consistent function names that did consistent things. But this would actually /still/ be inconsistent! Because tmux and screen windows are different to iTerm windows, kitty and iTerm tabs map to tmux/screen windows, kitty `terminals` could be either a split in the current window or a whole new OS window, and so on. These are all different systems that each do things differently in both big and obvious ways, but also in subtle and confusing ways. I really don't think it's a good approach to try and treat them as interchangeable or able to be commingled - I think the current approach of having a fairly minimal basic external interface and then letting the modules implement whatever functionality is appropriate makes the most sense. And in that context, I don't think it makes sense to load more than one module at a time.

lenormf:
Here's my implementation. It fixes your problem, and allows users to interact with a detected multiplexer of their choosing.

```patch
diff --git a/rc/windowing/detection.kak b/rc/windowing/detection.kak
new file mode 100644
index 00000000..5a127f92
--- /dev/null
+++ b/rc/windowing/detection.kak
@@ -0,0 +1,12 @@
+declare-option str-list windowing_modules 'tmux' 'screen' 'kitty' 'iterm' 'x11'
+
+hook global KakBegin .* %{ evaluate-commands %sh{
+    eval "set -- ${kak_quoted_opt_windowing_modules}"
+
+    i=$#
+    while [ "${i}" -gt 0 ]; do
+        module="$(eval "printf %s \"\$${i}\"")"
+        printf 'try-load-module-%s\n' "${module}"
+        i=$((i - 1))
+    done
+} }
diff --git a/rc/windowing/iterm.kak b/rc/windowing/iterm.kak
index c48aebc8..442f0031 100644
--- a/rc/windowing/iterm.kak
+++ b/rc/windowing/iterm.kak
@@ -1,12 +1,12 @@
 # https://www.iterm2.com
 # ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
 
-## The default behaviour for the `terminal` command is to open a vertical pane in
-## an iTerm session if not in a tmux session.
-hook global KakBegin .* %sh{
-    if [ "$TERM_PROGRAM" = "iTerm.app" ] && [ -z "$TMUX" ]; then
-        echo "require-module iterm"
-    fi
+define-command -hidden try-load-module-iterm %{
+    evaluate-commands %sh{
+        if [ "$TERM_PROGRAM" = "iTerm.app" ]; then
+            echo "require-module iterm"
+        fi
+    }
 }
 
 provide-module iterm %{
diff --git a/rc/windowing/kitty.kak b/rc/windowing/kitty.kak
index d847f863..bfc06c12 100644
--- a/rc/windowing/kitty.kak
+++ b/rc/windowing/kitty.kak
@@ -1,8 +1,12 @@
+# https://sw.kovidgoyal.net/kitty/index.html
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
 
-hook -group kitty-hooks global KakBegin .* %sh{
-    if [ "$TERM" = "xterm-kitty" ] && [ -z "$TMUX" ]; then
-        echo "require-module kitty"
-    fi
+define-command -hidden try-load-module-kitty %{
+    evaluate-commands %sh{
+        if [ -n "${KITTY_WINDOW_ID}" ]; then
+            echo "require-module kitty"
+        fi
+    }
 }
 
 provide-module kitty %{
diff --git a/rc/windowing/screen.kak b/rc/windowing/screen.kak
index 8c77e1e9..bcb25f4b 100644
--- a/rc/windowing/screen.kak
+++ b/rc/windowing/screen.kak
@@ -1,9 +1,13 @@
 # http://gnu.org/software/screen/
 # ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
 
-hook -group GNUscreen global KakBegin .* %sh{
-    [ -z "${STY}" ] && exit
-    echo "require-module screen"
+
+define-command -hidden try-load-module-screen %{
+    evaluate-commands %sh{
+        if [ -n "$STY" ]; then
+            echo "require-module screen"
+        fi
+    }
 }
 
 provide-module screen %{
diff --git a/rc/windowing/tmux.kak b/rc/windowing/tmux.kak
index b3c3dbdc..a89549cf 100644
--- a/rc/windowing/tmux.kak
+++ b/rc/windowing/tmux.kak
@@ -1,10 +1,13 @@
 # http://tmux.github.io/
 # ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
 
-hook global KakBegin .* %sh{
-    if [ -n "$TMUX" ]; then
-        echo "require-module tmux"
-    fi
+
+define-command -hidden try-load-module-tmux %{
+    evaluate-commands %sh{
+        if [ -n "$TMUX" ]; then
+            echo "require-module tmux"
+        fi
+    }
 }
 
 provide-module tmux %{
diff --git a/rc/windowing/x11.kak b/rc/windowing/x11.kak
index 6bbb842e..73d377cd 100644
--- a/rc/windowing/x11.kak
+++ b/rc/windowing/x11.kak
@@ -1,9 +1,11 @@
 # x11
 
-hook global KakBegin .* %sh{
-    if [ -n "$DISPLAY" ] && [ -z "$TMUX" ]; then
-        echo "require-module x11"
-    fi
+define-command -hidden try-load-module-x11 %{
+    evaluate-commands %sh{
+        if [ -n "$DISPLAY" ]; then
+            echo "require-module x11"
+        fi
+    }
 }
 
 provide-module x11 %{
```

sjjf:
@lenormf I can confirm that does resolve my initial problem (it would be nice if POSIX sh had a cleaner mechanism for iterating over a list in reverse).

I still don't think it's a good idea to load more than one of the windowing modules at once, though, certainly not without work to clean them up and fix the various inconsistencies I noted, and in my opinion not at all. But this is @mawww 's project, so the decision is obviously up to him.

rc/windowing/kitty.kak
```diff
@@ -1,8 +1,15 @@
...
-hook -group kitty-hooks global KakBegin .* %sh{
-    if [ "$TERM" = "xterm-kitty" ] && [ -z "$TMUX" ]; then
-        echo "require-module kitty"
-    fi
+define-command -hidden detect-module-kitty -docstring '
+detect-module-kitty: determine if the kitty module is relevant to the current
+environment' \
+%{
+    evaluate-commands %sh{
+        if [ "$TERM" = "xterm-kitty" ]; then
```
lenormf, line 9: I think the following is a better heuristic here, might as well fix it:

```suggestion
        if [ -n "${KITTY_WINDOW_ID}" ]; then
```

Screwtapello:
I think it's pretty clear that the different windowing modules were *intended* to be loaded up at once - otherwise they'd each define, say `:terminal`, instead of each defining their own variant and letting the user set an alias to point to their preferred implementation.

On the other hand, it's not really practical to load more than one at a time:

- you can't be running inside both Kitty and iTerm2 (since they don't nest)
- you can't be running inside both iTerm2 and X11 (since iTerm2 is not ported to X11)
- You *can* run inside both X11 and tmux, X11 and screen, Kitty and tmux, Kitty and screen, iTerm2 and tmux, or iTerm2 and screen... but they both can detach from the outer terminal, so just because you *started* with a particular outer terminal doesn't mean it's still around
- You can run inside both tmux and screen, but (1) why would you want to? and (2) just like X11, the outer terminal might no longer be around

Is there some combination of inner and outer that's practically useful?

sjjf:
@mawww I've now pushed a cleaned up, better documented version which I've squashed down to a single commit. I'd like you to consider this as my preferred version for being merged (after extensive discussion ;-)

Obviously any further comments are welcome.

rc/windowing/kitty.kak
sjjf: Either version actually works, but I think the $TERM value is more "correct", since kitty explicitly documents that $TERM must be set to "xterm-kitty" otherwise things may break

lenormf: Some values for this variable are known to be interchangeable.

That's why `tmux.kak` checks the `$TMUX` environment variable (same with `screen.kak` and `$STY`), and not if `$TERM = tmux-256color` - because you can also use `screen-256color` simultaneously for Tmux, Screen and Kitty.

In fact, when debugging Kakoune rendering issues with users, we often ask them to test with `TERM=screen-256color`.

lenormf:
@Screwtapello 

>you can't be running inside both Kitty and iTerm2 (since they don't nest)

That's why a new option was introduced (`windowing_modules` in the original patch), to allow users to specify exactly what tools they're using and the loading priority (which is their position in the list).

sjjf:
@Screwtapello I interpreted the `:terminal` and `:focus` aliases as more of an external interface, exposing the various module-specific implementations in a consistent way - if each module implemented the same functions then you'd /need/ a centralised location that loaded exactly one module, which means the modules can't be effectively standalone - they can only exist in conjunction with the loading code. Aliases simply overwrite whatever was there before, rather than conflicting.

It looks to me like things just kind of grew into the current system, with an initial implementation that only handled one environment, then a second implementation (with a simple way to ensure that they didn't load over the top of each other - hence the explicit checks for tmux), then more implementations that copied the existing code along with the explicit checks for tmux, but with no one stepping back and thinking through the design of the whole thing or how all the modules interacted.

I'm sure there /are/ combinations of windowing modules that could be made to do something useful, but I seriously doubt that it's worth trying to build the core system in order to support that - if someone wants to do their own thing they can disable the automated detection and manually load the modules they want.

mawww:
I am still a bit unhappy about that whole set of `detect-module-blah` commands, something about them rubs me the wrong way.

One idea came to mind, what if loading a module could fail ? In that case we could just do

```
try %{ require-module foo } catch %{ require-module bar } catch %{ require-module baz }...
```

Generating this command from a `str-list windowing_modules` would be trivial. `require-module` would fail either because the module does not exist, or because loading the module triggered a failure.

It turns out this is already supported:

```
provide-module x11 %{
   evaluate-commands %sh{ [ -n "$DISPLAY" || echo "fail DISPLAY is not set" }
   ...
```

Of course that means a module cannot be force loaded anymore, which can be an issue for some modules such as tmux, that are able to work using the client env (so the server might not run in a TMUX session, but if a client runs in one, using $kak_client_env_TMUX allows us to still open a pane in that session). it would be nice to solve that issue as well.

One way to solve it would be to make this pattern not set the module as loaded (currently a fail inside module still sets it as loaded), and the validation logic be a bit more tolerant (so say tmux would fail to load from the server, but if tryied again from a client that is in a tmux session, the kak_client_env_TMUX value would make the module accept to load, allowing the user to explicitely reload it (or hook on ClientCreate).

how does that sound ?

sjjf:
@mawww If you're not comfortable with the `detect-module-foo` approach I'm happy to drop it - that was one of the things I wanted guidance on from you.

With the chained try/catch model it should be easy enough to set it up to handle force-loading of selected modules - the logic can be that if you provide a non-empty (or default) `windowing_modules` list then the environment tests are applied, but if it's an empty list then any attempt to load the module manually bypasses the environment tests. So something like:

```sh
evaluate-commands %sh{
    set -- ${kak_opt_windowing_modules}
    [  $# -gt 0 ] && [ -n "$DISPLAY" ] || echo "fail DISPLAY is not set"
}
```
Then the user just sets the `windowing_modules` option to an empty list in their kakrc (which means the generated try/catch chain will be empty and nothing will load automatically) and manually sets everything up however they want.

Wouldn't using a per-client windowing module run up against the issue I noted of inconsistent module interfaces (i.e. different sets of aliases, different sets of functions, functions with the same name but different meanings)? If you're assuming the user is doing all this themselves (manually loading the exact set of modules they want, and setting up all the aliases the way they want) then that's not an issue, but in the default scenario I still don't think loading multiple windowing modules is a good idea.

I don't know enough about the way state is shared between server and client to be able to comment usefully beyond that . . .

sjjf:
In the meantime, here's an implementation that works as I suggested.

I need to update the docstring for the `windowing_modules` option before this is committed . . .

lenormf:
Feel free to set this PR as "draft", since all commits will ultimately have to be squashed anyway, and we're nowhere near a good solution for this problem yet.

sjjf:
@lenormf thanks for the suggestion.

There's certainly been a lot of back and forth over what I thought was a dead simple ten minute fix when I started . . .

useredsa:
> That said, the detection script should tolerate missing modules and correctly ignore them, the issue is that the only way to know if a module is available it to try to load it.
What's your opinion on setting a global option like `x11-module-available` after the `provide-module`? It could be done manually or automatically by kakoune.

sjjf:
Rebased against the current head, and tidied up some of the documentation. Assuming it passes all tests I'll remove the WIP and see if @mawww wants to merge this.

sjjf:
Dammit, I forgot to update the comment as well. I think that's three versions of the same basic text now: the docstring, the comment at the top of the file, and the commit message . . .

rc/windowing/detection.kak
```diff
@@ -0,0 +1,63 @@
...
+
+ set-option global windowing_modules 'mod1' 'mod2' 'mod3'
+
+To disble automatic loading of windowing modules so you can
+customise the module loading yourself set the windowing_modules
+option to an empty list:
+
+ set-option global windowing_modules
+
+and add your own module loading code." \
```
mawww, line 47: I think this docstring goes a bit too much into details. I dont think options should document their default value and what syntax is used to set them in the docstring, current value is displayed as a completion, and how set-option work is general Kakoune knowledge.

rc/windowing/detection.kak
```diff
@@ -0,0 +1,63 @@
...
+#
+# On load each module must attempt to detect the environment it's appropriate
+# for, and if the environment isn't appropriate it must fail with an error.
+# In addition, each module must check for the length of the windowing_modules
+# str-list option defined below, and must /not/ check for an appropriate
+# environment if the list is empty. An example of this test:
+#
+# evaluate-commands %sh{
+#     set -- ${kak_opt_windowing_modules}
+#     [ $# -eq 0 ] || [ -n "$TMUX" ] || echo 'fail tmux not detected'
```
mawww, line 15: `[ -z "$kak_opt_windowing_modules" ]` would work as well, avoiding the need to the `set` line.

More generally, I wonder about using that option in each module. It seems like convoluted way to force loading of modules, I guess it can be good enough for now, but not entirely satisfying.

rc/windowing/detection.kak
sjjf: Trying to be self-documenting and discoverable, I guess that docstring went way overboard . . .

rc/windowing/detection.kak
sjjf: The two ideas "enable/disable environment detection" and "load modules in this order" are pretty orthogonal, and we could configure them separately with different options, but then we have to decide how those two options would interact, and we're right back into the weeds we were in at the start of the discussion. That's why I've been trying to focus on the user experience rather than the abstract logic - almost all users will want a single module loaded which is relevant to their current environment, and anything else is just going to be confusing. Anyone who /doesn't/ want that is likely to be doing something so weird that there's no point us trying to implement anything to support it, and the best option is going to be getting completely out of the way and letting them do whatever they want. Using an empty list to indicate that is overloading the meaning a /little/, but only a little, and it makes for a simple implementation as well as an interface that's pretty easy to understand even if it's not perfectly elegant.

sjjf:
Cleaned up the docstring and the unnecessary use of `set -- ${kak_opt_windowing_modules}`

mawww:
Thanks for working on that, merged.

PR #3459, opened 2020-04-30
Fix +line:col initial buffer position when connecting to session

krobelus:
A command line argument like `+line[:column]` can be used to specify a
target line and column for the first file.

This did not work when connecting to a session, because the client
opens its file parameters with `-e "edit file1; edit file2"` which is
executed after the initial buffer position is set. Work around this by
setting the position for the first file and avoid moving the position
in unrelated files.

Reproduce:

	kak -s foo
	kak -c foo +4:11 README.asciidoc

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -1164,8 +1164,14 @@ int main(int argc, char* argv[])
                 }
             }
             String new_files;
-            for (auto name : files)
-                new_files += format("edit '{}';", escape(real_path(name), "'", '\\'));
+            for (auto name : files) {
+                new_files += format("edit '{}'", escape(real_path(name), "'", '\\'));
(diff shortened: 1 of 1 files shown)


src/main.cc
```diff
@@ -1164,8 +1164,16 @@ int main(int argc, char* argv[])
                 }
             }
             String new_files;
-            for (auto name : files)
+            for (auto name : files) {
                 new_files += format("edit '{}';", escape(real_path(name), "'", '\\'));
```
lenormf, line 1168: It'd be cleaner if you passed the coordinate to the `edit` command directly.

krobelus: Yeah, thanks!

PR #3466, opened 2020-05-05
Fix perl regex highlighting for operator //

buffet:
It used to see the second / of the // operator as the potential start of a regex, turning the entire document into a string in situations like `$path // "some/default"`.


mlvzk:
+1 I need this.

PR #3467, opened 2020-05-05
Remove extra space added after comment_line option when using :comment-line

SeerLite:
```sh
# This line is commented
```
VS.
```sh
#This line is commented
```

It's something that has been bugging me a lot. I don't usually comment lines with code with that extra space (first example).

If it's wanted, one could add the space to `comment_line` and it's (almost) the same as the old behavior.
For instance, the old behavior can be replicated like this:
```
hook global WinSetOption filetype=.* %{
	set-option buffer comment_line "%opt{comment_line} "
}
```

If that's too hacky and you think most people wouldn't like this change, I could add that extra space to all the `comment_line`s defined for each filetype in `comment.kak`. This way at least it's possible to remove the space from within user configuration without manually editing `comment.kak`.


Delapouite:
Looks like it conflicts with https://github.com/mawww/kakoune/pull/2069

SeerLite:
That's... Wow I never thought someone would want to have it hardcoded instead of keeping it as an option. I thought this was just how it was implemented originally.

IMO that shouldn't be hardcoded. It can easily be set up with the hook I gave as example.

SeerLite:
I just want to clarify that I didn't want my previous comment to mean that I thought hardcoding it was a dumb idea, I was just looking at it as a bug instead of a feature (heh) and it didn't cross my mind that it would be the other way around.

Do you think there's any downsides to using a hook like my example, instead of having it harcoded in comment.kak? Maybe I'm missing something important about other languages as I mostly deal with config/dot files, shell scripts and occasionally Python.

SeerLite:
Ah. Commands can be overriden with `define-command -override`. I guess I should have looked into that first before making this PR heh.

I still think this is a nice solution and could be implemented, but I haven't seen anyone using comments in the same way I do so I don't think anyone really cares about this at all :/

PR #3484, opened 2020-05-15
Use iTerm2 "synchronised update" sequences around redraws.

### LINKED ISSUE #3482: [builtin-terminal-ui] Support the iTerm2 "synchronised updates" feature
One of the potential problems with the builtin-terminal-ui branch is that most terminals don't provide a way for the application to signal that it's repainting the display, so there's a good chance that the the terminal will repaint *its* display halfway through, and then again once the application update has finished, leading to distracting flickering.

The most portable approach to this problem is to just ensure repaints are as concise as possible, but since there's no guaranteed maximum atomic size, and since repaints can be arbitrarily large, this is not a solution.

iTerm2 supports a [synchronised update control sequence](https://gitlab.com/gnachman/iterm2/-/wikis/synchronized-updates-spec). The linked specification has more details, but it basically amounts to:

  - send `ESC P = 1 s ESC \` when a repaint begins
  - send `ESC P = 2 s ESC \` when a repaint ends

Currently, iTerm2 is the only terminal to support this feature, other terminals should ignore it without incident. In that respect, it's like the "enable focus in/out events" sequence Kakoune already uses.

According to [tmux's maintainer](https://github.com/mintty/mintty/issues/881#issuecomment-628475165), tmux only [sends these sequences](https://github.com/tmux/tmux/blob/31621036ad38ef3a2d88da838eb021430ea3c944/tty.c#L1437-L1465) to the outer terminal and does not implement them for the inner terminal, "but I will do that as soon as there are some programs apart from tmux that use it..."

Screwtapello:
In terminals that support it, this sequence causes the terminal to not redraw *its* output until the application has finished, reducing redraw flickering.

The sequence is defined in:

    https://gitlab.com/gnachman/iterm2/-/wikis/synchronized-updates-spec

...and is apparently also supported by kitty and libVTE.

Fixes #3482.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -107,7 +107,6 @@ Kakoune dependencies are:
 
  * A {cpp}17 compliant compiler (GCC >= 7 or clang >= 6) along with its
    associated {cpp} standard library (libstdc{pp} or libc{pp})
- * ncurses with wide-characters support (>= 5.3, generally referred to as libncursesw)
 
 To build, just type *make* in the src directory.
 To generate man pages, type *make man* in the src directory.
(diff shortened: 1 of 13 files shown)


mawww:
That has been cherry-picked

PR #3486, opened 2020-05-16
Handle dedent in rust syntax

pickfire:
Naively handle dedent after , and ; which works most of the cases but
does not work with multi level dedent.


mawww:
I dont think I have a copyright waiver commit from you, as this change is reasonably big could you attach one (see CONTRIBUTING)

pickfire:
I added a copyright weaver.

@mawww Do you know if there is any method to fix auto-formatting in rust? I mean to auto-format while preserving the comments, so far I have yet seen any external formatting tools that does stuff in the way that wraps all types of comments correctly.

krobelus:
I think you mean to reflow lines; `|fmt -p//` works for me.

If you want handle different types of comments, then you could create a script `fmt.sh`:
```sh
#!/bin/sh
input="$(cat)"
prefix="$(printf %s "$input" | grep -E --only-matching --max-count=1 '^//[!/]? ')"
printf %s "$input" | fmt --prefix="$prefix" "$@"
```

Or if you prefer kak script:

```kak
define-command my-fmt %{
	evaluate-commands -save-regs | %{
		evaluate-commands -draft %{
			execute-keys ';<a-x>s'
			try %{
				execute-keys '^//[/!]? <ret>'
				set-register | "fmt --prefix='%val{selection}'"
			} catch %{
				set-register | "fmt"
			}
		}
		execute-keys '|<ret>'
	}
}
```

pickfire:
@krobelus Thanks but either method does not work that nice to me, is there any one liner? By the way, how did you set the kak script for `autowrap_fmtcmd`?

krobelus:
Whoops I didn't realize you mean autowrap

When a line overflows the column limit by appending at the end, the formatcmd is not invoked, so it doesn't work yet. When inserting in the middle of a line it should already work.
Here's a rough draft how the former could be changed. I used `set-option global autowrap_fmtcmd 'fmt.sh -w %c'`

```diff
diff --git a/rc/tools/autowrap.kak b/rc/tools/autowrap.kak
index c900abc6..36ca1b2f 100644
--- a/rc/tools/autowrap.kak
+++ b/rc/tools/autowrap.kak
@@ -21,3 +21,12 @@ define-command -hidden autowrap-cursor %{ evaluate-commands -save-regs '/"|^@m'
             ## if we're adding characters past the limit, just wrap them around
-            execute-keys -draft "<a-h><a-k>.{%opt{autowrap_column}}\h*[^\s]*<ret>1s(\h+)[^\h]*\z<ret>c<ret>"
+            evaluate-commands %sh{
+                format_cmd=$(printf %s "${kak_opt_autowrap_fmtcmd}" \
+                             | sed "s/%c/${kak_opt_autowrap_column}/g")
+                printf %s "
+                    evaluate-commands %{
+                        execute-keys %{<a-;><a-x><a-;>|${format_cmd}<ret><a-;>;}
+                    }
+                "
+            }
+
         } catch %{
```

pickfire:
@krobelus Thanks, should we send a autowrap patch for that?

mawww:
Tests are currently failing.

pickfire:
@mawww Thanks for the ping. I forgot to wait for the build in CI since I am lazy to take time building and testing kakoune on my computer which takes around 5 - 10 minutes.

krobelus:
>. I forgot to wait for the build in CI since I am lazy to take time building and testing kakoune on my computer which takes around 5 - 10 minutes.

If you are changing `rc/` files only, you only really need to build Kakoune once. Use `make debug=yes -j` for a potentially faster build.

PR #3493, opened 2020-05-17
terminal_ui: restore termios settings before suspending Kakoune.

### LINKED ISSUE #3488: [builtin-terminal-ui] Suspend with ctrl-Z messes up terminal
### Steps 

Running the builtin-terminal-ui branch, suspend kakoune with ctrl-Z and then run, for example, `ls -l`.

### Outcome 
image

### Expected 
image

Screwtapello:
`TerminalUI::suspend()` is responsible for undoing all Kakoune's modifications to the terminal state, actually suspending the process, then re-applying all the modifications after Kakoune wakes back up.

Previously, the "undo" and "reapply" steps for termios settings were both after the suspend point, so on some platforms they were incorrect when the user arrived back at the shell prompt.

Now, the termios "undo" step is back before the suspend point, and the undo and reapply steps should be in exactly reversed order.

Fixes #3488.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -107,7 +107,6 @@ Kakoune dependencies are:
 
  * A {cpp}17 compliant compiler (GCC >= 7 or clang >= 6) along with its
    associated {cpp} standard library (libstdc{pp} or libc{pp})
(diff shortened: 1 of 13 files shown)


mawww:
This was cherry-picked.

Issue #2377, opened 2018-09-10
Backslashes in the documentation are mis-handled


lenormf:
### Steps 

* `:doc command-parsing non-quoted-words`

In the asciidoc source, the sentence reads:

```
If they start with `\\` followed by `%`, `'` or `"`, then that leading
`\\` character is discarded.
```

### Outcome 

The sentence is rendered as:

```
If they start with \` followed by `%, ' or ", then that leading `\` is discarded.
```

ManDay:
See also section *Disabling Hooks* under `:doc keys` and `:doc hooks`. Both render incorrectly. One could say, the feature is therefore not documented.

lenormf:
Fixed.

PR #3502, opened 2020-05-23
Fix inline code issue in hooks.asciidoc

m-bdf:


krobelus:
Unfortunately this makes the backslash disappear in the rendering of `:doc hooks`, (pressing backslash disables hooks for the next command).
There is a similar issue in the "Non Quoted words" section of `:doc command-parsing`. They should all render as just one backslash.

Running
```sh
git grep '\\`'
```

shows that the only cases where we actually escape backticks are when we want to render a single backslash. One possible way to fix would be to special case "\`\\\`" to also be highlighted as code.

```diff
diff --git a/rc/tools/doc.kak b/rc/tools/doc.kak
index de799cdc..7b8bdde3 100644
--- a/rc/tools/doc.kak
+++ b/rc/tools/doc.kak
@@ -119 +119 @@ define-command -params 1 -hidden doc-render %{
-    doc-render-regex \B(?<!\\)`(?=\S)[^\n]+?(?<=\S)(?<!\\)`\B \A|.\z 'H' mono
+    doc-render-regex \B(?:`\\`|(?<!\\)`(?=\S)[^\n]+?(?<=\S)(?<!\\)`)\B \A|.\z 'H' mono
```

Or we disallow escaping of the backticks, since it's not used.
Then the above regex would simply be

```
 \B`(?=\S)[^\n]+?(?<=\S)`\B
```

lenormf:
#2377

PR #3508, opened 2020-05-27
Fix setext-style markdown header highlight

pickfire:
Reproduce:

```markdown
- item

header
------
```

### DIFF
--- a/rc/filetype/markdown.kak
+++ b/rc/filetype/markdown.kak
@@ -66,8 +66,8 @@ add-highlighter shared/markdown/listblock/content default-region ref markdown/in
 add-highlighter shared/markdown/inline/code region -match-capture (`+) (`+) fill mono
 
 # Setext-style header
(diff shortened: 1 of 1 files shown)


Screwtapello:
What happens in a situation like this?

```
abcdef
ghijkl
------
```

I think headings require `\n\n` to ensure there is a blank line before the heading text. I'm not sure why headings need *three* newlines after a bulleted-list, though.

pickfire:
@Screwtapello That will still highlight the header but the non header won't be highlighted.

I also found that this does not work.

```
three
---

four
---
```

I believe https://github.com/mawww/kakoune/blob/09a45a2e964e1efb39335926b2b88ce71e460ee4/rc/filetype/markdown.kak#L62 is probably wrong, commenting that affects the header.

pickfire:
@Screwtapello I sent a fix for the issue you mentioned.

Screwtapello:
I haven't tested your change yet, but it looks much better!

pickfire:
How should one add tests for highlighting? I don't find it a simple task to test out highlighting.

PR #3510, opened 2020-05-27
Add default tabstop and max width for rust

pickfire:
The default should follow rust default style mentioned in
https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/guide.md#indentation-and-line-width


lenormf:
We don't want to have a dependency to the autowrap script, which defines `autowrap_column`.

As for the `tabstop` option, language support scripts should not enforce coding style conventions on users.The linked document also mentions "multiples of four", so users who prefer 8 spaces would be effected.

pickfire:
@lenormf Yes, I understand. The documentation does not enforce but recommends, I believe we should take that recommendation as a default rather than some random 8 values that come from nowhere and is more of a surprise to a new kakoune user using rust.

Is there any way to only add `autowrap_column` if `autowrap` is enabled? I believe this should be the same as before, use the one that is recommended as the default rather than some random values. I personally prefer 80 but that is what was recommended.

lenormf:
Users can select which scripts will be loaded by creating an `autoload` directory. Users who don't select `autowrap.kak` will be met with an error once `rust.kak` loads.

The only way to get around that (since there's no `autowrap` module), is to [redeclare `autowrap_column`](https://github.com/mawww/kakoune/blob/master/rc/tools/autowrap.kak#L1-L2) in `rust.kak`.

In any case, the values set in this PR are the result of your interpretation of *recommendations* which will override any user settings. Not good.

pickfire:
@lenormf Is there any way for it to be the default but not to override any user settings?

pickfire:
I also noticed that autowrap column on 100 means 99 in kakoune unlike 100 in vim.

lenormf:
A more efficient way to encourage users to follow the recommended code style would be to edit the command line documented in [the wiki](https://github.com/mawww/kakoune/wiki/Format#rustfmt) to make it use a specific indentation level and column limit. Similarly to how the command for [jq](https://github.com/mawww/kakoune/wiki/Format#jq) is setup.

pickfire:
I believe that is the less efficient way, I mean wiki is good but the reason why I looked into the wiki in the first place is because it is missing the default stuff. If stuff is in place (like what fish shell did), people don't need to look into the wiki for these kind of basic stuff.

lenormf:
Now we're back to square one.

*Recommendations* are **not standard** defaults. Users can write valid Rust code with whatever indentation level they've set in their configuration, the editor should not try to second-guess them.

The idiomatic way of writing code that follows a standard in Kakoune is to use a formatter, or eventually something like EditorConfig.

pickfire:
But why **standard** defaults are set? These should be a reason right. I believe that is because people used to use 80 characters back then with tabs for C code, same goes for kakoune such that it made its way as the default.

But say, if a new editor were to be written in rust, I bet there might be a high chance the default tabstop is 4 characters and uses space instead of tab, do you think it would be good to make that the default for all languages including Makefile?

Say, Makefile have a hard requirement on tab but they editor people say they don't care. Wouldn't that be bad on Makefile people?

I believe the recommendations are the same, they made it recommendation for this reason, people can override it but I think the editor should not override it. I believe the order of setting indentation boils down to:

1. user
2. language
3. editor

In my humble opinion, the editor should not interfere with language recommendations.

Linus torvalds even have a say on 80-column terminals https://lkml.org/lkml/2020/5/29/1038. (I don't really like what he said personally, like raspberry pi 4 slow compilation and 100+ widths)

pickfire:
What can we do about this? I don't think recommending the users the editor's default is a good idea, the language's default should come which I think is a better choice.

lenormf:
C.f. [comment](https://github.com/mawww/kakoune/pull/3510#issuecomment-634706841).

pickfire:
C.f.?

pickfire:
> As for the tabstop option, language support scripts should not enforce coding style conventions on users.The linked document also mentions "multiples of four", so users who prefer 8 spaces would be effected.

We can keep four as default. But I think this might affect other users? What if we parse the rustfmt configuration and editorconfig to know how many space do they prefer but still default back to 4? This allows a zero configuration method to use sensible default when no custom spacing is used.

Ordoviz:
The `tabstop` option only changes the displayed width of a tab character. It does not affect file contents like `indentwidth`, which already defaults to 4.

I find `autowrap` annoying when writing code. `rustfmt` does a better job at this. Kakoune should not behave smart and surprising by default.

pickfire:
> The tabstop option only changes the displayed width of a tab character. It does not affect file contents like indentwidth, which already defaults to 4.

Yes, tab not able to insert space, kakoune tabbing and wrapping I say is very bad.

> I find autowrap annoying when writing code. rustfmt does a better job at this. Kakoune should not behave smart and surprising by default.

Kakoune rust indentation is smart and I think better rust.vim but the wrapping story of kakoune is really bad, almost useless. I find myself checking the line manually and do the counting myself, pipe to fmt is useless for rust. I wonder if we can use a kakoune function for formatting or we should write ourselves a partial formatter just for kakoune rust?

PR #3513, opened 2020-05-28
Fix regions highlighter's region computation with replace-ranges

### LINKED ISSUE #3495: Crash trying to highlight lines hidden by multi-line replace-ranges
Hi, thanks for the new mutli-line replace-ranges feature, it feels really powerful.

I found an issue that consistently reproduces when replacing some highlighted range.

### Steps

On 5fd548d5cfa9a96a247319398654c944eb52a671, using `make debug=yes`.

1. In the root of the Kakoune repo create the two files `repro.kak` and `in.kak` with the contents below.
2. Run `kak -n -e 'source repro.kak; edit in.kak; fold'`
3. Press `j` and observe the crash with the stacktrace below.

It seems like the iterator in `to_next` goes beyond the last line of the file.
My guess is that it tries to highlight `"some-quoted-string"`, which is hidden by the replace-ranges.
This reproducer needs highlighters for `sh` and `kak` filetypes. I'll try to create a minimal reproducer eventually.

```kak
$ cat repro.kak
source rc/detection/file.kak
source rc/filetype/kakrc.kak
source rc/filetype/sh.kak

declare-option range-specs test_ranges
add-highlighter global/ replace-ranges test_ranges

define-command fold %{
	evaluate-commands -draft %{
		# Select from START to EOF.
		execute-keys %{%sSTART.*<ret>}
		# Replace the selection with PLACEHOLDER.
		evaluate-commands %sh{
			printf %s "set-option global test_ranges $kak_timestamp $(
				printf ' %s|PLACEHOLDER' $kak_selections_desc
			)"
		}
	}
}
```

```kak
$ cat in.kak
nop %sh{
# START
}

nop %sh{
: "some-quoted-string"
}
```

### Outcome/Stacktrace 

```
Received SIGSEGV, exiting.
Pid: 3865281
Callstack:
./src/kak(Kakoune::Backtrace::Backtrace()+0x2b) [0x556efbb2b32f]
./src/kak(Kakoune::signal_handler(int)+0xcc) [0x556efbd175a0]
/usr/lib/libc.so.6(+0x3c370) [0x7f2920c68370]
./src/kak(void Kakoune::utf8::to_next<Kakoune::BufferIterator, Kakoune::BufferCoord>(Kakoune::BufferIterator&, Kakoune::BufferCoord const&)+0x97) [0x556efbc96c1f]
./src/kak(Kakoune::ThreadedRegexVM<Kakoune::BufferIterator, (Kakoune::RegexMode)5>::to_next_start(Kakoune::BufferIterator&, Kakoune::ThreadedRegexVM<Kakoune::BufferIterator, (Kakoune::RegexMode)5>::ExecConfig const&, Kakoune::CompiledRegex::StartDesc const&)+0xbe) [0x556efbc91db4]
./src/kak(Kakoune::ThreadedRegexVM<Kakoune::BufferIterator, (Kakoune::RegexMode)5>::exec(Kakoune::BufferIterator const&, Kakoune::BufferIterator const&, Kakoune::BufferIterator const&, Kakoune::BufferIterator const&, Kakoune::RegexExecFlags)+0x1cd) [0x556efbc8b7a3]
./src/kak(Kakoune::RegexIterator<Kakoune::BufferIterator, (Kakoune::RegexMode)1, Kakoune::Regex>::next()+0x173) [0x556efbc831b9]
./src/kak(Kakoune::RegexIterator<Kakoune::BufferIterator, (Kakoune::RegexMode)1, Kakoune::Regex>::It::It(Kakoune::RegexIterator<Kakoune::BufferIterator, (Kakoune::RegexMode)1, Kakoune::Regex>&)+0x2a) [0x556efbc8301a]
./src/kak(Kakoune::RegexIterator<Kakoune::BufferIterator, (Kakoune::RegexMode)1, Kakoune::Regex>::begin()+0x2e) [0x556efbc7a7da]
./src/kak(Kakoune::RegexHighlighter::add_matches(Kakoune::Buffer const&, std::vector<Kakoune::Range<Kakoune::BufferCoord>, Kakoune::Allocator<Kakoune::Range<Kakoune::BufferCoord>, (Kakoune::MemoryDomain)6> >&, Kakoune::Range<Kakoune::BufferCoord>)+0x222) [0x556efbc6b466]
./src/kak(Kakoune::RegexHighlighter::get_matches(Kakoune::Buffer const&, Kakoune::Range<Kakoune::BufferCoord>, Kakoune::Range<Kakoune::BufferCoord>)+0x499) [0x556efbc6be05]
./src/kak(Kakoune::RegexHighlighter::do_highlight(Kakoune::HighlightContext, Kakoune::DisplayBuffer&)
```

krobelus:
The test case shows an example where get_regions_for_range() would compute
an invalid region (0.3,0.2) because of a replace-ranges highlighter. The
bad range was the last region in the range. Clamp it to make it fit in
the range.

Fixes #3495


mawww:
Seems we got to almost the same code change for that, thanks for investigating this.

Issue #3480, opened 2020-05-15
replace-ranges highlighting has its style overriden if its followed by any non-newline character.


SolitudeSF:
![image](https://user-images.githubusercontent.com/10168645/82012214-f2fe8180-967f-11ea-817c-a05fe169aaed.png)
![image](https://user-images.githubusercontent.com/10168645/82012228-fd208000-967f-11ea-96f1-dd7df7b47c72.png)
if preceding text doesnt have highlighters applied, then range is highlighted correctly
![image](https://user-images.githubusercontent.com/10168645/82012325-483a9300-9680-11ea-88a6-55e3829460db.png)

krobelus:
Something that feels like the same issue, but I'm not sure:

The `{yellow}` markup from the replace-ranges style renders correctly in the first paragraph, but it is overridden by the regex highlighter in the second paragraph.

```
$ echo > in '
    word
    word

    keyword
    keyword'

$ echo > rc.kak 'add-highlighter global/ regex keyword 0:blue

declare-option range-specs fold_ranges %val{timestamp} \
    '2.1,3.8|    {yellow}--- 2 lines:{default} word' \
    '5.1,6.11|    {yellow}--- 2 lines:{default} keyword' \

add-highlighter global/fold replace-ranges fold_ranges'

$ kak -n -e 'source rc.kak; edit in'
```
![image](https://user-images.githubusercontent.com/6853872/82758728-966e3580-9de8-11ea-88bc-505f1136ba4d.png)

mawww:
I believe the current behaviour is correct, we do want replace-range to appear selected if inside a selection for example. To opt out of this the final face attributes can be used.

PR #3517, opened 2020-05-28
Do not apply user-defined highlighters to replaced ranges

krobelus:
I haven't tested this thoroughly but would appreciate feedback on direction.

Regex-based highlighting is not yet supported in text that
was replaced by a replace-ranges highlighter.
Skip matches that fall in a replaced range.
This means that replace-ranges highlighters have to format their replacement
text themselves.

Issue described in this comment:
https://github.com/mawww/kakoune/issues/3480#issuecomment-633253043


mawww:
I am not sure about that, currently this should be achievable by setting the "final" attribute in the replace range faces (`+F`) which would prevent following highlighters from changing that.

I wonder what makes more sense, on one hand it seems that replace-ranges would usually want full control over how its displayed, but on the other hand, I can imagine some highlighters that would make sense to apply to them (selection highlighting, `#if 0` commenting...)

krobelus:
I see, I didn't think about `+F`, which is enough for my use case, so this patch is not necessary.

If a replace-ranges highlighter replaces some code with other code, highlighting could be useful.
For example it could replace several lines by joining them all to one line, which should still be highlighted. However, that's a bit contrived, maybe not worth supporting.
For folding the highlighting of replaced text is probably not necessary, because it's fine if context line is not highlighted. I can't think of a good use case now, but I'm sure there are some.

PR #3523, opened 2020-05-30
Add support for Elvish

alexherbo2:
https://elv.sh

### DIFF
--- a/rc/filetype/elvish.kak
+++ b/rc/filetype/elvish.kak
@@ -1,0 +1,89 @@
+# Elvish
+# https://elv.sh
+
+hook global BufCreate '.*\.elv' %{
+  set-option buffer filetype elvish
+}
+
+hook global WinSetOption filetype=elvish %{
+  require-module elvish
+  evaluate-commands set-option window static_words %opt{elvish_keywords}
+  add-highlighter window/elvish ref elvish
+  hook -group elvish window InsertChar '\n' elvish-new-line-inserted
+  set-option window extra_word_chars '$' '~' '-'
+  hook -always -once window WinSetOption filetype=.* %{
+    remove-highlighter window/elvish
+    remove-hooks window elvish
+  }
+}
+
+provide-module elvish %{
+  declare-option -hidden str-list elvish_keywords 'all' 'assoc' 'base' 'bool' 'break' 'cd' 'chr' 'constantly' 'continue' 'count' 'dir-history' 'dissoc' 'drop' 'each' 'eawk' 'echo' 'eq' 'esleep' 'eval-symlinks' 'exec' 'exit' 'explode' 'external' 'fail' 'fclose' 'fg' 'float64' 'fopen' 'from-json' 'from-lines' '-gc' 'get-env' 'has-env' 'has-external' 'has-key' 'has-prefix' 'has-suffix' 'has-value' '-ifaddrs' 'is' '-is-dir' 'joins' 'keys' 'kind-of' '-log' 'make-map' 'multi-error' 'nop' 'not' 'not-eq' 'ns' 'one' 'only-bytes' 'only-values' 'ord' 'order' '-override-wcwidth' 'path-abs' 'path-base' 'path-clean' 'path-dir' 'path-ext' 'peach' 'pipe' 'pprint' 'prclose' 'print' 'put' 'pwclose' 'rand' 'randint' 'range' 'read-line' 'read-upto' 'repeat' 'replaces' 'repr' 'resolve' 'return' 'run-parallel' 'search-external' 'set-env' 'slurp' '-source' 'splits' 'src' '-stack' 'styled' 'styled-segment' 'take' 'tilde-abbr' '-time' 'time' 'to-json' 'to-lines' 'to-string' 'unset-env' 'wcswidth'
+
+  add-highlighter shared/elvish regions
+  add-highlighter shared/elvish/code default-region group
+
+  # Comments
+  # https://elv.sh/ref/language.html#syntax-convention
+  add-highlighter shared/elvish/comment region '#' '$' fill comment
+
+  # String
+  # https://elv.sh/ref/language.html#string
+  add-highlighter shared/elvish/double_string region '"' '(?<!\\)"' regions
+  add-highlighter shared/elvish/single_string region -recurse "(?<!')('')+(?!')" "(^|\h)\K'" "'(?!')" regions
+
+  # Variables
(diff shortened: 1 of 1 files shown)


rc/filetype/elvish.kak
```diff
@@ -0,0 +1,89 @@
...
+    }
+    # Remove empty line indent
+    try %{
+      execute-keys -draft 'k<a-x>s^\h+$<ret>d'
+    }
+  }
+
+  define-command -hidden elvish-fetch-keywords %{
+    set-register dquote %sh{
+      curl \
```
SolitudeSF, line 72: hell no, that should never be a thing

alexherbo2: ?

rc/filetype/elvish.kak
```diff
@@ -0,0 +1,89 @@
...
```
Screwtapello, line 72: This is a bit scary to see in a file-handling module, for all the predictable reasons (`curl` isn't POSIX, we don't want to make network requests at runtime, etc.) so it would be useful to have a comment clearly pointing out that this command is not actually used at runtime, it's just for people maintaining and updating this script in the future. Perhaps the whole function should be commented out, just to make it clear that it's not used at runtime - it's easy enough for somebody who wants to work on this script to uncomment it.

I'm also a little worried that this function is brittle - by the time this file needs maintenance, what are the odds that Elvish has added a new source file, or renamed an old one, or reformatted the source so that the "extract keywords" logic no longer works?

rc/filetype/elvish.kak
```diff
@@ -0,0 +1,89 @@
...
+  # Variables
+  # https://elv.sh/ref/language.html#variable
+  add-highlighter shared/elvish/code/variable regex '\$@?(?:[\w-]+:)*[\w-]+' 0:variable
+
+  evaluate-commands %sh{
+    # Keywords
+    eval "set -- $kak_quoted_opt_elvish_keywords"
+    regex="\\b(?:\\Q$1\\E"
+    shift
+    for keyword do
```
lenormf, line 44: I think there's a syntax error here: `for keyword; do`.

You could also use the same `join()` function from `c-family.kak` to join up a list of tokens efficiently.

rc/filetype/elvish.kak
```diff
@@ -0,0 +1,89 @@
...
+      regex="$regex|\\Q$keyword\\E"
+    done
+    regex="$regex)\\b"
+    printf 'add-highlighter shared/elvish/code/keywords regex %s 0:keyword\n' "$regex"
+
+    # String
+    # https://elv.sh/ref/language.html#string
+    for id in double_string single_string; do
```
lenormf, line 52: I think inlining the lines would be more readable than iterating over two elements here.

```
add-highlighter shared/elvish/single_string/fill default-region fill string
add-highlighter shared/elvish/double_string/fill default-region fill string
```

rc/filetype/elvish.kak
```diff
@@ -0,0 +1,89 @@
...
+  evaluate-commands %sh{
+    # Keywords
+    eval "set -- $kak_quoted_opt_elvish_keywords"
+    regex="\\b(?:\\Q$1\\E"
+    shift
+    for keyword do
+      regex="$regex|\\Q$keyword\\E"
+    done
+    regex="$regex)\\b"
+    printf 'add-highlighter shared/elvish/code/keywords regex %s 0:keyword\n' "$regex"
```
lenormf, line 48: You could also use that opportunity to populate the `static_words` option!

rc/filetype/elvish.kak
krobelus: Maybe this can be based on something like `elvish -c 'use builtin; keys $builtin:'` for more flexibility. The output has some adornments, probably easy to filter them away.

rc/filetype/elvish.kak
```diff
@@ -0,0 +1,89 @@
...
+    try %{
+      execute-keys -draft 'K<a-&>'
+    }
+    # Remove empty line indent
+    try %{
+      execute-keys -draft 'k<a-x>s^\h+$<ret>d'
+    }
+  }
+
+  define-command -hidden elvish-fetch-keywords %{
```
mawww, line 70: I dont think this should be here, I understand that you'd like to save that logic somewhere but I dont think we need to have it in every Kakoune installation.

rc/filetype/elvish.kak
```diff
@@ -0,0 +1,89 @@
...
+  hook -always -once window WinSetOption filetype=.* %{
+    remove-highlighter window/elvish
+    remove-hooks window elvish
+  }
+}
+
+provide-module elvish %{
+  declare-option -hidden str-list elvish_keywords 'all' 'assoc' 'base' 'bool' 'break' 'cd' 'chr' 'constantly' 'continue' 'count' 'dir-history' 'dissoc' 'drop' 'each' 'eawk' 'echo' 'eq' 'esleep' 'eval-symlinks' 'exec' 'exit' 'explode' 'external' 'fail' 'fclose' 'fg' 'float64' 'fopen' 'from-json' 'from-lines' '-gc' 'get-env' 'has-env' 'has-external' 'has-key' 'has-prefix' 'has-suffix' 'has-value' '-ifaddrs' 'is' '-is-dir' 'joins' 'keys' 'kind-of' '-log' 'make-map' 'multi-error' 'nop' 'not' 'not-eq' 'ns' 'one' 'only-bytes' 'only-values' 'ord' 'order' '-override-wcwidth' 'path-abs' 'path-base' 'path-clean' 'path-dir' 'path-ext' 'peach' 'pipe' 'pprint' 'prclose' 'print' 'put' 'pwclose' 'rand' 'randint' 'range' 'read-line' 'read-upto' 'repeat' 'replaces' 'repr' 'resolve' 'return' 'run-parallel' 'search-external' 'set-env' 'slurp' '-source' 'splits' 'src' '-stack' 'styled' 'styled-segment' 'take' 'tilde-abbr' '-time' 'time' 'to-json' 'to-lines' 'to-string' 'unset-env' 'wcswidth'
```
mawww, line 21: I dont like introducing Kakoune options that are only used once, why cant that be directly in the shell block as a shell variable ?

rc/filetype/elvish.kak
lenormf: The Crystal module also needs to be cleaned up, it has the issues.

alexherbo2: For maintenance and further reference, I like to remember I can easily update the keywords, and change the method if needed.

PR #3524, opened 2020-05-30
Crystal: Slight style change

alexherbo2:


lenormf:
The comments on #3523 revealed a few problems that were reproduced from this file, so I guess comments for one apply to the other.

The only other thing is using unicode characters for separators. As long as it's an actual letter e.g. § it should be fine, but it would probably do the community a disservice if the built-in scripts used cats.

@mawww shouldn't there be a recommendation about that, documented somewhere?

PR #3527, opened 2020-06-01
Fixed need for double escaping \ in custom object desc (\\\\ to match \)

mreppen:
In custom object desc (eg `<a-a>c`), matching `\` requires entering `\\\\`. This is because of unescaping `\\` at the same time as `\,`. This fix simply avoids unescaping `\\`, so that the input behaves like a normal regex.

Example:
Buffer: `\\_,` with cursor at `_`
`<a-a>c\\,\,` now gives `\[\_,]` as expected.

### DIFF
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -1302,7 +1302,8 @@ void select_object(Context& context, NormalParams params)
                     struct error : runtime_error { error(size_t) : runtime_error{"desc parsing failed, expected <open>,<close>"} {} };
 
                     auto params = cmdline | split<StringView>(',', '\\') |
-                        transform(unescape<',', '\\'>) | static_gather<error, 2>();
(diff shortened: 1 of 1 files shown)


mawww:
I think a better way to fix that would be to go back to the initial design where we get two prompts, one for opening, one for closing, so the separator becomes the <ret> key.

mreppen:
There is one important feature missing, for which I would gladly open an issue: custom object history. I often (would like to) define custom objects on the fly or modify one. For that use-case a history would really help, even if it is only of length 1. I currently do it with macros, which is a bit tedious.

But I imagine a history would be easier to handle/work with in the current one-prompt design. (although I do have an idea for how to do it in two-prompt)

I was not using kakoune regularly before the current design so I can't otherwise comment further.

PR #3529, opened 2020-06-01
Fix default value of $XDG_CONFIG_HOME in doc

m-bdf:
`$XDG_CONFIG_HOME` is the directory containing user-specific configurations for many apps, it should default to `$HOME/.config`, instead of `$HOME/.config/kak` (which is `$KAKOUNE_CONFIG_DIR`) as currently stated in the documentation.

### DIFF
--- a/doc/kak.1
+++ b/doc/kak.1
@@ -175,7 +175,7 @@ If not started with the \fB\-n\fR switch, Kakoune will first load
 
 .nf
 .RS
-if the \fI$KAKOUNE_CONFIG_DIR/kak/autoload\fR directory exists, recursively load every
+if the \fI$KAKOUNE_CONFIG_DIR/autoload\fR directory exists, recursively load every
 \fI*.kak\fR files in it and its sub-directories
 .RE
 .fi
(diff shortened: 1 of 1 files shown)


lenormf:
Thanks!

m-bdf:
Similar issue when referencing `$KAKOUNE_CONFIG_DIR`, the subdirectory `kak` is already included in the expansion of the variable, so writing `$KAKOUNE_CONFIG_DIR/kak` is wrong ;)

m-bdf:
There also seems to be an issue with `$XDG_RUNTIME_DIR`, which should default to `$TMPDIR` instead of `$TMPDIR/kakoune`... but then the fact that Kakoune works only in the subdirectory `kakoune` (or even `kakoune/$USER`) of `$XDG_RUNTIME_DIR` won't appear anymore in this section of the documentation ^^`

This is not an issue for `$XDG_CONFIG_HOME`, as the actual directory used by Kakoune is specified in another environment variable (`$KAKOUNE_CONFIG_DIR`, which defaults to `$XDG_CONFIG_HOME/kak`).

Unfortunately, there is no similar `$KAKOUNE_SESSION_DIR` environment variable defaulting to `$XDG_RUNTIME_DIR/kakoune` (or `$TMPDIR/kakoune/$USER` if `$XDG_RUNTIME_DIR` is not set, which actually makes `$XDG_RUNTIME_DIR` having a default value a bit useless imho)...

PR #3532, opened 2020-06-02
Add comment highlighting and options for asciidoc.

### LINKED ISSUE #3068: :comment-line inserts wrong character in asciidoc documents
Using `:comment-line` in an asciidoc document does insert a `#` character instead of the correct `//` comment character. 
Also the syntax highlighting for comments in asciidoc documents does not work correctly. Commented lines show the same highlighting as normal text lines.

tmccombs:
Fixes #3068.

### DIFF
--- a/rc/filetype/asciidoc.kak
+++ b/rc/filetype/asciidoc.kak
@@ -33,6 +33,9 @@ add-highlighter shared/asciidoc/ regex (\A|\n\n)[^\n]+\n\^{2,}\h*$ 0:header
 add-highlighter shared/asciidoc/ regex (\A|\n\n)=\h+[^\n]+$ 0:title
 add-highlighter shared/asciidoc/ regex (\A|\n\n)={2,}\h+[^\n]+$ 0:header
 
+add-highlighter shared/asciidoc/ regex ^//(?:[^\n/][^\n]*|)$ 0:comment
+add-highlighter shared/asciidoc/ regex ^(/{4,}).*?\n(/{4,})$ 0:comment
+
 add-highlighter shared/asciidoc/ regex ^\h+([-\*])\h+[^\n]*(\n\h+[^-\*]\S+[^\n]*)*$ 0:list 1:bullet
 add-highlighter shared/asciidoc/ regex ^(-{3,})\n[^\n\h].*?\n(-{3,})$ 0:block
 add-highlighter shared/asciidoc/ regex ^(={3,})\n[^\n\h].*?\n(={3,})$ 0:block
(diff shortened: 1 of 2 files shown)


rc/filetype/asciidoc.kak
```diff
@@ -33,6 +33,9 @@ add-highlighter shared/asciidoc/ regex (\A|\n\n)[^\n]+\n\^{2,}\h*$ 0:header
 add-highlighter shared/asciidoc/ regex (\A|\n\n)=\h+[^\n]+$ 0:title
 add-highlighter shared/asciidoc/ regex (\A|\n\n)={2,}\h+[^\n]+$ 0:header
 
+add-highlighter shared/asciidoc/ regex ^//[^\n]*$ 0:comment
+add-highlighter shared/asciidoc/ regex ^////.*?^////$ 0:comment
```
lenormf, line 37: The [spec](https://asciidoc.org/userguide.html#X26) doesn't mention the minimum amount of slash signs required to start a comment, but looking at the example I would assume it's more than three, and the line has to contain only slash signs. Do you have more info about that?

Additionally, the idiomatic way of handling that would be with a region. Added to the consideration above, it would look something like (untested):

```kak
add-highlighter shared/asciidoc/regions regions
add-highlighter shared/asciidoc/regions/comment_block region ^/{3,}$ ^/{3,}$ fill comment
```

rc/tools/comment.kak
```diff
@@ -10,8 +10,9 @@ declare-option -docstring "characters inserted after a commented block" \
 
 # Default comments for all languages
 hook global BufSetOption filetype=asciidoc %{
-    set-option buffer comment_block_begin '///'
-    set-option buffer comment_block_end '///'
```
lenormf, line 14: C.f. my comment above, this might be correct?

rc/filetype/asciidoc.kak
tmccombs: according to this https://github.com/asciidoc/asciidoc/blob/master/asciidoc.conf, it must be at least 4 slashes, and the line should only be slashes.

lenormf: https://github.com/asciidoc/asciidoc/blob/0beddc73873acb3724b3cb93999acf0ddba84fc5/asciidoc.conf#L304-L308

Got it.

The regex still has to be fixed though.

rc/tools/comment.kak
lenormf: Nevermind, it seems to be 4.

https://github.com/asciidoc/asciidoc/blob/0beddc73873acb3724b3cb93999acf0ddba84fc5/asciidoc.conf#L304-L308

rc/filetype/asciidoc.kak
tmccombs: Yep, I'll push up a fix later today.

mawww:
apply that fixup commit ?

somasis:
Is there anything else preventing this from being merged?

PR #3535, opened 2020-06-02
Rust highlight

pickfire:


rc/filetype/rust.kak
```diff
@@ -39,10 +39,9 @@ add-highlighter shared/rust/line_doctest region ^\h*//[!/]\h*```($|should_panic|
 add-highlighter shared/rust/line_doctest/marker region ```.* $ group
 add-highlighter shared/rust/line_doctest/marker/fence regex ``` 0:meta
 add-highlighter shared/rust/line_doctest/marker/keywords regex [\d\w] 0:meta # already matched above, just ignore comma
-add-highlighter shared/rust/line_doctest/inner region '^\h*//[!/]\h+#' ' ' group
+add-highlighter shared/rust/line_doctest/inner region '^\h*//[!/]( #(?= )|)' ' ' group
```
krobelus, line 42: I guess `( #(?= ))?` would be more idiomatic.

I'm wondering, what's special about `#` followed by a space in doc comments? I thought only `#!` has special meaning and everything else should be treated the same way.
If that's actually true, then something like this would look more correct to me:
```diff
diff --git a/rc/filetype/rust.kak b/rc/filetype/rust.kak
index 8a57ce9e..881eeeec 100644
--- a/rc/filetype/rust.kak
+++ b/rc/filetype/rust.kak
@@ -42 +42 @@ add-highlighter shared/rust/line_doctest/marker/keywords regex [\d\w] 0:meta # a
-add-highlighter shared/rust/line_doctest/inner region '^\h*//[!/]( #(?= )|)' ' ' group
+add-highlighter shared/rust/line_doctest/inner region '^\h*//[!/]' ' ' group
@@ -58 +58 @@ add-highlighter shared/rust/block_doc/doctest/inner default-region group
-add-highlighter shared/rust/block_doc/doctest/inner/hidden regex '^\h*\**\h*#' 0:meta
+add-highlighter shared/rust/block_doc/doctest/inner/hidden regex '^\h*\**\h*' 0:meta
```

Screwtapello: A `#` inside a doctest is used for text that's part of the test but not part of the documentation: https://doc.rust-lang.org/rustdoc/documentation-tests.html#hiding-portions-of-the-example

krobelus: Oh that's nifty, thanks!

pickfire:
Looks like the tests won't be able to accomodate testing highlight too long (out of the view). How should we handle that?

Also, is there a test that can be visually distinctive?

mawww:
You should be able to change the dimension of the window in the script file by using

```
ui_in '{ "jsonrpc": "2.0", "method": "resize", "params": [ 100, 80 ] }'
```

in the `script` file (to have a 100 line x 80 column window here).

pickfire:
I thought script is the output? How do I change it like that? Currently I use
```fish
mv in out; ../../../../src/kak -n -e 'source rc' -s kak-tests -ui json out < /dev/null | sed "s/.*/ui_out '&'/" > script; mv out in
```

mawww:
`script` is a shell script, but it usually uses two provided functions: ui_out (expects some json-rpc message from kakoune) and ui_in (send some json-rpc message to kakoune). If you add the suggested command at the top of your file, Kakoune will receive a UI resize message and should then send display command for a 100x80 window, which should allow you to match against more output in your `ui_out` call.

pickfire:
@mawww But the `script` is being written by the output I want to test, how do I do that? I did not use `./run` though, it is probably more troublesome to do that. I wish there is `./run --bless` like the one in rust to auto fix the tests.

mawww:
Your `script` file here has two ui_out commands, you just have to add a ui_in command before them and fix the second ui_out to take into account the additional lines that will get output.

It definitely would be nice to automate updating script files when we know whatever Kakoune outputs is correct.

pickfire:
@mawww Too troublesome, I need to wrap hundreds of lines manually, I rather split it into 10 separate tests or write fewer tests next time.

mawww:
Tried to merge that, but tests fail.

pickfire:
Weird, tests didn't fail locally.

lenormf:
Rebase the branch on `master`, then force push.

pickfire:
@lenormf Still broken, it's stuck. I also don't know why the output is split into multiple lines.

pickfire:
I also added fix for https://github.com/mawww/kakoune/issues/3586, I feel lazy to open a new pull request and then having to rebase it later, just dump all pull requests into one except those that is controversial.

pickfire:
I don't understand why is this failing, it passed locally though. Is there any way to just change the test snapshot output without having me to write manually?

mawww:
That was an escaping issue in the 'value' test. Thanks for that.

pickfire:
@mawww @lenormf I wonder if we only supports regex for highlighting, I noticed that some parts in javascript highlighing or html highlighting breaks, the javascript ones is a bit worse such that the whole page have the same color pink. I don't really know if regex is able to get if `>` is part of `<tag>` or just `<tag text="hello > world">` (this broke). Or maybe we can delegate the highlighting to language servers which I think most of them supports highlighting, some even semantic highlighting IIRC?

krobelus:
> I don't really know if regex is able to get if > is part of <tag> or just <tag text="hello > world"> 

This should be possible with regex, you'd just need to skip quoted strings.
However, [regex can't parse HTML](https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags/) in general, so there will always be cases where it fails.

> we can delegate the highlighting to language servers

Good idea, if you know how to do this maybe document it in the Wiki.

> I also added fix for #3586, I feel lazy to open a new pull request and then having to rebase it later, just dump all pull requests into one except those that is controversial.

To avoid having to checkout and rebase repeatedly when working on multiple pull requests I use [this script](https://git.sr.ht/~krobelus/git-branchless) which you may find useful.

PR #3537, opened 2020-06-03
src: Introduce a <percent> named key

lenormf:
Similarly to the <semicolon> key, make it easier to write
`:execute-keys` commands by replacing <percent> with `%`.

Highlighters can keep escaping the sign when regular expressions are
not quoted, but built-in scripts that use `%` as an editing primitive
have been modified to use the named key, for clarity.


Delapouite:
Should this new key be mentioned in https://github.com/mawww/kakoune/blob/master/doc/pages/keys.asciidoc like `<semicolon>`?

lenormf:
Ooops! Thanks!

PR #3540, opened 2020-06-03
src: Implement a `-N` flag to ignore user configuration

lenormf:
Users with a heavy configuration that customise a lot of default
settings/side-effects cannot easily have it ignored by the editor.

This commit implements a `-N` flag that instructs the editor to load
system files, but skip those written by the user. In terms of built-in
options:

* load files in %val{runtime}
* ignore files in %val{config}

Note that, despite being user-created, the `kakrc.local` file will
still be loaded.

### DIFF
--- a/doc/kak.1
+++ b/doc/kak.1
@@ -32,7 +32,7 @@ kak \- a vim inspired, selection oriented code editor
 .B kak \-s
 .I session_id
 .B \-d
-[\fB\-n\fR] [\fB\-ro\fR]
+[\fB\-n\fR|\fB\-N\fR] [\fB\-ro\fR]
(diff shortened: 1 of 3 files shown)


lenormf:
My usecase is when I want to reproduce an issue, I generally have to rename `~/.config/kak` to prevent it from being loaded, but still have all system scripts (i.e. installed by my distro's package) loaded.

And running `kak -n` really is too much work, so `kak -N` sounds good.

krobelus:
`HOME=$(mktemp -d) kak` should work in some scenarios

lenormf:
PING

lenormf:
Rebased on master.

PR #3541, opened 2020-06-03
doc: add comments for escape sequences in setup/restore terminal

Delapouite:
Hi

Most of the other ANSI escape sequences are already commented in the file, but these ones in the PR were missing.

### DIFF
--- a/src/terminal_ui.cc
+++ b/src/terminal_ui.cc
@@ -48,19 +48,36 @@ struct TerminalUI::Window::Line
     struct Atom
     {
         String text;
+        ColumnCount skip = 0;
         Face face;
 
-        friend bool operator==(const Atom& lhs, const Atom& rhs) { return lhs.text == rhs.text and lhs.face == rhs.face; }
+        ColumnCount length() const { return text.column_length() + skip; }
+        void resize(ColumnCount size)
+        {
+            auto text_len = text.column_length();
+            if (size < text_len)
+            {
+                text.resize(text.byte_count_to(size), 0);
+                skip = 0;
+            }
+            else
+                skip = size - text_len; 
+        }
+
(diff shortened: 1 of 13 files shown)


src/terminal_ui.cc
```diff
@@ -1250,9 +1250,13 @@ void TerminalUI::set_resize_pending()
 
 void TerminalUI::setup_terminal()
 {
+    // enable alternative screen buffer
     fputs("\033[?1049h", stdout);
+    // enable focus notify
     fputs("\033[?1004h", stdout);
+    // hide cursor
     fputs("\033[?25l", stdout);
+    // set special keypad mode
```
Screwtapello, line 1259: This is usually (i.e. in the original DEC terminal documentation) known as "application keypad" mode. If we're being explanatory, we might as well go all the way:
```suggestion
    // set application keypad mode, so the keypad keys send unique codes
```

src/terminal_ui.cc
```diff
@@ -1263,6 +1267,7 @@ void TerminalUI::restore_terminal()
     fputs("\033[?1004l", stdout);
     fputs("\033[?25h", stdout);
     fputs("\033>", stdout);
+    // reset SGR
```
Screwtapello, line 1270: Again, we might as well explain all the way:
```suggestion
    // set the terminal output back to default colours and style
```

lenormf:
I've actually been thinking we could make this more C++-y by storing the escape sequences as `constexpr` variables in the global scope (in a `struct` or whatever), which we would refer to by name, in the code.

The documentation added by this PR would be centralised where the sequences are declared.

Delapouite:
@Screwtapello thanks for your expertise on this topic. I've updated the comments accordingly.
@lenormf Yes centralising this magic values in proper constants would be nice. But I don't feel comfortable doing these now. So a future PR can go this way.

mawww:
cherry-picked.

PR #3549, opened 2020-06-06
doc faq: Subsections, more items

lenormf:


doc/pages/faq.asciidoc
```diff
@@ -155,36 +225,42 @@ for a given format for instance), but not much can be done about those
...
+is too much of a drawback when balanced against the low cost of using
+an interpreter in a regular shell scope (e.g. `%sh{ python -c "..." }`).
+The shell scope allows users to spawn any interpreter they want, for a minimal
+cost in terms of performance, it is therefore the reason why it's the only
+one available by default.
+
+=== What shell is used to expand `%sh{}` scopes ?
+
+The server expands shell scopes using the `sh` binary, stored in one of the
+directories where all the POSIX standard utilities can be found -this list
```
Screwtapello, line 249: `found -this` should probably be `found - this` (space before and after the punctuation) or `found-this` (no space)

doc/pages/faq.asciidoc
```diff
@@ -59,14 +63,110 @@ $ tic /path/to/kakoune/contrib/tmux-256color.terminfo
...
+You need a UTF-8 compliant locale, in order to render special characters.
+
+The way to do so might vary from distribution to another, but an easy way
+to verify that your locale is set correctly is to run the following command:
+
+----
+$ locale
+----
+
+The output should look similar to the following (with eventually another
```
krobelus, line 77: I think "eventually" is a false friend here, maybe: 

```
with your language and country code instead of `en_US`
```

doc/pages/faq.asciidoc
lenormf: Thanks for the review, I've addressed the comment in a "prose fix" commit.

doc/pages/faq.asciidoc
lenormf: Thanks for the review, I've addressed the comment in a "prose fix" commit.

doc/pages/faq.asciidoc
```diff
@@ -1,12 +1,14 @@
 = FAQ
 
-== How to pronounce the name of the project and what does it mean ?
+== The project
+
+=== How to pronounce the name of the project and what does it mean ?
```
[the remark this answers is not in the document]
lenormf: Correct, that's a French thing. Fixed.

doc/pages/faq.asciidoc
```diff
@@ -59,14 +63,42 @@ $ tic /path/to/kakoune/contrib/tmux-256color.terminfo
...
+(or one of the supported multiplexers), run Kakoune in a `tmux` session,
+and simply use the `:new` command to spawn new clients as you would
+have otherwise in an X11 environment.
+
+== Generic functionalities
+
+=== How can I explore the filesystem the way Vim's NerdTree does ?
+
+The builtin file completion engine used when opening a file for editing
+(using the `:edit` command and letting the suggestions popup in the menu
```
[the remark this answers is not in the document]
lenormf: Fixed.

doc/pages/faq.asciidoc
```diff
@@ -174,8 +174,9 @@ the language you're using with the help of the `|` key.
 
 Example: `%|indent<ret>` to indent an entire buffer with C code.
 
-Note that some languages have a default formatter set, which you can use
-with the `:format` command.
+Note that you can also set the `formatcmd` option, used by the `format`
+command to handle formatting in the current buffer without further
+interaction.
```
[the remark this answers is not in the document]
lenormf: Fixed.

doc/pages/faq.asciidoc
```diff
@@ -6,16 +6,16 @@
...
-much less efficient on a Windows system, the incentive to porting the
-project to this operating system is pretty low.
+As many features provided by UNIX systems would be missing, or albeit much
+less efficient, on a Windows system, the incentive to porting the project
+to this operating system is low.
 
-Moreover, you can get pretty decent performance by using Kakoune on Cygwin
-(which is officially supported).
+Moreover, you can achieve pretty decent performance by using Kakoune on
+Cygwin (which is officially supported).
```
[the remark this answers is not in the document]
lenormf: Only Cygwin is officially supported, I think we can let users pick alternatives (there's also MSYS, as far as I know).

doc/pages/faq.asciidoc
```diff
@@ -269,6 +269,9 @@ order to select data in a line without any trailing newline characters, one coul
 use the `[^\n]+` pattern, which is arguably a good compromise when
 balanced against the ability to select data over several lines.
 
+You can instruct the regex engine to stop matching newline characters with
+`.` by disabling the appropriate flag (`(?S)`).
```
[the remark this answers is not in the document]
lenormf: I didn't want to be too specific because flags can currently be enabled/disabled anywhere in a pattern (not just at the beginning).

I'd like the FAQ to be just pointing fingers at topics to look into, rather than instructions that might become invalid in the future (not that these specific regex details have any real chances to be).

doc/pages/faq.asciidoc
```diff
@@ -6,16 +6,16 @@
...
-but generally refers to a blow into which all of one's strength went.
+but generally refers to a strike into which all of one's strength went.
 
 === Is there going to be a Windows port of Kakoune ?
 
-As many features provided by UNIX systems would be missing, or if anything
-much less efficient on a Windows system, the incentive to porting the
-project to this operating system is pretty low.
+As many features provided by UNIX systems would be missing, or albeit much
+less efficient, on a Windows system, the incentive to porting the project
```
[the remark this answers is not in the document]
lenormf: Fixed.

doc/pages/faq.asciidoc
```diff
@@ -114,7 +114,7 @@ to a short time, e.g. 25ms: `set -sg escape-time 25`
 As a fairly compliant follower of the UNIX philosophy, Kakoune does not
 try to implement features that are best handled by separate, dedicated
 tools. Window splitting in terminals is a prime example of that
-concept, where the editor provides commands to interact with several
+concept, whereby the editor provides commands to interact with several
```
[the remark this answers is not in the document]
lenormf: Whenever in doubt, I pick the simplest version. I reverted back to "where".

doc/pages/faq.asciidoc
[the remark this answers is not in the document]
krobelus: Right, also if for example a user runs Ubuntu in WSL, then the installation instructions for Ubuntu apply, it's not really Windows anymore.

doc/pages/faq.asciidoc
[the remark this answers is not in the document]
krobelus: I agree, no need for details here so let's keep it.

doc/pages/faq.asciidoc
```diff
@@ -101,97 +206,104 @@ However, you can easily declare key mappings in your configuration file
...
-(or one of the supported multiplexers), run Kakoune in a `tmux` session,
-and simply use the `:new` command to spawn new clients as you would
-have otherwise in an X11 environment.
-
-== Why does `a` extend the current selection, but `i` leaves it untouched ?
+=== Why does `a` extend the current selection, but `i` leaves it untouched?
 
 Selections are ranges of characters whose delimiters are an "anchor" and
-a "cursor", and inserting characters is always done before the cursor in
+a "cursor", and inserting characters is always done before the anchor in
```
mawww, line 276: I believe this should remain "cursor"

lenormf: I changed that because when a user has a left-to-right selection and wonders where new text is going inserted if they hit <kbd>i</kbd>, the text will be inserted before the anchor - even if technically, <kdb>i</kbd> swaps cursor/anchor around at the moment you hit it.

The problem with this sentence is that it assumes the selections are left-to-right oriented. If the selection is right-to-left oriented, <kbd>i</kbd> won't move the cursor.

I think we should rewrite this bit entirely to take into account both the above issues. I don't know how, it challenges my skills at English.

Screwtapello: If I understand correctly, the real answer to this FAQ is something like:

- A selection is a range of characters from the "cursor" (the end you move around directly, sometimes called the head) to the "anchor" (the end that follows the previous position of the cursor, sometimes called the tail)
- Kakoune has (somewhat arbitrarily) decided that inserted text always appears to the left of the cursor
- When you enter insert mode, Kakoune ensures that each selection's cursor is at the left-hand end of the selection — because inserted text appears to the left of the cursor, it appears to the left of the selection, and is therefore not part of the selection
- When you enter append mode, Kakoune ensures that each selection's cursor is at the right-hand end of the selection, and grows the selection by one character — because inserted text appears to the left of the cursor, it appears just inside the selection and is therefore part of the selection

I don't know if there's a specific reason why inserted text always appears to the left of the cursor.

mawww: @Screwtapello I am not sure what other desirable alternative we would have:

* We could insert *after* the cursor in append mode then move the cursor by the inserted amount, thats quite equivalent to what we do now except that the cursor is always onto the last inserted character and it is a bit more stateful (insert mode behaves differently depending how we entered it).

* We could insert in front of the selection/past the end of the selection and grow it, effectively using a hidden cursor concept that gets decoupled from the selection cursor.

Not growing the selection on append would be very strange, and that would mean the inserted text progressively goes away from the selected text, requiring an additional visual cue on where the text gets inserted, this is not an issue on insert because the selection naturally moves due to the inserted text.

Did you have something specific in mind ?

Screwtapello: Sorry, I didn't mean to imply that there was some obvious better alternative that you'd missed. Kakoune's selections are minimum-length-1 for reasons I'm confident I can explain, which means that a newly-inserted character must go to the left or the right — but I wasn't confident I could explain why either choice was better than the other.

Now that I've had time to think about it a little more, I think the *real* source of Kakoune's left-bias is the fact that it's designed for Latin script (like terminals and the QWERTY keyboard layout), which is written left-to-right. So, when you insert text, the left-end of the selection moves right to make room for it. When you append text, the right-end of the selection moves right to make room for it. Kakoune's behaviour is mechanically consistent, it's just not visually consistent because of the way the Latin script works. So, here's another attempt at answering this FAQ:

> A selection is a range of characters from the "cursor" (the end you move around directly, sometimes called the head) to the "anchor" (the end that follows the previous position of the cursor, sometimes called the tail). When you type a letter in insert mode, Kakoune will always place it under the cursor (shifting any existing text to the right to make room), and then move the cursor to the right ready for the next letter.
>
> When you enter insert mode with `i`, Kakoune will ensure that each selection's cursor is at the left-hand end before you start typing, so that the text you type appears before the originally selected text. As you type, the cursor (and any following text) move right to make room, leaving the new text outside the selection.
>
> When you enter insert mode with `a`, Kakoune will ensure each selection's cursor is at the right-hand end before you start typing, so that the text you type appears after the originally selected text. As you type, the cursor (and any following text) move right to make room, leaving the new text inside the selection.
>
> Although it's visually inconsistent for appended text to wind up inside the selection and inserted text to be outside it, it's mechanically consistent.

mawww: No need to be sorry, its actually good to challenge the design and make sure we have good reasons for making things work the way they do.

Regarding the description proposal, in my mind a selection goes from the anchor to the cursor rather than the other way around because the anchor is what might remain fixed (as its name implies) and the cursor is the one moving.

lenormf: So, what do we do with the hunk?

mawww: Can we restore cursor for now until we find a better wording, "inserting characters is always done before the cursor in insert mode" might be confusing, but is not wrong, if Kakoune is in insert mode, regardless of how it got there, characters will be inserted before the cursor.

doc/pages/faq.asciidoc
```diff
@@ -101,97 +206,117 @@ However, you can easily declare key mappings in your configuration file
...
+Rendering the cursor as a beam is a common feature of other modal editors,
+it however doesn't fit within Kakoune's selection-first paradigm.
+
+There is a selection on screen at all times, containing either data selected
+by the user, or a newline character when the buffer is empty.
+
+A selection is bound by an anchor and a cursor. They can overlap, but
+ultimately must both be placed *over* a character. A beam cursor placed
+*between* two characters doesn't fulfil that requirement, and is thus
+not allowed.
```
Screwtapello, line 253: I think this is backwards: it's not that the cursor must be a block because a selection is minimum-length 1; selections are minimum-length 1 because the cursor must be a block.

The real reason the cursor must be a block is that terminals can only draw one "real" cursor (beam or otherwise) at a time, so when Kakoune has multiple selections, it has to draw them itself by setting foreground/background colours. Because a custom background colour must cover at least one character, therefore the Kakoune selection model has minimum-length 1.

lenormf: There always has to be at least one character selected, not because of a technical detail related to colours or the terminal, but because of the selection-first paradigm.

0-length selections simply don't make sense within the editing language.

doc/pages/faq.asciidoc
```diff
@@ -1,44 +1,48 @@
...
 
 The name of the project is pronounced "Kak-oon", and is a word taken from a
 New Caledonian dialect based on French. It means a hard blow, usually a punch,
-but generally refers to a blow into which all of one's strength went.
+but generally refers to a strike into which all of one's strength went.
+
+=== Is there going to be a Windows port of Kakoune ?
 
-== Is there going to be a Windows port of Kakoune ?
+As many features provided by UNIX systems are missing or less efficient on
```
pickfire, line 13:
```suggestion
No, as many features provided by UNIX systems are missing or less efficient on
```

lenormf: The FAQ doesn't claim there will not be a Windows port in the future, only that the incentive is low.

Answering the question with a "no" would be less future-proof that what we have currently, so I don't think we should apply this hunk.

PR #3552, opened 2020-06-11
rc lint: Utilize toolsclient if defined

caksoylar:
A small change to open the `*lint-output*` buffer in `toolsclient` if defined, similar to `make` and `grep` tools. Since the buffer is already of `make` filetype it utilizes `jumpclient`.

### DIFF
--- a/rc/tools/lint.kak
+++ b/rc/tools/lint.kak
@@ -17,6 +17,9 @@ declare-option -hidden line-specs lint_messages
 declare-option -hidden int lint_error_count
 declare-option -hidden int lint_warning_count
 
+declare-option -docstring "name of the client in which utilities display information" \
+    str toolsclient
+
 define-command \
     -hidden \
     -params 1 \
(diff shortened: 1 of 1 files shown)


rc/tools/lint.kak
```diff
@@ -46,9 +49,8 @@ define-command \
         dir=$(mktemp -d "${TMPDIR:-/tmp}"/kak-lint.XXXXXXXX)
 
         # A fifo to send the results back to a Kakoune buffer.
-        # FIXME: Should we put the lint output in toolsclient?
         mkfifo "$dir"/fifo
-        printf '%s\n' "evaluate-commands -draft %{
+        printf '%s\n' "evaluate-commands -try-client '$kak_opt_toolsclient' %{
```
lenormf, line 53: I think the `-draft` should stay. The primary way of reading errors is via flag lines set on the buffer itself, I don't want the focus to change to another buffer every time I call `:lint`.

caksoylar: I wasn't sure if it is possible to use both `-draft` and `-try-client`, since both are specifying a client context. Also, I don't get a buffer change when I use it without a `toolsclient` definition (which is a bit strange when I think about it), but let me know if it is happening for you.

lenormf: It seems that you are correct, using `-try-client` doesn't change the focus. I always thought it did. Sorry for the noise!

PR #3561, opened 2020-06-23
Syntax highlighting for Standard ML

ii8:
I also have highlighters for [bale](/ii8/bale), wasn't sure if you'd want it since only a few people use bale so let me know.

### DIFF
--- a/rc/filetype/mlb.kak
+++ b/rc/filetype/mlb.kak
@@ -1,0 +1,50 @@
+# http://mlton.org/MLBasis
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+# Detection
+# ‾‾‾‾‾‾‾‾‾
+
+hook global BufCreate .*\.mlb %{
+    set-option buffer filetype mlb
+}
+
+# Initialization
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+hook global WinSetOption filetype=mlb %{
+    require-module mlb
+    set-option buffer extra_word_chars '_' '-' '.'
+    set-option window static_words %opt{mlb_static_words}
+}
+
+hook -group mlb-highlight global WinSetOption filetype=mlb %{
+    add-highlighter window/mlb ref mlb
+    hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/mlb }
+}
+
+provide-module mlb %[
+
+# Highlighters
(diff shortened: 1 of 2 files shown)


rc/filetype/mlb.kak
```diff
@@ -0,0 +1,50 @@
...
+add-highlighter shared/mlb/string region '"' '(?<!\\)(\\\\)*"' group
+add-highlighter shared/mlb/string/fill fill string
+add-highlighter shared/mlb/comment region -recurse '\(\*' '\(\*' '\*\)' fill comment
+
+evaluate-commands %sh{
+    keywords='basis bas and open local let in end structure signature functor ann'
+
+    join() { sep=$2; eval set -- $1; IFS="$sep"; echo "$*"; }
+
+    printf %s\\n "declare-option str-list mlb_static_words $(join "${keywords}" ' ')"
```
lenormf, line 41: We generally try to be as conservative as possible with new options, why not just set `static_words` directly, instead of storing keywords in `mlb_static_words`?

rc/filetype/sml.kak
```diff
@@ -0,0 +1,86 @@
...
+              sharing sig signature struct structure'
+    types='unit exn ref'
+    ops='before ignore o
+         div mod quot rem abs
+         not chr ord ceil floor round trunc
+         andb orb xorb notb'
+
+    join() { sep=$2; eval set -- $1; IFS="$sep"; echo "$*"; }
+
+    printf %s\\n "declare-option str-list sml_static_words $(join "${keywords} ${types} ${ops}" ' ')"
```
lenormf, line 48: Same remark as above.

rc/filetype/mlb.kak
ii8: I was cargo culting, most of the other filetypes do it like this.

lenormf: Hmm you're right, apparently commit 80ac46e09de0a30c0bf590fd50b6cbf9bfd1c41b introduced options for all files…

Ignore me, then!

lenormf:
Feel free to submit highlighters for filetype that might not make the cut for this repository to https://github.com/kakoune-editor/kakoune-extra-filetypes.

rc/filetype/mlb.kak
ii8: In fact, the ones that don't set an option(fsharp and awk for example) like this don't actually work because the hook is created after the window already has the `filetype=...` option set.

PR #3569, opened 2020-06-28
Fix 'error: non-constant-expression cannot be narrowed...'

1g0rb0hm:
Fix **clang** based builds that raise the following error:

```
ncurses_ui.cc:759:59: error: non-constant-expression cannot be narrowed from type 'unsigned int' to 'Key::MouseButton' in initializer list [-Wc++11-narrowing]
                return mouse_button(mod, Key::MouseButton{code}, coord, c == 'm');
                                                          ^~~~
```


mawww:
Thanks.

PR #3574, opened 2020-07-01
expansions.asciidoc: Add an example of file expansion.

Screwtapello:
Properly speaking, all expansions are of the form "%abc{}" where "abc" is the expansion type. All expansion types are listed in the documentation, as they should be. However, when searching through the documentation, a keyword like "file" is likely to hit a lot of false-positives, so a smart user is likely to search for the more-unique "%file" instead. Unfortunately, unlike every other expansion-type, file-expansions did not include an example, so a search for "%file" resulted in no hits, potentially leading people to believe it was undocumented.

Now there's an example, so people searching for "%file" will find what they're looking for.

### DIFF
--- a/doc/pages/expansions.asciidoc
+++ b/doc/pages/expansions.asciidoc
@@ -171,6 +171,13 @@ Kakoune uses a shell command, such as the `|`, `!` or `$` normal mode commands
 Expansions with the type `file` will expand to the content of the filename
 given in argument as read from the host filesystem.
 
+For example, this command stores the entire content of `/etc/passwd` into the
+`a` register.
+
+----
+set-register a %file{/etc/passwd}
+----
+
 == Value expansions
 
 Expansions with the type `val` give access to Kakoune internal data that is


Screwtapello:
Admission: Earlier today I wanted to link somebody to the `%file{}` docs, so I searched for `%file`. When I didn't find it, I decided to make a PR documenting it, only to find it had actually been documented all along. So this is as much for the benefit of my laziness as anybody else. :)

alexherbo2:
How to handle arbitrary filenames?

Screwtapello:
In some circumstances (such as piping to `kak -p`) you can use NUL as the argument delimiter. Since NUL cannot legally appear in a filename, it's safe to use it to delimit filenames. However, bash doesn't like NUL bytes appearing in command-substitutions, and I can't figure out how to type NUL directly in bash, or in Kakoune, so it's not a universally-applicable solution.

alexherbo2:
@Screwtapello It could be nice to add an example with a filename from an option.

Screwtapello:
It would, but off the top of my head I'm not even sure how to go about it. I suspect it would be too complex to make a good example.

PR #3580, opened 2020-07-06
Fix syntax errors.

SolitudeSF:

PR #3583, opened 2020-07-07
Fix missing whitespace from improved windowing module detection.

sjjf:
A space was missing from some of the tests for a zero length `windowing_modules` option, making them invalid shell code.

This has been tested with kitty, tmux, screen and x11 environments.


krobelus:
Same as #3580

sjjf:
Yeah, you're right. I'll close this, then.

PR #3589, opened 2020-07-09
[doc] Escape asciidoc meta-character

occivink:


doc/pages/regex.asciidoc
```diff
@@ -38,8 +38,8 @@ class is negated, meaning that it matches every character not specified
 in the character class.
 
 Literals match themselves, including syntax characters, so `^`
-does not need to be escaped in a character class. `[*+]` matches both
-the `*` character and the `+` character. Literal escape sequences are
+does not need to be escaped in a character class. `[\*+]` matches both
```
mawww, line 41: I am surprised that we need that inside code blocks.

occivink: This spurious emboldening occurs with both the github viewer and the kakoune renderer, so I'm assuming that it is compliant with the asciidoc spec

PR #3594, opened 2020-07-10
Disable zsh autotitle

sucrecacao:
Zsh has a setting to set the title automatically and it is enable by default. To make zsh work with this script without having the user change they zsh config, we have to disable it by setting `DISABLE_AUTO_TITLE=false` .

### DIFF
--- a/rc/windowing/repl/x11.kak
+++ b/rc/windowing/repl/x11.kak
@@ -18,8 +18,10 @@ define-command -docstring %{
         fi
         if [ $# -eq 0 ]; then cmd="${SHELL:-sh}"; else cmd="$@"; fi
         # The escape sequence in the printf command sets the terminal's title:
-        setsid ${kak_opt_termcmd} "printf '\e]2;kak_repl_window\a' \
+        setsid ${kak_opt_termcmd} "DISABLE_AUTO_TITLE=false \ # disable zsh auto title
+        	&& printf '\e]2;kak_repl_window\a' \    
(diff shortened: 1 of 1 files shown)


sucrecacao:
If the user doesn't use zsh this might create side effects even if its really unlikely. 
Maybe it should detect zsh first ?

Screwtapello:
I don't understand how zsh is involved. As I understand it, if I use Kakoune's REPL support to launch a process like Python, I expect the process tree to look something like "kakoune → xterm → python", with maybe an extra "setsid" or a non-interactive "/bin/sh" for plumbing purposes. My login shell wouldn't be involved at all.

sucrecacao:
I could come up with a least hree reason:
1) you might want a bash repl inside zsh 
2) you might want to start with bash so you can use your fancy zsh config to easily jumps to a dir and then start a repl  
3) For all the new user using zsh ( which is not a negligible minority ) who want to try the windowing feature. They will try to run `:repl` and `:send-text` and will see that it doesn't work, without any feedback and believe kakoune is broken. 

In a more broader way: the x11 REPL rely on windows title, which is risky: zsh, other shell or terminal emulator can change the title. 
Maybe a better solution would be to catch if we fail to send text to the REPL, and  return an explicit message with maybe a link to a troubleshooting page.

mawww:
I think we need to change the x11-repl logic not to use the window title. Most x11 terminal emulator expose a `$WINDOWID` env-var, I think using that would be a better approach.

PR #3596, opened 2020-07-11
Highlight markdown code block with space

pickfire:


rc/filetype/markdown.kak
```diff
@@ -81,6 +81,9 @@ add-highlighter shared/markdown/inline/text/ regex ^\[[^\]\n]*\]:\h*([^\n]*) 1:l
 add-highlighter shared/markdown/inline/text/ regex ^\h*(>\h*)+ 0:comment
 add-highlighter shared/markdown/inline/text/ regex \H\K\h\h$ 0:PrimarySelection
 
+# Inline code
+add-highlighter shared/markdown/inline/text/ regex "^    .*" 0:meta
```
krobelus, line 85: The comment should refer to [indented code blocks](https://spec.commonmark.org/0.29/#indented-code-blocks).

The current regex highlights too much. Consider:

```markdown
block:

    code

not code
```

The highlighting leaks to `not code`.

I suggest this alternative where we also support tabs in addition to four-space indents:

    add-highlighter shared/markdown/inline/text/ regex "^\n(?:(?:\t| {4})[^\n]*\n)+" 0:meta

It does require an empty line before a code block, just like CommonMark.
However, CommonMark allows to [omit the empty line after another kind of block, such as headlines](https://spec.commonmark.org/0.29/#example-85). Maybe we should just not require an empty line before the code block (dropping the first `\n` from the regex). This would work well if indented lines are meant to be code blocks in most Markdown files out there. Except for lists, they are already highlighted correctly.

pickfire: > The highlighting leaks to not code.

This is fixed already. The original version (`^    .*$`) does highlight non code. I don't see why it is an issue.

krobelus:
```sh
~/kakoune ((b3c8d419…))> printf '
block:

    code

not code' | HOME=$PWD ./src/kak -e 'set buffer filetype markdown'
```

should not highlight the last line

pickfire: Yeah, it didn't.

PR #3606, opened 2020-07-17
gruvbox: invert colors for info box

maximbaz:
I've been using `gruvbox` colorscheme for a long time, and one thing that always bothered me is that while the theme is generally dark, the amount of "white" color that info box produces is simply blinding, especially at night.

I inverted the colors and was really happy with the result, so I propose we do it for everyone.

Before:

image

After:

image

### DIFF
--- a/colors/gruvbox.kak
+++ b/colors/gruvbox.kak
@@ -60,7 +60,7 @@ evaluate-commands %sh{
         face global MenuForeground     ${bg2},${blue}
         face global MenuBackground     ${fg},${bg2}
         face global MenuInfo           ${bg}
-        face global Information        ${bg},${fg}
+        face global Information        ${fg},${bg}
         face global Error              ${bg},${red}
         face global StatusLine         ${fg},${bg}
         face global StatusLineMode     ${yellow}+b


mawww:
This breaks inline info boxes (like the ones we get as completion documentation), as they are not distinguishable from buffer text. I think we still need a different background from the default one, but I am happy to change the look of info boxes to be darker.

maximbaz:
Good catch! This is especially noticeable on a small 1-2 line inline info-boxes...

Before:

image

After:

image

I tried different grey colors that are already used in this colorscheme, the darkest one that has good enough contrast is this:

image

But regular info boxes don't look as slick anymore 😞 (better than before, still!)

Have you considered drawing borders for inline info-boxes just like you do for regular info-boxes? If we had borders, we would gain consistency and more possibilities to color info boxes, because a choice that works well would work well in all places, inline or not 🙂 

Just a couple of examples to demonstrate:

image

image

mawww:
Adding borders to inline boxes would make them bigger (by two lines and two columns), which I'd like to avoid. Ideally we could add some kind of border at the character edge which would allow for a similar effect without consuming a row/column. That might be possible with recent terminals, but its a longer term project.

mawww:
Gave the new version a try, I remain a bit unhappy about that lack of visual distinction between menus and info boxes that we have now, I'd prefer to use a distinct background for those, as the foreground is a bit harder to control (we can have arbitrary formating in both menus and info boxes).

maximbaz:
> Adding borders to inline boxes would make them bigger (by two lines and two columns), which I'd like to avoid.

Could you explain a bit more why you would like to avoid it? Is it because some info boxes will no longer fit the screen? Or will begin to wrap and take even more space? I might be missing something but I can't think of anything that will make using extra two lines and two columns too terrible (aside from just desire not to waste space, but the same argument can be used for the regular info box which already does it).

> Ideally we could add some kind of border at the character edge which would allow for a similar effect without consuming a row/column.

This is an interesting idea, could also be applied for regular info boxes then! But to be honest I couldn't find anything on how to make it possible, I imagine it would require drawing two characters in a cell... Do you have any pointers for me to investigate? Or do you suggest we drop this for now?

 > Gave the new version a try, I remain a bit unhappy about that lack of visual distinction between menus and info boxes that we have now, I'd prefer to use a distinct background for those, as the foreground is a bit harder to control (we can have arbitrary formating in both menus and info boxes).

To be honest I'm not satisfied as well but for different reasons, and personally I'd rather stick with black background (the first commit) and disable inline info boxes (I can only see them in `kak-lsp` I think).

If you prefer distinct background, does that mean that even if we had a possibility for borders for inline info boxes, you would still prefer not to have a black background for them? If so, shall we abandon this PR altogether?

By the way I initially picked the idea for using the same background for info box and buffer from @Delapouite's screenshots in his [kakoune-colors](https://github.com/Delapouite/kakoune-colors/) repo where I think info boxes look very slick as well. The actual color schemes now use different background for info boxes, so they don't reflect screenshots anymore. @Delapouite out of curiosity, what would you think if inline info boxes had borders?

mawww:
Using `ctags-func-info` to display inline signature info  hides much more context if we add borders, also it makes it impossible to correctly anchor the info text (to align the info text with buffer text) on the first character of a line because we'll always get the bordere character + the following padding space there.

capture
capture2

Regarding terminal assisted borders, I was referring to possible support for "SGR 51 (framed)" in terminals, but that does not seem to be available in any widely used ones.

lobre:
Hello,

If I understand the problem correctly, it seems we want better design/contrast of info boxes, but without wasting space for anchored ones.

> By the way I initially picked the idea for using the same background for info box and buffer from @Delapouite's screenshots in his [kakoune-colors](https://github.com/Delapouite/kakoune-colors/) repo where I think info boxes look very slick as well.

As @maximbaz stated, I also find that the info box having the same background as the terminal background is really nice visually speaking. But I also share your concern @mawww with overall spaces wasted if we implement the borders on small anchored boxes. I mainly use the automatic signature that pops up from `ctags`. It is a nice small and effective feature that makes a difference to me.

I am not sure that terminal-assisted borders will be generally available for all terminals soon. And I think that Kakoune should stay robust and work consistently on all terminals.

To still answer the initial concern, I would suggest adding a face–such as `InformationAnchored`–that would apply when drawing anchored info boxes and that would default to the style of `Information` (for compatibility purposes). This specific face could be overridden by colorschemes to allow:

- keeping the main info boxes background-less (same as in [kakoune-colors](https://github.com/Delapouite/kakoune-colors/)),
- making small anchored info boxes have a contrasting background to differentiate them from the terminal background without having to use proper borders.

It feels to me that it is a simple solution that does not add much to the codebase while still responding to the need.

In terms of implementation, I did not look at the code, but I guess there is already a condition that exists to decide whether to print the borders on not, so applying a different face for one of the use cases does not seem to be too many changes.

What do you think?

mawww:
Sounds like a reasonable approach, I am usually cautious about adding extra faces, but here we do distinguish between inline/non inline info boxes through borders, as you mentioned, so adding an `InlineInformation` face could make sense.

One potential issue is that we have more than two case here, internally we have the InfoStyle enum that can be Prompt, Inline, InlineAbove, InlineBelow, MenuDoc and Modal. I would like to avoid having a different face for each. I think we could have Inline* and MenuDoc be considered Inline, while Prompt and Modal would not.

lobre:
I do like your restrictive approach to keep kakoune as bloat-less as possible. And this is really what attracts me in kakoune, compared to a codebase such as vim, or helix. From the outside, kakoune’s codebase feels simple, and well thought (and it is amazing that it is dependency-free). I strive for similar goals on my personal project, due to past experiences where projects explode in complexity because many don’t understand that adding features has a high price on maintainability.

That being said, what becomes difficult is to assess when things should really be added. I think your idea of adding an `InlineInformation` face makes sense. Discoverability is at the core of kakoune’s principles. Improving info boxes by having them visually attractive seems going in the correct direction.

As I am not familiar with the C++ codebase, I did not know about those other kinds of boxes. But from the outside, it seems valid to categorize them into inline and non-inline.

If you think that adding this feature is reasonable, I can take a look at it as a "first contribution". Maybe I am wrong, but it looks like a rather "simple" one. I don’t know if @maximbaz is also still interested in this, as the pull request has been created a while ago, but I guess he could also assist me as he appears to have already contributed in the past. Let me know!

maximbaz:
I'm happy to help out with testing, to be honest I still use the first commit in this PR, together with inline help disabled, I still find it most pleasing to the eye in terms of colors and getting the most context around the cursor, especially on very wide help texts.

<details>
<summary>Screenshot</summary>

image

image

</details>

<details>
<summary>&lt;offtopic&gt;Now that you mention helix, I'm so much envious about their syntax highlight in the help info boxes 😭 &lt;/offtopic&gt;</summary>

image

</details>

lobre:
If you have background and foreground reversed, that means you try to make sure to avoid anchored boxes @maximbaz?

So I guess that for example, you have `set global lsp_hover_anchor false` for kak-lsp? How do you do for lsp completions? I think there is no configuration to make them non-inline...

I do personally use this hack at the moment:

```
face global Information default+g

# set background during lsp completions popups
hook global InsertCompletionShow .* %{ face global Information default,black+g }
hook global InsertCompletionHide .* %{ face global Information default+g }
```

As for helix, I agree that syntax is great in boxes. They also seem to add an empty space as padding inside boxes. Same as a border, this is a waste of space but it is nice looking. By the way, this topic is getting discussed at https://github.com/kak-lsp/kak-lsp/issues/588.

Before moving forward and trying to implement the `InlineInformation` face (that would avoid my above hack), I will wait for @mawww’s approval.

maximbaz:
> So I guess that for example, you have `set global lsp_hover_anchor false` for kak-lsp?

Yeah! (it's `false` by default)

> How do you do for lsp completions? I think there is no configuration to make them non-inline...

The completions already use a different background color though, so that "magically" works, without any hooks... 🙂 Or did I misunderstand what you refer to?

image

Same in helix btw:

image

lobre:
> The completions already use a different background color though, so that "magically" works, without any hooks... slightly_smiling_face Or did I misunderstand what you refer to?

I think we don’t talk about the same thing. Effectively the completions seem to be impacted by the `MenuBackground` face. That’s why it "magically" works on your screenshot. However, I was talking about the inline box that should normally appear next to the completion box, showing "hover" information. I don’t know why you don’t have them on your screenshot, as I did not find a way to disable them on my side ^^.

See when I have a transparent background:

transparent_bg

This menu does not render well inline:

transparent_inline

In comparison, see when I apply a grey background to `Information` boxes:

grey_bg

And inline:

grey_inline

I see that @mawww did not give his opinion on whether I should go ahead or not with the `InlineInformation` face, but I will maybe start looking at it anyway.

maximbaz:
Ahhh I see what you mean, yes of course I reproduce the problem!

image

mawww:
I think InlineInformation makes sense, we can default it to Information.

lobre:
I am back in Kakoune (after a Neovim journey), and it has been a while, but I am still interested in this `InlineInformation` face.

I am trying to explore the Kakoune codebase, but as a newcomer, I would appreciate it if you have guidance on where this change should be applied in the code (maybe @maximbaz or @mawww).

I have seen that the `info` command will call the `info_show` function of the client.

https://github.com/mawww/kakoune/blob/871631bb00dc4e531f0490927d5d7a73209cd249/src/commands.cc#L2381

Then, it will set an empty `Face{}`.

https://github.com/mawww/kakoune/blob/871631bb00dc4e531f0490927d5d7a73209cd249/src/client.cc#L472

And this `info_show` function seems to set global variables (`m_info` and `m_ui_pending`) to indicate that an infobox should be displayed.

I just want to know if I am digging at the correct place, and if changing this `Face{}` is the way to go.

mawww:
I think the change should be on src/client.cc line 281:
`face["Information"]` should become `face[m_info.style == InfoStyle::Inline ? "InlineInformation" : "Inline"]` and "InlineInformation" should be added to face registry as a default face (possibly defaulting to default,default@Information)

lobre:
Thank you for guiding me through this change.

When you say `possibly defaulting to default,default@Information`, I am not sure to understand what "defaulting" means here. In `src/face_registry.cc`, I don't see any way of defaulting to something else.

Does it mean I just have to set the same value as `Information`? Such as:

```
         { "Information", {Face{ Color::Black, Color::Yellow }} },
+        { "InlineInformation", {Face{ Color::Black, Color::Yellow }} },
```

mawww:
Something like this:
```diff
         { "Information", {Face{ Color::Black, Color::Yellow }} },
+        { "InlineInformation", {Face{}, "Information"} },
 ```
With this, changing Information will change InlineInformation as well, until InlineInformation is defined differently.

lobre:
Thanks again, I have submitted a PR.

PR #3613, opened 2020-07-22
fix shell fi always deindenting after newline

pickfire:
Always align fi to the line before before deindenting it
```
    ...
        test
    fi%( )
```
always deindent fi on `o`
I feel lazy to add tests for this.

### DIFF
--- a/rc/filetype/sh.kak
+++ b/rc/filetype/sh.kak
@@ -128,7 +128,7 @@ define-command -hidden sh-indent-on-new-line %[
         # indent after then
         try %{ execute-keys -draft <space> k <a-x> <a-k> then$ <ret> j <a-gt> }
         # deindent after fi
-        try %{ execute-keys -draft <space> k <a-x> <a-k> fi$ <ret> <a-lt> j K <a-&> }
+        try %{ execute-keys -draft <space> k <a-x> <a-k> fi$ <ret> K <a-&> j <a-lt> j K <a-&> }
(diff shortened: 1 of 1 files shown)


krobelus:
This is awesome! The fact that `fi` was indented incorrectly after `o` and `O` has irritated me for some time.

pickfire:
This may break someone's behavior for those that write `fi` before they write the function body, but this improves a lot although it is still not good.

Something better would be to reindent when they type fi and reindent when they changed fi to something else, like how fish shell does indentation, not sure if the second thing is possible since we don't know the previous state.

@mawww do we get to track the previous word or state changed? Like to know if the word was `fi` before it was changed.

> This is awesome! The fact that fi was indented incorrectly after o and O has irritated me for some time.

Ah, just irritated you? I cannot do stuff at all since it broke my workflow.

eraserhd:
This has bothered me for a long time!

It also happens for me with do/done.

I think I just notice that I always close brackets and ending keywords, then insert between them.  I just noticed because I'm pairing with someone who doesn't, and it is absolutely terrifying.  It's like a psychological thriller to watch all those open constructs just ... hanging there, incomplete, like knives.

pickfire:
Both python, typescript and html indentation method is not so useful for me, sometimes I find it annoying. This shell indentation really broke me to not being able to do anything without sending in a patch. I sent some patches for rust indentation so it improves a lot now. Maybe next, I would fix the python indentation, not sure how to handle this shell indentation.

Issue #3618, opened 2020-07-26
kakoune daemon without forking?


theamazingwaffle:
I have `kak -d -s "$KAKSESSION"` run on login, but when i log out, the forked process doesn't get killed, but i can't access the old session either. I have to `killall kak` because i don't see the process id kakoune gives me, i go straight into xorg. And i don't want to manually kill the demon anyway. I think a `--no-fork` flag would be nice so the users can decide what to do with the headless devil themselves.

```sh
# instead of
kak -d -s "$KAKSESSION"
# do
kak -d -s "$KAKSESSION" &
# now on logout the shell will kill the kak daemon, i think.
```

johnisom:
A potential workaround for your situation is to use `kak -l` to list the active sessions, and for each one, execute `echo kill | kak -p <session_name>`. To kill all sessions you could use this:

``` bash
for session in $(kak -l); do
  echo kill | kak -p $session
done
```

And if you want to kill sessions that have modified & unsaved buffers, use this variant:

``` bash
for session in $(kak -l); do
  echo kill! | kak -p $session
done
```

alexherbo2:
More interactively, you can do `kak -c "$KAKSESSION"` to reconnect to the session, then `kill` or `kill!`, like mentioned above.

johnisom:
I realize that my comment is definitely not the answer you're looking for, and I do think a `--no-fork` option would be nice. Though depending on your specific situation, I think using `killall kak`, my proposed solution, or something similar would work nicely. 

To clarify, when you say "log in" and "log out", are you referring to logging in and out from your display manager? Was the code block you provided part of your `.xprofile`/`.xinitrc`? If so, then using `--no-fork` and running in the background with `&` does indeed make the most sense.

theamazingwaffle:
>To clarify, when you say "log in" and "log out", are you referring to logging in and out from your display manager? Was the code block you provided part of your .xprofile/.xinitrc? If so, then using --no-fork and running in the background with & does indeed make the most sense.

The code is in my `.profile` which is executed when i log into the tty.

johnisom:
> > To clarify, when you say "log in" and "log out", are you referring to logging in and out from your display manager? Was the code block you provided part of your .xprofile/.xinitrc? If so, then using --no-fork and running in the background with & does indeed make the most sense.
> 
> The code is in my `.profile` which is executed when i log into the tty.

[...]

PR #3619, opened 2020-07-26
Add a flag to not fork when running the server

theamazingwaffle:
#3618
Like @eraserhd said, there are cases where a non-forking variant of the server are required. Furthermore, kakoune does a job that it isn't supposed to - manage processes. Calling `fork()` is unnecessary in this case and this functionality can be easily replaced by any other tool that is better at managing processes, like the init system of choice or another program that focuses specifically on managing processes.


theamazingwaffle:
@mawww fixed this

Issue #3617, opened 2020-07-25
Deindent closing brace when after cursor


johnisom:
### Steps 

1. Open any file that uses braces for blocks (`touch main.c; kak main.c`)
2. Insert opening and closing brace (`ivoid failed_indent(void) {}<esc>`)
   ```
   void failed_indent(void) {}▌
   ```
3. Place cursor in-between braces (`h`)
   ```
   void failed_indent(void) {▌}
   ```
4. Insert new line (`i<return>`)

### Outcome 

(`▌` = cursor)

```
void failed_indent(void) {
    ▌}
```

### Expected 

(`▌` = cursor)

```
void failed_indent(void) {
▌}
```

### Description

This issue is properly described in the discussion of this issue:
alexherbo2/auto-pairs.kak#39

> It looks to be the responsability of the [filetypes](https://github.com/mawww/kakoune/blob/master/rc/filetype) to produce a correct indenting when pressing Return.

> I’m afraid all filetypes are affected to not handle indenting in-between content, assuming new lines are inserted at the end of the line.

...

> It is hard, because all [filetypes](https://github.com/mawww/kakoune/blob/master/rc/filetype) are wrong. I feel like if I fix it in auto-pairs, upstream will never get fixed.

> We just need the [filetypes](https://github.com/mawww/kakoune/blob/master/rc/filetype) to pass from:

> ```
> int main() {
>   ▌}
> ```

> To:

> ```
> int main() {
> ▌}
> ```

> Then auto-pairs can do what we would do interactively: Return—Escape—O.

...

> If you feel courageous, fix all `<language>-new-line-inserted` commands in [filetypes](https://github.com/mawww/kakoune/blob/master/rc/filetype).

Is this a valid issue? It seems that the correct behavior would indeed entail what is described.

krobelus:
Yeah I think we should handle this.
There is some nice documentation on how indentation after braces is implemented in `doc/autoedit.asciidoc`.
Here is a patch that should make this work for C, Rust and Go. There are plenty of other languages that use the same logic - grep for `indent after lines? ending with`. We probably want to change all of them. Let's make sure this patch works in all reasonable scenarios.

```diff
diff --git a/rc/filetype/c-family.kak b/rc/filetype/c-family.kak
index 837c5059..1b430074 100644
--- a/rc/filetype/c-family.kak
+++ b/rc/filetype/c-family.kak
[...]

PR #3620, opened 2020-07-27
rc: Deindent closing brace when after cursor (all filetypes)

johnisom:
Fixes issues described in #3617.

Example (▌ = cursor):

``` c
void foo(void) {▌}
/* currently on <ret> becomes: */
void foo(void) {
    ▌}
/* but with my fix becomes: */
void foo(void) {
▌}
```


alexherbo2:
Will this fix apply to all filetypes?

johnisom:
Yes, that's the indented scope of this PR.

johnisom:
I realized my specific implementation will fail in the case of bad indentation:

``` c
  void foobar(void) {
         foobar();▌}
/* which becomes */
  void foobar(void) {
         foobar();
       ▌}
/* but should be */
  void foobar(void) {
         foobar();
  ▌}
```

This should be a fairly easy fix: I just need to make it so that when deindenting, it doesn't use `<a-lt>` but rather finds the matching brace (`m`) and indent to the same level as the line that contains the opening brace.

johnisom:
I haven't gotten to clojure.kak or lisp.kak. Since it's been some time since I've worked with Lisp syntax, I'll probably get to them last. If anyone wants to volunteer to handle them, though please do and submit a pull request on johnisom/kakoune (on the `deindent-closing-braces` branch).

johnisom:
Just for context and if people don't want to read the actual commits, here's the code that's being added to each file, tailored to the specific language:

``` kakrc
# ...
try %< execute-keys -draft <a-x> <a-k> ^\h*[})\]] <ret> gh / [})\]] <ret> m <a-S> 1<a-&> >
# ...
```
Let's break it apart:
| Keys | Explanation |
|---|---|
| `<a-x>` | Select whole line, placing cursor at end of line |
| `<a-k> ... <ret>` | Keep selections that match regexp (in this case, throws error if current single selection does not match, ending execution early) |
| `^\h*[})\]]` | Regex that matches the beginning of a line (`^`) followed by zero or more white spaces (tabs, spaces, etc) (`\h*`) and one of: `}`, `)`, or `]` |
| `gh` | Move cursor to beginning of line, reducing selection to size 1 |
| `/ ... <ret>` | Find and select the first occurrence of a regexp |
| `[})\]]` | Regex that matches one of: `}`, `)`, or `]` |
| `m` | Move cursor to matching brace (matching brace of `}` is `{`), extending selection to encompass whole area in-between |
| `<a-S>` | Split selection into 2 selections of length 1 at beginning and end of original selection |
| `1<a-&>` | Indent line that primary selection (closing brace) resides on to the level of indentation that the secondary selection (opening brace) resides on |

johnisom:
Now that all filetypes have been changed, as well as some test cases added, I'm taking this out of draft (WIP) stage! There are still a few tests to be added, though no tests for these languages currently exist outside the scope of this PR:

- javascript
- awk
- cabal
- css
- cue
- dart
- d
- gluon
- java
- json
- kakrc
- latex
- nix
- perl
- php
- ragel
- python
- protobuf
- sass
- scala
- sh
- zig

This PR is now open for review and merge!

mawww:
Wow, thanks a lot for this through PR, it reminds me that now that we have modules it might make sense to have a indentation utilities one to share the indentation logic and give them nice names, but thats a cleanup for later (assuming it proves to work well and not impact performance too much).

PR #3622, opened 2020-07-28
Fix missing quotation marks.

leira:


doc/pages/expansions.asciidoc
```diff
@@ -48,9 +48,9 @@ for full details). For example:
 
 * `echo "x%val{session}x"` echoes the current session ID, surrounded by `x`
 
-* `echo %{%val{session}}"` echoes the the literal text `%val{session}`
+* `echo "%{%val{session}}"` echoes the the literal text `%val{session}`
```
krobelus, line 51: I think both these quotes are not necessary.

`%val{session}` is already quoted by being surrounded by `%{` and `}`.

Screwtapello: Yeah, the point of this example is to show that a single layer of %{} quoting will prevent the contents from being expanded. This is a case of too *many* quotation marks, not missing quotation marks.

doc/pages/expansions.asciidoc
```diff
@@ -48,9 +48,9 @@ for full details). For example:
...
 
-* `echo %sh{ echo %val{session} }"` echoes the literal text `%val{session}`
+* `echo "%sh{ echo %val{session} }"` echoes the literal text `%val{session}`
```
Screwtapello, line 53: Likewise, this example is supposed to show that a single layer of %sh{} will prevent the contents from being expanded. The extra double-quote should be removed.

Screwtapello:
Thank you very much for making this PR! I'm sorry the documentation you were looking for was misleading, but I appreciate your effort in helping to clean it up.

Issue #3591, opened 2020-07-09
Be verbose about requirement 


sucrecacao:
Too often I try a command and nothing happen, like in #3590 . When I dig in the code I realise I miss a dependency. Sometime it will be explicit in the readme, the doc, or commented on top of a file, but sometime not. 

How could we introduce a way to be more verbose about requirement ? The fact that nothing happen at all instead of an error can be very confusing for new users ( and old one too ! ) ,   #3590  is an example that just happen to me, be it's not the first time.

Maybe just ask contributor and pluggin writter to had something like: ( pseudocode)
``` 
if  ! exist(xsel):
    echo " xsel required"
```

krobelus:
usually there should be a "command not found" error in the debug buffer

mawww:
I think the pattern should be:

```
if ! <my command line that requires a tool that might be missing>; then
    echo "fail failed to run my command, see *debug* buffer"
fi
```

We would get the `command not found`  error from the shell or stderr from the command itself in the *debug* buffer.

PR #3628, opened 2020-07-28
rc/x11: failure message when some external tools are missing

krobelus:
Addresses #3591

Thanks to @sucrecacao for reporting.

### DIFF
--- a/rc/windowing/repl/x11.kak
+++ b/rc/windowing/repl/x11.kak
@@ -23,9 +23,11 @@ define-command -docstring %{
 }}
 
 define-command x11-send-text -docstring "send the selected text to the repl window" %{
-    nop %sh{
-        printf %s\\n "${kak_selection}" | xsel -i
-        xdotool search --name kak_repl_window key --clearmodifiers Shift+Insert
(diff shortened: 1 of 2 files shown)


lenormf:
Related #1726

krobelus:
Thanks for the pointer! From https://github.com/mawww/kakoune/pull/1726#issuecomment-346923163

> If this PR has to be denied, can I make another-one that removes the check in go-tools.kak to keep things consistent? After all that shell scope was the only thing that prompted this PR.

I'm running the default configuration where all `rc` files are loaded and the `go tools` warning has confused me repeatedly.
So I think it could be nice if that warning was not shown at startup, especially for users who don't use Go. We could do the check for the go tools lazily the first time one of the go-tools commands is used.

sucrecacao:
Maybe the error message could be more explicit like:
echo 'fail x11-send-text: failed to run xdotool... is xdotool installed ? see *debug* buffer for details'
Because one could think `xdotool` or `xsel` are part of kakoune ( and blame it for being buggy )  and not an external tool you need to install.

PR #3634, opened 2020-08-01
rc: Fix Ruby indent/insert hooks

johnisom:
The intended behavior that already exists is that when `else`/`elsif`/`when`/etc is typed, it is immediately deindented. Currently, because of a slight error of implementation, none of it works. This fixes that and makes Ruby a joy to work with in Kakoune.

### DIFF
--- a/rc/filetype/ruby.kak
+++ b/rc/filetype/ruby.kak
@@ -140,7 +140,7 @@ define-command -hidden ruby-indent-on-char %{
     evaluate-commands -no-hooks -draft -itersel %{
         # align middle and end structures to start
         try %{ execute-keys -draft <a-x> <a-k> ^ \h * (else|elsif) $ <ret> <a-a> i <a-semicolon> <a-?> ^ \h * (if)                                                       <ret> <a-S> 1<a-&> }
-        try %{ execute-keys -draft <a-x> <a-k> ^ \h * (when)       $ <ret> <a-a> i <a-semicolon> <a-?> ^ \h * (case)                                                     <ret> <a-S> 1<a-&> }
+        try %{ execute-keys -draft <a-x> <a-k> ^ \h * (else|when)  $ <ret> <a-a> i <a-semicolon> <a-?> ^ \h * (case)                                                     <ret> <a-S> 1<a-&> }
         try %{ execute-keys -draft <a-x> <a-k> ^ \h * (rescue)     $ <ret> <a-a> i <a-semicolon> <a-?> ^ \h * (begin|def)                                                <ret> <a-S> 1<a-&> }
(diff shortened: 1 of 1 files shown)


alexherbo2:
@johnisom Is [Crystal] affected too?

[Crystal]: https://crystal-lang.org

johnisom:
> @johnisom Is [Crystal] affected too?
> 
> 
> 
> [Crystal]: https://crystal-lang.org

Good question! It would seem to be, because there isn't even a `crystal-indent-on-insert-char` (or something to that effect) function. I haven't tested it out but will soon.

My main programming language is actually Ruby, but since Crystal syntax is based heavily off of Ruby, I should have no trouble working this out.

johnisom:
@alexherbo2 it seems that Crystal isn't affected, but for more horrible reasons. As it currently stands, `rc/filetype/crystal.kak` doesn't handle indenting after new line when entering the beginning of a block statement like `if`/`else`, `begin`/`rescue`, etc. Nor does it nicely complete the `end` for you when you enter a new line after an opening statement. The smartness of indenting and inserting is super minimal, and I will open a PR that makes it behave a bit more like `ruby.kak`.

johnisom:
The current implementation doesn't account for this style of syntax:

``` ruby
def foo
  do_risky_stuff_that_might_raise_an_error
rescue ExceptionType => e
  recover_from_error
end
```

This is perfectly valid Ruby syntax, so I'll go ahead and make a change that supports this.

johnisom:
I also just noticed that with the current implementation, the `end` statement only aligns with the most recent opener, even if it has been closed. If perhaps it were dependent on `<a-a>i` to select everything in the current indentation level and then move up just 1 more line, it would be able to align `end` to the correct indentation level.

johnisom:
Disabling `ruby-insert-on-new-line`, here's what it currently looks like to have nested statements:

GIF of incorrect deindenting

And here is what it looks like with my fix:

GIF of correct deindenting

johnisom:
There's also another problem I noticed. When pressing `<ret>` after a `case` statement, it will insert `end`, even if it exists *if* the `when` clauses are on the same indentation level as the `case` keyword.

Example of the problem (▌ = cursor):

``` ruby
case foo▌
when 'bar' then 'bam'
when 'baz' then 'qux'
end
```

On `<ret>` becomes:

``` ruby
case foo
▌
end
when 'bar' then 'bam'
when 'baz' then 'qux'
end
```

johnisom:
One more thing: `case` statements can have an `else` clause for when none of the `when`s match. Currently, it isn't being aligned to the level of the `case` keyword like it should. I'll also publish a quick fix for that.

alexherbo2:
@johnisom I did not do it because I prefer [manual indent][manual-indent.kak].

[manual-indent.kak]: https://github.com/alexherbo2/manual-indent.kak

johnisom:
> @johnisom I did not do it because I prefer [manual indent](https://github.com/alexherbo2/manual-indent.kak).

@alexherbo2 are you referencing `case..when..else..end`, nested blocks, or `def..rescue..end`?

PR #3637, opened 2020-08-01
rc: Fill out `rc/filetype/crystal.kak`

johnisom:
Crystal's syntax is based heavily off of ruby, yet the experience of programming Crystal in Kakoune is not comparable to that of Ruby. I added indenting on new line, inserting the closing `end` token on newline, deindenting on completion of intermediate/closing token, and copying comment prefixes from line to line. In general, I cleaned up the file with indentation and comments, making it easier to hack for the next person that comes along.

### DIFF
--- a/rc/filetype/crystal.kak
+++ b/rc/filetype/crystal.kak
@@ -190,8 +190,8 @@ define-command -hidden crystal-indent-on-new-line %{
     evaluate-commands -no-hooks -draft -itersel %{
         # Copy previous line indent
         try %{ execute-keys -draft K <a-&> }
-        # Remove empty line indent
-        try %{ execute-keys -draft k <a-x> s ^\h+$ <ret> d }
+        # Remove previos line's trailing spaces
+        try %{ execute-keys -draft k :ruby-trim-indent <ret> }
     }
 }
 
--- a/rc/filetype/crystal.kak
+++ b/rc/filetype/crystal.kak
@@ -204,6 +204,8 @@ define-command -hidden crystal-indent-on-char %{
     evaluate-commands -no-hooks -draft -itersel %{
         # align 'else/elsif' to 'if'
         try %{ execute-keys -draft <a-x> <a-k> ^\h*(?:else|elsif)$ <ret> <a-semicolon> <a-?> ^\h*(?:if) <ret> <a-S> 1<a-&> }
+        # align 'when' to 'case'
+        try %{ execute-keys -draft <a-x> <a-k> ^\h*(?:when)$ <ret> <a-semicolon> <a-?> ^\h*(?:case) <ret> <a-S> 1<a-&> }
     }
 }
 
--- a/rc/filetype/crystal.kak
+++ b/rc/filetype/crystal.kak
@@ -181,9 +181,9 @@ evaluate-commands %sh[
 define-command -hidden crystal-indent-on-new-line %{
     evaluate-commands -no-hooks -draft -itersel %{
         # Copy previous line indent
-        try %{ execute-keys -draft 'K<a-&>' }
+        try %{ execute-keys -draft K <a-&> }
         # Remove empty line indent
-        try %{ execute-keys -draft 'k<a-x>s^\h+$<ret>d' }
+        try %{ execute-keys -draft k <a-x> s ^\h+$ <ret> d }
     }
 }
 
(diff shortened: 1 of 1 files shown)


johnisom:
Many of the changes were inspired by what already was in the `ruby.kak` file, yet were more robust. Check out #3634 for details explaining some choices I made.

mawww:
@alexherbo2 Can you give this PR a quick review ?

johnisom:
There a couple small issues that need to be sorted out before this can be merged. I'll comment back here when they're fixed

alexherbo2:
How to escape comments if we don’t want them, or wrong `end` insertion.  I ended up disabling filetype hooks because they were too intruisive, and `ruby.kak` looked quite a mess when I stopped to write in it.

@mawww By looking at @johnisom commits, it looks he has carefully cherry picked the features from `ruby.kak`.

@mawww, @johnisom What do you think to rewrite all filetypes with named commands sharing the logic?

johnisom:
@alexherbo2 @mawww I just fixed 2 small bugs, I think this is good to review now

johnisom:
> @mawww By looking at @johnisom commits, it looks he has carefully cherry picked the features from `ruby.kak`.

This is correct

alexherbo2:
@johnisom I have not checked, but does your implementation work with multiple selections?, if one needs an `end` or a comment, but not other selections.

johnisom:
> How to escape comments if we don’t want them, or wrong `end` insertion. I ended up disabling filetype hooks because they were too intruisive, and `ruby.kak` looked quite a mess when I stopped to write in it.

Yes, I agree that the filetypes for Ruby and what I cherry-picked to Crystal are quite a mess. There's a lot of "smart" auto-indentation and auto-insertion, to the point where it is understandable for one to want to disable filetype hooks. Because of how complicated each file is, it means any change is somewhat likely to introduce a new, unexpected behavior. I would be open to refactoring the `ruby.kak` and `crystal.kak` filetypes.

johnisom:
> @johnisom I have not checked, but does your implementation work with multiple selections?, if one needs an `end` or a comment, but not other selections.

It seems to be working with multiple selections for comments, typing `end` or other code, etc. Each of the hook functions has  a top level `evaluate-commands -no-hooks -draft -itersel` in which everything else is contained. The `-itersel` means that it will execute the following commands once per selection individually.

alexherbo2:
@johnisom `-itersel` is awesome :p

johnisom:
> How to escape comments if we don’t want them

Are you asking if you are typing in comments and want to type code the next line, how would you go about that?

> , or wrong `end` insertion.

I fixed an issue with wrong `end` insertion in my most recent commit to this branch. The issue before was that if you typed an opening statement (`def`, `if`, `begin`, `case`, etc.) in the top level indentation (no preceding white space), the `end` wouldn't be automatically added for you. It only worked if you were nested at least 1 level deep in indentation. I just fixed this and now it's working at the top level.

There was also another issue that I fixed today in 6cf39a5. There was a similar issue in `ruby.kak` (which I sadly introduced) and I opened PR #3640 to fix it.

alexherbo2:
I liked [double-tap] idea.

[double-tap]: https://github.com/Osse/double-tap

johnisom:
> I liked [double-tap](https://github.com/Osse/double-tap) idea.

I like this idea. It's very clean. It does strike me as more of a plugin than a core feature though.

johnisom:
> @mawww, @johnisom What do you think to rewrite all filetypes with named commands sharing the logic?

Would you mind opening an issue that describes this more in detail? I'd be interested.

alexherbo2:
The same applies for [vim-endwise] imo.

[vim-endwise]: https://github.com/tpope/vim-endwise

alexherbo2:
> It does strike me as more of a plugin than a core feature though.

@johnisom Hooks are sadly not very composable with the order of execution / results.

johnisom:
> The same applies for [vim-endwise](https://github.com/tpope/vim-endwise) imo.

I also agree here

johnisom:
> > It does strike me as more of a plugin than a core feature though.
> 
> @johnisom Hooks are sadly not very composable with the order of execution / results.

Could a mapping be used instead? Remap `<ret>` to run the plugin function, which checks if cursor is currently inside empty comment. If so, do special behavior (without hooks). Otherwise, execute `<ret>` with all the hooks. I'm pretty new to developing kakscript, so correct me if my thought process is incorrect.

alexherbo2:
The cleanest approach is endeed mappings.  It is a bit more involved, but if a user wants multiple functionalities on the same key, he can just chain the commands with the desired order.

``` kak
define-command insert-on-return %{
  command-1
  command-2
}

map global insert <ret> '<a-;>: insert-on-return<ret>'
```

johnisom:
@alexherbo2 Is this PR ready for merge or awaiting changes?

alexherbo2:
It looks good.

PR #3642, opened 2020-08-04
Update spec file to newer version

jkonecny12:
Adapt spec file for daily builds on Fedora.

### DIFF
--- a/contrib/kakoune.spec
+++ b/contrib/kakoune.spec
@@ -5,7 +5,7 @@
 %global build_ldflags   %{build_ldflags} -flto
 
 Name:           kakoune
-Version:        2020.01.16
+Version:        2020.08.04
 Release:        1%{?dist}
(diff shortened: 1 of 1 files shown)


contrib/kakoune.spec
```diff
@@ -83,6 +83,9 @@ popd
 
 
 %changelog
+* Thu Jan 16 2020 Jiri Knoceny <jkonecny@redhat.com> - 2020.08.04-1
```
caksoylar, line 86: Fully aware of the irony of correcting the author's name but thought I'd ask anyway, is the last name a typo?

jkonecny12: :D :+1: Good one!

jkonecny12: Even the date was wrong. Should be fine now.

jkonecny12:
Ping, could you please merge this?

PR #3656, opened 2020-08-13
bring *grep* buffer to front in context of toolsclient

vbauerster:
When `toolsclient` is set and contains not a `grep` buffer, grep next/prev never brings it upfront. This PR fixes this.

### DIFF
--- a/rc/tools/grep.kak
+++ b/rc/tools/grep.kak
@@ -53,13 +53,18 @@ define-command -hidden grep-jump %{
 define-command grep-next-match -docstring 'Jump to the next grep match' %{
     evaluate-commands -try-client %opt{jumpclient} %{
         buffer '*grep*'
-        # First jump to enf of buffer so that if grep_current_line == 0
+        # First jump to end of buffer so that if grep_current_line == 0
         # 0g<a-l> will be a no-op and we'll jump to the first result.
         # Yeah, thats ugly...
         execute-keys "ge %opt{grep_current_line}g<a-l> /^[^:]+:\d+:<ret>"
         grep-jump
     }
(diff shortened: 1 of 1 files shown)


vbauerster:
Actually it's a little bug fix as well, because if `toolsclient` doesn't contain a `grep` buffer it ends up opening `goto` menu on every next/prev.

rc/tools/grep.kak
```diff
@@ -53,13 +53,18 @@ define-command -hidden grep-jump %{
...
         # 0g<a-l> will be a no-op and we'll jump to the first result.
         # Yeah, thats ugly...
         execute-keys "ge %opt{grep_current_line}g<a-l> /^[^:]+:\d+:<ret>"
         grep-jump
     }
-    try %{ evaluate-commands -client %opt{toolsclient} %{ execute-keys gg %opt{grep_current_line}g } }
+    try %{
+        evaluate-commands -client %opt{toolsclient} %{
+            buffer '*grep*'
+            execute-keys gg %opt{grep_current_line}g
```
mawww, line 65: I wonder why we have that `gg` there, it does not seem necessary.

vbauerster: Indeed, just checked without `gg` and it worked, also worked without `ge` but comment suggests it is for edge cases.
@lenormf Can you please suggest?

vbauerster:
I never set `jumpclient` explicitly, instead use following command:
```
define-command toolsclient -params ..1 -client-completion \
-docstring 'toolsclient [<jumpclient>]: set current client as toolsclient' \
%{
    evaluate-commands %sh{
        if [ $# -eq 1 ]; then
            printf "set current jumpclient '%s'" "$1"
        fi
    }
    set current toolsclient %val{client}
    hook -once global ClientClose %val{client} %{
        set current toolsclient ''
    }
}
```

PR #3658, opened 2020-08-14
Add a find.kak module

Eluminae:
Hello there,

As discussed [here](https://discuss.kakoune.com/t/command-wildcard-expansion/1206/17), I needed a simple way to browse subdirectories and open buffers based on some wildcard rules.

I wrote a find module pretty similar to the grep one.

This module wrap the find command and open a scratch dedicated buffer to open found files.

    :find ./ -name *document*
    :find app/controllers -name user*.rb

Futhermore, we got all the possibilities from this venerable find.

There still is some commands as

    :find-next-match
    :find-previous-match

And the toolsclient is used if it exists.

I though it was a very basic way to allow kakoune to close this gap without embedding complex logic in its core and I'd  like to have this find command in a vanilla kakoune installation. Plus, I think that find, similary to grep, is a very basic dependency that all distributions will have easily.

Thanks a lot for this amazing project.

### DIFF
--- a/rc/tools/find.kak
+++ b/rc/tools/find.kak
@@ -1,0 +1,73 @@
+declare-option -docstring "shell command run to search for file in a directory" \
+    str findcmd 'find'
+declare-option -docstring "name of the client in which utilities display information" \
+    str toolsclient
+declare-option -hidden int find_current_line 0
+
+define-command -params .. -file-completion \
+    -docstring %{find [<arguments>]: find utility wrapper
+All the optional arguments are forwarded to the find utility} \
+    find %{ evaluate-commands %sh{
+     if [ $# -eq 0 ]; then
+         set -- "${kak_selection}"
+     fi
+
+     output=$(mktemp -d "${TMPDIR:-/tmp}"/kak-find.XXXXXXXX)/fifo
+     mkfifo ${output}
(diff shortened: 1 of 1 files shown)


krobelus:
Unfortunately this is not a good name because both the `:find` command and the `*find*` buffer are used by the
[kakoune-find](https://github.com/occivink/kakoune-find) plugin.

For selecting a single file to edit, Kakoune offers nice completion filtering which you can feed with `find`:

```kak
define-command -override my-find-edit -params 1.. %{
	edit -- %arg{@}
} -shell-script-candidates %{ find -type f | sed s,^./,, }
```

For browsing a directory tree it is a bit more convenient than running `!find` in a scratch buffer followed by `<a-a><a-w>gf`.

Eluminae:
Woooh it's just amazing. This kakoune-find seems super nice and this code example gave me a lot of ideas. 

Thanks for this inspiring return !

Eluminae:
I close this as I got better alternatives to my need. As default, the file completion builtin is enough.

I will build my own merged version of find.kak as I think finding words in existing buffer is not a need I got but finding filename based on rules is. Anyway, the find-apply-changes of find.kak is awesome with the grep.kak.

PR #3662, opened 2020-08-16
Crystal: Add support for macros

alexherbo2:
https://crystal-lang.org/api/toplevel.html#macro-detail

### DIFF
--- a/rc/filetype/crystal.kak
+++ b/rc/filetype/crystal.kak
@@ -15,7 +15,7 @@ hook global WinSetOption filetype=crystal %{
     require-module crystal
 
     add-highlighter window/crystal ref crystal
-    evaluate-commands set-option window static_words %opt{crystal_keywords} %opt{crystal_attributes} %opt{crystal_objects}
+    evaluate-commands set-option window static_words %opt{crystal_keywords} %opt{crystal_macros} %opt{crystal_attributes} %opt{crystal_objects}
 
     hook window InsertChar .*   -group crystal-indent crystal-indent-on-char
     hook window InsertChar '\n' -group crystal-indent crystal-indent-on-new-line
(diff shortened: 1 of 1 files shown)


rc/filetype/crystal.kak
```diff
@@ -239,6 +250,14 @@ define-command -hidden crystal-fetch-keywords %{
     }
 }
 
+# https://crystal-lang.org/api/toplevel.html#macro-detail
+define-command -hidden crystal-fetch-macros %{
+    set-register dquote %sh{
+        curl --location https://crystal-lang.org/api/toplevel.html |
+        kak -f '%1smacro\h+<lt>strong<gt>(\w+[?!]?)<lt>/strong<gt><ret>y%<a-R>a<ret><esc><a-_>a<del><esc>'
```
lenormf, line 257: As mentioned in #3523, assuming `curl` is on the system, fetching arbitrary data and running `kak` on top is probably not the play.

alexherbo2: It is just a command for the maintainer to see the updates.

mawww: I'd like those to be removed from the .kak file, its a useful snippet for you, but it is not really something we need to keep around for all users, and it also depends on how that webpage is formatted which might change.

alexherbo2: If someone else want to maintain the language he has the helpers to be sure to not miss any keyword; if the command fails he will see it in its clipboard and adapts the command.

sidkshatriya: https://crystal-lang.org/api/toplevel.html is a trust worthy URL at this point but there no guarantee what content it may serve in the future. Piping the contents of the URL to `kak -f` does seem a bit risky does it not?

alexherbo2: The command is only executed by the maintainer when he wants to update the keywords.  My workflow has been to find a way to collect the keywords and create a command to see the updates; when I want to check new keywords, I execute the command to fetch, paste the result into the buffer and update the keywords.

lenormf: You could centralise the keywords fetching commands into a script, which we could store into `contrib`. Support several languages even (e.g. Python).

krobelus: If we have multiple of those, then a single script in `contrib` that runs all of them would make it easier for maintainers!
Or maybe putting the script in a comment would be less contentious. When you run it, you want to edit this file anyway, so it's just an `eval %val{selection}` away?

PR #3664, opened 2020-08-17
Rename base16 theme to base16-old, add base16-light and base16-dark themes

clarfonthey:
Base16 was original intended to be a colour scheme template, not a colour scheme in its own right, although it makes sense that the provided `base16` colour scheme matches the default-dark colour scheme provided by the author.

However, it would be extremely useful if the colour scheme used the same colour names as the original (i.e. 0 - F) so that folks can modify the theme to create their own base16-based kakoune themes.

I ended up having to modify the existing theme so it would actually match what's provided, and as such I renamed the old theme to base16-old for folks who liked the original. I'm open to alternatives, although this is mostly what made sense to me, as simply calling it base16 when the default doesn't even prefer the light or dark theme seems wrong. The current colours are based upon the documentation on the website and the vim implementation, although there is always some guesswork as the "spec" is just a short list of words for each colour.

Note: the existing faces don't 100% match the original specifications, as there are things that are separate there which aren't separate faces in kakoune -- e.g. the "meta" colour is supposed to be different from closing/opening tags in PHP, but here they're the same colour.

There is also one major change from the original which is that comments are given the "status line" colour (slightly dimmed foreground) as opposed to the "invisibles" colour because:

1. You shouldn't have to squint to read documentation
2. The invisibles colour should be different from the comment colour, as comments contain invisibles


lenormf:
Could you please post screenshots?

The colourscheme was based on the colours seen in the [repository for the Vim files](https://github.com/chriskempson/base16-vim), which were what most users had in mind when talking about "base16".

I don't think the renaming is necessary, as `base16-dark` should be very similar to the current `base16`, right? We can mention in the changelog that the appropriate flavour for the latter is the former.

clarfonthey:
Closing this since I ended up changing a bunch of the work I was doing on making my own colour scheme, and think that I don't really have the time to work on this. Will leave code on the branch though in case someone else wants to pick it up.

PR #3679, opened 2020-08-23
Avoid accidentally using environment variables in sh scopes

### LINKED ISSUE #3673: Arbitrary file inclusion/code execution, when no user directory exists, via the environment
Reported by @chambln 

Symptoms:

```
shell stderr: <<<
+ [ -d /home/greg/.config/kak/colors ]
+ [ -z /home/greg/.bashrc ]
+ [ -n /home/greg/.bashrc ]
+ printf source %%{%s} /home/greg/.bashrc
>>>
/home/greg/.bashrc:no such command: 'case'
/usr/share/kak/kakrc:5:864: 'evaluate-commands' 1:1: 'colorscheme' 1:2: 'evaluate-commands' 1:1: 'source' no such command: 'case'
error while parsing kakrc:
    1:1: 'source' 5:864: 'evaluate-commands' 1:1: 'colorscheme' 1:2: 'evaluate-commands' 1:1: 'source' no such command: 'case'
```

### Steps 

* move the user directory if it exists, for example `mv ~/.config/kak ~/.config/kak.bak`
* run `env filename=a kak`

### Outcome 

Startup error:

```
/usr/share/kak/kakrc:28:1: 'evaluate-commands' 1:1: 'colorscheme' 1:2: 'evaluate-commands' 1:1: 'source' /home/lenormf/a: No such file or directory
error while parsing kakrc:
    1:1: 'source' 28:1: 'evaluate-commands' 1:1: 'colorscheme' 1:2: 'evaluate-commands' 1:1: 'source' /home/lenormf/a: No such file or directory
```

### Expected 

No errors, the editor starts normally.

### Notes

The issue stems from this part of the global `kakrc` file:

https://github.com/mawww/kakoune/blob/033a28dca2a5994bb9f7756c7cbd54e68554654d/share/kak/kakrc#L14-L22

* line 14: the `${kak_config}/colors` directory doesn't exist
* line 17: the `filename` environment variable is set
* line 22: source whatever the variable is pointing to

The issue was made hard to diagnose because `~/.bashrc`, which was the interactive shell being used, set the `filename` variable internally, and was therefore inherited by all the Kakoune servers.

Since this vulnerability could be exploited to execute arbitrary shell code, by making `filename` point to a file that contains Kakoune code e.g. `nop %sh{…}`, I recommend a new release is produced once the issue is fixed.

### Fix

We might have to go over all shell scopes and set a default value to variables who are set conditionally, but maybe there's a portable way of handling that better?

krobelus:
On the instances with `altfile` this was already the case, but this makes it more obvious.

Closes #3673

### DIFF
--- a/rc/detection/modeline.kak
+++ b/rc/detection/modeline.kak
@@ -58,9 +58,8 @@ define-command -hidden modeline-parse-impl %{
         case "${kak_selection}" in
             *vi:*|*vim:*) type_selection="vim";;
             *kak:*|*kakoune:*) type_selection="kakoune";;
-            *) echo "echo -debug Unsupported modeline format";;
+            *) echo "echo -debug Unsupported modeline format"; exit 1 ;;
         esac
-        [ -n "${type_selection}" ] || exit 1
 
         # The following subshell will keep the actual options of the modeline, and strip:
         # - the text that leads the first option, according to the official vim modeline format
@@ -82,6 +81,7 @@ define-command -hidden modeline-parse-impl %{
             case "${type_selection}" in
                 vim) tr=$(translate_opt_vim "${name_option}" "${value_option}");;
                 kakoune) tr=$(translate_opt_kakoune "${name_option}" "${value_option}");;
+                *) tr="";;
             esac
 
             [ -n "${tr}" ] && printf %s\\n "${tr}"
--- a/rc/filetype/lua.kak
+++ b/rc/filetype/lua.kak
@@ -65,6 +65,8 @@ define-command lua-alternative-file -docstring 'Jump to the alternate file (impl
             [ ! -f $altfile ] && echo "fail 'implementation file not found'" && exit
         ;;
         *.lua)
+            altfile=""
+            altdir=""
             path=$kak_buffile
             dirs=$(while [ $path ]; do printf %s\\n $path; path=${path%/*}; done | tail -n +2)
             for dir in $dirs; do
@@ -74,7 +76,7 @@ define-command lua-alternative-file -docstring 'Jump to the alternate file (impl
                     break
                 fi
             done
-            [ ! -d $altdir ] && echo "fail 'spec/ not found'" && exit
+            [ ! -d "$altdir" ] && echo "fail 'spec/ not found'" && exit
         ;;
         *)
             echo "fail 'alternative file not found'" && exit
--- a/rc/filetype/ruby.kak
+++ b/rc/filetype/ruby.kak
@@ -120,6 +120,8 @@ define-command ruby-alternative-file -docstring 'Jump to the alternate file (imp
             [ ! -f $altfile ] && echo "fail 'implementation file not found'" && exit
         ;;
         *.rb)
+            altfile=""
+            altdir=""
             path=$kak_buffile
             dirs=$(while [ $path ]; do echo $path; path=${path%/*}; done | tail -n +2)
             for dir in $dirs; do
@@ -130,7 +132,7 @@ define-command ruby-alternative-file -docstring 'Jump to the alternate file (imp
                     break
                 fi
             done
-            [ ! -d $altdir ] && echo "fail 'spec/ and test/ not found'" && exit
+            [ ! -d "$altdir" ] && echo "fail 'spec/ and test/ not found'" && exit
         ;;
         *)
             echo "fail 'alternative file not found'" && exit
--- a/rc/tools/clang.kak
+++ b/rc/tools/clang.kak
@@ -173,6 +173,8 @@ define-command clang-diagnostics-next -docstring "Jump to the next line that con
     evaluate-commands %sh{
         eval "set -- ${kak_quoted_opt_clang_errors}"
         shift # skip timestamp
+        unset line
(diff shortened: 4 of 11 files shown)


Screwtapello (approved):

rc/filetype/lua.kak
```diff
@@ -59,6 +59,7 @@ add-highlighter shared/lua/code/function_call regex \b([a-zA-Z_]\w*)\h*(?=[\(\{]
 # ‾‾‾‾‾‾‾‾
 
 define-command lua-alternative-file -docstring 'Jump to the alternate file (implementation ↔ test)' %{ evaluate-commands %sh{
+    altfile=""
```
lenormf, line 62: What about `$altdir`?

rc/filetype/ruby.kak
```diff
@@ -110,6 +110,7 @@ add-highlighter shared/ruby/code/ regex \b(\w+:(?!:))|(:?(\$(-[0FIKWadilpvw]|["'
 # ‾‾‾‾‾‾‾‾
 
 define-command ruby-alternative-file -docstring 'Jump to the alternate file (implementation ↔ test)' %{ evaluate-commands %sh{
+    altfile=""
```
lenormf, line 113: What about `$altdir`?

rc/tools/clang.kak
```diff
@@ -173,6 +173,8 @@ define-command clang-diagnostics-next -docstring "Jump to the next line that con
     evaluate-commands %sh{
         eval "set -- ${kak_quoted_opt_clang_errors}"
         shift # skip timestamp
+        unset line
+        unset first_line
```
lenormf, line 177: Do we need those? `$line` is explicitly assigned to on line 186, and `$first_line` then doesn't need to be initialised.

rc/tools/git.kak
```diff
@@ -247,7 +248,7 @@ define-command -params 1.. \
             fi
         fi
 
-        if [ "$wrapped" = true ]; then
+        if "$wrapped"; then
```
lenormf, line 251: Why change this? I'm not even sure this works as expected either.

rc/detection/modeline.kak
```diff
@@ -55,6 +55,7 @@ define-command -hidden modeline-parse-impl %{
             printf %s\\n "set-option buffer ${key} ${value}"
         }
 
+        type_selection=""
```
mawww, line 58: I'd prefer to `exit 1` in the default case (line 61) and remove the test at line 63

rc/filetype/lua.kak
mawww: Move that inside the lua case ? the default case already exits.

rc/tools/man.kak
```diff
@@ -68,6 +68,7 @@ define-command -params ..1 \
     subject=${1-$kak_selection}
 
     ## The completion suggestions display the page number, strip them if present
+    pagenum=""
```
mawww, line 71: move that to a default case ?

rc/tools/git.kak
Screwtapello: Previously it was comparing the string stored in $wrapped to the string "true".

Now it's executing the program named in $wrapped, which should effectively be `/bin/true` or `/bin/false`. It's safe because $wrapped is always set to one or the other now, it's not going to execute some random other thing.

It's probably less efficient to launch an external command than to do a string comparison, but it's not like this executed frequently.

lenormf: But why modify it? The hunk is not related to the changeset, it doesn't fix any issue, it's not a performance improvement, and it's harder (slightly) to understand what the new code does.

krobelus: I agree that this change should not really be here, reverted.

Correct me if I'm wrong, but I believe that this executes shell-builtins `true`/`false` if they exist (not guaranteed by POSIX). I find it more elegant than `[ $bool = true ]` or `test -n` but I guess it's not so common.

rc/tools/clang.kak
krobelus: If `$line` was already set from the environment and is never set in the loop, then
`line=${line-$first_line}` will not change `$line`, so the inherited value will be executed.

Similarly, if `$first_line` is inherited from the environment, then its value will never change, and it could be assigned to `$line`.

Apparently `${var-default}` expands to `default` if `$var` is unset,
whereas `${var:-default}` expands to `default` if `$var` is unset or empty.

That's why I use `unset` over `line=""`. Maybe it should be rewritten.

rc/filetype/lua.kak
krobelus: Done, also added initialization of `altdir`.

I need to quote it in `[ -d "$altdir" ]` if it can be empty, because in that case it would be true - `[` with a single non-empty argument is always true.

rc/tools/git.kak
Screwtapello: I checked `dash(1)` and `bash(1)`; only dash lists `true` as a shell builtin (as a synonym for `:`), and bash lists neither.

Now that I look more closely, `command -V` says both are builtins in both shells.

PR #3682, opened 2020-08-27
Add information on leaving multiple selection in TRAMPOLINE.

m-kru:


lenormf:
I'd rather the space key had its own paragraph with a clear explanation of its side effects. Mentioning it like an afterthought of (only) one of the primitives that select text is a sure way to have users miss it.

sidkshatriya:
Yes, this was one of the major issues I had as a newbie to kak. It was easy to create multiple selections but difficult to figure out how to remove them!

The effect of space must have its own paragraph and it should be mentioned when introducing multiple selections also. Its an important concept and so some repetition is good.

m-kru:
@lenormf I do not know these side effects. It would be really good if someone described `space`. As a beginner I had the same difficulties as @sidkshatriya .

lenormf:
You could go for a paraphrasing of [the documentation](https://github.com/mawww/kakoune/blob/master/doc/pages/keys.asciidoc#L573-L574). I don't think there are any other side effects, so it would be a shoe in for TRAMPOLINE.

PR #3698, opened 2020-09-02
rc git: Create individual Git commands

lenormf:
Split all the subcommands supported by the `:git` wrapper into
individual commands. This approach has several advantages:

* every `git-` command is discoverable through the command prompt's
  fuzzy matcher, as opposed to being hidden behind `:git` as an
  argument
* remain consistent with the rest of the built-in commands, which
  all follow the "prefix paradigm"
* every `git-` command has its own docstring, completion command etc.
* calling a `git-` command doesn't result in a shell being spawned
  regardless (`hide-diff`, `hide-blame` etc.)
* it's harder to leave a command undocumented (hello `:git reset`)

The old behaviour can be kept with the following user command:

```
define-command -params .. git %{
    evaluate-commands "git-%arg{@}"
}
```


lenormf:
I'm publishing this as draft, it should be cross-compatible with the current implementation (not tested much though).

More commits could be added to the branch, I'm still unsure if we couldn't turn some `(` subshells into their `{` equivalent, keeping #2955 in mind.

The completion callbacks of some commands could also be implemented.

rc/tools/git.kak
```diff
@@ -163,14 +163,14 @@ define-command -params .. -docstring %{
 define-command -params .. -docstring %{
     git-diff [<file>]...: show the diff of the given files
 } git-diff %{ evaluate-commands %sh{
-    output=$(mktemp -d "${TMPDIR:-/tmp}"/kak-git-diff.XXXXXXXX)/fifo
+    output=$(mktemp "${TMPDIR:-/tmp}"/kak-git-diff.XXXXXXXX)
     mkfifo "${output}"
```
krobelus, line 234: This leads to an error like
```
mkfifo: cannot create fifo '/tmp/kak-git-diff.kzg6uxub': File exists
```

krobelus (commented):
I like that this makes it possible to override individual git commands.
The old `:git xyz` was nice but I guess it's easy to add if we disregard escaping.

rc/tools/git.kak
lenormf: Thanks for testing and sorry about that, I messed up my `git add -p`. It should be fixed now, I added the appropriate `rm "${output}"` call.

lenormf:
Is this merge-able in principle? If it is, I'll test it more so this can be included in the next release.

Issue #2306, opened 2018-08-15
Add support for a default value in typed expansions


alexherbo2:
### Syntax

```
%type(name:default)
```

#### 1

```
%val(count:1)
```

#### 2

```
%arg(1:foo)
```

Screwtapello:
Use-case: some commands treat "no count" and "count=1" differently, so `%val{count}` needs to distinguish those cases. On the other hand, many commands (especially the most common commands, like `hjkl`) treat "no count" and "count=1" the same, so it would be nice if it were easy to make custom commands follow that convention.

A similar argument applies to `%val{register}`, where custom commands that resemble delete/yank/paste will want to use the `"` register; custom commands that search will want to use the `/` register, etc.

It is possible to do this entirely in shell, but it's quite ugly. Compare:

    %val{count:1}

    %sh{ if [ "$kak_count" -eq 0 ] ; then echo 1; else echo $kak_count; fi }

alexherbo2:
Could be less — but still — ugly if `count` was empty.

Delapouite:
This reminds me of this discussion trying to distinguish when numbers are used as a *count* vs when they are used as an *index* by commands: https://github.com/mawww/kakoune/issues/1689

Delapouite:
> Use-case: some commands treat "no count" and "count=1" differently, so %val{count} needs to distinguish those cases

I just found out that vim has distinct `v:count` and `v:count1` for this exact purpose: https://neovim.io/doc/user/eval.html#v:count

PR #3700, opened 2020-09-02
src: Make count on <space> relative to the main selection

### LINKED ISSUE #1323: make `count` <space> commands relative to main selection

lenormf:
This commit makes counts passed to the <space> primitive be added
to the index of the main selection, instead of considering them the
direct index of the selection to keep.

This logic seems more helpful interactively, in cases where many
selections are off-screen, or counting the visible ones is too tedious.

Counting wraps around the selection, e.g. if the main selection if
the last one, a count of 1 will refer to the first one.

The modifications assume the selection is never empty, and uses
`selections.size()` as a modulo operand without any prior sanity
checks.

Fixes #1323.

### DIFF
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -2082,7 +2082,7 @@ void select_whole_buffer(Context& context, NormalParams)
 void keep_selection(Context& context, NormalParams p)
 {
     auto& selections = context.selections();
-    const int index = p.count ? p.count-1 : selections.main_index();
+    const int index = (selections.main_index() + p.count) % selections.size();
     if (index >= selections.size())
         throw runtime_error{format("invalid selection index: {}", index)};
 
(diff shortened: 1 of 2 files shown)


src/normal.cc
```diff
@@ -2082,7 +2082,7 @@ void select_whole_buffer(Context& context, NormalParams)
 void keep_selection(Context& context, NormalParams p)
 {
     auto& selections = context.selections();
-    const int index = p.count ? p.count-1 : selections.main_index();
+    const int index = (selections.main_index() + p.count) % selections.size();
```
krobelus, line 2085: What's slightly surprising here is that `1<space>` behaves differently than `<space>`.
For example on

```
%(foo)
%(bar)
```

where the first selection is the main one.

I'm not sure what's the best behavior here.

lenormf: The default count in Kakoune is 0, so I think this behaviour makes sense and is predictable.

krobelus: But from the user perspective all commands default to 1, right? At least I couldn't find one that doesn't.

krobelus:
Does it make sense to do the same for `<a-&>`?

src/normal.cc
occivink: You could say that `s`,`<a-k>` and their variants are 0-based, since the default behavior is to use the 0 capture group. I think it's fine that different keys have a different default, depending on what makes sense.

lenormf: Correct, the default count value is not consistent across all primitives. C.f. #2306

mawww:
I dont think this would be a good change, as the new behaviour is equivalent to `<count>)<space>`. The old behaviour is useful in the common case where all selections are visible, I think most users would find it easier to count selections in buffer order than from the current main, and it would be especially cumbersome to refer to a selection before the main one as we dont have negative counts.

lenormf:
Alright, we can close #1323 now, then.

Issue #2278, opened 2018-08-01
Hitting an un-mapped key after `:enter-user-mode -lock` shouldn't leave the user mode


lenormf:
### Steps 

* declare a user mode with a mapping in it
* enter the user mode in lock mode i.e. `:enter-user-mode -lock …`
* hit a key that's not mapped in the user mode

### Outcome 

The user mode is left, as if <kbd>escape</kbd> had been hit.

### Expected 

The context should remain in user mode.

PR #3703, opened 2020-09-02
src: Implement display-line selectors for `g`/`G`

### LINKED ISSUE #1451: Add a goto primitive to move the anchor to the window's border
Hi,

When soft wrapping is enabled and the anchor is in the middle of a wrapped line, there's no easy way to send the anchor to the beginning of the /display/ line. The traditional `gh`/`gl` combination sends us back to the beginning/end of the /buffer/ line, I think we could benefit from new `goto` shortcuts to send the anchor to the beginning/end of the current /display/ line.

Thanks.

lenormf:
This commit allows moving the cursor around the current display-line,
in replace mode (`g`) or extend mode (`G`).

The `g` primitive is now case sensitive, to prevent the default
mappings to be overriden by `gh`/`gj`/`gk`/`gl`.

Selections whose cursors are not within the window at the time of
trigger of the mapping are dropped, since the concept of display-line
is bound to the window and doesn't exist "outside" of it.

Fixes #1451.


lenormf:
I created this draft PR to garner interest in the feature. The code is probably buggy, needs to be factorised, but it's helpful.

The side effect is dirty, which probably prevents the PR from being merged, but if enough users want to patch their local instance to add the feature, I'll clean the PR up.

Screwtapello:
For anybody else who wants to try this, assuming you've already cloned the Kakoune git repo:

    git fetch origin pull/3703/head:display-line-movement
    git switch display-line-movement
    make

...now `src/kak` should have these changes.

These new commands don't show up in the list of keybindings for "goto" mode yet, but they still work.

> Selections whose cursors are not within the window at the time of trigger of the mapping are dropped, since the concept of display-line is bound to the window and doesn't exist "outside" of it.

At least for a first draft, I'd suggest that moving by display-line should automatically drop *all* other selections, for a few reasons:

  - the only existing display-relative-movement command is the mouse-click, and that drops all other selections
  - there's nothing else in Kakoune that works inside the viewport and silently fails everywhere else, while there's other things that only work on the main selection but fail for other selections
  - better to reliably fail in the multiple-cursor case so people learn the limitation immediately, than to have it sometimes work and sometimes not
  - people who want this feature are almost certainly coming from Vim which only has a single cursor, so it won't be a big limitation for them

lenormf:
Thanks for the review, I forgot to add `H`/`J`/`K` to the menu :/

It probably makes sense to only move the main selection and drop all others, I originally thought it was easier to go from all to main than the opposite (in terms of code changes).

Screwtapello:
Some additional thoughts:

  - Making the goto mode case-sensitive might be a disruptive change; perhaps the new actions could be added with the Control modifier instead of Shift?
      - Alternatively, view-relative operations currently live in the View mode, maybe they could be moved there?
      - Either way, I expect most people to use these commands via mappings rather than directly, so I don't think it matters too much exactly what keys they use.
  - Normal j/k movement remembers the "intended" column - if the cursor is in column 40, then you press "j" down onto an empty line the cursor goes to column 1, but if you press "j" again down onto a long line it jumps back to column 40. gJ and gK do not respect or update this column memory.
  - It would be nice to have "gI" that works like "gi" on an unwrapped line, but "gH" on a wrapped line.

I've written this comment with the following setup:

    add-highlighter window/ wrap -word -indent -marker "… "
    map window normal j gJ
    map window normal k gK
    map window <a-h> gH
    map window <a-l> gL
    map window goto h H
    map window goto l L

lenormf:
I went for <kbd>alt</kbd> as a modifier because it felt more consistent, but I feel like it's not really convenient, so I'm open to changing that.

I think the goto mode is more appropriate than the view mode to handle those mappings, despite not handling counts and acting on the view, because in my view (pun intended) the former doesn't act directly on selections. Open to change as well, not hiding those mappings in a menu would be the best option, but I don't know what keys to bind them to.

I implemented <kbd>g<kbd><kbd>i<kbd> for the display lines.

I'm still not clear on how the editor currently keeps track of the previous location of the cursor upon hitting movement keys, so that's not implemented (yet).

Screwtapello:
I noticed that the view mode is not the only one with view-relative movement: the goto menu has "window top", "window center", and "window bottom" which are all view-relative, so that's another justification for putting these commands in the goto mode. Also, gt, gc, and gb also drop all non-main selections, so that's another justification for *that* behaviour too.

I think you forgot to add `<a-i>` to the goto menu again. :)

krobelus:
> I went for alt as a modifier because it felt more consistent, but I feel like it's not really convenient, so I'm open to changing that.

Yeah, it should be more convenient, especially repeating `g<a-j>` is a bit awkward. I considered adding a user mode so I can use `hjkl`. An `enter-user-mode -lock` might be useful then (unlike for `goto`) but I want to seamlessly use other movement keys. Since #2278 isn't a thing, I feel like the least surprising behavior is to add a convenient way to change the meaning of `j` in normal mode. I'll try this for some time:

```kak
map global normal <a-g> %{: my-toggle-display-lines-movement<ret>}
declare-option bool move_by_display_lines false
define-command -override my-toggle-display-lines-movement %{
    try %{
        evaluate-commands %sh{ $kak_opt_move_by_display_lines && echo fail }
        set-option global move_by_display_lines true
        echo Enabled moving by display lines.
        map global normal j g<a-j>
        map global normal k g<a-k>
        map global normal <a-h> g<a-h>
        map global normal <a-l> g<a-l>
        map global goto i g<a-i>
    } catch %{
        echo Disabled moving by display lines.
        set-option global move_by_display_lines false
        unmap global normal j
        unmap global normal k
        unmap global normal <a-h>
        unmap global normal <a-l>
        unmap global goto i
    }
}
```
> I'm still not clear on how the editor currently keeps track of the previous location of the cursor upon hitting movement keys, so that's not implemented (yet).

That'll be a nice addition!

I noticed that `g<a-j>` does not go beyond the last line on the display, but we can leave that, I don't think it's really important.

krobelus:
`g<a-l>` currently goes to the newline character, instead of the last character of the line. I believe this should be changed - I'll try to when I get to it.

lenormf:
Thanks for the feedback!

>`g<a-j>` does not go beyond the last line on the display

What do you mean by that? I tried jumping to the last display-line that was also part of the last buffer line in the file, and it worked.

>`g<a-l>` currently goes to the newline character, instead of the last character of the line

That seems to be the case when the current display line is leading to the end of the buffer line. It's inconsistent, I'll fix it.

lenormf:
Forgot to address the following:

>Also it does not select anything. I believe this should be changed

<kbd>g</kbd><kbd>a-l</kbd> selects the last character of the line, do you mean that it should extend the selection to it? <kbd>g</kbd><kbd>l</kbd> doesn't, if you want that behaviour you need to use <kbd>G</kbd><kbd>l</kbd>. Which is inconvenient, but I believe my display-line mappings are consistent with buffer line ones.

krobelus:
> > g<a-j> does not go beyond the last line on the display

> What do you mean by that? I tried jumping to the last display-line that was also part of the last buffer line in the file, and it worked.

When the cursor is on the last line of the screen, for example after using <kbd>gb</kbd>. Then <kbd>g<a-j></kbd> will do nothing.

> gl doesn't, if you want that behaviour you need to use Gl

Yeah, that's consistent, I was just confused because I mapped it to `<a-l>` in normal mode.
I can use something like `Zg<a-l><a-z>u`.

lenormf:
Got it, though I had trouble reproducing the behaviour you describe, because the editor randomly didn't render entire parts of the buffer. These new primitives assume the parts of the line you're trying to jump to are visible, what they should do otherwise is something I didn't consider. Maybe reverting to a normal <kdb>j</kbd>? Not sure that's what users would actually want to happen.

My patch should support <kbd>G</kbd><kbd>a-l</kbd>, I don't expect users to need combining selections. Does it not work with that combination?

krobelus:
> My patch should support Ga-l, I don't expect users to need combining selections. Does it not work with that combination?

That always extends the selection, as opposed to normal mode's <kbd>a-l</kbd> which creates a new selection between the cursor and the EOL, dropping the selected part left of the old cursor.

Screwtapello:
Just checking in - is there anything in particular stopping this PR from leaving "draft" status and becoming an actual proposal? I don't use auto-wrapped text *very* often, but when I do, I miss this functionality.

lenormf:
Thanks for the update. I want to go over the code again, clean it up, maybe pick better mappings… I will get to it, it's only a matter of when.

lenormf:
After re-reading the thread, here are unsolved issues that have been discussed:

- [ ] modifier based mappings in the `goto` menu are not convenient
- [ ] <kbd>g</kbd><kbd>a-j</kbd> on the last display line of the window doesn't select the next buffer line
- [ ] <kbd>g</kbd><kbd>a-j</kbd> and <kbd>g</kbd><kbd>a-k</kbd> don't jump back to the column that was selected prior to moving

lenormf:
What if all the window-based movement mappings were moved to an <kbd>a-g</kbd> menu?

<kbd>g</kbd><kbd>t</kbd> → <kbd>a-g</kbd><kbd>t</kbd>
<kbd>g</kbd><kbd>b</kbd> → <kbd>a-g</kbd><kbd>b</kbd>
<kbd>g</kbd><kbd>c</kbd> → <kbd>a-g</kbd><kbd>c</kbd>

<kbd>g</kbd><kbd>a-h</kbd> → <kbd>a-g</kbd><kbd>h</kbd>
<kbd>g</kbd><kbd>a-j</kbd> → <kbd>a-g</kbd><kbd>j</kbd>
<kbd>g</kbd><kbd>a-k</kbd> → <kbd>a-g</kbd><kbd>k</kbd>
<kbd>g</kbd><kbd>a-l</kbd> → <kbd>a-g</kbd><kbd>l</kbd>
<kbd>g</kbd><kbd>a-i</kbd> → <kbd>a-g</kbd><kbd>i</kbd>

Probably a bit too brutal, redundant with what could be a simple user mode, and doesn't really help with the modifier awkwardness when chaining the <kbd>g</kbd> based display line movements.

I also considered using another row of consecutive keys to handle display line movements, for example `y`, `u`, `i`, `o` (right above `h`, `j`, `k`, `l`), but `i` is already taken.

Screwtapello:
For myself, I don't really care about inconvenient default mappings; as long as they exist, I can map them in places of the regular movement commands (and add a wrap highlighter) when I'm editing a file that needs them. The current mappings are fine by me.

lenormf:
I factorised the code, I hope I didn't introduce more bugs.

The mappings stayed as they were (<kbd>alt</kbd> based).

Hitting <kbd>g</kbd><kbd>a-j</kbd> at the bottom of a window will jump to the next buffer-line when not at the end of the buffer.

Hitting <kbd>g</kbd><kbd>a-j</kbd> on the last display-line of the buffer doesn't jump to the last column any more.

Screwtapello:
Thanks for this update! I should go try and write some documentation or GitHub comments or something now, and see how it goes. :)

ficcdaf:
I would love to revive this! Sadly, I know nothing about C++ and Kakoune's codebase, so I'd really appreciate some help if anyone would be interested.

PR #3706, opened 2020-09-02
Gopls support for kakoune

### LINKED ISSUE #2976: gocode is not maintained
from `rc/tools/go/go-tools.kak`:

> ```
> # Provides integration of the following tools:
> # - gocode for code completion (github.com/nsf/gocode)
> ```

from `github.com/nsf/gocode`:
> this project is not maintained anymore, look for alternatives or forks if you need Go autocompletion tool

Now, that project has a maintained fork at [mdempsky/gocode](https://github.com/mdempsky/gocode), and there is also a fork of mdempsky's fork that supports Go modules at [stamblerre/gocode](https://github.com/stamblerre/gocode).

The natural thing to do first of all is to switch tho go-tools.kak comment to use mdempsky's fork. Then, we might consider if we want to use stamblerre's fork after, since it would allow for Go modules

voroskoi:
Kakoune go-tools.kak does the following:

- [x] ~~code completion (gocode)~~ _(broken)_
- [X] code formatting (goimports)
- [x] documentation display and code jump (gogetdoc)

Because of various API changes these tools tend to fail, it is recommended to replace them with **gopls**:
https://github.com/mawww/kakoune/issues/2976#issuecomment-502542757

I do not plan to implement code completion, as kak-lsp already supports gopls.
The code formatting part works already, I plan to add gopls definition and references support for a decent out-of-the-box go support in kakoune.

I suggest removing go-tools.kak from the repo when it is done.

closes #2976


rc/tools/go/gopls.kak
```diff
@@ -0,0 +1,49 @@
+hook -once global BufSetOption filetype=go %{
+    require-module gopls
+}
+
+evaluate-commands %sh{
+    if ! command -v gopls > /dev/null 2>&1; then
```
krobelus, line 6: I think this check should be done when the module is loaded.

rc/tools/go/gopls.kak
```diff
@@ -0,0 +1,49 @@
...
+        echo "fail Please install gopls or add to PATH!"
```
krobelus, line 7: I wonder if it's better to print that to the debug buffer, like go-tools does.

rc/tools/go/gopls.kak
```diff
@@ -0,0 +1,49 @@
...
+define-command -hidden -params 1 gopls-cmd \
+    -docstring "run various gopls commands" %{
+    evaluate-commands %sh{
+        dir=$(mktemp -d "${TMPDIR:-/tmp}"/kak-gopls.XXXXXXXX)
+        printf %s\\n "set-option buffer gopls_tmp_dir ${dir}"
+    }
+    evaluate-commands %sh{
+        dir=${kak_opt_gopls_tmp_dir}
+        eval "gopls $1 -w ${kak_buffile} 2> ${dir}/stderr"
+        if [ ! $? ]; then
```
krobelus, line 40: This condition is always false, use `[ $? -ne 0 ]`

krobelus (commented):
Makes sense - I'll try it soon!

rc/tools/go/gopls.kak
voroskoi: Thanks for your feedback! I have moved this part under the provide-module line.

rc/tools/go/gopls.kak
voroskoi: I like the big fat red error message as without gopls is useless. On the other hand if we autoload it with every go file it can be annoying if someone do not want to use it.
So I think the main question is if we want to autoload gopls.kak for go files, or we just keep the gopls-mode command. I prefer the latter, as gopls has no interactive part, so users will invoke it manually every time, autoloading is not useful IMO.

rc/tools/go/gopls.kak
voroskoi: Fixed, thanks.

rc/tools/go/gopls.kak
```diff
@@ -0,0 +1,98 @@
+# gopls.kak: gopls bindings for kakoune
+
+define-command -docstring "Enter gopls mode.
```
lenormf, line 3: You could make this docstring more readable by using the relatively unused format:

https://github.com/mawww/kakoune/blob/ec3d7c31040afaee65afde15832d9b0ed6b2224d/rc/tools/lint.kak#L2-L12

rc/tools/go/gopls.kak
```diff
@@ -0,0 +1,98 @@
...
+Bindings for gopls in kakoune.
+
+Best used with mapping like:
+	map global normal '<whatever>' ': gopls<ret>'
+" \
+gopls %{ require-module gopls; evaluate-commands 'enter-user-mode gopls' }
```
lenormf, line 9: Isn't `eval` redundant here?

rc/tools/go/gopls.kak
```diff
@@ -0,0 +1,98 @@
...
+
+provide-module gopls %§
+
+evaluate-commands %sh{
+    if ! command -v gopls > /dev/null 2>&1; then
+        echo "fail Please install gopls or add to PATH!"
+    fi
+}
+
+try %{ declare-user-mode gopls }
```
lenormf, line 19: I'm not sure built-in script should declare user modes. The name might conflict with a user-declared user mode, plus having the editor pick mappings (even out of convenience) doesn't sound good. @mawww

rc/tools/go/gopls.kak
```diff
@@ -0,0 +1,98 @@
...
+        dir=$(mktemp -d "${TMPDIR:-/tmp}"/kak-gopls.XXXXXXXX)
+        printf %s\\n "set-option buffer gopls_tmp_dir ${dir}"
+    }
+}
+
+define-command -hidden -params 1 gopls-cmd %{
+    gopls-prepare
+    evaluate-commands %sh{
+        dir=${kak_opt_gopls_tmp_dir}
+        eval "gopls $1 -w ${kak_buffile} 2> ${dir}/stderr"
```
lenormf, line 40: `${dir}` needs to be escaped (everywhere), in case `$TMPDIR` contains spaces for example.

rc/tools/go/gopls.kak
```diff
@@ -0,0 +1,98 @@
...
+        if [ $? -ne 0 ]; then
+            # show error messages in *debug* buffer
+            printf %s\\n "echo -debug '$(cat ${dir}/stderr)'"
```
lenormf, line 43: You could use the `%file{…}` expansion here.

rc/tools/go/gopls.kak
voroskoi: Thanks, I will.

rc/tools/go/gopls.kak
voroskoi: Sure, thanks.

rc/tools/go/gopls.kak
voroskoi: It is, fixed.

rc/tools/go/gopls.kak
voroskoi: Ok, I will rework this part something like in git.kak, so there will be only a gopls command.

rc/tools/go/gopls.kak
voroskoi: Should be fixed now.

rc/tools/go/gopls.kak
```diff
@@ -0,0 +1,116 @@
...
+    edit!
+}
+
+# gopls definition
+define-command -hidden -params 0 gopls-def %{
+    gopls-prepare
+    evaluate-commands %sh{
+        dir=${kak_opt_gopls_tmp_dir}
+        eval "gopls definition ${kak_buffile}:${kak_cursor_line}:${kak_cursor_column} \
+            |sed -e 's/-.*//' -e 's/:/ /g' > '${dir}/jump'"
```
krobelus, line 74: `gopls` will give an output like

```
/usr/lib/go/src/fmt/print.go:212:6-12: defined here as func fmt.Printf(format string, a ...interface{}) (n int, err error)
Printf formats according to a format specifier and writes to standard output.
It returns the number of bytes written and any write error encountered.
```

we only want to pick the first line, for example by using `sed 's/-.*//; s/:/ /g; q'`

rc/tools/go/gopls.kak
```diff
@@ -0,0 +1,116 @@
...
+            printf %s\\n "echo -debug %file{ '${dir}/stderr' }"
+        fi
+        rm -r "${dir}"
+    }
+    edit!
+}
+
+# gopls definition
+define-command -hidden -params 0 gopls-def %{
+    gopls-prepare
```
krobelus, line 70: The temporary directory+file seems unnecessary for this command. We can use a shell variable instead.

rc/tools/go/gopls.kak
```diff
@@ -0,0 +1,116 @@
...
+    }
+}
+
+# gopls references
+define-command -hidden -params 0 gopls-ref %{
+    gopls-prepare
+    evaluate-commands %sh{
+        dir=${kak_opt_gopls_tmp_dir}
+        mkfifo "${dir}/fifo"
+        ( gopls references ${kak_buffile}:${kak_cursor_line}:${kak_cursor_column} \
```
krobelus, line 90: Also here it would be nice if this command worked reliably in an unsaved buffer. I wonder if saving is the best option.

rc/tools/go/gopls.kak
```diff
@@ -0,0 +1,116 @@
...
+        printf %s\\n "set-option buffer gopls_tmp_dir ${dir}"
+    }
+}
+
+# gopls format/imports
+define-command -hidden -params 1 gopls-cmd %{
+    gopls-prepare
+    evaluate-commands %sh{
+        dir=${kak_opt_gopls_tmp_dir}
+        eval "gopls $1 -w ${kak_buffile} 2> '${dir}/stderr'"
```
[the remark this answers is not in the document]
voroskoi: Sure, eval is not necessary, but i do not understand the second part. Neither why is that good, nor how to do that. Could you explain?

rc/tools/go/gopls.kak
```diff
@@ -0,0 +1,116 @@
...
+define-command -hidden -params 0 gopls-ref %{
+    gopls-prepare
+    evaluate-commands %sh{
+        dir=${kak_opt_gopls_tmp_dir}
+        mkfifo "${dir}/fifo"
+        ( gopls references ${kak_buffile}:${kak_cursor_line}:${kak_cursor_column} \
+            > "${dir}/fifo" 2> /dev/null & ) > /dev/null 2>&1 < /dev/null
+        printf %s\\n "evaluate-commands -try-client '${kak_opt_toolsclient}' %{
+            edit! -fifo '${dir}/fifo' *gopls-refs*
+            set-option buffer filetype gopls
```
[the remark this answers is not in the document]
voroskoi: I have though depending on other modules is bad, but simpler for sure. Happy to do that.

rc/tools/go/gopls.kak
voroskoi: Indeed, rewritten.

rc/tools/go/gopls.kak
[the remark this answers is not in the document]
krobelus: Good point, scripts are generally standalone. Probably better to leave it like that, though I'm happy to hear other ideas.

rc/tools/go/gopls.kak
voroskoi: I was unable to get gops references working without a file. Also it would not find references in other files in the project. So I am confused here too.

rc/tools/go/gopls.kak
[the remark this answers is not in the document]
krobelus: If you type something like `fmt.Printf()` in your buffer and then use `gopls reference` on it, it won't work unless you save it.

The ideal solution would be to pipe the buffer contents to gopls, using the `set-register '|'` idiom like `format.kak`.
I don't know if that's possible, I'd assume gopls needs to receive the correct filename.
(`clang-format` has a nice option for that use case: `-assume-filename`)

PR #3712, opened 2020-09-04
Provide better support for Go language.

m-kru:
1. Highlight short variable declaration operator :=.
2. 'while' is not go keyword.
3. Auto insert ')', '}', when line ends with '(' or '{'.


m-kru:
@mawww why do tests fail? I have not even touched source files.

m-kru:
Fixing tests blocked by https://github.com/mawww/kakoune/issues/3713 .

mawww:
I dont think I have a copyright waiver commit from you, could you add one in that PR as this does add a sizeable amount of lines ? (see the CONTRIBUTING file)

m-kru:
@mawww done.

rc/filetype/go.kak
```diff
@@ -87,8 +89,8 @@ define-command -hidden go-indent-on-new-line %~
         try %{ execute-keys -draft <semicolon><c-s>k<a-x> s ^\h*\K/{2,} <ret> y<c-o>P<esc> }
         # indent after a switch's case/default statements
         try %[ execute-keys -draft k<a-x> <a-k> ^\h*(case|default).*:$ <ret> j<a-gt> ]
-        # indent after if|else|while|for
-        try %[ execute-keys -draft <semicolon><a-F>)MB <a-k> \A(if|else|while|for)\h*\(.*\)\h*\n\h*\n?\z <ret> s \A|.\z <ret> 1<a-&>1<a-space><a-gt> ]
+        # indent after if|else|for
+        try %[ execute-keys -draft <semicolon><a-F>)MB <a-k> \A(if|else|for)\h*\(.*\)\h*\n\h*\n?\z <ret> s \A|.\z <ret> 1<a-&>1<a-space><a-gt> ]
```
krobelus, line 93: Hold on, this line should be dropped altogether.
Go does not have [if](https://golang.org/ref/spec#If_statements)/else/for statements without a `{`-prefixed block so they are already handled above.

(Also this indentation only triggers for parenthesized conditions like `if (x)` but in Go it's more idiomatic to write `if x`.)

m-kru: You are right, looks like someone copied it without understanding. I have removed this line.

krobelus:
I generally like the autoinsert feature, it is less intrusive than in most IDEs.
However there is a case where it inserts an extra closing parenthesis.

```go
func min(%()
  a int,
)
```

Adding a newline here make us end up with

```go
func min(%()
  %()
)
  a int,
)
```

There is a generic plugin that handles this case: [auto-pairs.kak](https://github.com/alexherbo2/auto-pairs.kak).
So I wonder if we should not add this auto-insertion to the defaults. auto-pairs seems much superior because it works for all filetypes.

I'm not sure. I work with Go and would love to have insert hooks that handle some of the language specifics.
For example, in composite literals, list elements that are terminated by a newline must have a comma before that newline.
It would be really nice to insert that as I type:

```go
x := [][]int{
  []int{%()
}
```

```go
x := [][]int{
  []int{
    %()
  }, // Added a comma for list element here.
}
```

m-kru:
@krobelus There was an error in one regex. I have fixed this.

> So I wonder if we should not add this auto-insertion to the defaults.

To the defaults of kakoune? Well, I guess it is a question to @mawww. I think I do not care how the things work, I just want a decent support for basic stuff.

krobelus:
Thanks for the fix. It's definitely brings a better out-of-the-box experience so I'd take it.
The interaction with `auto-pairs.kak` is also bening - they don't mess up each other.

If we use something similar for other filetypes we can put it into a module and share most of the code.

m-kru:
I was wondering why kakoune does not support such things out of the box.

Issue #3340, opened 2020-02-04
Add a kakquote function to the prelude of shell blocks


Screwtapello:
Since Kakoune's quoting system was reworked, it's pretty easy to reliably quote Kakoune strings by just doubling apostrophes. In shell, it looks something like this:

    kakquote() {
        printf "'"
        printf "$*" | sed "s/'/''/g"
        printf "'"
    }

However, while working on #3336, I've copied and pasted this fragment into three or four shell blocks already; I expect it (or something like it) is already present in a bunch of other scripts too.

We already inject environment variables into shell blocks; how about we prepend this helpful and near-universally-useful function as well?

(it should be possible to implement this in pure POSIX shell without `sed`, but it would be too complex. If avoiding a fork turns out to be a performance improvement, it'd be nice to make the change once in Kakoune's shell prelude, and not have to copy/paste the change into thousands of existing Kakoune scripts)

lenormf:
Related #1224 #1430

alexherbo2:
@Screwtapello How about quoting each argument, instead of splatting?

``` sh
kak_escape() {
  for argument do
    printf "'"
    printf '%s' "$argument" | sed "s/'/''/g"
    printf "'"
    printf ' '
  done
}
```

alexherbo2:
@Screwtapello I’m using the following pattern for my scripts.

`rc/prelude.sh`

``` sh
kak_escape() {
  for argument do
    printf "'"
    printf '%s' "$argument" | sed "s/'/''/g"
    printf "'"
    printf ' '
  done
}
```

`rc/chronic.kak`

``` kak
declare-option -hidden str chronic_path %sh(dirname "$kak_source")

provide-module chronic %{
  define-command chronic -params .. -docstring 'Pipe to Chronic' %{
    evaluate-commands -save-regs '|' %{
      set-register | %sh{
        . "$kak_opt_chronic_path/prelude.sh"
        kak_quoted_arguments=$(kak_escape "$@")
        printf 'chronic %s' "$kak_quoted_arguments"
      }
      execute-keys '|<ret>'
    }
  }
}

require-module chronic
```

Screwtapello:
> How about quoting each argument, instead of splatting?

I like the modularity of having `kakquote` do exactly one thing, so I can use a printf format string to build up the command, and be sure I've quoted each argument properly:

```sh
printf "eval -try-client %s echo -markup {Information}%s" \
    "$(kakquote "$kak_client")"
[...]

PR #3716, opened 2020-09-06
Add :alt command for go files

deviant:
Like those for Lua, Moonscript, and Ruby, this adds the `:alt` command when editing a go file to jump between it and its corresponding test file.

e.g. editing a file `foo.go`, `:alt` would open `foo_test.go` if it exists, and `:alt` again would take you back.

### DIFF
--- a/rc/filetype/go.kak
+++ b/rc/filetype/go.kak
@@ -22,7 +22,12 @@ hook global WinSetOption filetype=go %{
     hook window InsertChar \{ -group go-indent go-indent-on-opening-curly-brace
     hook window InsertChar \} -group go-indent go-indent-on-closing-curly-brace
 
-    hook -once -always window WinSetOption filetype=.* %{ remove-hooks window go-.+ }
+    alias window alt go-alternative-file
+
+    hook -once -always window WinSetOption filetype=.* %{
+        remove-hooks window go-.+
+        unalias window alt go-alternative-file
+    }
 }
 
 hook -group go-highlight global WinSetOption filetype=go %{
@@ -72,6 +77,23 @@ evaluate-commands %sh{
 
 # Commands
 # ‾‾‾‾‾‾‾‾
+
+define-command go-alternative-file -docstring 'Jump to the alternate file (implementation ↔ test)' %{ evaluate-commands %sh{
+    case $kak_buffile in
+        *_test.go)
+            altfile=${kak_buffile%_test.go}.go
+            test ! -f "$altfile" && echo "fail 'implementation file not found'" && exit
+        ;;
+        *.go)
+            altfile=${kak_buffile%.go}_test.go
+            test ! -f "$altfile" && echo "fail 'test file not found'" && exit
+        ;;
+        *)
+            echo "fail 'alternative file not found'" && exit
+        ;;
+    esac
+    printf "edit -- '%s'" "$(printf %s "$altfile" | sed "s/'/''/g")"
+}}
 
 define-command -hidden go-indent-on-new-line %~
     evaluate-commands -draft -itersel %=


rc/filetype/go.kak
```diff
@@ -73,6 +78,23 @@ evaluate-commands %sh{
 # Commands
 # ‾‾‾‾‾‾‾‾
 
+define-command go-alternative-file -docstring 'Jump to the alternate file (implementation ↔ test)' %{ evaluate-commands %sh{
+    case $kak_buffile in
+        *_test.go)
+            altfile=$(echo $kak_buffile | sed 's/_test\.go$/.go/')
```
mawww, line 84: `altfile="${kak_buffile%_test.go}.go"` would avoid a subshell and forking a sed

rc/filetype/go.kak
```diff
@@ -73,6 +78,23 @@ evaluate-commands %sh{
...
+            test ! -f $altfile && echo "fail 'implementation file not found'" && exit
+        ;;
+        *.go)
+            altfile=$(echo $kak_buffile | sed 's/\.go$/_test.go/')
```
mawww, line 88: same

rc/filetype/go.kak
```diff
@@ -73,6 +78,23 @@ evaluate-commands %sh{
...
+            test ! -f $altfile && echo "fail 'test file not found'" && exit
+        ;;
+        *)
+            echo "fail 'alternative file not found'" && exit
+        ;;
+    esac
+    echo "edit $altfile"
```
mawww, line 95: quoting altfile would be safer, use printf to avoid portability issues with echo.

mawww (changes requested):

rc/filetype/go.kak
deviant: What portability issues are there here other than if `$altfile` contains backslashes? And how should this be quoted? If I use single quotes, the filename could break out of that with another; if I use braces, the filename could break out of that with a closing brace. This is overall handled very inconsistently through the rest of the codebase, here's a quick grep:

- `echo`, no quoting, no newline:
https://github.com/mawww/kakoune/blob/cfbff4e132c8d0aed8bcaf677b25f5ab3a9c3221/rc/filetype/ruby.kak#L141
- `printf`, no quoting, newline
https://github.com/mawww/kakoune/blob/cfbff4e132c8d0aed8bcaf677b25f5ab3a9c3221/rc/filetype/lua.kak#L85
https://github.com/mawww/kakoune/blob/cfbff4e132c8d0aed8bcaf677b25f5ab3a9c3221/rc/filetype/moon.kak#L77
https://github.com/mawww/kakoune/blob/cfbff4e132c8d0aed8bcaf677b25f5ab3a9c3221/rc/tools/go/go-tools.kak#L157-L160
- `printf`, single quotes, no newline
https://github.com/mawww/kakoune/blob/cfbff4e132c8d0aed8bcaf677b25f5ab3a9c3221/rc/tools/git.kak#L273-L280
- `printf`, single quotes, newline
https://github.com/mawww/kakoune/blob/cfbff4e132c8d0aed8bcaf677b25f5ab3a9c3221/rc/tools/rust/racer.kak#L80
- `printf`, brace quotes, newline:
https://github.com/mawww/kakoune/blob/cfbff4e132c8d0aed8bcaf677b25f5ab3a9c3221/rc/filetype/c-family.kak#L420
https://github.com/mawww/kakoune/blob/cfbff4e132c8d0aed8bcaf677b25f5ab3a9c3221/rc/filetype/c-family.kak#L431
- `awk`, `%|...|` quoting, implicit newline
https://github.com/mawww/kakoune/blob/cfbff4e132c8d0aed8bcaf677b25f5ab3a9c3221/rc/tools/ctags.kak#L40-L64

AFAICT, newlines are completely unneeded for kakoune to run the command, and could be dropped. If you pick a common style to use here, I can add a commit that makes these consistent.

lenormf: [Backslashes and flags are an issue with `echo`](https://www.etalabs.net/sh_tricks.html).

The path should be quoted for Kakoune, e.g. `printf 'edit "%s"' "$(printf %s "${altfile}" | sed 's,",&&,g')"`.

There's a lot of technical debt in the codebase, with unquoted variables, calls to `echo`, or lazy `printf %s\\n 'static string'`… I try to fix them when I see them in my own PRs.

Feel free to create a commit that addresses all these, but be aware that it's not gratifying work ;)

deviant: I'm aware of that page, and flags aren't relevant here. Is it really only possible to escape things properly by running stuff through sed first? You don't have a function for that or such?

krobelus: I think the quoting can also be done in a follow up.

> The path should be quoted for Kakoune, e.g. printf 'edit "%s"' "$(printf %s "${altfile}" | sed 's,",&&,g')".

I believe this should use single quotes so `%val{foo}` expansion does not take place.

> There's a lot of technical debt in the codebase, with unquoted variables

Yeah, this on rarely matters in practise but is totally avoidable.

One more thing I'd do is use `edit -- '%s'` so it works for files that start with `-`, though I guess I'd be okay with ignoring those, too.

> You don't have a function for that or such?

Not yet, that's #3340

This one is used sometimes:
`kakquote() { printf "%s" "$*" | sed "s/'/''/g; 1s/^/'/; \$s/\$/'/"; }`
using it will make it easy to refactor, once we have some kind of prelude.

There is a version without `sed` calls in [prelude.kak](https://github.com/alexherbo2/prelude.kak/blob/master/rc/prelude.sh) but I don't think that's warranted here.

deviant: @krobelus Thanks a lot for that function! Dropped that in and it works like a charm.

deviant: (Don't merge yet, I'm adding commits to clean up the other places I found.)

lenormf: Why declare a function that is made to escape a *list* of arguments, in stead of the simple&stupid command I posted above?

deviant: Because it would be better to have a common function that's used to escape things used throughout the entire codebase. It's easier to grep for instances of `kakquote` than to find every possible way someone's manually written out escaping logic.

lenormf: The function is only used in that one scope, in a single location. And it complicates the pattern passed to `sed` for no visible reason, hence my remark.

rc/filetype/go.kak
```diff
@@ -73,6 +78,24 @@ evaluate-commands %sh{
...
+        *.go)
+            altfile=${kak_buffile%.go}_test.go
+            test ! -f "$altfile" && echo "fail 'test file not found'" && exit
+        ;;
+        *)
+            echo "fail 'alternative file not found'" && exit
+        ;;
+    esac
+    kakquote() { printf "%s" "$*" | sed "s/'/''/g; 1s/^/'/; \$s/\$/'/"; }
+    printf %s "edit -- $(kakquote "$altfile")"
```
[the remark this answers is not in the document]
deviant: Seems like there's plenty of places in the codebase that does it like that. I wasn't aware that this was somehow 'improper'. This is a substitute for echo, after all...

rc/filetype/go.kak
deviant: "using it will make it easy to refactor, once we have some kind of prelude"

deviant: Actually, probably better to merge this for now, I'm busier than expected. I'm going to make an issue regarding consistent shell script styling across the codebase, since that would make it easier to know what the correct way to implement things would be and avoid bikeshedding.

krobelus: Yeah, the shell script changes can be non-obvious and should be discussed in a follow-up.
My argument that it's easier to refactor `kakquote` is kind of weak, since we'll have to inspect every use of `sed` anyway.
So maybe the simple solution by lenormf is a better fit for the script in its current form.
Anyway, it really doesn't matter.

deviant: ¯\\\_(ツ)\_/¯

PR #3719, opened 2020-09-08
Add illumos support for Kakoune

luka-devnull:
This PR adds support for compilation on the illumos family of operating systems. It also probably adds support for Solaris, but I'm not able to test that.

This is tested on OpenIndiana using GCC 7.5.0, GNU make 4.2.1, and GNU install from coreutils 8.31. I tried using illumos install(1M), but it acted really strangely so I added a little extra to the makefile to decide on what install to use and use ginstall on illumos. All you need to do really is run `gmake` and it works! Don't even need to redefine what C++ compiler to use, the default one works just fine.

kakoune running on OpenIndiana, showing uname -a output in the status line

If there's anything else that needs done, let me know!

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -80,6 +80,9 @@ else ifeq ($(os),OpenBSD)
 else ifneq (,$(findstring CYGWIN,$(os)))
     CPPFLAGS += -D_XOPEN_SOURCE=700
     LIBS += -lncursesw -ldbghelp
+else ifeq ($(os),SunOS)
+    LIBS += -lncursesw 
+    LDFLAGS += -lsocket -rdynamic
 else
     ifeq ($(PKG_CONFIG),)
     	$(error "pkg-config not found in PATH")
--- a/src/file.cc
+++ b/src/file.cc
@@ -654,6 +654,11 @@ String get_kak_binary_path()
     return buffer;
 #elif defined(__OpenBSD__)
     return KAK_BIN_PATH;
+#elif defined(__sun__)
+    ssize_t res = readlink("/proc/self/path/a.out", buffer, 2048);
+    kak_assert(res != -1);
+    buffer[res] = '\0';
+    return buffer;
 #else
 # error "finding executable path is not implemented on this platform"
 #endif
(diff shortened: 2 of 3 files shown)


luka-devnull:
Also just tested: clang++ 9.0.1, but it complained about `-rdynamic` being unused

I'm not sure what `-rdynamic` is for really, I just copied it from the bottom else statement for what I assume is used when building for Linux and it seemed to have built OK

lenormf:
Shouldn't this patch be sent out to the package manager? Creating a separate variable in the `Makefile` to pick an `install` binary to support an exotic distribution isn't appealing.

src/Makefile
```diff
@@ -157,26 +166,26 @@ distclean: clean
 	find ../doc -type f \( -name \*\\.gz -o -name \*\\.1 \) -exec rm -f '{}' +
 
 installdirs:
-	install -d $(bindir) \
+	$(INSTALL_CMD) -d $(bindir) \
```
krobelus, line 169: From a quick look at [`install(1m)`](https://illumos.org/man/1m/install): If this is the source of incompatibility, can't we just `mkdir -p` here instead to do what GNU and other `install`s do.

luka-devnull: I kid y'all not, I looked over it all again today, edited install(1m) back in and and it seems to be working just fine. Guess I'll just remove that little kludge!

src/file.cc
```diff
@@ -654,6 +654,11 @@ String get_kak_binary_path()
     return buffer;
 #elif defined(__OpenBSD__)
     return KAK_BIN_PATH;
+#elif defined(__sun__)
+    ssize_t res = readlink("/proc/self/path/a.out", buffer, 2048);
+    kak_assert(res != -1);
+    buffer[res] ='\0';
```
krobelus, line 660:
```suggestion
    buffer[res] = '\0';
```
:detective:

luka-devnull: :+1:

mawww:
Hello, could you rebase those into a single commit please ?

luka-devnull:
squished!

PR #3732, opened 2020-09-14
fix justfile highlighter

blaggacao:
This fixes serveral shortcomings of the current implementation:

- valid recipt definitions eg 
foo bar="quz":
where previously interrupted by justfile/double_string
and therefore they where not highlighted correctly

- global variable assignments where not captured at all

**Now**:
- see how globals in first line are shown red `nixEnvs`
- see how `deploy env="active"` is shown correctly

image

image

Also: 
image

### DIFF
--- a/rc/filetype/just.kak
+++ b/rc/filetype/just.kak
@@ -57,6 +57,9 @@ add-highlighter shared/justfile/body/interpreters/defaultshell default-region gr
 add-highlighter shared/justfile/body/interpreters/defaultshell/ ref sh
 add-highlighter shared/justfile/body/interpreters/defaultshell/ regex '^\h+(@)' 1:operator
 
+add-highlighter shared/justfile/body/interpreters/bash region '^\h+#!\h?/usr/bin/env bash' '^[^\h]' ref sh
+add-highlighter shared/justfile/body/interpreters/sh region '^\h+#!\h?/usr/bin/env sh' '^[^\h]' ref sh
+
 add-highlighter shared/justfile/body/ regex '(\{{2})([\w-]+)(\}{2})' 1:operator 2:variable 3:operator
 
 
--- a/rc/filetype/just.kak
+++ b/rc/filetype/just.kak
@@ -38,18 +38,23 @@ define-command -hidden just-indent-on-new-line %{
 # ‾‾‾‾‾‾‾‾‾‾‾‾
 
 add-highlighter shared/justfile regions
+
 add-highlighter shared/justfile/content default-region group
+add-highlighter shared/justfile/content/recipe regex '^@?([\w-]+)([^\n]*):(?!=)([^\n]*)' 1:function 2:meta 3:keyword
+add-highlighter shared/justfile/content/assignments regex ^([\w-]+\h*:=\h*[^\n]*) 1:meta
+add-highlighter shared/justfile/content/operator regex '((^@|:=|=|\+|\(|\)))' 1:operator
+add-highlighter shared/justfile/content/strings regions
+add-highlighter shared/justfile/content/strings/double region '"' (?<!\\)(\\\\)*" fill string
+add-highlighter shared/justfile/content/strings/single region "'" (?<!\\)(\\\\)*' fill string
+
 add-highlighter shared/justfile/comment  region '#' '$'  fill comment
-add-highlighter shared/justfile/double_string region '"' (?<!\\)(\\\\)*" fill string
-add-highlighter shared/justfile/single_string region "'" (?<!\\)(\\\\)*' fill string
+
 add-highlighter shared/justfile/inline   region '`' '`' ref sh
 
 add-highlighter shared/justfile/shell  region '^\h+' '^[^\h]' group
 add-highlighter shared/justfile/shell/ ref sh
 add-highlighter shared/justfile/shell/ regex '(\{{2})([\w-]+)(\}{2})' 1:operator 2:variable 3:operator
+add-highlighter shared/justfile/shell/ regex '^\h+(@)' 1:operator
 
-add-highlighter shared/justfile/content/ regex '^(@)?([\w-]+)(?:\s(.+))?\s?(:)(.+)?$' 1:operator 2:function 3:value 4:operator 5:type
-add-highlighter shared/justfile/content/ regex '([=+])' 1:operator
-add-highlighter shared/justfile/content/ regex '^([\w-]+)\s=' 1:value
 
 }
--- a/rc/filetype/just.kak
+++ b/rc/filetype/just.kak
@@ -51,10 +51,13 @@ add-highlighter shared/justfile/comment  region '#' '$'  fill comment
 
 add-highlighter shared/justfile/inline   region '`' '`' ref sh
 
-add-highlighter shared/justfile/shell  region '^\h+' '^[^\h]' group
-add-highlighter shared/justfile/shell/ ref sh
-add-highlighter shared/justfile/shell/ regex '(\{{2})([\w-]+)(\}{2})' 1:operator 2:variable 3:operator
-add-highlighter shared/justfile/shell/ regex '^\h+(@)' 1:operator
+add-highlighter shared/justfile/body  region '^\h+' '^[^\h]' group
+add-highlighter shared/justfile/body/interpreters regions
+add-highlighter shared/justfile/body/interpreters/defaultshell default-region group
+add-highlighter shared/justfile/body/interpreters/defaultshell/ ref sh
+add-highlighter shared/justfile/body/interpreters/defaultshell/ regex '^\h+(@)' 1:operator
+
+add-highlighter shared/justfile/body/ regex '(\{{2})([\w-]+)(\}{2})' 1:operator 2:variable 3:operator
 
 
 }


rc/filetype/just.kak
```diff
@@ -39,6 +39,8 @@ define-command -hidden just-indent-on-new-line %{
 
 add-highlighter shared/justfile regions
 add-highlighter shared/justfile/content default-region group
+add-highlighter shared/justfile/recipe region '^[\w-]+(\s.+)?\s?:(?!=).*' '$' group
```
lenormf, line 42: Is the intent to match across lines? Maybe `\s` could be turned into `\h` to make it clearer.

And why declare a region, when this line could be a one-line regex? For clarity, or is there a technical obstacle?

blaggacao: No the intent is only to match one single line (the recipe line, similar to a `make` target line). `\s` is indeed a single white space within the recipe regex.

> And why declare a region, when this line could be a one-line regex? For clarity, or is there a technical obstacle?

Further down, I declare an operator against this region. since there can be multiple arguments, with or without operators, the region regex would become overly complex: right now args are simly catched by: `(\s.+)?\s?`

and then:
`add-highlighter shared/justfile/recipe/ regex '([=+])' 1:operator` - simple but effective

blaggacao: `ssh ctx="current": (_check ctx)` &mdash; example recipe line

krobelus: I'd use `[^\n]` instead of `.` to prevent matching newlines. Also, quotes are
usually not necessary for regex literals. I believe it can be simplified to
(untested)

```
add-highlighter shared/justfile/recipe region ^[\w-]+(\h[^\n]*|\h?):(?!=) $ group
```

I tink comments are not highlighted correctly in `ssh: # comment`.

Maybe you can do without a region, and just match all parts in one regex.

blaggacao: Oh I was confused about `\s` & `\h` - indeed it should be `\h`

rc/filetype/just.kak
```diff
@@ -39,6 +39,8 @@ define-command -hidden just-indent-on-new-line %{
...
+add-highlighter shared/justfile/globals region '^([\w-]+)\s:=.*' '$' regex '^([\w-]+)\s(:=)\s(.+)' 1:meta 2:operator 3:value
```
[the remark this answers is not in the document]
blaggacao: Thanks fro this educative suggestion!

rc/filetype/just.kak
blaggacao, line 42:
```suggestion
add-highlighter shared/justfile/recipe region ^[\w-]+(\h[^\n]*|\h?):(?!=) $ group
```

blaggacao: @krobelus I think this was your suggestion. It looks good to me, will test shortly.

blaggacao:
@krobelus 
> I tink comments are not highlighted correctly in `ssh: # comment.`

you are right, will have a look

rc/filetype/just.kak
```diff
@@ -39,6 +39,8 @@ define-command -hidden just-indent-on-new-line %{
 
 add-highlighter shared/justfile regions
 add-highlighter shared/justfile/content default-region group
+add-highlighter shared/justfile/recipe region ^[\w-]+(\h[^\n]*|\h?):(?!=) $ group
+add-highlighter shared/justfile/contents/ regex ^([\w-]+)\h*(:=)\h*([^\n]+) 1:meta 2:operator 3:value
```
blaggacao, line 43:
```suggestion
add-highlighter shared/justfile/content/ regex ^([\w-]+)\h*(:=)\h*([^\n]+) 1:meta 2:operator 3:value
```

rc/filetype/just.kak
```diff
@@ -39,6 +39,8 @@ define-command -hidden just-indent-on-new-line %{
 
 add-highlighter shared/justfile regions
 add-highlighter shared/justfile/content default-region group
+add-highlighter shared/justfile/recipe region ^[\w-]+(\h[^\n]*|\h?):(?!=) $ group
+add-highlighter shared/justfile/content/ regex ^([\w-]+)\h*(:=)\h*([^\n]+) 1:meta 2:operator 3:value
```
blaggacao, line 43:
```suggestion
add-highlighter shared/justfile/content/globals regex ^([\w-]+)\h*(:=)\h*([^\n]+) 1:meta 2:operator 3:value
```If this is not reagino, then we need to remove double_string and single_string they take precedence and break this matching

Comment (blaggacao) [inline rc/filetype/just.kak:43]:
```suggestion
add-highlighter shared/justfile/content/recipe regex '^(@)?([\w-]+)((?:\h[\w-]+((=)("[^\n]*?"))?)*)(:(?!=))(\h(?:[\w-]+|\([\w-]+\h[\w-]+\)))?' 1:operator 2:function 3:value 5:operator 7:operator 8:type
add-highlighter shared/justfile/content/ regex ^([\w-]+)\h*(:=)\h*([^\n]+) 1:meta 2:operator 3:value
```

If this is not regino, then we need to remove double_string and single_string they take precedence and break this matching

rc/filetype/just.kak
```diff
@@ -39,6 +39,8 @@ define-command -hidden just-indent-on-new-line %{
 
 add-highlighter shared/justfile regions
 add-highlighter shared/justfile/content default-region group
+add-highlighter shared/justfile/content/recipe regex '^(@)?([\w-]+)((?:\h[\w-]+(?:(=)("[^\n]*?"))?)*)(:(?!=))(\h(?:[\w-]+|\([\w-]+\h[\w-]+\)))?' 1:operator 2:function 3:value 4:operator 6:operator 7:type
+add-highlighter shared/justfile/content/ regex ^([\w-]+)\h*(:=)\h*([^\n]+) 1:meta 2:operator 3:value
```
blaggacao, line 43:
```suggestion
add-highlighter shared/justfile/content/globals regex ^([\w-]+)\h*(:=)\h*([^\n]+) 1:meta 2:operator 3:value
```

rc/filetype/just.kak
```diff
@@ -48,8 +50,7 @@ add-highlighter shared/justfile/shell  region '^\h+' '^[^\h]' group
 add-highlighter shared/justfile/shell/ ref sh
 add-highlighter shared/justfile/shell/ regex '(\{{2})([\w-]+)(\}{2})' 1:operator 2:variable 3:operator
 
-add-highlighter shared/justfile/content/ regex '^(@)?([\w-]+)(?:\s(.+))?\s?(:)(.+)?$' 1:operator 2:function 3:value 4:operator 5:type
-add-highlighter shared/justfile/content/ regex '([=+])' 1:operator
-add-highlighter shared/justfile/content/ regex '^([\w-]+)\s=' 1:value
+add-highlighter shared/justfile/recipe/ regex '^(@)?([\w-]+)(?:\s(.+))?\s?(:)(.+)?$' 1:operator 2:function 3:value 4:operator 5:type
+add-highlighter shared/justfile/recipe/ regex '([=+])' 1:operator
```
blaggacao, line 54:
```suggestion
add-highlighter shared/justfile/content/operator regex '([:=|=+])' 1:operator
```

blaggacao:
Now it looks even better:  &mdash; Thanks for your suggestions!

image

The only problem now is that the operator _overlay_ also hits within _value_ group: &mdash; not too bad, though
image

blaggacao:
Solved that, too:

image

blaggacao:
`=` - Operator: OK

image

No operator within string: OK

image

rc/filetype/just.kak
```diff
@@ -38,18 +38,19 @@ define-command -hidden just-indent-on-new-line %{
 # ‾‾‾‾‾‾‾‾‾‾‾‾
 
 add-highlighter shared/justfile regions
+
 add-highlighter shared/justfile/content default-region group
+add-highlighter shared/justfile/content/recipe regex '^@?([\w-]+)((?:\h[\w-]+(?:=(?:"[^\n]*?"))?)*):(?!=)(\h(?:[\w-]+|\([\w-]+\h[\w-]+\)))?' 1:function 2:value 3:type
+add-highlighter shared/justfile/content/globals regex ^([\w-]+)\h*:=\h*(["|'][^\n]*?["|'])((?:\h*\+\h*[\w-]+)*) 1:meta 2:value 3:meta
```
krobelus, line 44: The [grammar](https://github.com/casey/just/blob/master/GRAMMAR.md) calls them "assignments".

rc/filetype/just.kak
```diff
@@ -38,18 +38,19 @@ define-command -hidden just-indent-on-new-line %{
...
+add-highlighter shared/justfile/content/operator regex '((^@|\h:=\h|=(?=")|\h\+\h))' 1:operator
```
krobelus, line 45: are the spaces around `:=` and `+` really required?

rc/filetype/just.kak
```diff
@@ -38,18 +38,19 @@ define-command -hidden just-indent-on-new-line %{
...
+
 add-highlighter shared/justfile/comment  region '#' '$'  fill comment
-add-highlighter shared/justfile/double_string region '"' (?<!\\)(\\\\)*" fill string
```
krobelus, line 43: It's a bit strange that strings are no longer colored as `string` like in other languages, but that's okay.

krobelus (commented):
I guess you can add keyword highlighting for things like `export` or `alias` at the beginning of a line. 

> as an aside, how can I override this in `kakrc` so I can keep it tweaking a week or so?

There are multiple ways to do that, I would just run `src/kak` from the repo.

rc/filetype/just.kak
blaggacao: Just figured out I can define another `strings` regions within the content regions and then define a `dq` region that matches those strings. Yeah! That will simplify everything a lot...

rc/filetype/just.kak
blaggacao: I put them as nested regions into the content region, that works perfeclty and makes a lot of things simpler to paint.

blaggacao:
Works:
image

Works:
image

Works:
image

blaggacao:
I think this is as far as I can get. Further ideas for future work:

- multiline string assignments
- detect recipe shebang and detect popular host scripting languages (_python_, _perl_, etc.)

rc/filetype/just.kak
```diff
@@ -38,18 +38,22 @@ define-command -hidden just-indent-on-new-line %{
 # ‾‾‾‾‾‾‾‾‾‾‾‾
 
 add-highlighter shared/justfile regions
+
 add-highlighter shared/justfile/content default-region group
+add-highlighter shared/justfile/content/recipe regex '^@?([\w-]+)([^\n]*):(?!=)([^\n]*)' 1:function 2:meta 3:keyword
+add-highlighter shared/justfile/content/assignments regex ^([\w-]+\h*:=\h*[^\n]*) 1:meta
+add-highlighter shared/justfile/content/operator regex '((^@|:=|=|\+|\(|\)))' 1:operator
+add-highlighter shared/justfile/content/strings regions
+add-highlighter shared/justfile/content/strings/dq region '"' (?<!\\)(\\\\)*" fill string
```
krobelus, line 47: Is there an example where this behaves differently from the old `shared/justfile/double_string`?
If yes, it should probably still be `shared/justfile/content/doule_string` to match other `rc` files.

krobelus (commented):
looks solid! I think you can squash your commits to get a cleaner history.

rc/filetype/just.kak
blaggacao: Yes, the golabl spaced region indiscriminately brakes the content group.

so this hfailed - which was the motivation to get this PR started:

`ssh m="":` - no syntax highlighting was done because the `""` string region broke up the possiblilty do match the line in the default region.

blaggacao: > If yes, it should probably still be shared/justfile/content/doule_string to match other rc files.

the region needs its own name. it would be 

``shared/justfile/content/strings/doule_string` &mdash; is that what you suggest for more homogenous grepping of the codebase?

blaggacao:
I squashed the commit history :+1:

rc/filetype/just.kak
krobelus: Oh yeah, that makes sense!

krobelus: > the region needs its own name. it would be

I see, maybe `strings/double` but it's also fine to leave it like this imo

blaggacao: ok adopted: 
- `strings/double`
- `strings/single`

blaggacao:
After my last two commits, when I try to implement `ref python` highlighter, it doesn't do anything, only with the `sh` highlighter it seems to work. Same for `ref nix` &mdash; while  the regions are identified correctly by their shebangs.

blaggacao:
@mawww thanks for merging! Much appreciated!

PR #3740, opened 2020-09-16
rc modeline: Prevent command execution

### LINKED ISSUE #3735: Arbitrary code execution using modeline
### Steps 
1. write in scratch:
```python
# kak: tabstop=2;set-option	buffer	autowrap_column	1
```
2. exec: modeline-parse

### Outcome 
`autowrap_column` is set to `1`.

Using `;` and tabs as space separator, we are able to bypass the `key` verification in the modeline parser.
It is parsed as `value` but is then interpreted as a separate command during eval.
It opens to arbritary code execution.

@lenormf suggested in https://github.com/mawww/kakoune/issues/3734#issuecomment-692797589:
> [...] quoting the value before passing the set command to eval should fix it.

lenormf:
This commit prevents specially crafted modelines from making the
editor execute arbitrary Kakoune commands.

By using a tabulation character as a separator, commands can be
injected in the value of the options listed in the modeline. For
example:

	# kak: tabstop=2;set-option	buffer	pwned	yes

Fixes #3735.

More commits unrelated to the issue were queued to this PR instead of getting their individual PR for simplicity's sake.


rc/detection/modeline.kak
```diff
@@ -30,14 +30,16 @@ define-command -hidden modeline-parse-impl %{
                     case "${value}" in
                         unix) tr="eolformat lf";;
                         dos) tr="eolformat crlf";;
-                        *) printf %s\\n "echo -debug 'Unsupported file format: ${value}'";;
+                        *) printf 'echo -debug %s' "$(kakquote "Unsupported file format: ${value}")" \
+                               | kak -p "${kak_session}";;
```
krobelus, line 34: Would printing to stderr do the same? I'm not sure.

lenormf: When writing the message to stderr, this is what the debug buffer looks like:

```
shell stderr: <<<
Unsupported modeline format: # foo: hey

>>>
shell stderr: <<<
Unsupported modeline format: shell stderr: <<<

>>>
```

krobelus: Oh sorry, this would have been easy to check myself. Thanks for elaborating!

PR #3746, opened 2020-09-18
Improve '}' auto inserting for go language.

m-kru:
Adding "} else if ... {" was not correctly handled.


rc/filetype/go.kak
```diff
@@ -134,7 +134,7 @@ define-command -hidden go-insert-on-new-line %[
             try %[ execute-keys -draft k<a-x>s^\h+<ret>"xy ] catch %[ reg x '' ]
             try %[
                 # Validate previous line and that it is not closed yet.
-                execute-keys -draft k<a-x> <a-k>^<c-r>x.*\{\h*\(?\h*$<ret> J}iJ<a-x> <a-K>^<c-r>x(\)?\h*\})$<ret>
+                execute-keys -draft k<a-x> <a-k>^<c-r>x.*\{\h*\(?\h*$<ret> j}iJ<a-x> <a-K>^<c-r>x(\)?\h*\}.*)$<ret>
```
krobelus, line 137: Instead of `.*` I'd normally use `[^\n]`. Actually, we know that the selection is a line, so why not just omit the `.*$`, since it will always match. Also I don't think the capturing parentheses do anything here.

```suggestion
                execute-keys -draft k<a-x> <a-k>^<c-r>x.*\{\h*\(?\h*$<ret> j}iJ<a-x> <a-K>^<c-r>x\)?\h*\}<ret>
```

krobelus (changes requested):
Thanks for the quick fix. I believe we also need to handle `)` the same way:
Consider adding a newline after `%()` in

```go
	a := []int{
		someFunction(%()
		),
	}
```

m-kru:
@krobelus you are right. I have applied changes.

PR #3749, opened 2020-09-18
*sh.kak: do not highlight expansions if their $ is escaped

krobelus:


lenormf:
Why? The dollar sign is as much part of the expansion as the word, it's not an unrelated artefact — the same way `%` is not separable from `opt{…}`.

krobelus:
Right, but I didn't change the highlighting for `$`, but disabled highlighting for the entire expansion if the dollar sign is escaped.

Before:
image

After:
image

lenormf:
Right, my bad! I misread the title and didn't understand the patch.

PR #3754, opened 2020-09-21
spell.kak: Present spelling suggestions with :prompt

Screwtapello:
Previously, spelling suggestions were presented with the :menu command, requiring the user to cycle through wild and fanciful alternatives to get to the one they wanted. Now, we present suggestions with the :prompt command, which allows the user to type to filter down the list, and also to customise the replacement after they've chose it (perhaps to fix capitalisation or add apostrophe-S).

We also use the mispelled word as the initial content of the prompt. That filters out the wildest alternatives by default, and allows the user to edit the original word instead of forcing them to choose from among the suggestions. To get the full list of suggestions, it's easy enough to just backspace until the word you want appears in the list.

### DIFF
--- a/rc/tools/spell.kak
+++ b/rc/tools/spell.kak
@@ -142,20 +142,27 @@ define-command spell-next %{ evaluate-commands %sh{
 
 define-command \
     -docstring "Suggest replacement words for the current selection, against the last language used by the spell-check command" \
-    spell-replace %{ evaluate-commands %sh{
-    options=""
-    if [ -n "$kak_opt_spell_last_lang" ]; then
-        options="-l '$kak_opt_spell_last_lang'"
-    fi
-    suggestions=$(printf %s "$kak_selection" | eval "aspell -a $options" | grep '^&' | cut -d: -f2)
-    menu=$(printf %s "${suggestions#?}" | awk -F', ' '
-        {
-            for (i=1; i<=NF; i++)
-                printf "%s", "%{"$i"}" "%{execute-keys -itersel %{c"$i"<esc>be}}"
+    spell-replace %{
+    prompt \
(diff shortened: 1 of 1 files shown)


rc/tools/spell.kak
```diff
@@ -142,20 +142,27 @@ define-command spell-next %{ evaluate-commands %sh{
...
+    prompt \
+        -init %val{selection} \
+        -shell-script-candidates %{
+            options=""
+            if [ -n "$kak_opt_spell_last_lang" ]; then
+                options="-l '$kak_opt_spell_last_lang'"
+            fi
+            printf %s "$kak_selection" |
+                eval "aspell -a $options" |
+                kak -f '<a-s><a-K>^&<ret>d%s^[^:]*: <ret>d%s, <ret>c<ret><esc>'
```
lenormf, line 155: Could we not re-implement this with Awk? Otherwise, using `-n` as well would help loading time.

Screwtapello: Using `-n` wouldn't help; Kakoune checks for and executes `-f` long before it even bothers checking for `-n`.

I used `kak -f` because I couldn't immediately see how to perform the same operations with more primitive tools, but now that I've stepped away for a few hours I figured out how to do it with pure `sed`.

mawww:
This goes in the right direction, I have been hoping to get rid of `:menu` at some point, as `:prompt` almost superseeds it.

lenormf:
I am sceptical of this change because `menu` preserves the order of the candidates returned by the spell checker. Their order is predicated on the confidence that they make more sense than the current word, so it makes more sense to me to want to advance to the next best candidate with a simple <kbd>tab</kbd>, rather than picking one out.

PR #3755, opened 2020-09-21
fix(ninja filetype): add subninja and include declarations and fix builddir highlights

tototest99:


rc/filetype/ninja.kak
```diff
@@ -52,29 +52,32 @@ add-highlighter shared/ninja/command/linebreak regex '\$$' 0:operator
 add-highlighter shared/ninja/command/variables regex '\$\w+|\$\{\w+\}' 0:value
 
 # `build`
-add-highlighter shared/ninja/build region '^build' '\n' group
-add-highlighter shared/ninja/build/build regex '^build\h+' 0:keyword
+add-highlighter shared/ninja/build region '^build ' '\n' group
```
mawww, line 55: maybe `^build\b` would be more correct ?

Issue #3587, opened 2020-07-09
Suggestion: require :w! for writing to a file other than the current file


Quelklef:
Apologies if this has been brought up before; I was unable to find a related issue

### Feature

If `:w FILE` is used where `FILE` is not the currently open file, refuse to write; require `:w!`.

### Usecase

QWERTY places `w` and `e` adjacent to each other. Several times when working with multiple files, I have meant to switch files by typing `:e OTHER_FILE` but accidentally typed `:w OTHER_FILE`, mistakenly overwriting the other file.

Most of the time, that file is in Kak's memory, so I am able to undo back to safety. Once, though, it wasn't. In that case, I was luckily able to `git checkout` without losing work. However, neither of these safeties are guaranteed. As such, it would be nice to include a guard against such an easy typo.

CharlesGueunet:
You may find [this discourse topic](https://discuss.kakoune.com/t/prevent-write-to-overrite-existing-file-by-default/992) interesting.

Here is [the final version](https://github.com/CharlesGueunet/kak/blob/af0de673c762e1790af9d78a6b58a56f3345c27b/kakrc#L126) on my config files.

Could this be a (really small) plugin ?

alexherbo2:
Same happened to me.

krobelus:
Refusing to write if the file exists sounds like a reasonable behavior.
Either chaning the `w` alias or `write` itself.

Quelklef:
@CharlesGueunet Thank you for the snippet!

caksoylar:
Sounds like https://github.com/mawww/kakoune/commit/40e3614cf458a982c783e899143bef81736faa27 resolve this, right?

mawww:
@voroskoi seems I ended up with something very similar to your previous PR, changed my mind about this seeing that it seems to regularly come up and that vim does refuse to overwrite by default as well.

Issue #3717, opened 2020-09-06
ask before overwitting a file with `:w`


sucrecacao:
### Feature 

When executing the command `:w foo.txt`, if `foo.txt` already exist ask for confirmation before overwriting it .

### Usecase 
Prevent lost of data.

krobelus:
Duplicate of #3587

PR #3762, opened 2020-09-24
write: do not overwrite existing file

voroskoi:
Do not let Kakoune to overwrite existing files with the buffer content without using w! or write! if the argument does not match the buffer name.
Writing without arguments works as before.

Fixes: #3587 #3717

- [x] Code changes
- [x] Make failing tests work again
- [x] Add tests for this change


lenormf:
The random side effect is not appealing, this is better handled with a custom command.

voroskoi:
Given the fact that `e` and `w` are next to each other at least on a QWERTY/QWERTZ keyboard this behavior is desired even in `kak -n` mode.
There are already two bug reports, a [forum thread](https://discuss.kakoune.com/t/prevent-write-to-overrite-existing-file-by-default/992) based on this issue and nvi/vim/neovim works this way too.

The tests (3 failing for me) has to be fixed, maybe there are plugins out there need a `s/write/write!/`, but I still think it is the right way.

- [ ] Make failing tests work again.

lenormf:
My reservation about this change is that the side effect is unpredictable unless you know about/experience it beforehand. The primitive is called `write`, and that's what the editor should do (or attempt to, c.f. filesystem permission errors) when you use it.

It's also more elegant than inverting the paradigm and delegating actual writing to `write!`.

Look at how `edit -existing` does it: the side effect is opt-in, and expected.

I also have fat finger issues, that's why I have the following (somewhat related) instruction in my configuration:

```kak
unalias global wq write-quit
```

voroskoi:
First of all, thank You for your valuable feedback, I appreciate.

I do understand your concern, but `edit` and `write` is two different things:
 - Using `edit -existing` will guarantee that you do not create a new file, so the default behavior do not make any harm to the user, worst case he/she ends up with an empty buffer.
 - OTOH `write existing.file` by default overwrites someone previous work on a file, which can be really annoying.

If you look at `edit` it will not reopen an existing file if it has a modified version in the buffer, even if reverting the reload is just pressing `u` once. So `edit` is safety first by default.

My the main point would be making write a bit safer. I can change the patch to add `-existing` option to `write` but I would like to **change default action**, adding `-existing` would allow *overwriting* without using `write!`. Does it sound acceptable for You?

I would like to also note that this would only change the behavior if you want to use `write some.file` (this is "save as" in GUI apps) with an existing file **and** that is not the one open in the buffer. So severity is much lower, you do not have to use `write!` all the time as you implied. I would not call it inverting the paradigm.

You are absolutely right, this could be handled in `kakrc` too (as seen in the linked forum thread), but as I said before, I think we should change the default behavior.

voroskoi:
I do not think it will be merged.

PR #3767, opened 2020-09-27
Rust indent

pickfire:
Before
```rust
foo%( )
```
```rust
foo
.bar
```
After
```rust
foo%( )
```
```rust
foo
    .bar
```
For operation `c bar`. I put checks on operators like `.`, and even those `-` or `&&`.


test/indent/rust/after-where/in
```diff
@@ -1,5 +1,3 @@
-    impl X for T where%( )
```
mawww, line 1: why did this test case go ?

pickfire: It is not important, also `rustfmt` does not forward with `where` at the back IIRC.

PR #3773, opened 2020-10-03
lua.kak fixes: Standardize indentation and correct lua-trim-indent hook group name

SeerLite:
Edit: [See comments below](https://github.com/mawww/kakoune/pull/3773#issuecomment-703153455)
~~Undoes some of the changes from #3419 by @Anfid:~~ 

> * ~~Keep indent on new line similar to previous **non-empty** line~~
> * ~~Add extra indent after opening block on previous **non-empty** line~~

~~This, so it acts like other Kakoune filetypes when inserting newlines on empty lines. I'm pretty sure the old behavior can be replicated in user configuration by overriding some functions (correct me if I'm wrong).~~

---
Fixes the `-trim-indent` hook name which was driving me crazy.
Also fixes a small [indentation bug](https://github.com/mawww/kakoune/pull/3773#issuecomment-703135577)

### DIFF
--- a/rc/filetype/lua.kak
+++ b/rc/filetype/lua.kak
@@ -14,7 +14,7 @@ hook global BufCreate .*[.](lua) %{
 hook global WinSetOption filetype=lua %{
     require-module lua
 
-    hook window ModeChange pop:insert:.* -group lua-indent lua-trim-indent
+    hook window ModeChange pop:insert:.* -group lua-trim-indent lua-trim-indent
     hook window InsertChar .* -group lua-indent lua-indent-on-char
     hook window InsertChar \n -group lua-indent lua-indent-on-new-line
     hook window InsertChar \n -group lua-insert lua-insert-on-new-line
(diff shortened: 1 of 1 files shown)


SeerLite:
I forgot to add the copyright waiver (also happened on my last PR). I hope it's fine if I add it now.

Anfid:
To be fair, i don't understand any reason to revert this behavior, except for consistency. If it is indeed so, I could start working on implementing similar behavior for other filetypes, although it may take some time.

SeerLite:
And this is why I pinged you <3

I was actually gonna make an issue first to ask what everyone thought about getting this behavior on other filetypes. Then I realized Python wouldn't work well with this because of its indentation syntax, so I just sent the PR to see what happened :o

Personally, I don't mind the behavior that much. I just noticed because of a bug it causes, where:

```lua
if true then|
end
```
(`|` is the cursor)
if you insert more than one newline, the indentation is gonna increase more than expected. I don't know how to fix this with regex myself.
(I only discovered this because I was messing around with some random file trying to fix `-trim-indent`, it isn't really an important bug).
I decided to just revert back to the original behavior, also because I thought it was more consistent with other filetypes.

It does sound like a nice feature to have for other filetypes, but I feel like it would cause some unexpected behavior (IMO this all comes down to how hacky kakscript is).
Maybe I really should make an issue to ask everyone about it? (if people don't see this here)

Anfid:
Oh, yeah, I see this bug now. It can be easily fixed just by removing indent before trying to preserve it. Swapping lines 104 and 106 does the thing.

I also agree that trim-indent should be moved to separate group.

As of python and other languages with similar behavior, I do not know any decent way to gather opinions of other people, other than implementing it and waiting for complains :)
Such issues seem to sink without much responses, but I would rather wait for @mawww to put his opinion on this topic.

Anfid:
As for me, I will slowly start working on implementing similar behavior for languages that I am more or less certain could benefit from it

SeerLite:
I just checked neovim and it looks like it also does this. It does it for other filetypes too (I tested shell, Lua and C).
So yeah I'm convinced, I think it's okay to do this on Kakoune. I'll play around with some filetypes too and see how to work it out (I'm not sure that I'll be trying to implement it though, kakscript is still really confusing to me)

Thanks for the easy fix for the bug too haha, I don't know how I didn't see that.

PR #3788, opened 2020-10-08
Update make.kak

alexherbo2:


lenormf:
As with #3789, `:mne<tab>`/`:mpe<tab>` take me to `:make-next-error`/`:make-previous-error`, making the aliases redundant.

Turning the hook into a mapping is probably a good move, but that would have to be discussed I think because it has the potential to break some configurations.

And making that same change in two different PRs would be confusing to users looking at the history of those files, better have one commit making the change and explaining the motivation.

alexherbo2:
4 vs 6 keystrokes, involving <kbd>Tab</kbd> and <kbd>Enter</kbd> is quite different.  It also depends on the list of available commands.

PR #3789, opened 2020-10-08
Add aliases for rename-buffer and rename-session

alexherbo2:
Completes alias for rename-client

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -818,7 +818,7 @@ const CommandDesc force_delete_buffer_cmd = {
 
 const CommandDesc rename_buffer_cmd = {
     "rename-buffer",
-    nullptr,
+    "nb",
     "rename-buffer <name>: change current buffer name",
     ParameterDesc{
         {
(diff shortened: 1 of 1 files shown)


lenormf:
`rename-session` is already complete-able easily when typing `:rs<tab>`, that's one more stroke than `:ns<ret>`. Only one more tabulation for `rename-buffer`, because another command is suggested as first candidate.

Why those specific aliases, too? I'm assuming they stand for "name session/client", but that goes against the name pattern of built-in commands (in which case you'd generally use fuzzy matching as above), and those aliases are not Vim ones either (to provide some sort of transitional API from one to the other).

alexherbo2:
`rename-client` has an alias on `nc`.

lenormf:
I see that this was introduced in 8aaeaa3187aa6cda3f2b264d474f15b4486c44b5 (6 years ago). The topic should probably have been discussed when the command names were normalised (I can't find the issue, only this commit dcd733e5fab9016afbf9d8b5c5774053c1ce24af), but it's possible that it was and the decision turned out to be not to remove anything, only add.

Regardless, if anything I think we should *remove* this alias rather than assign one to all renaming commands, unless you have pragmatic reasons to do otherwise?

mawww:
Agreed with lenormf, I'd rather remove the "nc" alias.

PR #3793, opened 2020-10-09
Perl regex highlighters

### LINKED ISSUE #3792: Perl Highlighting Error
### Steps 
1. Open a perl file
2. Write something like
```
my $str = 'String with "double" quotes';
   $str =~ s{"}{&quot;};
#            ^
```
The statement above is highlighted like the marked double quote is the start of the string. This should be handled like a proper regex. Perl syntax is ~~horrifyingly weird~~ good.
### Example
image

### Expected
Should match other sregexes, look like this:
image

Carsenal:
Added support for curly brace sregexes in Perl.

Closes #3792

### DIFF
--- a/rc/filetype/perl.kak
+++ b/rc/filetype/perl.kak
@@ -44,6 +44,7 @@ add-highlighter shared/perl/comment        region (?<!\$)(?<!\\)#   $           
 
 add-highlighter shared/perl/regex          region m?(?<!/)(?<!qr)/[^/\n]+(?=/)  /\w?   fill meta
 add-highlighter shared/perl/sregex         region s/[^/\n]+/[^/\n]+(?=/) /\w?          fill meta
+add-highlighter shared/perl/bregex         region s\{[^\}\n]+\}\{[^\}\n]+(?=\}) /\w?   fill meta
 
 add-highlighter shared/perl/quote_brace    region -recurse \{ \bq[qrwx]?\{ \}          fill string
 add-highlighter shared/perl/quote_paren    region -recurse \( \bq[qrwx]?\( \)          fill string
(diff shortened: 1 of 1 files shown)


rc/filetype/perl.kak
```diff
@@ -44,6 +44,7 @@ add-highlighter shared/perl/comment        region (?<!\$)(?<!\\)#   $
 
 add-highlighter shared/perl/regex          region m?(?<!/)(?<!qr)/[^/\n]+(?=/)  /\w?   fill meta
 add-highlighter shared/perl/sregex         region s/[^/\n]+/[^/\n]+(?=/) /\w?          fill meta
+add-highlighter shared/perl/bregex         region s\{[^\}\n]+\}\{[^\}\n]+(?=\}) /\w?   fill meta
```
krobelus, line 47: The second argument to a `region` highlighter is the closing pattern, so I think the `/\w?` should be `\}`.

(It would be simpler if we could use a `regex` highlighter instead of the `region` highlighter, but that's only possible for things that are in the default region. We can't do that for regex, since they can contain `"` like your example, so they need to be in a separate region from quoted strings..)

Carsenal: Thanks for the feedback. Its my inaugural kakoune commit, so I apologize as I'm still figuring out some bits of the sytnax. I've just added a commit to fix the issue.

krobelus: It's great to have people contribute - Perl highlighting is a monumental task. We need pick our battles, it seems impossible to support every detail.

`/}` is not quite right though, the closing should be `\}` (an escaped brace), otherwise the region would go until the literl sequence `/}`. I usually test `rc/` changes by running `src/kak` on a test file. Kakoune detects that it's running directly from the repository and uses the `rc` files in the repository instead of the system ones in `/usr/share`.

An alternative would be to include the closing brace in the region start, and as region end use a zero-width regex that always matches, like `region s\{[^\}\n]+\}\{[^\}\n]+\} (?=)`. That feels better for these regions that want to match a lot in the opening regex already, why not just match the full thing.

We can include `\w?` to also highlight regex modifiers like in line 46, or even `\w*` to support multiple modifiers.

One more thing I forgot, maybe it's better to use `*` instead of `+`, to highlight something like `s{delete-this}{}`.
I think `+` is used for the `/` variants to avoid some ambiguity, but I don't think we have that problem with matching braces.

Also before this gets merged you should squash the commits that merely fixup previous commits, to keep the history clean.
https://git-rebase.io/ has a nice tutorial on how to do that.

PR #3797, opened 2020-10-11
Support highlighting of pandoc's markdown raw_attribute extension

SeerLite:
https://pandoc.org/MANUAL.html#extension-raw_attribute
I only enabled highlighting for code blocks, not inline.

Let me know if there's a better way to regex this or if I missed anything!

### DIFF
--- a/rc/filetype/markdown.kak
+++ b/rc/filetype/markdown.kak
@@ -47,7 +47,7 @@ evaluate-commands %sh{
     ruby rust sass scala scss sh swift toml tupfile typescript yaml sql
   "
   for lang in ${languages}; do
-    printf 'add-highlighter shared/markdown/%s region -match-capture ^(\h*)```\h*%s\\b   ^(\h*)``` regions\n' "${lang}" "${lang}"
+    printf 'add-highlighter shared/markdown/%s region -match-capture ^(\h*)```\h*(%s|\\{=%s\\}))\\b   ^(\h*)``` regions\n' "${lang}" "${lang}" "${lang}"
     printf 'add-highlighter shared/markdown/%s/ default-region fill meta\n' "${lang}"
     [ "${lang}" = kak ] && ref=kakrc || ref="${lang}"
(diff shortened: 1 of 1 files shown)


SeerLite:
TIL FreeBSD's `printf` removes unescaped `\`s.

Tests should be fixed now!

alexherbo2:
Can you add a comment with the link in the code?

SeerLite:
You mean something like this?
```kak
# Also supports https://pandoc.org/MANUAL.html#extension-raw_attribute
```

Sure! Not all markdown flavours support this (I think pandoc may even be the only one?) so I think clarifying may be a good idea.
But first I'd just like to know, @lenormf why do you think otherwise. (?

alexherbo2:
Yep, it is nice to know what a regex refers to.  I’d add a list of the supported syntax (a space after the 3 quotes, before the language, etc.), so it is clear by looking at the comments of what is intended in the regex.  A few times I wonder “Is this character really needed in the regex?” or “What the hell does it does?”.

lenormf:
>But first I'd just like to know, @lenormf why do you think otherwise. (?

I'd rather have the information about why a change was necessary be stored into the commit message. A comment referencing a handful of characters in a file littered with regex seems out of place, plus that's what we do with C++ code already.

mawww:
I'd be ok for a comment like `pandoc raw_attribute extension` at the end of the line but I'd rather avoid pointing to urls for reference for a single line. Its also fine without comment, a git blame would point to this commit.

SeerLite:
Alright, I think I'll be leaving it without the comment.

> I'd rather have the information about why a change was necessary be stored into the commit message. A comment referencing a handful of characters in a file littered with regex seems out of place, plus that's what we do with C++ code already.

That's what I was guessing. Thanks for clarifying, I'm gonna keep this in mind for future contributions!

I'm fine with using `git blame` for this. I have even done it before, also for understanding regex.

PR #3811, opened 2020-10-16
Fix `wcwidth()` compile error with MSYS2

6112:
We try to detect when compiling under Cygwin, so we can set the
_XOPEN_SOURCE define which lets us use wcwidth(). We look for the
string "CYGWIN" in the uname, which looks like:

    CYGWIN-NT_10.0

MSYS2 is also based on Cygwin, so all of the above should still work.
However, we detect Cygwin by looking for the string "CYGWIN" in the
uname. In MSYS2, the uname looks like:

    MSYS_NT-10.0-18362

This patch looks for the string "_NT" instead of the string "CYGWIN"
in the uname, since it's common to both environments. This fixes a
compilation error on MSYS2.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -77,7 +77,8 @@ else ifeq ($(os),OpenBSD)
     CPPFLAGS += -D'KAK_BIN_PATH="$(bindir)/kak"' -I/usr/local/include
     LDFLAGS += -L/usr/local/lib
     mandir := $(DESTDIR)$(PREFIX)/man/man1
-else ifneq (,$(findstring CYGWIN,$(os)))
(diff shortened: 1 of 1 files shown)


caksoylar:
> CYGWIN-NT_10.0

Is this a typo? Above doesn't seem to contain `_NT`.

6112:
> > CYGWIN-NT_10.0
> 
> Is this a typo? Above doesn't seem to contain `_NT`.

Yes, it was a typo. Thanks for catching it. The real string was `CYGWIN_NT-10.0`

mawww:
Sorry for the delay, thanks for that fix.

PR #3812, opened 2020-10-16
gopls: Fix format/imports error reporting

voroskoi:
Remove unnecessary single quotes and whitespaces in %file{}
Do not delete error file before sending to debug buffer
Fix gopls definition error handling and empty sting check
Silence shellcheck warnings

### DIFF
--- a/rc/tools/go/gopls.kak
+++ b/rc/tools/go/gopls.kak
@@ -55,22 +55,22 @@ define-command -hidden -params 1 gopls-cmd %{
     gopls-prepare
     evaluate-commands %sh{
         dir=${kak_opt_gopls_tmp_dir}
-        gopls $1 -w ${kak_buffile} 2> "${dir}/stderr"
+        gopls "$1" -w "${kak_buffile}" 2> "${dir}/stderr"
         if [ $? -ne 0 ]; then
             # show error messages in *debug* buffer
-            printf %s\\n "echo -debug %file{ '${dir}/stderr' }"
+            printf %s\\n "echo -debug %file{${dir}/stderr}"
         fi
-        rm -r "${dir}"
     }
     edit!
+    evaluate-commands %sh{ rm -r "${kak_opt_gopls_tmp_dir}" }
 }
 
 # gopls definition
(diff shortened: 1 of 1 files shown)


rc/tools/go/gopls.kak
```diff
@@ -55,22 +55,22 @@ define-command -hidden -params 1 gopls-cmd %{
...
+        gopls "$1" -w "${kak_buffile}" 2> "${dir}/stderr"
         if [ $? -ne 0 ]; then
             # show error messages in *debug* buffer
-            printf %s\\n "echo -debug %file{ '${dir}/stderr' }"
+            printf %s\\n "echo -debug %file{${dir}/stderr}"
         fi
-        rm -r "${dir}"
     }
     edit!
+    evaluate-commands %sh{ rm -r "${kak_opt_gopls_tmp_dir}" }
```
lenormf, line 65: If we don't care about the commands printed by the shell we can use the `nop` command (`-f` added for completeness):

```suggestion
    nop %sh{ rm -rf "${kak_opt_gopls_tmp_dir}" }
```

krobelus: I would not use `-f` because we expect the directory to exist, so printing an error if it doesn't seems right. It's not very important though.

lenormf: I use it because it's only a temporary directory, so its being there doesn't impact the functionality of the tool, and we can afford to ignore unexpected/unlikely errors.

pickfire:
Last time I tried formatting a go file with 15k+ lines, it only returns 1k lines. Heh

voroskoi:
> Last time I tried formatting a go file with 15k+ lines, it only returns 1k lines. Heh

That is really surprising, as it just calls gopls. Could you try gopls from the command line to see if it is a gopls or a wrapper problem?
`gopls format -w file.go` should work.

pickfire:
Or maybe it is related to kak-lsp because I use that for gopls. I probably don't want to go back and check this. `gofmt` have no issue but not sure if it's gopls issue.

Issue #3567, opened 2020-06-26
Factorise `spell.kak` to accommodate Hunspell support


lenormf:
Hi,

A user on IRC asked about Hunspell support, because [their distribution didn't want to create an Aspell package](https://dev.getsol.us/T5602#173616).

I ended up writing the following patch, to get things to work:

```diff
diff --git a/rc/tools/spell.kak b/rc/tools/spell.kak
index ca231ee5..e007528d 100644
--- a/rc/tools/spell.kak
+++ b/rc/tools/spell.kak
@@ -25,7 +25,7 @@ define-command -params ..1 -docstring %{
                 rm -rf "$(dirname "$kak_opt_spell_tmp_file")"
                 exit 1
             else
-                options="-l '$1'"
+                options="-d '$1'"
                 printf 'set-option buffer spell_last_lang %s\n' "$1"
             fi
         }
@@ -37,8 +37,14 @@ define-command -params ..1 -docstring %{
         fi
 
         {
-            sed 's/^/^/' "$kak_opt_spell_tmp_file" | eval "aspell --byte-offsets -a $options" 2>&1 | awk '
+            eval "hunspell -a $options '${kak_opt_spell_tmp_file}'" | awk '
+                function escape_quote(s) {
+                    gsub(Q, Q "\"" Q "\"" Q, s)
+                    return s
+                }
+
                 BEGIN {
+                    Q = "\047"
                     line_num = 1
                     regions = ENVIRON["kak_timestamp"]
                     server_command = sprintf("kak -p \"%s\"", ENVIRON["kak_session"])
@@ -58,8 +64,11 @@ define-command -params ..1 -docstring %{
                     }
 
                     else if (/^[#&]/) {
-                        word_len = length($2)
+                        word_len_cmd = "printf %s " Q escape_quote($2) Q " | wc -c"
+                        word_len_cmd | getline word_len
+                        close(word_len_cmd)
                         word_pos = substr($0, 1, 1) == "&" ? substr($4, 1, length($4) - 1) : $3;
+                        word_pos++;
                         regions = regions " " line_num "." word_pos "+" word_len "|Error"
                     }
 
@@ -144,9 +153,9 @@ define-command \
     -docstring "Suggest replacement words for the current selection, against the last language used by the spell-check command" \
     spell-replace %{ evaluate-commands %sh{
     if [ -n "$kak_opt_spell_last_lang" ]; then
-        options="-l '$kak_opt_spell_last_lang'"
[...]

PR #3813, opened 2020-10-17
spell.kak: fix multibyte character handling

voroskoi:
Hi,

I have no idea if it have some unwanted consequences or not. I have tested it with `en_US` and `hu_HU`, works for me.

The root and compound handling part should not be a problem, but the `awk -b` thing could be. I have made some research and I do *not* find this option in BSD awk, mawk, busybox awk.

Maybe this should be handled with a spell_awk option default to awk and write some nice docs for the users. What do You think?

### DIFF
--- a/rc/tools/spell.kak
+++ b/rc/tools/spell.kak
@@ -37,7 +37,7 @@ define-command -params ..1 -docstring %{
         fi
 
         {
-            sed 's/^/^/' "$kak_opt_spell_tmp_file" | eval "aspell --byte-offsets -a $options" 2>&1 | awk '
+            sed 's/^/^/' "$kak_opt_spell_tmp_file" | eval "aspell --byte-offsets -a $options" 2>&1 | awk -b '
                 BEGIN {
                     line_num = 1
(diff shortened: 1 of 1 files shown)


pickfire:
Is `hu_HU` multibyte?

voroskoi:
> Is `hu_HU` multibyte?

We have accent chars in Hungarian: `íéáűúőóüö` for example, their byte size is 2 in UTF-8.

Screwtapello:
I was thinking about writing a patch to add a `spellcmd` option that would default to aspell, but which you could set to hunspell or enchant if aspell doesn't do a good job for your language. Unfortunately, aspell is the only spell-checker that supports the `--byte-offsets` flag, so non-English-speakers would have to choose between poor-quality suggestions or poor-quality offsets.

Given how unportable byte-offset options are, I wonder if it would be easier all around to teach `:update-option` a new `-chars-to-bytes` flag that converts units at the same time as it updates timestamps.

lenormf:
#3567

Issue #1993, opened 2018-04-06
Keep the entire history of 'previous buffer'


occivink:
Currently only the last buffer is recorded (accessible via `ga` or by deleting the current buffer) so if you delete two buffers in succession, the client will jump to the first buffer in the list (because there is no "last buffer" at this point). 

It would be nicer if the whole history was kept, so that deleting buffers in succession would be more predictable.

Delapouite:
I think this feature could be implemented if the jump-list was exposed somehow. (so it's a bit like the issue on exposing the undo history tree nodes)

PR #3814, opened 2020-10-17
Add last modified buffer (gm)

### LINKED ISSUE #3607: Last modified buffer
### Feature 

`^` in vim. To jump to the previous modified buffer. When opening like 10+ buffer, I find myself frequently typing a few times (~5x) to `:bp` or `:bn` and I get lost easily when switching to previous buffer.

### Usecase 

It would be good to quickly switch to last modified buffer easily. It should be similar to `:bp` and `:bn` except the ordering is based on last modified rather than buffer creation time.

pickfire:
Behaves similar to vim ^ such that it jumps to buffer that was modified.
Internally, it keeps two last modified buffer, it behaves like a stack
and push the buffer when there is a new modified buffer and it also
prevents both last modified buffers pointing to the same buffer.

This broke my workflow a lot so I came in and write some C++, hopefully
I got it correct.

Fix #3607


pickfire:
> why two pointers?

If there is only one pointers, if we switched to another buffer without modifying the current one, it we always jump to the original one. If say the second buffer is the previous modified file, it should jump to the same file? So we need two buffers to jump between alternative modified buffers as we cannot guarantee that the buffers are always modified when we change buffers.

> why use a sentinel (m_modified_since_accessed)?

Err, I tried using history but I didn't get that method working, we need a sentinel. Say if we changed many files, you happened to jump from one of the files that was previously modified but not one of the last two pointers, if would jump to that file even though that was modified after many modified buffers jumps.

> is #1993 impacted by this patch?

No, having the sentinel does not impact that, if we chose two change the history like vim default `nohidden`, we could not keep the history of the previous buffers which I think is bad.

pickfire:
Should we add changelog for this?

pickfire:
@mawww Ping

PR #3817, opened 2020-10-19
rc *sh.kak: include "-" as word character

krobelus:
This seems better given that kakrc.kak does the same, however the change to
sh.kak might be contentious.


krobelus:
probably not worth it because this is not expected by many users

PR #3821, opened 2020-10-19
rc perl.kak: fix regex highlighting of s{foo}{bar}

krobelus:
- rc perl.kak: highlight multiple regex modifiers
- rc perl.kak: highlight substitution with empty replacement, like `s{foo}{}`

### DIFF
--- a/rc/filetype/perl.kak
+++ b/rc/filetype/perl.kak
@@ -42,9 +42,9 @@ add-highlighter shared/perl/double_string  region (?<!\$)(?<!\\)"   (?<!\\)(\\\\
 add-highlighter shared/perl/single_string  region (?<!\$)(?<!\\\\)' (?<!\\)(\\\\)*' fill string
 add-highlighter shared/perl/comment        region (?<!\$)(?<!\\)#   $               fill comment
 
-add-highlighter shared/perl/regex          region m?(?<!/)(?<!qr)/[^/\n]+(?=/)  /\w?   fill meta
-add-highlighter shared/perl/sregex         region s/[^/\n]+/[^/\n]+(?=/) /\w?          fill meta
-add-highlighter shared/perl/bregex         region s\{[^\}\n]+\}\{[^\}\n]+(?=\}) \}     fill meta
+add-highlighter shared/perl/regex          region m?(?<!/)(?<!qr)/[^/\n]+(?=/)  /\w*   fill meta
+add-highlighter shared/perl/sregex         region s/[^/\n]+/[^/\n]+(?=/)        /\w*   fill meta
+add-highlighter shared/perl/bregex         region s\{[^\}\n]+\}\{[^\}\n]+(?=\}) \}\w*  fill meta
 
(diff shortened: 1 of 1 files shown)


rc/filetype/perl.kak
```diff
@@ -42,9 +42,9 @@ add-highlighter shared/perl/double_string  region (?<!\$)(?<!\\)"   (?<!\\)(\\\\
 add-highlighter shared/perl/single_string  region (?<!\$)(?<!\\\\)' (?<!\\)(\\\\)*' fill string
 add-highlighter shared/perl/comment        region (?<!\$)(?<!\\)#   $               fill comment
 
-add-highlighter shared/perl/regex          region m?(?<!/)(?<!qr)/[^/\n]+(?=/)  /\w?   fill meta
-add-highlighter shared/perl/sregex         region s/[^/\n]+/[^/\n]+(?=/) /\w?          fill meta
-add-highlighter shared/perl/bregex         region s\{[^\}\n]+\}\{[^\}\n]+(?=\}) /}     fill meta
+add-highlighter shared/perl/regex          region m?(?<!/)(?<!qr)/[^/\n]+/\w*  (?=)    fill meta
+add-highlighter shared/perl/sregex         region s/[^/\n]+/[^/\n]*/\w*        (?=)    fill meta
+add-highlighter shared/perl/bregex         region s\{[^\}\n]+\}\{[^\}\n]*\}\w* (?=)    fill meta
```
mawww, line 47: `(?=)` is going to match at every single position in the buffer, and for performance reason region highlighters cache those results, so I would expect this to introduce a big memory overhead on perl highlighting.

A quick test on Markdown.pl makes the `Regions` memory domain go from 2MB to 5MB with this change, I'd perfer another closing regex that would not trigger this.

krobelus: Makes sense, I dropped that change.

I assumed that a "closing" match is only computed on lines where the opening regex matches.
The loop in `src/highlighters.cc 2126` seems to do that - `find_matching_end` is only called after `find_next_begin` suceeded. So I don't know yet where that memory usage is coming from. I'll try heaptrack.

krobelus: Heaptrack doesn't show me any significant difference on [Markdown.pl](https://fastapi.metacpan.org/source/BOBTFISH/Text-Markdown-1.000031/script/Markdown.pl), the debug build of kak with an empty config peaks at around 1.9MB either way.

Is there an easy way to show stats for the memory domains?

mawww: If you look at the implementation of `find_matching_end` you'll see it relies on an already cached `RegexMatchList`, all matches are cached and only modified lines get their matches updated.

Not familiar with Heaptrack, but I use `:debug memory` to see memory usage as tracked by Kakoune and could see the `Regions` memory domain raise from 2MB to 5MB between the two versions.

krobelus: Thanks, I was able to reproduce a difference of the same scale now!

PR #3829, opened 2020-10-22
Revive customizable title line 

### LINKED ISSUE #2217: Separate ncurses title from mode line
Currently the window title is the mode line with ' - Kakoune' appended to it. This is not very nice. Since mode line generally has much more real estate than the window title.

A window title is more likely to be used for just identifying the window, while the mode line is expected to contain as much useful information as possible.

Given that, it would be nice if we can set the title and mode line separately.

yshui:
When ncurses_set_title is set to a string other than "yes" and
"true", it will be used to set the title line. This give the user
the ability to configure the title line.

Co-Authored-By: yshui <yshuiv7@gmail.com>

Was #2249, Fixes #2217

### DIFF
--- a/doc/pages/options.asciidoc
+++ b/doc/pages/options.asciidoc
@@ -327,8 +327,10 @@ are exclusively available to built-in options.
     interface implementation. The NCurses UI support the following options:
 
         *ncurses_set_title*:::
-            if *yes* or *true*, the terminal emulator title will
-            be changed
+            if not set to *no* or *false*, the terminal emulator
+            title will be changed. if set to *yes* or *true*, the
+            default title line will be used, otherwise the value
+            of this option will be used as the title line
 
         *ncurses_status_on_top*:::
             if *yes*, or *true* the status line will be placed
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -512,19 +512,34 @@ void NCursesUI::draw_status(const DisplayLine& status_line,
 
     if (m_set_title)
     {
-        constexpr char suffix[] = " - Kakoune\007";
         char buf[4 + 511 + 2] = "\033]2;";
-        // Fill title escape sequence buffer, removing non ascii characters
-        auto buf_it = &buf[4], buf_end = &buf[4 + 511 - (sizeof(suffix) - 2)];
-        for (auto& atom : mode_line)
+        auto buf_it = &buf[4];
+        if (not m_title_line.empty())
         {
-            const auto str = atom.content();
-            for (auto it = str.begin(), end = str.end();
-                 it != end and buf_it != buf_end; utf8::to_next(it, end))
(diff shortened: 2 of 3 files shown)


Screwtapello:
I'm a little confused by this PR. There's only one `m_title_line` field in the `NCursesUI` class; does that mean every Kakoune window would share the exact same title? Would `%val{}` expansion in the option value be expanded dynamically? Is the intention that the option be set at window-scope so every window can have a different title?

If this PR means every window gets the same title, I'm not sure this is a very useful addition.

If this PR means every window can get its own title, either by embedding expansions in the value, or by setting the option at different scopes, that should be mentioned in the documentation.

danr:
I'm using this PR like so:
```
set global ui_options ncurses_assistant=none ncurses_set_title=false
hook global WinCreate .* %{
    set -add window ui_options "ncurses_set_title=%val{bufname} - kakoune"
}
```
Each window gets its own title and when I change between (kakoune) windows the title updates. And each X window gets the correct kakoune window title.

alexherbo2:
It looks the same approach than how `dynregex` can be replaced with hooks.

https://github.com/mawww/config/commit/92669dbe31a1ca29ee951769dac40270e83b843e

doc/pages/options.asciidoc
```diff
@@ -327,8 +327,10 @@ are exclusively available to built-in options.
     interface implementation. The NCurses UI support the following options:
 
         *ncurses_set_title*:::
-            if *yes* or *true*, the terminal emulator title will
-            be changed
+            if not set to *no* or *false*, the terminal emulator
+            title will be changed. if set to *yes* or *true*, the
+            default title line will be used, otherwise the value
+            of this option will be used as the title line
```
Screwtapello, line 333: When I read this, it wasn't obvious to me how to make different windows have different titles (which seemed an important use-case). How about:

```suggestion
            If set to *no* or *false*, the terminal emulator
            title will not be changed.
            
            If set to *yes* or *true*, each window's title will
            mirror the contents of the window's modeline
            (as set by the `modelinefmt` option).
            
            Otherwise, each window will use the value of this
            option at window scope as the title text, without
            further expansion.
```

danr: That's clearer!

caksoylar:
Would it make sense to make this dynamic, like `dynregex` highlighter or `modeline` option itself? Then we wouldn't need to hook into window events to modify the value either.

danr:
That would indeed be more convenient!

Gskartwii:
This PR probably conflicts with builtin-terminal-ui pretty severely. Are there plans to resolve the conflicts? I might be interested in taking over the PR in the future, if it's not maintained right now.

yshui:
@Gskartwii please go ahead if you want to take over.

yshui:
Closing in favor of #4265

PR #3831, opened 2020-10-22
Disabled backtick escaping

ManDay:
Fixes issue #2377 by removing the ability to escape backticks in backticks in compliance with asciidoc's behaviour.

### DIFF
--- a/doc/pages/command-parsing.asciidoc
+++ b/doc/pages/command-parsing.asciidoc
@@ -69,12 +69,12 @@ No other escaping takes place in balanced strings.
 Other words are non-quoted. Non-quoted words end either on a whitespaces
 or a `;`.
 
-If they start with `\\` followed by `%`, `'` or `"`, then that leading
-`\\` is discarded.
+If they start with `\` followed by `%`, `'` or `"`, then that leading
+`\` is discarded.
 
-If a whitespace or `;` is preceded by `\\`, then the `\\` is discarded
-and the whitespace or `;` becomes part of the word. Any other `\\`
-is treated as a literal `\\`.
+If a whitespace or `;` is preceded by `\`, then the `\` is discarded
+and the whitespace or `;` becomes part of the word. Any other `\`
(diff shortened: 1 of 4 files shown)


mawww:
Happy to merge that, however did you check that this backtick escaping was not used anywhere in the doc pages ?

ManDay:
Using `grep -Rho '<bt>.*<bt>' . | grep '^\(<bt>[^<bt>]*\)\{2\}<bt>$'` where <bt> is a backtick, yes. But as mentioned (and apparent right here), it would be a mistake in the first place (and thus a separate bug), because asciidoc does not properly support it, to all I have learned.

krobelus:
```git grep '\\`' doc``` shows that there is another place that needs updating to render correctly. This is the one mentioned in #2377

```diff
diff --git a/doc/pages/command-parsing.asciidoc b/doc/pages/command-parsing.asciidoc
index dcf3724f..e4b62123 100644
--- a/doc/pages/command-parsing.asciidoc
+++ b/doc/pages/command-parsing.asciidoc
@@ -69,12 +69,12 @@ No other escaping takes place in balanced strings.
 Other words are non-quoted. Non-quoted words end either on a whitespaces
 or a `;`.
 
-If they start with `\\` followed by `%`, `'` or `"`, then that leading
-`\\` is discarded.
+If they start with `\` followed by `%`, `'` or `"`, then that leading
+`\` is discarded.
 
-If a whitespace or `;` is preceded by `\\`, then the `\\` is discarded
-and the whitespace or `;` becomes part of the word. Any other `\\`
-is treated as a literal `\\`.
+If a whitespace or `;` is preceded by `\`, then the `\` is discarded
+and the whitespace or `;` becomes part of the word. Any other `\`
+is treated as a literal `\`.
 
 == Typed Expansions
 
```

mawww:
Looks like I dont have a copyright waiver commit from you, could you add one ?

PR #3835, opened 2020-10-23
rc markdown: Fix bullet highlighting

### LINKED ISSUE #3582: Markdown highlighter doesn't work with nested code block in list item
### Steps 
1. Edit a simple .md file in kakoune with content:
~~~markdown
* 1st item
* ```
  funcall()
  ```
* 3rd item

## new header
~~~

### Outcome 
The highlighter doesn't recognize the start of a code block after `* `, but take the ending "```" as the start of a code block, thus messed up all the highlighting. Like this:

Screenshot 2020-07-06 at 5 03 09 PM

### Expected 
Expect the highlighter to recognize a code blocker nested in a list item.

lenormf:
This commit prevents the lines following the one that holds the bullet
from being highlighted with the `bullet` face when they're indented:

- The bullet is highlighted properly, so is this sentence
  but this line the ones that would follow are not

Fixes #3582

### DIFF
--- a/rc/filetype/markdown.kak
+++ b/rc/filetype/markdown.kak
@@ -59,9 +59,9 @@ add-highlighter shared/markdown/codeblock region -match-capture \
     ^(\h*)```\h*$ \
     fill meta
 
-add-highlighter shared/markdown/listblock region ^\h*[-*]\s ^\h*((?=[-*])|$) regions
-add-highlighter shared/markdown/listblock/marker region \A [-*]\s fill bullet
-add-highlighter shared/markdown/listblock/content default-region ref markdown/inline
+add-highlighter shared/markdown/listblock region ^\h*[-*]\s ^\h*((?=[-*])|$) group
+add-highlighter shared/markdown/listblock/ ref markdown/inline
+add-highlighter shared/markdown/listblock/marker regex ^\h*([-*])\s 1:bullet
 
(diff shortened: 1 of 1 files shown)


rc/filetype/markdown.kak
```diff
@@ -59,9 +59,10 @@ add-highlighter shared/markdown/codeblock region -match-capture \
     ^(\h*)```\h*$ \
     fill meta
 
-add-highlighter shared/markdown/listblock region ^\h*[-*]\s ^\h*((?=[-*])|$) regions
-add-highlighter shared/markdown/listblock/marker region \A [-*]\s fill bullet
-add-highlighter shared/markdown/listblock/content default-region ref markdown/inline
+add-highlighter shared/markdown/listblock region ^\h*[-*]\s ^\h*((?=[-*])|$) group
+add-highlighter shared/markdown/listblock/regions regions
+add-highlighter shared/markdown/listblock/regions/ default-region ref markdown/inline
```
mawww, line 64: Why does this introduce a region with only a default-region ? adding a `ref markdown/inline` to the group would have the same effect would'nt it ?

lenormf: The recent tendency to create regions for everything has triggered by habit of grouping all regions into a single group. I've made the change!

PR #3836, opened 2020-10-23
rc markdown: Highlight inline code blocks properly

lenormf:
This commit allows code blocks to be prefixed with tabulation
characters to be picked up and highlighted by the editor.

Indenting caused by the inclusion of an inline code block into a
list item is also taken into account. However, that might cause false
positives, for example with a hard wrapped list item indented with
an amount of spaces congruent to 4.

### DIFF
--- a/rc/filetype/markdown.kak
+++ b/rc/filetype/markdown.kak
@@ -82,7 +82,7 @@ add-highlighter shared/markdown/inline/text/ regex ^\h*(>\h*)+ 0:comment
 add-highlighter shared/markdown/inline/text/ regex "\H( {2,})$" 1:+r@meta
 
 # Inline code
-add-highlighter shared/markdown/inline/text/ regex "^    [^\n]*" 0:meta
+add-highlighter shared/markdown/inline/text/ regex "^( {4}|\t)+([^\n]+)" 2:meta
 
 # Commands
 # ‾‾‾‾‾‾‾‾


rc/filetype/markdown.kak
```diff
@@ -82,7 +82,7 @@ add-highlighter shared/markdown/inline/text/ regex ^\h*(>\h*)+ 0:comment
 add-highlighter shared/markdown/inline/text/ regex "\H( {2,})$" 1:+r@meta
 
 # Inline code
-add-highlighter shared/markdown/inline/text/ regex "^    [^\n]*" 0:meta
+add-highlighter shared/markdown/inline/text/ regex "^( {4}|\t)+([^\n]+)" 2:meta
```
krobelus, line 85: I believe we can avoid false positives on lists by requiring an empty line before an inline code block.
According to the [CommonMark Spec](https://spec.commonmark.org/0.29/#indented-code-blocks) it's not necessary after headlines, but I don't think I've ever seen an indented code block without a preceding empty line.

So I'd suggest:

    add-highlighter shared/markdown/inline/text/ regex "^\n(?:(?:\t| {4})[^\n]*\n)+" 0:meta

lenormf: An empty line is not required to precede a code block, according to the references I could find online. It's also not clear that we should assume that code blocks all follow that principle, after all the predicate is restrictive enough (multiples of 4 spaces indentation or plain tabulation characters) that false positives shouldn't happen often.

I also don't want to apply the `meta` face to all the whitespace preceding the code.

As for the rest of the regex, it seems to be a more complicated rewrite of the one already in the commit? Non-capturing parenthesis, `[^\n]*\n instead of `[^\n]+`.

Screwtapello: Looking through the spec krobelus linked, a four-space indent makes that line part of what went before, if that makes sense:

- 4-space indent after a heading: code block
- 4-space indent after a blank line: code block
- 4-space indent after a line of a paragraph: more paragraph
- 4-space indent after a list item: more list item
- 8-space indent after a list item: code block inside a list item

lenormf: From the document:

>Is a blank line needed before an indented code block? (Markdown.pl requires it, but this is not mentioned in the documentation, and some implementations do not require it.)

I realise some false positives will arise in lists, but the current behaviour is consistent across most implementations (although not necessarily across the most popular ones, whichever those are) albeit impractical.

lenormf: Although on second thought, it might be better to assume an empty line is needed, and let people indent blocks with tabulation characters to avoid a false negative.

krobelus: Either version is fine by me, I don't mind having (rare) false positives for now.

Treating tabs the same as 4 spaces is definitely an improvement, it's
specififed in https://spec.commonmark.org/0.29/#tabs.

> 4-space indent after a list item: more list item
8-space indent after a list item: code block inside a list item

Ugh yeah that's hard to implement properly. We currently have a `listblock` region but it just highlights what's inside as normal markdown. I wonder if it's possible to support nested list.

> I also don't want to apply the meta face to all the whitespace preceding the code.

Right, I didn't take care of that, applying only to the code itself seems better although I don't know how it makes a difference.

Issue #3011, opened 2019-07-07
x11-repl does not work with termite


JJK96:
### Steps 

set buffer termcmd 'termite -e'
x11-repl

### Outcome 

termite is started but immediately closes

### Expected 

termite is started and stays open

### Probable cause

The `-e` flag to termite does not support passing multiple commands using `&&` as is done in `repl/x11.kak`

lenormf:
`termite -e` executes the command passed as argument as-is, it doesn't pass it to a shell - you had the right diagnosis. And setting `termcmd` to `termite -e sh -c` won't work either because Termite expects the command to be passed as a single CLI argument, i.e. the resulting command has to be `termite -e 'sh -c "$CMD"'`.

It was bound to happen one day or the other, the way `termcmd` is being used by scripts might have to be modified to fix this bug.

lenormf:
It looks like Tilix also suffers from the same issue. [Doesn't look like the commands are forwarded to a shell](https://github.com/gnunn1/tilix/blob/master/source/app.d#L152).

Issue #1795, opened 2018-01-07
X11 repl only works for some termcmds


atp30:
Hi,

Firstly - apologies if I have got the wrong end of the stick - I have only recently discovered Kakoune and am still getting my head around the approach to configuration.

I am currently experiminting considering using kakoune as my main editor. However, I have not had any success in launching repl windows from x11-repl.kak (calling x11-repl from an unconfigured install of kakoune from nixpkgs-unstable launches correctly with urxvt and bash, but  the window title is not set, and x11-send-text fails silently).

As far is I can tell, the current version assumes it is able able to set the window title by appending '-t kak_repl_window' to the termcmd. This works correctly with termite, where this option disables dynamic title updates, but not with urxvt or xterm, which use '-T' instead, and permit the title to be ovrewritten by the repl process. (I have not tested other terminals).

I have found an approach that seems to work for [my own configuration](https://gist.github.com/atp30/bb71c4981ecaa063b8fa5db023dc3e29): I have modified `x11-repl.kak` to store the processID of the launched terminal in an option, using this to search for the correct window. This has the advantage of permitting per-buffer or per-window repls, but I haven't been able to configure it use setsid - although the terminals do nonetheless seem to survive the termination of all kakoune processes regardless.

mawww:
Hello,

Using the title to find the repl terminal is quite a big limitation of x11-repl, unfortunately using the pid is not going to be much more robust, as some emulators, such as rxvt-unicode, have a daemon mode where all windows are managed by a server process, and launching a new terminal just tells that server process to create a new window (so the launched process is not tied with the actual window).

An alternative approach would be to change the actual command we launch in that terminal so that it gathers the WINDOWID env var and communicate it back so that the repl window is correctly identified.

PR #3841, opened 2020-10-27
rc x11-repl: identify the REPL window by window id instead of title

listentolist:
Now it uses the window id to identify the REPL window. It is stored in the option `x11_repl_id`. That way it is possible to have different REPLs for different buffers or windows. And AFAICT there are less issues with different termcmds and shells.
#1795 #3011 #2973

### DIFF
--- a/rc/windowing/repl/x11.kak
+++ b/rc/windowing/repl/x11.kak
@@ -13,8 +13,9 @@ define-command -docstring %{
     -params .. \
     -shell-completion \
     x11-repl %{ x11-terminal sh -c %{
+        winid="${WINDOWID:-$(xdotool search --pid ${PPID} | tail -1)}"
         printf "evaluate-commands -try-client $1 \
-            'set-option current x11_repl_id ${WINDOWID}'" | kak -p "$2"
+            'set-option current x11_repl_id ${winid}'" | kak -p "$2"
         shift 2;
         [ "$1" ] && "$@" || "$SHELL"
     } -- %val{client} %val{session} %arg{@}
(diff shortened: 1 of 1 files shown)


Screwtapello:
I think this will not be reliable - when Kakoune launches a new terminal for the REPL, that terminal will not necessarily be made the active window; lots of window managers let the user configure whether new windows should automatically be made active. Even if it is made active, it might not be made active immediately, so there's a race condition between the window-manager switching the active window, and `xdotool getactivewindow` running.

That said, there are other approaches. Many X11 terminal emulators set `$WINDOWID`, which (if present) is probably more reliable than messing with xdotool. If that's not set, a script could use the existing escape sequence to set the title to a string including the shell's pid (`$$`) then use `xdotool searchbyname` to look for it — which still depends on `searchbyname` but at least it keeps working if the window title changes later (for example, shells that put $PWD in the title).

listentolist:
Yes, most terminal emulators set `$WINDOWID`. Then we can just change `$(xdotool getactivewindow)` to $WINDOWID`, although it doesn't work for all termcmds (e.g. gnome-terminal does not set this variable). I just changed it.

alexherbo2:
@listentolist How to have multiple REPLs?  If you use `repl-new` two times, the second call erases the option storing the window ID of the first REPL.

How about using a `int-list` option storing the window ID of the REPLs and adding a `repl-connect` command completing the IDs to change the connection?

listentolist:
@alexherbo2 `repl-new` opens a REPL for the current scope. So to get new REPL for the current buffer only you have to make sure that `x11_repl_id` is set in the buffer scope:
```
set-option buffer x11_repl_id ''; repl-new
```

PR #3846, opened 2020-10-27
Add a shebang filetype detection hook

Eluminae:
This work with filetypes as

* #!/bin/sh
* #!/bin/sh -x
* #!/bin/env sh
* #!/bin/env python
* #!/bin/env python3

This will strip last digit to transfrom python3 to python

This will cause trouble with shebang as /bin/bash as bash is not
the correct filetype which is sh.

You still can use this to map this kind of filetype

hook global WinSetOption filetype=bash %{
	set buffer filetype sh
}

edit: a mapping has been added to handle python3 and bash execs.

### DIFF
--- a/rc/detection/shebang.kak
+++ b/rc/detection/shebang.kak
@@ -1,0 +1,20 @@
+declare-option -docstring "mapping between shebang exec and actual filetype" \
+    str shebang_filetypes 'bash=sh zsh=sh dash=sh python3=python'
+
+# Add the following function to a hook on BufWritePost to automatically parse shebangs
+define-command -hidden shebang-parse %{
+    evaluate-commands %sh{
+        if [ -z "${kak_opt_filetype}" ] || [ "plain" = "${kak_opt_filetype}" ]; then
+            first_line="$(head -n+1 "${kak_buffile}")"
+
+            if [ 0 -eq "$(expr "$first_line" : "#!")" ]; then
+                return
+            fi
+
+            filetype="$(basename "$(echo "$first_line" | awk -F '(/| )' '{print $NF}')")"
+            filetype="$(echo "$kak_opt_shebang_filetypes" | sed -E "s/.*${filetype}=([a-z0-9^\s]*).*/\1/")"
+
+            printf "set-option buffer filetype '%s'\n" "${filetype}"
+        fi
+    }
+}


alexherbo2:
Can we add the check on save, when the filetype is not set?  The use case is when creating new scripts.

SeerLite:
What about using a command instead of a hook? Something like `shebang-parse`, to make it consistent with `modeline-parse`.
That way you can set up a `BufWrite` hook to set it up like @alexherbo2. I also have exactly that set up for `modeline-parse`.

SeerLite:
I came up with something to deal with python3 and bash/shells:
Declaring a new `str-to-str-map` that acts as a shebang/binary -> filetype map.

This can then be parsed:

```sh
# kak_opt_shebang_filetypes='bash=sh zsh=sh dash=sh python3=python'
# filetype=python3
echo "$kak_opt_shebang_filetypes" | sed -E "s/.*${filetype}=([a-z0-9^\s]*).*/\1/"

# with set-option included (too unreadable?)
echo "$kak_opt_shebang_filetypes" | sed -E "s/.*${filetype}=([a-z0-9^\s]*).*/set-option buffer filetype '\1'/"
```

What do you all think?

---

I also noticed the shebang/filetype detection could be simplified:
```
echo "$first_line" | awk -F '(/| )' '{print $NF}'
```
(Uses both `/` and space as field separator and prints last field.)

alexherbo2:
I would prefer to extract the [existing detection][`file.kak`] in a command and use a [`magic`] file.

[`file.kak`]: https://github.com/mawww/kakoune/blob/master/rc/detection/file.kak
[`magic`]: https://github.com/alexherbo2/configuration/blob/master/home/magic

krobelus:
For which cases does the existing detection not work?

lenormf:
Related #711

SeerLite:
As for me I didn't even know about `file.kak` (I never really bothered investigating shebang detections and just set `kak:filetype=` where I needed it).
So today I saw this and thought it was the first time it was being looked into. I'm gonna be checking out that magic file stuff though.

I wonder then, @alexherbo2 what's the usecase you see for this one (seeing as you reacted/commented)? Or does this one do something different that I missed?

Eluminae:
I pushed a v2 thanks to all of you

* kak_opt_shebang_filetypes to map bash to sh, python3 to python
* simplified the filetype detection using a better awk command
* used a command that we can manually hook to BufWritePost as modeline-parse

lenormf:
The problem here is that this code is redundant with the functionality provided by the `file` UNIX utility.

As mentioned by @alexherbo2 here and by myself in #711, the idiomatic way to have shebangs parsed is to leverage `file.kak`, and the magic file it reads to customise its heuristics. The [`file` implementation ArchLinux uses](https://www.darwinsys.com/file/) seems to be properly detecting the shebangs mentioned here, you should give it a try if it doesn't work on your distro.

A nice contribution touching on contents detection could be to move the code in `file.kak` into a separate command, which could be used to implement [@alexherbo2 suggestion](https://github.com/mawww/kakoune/pull/3846#issuecomment-717628677).

Eluminae:
@lenormf I'm okay with that and it seems a reasonable decision. I pushed the better version of this code anyway :)

Anyone could update the file to make it hookable ?

alexherbo2:
@Eluminae In the file detection support, you can extract the code in the hook into a named command, on opening try the command to set the filetype, if not set, add a hook once to try an additional check on save.

The ideal scenario would be to use the file detection when the buffer has been linked to a file, and not just on opening a file.

mawww:
I think relying on `file.kak` and the file command is the preferred approach here, it has served us well for a long time and extending the magic file will be useful to many other use cases outside of Kakoune, so we should encourage improvements to be done upstream.

PR #3854, opened 2020-10-29
Use the right tabstop when converting to spaces

### LINKED ISSUE #3853: Weird behavior when `@`ing multiple tabs with count arg and tabstop is less 4
### Steps 
1. Set the `tabstop` to 2.
2. Fill a buffer with 2 tabs.
3. Select both tabs in normal mode.
4. Press `4@`.

### Outcome 
The 2 tabs are converted to 6 spaces.

### Expected 
The tabs are converted to 8 spaces, because of the `4` passed in to `@`.

I could find similar behavior with `tabstop` set to 1 and 3 (probably had to add more tabs to reproduce though).

SeerLite:
Fixes #3853.
Big disclaimer: I don't know C++ (yet), but fixing this seemed pretty simple. I don't know why `opt_tabstop` was used instead of `tabstop` in the first place though, so maybe there's something deeper that I missed.


Screwtapello:
It's not clear to me what this is doing, other than removing the `opt_tabstop` intermediate variable.

lenormf:
Counts seem to be ignored in some cases because `opt_tabstop` is used instead of the `tabstop` variable that does take counts into account.

Screwtapello:
Ah, right, I missed that `tabstop` was being calculated, but not always used. Thanks.

mawww:
I believe the reason the existing code uses opt_tabstop is because it first computes the actual display position of the spaces (so, opt_tabstop is the correct value there, because this is what gets used to expand tabs for display), then replaces those spaces with tab character using the explicitely given tabstop value.

The proposed change says that even the initial column computation should work as-if tabstop was the explicit value for the whole line, not just for the spaces we convert, so tabs appearing earlier in the line will impact the computed column position of the selected spaces to something that does not match the display column (but would match it *if* opt_tabstop had been this value).

I tend to think the new behaviour makes a bit more sense but not by much, any opinions on that ?

PR #3858, opened 2020-10-30
src/ncurses_ui.cc: Teach Kakoune about ctrl-symbol keys.

### LINKED ISSUE #2553: Cannot map <c-space>
### Steps 

1. `kak -n`
1. `:map global normal <c-space> ": echo ctrl space<ret>"`
1. Press `<c-space>`

### Outcome 

Nothing happens.

### Expected 

"ctrl space" should appear in the bottom-left of the terminal.

### Notes

`<c-space>` is not actually a character that appears in an ASCII table, but by the convention that the Control key means "clear the top three bits of the character", `<c-space>` produces ASCII NUL, and in fact most terminals will send NUL when the user presses `<c-space>`. Run `cat` and try it yourself.

In slightly more conventional notation, NUL could also be written as `<c-@>`, but that form doesn't work either (and is much, much more awkward to type). At least with `gnome-terminal`, I can also send NUL with `<c-2>` and ctrl-backtick which I can't type in markdown.

I also can't make `<c-space>` do anything if I prefix it with `<c-v>`, or type it in insert mode.

Other non-alphabetic control characters like `<c-^>` (aka `<c-6>` and `<c-~>`) and `<c-_>` (aka `<c-7>`) also can't be mapped, but I can type them on the command-line with `<c-v>` and I can type them in insert mode.

Screwtapello:
Previously, Kakoune only handled ctrl-codes less than 27, representing them as lower-case ASCII codes. For regular keys like `<c-a>`, that worked fine. However, NUL became the unorthodox ``<c-`>`` and other ctrl-symbols (`<c-\>`, `<c-]>`, `<c-_>`) weren't supported at all.

Now NUL is rendered as the more comfortable `<c-space>`, and the other ctrl-symbol codes are properly decoded.

Fixes #2553.

### DIFF
--- a/src/keys.cc
+++ b/src/keys.cc
@@ -229,6 +229,10 @@ UnitTest test_keys{[]()
          alt('j'),
          ctrl('r'),
          shift(Key::Up),
+         ctrl('['),
+         ctrl('\\'),
+         ctrl(']'),
+         ctrl('_'),
     };
     String keys_as_str;
     for (auto& key : keys)
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -623,8 +623,19 @@ Optional<Key> NCursesUI::get_next_key()
             kill(0, SIGTSTP); // We suspend at this line
             return {};
         }
+        // Special case: you can type NUL with Ctrl-2 or Ctrl-Shift-2 or
+        // Ctrl-Backtick, but the most straightforward way is Ctrl-Space.
+        if (c == 0)
+            return ctrl(' ');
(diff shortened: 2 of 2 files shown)


src/keys.cc
```diff
@@ -229,6 +229,10 @@ UnitTest test_keys{[]()
          alt('j'),
          ctrl('r'),
          shift(Key::Up),
+         ctrl('['),
+         ctrl('\\'),
```
Screwtapello, line 233: This is here just to double-check that no accidental escaping shenanigans would break people's ability to map `<c-\>`.

PR #3859, opened 2020-10-30
Add very basic svelte support

JJK96:
Basically just a wrapper for the html support, but this can be extended later.

### DIFF
--- a/rc/filetype/svelte.kak
+++ b/rc/filetype/svelte.kak
@@ -5,18 +5,18 @@ hook global WinCreate .*\.svelte %[
 hook global WinSetOption filetype=(svelte) %{
     require-module html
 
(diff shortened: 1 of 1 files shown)


rc/filetype/svelte.kak
```diff
@@ -0,0 +1,22 @@
+hook global WinCreate .*\.svelte %[
+    set-option window filetype svelte
+]
+
+hook global WinSetOption filetype=(svelte) %{
+    require-module html
+
+    hook window ModeChange pop:insert:.* -group "%val{hook_param_capture_1}-trim-indent"  html-trim-indent
```
mawww, line 8: Any strong reason to use hook capture and string expansions here instead of just hard-coding "svelt" for those hook groups ?

JJK96: Not really, I kept it in so it could support different filetypes in the future, but I don't think there are any other filetypes for svelte. 
So I'll hardcode it.

PR #3862, opened 2020-11-02
src: Support strikethrough faces

eraserhd:
ncurses does not support strikethrough (intentionally, because they ran out of bits?) but Kitty does (at least), and the code reserved for it is quite old, so it should be safe.

[See here](https://stackoverflow.com/questions/8357203/is-it-possible-to-display-text-in-a-console-with-a-strike-through-effect)

We could support the face but do nothing with it on `master` if you like.

### DIFF
--- a/src/face.hh
+++ b/src/face.hh
@@ -9,17 +9,18 @@ namespace Kakoune
 
 enum class Attribute : int
 {
-    Normal    = 0,
-    Underline = 1 << 1,
-    Reverse   = 1 << 2,
-    Blink     = 1 << 3,
-    Bold      = 1 << 4,
-    Dim       = 1 << 5,
-    Italic    = 1 << 6,
-    FinalFg   = 1 << 7,
-    FinalBg   = 1 << 8,
-    FinalAttr = 1 << 9,
-    Final     = FinalFg | FinalBg | FinalAttr
+    Normal        = 0,
+    Underline     = 1 << 1,
+    Reverse       = 1 << 2,
+    Blink         = 1 << 3,
+    Bold          = 1 << 4,
+    Dim           = 1 << 5,
+    Italic        = 1 << 6,
+    Strikethrough = 1 << 7,
+    FinalFg       = 1 << 8,
+    FinalBg       = 1 << 9,
+    FinalAttr     = 1 << 10,
+    Final         = FinalFg | FinalBg | FinalAttr
 };
 
 constexpr bool with_bit_ops(Meta::Type<Attribute>) { return true; }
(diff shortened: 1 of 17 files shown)


lenormf:
Do you have a usecase for this? Markup?

eraserhd:
This came up working on kak-jira.  JIRA has markup for added and removed text.  These are rendered on the web as underline and strikethrough.

That being said, IMHO text rendering in Kakoune doesn’t need to match anything else — this seemed like it could be generally useful.

src/face.hh
```diff
@@ -9,17 +9,18 @@ namespace Kakoune
...
+    Underline     = 1 << 1,
+    Reverse       = 1 << 2,
+    Blink         = 1 << 3,
+    Bold          = 1 << 4,
+    Dim           = 1 << 5,
+    Italic        = 1 << 6,
+    FinalFg       = 1 << 7,
+    FinalBg       = 1 << 8,
+    FinalAttr     = 1 << 9,
+    Strikethrough = 1 << 10,
```
mawww, line 22: I'd rather have all non-final attributes together, so could you move it next to italic, as `1 << 7`

src/terminal_ui.cc
```diff
@@ -181,7 +181,22 @@ void TerminalUI::Screen::set_face(const Face& face)
 {
     static constexpr int fg_table[]{ 39, 30, 31, 32, 33, 34, 35, 36, 37, 90, 91, 92, 93, 94, 95, 96, 97 };
     static constexpr int bg_table[]{ 49, 40, 41, 42, 43, 44, 45, 46, 47, 100, 101, 102, 103, 104, 105, 106, 107 };
-    static constexpr int attr_table[]{ 0, 4, 7, 5, 1, 2, 3 };
```
mawww: while the new code is a bit more explicit, I kind of like the simplicity of having a onliner table, which should be possible again if you change Strikethrough as requested above.

src/terminal_ui.cc
```diff
@@ -199,10 +214,10 @@ void TerminalUI::Screen::set_face(const Face& face)
     bool join = false;
     if (face.attributes != m_active_face.attributes)
     {
-        for (int i = 0; i < sizeof(attr_table) / sizeof(int); ++i)
+        for (int i = 0; i < sizeof(attr_table) / sizeof(attr_table[0]); ++i)
```
mawww, line 217: we could use `std::size(attr_table)` here to avoid having to do the division dance.

mawww (changes requested):

eraserhd:
@mawww All requested changes made.

(I also tried to make the attribute bit numbers equal to the terminal codes, which required skipping a few, but it ended up not being an improvement.)

mawww:
cherry-picked

PR #3868, opened 2020-11-04
Check for division before we check for regexes

### LINKED ISSUE #3856: HTML/js highlighter wrong on `/`
Filetype `html`

File contents:
```
<script>
    let a = 1/2
    let b = 3
</script>
```
Screen Capture_select-area_20201029184831

Carsenal:
Closes #3856 
This seems a little hacky, and if it's deemed ***too*** hacky, I can rewrite it.


mawww:
Priority for regions that start at the same place are supposed to be handled through ordering, so this is the expected approach.

PR #3871, opened 2020-11-06
Aspell undocumented buggy-behavior fixed.

npatsakula:
Description of behavior:
https://lists.gnu.org/archive/html/aspell-user/2009-11/msg00003.html

tldr:
> Note that this result begins with a plus sign + instead of & * or #. I believe this means that "rsA'l" is the root word of "brsAilnA" (since b+ and +nA are affixes), but this does not seem to be part of the usual aspell/ispell pipe api.

### DIFF
--- a/rc/tools/spell.kak
+++ b/rc/tools/spell.kak
@@ -53,6 +53,10 @@ define-command -params ..1 -docstring %{
                         # nothing
                     }
 
+                    else if (/^\+/) {
+                        # required to ignore undocumented aspell functionality
+                    }
+
                     else if (/^$/) {
                         line_num++
                     }


lenormf:
Note that this character is not produced by the latest stable release (`v0.60.8`), and that that version that allegedly does (`v0.61`, although I haven't checked) is labelled as a “Historical Development Version”.

I looked at the code and found that `v0.60.8` can also produce a `?` leading control character (`aspell-0.60.8/prog/aspell.cpp:930`, which apparently isn't expected either by the person who wrote the mailing list post.

npatsakula:
> is not produced by the latest stable release

This is not true. Minimal reproducible example: `echo "разделе Подзапросы" | aspell -l ru -d ru -a`.

aspell version: 0.60.8
aspell-ru version: 0.99f7-8

lenormf:
Weird, I couldn't find the code that prints that sign in the official source. Not important.

PR #3882, opened 2020-11-12
Improve svelte highlighter

JJK96:
Uses javascript highlighter for code blocks in the html part.


JJK96:
See before and after:

https://imgur.com/a/DSDDbZK

PR #3885, opened 2020-11-16
Ignore indent on line completion

### LINKED ISSUE #3071: Insert mode: <c-x> l/L completion candidates should ignore indentation
Complete line in insert mode will not list matching candidates because it treats same lines with different indentation levels to be different.
```
// I have this line with indentation
    PRINTLN()

// next I type PRI without any indentation followed by <c-x>l
// sadly but there's no complete candidate, unless I begin the line with same indentation
PRI
```
I reckon lines should be trimmed in order to populate matching candidates, otherwise it's not so useful.

AndrewVos:
- Ignore indent when completing lines
- Andrew Vos Copyright Waiver

This is my first pull request to kakoune, so I'm hoping the test I added is in the right place!

Closes #3071


mawww:
Hum, admitting we do want to ignore indent on line completion, would'nt we need also to insert the completion respecting the existing indentation ?

AndrewVos:
@mawww yes, that's a great idea, in my opinion. Should I add that to this pull request?

mawww:
Yes please.

AndrewVos:
Actually, I wasn't thinking things through last night.

This pull request actually exhibits the behaviour you were talking about.
There's no need to add back in the indent as far as I understand. Is it
confusing maybe because the test input is on zero indent? If the end of the
file had some indent then the completed word would be at that indent.

AndrewVos:
Actually, ignore me. I just checked and it seems you're right. I'll fix that, thanks for catching it!

AndrewVos:
@mawww Ok, this seems to be working now. I think I could clean up the code a bit with regard to variable names and such, but I would love a review on the `steal_indent` method. I'm not sure I'm even doing that right at all.

mawww (changes requested):
I think we can go with a different approach:

`InsertCompletion` contains a begin and end coordinate, which is the text in the buffer to replace, I think this should be the text from one-past-indent till the cursor position, and the candidate text should be the full line without indent.

With that approach there is no need for `steal_indent` and I think it maps the mental model of what an indent-agnostic line completion should do.

does that make sense ?

AndrewVos:
That makes a lot more sense than what I've done, thanks.

AndrewVos:
@mawww I believe this is ready to go now.

Thanks for helping me through this.

By the way, I updated the functional test because it wasn't actually using line completion before. I believe it was using word completion to pass the test.

PR #3888, opened 2020-11-17
doc commands: Align mis-indented command flags

lenormf:


lenormf:
Found #3889 in the process of doing this.

PR #3893, opened 2020-11-17
Detect filetype on buffer save too

Eluminae:
This allow kaking a new file, add a shebang in the first line, and get
the right filetype detected automatically on save.

### DIFF
--- a/rc/detection/file.kak
+++ b/rc/detection/file.kak
@@ -1,4 +1,4 @@
-hook global BufOpenFile .* %{ evaluate-commands %sh{
+define-command -hidden file-detection %{ evaluate-commands %sh{
     if [ -z "${kak_opt_filetype}" ]; then
         mime=$(file -b --mime-type -L "${kak_buffile}")
         mime=${mime%;*}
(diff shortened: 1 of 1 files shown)


alexherbo2:
@Eluminae How about exposing and documenting the `file-detection` command?

Eluminae:
I used to. But in fact, I don't think we need more than those two hooks. So I'm now not sure this is necessary.

krobelus:
I'd love to know why Vim chooses not to do this.

Eluminae:
Vim allow to `:e!` which will reload the file and trigger filetype detection.

Screwtapello (approved):

mawww:
This is a nice change but I would like one tweak: only change the filetype of the buffer if its currently unset.

I think it would get quite frustrating if file misdetects the filetype and overwrite the user specified value at each write.

krobelus (approved):
> only change the filetype of the buffer if its currently unset.

I think that's already the case via `if [ -z "${kak_opt_filetype}" ]`.

(It would be nice to also get rid of the fork if the filetype is already set, but that's not a blocker)

krobelus:
Maybe the hooks should have groups, so they can be disabled dynamically?
Something like `file-open` and `file-write`.
Some of my plain text files (where `""` is the correct filetype) are detected as ruby for some reason..

PR #3900, opened 2020-11-20
spell.kak: Add arbitrary backend support

voroskoi:
Add `spell_cmd` option to run various (ispell compatible) backends
Add `spell_awk_options`, so one can use `awk -b` for multibyte handling

Ignore lines starting with -, * or +: this will fix backends with
affiliations support

Add spell.asciidoc to document spell.kak

This PR is an upsteam compatible version of #3813

I have tried this with aspell, hunspell and nuspell with en_US and hu_HU:
 - aspell works fine, and it is the fastest by far
 - hunspell does not work with hu_HU as there is no --byte-offsets mode in hunspell.
 - nuspell works fine beside the spell-add command which is not implemented in nuspell yet.


rc/tools/spell.kak
```diff
@@ -37,7 +39,7 @@ define-command -params ..1 -docstring %{
         fi
 
         {
-            sed 's/^/^/' "$kak_opt_spell_tmp_file" | eval "aspell --byte-offsets -a $options" 2>&1 | awk '
+            sed 's/^/^/' "$kak_opt_spell_tmp_file" | eval "$kak_opt_spell_cmd $options" 2> /dev/null | awk "$kak_opt_spell_awk_options" '
```
krobelus, line 42: I wonder if we should simply ask awk if -b is supported. Then we can probably get rid of the configuration variable.

```sh
awk_options=
awk -b "" >/dev/null 2>&1 && awk_options=-b
```

Only tested with gawk but I'm pretty sure it will not add the -b for BSD awk.

rc/tools/spell.kak
```diff
@@ -3,6 +3,8 @@ declare-option -hidden str spell_last_lang
 declare-option -hidden str spell_tmp_file
 
 declare-option -docstring "default language to use when none is passed to the spell-check command" str spell_lang
+declare-option -docstring "spell checking engine to use" str spell_cmd 'aspell --byte-offsets -a -l'
```
Screwtapello, line 6: `makecmd` and `grepcmd` and `lintcmd` and `formatcmd` and `termcmd` all lack an underscore, so `spell_cmd` probably should too.

rc/tools/spell.kak
lenormf, line 42: We want errors to be reported, not ignored, plus that change is made in an unrelated hunk.

rc/tools/spell.kak
```diff
@@ -49,14 +51,10 @@ define-command -params ..1 -docstring %{
                         # drop the identification message
                     }
 
-                    else if (/^\*/) {
+                    else if (/^[-*+]/) {
```
lenormf, line 54: Is `-` a sign produced by Awk that wasn't supported? Why factorise the following scope into this one? Why is this hunk in the commit that only introduces options?

rc/tools/spell.kak
```diff
@@ -49,14 +51,10 @@ define-command -params ..1 -docstring %{
...
                         # nothing
                     }
 
-                    else if (/^\+/) {
-                        # required to ignore undocumented aspell functionality
```
lenormf, line 57: We lost the comment here, which documents that `+` is an undocumented control character.

rc/tools/spell.kak
lenormf, line 6: I think this default command would force users to pass a language to `:spell`, because it is trailed with `-l`.

lenormf (commented):
I don't think that's the play.

Having a `spellcmd` option to allow specifying an implementation of Aspell is not good enough because the flag to pick a dictionary is different (`-l` on Aspell, `-d` on Hunspell), hardcoded in the plugin's implementation, and added on demand.

Having a visible option (`spell_awk_options`) to control an implementation detail that is not portable also isn't appealing.

The problem here is that Awk with multibyte handling is not portable. Maybe we should find a tool that converts its input, or change it entirely (to use… Perl? I don't know).

rc/tools/spell.kak
voroskoi: Much better, thanks.

rc/tools/spell.kak
voroskoi: Sure.

rc/tools/spell.kak
voroskoi: That was accidental, I do not want that either.

rc/tools/spell.kak
voroskoi: Actually as the commit message says it is necessary for languages with affiliation, hunspell use `-` for compounds and `+` for root.

rc/tools/spell.kak
voroskoi: I just thought it does not matter if we ignore these because of aspell bug or because of hunspell. But You are right, we should not drop the comment. I think I will add the hunspell info too as that is also related info.

rc/tools/spell.kak
voroskoi: I will make a v2 of this patch with a couple of if statements for setting up the backend. Maybe we could try the backends and use the first one found, something similar we do with windowing.

Thanks for the awesome feedback!

voroskoi:
> Having a `spellcmd` option to allow specifying an implementation of Aspell is not good enough because the flag to pick a dictionary is different (`-l` on Aspell, `-d` on Hunspell), hardcoded in the plugin's implementation, and added on demand.

Let me create v2, that should be much better.

> Having a visible option (`spell_awk_options`) to control an implementation detail that is not portable also isn't appealing.

Yeah, that is lame. But I like the idea @krobelus suggested.

> The problem here is that Awk with multibyte handling is not portable. Maybe we should find a tool that converts its input, or change it entirely (to use… Perl? I don't know).

I like the fact that spell.kak is written in awk as most of us have some kind of awk installed. If you go for perl/python/lua whatever someone will hate it. Lua is the cheapest one, but I am sure there are more people out here with awk, than lua.

voroskoi:
Second version pushed. Let me know what You think about it.

I will update the doc, once the code is in final form.

voroskoi:
I think this is the final version. I have used it quite a lot today, tested all the backends. Please review!

voroskoi:
I am using since then and have not run into any issue. I really think this should be merged, as I can not see any regression or whatever, but it makes life much easier.

PR #3903, opened 2020-11-21
Coalesce User Mode Mappings

### LINKED ISSUE #3902: Multiple Key Options for a User Mode Selection
Some user-mode-like popups such as the one in when doing `<a-i>` offer multiple options for the same command. For example, the `parenthesis block` after doing `<a-i>` can be triggered with `b,(,)`. **How can I get a user mode to also show multiple key options for the same command?** I tried
```
map global mymode r,[,] ": my-command" -docstring "doc"
```
with no luck. Is there a way to do this currently?

enricozb:
This patch resolves #3902. Some user-mode-like autoinfos have multiple keys bound to the same action, such as `<a-i>`, which has `b,(,)` mapping to an action. There is currently no way to combine multiple mappings in a user mode.

This PR combines (non `built_in`) keymaps in an autoinfo that have the **exact same command and docstring**. For example,
```
map global user p ": echo test" -docstring "some command"
map global user n ": echo test" -docstring "some command"
```
Will now show up as:
image

This change also applies to custom user modes.

## Disclaimer
This is my first time attempting to contribute to Kakoune's codebase, so I'm not super familiar with the containers and custom string types. I'm also not that experienced in C++, so I'm sure I'm performing some unnecessary copies somewhere in here. And lastly of course please reject this if this change isn't desired.

### DIFF
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -195,13 +195,11 @@ String build_autoinfo_for_mapping(const Context& context, KeymapMode mode,
     for (auto& key : keymaps.get_mapped_keys(mode))
     {
         KeymapManager::KeymapInfo key_info = keymaps.get_mapping(key, mode);
-        std::pair<String, String> cmd_and_docstr = {join(key_info.keys | transform(key_to_str)),
+        std::pair<String, String> cmd_and_docstr = {accumulate(key_info.keys | transform(key_to_str),
+                                                               String{},
+                                                               std::plus{}),
                                                     key_info.docstring};
-        if (synonyms.contains(cmd_and_docstr)) {
-          synonyms[cmd_and_docstr].push_back(key_to_str(key));
-        } else {
-          synonyms.insert({std::move(cmd_and_docstr), {key_to_str(key)}});
-        }
+        synonyms[cmd_and_docstr].push_back(key_to_str(key));
     }
 
     for (auto& [cmd_and_docstr, keys] : synonyms) {
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -189,9 +189,24 @@ String build_autoinfo_for_mapping(const Context& context, KeymapMode mode,
             descs.emplace_back(std::move(keys), built_in.docstring);
     }
 
+    // To combine mapped keys with the same command and docstrings (synomyms),
+    // we need a map (command, docstring) -> mapped keys
+    HashMap<std::pair<String, String>, Vector<String>> synonyms;
     for (auto& key : keymaps.get_mapped_keys(mode))
-        descs.emplace_back(key_to_str(key),
-                           keymaps.get_mapping(key, mode).docstring);
+    {
+        KeymapManager::KeymapInfo key_info = keymaps.get_mapping(key, mode);
+        std::pair<String, String> cmd_and_docstr = {join(key_info.keys | transform(key_to_str)),
+                                                    key_info.docstring};
+        if (synonyms.contains(cmd_and_docstr)) {
+          synonyms[cmd_and_docstr].push_back(key_to_str(key));
+        } else {
+          synonyms.insert({std::move(cmd_and_docstr), {key_to_str(key)}});
+        }
+    }
+
+    for (auto& [cmd_and_docstr, keys] : synonyms) {
+        descs.emplace_back(join(keys, ',', false), cmd_and_docstr.second);
+    }
 
     auto max_len = 0_col;
     for (auto& desc : descs)
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -202,9 +202,8 @@ String build_autoinfo_for_mapping(const Context& context, KeymapMode mode,
         synonyms[cmd_and_docstr].push_back(key_to_str(key));
     }
 
-    for (auto& [cmd_and_docstr, keys] : synonyms) {
+    for (auto& [cmd_and_docstr, keys] : synonyms)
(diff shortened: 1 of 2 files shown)


src/normal.cc
```diff
@@ -189,9 +189,24 @@ String build_autoinfo_for_mapping(const Context& context, KeymapMode mode,
...
+    HashMap<std::pair<String, String>, Vector<String>> synonyms;
     for (auto& key : keymaps.get_mapped_keys(mode))
-        descs.emplace_back(key_to_str(key),
-                           keymaps.get_mapping(key, mode).docstring);
+    {
+        KeymapManager::KeymapInfo key_info = keymaps.get_mapping(key, mode);
+        std::pair<String, String> cmd_and_docstr = {join(key_info.keys | transform(key_to_str)),
+                                                    key_info.docstring};
+        if (synonyms.contains(cmd_and_docstr)) {
+          synonyms[cmd_and_docstr].push_back(key_to_str(key));
```
CharlesGueunet, line 201: is there a reason you did not use `emplace_back` here ?

enricozb: I thought `emplace_back` was useful for forwarding params to a constructor, why should I prefer it in this case?

mawww: Both are fine by me, I would tend to use push_back in the case where emplace_back would call the copy constructor anyway, such as here.

src/normal.cc
```diff
@@ -189,9 +189,24 @@ String build_autoinfo_for_mapping(const Context& context, KeymapMode mode,
...
+    // we need a map (command, docstring) -> mapped keys
+    HashMap<std::pair<String, String>, Vector<String>> synonyms;
     for (auto& key : keymaps.get_mapped_keys(mode))
-        descs.emplace_back(key_to_str(key),
-                           keymaps.get_mapping(key, mode).docstring);
+    {
+        KeymapManager::KeymapInfo key_info = keymaps.get_mapping(key, mode);
+        std::pair<String, String> cmd_and_docstr = {join(key_info.keys | transform(key_to_str)),
+                                                    key_info.docstring};
+        if (synonyms.contains(cmd_and_docstr)) {
```
mawww, line 200: I dont think we need that `if`, both cases should work fine with `synonyms[std::move(cmd_and_docstr)].push_back(key_to_str(key));`

src/normal.cc
```diff
@@ -189,9 +189,24 @@ String build_autoinfo_for_mapping(const Context& context, KeymapMode mode,
...
 
+    // To combine mapped keys with the same command and docstrings (synomyms),
+    // we need a map (command, docstring) -> mapped keys
+    HashMap<std::pair<String, String>, Vector<String>> synonyms;
     for (auto& key : keymaps.get_mapped_keys(mode))
-        descs.emplace_back(key_to_str(key),
-                           keymaps.get_mapping(key, mode).docstring);
+    {
+        KeymapManager::KeymapInfo key_info = keymaps.get_mapping(key, mode);
+        std::pair<String, String> cmd_and_docstr = {join(key_info.keys | transform(key_to_str)),
```
mawww, line 198: I think you can do `accumulate(..., String{}, std::plus{})` in place of join here, avoiding the need to define a new one in string_utils.

mawww (changes requested):

src/normal.cc
```diff
@@ -189,9 +189,22 @@ String build_autoinfo_for_mapping(const Context& context, KeymapMode mode,
...
+    {
+        KeymapManager::KeymapInfo key_info = keymaps.get_mapping(key, mode);
+        std::pair<String, String> cmd_and_docstr = {accumulate(key_info.keys | transform(key_to_str),
+                                                               String{},
+                                                               std::plus{}),
+                                                    key_info.docstring};
+        synonyms[cmd_and_docstr].push_back(key_to_str(key));
+    }
+
+    for (auto& [cmd_and_docstr, keys] : synonyms) {
```
mawww, line 205: No curly braces around single statement bodies please.

mawww (commented):
Looks good, could you add a copyright waiver commit as well (see CONTRIBUTING) I dont think I have one from you.

enricozb:
Done. Do I only need this commit once?

vbauerster:
imho it's better to introduce a new switch like `-alias` for this purpose.
So instead of many lines
```
map global user p ": echo test" -docstring "some command"
map global user n ": echo test" -docstring "some command"
```
one line is suffice
```
map global user p ": echo test" -alias n -alias x -docstring "some command"
```

enricozb:
@vbauerster 
I think your idea is a better design. One issue though is that with this switch you can't add aliases to existing mappings, only to new ones.

vbauerster:
By `existing mappings` do you mean user defined ones? If so just adding the switch to existing mapping will do the trick.
Built-in mappings (modes) like `goto` and `view` for example are user editable, so I believe it should be possible to redefine any of them. 

One issue though is that it's possible to define one to many alias like so, which I think should be an error or the last one wins:

```
map global user p ": echo test_a" -alias a
map global user b ": echo test_b" -alias a
```

enricozb:
Please disregard my previous comment.

With respect to multiple mappings with conflicting aliases, I'm not sure whether an error or "last one wins" makes more sense. I would probably prefer "last one wins". And I think that would work well with different scopes too:
```
map global user p ": echo test_a" -alias a
map buffer user b ": echo test_b" -alias a
```
@mawww any thoughts on this design?

PR #3912, opened 2020-11-24
Distinguish <c-h> and Backspace on terminals where that is possible.

### LINKED ISSUE #3863: Distinguish <c-h> and Backspace on terminals where that is possible.
## Feature 

Some (many?) terminals in common use today send different codes for Ctrl-H versus Backspace. Kakoune should support that distinction where it's available.

### Usecase 

See #1030, #1270, #3250 

## Notes

The original VT100 terminal had a key above the Enter key labelled "Backspace" that sent 0x08, and another nearby key labelled "Delete" that sent 0x7F.

The VT220 terminal had a key above the Enter key labelled ⌫ that sent 0x7F, and another nearby key labelled "Delete" that sent 0x1B 0x5B 0x33 0x7E, although it could be configured to use the VT100 conventions for compatibility.

Since different terminals followed different conventions, the Unix kernel added support for tracking "what does backspace send" so the kernel's built-in line-editing would work everywhere. Since this is one of the terminal settings from the Before Time (predating termcap/terminfo/curses, I believe) and since getting it wrong is near-instantly visible (backspace doesn't work in the shell!), it's extremely reliable.

(if you'd like to see what *your* terminal is configured to send for backspace, try running `stty -a | grep erase`)

In fact, Kakoune already reads this setting at startup, it just doesn't do anything with it. Inside `NCursesUI::get_next_key()`, instead of detecting Backspace by comparing the incoming character code to `ctrl('h')` or `127`, we could compare it to `m_original_termios.c_cc[V_ERASE]`, and then whichever keycode was *not* backspace could be decoded normally (0x08 as `<c-h>`, 0x7F as `<delete>`).

Unfortunately, Kakoune's current input-parsing code is built on a bunch of inline lambda functions, which are very picky about which values from outer contexts they can access, and apparently "member variables of the enclosing class" are not on the list. I'd make a PR, but I lack the C++ skill to rearrange the code to fit.

Screwtapello:
Different terminals send different codes to indicate backspace, usually one of `\x08` or `\x7f`, so Kakoune blindly treated both as backspace. However, a given terminal is only likely to use one of those, and mnemonic control codes like `<c-h>` are a precious resource so we should endeavour to keep backspace and `<c-h>` separate when we can. Luckily, termios tells us what code our terminal is currently using, and Kakoune already reads the information at startup, so we can just use that information.

Thanks to @krobelus for figuring out the C++ syntax required.

Fixes #3863.

### DIFF
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -610,16 +610,19 @@ Optional<Key> NCursesUI::get_next_key()
     if (not c)
         return {};
 
-    static constexpr auto convert = [](Codepoint c) -> Codepoint {
+    const cc_t erase = m_original_termios.c_cc[VERASE];
+    auto convert = [erase](Codepoint c) -> Codepoint {
         if (c == control('m') or c == control('j'))
             return Key::Return;
         if (c == control('i'))
             return Key::Tab;
-        if (c == control('h') or c == 127)
+        if (c == erase)
             return Key::Backspace;
+        if (c == 127) // when it's not backspace
+            return Key::Delete;
         return c;
     };
-    auto parse_key = [](unsigned char c) -> Key {
+    auto parse_key = [&convert](unsigned char c) -> Key {
         if (Codepoint cp = convert(c); cp > 255)
             return Key{cp};
         if (c == control('z'))
@@ -652,7 +655,7 @@ Optional<Key> NCursesUI::get_next_key()
        return Key{utf8::codepoint(CharIterator{c}, Sentinel{})};
     };
 
-    auto parse_csi = [this]() -> Optional<Key> {
+    auto parse_csi = [this, &convert]() -> Optional<Key> {
         auto next_char = [] { return get_char().value_or((unsigned char)0xff); };
         int params[16] = {};
         auto c = next_char();


lenormf:
That's good, I have backspace mapped already, and wanted to map <kbd>c-h</kbd> to `:man<ret>`!

src/ncurses_ui.cc
```diff
@@ -610,13 +610,16 @@ Optional<Key> NCursesUI::get_next_key()
     if (not c)
         return {};
 
-    static constexpr auto convert = [](Codepoint c) -> Codepoint {
+    const cc_t erase = m_original_termios.c_cc[VERASE];
+    static auto convert = [erase](Codepoint c) -> Codepoint {
```
mawww, line 614: This was static constexpr so that `parse_key` and `parse_csi` did not need to capture it, but now making it just static is weird as it means it will use the value of m_original_termios at first call, which should not change in the current way Kakoune works, but would break if we moved one day to say have the server handle multiple ncurses clients directly instead of having multiple client processes.

So this should probably be non-static and I guess we could make `parse_key` and `parse_csi` capture it by reference for now, until I find an idea I like better and refactor that code.

mawww (changes requested):

lenormf:
Bump!

krobelus:
This should match @mawww's suggestion. Works just as well.

```diff
diff --git a/src/ncurses_ui.cc b/src/ncurses_ui.cc
index e52748c9..812e1612 100644
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -610,16 +610,19 @@ Optional<Key> NCursesUI::get_next_key()
     if (not c)
         return {};
 
-    static constexpr auto convert = [](Codepoint c) -> Codepoint {
+    const cc_t erase = m_original_termios.c_cc[VERASE];
+    auto convert = [erase](Codepoint c) -> Codepoint {
         if (c == control('m') or c == control('j'))
             return Key::Return;
         if (c == control('i'))
             return Key::Tab;
-        if (c == control('h') or c == 127)
+        if (c == erase)
             return Key::Backspace;
+        if (c == 127) // when it's not backspace
+            return Key::Delete;
         return c;
     };
-    auto parse_key = [](unsigned char c) -> Key {
+    auto parse_key = [&convert](unsigned char c) -> Key {
         if (Codepoint cp = convert(c); cp > 255)
             return Key{cp};
         if (c == control('z'))
@@ -652,7 +655,7 @@ Optional<Key> NCursesUI::get_next_key()
        return Key{utf8::codepoint(CharIterator{c}, Sentinel{})};
     };
 
-    auto parse_csi = [this]() -> Optional<Key> {
+    auto parse_csi = [this, &convert]() -> Optional<Key> {
         auto next_char = [] { return get_char().value_or((unsigned char)0xff); };
         int params[16] = {};
         auto c = next_char();
```

Screwtapello:
Once again, thank you @krobelus! I've included those changes and rebased my branch onto the latest master.

PR #3923, opened 2020-12-01
Match the end of heredoc with horizontal spacing 

stevengubler:
If using spaces for indentation, heredocs should still terminate on indented end-token, as is currently allowed for indentation with tabs.

### DIFF
--- a/rc/filetype/sh.kak
+++ b/rc/filetype/sh.kak
@@ -27,7 +27,7 @@ add-highlighter shared/sh/double_string region  %{(?<!\\)(?:\\\\)*\K"} %{(?<!\\)
 add-highlighter shared/sh/single_string region %{(?<!\\)(?:\\\\)*\K'} %{'} fill string
 add-highlighter shared/sh/expansion region -recurse (?<!\\)(?:\\\\)*\K\$\{ (?<!\\)(?:\\\\)*\K\$\{ \}|\n fill value
 add-highlighter shared/sh/comment region (?<!\\)(?:\\\\)*(?:^|\h)\K# '$' fill comment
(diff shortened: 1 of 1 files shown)


stevengubler:
Any ideas why this stalled on the MacOS tests? I didn't change anything that should effect the build...

rc/filetype/sh.kak
```diff
@@ -27,7 +27,7 @@ add-highlighter shared/sh/double_string region  %{(?<!\\)(?:\\\\)*\K"} %{(?<!\\)
 add-highlighter shared/sh/single_string region %{(?<!\\)(?:\\\\)*\K'} %{'} fill string
 add-highlighter shared/sh/expansion region -recurse (?<!\\)(?:\\\\)*\K\$\{ (?<!\\)(?:\\\\)*\K\$\{ \}|\n fill value
 add-highlighter shared/sh/comment region (?<!\\)(?:\\\\)*(?:^|\h)\K# '$' fill comment
-add-highlighter shared/sh/heredoc region -match-capture '<<-?\h*''?(\w+)''?' '^\t*(\w+)$' fill string
+add-highlighter shared/sh/heredoc region -match-capture '<<-?\h*''?(\w+)''?' '^\h*(\w+)$' fill string
```
krobelus, line 30: Thanks for the contribution; can you provide an example script where spaces precede the heredoc delimiter?

I didn't find one, and Bash's man page only mentions tabs:
> If  the  redirection operator is <<-, then all leading tab characters are stripped from input lines and the line containing delimiter.  This allows here-documents within shell scripts to be indented in a natural fashion.

stevengubler: Actually, I do apologize. It appears that bash does specifically check for tabs and not spaces. That's both terribly annoying, and a fact which actually renders this PR irrelevant. Thanks for the catch.

PR #3934, opened 2020-12-14
fix(doc): put <height> and <width> from window_range in the right order

tototest99:


occivink:
As far as I can tell, `coord_x` and `coord_y` are swapped as well. But instead of changing the doc, I think it would make more sense to change the variable, since this format seems unusual.

Screwtapello:
My only guess is that `window_range` uses "y x h w" order because [that's what ncurses uses](https://linux.die.net/man/3/mvaddch), presumably because [that's what terminals use](https://vt100.net/docs/vt102-ug/chapter5.html#S5.5.2.6) (search for "CUP").

Of course, that doesn't make it a *good* reason, and if we don't care about compatibility with existing scripts, we should make the implementation match the sensible-sounding docs.

tototest99:
I considered that `coord_x` was abstract and thus fitted the first coordinate (maybe it’s because I’m used to it for image manipulation), unlike `height` wich designate, well, a vertical thing, but I understand your points. I have no strong opinion.

Delapouite:
The order of `window_range` values has been fixed in another commit https://github.com/mawww/kakoune/commit/3b0f69a6d68d1a8a5d0abcf8a611f401cf0bb9af (2022)

This issue can be closed

Issue #3940, opened 2020-12-14
The documentation should be organised better


lenormf:
### Feature 

The information in the documentation should be organised better. There's also a lot of fragmentation between it and the wiki, it would be nice to bring them closer together.

HTH.

lenormf:
Related #3686

naringas:
also IMHO the documentation should be put on the website kakoune.org

PR #3955, opened 2020-12-15
doc commands: Sort command names

lenormf:
Back when those documentation pages were written, names (options, faces, hooks, key names, flags…) were sorted alphabetically first, and then by functionality. It worked fine to have, for example, `NormalIdle` close to `InsertIdle` at the top of the list, because the two modes are the most well known, and beneath them all related information was unrolled.

As time went by, more names came to be documented, and those lists more or less became append-only.

It's reasonably easy to find your way around them when one knows a priori how they are sorted, but confusing to a newcomer leveraging the ubiquitous prefix notation (`Normal*` hooks, `Menu*` faces etc.) or expecting names to be sorted alphabetically (option names).

So despite being well-intentioned originally, the double sorting has become counter productive, especially when names are mindlessly appended to them. This PR sorts all names alphabetically, which should help with the “organisation” complaints that bubbled up in the latest survey.

The keys page was not modified, but I think it'd be worth it to sort as well (somehow), if this PR gets merged.

Related to #3940

### DIFF
--- a/doc/pages/commands.asciidoc
+++ b/doc/pages/commands.asciidoc
@@ -20,10 +20,30 @@ of the file onto the filesystem
     given. Buffers that do not appear in the parameters will remain at the
     end of the list, keeping their current order.
 
+*buffer* <name>::
+    *alias* b +
+    switch to buffer <name>
+
+*buffer-next*::
+    *alias* bn +
+    switch to the next buffer.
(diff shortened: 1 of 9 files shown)


lenormf:
PING

lenormf:
PING

lenormf:
PING

PR #3958, opened 2020-12-17
rc lint: Always cleanup files/worker upon quitting

### LINKED ISSUE #3957: Temporary files are left behind when the editor is quit as linting is being performed
### Steps 

* Open a buffer, configure a linter
* Run `:lint`
* Quit the editor before the diagnostics have time to show up
* List files in your temporary directory (`/tmp` here)

### Outcome 

Temporary files were left behind in a `kak-lint.XXXXXX` directory.

### Expected 

The editor makes sure temporary files are removed if it's being quit before the linting process succeeds (after which those files are removed under normal circumstances).

lenormf:
This commit assigns a hook to buffers that are being linted that makes
sure all temporary files and asynchronous processes are cleaned up.

When a session is exited as the linting command is being executed in
the background, a `kak-lint.XXXXXX` directory remains in `$TMPDIR`,
and the process that parses the output of the linting command are
left dangling.

The buffer hook cleans both up when the buffer is closed, and is
unset when linting has completed.

Fixes #3957

### DIFF
--- a/rc/tools/lint.kak
+++ b/rc/tools/lint.kak
@@ -216,7 +216,16 @@ define-command \
         cat "$dir"/result > "$dir"/fifo
         rm -rf "$dir"
 
-        } & ) >"$dir"/stderr 2>&1 </dev/null
+        # We don't need the cleanup hook any more, the files are gone
+        printf "eval -buffer '%s' 'remove-hooks buffer lint-cleanup'" "$kak_bufname" | kak -p "$kak_session"
+
+        } & ) >"$dir"/stderr 2>&1 </dev/null &
+
+        readonly pid_async_worker=$!
+        # Make sure temporary files are cleaned up, asynchronous processes killed
+        printf "hook -once -always -group lint-cleanup buffer BufClose .* '
+            nop %%sh{ rm -rf \"%s\"; kill %d }
+        '" "$dir" "$pid_async_worker"
     }
 }
 


rc/tools/lint.kak
```diff
@@ -216,7 +216,16 @@ define-command \
...
+        # We don't need the cleanup hook any more, the files are gone
+        printf "eval -buffer '%s' 'remove-hooks buffer lint-cleanup'" "$kak_bufname" | kak -p "$kak_session"
+
+        } & ) >"$dir"/stderr 2>&1 </dev/null &
+
+        readonly pid_async_worker=$!
+        # Make sure temporary files are cleaned up, asynchronous processes killed
+        printf "hook -once -always -group lint-cleanup buffer BufClose .* '
+            nop %%sh{ rm -rf \"%s\"; kill %d }
+        '" "$dir" "$pid_async_worker"
```
mawww, line 228: This looks dangerous, is it guaranteed that $pid_async_worker still points to that process here and has not been reused ?

lenormf: I think somebody more knowledgeable than me about how shell works should confirm that.

[POSIX specification](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_05_02):

> !
> Expands to the decimal process ID of the most recent background command (see Lists) executed from the current shell. (For example, background commands executed from subshells do not affect the value of "$!" in the current shell environment.) For a pipeline, the process ID is that of the last command in the pipeline.

I assume `( … ) & pid=$!; …; kill $pid` should behave as expected.

Screwtapello: In general, there's no way to guarantee that a PID obtained at time X still refers to the same process at time Y. If you use the Linux-specific syscall [`pidfd_open()`](https://man7.org/linux/man-pages/man2/pidfd_open.2.html) to obtain a permanent process identifier, then a later call to `pidfd_send_signal()` will fail if the process has already exited, even if the PID has been re-used. However, there's still a chance that the PID has been recycled between the time it is obtained and the time `pidfd_open()` is called

lenormf: The PIDs on my system get into the nine digits over time, so I assumed they could not be re-used.

It's also a common pattern in service management to save the PID of a server into a `.pid` file, and try `kill`ing that number at an arbitrary, later time. Is that pattern wrong?

Are reusable PIDs really common in the wild, or is that irrelevant because we might end up killing the wrong process due to how linux works anyway? What about other kernels?

lenormf:
We cannot assume PIDs are not being re-used by the system, therefore this patch is likely to break.

PR #3959, opened 2020-12-17
rc scheme: Fix an Awk error

lenormf:
This commit prevents Awk from printing the following error:

```
awk: -:106: warning: escape sequence `\.' treated as plain `.'
```


eraserhd:
Well, oops.  Looks right to me.

lenormf:
PING

PR #3960, opened 2020-12-17
rc kitty: Correctly detect the terminal emulator

lenormf:
This commit checks the (non-standard) `$TERMINAL` environment variable, which Kitty sets to `kitty`.

Using `$TERM` as heuristic is not reliable, for example in the case when a multiplexer (Tmux) runs within Kitty and overrides this variable.

### DIFF
--- a/rc/windowing/kitty.kak
+++ b/rc/windowing/kitty.kak
@@ -5,7 +5,7 @@ provide-module kitty %{
 
 # ensure that we're running on kitty
 evaluate-commands %sh{
-    [ -z "${kak_opt_windowing_modules}" ] || [ "$TERM" = "xterm-kitty" ] || echo 'fail Kitty not detected'
+    [ -z "${kak_opt_windowing_modules}" ] || [ "$TERMINAL" = "kitty" ] || echo 'fail Kitty not detected'
 }
 
 declare-option -docstring %{window type that kitty creates on new and repl calls (kitty|os)} str kitty_window_type kitty


kovidgoyal:
FYI, kitty does not set any such environment variable, easily verified by grepping the kitty surce for TERMINAL. In general, env vars are not a reliable way to do terminal detection, since they are not propagated over ssh. The best way is to use escape codes to query the terminal. There are many choices for these, you can query the terminal name, the terminal's terminfo capabilities, etc. See for example here: https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Device-Control-functions

And for a kitty specific one, that allow querying not just terminal name, but version and arbitrary other data: https://sw.kovidgoyal.net/kitty/kittens/query_terminal.html

lenormf:
Oh, I set this variable myself, in my shell configuration! Thanks for the message.

I think `kitty +kitten query_terminal name` is appropriate, as we can't really play with escape sequences there.

That won't handle the SSH usecase unless Kitty is installed on the remote side either, but nobody has complained about that yet — it would still be nice to handle that.

kovidgoyal:
Just FYI, you dont have to use the kitten, it uses a standard escape
code to do its work. You can produce and consume that yourself, without
relying on the kitten.

lenormf:
Can that be done from a shell, with portable tools?

kovidgoyal:
No idea, I avoid shell, since python exists on all systems I care about.
It's trivial to do in python.

lenormf:
Thanks for your help!

Screwtapello:
Sending the escape sequence from the shell is easy.

Reading the response is more difficult. It's doable in `bash`, because bash's version of the `read` built-in supports a custom delimiter (escape sequences do not end with newline) and a custom timeout (if the terminal does *not* support the query, it won't respond at all). However, POSIX shell's `read` supports neither of those things.

It's worth pointing out how important the timeout is - the `query_terminal` kitten is documented as defaulting to a 10-second timeout, so using it to detect whether Kakoune is running inside Kitty implies adding worst-case 10 seconds to Kakoune's startup time.

kovidgoyal:
You dont actually need the timeout, you can use the same strategy as
described here: https://sw.kovidgoyal.net/kitty/graphics-protocol.html#querying-support-and-available-transmission-mediums

i.e. send the query escape code and also a more widely supported one
that requires a response, and only wait till the widely supported
response arrives.

PR #3965, opened 2020-12-18
feat(comment.kak): add OCaml and Coq cases

tototest99:


krobelus:
Thanks, this looks good. BTW you don't need to create a new PR to update commits - you can simply amend your commit and use `git push --force`.

Issue #3941, opened 2020-12-14
Improve commands discoverability


lenormf:
### Feature 

It's currently difficult to map out the possible commands available to the user, there should be a way to discover those more easily.

HTH.

alexherbo2:
On Discord we discussed about the behaviour of `n` and `/<ret>`.  What do you think about the discoverability of `/<ret>` for example?

- `n` will change the main selection (deselect) to select the next occurrence.
- `/<ret>` will move each selection to the next occurrence.

lenormf:
I assume the original report was referring to prompt commands, not primitives, although I can't be sure.

Regarding prompt commands, I assume the respondents mean that fuzzy matching is not enough, I suppose because if they don't know what they're looking for, they have nothing to type in.

Primitives and their side effects are generally covered in `:doc keys`, I don't know how else they could be better discovered (related #3945).

PR #3969, opened 2020-12-20
[RFC] Autogeneration of plugin documentation

lenormf:
Hi,

Some users have always complained that the documentation could be better, but the [last community survey](https://kakoune-editor.github.io/community-articles/2020/12/12/results_interpretation_first_community_survey.html) has revealed that they (not many, but still) expect it to contain *all* the information they need, they don't necessarily understand Kakoune script (nor do they want to), and they don't want to have to read code — all of these impact “discoverability”.

In order to make the documentation more exhaustive, thus helping discoverability that doesn't depend on users reading code, I did the following:

- modified the editor to print messages every time a significant piece of code (declaring an alias, a command, a face, a module, an option) was executed after `source`ing the built-in scripts
- wrote a Python script that processed those messages to generate an individual Asciidoc documentation file
- merged and edited the generated documentation with the existing one (as of writing, `lint.asciidoc`, `autorestore.asciidoc` and `doc.asciidoc`)

I would like to know what changes should be applied to the template the Python script uses to generate the documentation, if any. I've already identified several problems, I'll try to address them in proper PRs in parallel: capitalisation in docstrings, some descriptions end with a dot (and others don't), some optional parameters to commands are not documented in the docstring…

I included the patch/script in the commits of this branch, I don't think we need to maintain them going forward, once we've figured out a format we want for the documentation, we can just have a commit that creates all the files and maintain that.

Related to #3940 #3941


lenormf:
Note that once we're done with this, I plan on using the same script to generate HTML documentation (which can later on be turned into a docset).

lenormf:
Added support for user modes, apparently `man.kak` declares one (why?).

PR #3973, opened 2020-12-22
src: Catch “kill exceptions”’cleanly

### LINKED ISSUE #3972: `kak -E 'kill 0'` results in uncaught exception
### Steps 

Occurs on both `Kakoune v2020.09.01` and `Kakoune v2020.09.01-246-g74b6c20a`.

1. Run `kak -E 'kill 0'`

### Outcome 

`uncaught exception` is printed to stdout, and nothing else.

### Expected 

Something more descriptive than `uncaught exception` should be printed, explaining the problem (if there is one). I'm not familiar enough with the codebase (yet) to understand why this happens.

lenormf:
When the `kill` command is called in the `-E` CLI flag, the resulting
exception is not caught and crashes the server.

This commit allows the server to terminate cleanly.

Since `KakEnd` hooks also need to be executed should the user run a
command like `kak -E 'kill 0'`, the execution of `KakBegin` hooks is
now performed *before* the command provided in `-E` is executed. The
documentation for `KakBegin` (executed after the `-E` command prior to
this commit) consequently becomes more truthful, as it states:

	KakBegin session name
		kakoune has started, this hook is called just after
		reading the user configuration files

Fixes #3972

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -792,21 +792,27 @@ int run_server(StringView session, StringView server_init,
                                      "    {}", error.what()));
     }
 
+    {
+        Context empty_context{Context::EmptyContextFlag{}};
+        global_scope.hooks().run_hook(Hook::KakBegin, session, empty_context);
+    }
+
     if (not server_init.empty()) try
     {
         Context init_context{Context::EmptyContextFlag{}};
         command_manager.execute(server_init, init_context);
+    }
+    catch (const kill_session& kill)
+    {
+        Context empty_context{Context::EmptyContextFlag{}};
+        global_scope.hooks().run_hook(Hook::KakEnd, "", empty_context);
+        return kill.exit_status;
     }
     catch (runtime_error& error)
     {
         startup_error = true;
         write_to_debug_buffer(format("error while running server init commands:\n"
                                      "    {}", error.what()));
-    }
-
-    {
-        Context empty_context{Context::EmptyContextFlag{}};
-        global_scope.hooks().run_hook(Hook::KakBegin, session, empty_context);
     }
 
     if (not files.empty()) try


cole-h (approved):
Thanks for the quick fix! I can confirm this no longer crashes kakoune when doing e.g. `kak -E 'kill 0'`.

lenormf:
One usecase for this would be: `kak -n -E 'echo -to-file /tmp/out.log %val{runtime}; kill!'`.

lenormf:
PING

PJungkamp:
I also need this. I want to put a sanity check into the server init commands and kill the session with different exit codes depending on the failure. This PR would fix this.

Putting the `kill` into a `KakBegin` hook also causes this error.

I think the whole main function should be wrapped in a `try` block that catches the `kill_session` exception during **all** evaluations, including kakrc, server_init and e.g. the `KakBegin` hook.....

krobelus:
I'm surprised that people use `-E` for nontrivial scenarios.
I thought that feature was just for making cases like

    kak -E 'hook BufCreate ...' file1 file2

easier to type. But there is always the long form with `-e "edit file1; ..."` which also makes the ordering much easier to see.

But yeah I think we should fix that.
What do you think about this:
The weird return value is to update the exit_status when KakEnd throws kill_session.

```diff
diff --git a/src/main.cc b/src/main.cc
index 336e87a5d..060aab2a2 100644
--- a/src/main.cc
+++ b/src/main.cc
@@ -794,10 +794,11 @@ enum class ServerFlags
 };
 constexpr bool with_bit_ops(Meta::Type<ServerFlags>) { return true; }
 
-int run_server(StringView session, StringView server_init,
+void run_server(StringView session, StringView server_init,
                 StringView client_init, StringView init_buffer, Optional<BufferCoord> init_coord,
                 ServerFlags flags, UIType ui_type, DebugFlags debug_flags,
-               ConstArrayView<StringView> files)
+                ConstArrayView<StringView> files,
+                int& exit_status)
 {
     static bool terminate = false;
     set_signal_handler(SIGTERM, [](int) { terminate = true; });
@@ -805,7 +806,8 @@ int run_server(StringView session, StringView server_init,
     if ((flags & ServerFlags::Daemon) and session.empty())
     {
         write_stderr("-d needs a session name to be specified with -s\n");
-        return -1;
+        exit_status = -1;
+        return;
     }
 
     EventManager        event_manager;
@@ -841,6 +843,9 @@ int run_server(StringView session, StringView server_init,
     }
 #endif
 
+    Optional<OnScopeEnd<std::function<void()>>> run_session_close_hooks;
+    try
+    {
         bool startup_error = false;
         if (not (flags & ServerFlags::IgnoreKakrc)) try
         {
@@ -855,6 +860,23 @@ int run_server(StringView session, StringView server_init,
                                          "    {}", error.what()));
         }
 
+        {
+            Context empty_context{Context::EmptyContextFlag{}};
+            global_scope.hooks().run_hook(Hook::EnterDirectory, real_path("."), empty_context);
+            global_scope.hooks().run_hook(Hook::KakBegin, session, empty_context);
+            run_session_close_hooks = on_scope_end(std::function{[&]() {
+                Context empty_context{Context::EmptyContextFlag{}};
+                try
+                {
+                    global_scope.hooks().run_hook(Hook::KakEnd, "", empty_context);
+                }
+                catch (const kill_session& kill)
+                {
+                    exit_status = kill.exit_status;
+                }
+            }});
+        }
+
         if (not server_init.empty()) try
         {
             Context init_context{Context::EmptyContextFlag{}};
@@ -867,12 +889,6 @@ int run_server(StringView session, StringView server_init,
                                          "    {}", error.what()));
         }
 
-    {
-        Context empty_context{Context::EmptyContextFlag{}};
-        global_scope.hooks().run_hook(Hook::EnterDirectory, real_path("."), empty_context);
-        global_scope.hooks().run_hook(Hook::KakBegin, session, empty_context);
-    }
-
         if (not files.empty()) try
         {
             for (auto& file : files)
@@ -899,9 +915,6 @@ int run_server(StringView session, StringView server_init,
              write_to_debug_buffer(format("error while opening command line files: {}", error.what()));
         }
 
-    int exit_status = 0;
-    try
-    {
         if (not server.is_daemon())
         {
             local_client = client_manager.create_client(
@@ -977,13 +990,6 @@ int run_server(StringView session, StringView server_init,
     {
         exit_status = kill.exit_status;
     }
-
-    {
-        Context empty_context{Context::EmptyContextFlag{}};
-        global_scope.hooks().run_hook(Hook::KakEnd, "", empty_context);
-    }
-
-    return exit_status;
 }
 
 int run_filter(StringView keystr, ConstArrayView<StringView> files, bool quiet, StringView suffix_backup)
@@ -1278,7 +1284,9 @@ int main(int argc, char* argv[])
                              ((argc == 1 or (ignore_kakrc and argc == 2))
                               and isatty(0)                               ? ServerFlags::StartupInfo : ServerFlags::None);
                 auto debug_flags = option_from_string(Meta::Type<DebugFlags>{}, parser.get_switch("debug").value_or(""));
-                return run_server(session, server_init, client_init, files.empty() ? StringView{} : files[0], init_coord, flags, ui_type, debug_flags, files);
+                int exit_status = 0;
+                run_server(session, server_init, client_init, files.empty() ? StringView{} : files[0], init_coord, flags, ui_type, debug_flags, files, exit_status);
+                return exit_status;
             }
             catch (convert_to_client_mode& convert)
             {
```

Issue #3978, opened 2020-12-23
Add `:debug register` to inspect the value of all registers


cole-h:
### Feature 

A new subcommand for `:debug` that displays the contents of all registers would be useful.

* Should the subcommand be `register` or `registers`?
* Should only registers with content be printed (if there's a way to differentiate)?
  * If there is a way to differentiate, should there be a `-all` flag to print all registers regardless of state?

### Usecase 

It would be nice to be able to inspect the value of all registers (the most obvious example I can make is to help newbies like me determine if they need to use `-save-regs` with `eval` or `exec`).

tototest99:
`echo -debug register a is: %reg{a}`
You can define a command for multiple registers based on this

caksoylar:
Also see: https://github.com/Delapouite/kakoune-registers

PR #3980, opened 2020-12-23
commands: add "registers" subcommand to :debug 

cole-h:
This prints all non-empty registers and their value(s) to the `*debug*`
buffer.

---

This is definitely not the prettiest code I've ever written. It's been a while since I've written C++, as well, so please point out any unidiomatic / otherwise bad code so I can fix it!

~~My main dissatisfaction with this implementation is how the registers are hardcoded. I was thinking I could iterate through all registers through the `m_registers` HashMap, but that is protected and I can't access it directly. Suggestions greatly appreciated.~~

cc @lenormf for putting me up to this ;)

Closes https://github.com/mawww/kakoune/issues/3978.

---

image

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -1392,7 +1392,7 @@ const CommandDesc debug_cmd = {
         [](const Context& context, CompletionFlags flags,
            const String& prefix, ByteCount cursor_pos) -> Completions {
                auto c = {"info", "buffers", "options", "memory", "shared-strings",
-                         "profile-hash-maps", "faces", "mappings", "regex"};
+                         "profile-hash-maps", "faces", "mappings", "regex", "registers"};
                return { 0_byte, cursor_pos, complete(prefix, cursor_pos, c) };
     }),
     [](const ParametersParser& parser, Context& context, const ShellContext&)
@@ -1484,6 +1484,20 @@ const CommandDesc debug_cmd = {
 
             write_to_debug_buffer(format(" * {}:\n{}",
                                   parser[1], dump_regex(compile_regex(parser[1], RegexCompileFlags::Optimize))));
+        }
+        else if (parser[0] == "registers")
+        {
+            write_to_debug_buffer("Register info:");
+            for (auto&& [name, reg] : RegisterManager::instance())
+            {
+                auto content = reg->get(context);
+
+                if (content.size() == 1 and content[0] == "")
+                    continue;
+
+                write_to_debug_buffer(format(" * {} = {}\n", name,
+                    join(content | transform(quote), "\n     = ")));
+            }
         }
         else
             throw runtime_error(format("no such debug command: '{}'", parser[0]));
--- a/src/register_manager.hh
+++ b/src/register_manager.hh
@@ -124,6 +124,9 @@ public:
     void add_register(Codepoint c, std::unique_ptr<Register> reg);
     CandidateList complete_register_name(StringView prefix, ByteCount cursor_pos) const;
 
+    auto begin() const { return m_registers.begin(); }
+    auto end() const { return m_registers.end(); }
+
 protected:
     HashMap<Codepoint, std::unique_ptr<Register>, MemoryDomain::Registers> m_registers;
 };


src/commands.cc
```diff
@@ -1485,6 +1485,35 @@ const CommandDesc debug_cmd = {
             write_to_debug_buffer(format(" * {}:\n{}",
                                   parser[1], dump_regex(compile_regex(parser[1], RegexCompileFlags::Optimize))));
         }
+        else if (parser[0] == "registers")
+        {
+            write_to_debug_buffer("Register info:");
+            const auto& register_manager = RegisterManager::instance();
+            for (Codepoint c : StringView{"\"/@^|%.#_:\\0123456789abcdefghijklmnopqrstuvwxyz"})
```
mawww, line 1492: Add a `begin/end` const method to RegisterManager that return `m_registers.begin/end`, then this line can become `for (auto&& [name, reg] : RegisterManager::instance())`.

src/commands.cc
```diff
@@ -1485,6 +1485,35 @@ const CommandDesc debug_cmd = {
...
+            {
+                bool first = true;
+                String s;
+
+                for (String val : register_manager[c].get(context))
```
mawww, line 1497: This whole loop looks like it could be written `String s = join(reg->get(context), "\n     = ");` However I think we should also quote the contents to make the output less ambiguous (like if some element is the empty string), so it go with something like:

```
auto content = reg->get(context);
if (content.size() == 1 and content[0] == "") // empty register always return a single empty string
    continue;
write_to_debug_buffer(format("* {} = {}\n", c, join(content | transform(quote), "\n     = ")));
```

mawww (changes requested):

src/commands.cc
cole-h: I opted to make a "custom" quote transform to avoid the `double_up("'")` in the `quote` transform (if any register's value is `'`, it would actually display as `''`, e.g. ` * a = ''''` rather than ` * a = '''`).

cole-h:
Output now looks like the following (minor change; contents surrounded in single quotes):

image

That one empty line of `:` history is due to my config in some way; `kak -n` doesn't exhibit the same behavior.

src/commands.cc
```diff
@@ -1485,6 +1485,20 @@ const CommandDesc debug_cmd = {
...
+            write_to_debug_buffer("Register info:");
+            for (auto&& [name, reg] : RegisterManager::instance())
+            {
+                auto content = reg->get(context);
+
+                if (content.size() == 1 and content[0] == "")
+                    continue;
+
+                write_to_debug_buffer(format(" * {} = {}\n", name,
+                    join(content | transform([](StringView s) { return format("'{}'", s); }), "\n     = ")));
```
mawww, line 1499: there is a function called `quote` in string_utils.hh it will also double up contained `'` to ensure the string matches what Kakoune can parse in command line.

src/register_manager.cc
```diff
@@ -102,4 +102,14 @@ CandidateList RegisterManager::complete_register_name(StringView prefix, ByteCou
     return complete(prefix, cursor_pos, reg_names | transform([](auto& i) { return i.key; }) | gather<Vector<String>>());
 }
 
+RegisterMap::const_iterator RegisterManager::begin() const
```
mawww, line 105: those could be directly inlined in the header as `auto begin() const { return m_registers.begin(); }`.

src/register_manager.hh
```diff
@@ -116,6 +116,8 @@ public:
     }
 };
 
+using RegisterMap = HashMap<Codepoint, std::unique_ptr<Register>, MemoryDomain::Registers>;
```
mawww, line 119: I think this should be nested inside RegisterManager, although it should not be necessary anymore if you use auto deduction for begin/end as suggested in the previous comment.

src/commands.cc
cole-h: Oh, I see now. I was looking at the visual representation, but indeed `echo ''''` just prints a single `'`. Will change that back to your suggestion.

src/register_manager.cc
cole-h: I didn't know what would be more preferable (e.g. `auto` vs explicit type); I'll do as you suggest.

PR #3983, opened 2020-12-23
Fix minor typos in VIMTOKAK

toddyamakawa:


lenormf:
I don't think these are mistakes, some new comers have trouble understanding what the `<a-…>` syntax means, so it seems adequate for a guide to use the “long form”.

toddyamakawa:
Does that mean that Line 42 is the typo where the short-form is used?
https://github.com/mawww/kakoune/blob/master/VIMTOKAK#L42

lenormf:
Probably. I have the feeling this line is formatted that way so that users can use it as-is into a prompt in the editor, but I suppose writing `alt-i` and spacing out the following keys would make everything more readable.

toddyamakawa:
Anyone reading the VIMTOKAK document is already coming from vim, and `<a-i>` is pretty readable to a lot of vim-users.

Regardless, `alt-ld` is a lot less readable than `<a-l>d` because the former could be misinterpreted as `<a-l><a-d>` or even typing out each character `alt-ld` rather than pressing the alt-key. Maybe instead of `<a-l>d`, something like `<alt-l>d` is even less ambiguous?

PR #3985, opened 2020-12-23
Remove non-POSIX -r from ln in Makefile

codesoap:
`-r` is not mentioned in [the POSIX standard](https://pubs.opengroup.org/onlinepubs/009695399/utilities/ln.html) and does lead to an error with OpenBSD's `ln`.

I'm not yet really sure what `-r` does, so feel free to correct me here.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -172,7 +172,7 @@ installdirs:
 
 install: kak man installdirs
 	install -m 0755 kak $(bindir)
-	ln -rsf $(bindir)/kak $(libexecdir)/kak
+	ln -sf $(bindir)/kak $(libexecdir)/kak
 	install -m 0644 ../share/kak/kakrc $(sharedir)
 	install -m 0644 ../doc/pages/*.asciidoc $(sharedir)/doc
 	cp -r ../rc/* $(sharedir)/rc


Screwtapello:
In GNU coreutils, `-r` means "ensure this is a relative symlink, even if the given path is absolute":

```
$ mkdir /tmp/ln-test/
$ cd /tmp/ln-test/
$ touch foo
$ ln -sf /tmp/ln-test/foo bar
$ ln -rsf /tmp/ln-test/foo qux
$ ls -al
total 576
drwxr-xr-x  2 st   st     4096 Dec 24 10:21 .
drwxrwxrwt 26 root root 581632 Dec 24 10:21 ..
lrwxrwxrwx  1 st   st       16 Dec 24 10:21 bar -> /tmp/ln-test/foo
-rw-r--r--  1 st   st        0 Dec 24 10:21 foo
lrwxrwxrwx  1 st   st        3 Dec 24 10:21 qux -> foo
```

In the context of `make install`, this means the `libexec` symlink will always be a relative link to the kak binary, which means you can tarball up the installation prefix and copy it to a different location without breaking anything, even if the installation prefix is absolute.

Instead, probably what we should do here is to remove all the libexecdir stuff from the Makefile, and instead change the line to read:

    ln -sf ../../../bin/kak $(bindir)/../libexec/kak/kak

...that way, the symlink will always be relative, and it will always be where Kakoune looks for it. Some may argue that Kakoune should respect `libexecdir` because somebody might want to install their libexec stuff somewhere else, but Kakoune [already hardcodes that path](https://github.com/mawww/kakoune/blob/74b6c20a0d19c9fcf2b30a93de1ff3b715151e65/src/shell_manager.cc#L75) so we might as well hard-code it the same way everywhere.

codesoap:
Thanks for the explanation, I now understand why `-r` was used here. What I don't yet understand is: Why is it important to use relative paths at all? I don't feel like it's a real-world-scenario to tar system directories, untar them on a different system with different directory hierarchy and expect it to work.

I took a closer look at my `/usr/local/libexec` directory and found, that `kak` is the only softlink there. All other files are either "normal files" or hardlinks. Using a hardlink seems like a good idea to me, what is the reason for kakoune using softlinks here?

Screwtapello:
> I don't feel like it's a real-world-scenario to tar system directories, untar them on a different system with different directory hierarchy and expect it to work.

The use-case here is a developer who is using a distro like Arch or Ubuntu (possibly in a VM) on their laptop, but they have to administer/interact with a production environment based on RHEL or CentOS which does not package Kakoune, packages an ancient version, and probably packages an ancient C++ compiler that cannot build the latest version of Kakoune.

From the Kakoune repo, you can `make -C src static=yes DESTDIR=/tmp/kakoune-staging`, then tar up the `/tmp/kakoune-staging` directory and extract it to `/usr/local/bin` or `~/.local/bin` or similar on the target machine.

codesoap:
> The use-case here is a developer who is using a distro like Arch or Ubuntu (possibly in a VM) on their laptop, but they have to administer/interact with a production environment based on RHEL or CentOS which does not package Kakoune [...]

Thanks for the insight! I had never considered doing something like this.

> As for soft-links versus hard-links, I guess somebody just didn't want to assume bindir and libexecdir would be on the same filesystem?

Sounds plausible. I had almost forgotten about this limitation of hardlinks. I agree that this makes softlinks a little more correct.

PR #3996, opened 2020-12-30
main: add buffilelist variable

cole-h:
This makes it easier to script on the list of buffers, especially when
using $kak_hook_param e.g. in a BufClose hook.

---

This helps with my attempt at scripting alternate buffer functionality (might be a bit over-engineered and / or could be done simpler, but...):

<details>

```kak
## bufhist list for better alternate buffer functionality
declare-option str-list bufhist
declare-option str bufhist_last

define-command -hidden last_buf %{
  evaluate-commands %sh{
    source "$kak_opt_prelude_path"

    # populate the previous buffer
    last="$(printf %s "$kak_opt_bufhist" | perl -pe "s@\\Q$kak_buffile\\E@@g" | tr -s ' ')"
    last="${last%% }"
    last="${last##* }"
    [ -n "$last" ] && [ "${kak_buffilelist#*"$last"}" != "$kak_buffilelist" ] && \
      kak_escape set-option global bufhist_last "$last"

    # if current buffer is the "last" buffer, move it to the beginning of
    # the history list (will be recalculated next time a buffer changes,
    # so this shouldn't change much)
    if [ "$kak_opt_bufhist_last" = "$kak_buffile" ]; then
      file="${kak_opt_bufhist##* }"
      last="$(printf %s "$kak_opt_bufhist" | perl -pe "s@\\Q$file\\E@@g" | tr -s ' ')"
      last="${last%% }"
      last="${last## }"
      last="$file $last"
      kak_escape set-option global bufhist "$last"
    fi
  }
}

define-command -hidden bufhist %{
  evaluate-commands %sh{
    source "$kak_opt_prelude_path"

    # populate the list of buffers
    bufhist="$(printf %s "$kak_opt_bufhist" | perl -pe "s@\\Q$kak_buffile\\E@@g" | tr -s ' ')"
    bufhist="${bufhist%% }"
    bufhist="${bufhist## }"
    bufhist="${bufhist:+$bufhist }$kak_buffile"
    kak_escape set-option global bufhist "$bufhist"
  }
}

hook -always global WinDisplay .* %{
  evaluate-commands %{
    # populate the previous buffer
    last_buf

    # populate the list of buffers
    bufhist
  }

  hook -once -always global BufClose .* %{
    evaluate-commands %sh{
      source "$kak_opt_prelude_path"

      # remove closed buffers
      closed="$kak_hook_param"
      if [ -n "$closed" ]; then
        bufhist="$(printf %s "$kak_opt_bufhist" | perl -pe "s@\\Q$closed\\E@@g" | tr -s ' ')"
        bufhist="${bufhist%% }"
        bufhist="${bufhist## }"
        kak_escape set-option global bufhist "$bufhist"
      fi

      # update last buffer now that a buffer has been removed
      printf "%s\n" last_buf
    }
  }
}
```

</details>

### DIFF
--- a/doc/pages/expansions.asciidoc
+++ b/doc/pages/expansions.asciidoc
@@ -201,7 +201,7 @@ The following expansions are supported (with required context _in italics_):
     `%val{bufname}`)
 
 *%val{buffilelist}*::
-    quoted list of the full path of currently-open buffers (as seen in
+    quoted list of the full file paths of currently-open buffers (as seen in
     `%val{buffile}`)
 
 *%val{bufname}*::
--- a/doc/pages/expansions.asciidoc
+++ b/doc/pages/expansions.asciidoc
@@ -200,6 +200,10 @@ The following expansions are supported (with required context _in italics_):
     quoted list of the names of currently-open buffers (as seen in
     `%val{bufname}`)
 
+*%val{buffilelist}*::
+    quoted list of the full path of currently-open buffers (as seen in
+    `%val{buffile}`)
+
 *%val{bufname}*::
     _in buffer, window scope_ +
     name of the current buffer
--- a/src/main.cc
+++ b/src/main.cc
@@ -195,6 +195,10 @@ static const EnvVarDesc builtin_env_vars[] = { {
         [](StringView name, const Context& context) -> Vector<String>
         { return BufferManager::instance() | transform(&Buffer::display_name) | gather<Vector>(); }
     }, {
+        "buffilelist", false,
+        [](StringView name, const Context& context) -> Vector<String>
+        { return BufferManager::instance() | transform(&Buffer::name) | gather<Vector>(); }
+    }, {
         "buf_line_count", false,
         [](StringView name, const Context& context) -> Vector<String>
         { return {to_string(context.buffer().line_count())}; }


Screwtapello:
I feel like your code snippet could be simplified to something like:

```
declare-option str-list bufhist

hook global WinDisplay .* %{
    # move the current buffer to the end of the list
    set-option -remove global bufhist %val{hook_param}
    set-option -add global bufhist %val{hook_param}
}

hook global BufClose .* %{
    # This is no longer a valid buffer, remove it from the list
    set-option -remove global bufhist %val{hook_param}
}
```

It's a little more difficult to switch to the buffer at the end of the list than at the beginning, but not too much so:

```
eval set -- "$kak_quoted_opt_bufhist"
lastbuf=""
for bufname; do
    lastbuf="$bufname"
done

if [ -z "$lastbuf" ]; then
    kak_escape fail No other buffers available.
else
    kak_escape buffer "$lastbuf"
fi
```

I'm not sure how a `%val{buffilelist}` expansion helps here.

cole-h:
I guess it was more than "a bit over-engineered"...

Thanks for the help, that actually made it a ton easier. I had to make minor modifications and came up with this:

(the `lastbuf` would actually be the current buffer, so strip that off. `perl` is necessary for its `quotemeta` / `\Q`/`\E` functionality to prevent `*debug*` and the like from being expanded)

```
define-command alt-buf %{
  evaluate-commands %sh{
    source "$kak_opt_prelude_path"

    eval set -- "$kak_quoted_opt_bufhist"
    lastbuf=""
    for bufname; do
      lastbuf="$bufname"
    done

    bufs="$(printf %s "$kak_opt_bufhist" | perl -pe "s@\\Q$lastbuf\\E@@g" | tr -s ' ')"
    bufs="${bufs%% }"
    bufs="${bufs## }"
    prev="${bufs##* }"

    if [ -z "$prev" ]; then
      kak_escape fail 'No other buffers available.'
    else
      kak_escape buffer "$prev"
    fi
  }
}
```

Screwtapello:
How about:

```
    eval set -- "$kak_quoted_opt_bufhist"
    # toss all but the last two items in the list
    shift $(( $# - 2 ))

    if [ $# -ne 2 ]; then
      # The history list didn't have enough items in it.
      kak_escape fail "No other buffers available"
    else
      # Having dropped all the preceding items,
      # $1 is the previous buffer and $2 is the current buffer
      kak_escape buffer "$1"
    fi
```

I'm trying to avoid the `sed` and string-slicing techniques you're using because I think they'll break horribly if a filename has a space in it. Using (argument) array operations means no worrying about escaping or shell-sensitive characters.

cole-h:
You're a magician (and 100% correct). Though I don't have many paths with spaces in them, this will surely be nice for the few that I end up editing that do.

Thanks a ton for all your help.

PR #3997, opened 2020-12-30
Bug fix when selection start with -

co-dh:
When $1 or ${kak_selection} start with dash, like "-1", the command will fail, because tmux think it's an argument flag.
-- prevent this.


co-dh:
Also the document of this function: "(append new line)" is no longer valid.  Preparing PR.

PR #4000, opened 2021-01-01
syntax highlighting for Q (kdb+)

co-dh:


co-dh:
@mawww  Any chance that this get merged in? or it's too niche a language?

mawww:
Hello,

I never heard of the Q language, looking it up it does sound very niche, so I think this should probably be provided as an external plugin.

co-dh:
I might be the only kdb programmer on the world that using Kakoune :(

co-dh:
As APL get syntax colour, I'd like to have this one merge if possible.
Q is a descendent of APL, a great language, and used a lot in finance.

PR #4002, opened 2021-01-01
Add new libexec directory to SPEC file packaging

jkonecny12:
Without that the build of RPM file will fail.

-------------------------------
The question is if instead we don't want to remove spec file from here completely. I moved daily builds automation to my fork so I don't need upstream spec file anymore, I will instead keep a copy on my fork. However, I don't have a problem to keep this spec file up-to-date for the community.
Do the upstream want the spec file here or you would rather want to remove it completely?

### DIFF
--- a/contrib/kakoune.spec
+++ b/contrib/kakoune.spec
@@ -6,7 +6,7 @@
 
 Name:           kakoune
 Version:        2020.09.01
-Release:        1%{?dist}
+Release:        2%{?dist}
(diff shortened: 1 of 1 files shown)


mawww:
It really depends if that file is useful to anybody else, if nobody really would use that spec file then I think we can remove it. I actually did use it as a base to generate internal rpms at my work, so I guess thats one data point in favor of keeping it.

Issue #3999, opened 2020-12-31
Cannot use `:` in builtin mode bindings


cole-h:
### Steps 

1. `map global goto a ':fail lol<ret>'`
2. `gh`
3. `ga`

### Outcome 

```
l lol
map global goto a ':fail lol<ret>'
```

Due to how mappings work, this means `goto_commands()` in `src/normal.cc` is seeing `:` as the key it needs to executed, which is obviously unbound in that mode. The same is true for `view` mode.

### Expected 

The `:` prompt should execute `fail lol`.

Screwtapello:
Mappings in insert mode are executed in insert mode, mappings in normal mode are executed in normal mode, mappings in prompt mode are executed in prompt mode. The outlier here is user modes - mappings in user modes are *also* executed in normal mode.

Mappings let you redefine which keys invoke which Kakoune operations, but the operations are just the built-in mappings. So if you `map global normal j k` the `j` is what gets redefined, and the `k` is the built-in mapping for "go up a line".

Insert mode has built-in mappings you can map onto; normal mode has built-in mappings you can map onto; prompt mode has built-in mappings you can map onto, but user modes by definition do not have any built-in mappings, so the right-hand-side has to be interpreted in some other mode, and Normal mode is the obvious choice.

Because modes like `goto` and `view` look like user-modes, it's reasonable to expect them to behave the same way, but they *do* have built-in mappings that you might want to map onto. In fact the similar `<a-a>` and `<a-i>` modes have built-in mappings (specifically, `c` and `<a-semicolon>`) whose only purpose is to have custom mappings mapped onto them!

So, the behaviour you describe is consistent, but the rule is "does this mode have built-in mappings", not "does this mode present an info box of available mappings".

cole-h:
Thanks for your explanation, that makes sense.

Do you think it would make sense to have a binding to `:` / `;` similar to that of `object` mode's `<a-semicolon>` in both `goto` and `view` modes? Or would their behaviors be too inconsistent for those contexts?

---

For now, though, I'll just have to settle for my own user mode, copying all the built-in commands except for the one I want to modify.

alexherbo2:
You can prefix the command with `<esc>` or `<a-semicolon>` for a single command to an object menu to acceed `%val{count}`, `%val{select_mode}` and `%val{object_flags}`.

Try:

```
2<a-i><a-semicolon>echo %val{count} %val{select_mode} %val{object_flags}<ret>
```

[...]

PR #4007, opened 2021-01-04
Support commands in more submodes

cole-h:
Limited to `goto` and `view` modes, but I'm sure it'd be easy to add to other modes. Ref https://github.com/mawww/kakoune/issues/3999#issuecomment-753874219.

Note: `command()` was moved up for it to be usable within `goto_commands()` and `view_commands()`.

I wonder what would be the best binding. I used `;` for now, but `<a-;>` would also work.


cole-h:
cc @alexherbo2

PR #4008, opened 2021-01-05
Add strikethrough text ability, 'smxx'

craigmac:


lenormf:
Have you verified that the upstream `tmux-256color.terminfo` source doesn't define it already? If it has changed, it'd be nice to update the whole `contrib/tmux-256color.terminfo` file (if other capacities have been modified), or otherwise have the upstream file fixed to support `smxx`.

craigmac:
> Have you verified that the upstream `tmux-256color.terminfo` source doesn't define it already? If it has changed, it'd be nice to update the whole `contrib/tmux-256color.terminfo` file (if other capacities have been modified), or otherwise have the upstream file fixed to support `smxx`.

I did a curl of the most current terminfo, `cd /tmp && curl -O https://invisible-mirror.net/archives/ncurses/current/terminfo.src.gz`  and installed it with `gunzip /tmp/terminfo.src.gz && /usr/bin/tic -x /tmp/terminfo.src`.

Then I dumped it with `infocmp -1x tmux-256color > /tmp/newest`. 

I diffed it with my current terminfo from this repo and there indeed is a `smxx=\E\9m` attribute in the latest terminfo file, so that's good news. The only other difference is the new terminfo uses `ind=^J` whereas the contrib file uses `ind=\n`, is that relevant to us?

I'm quite new to all the terminfo business, but if you need further digging I'm game!

lenormf:
If the upstream `tmux-256color.terminfo` works for you as-is, you could simply add `ccc` to it and update the file in `contrib` in this PR, instead of manually backporting capacities we're missing.

I don't know what the `ind=^J` really involves for the terminal, I only know that hitting <kbd>c-j</kbd> in one inserts a new line, so I'm not too worried it will break anything. Maybe someone who knows more about this could confirm.

Screwtapello:
`man 5 terminfo` says that `ind` is the short name for `scroll_forward`, described as "scroll text up" (it's not mentioned, but I believe it's short for "index").

Regardless, `J` in ASCII is `\x4A`, so Ctrl-J clears the top 3 bits to make `\x0A`, which is also known as LF or (on Unix) "newline". Writing `^J` or `\n` is exactly equivalent.

craigmac:
Ok I can change it so that we use the upstream version for ind so it’s one less difference, and just add the ccc to our contrib file.

PR #4011, opened 2021-01-05
Markdown: Add HTML and improve URL highlighting

### LINKED ISSUE #3979: markdown highlighting with html attributes
is there a way to exclude the underscore `_` value of an html attribute from being identified/highlighted as italics text from a markdown file, as long as, and obviously, is enclosed within its html respective opening and closing tags `<`, `>`. Then only reason I mentioned the delimiters/symbols/signs `<` and `>` is simply because I think is customary in html to write it within double quotes. And with or without it makes no difference on the error. Looking at the regex of `markdown.kak` I wouldn't know where to start really. 

e.g of the annoyance`<a href="https://google.com" target="_blank">this text should not be emphasized</a>` 

Other html elements that I presume would have the same issue 
```	
_self	
_parent	
_top		
```

possibly related after writing the above https://github.com/mawww/kakoune/issues/2111

Ordoviz:
Closes #3979:
-  Prevent underscores in URLs and HTML tags cause italic highlighting
- Add highlighting for inline links and HTML tags

Relative links (i.e. all links without a URI scheme like `http://` or `mailto:`) are not highlighted—and underscores in them can still cause problems—but the full [CommonMark link spec](https://spec.commonmark.org/0.29/#links) is just too complex.

Before:
<pre>
http://example.org#lorum_<i>ipsum not emphasized</i>_
<a href="http://example.org#lorum ipsum_dolor">&lt;http://example.org#lorum ipsum_<i>dolor></i></a> <i>not emphasized</i>_
&lt;a target="_<i>blank">not emphasized</i>_</a>
[Wikipedia](https://en.wikipedia.org/wiki/Element_<i>(software)) not emphasized</i>_
</pre>

After:
<pre>
<a href="http://example.org#lorum_ipsum">http://example.org#lorum_ipsum</a> not emphasized_
&lt;<a href="http://example.org#lorum ipsum_dolor">http://example.org#lorum ipsum_dolor</a>> not emphasized_
&lt;<b>a</b> target="_blank">not emphasized_</a>
[Wikipedia](<a href="https://en.wikipedia.org/wiki/Element_(software)">https://en.wikipedia.org/wiki/Element_(software)</a>) not emphasized_
</pre>

### DIFF
--- a/rc/filetype/markdown.kak
+++ b/rc/filetype/markdown.kak
@@ -59,9 +59,17 @@ add-highlighter shared/markdown/codeblock region -match-capture \
     ^(\h*)```\h*$ \
     fill meta
 
-add-highlighter shared/markdown/listblock region ^\h*[-*]\s ^\h*((?=[-*])|$) group
-add-highlighter shared/markdown/listblock/ ref markdown/inline
-add-highlighter shared/markdown/listblock/marker regex ^\h*([-*])\s 1:bullet
+add-highlighter shared/markdown/listblock region ^\h*[-*]\s ^\h*((?=[-*])|$) regions
+add-highlighter shared/markdown/listblock/g default-region group
+add-highlighter shared/markdown/listblock/g/ ref markdown/inline
+add-highlighter shared/markdown/listblock/g/marker regex ^\h*([-*])\s 1:bullet
+
+# https://spec.commonmark.org/0.29/#link-destination
+# This repetition is not pretty but shell escaping is worse
+add-highlighter shared/markdown/angle_bracket_url region (?<=<)([a-z]+://|(mailto|magnet|xmpp):) (?!\\).(?=>)|\n fill link
+add-highlighter shared/markdown/url region -recurse \( ([a-z]+://|(mailto|magnet|xmpp):) (?!\\).(?=\))|\s fill link
+add-highlighter shared/markdown/listblock/angle_bracket_url region (?<=<)([a-z]+://|(mailto|magnet|xmpp):) (?!\\).(?=>)|\n fill link
+add-highlighter shared/markdown/listblock/url region -recurse \( ([a-z]+://|(mailto|magnet|xmpp):) (?!\\).(?=\))|\s fill link
 
 add-highlighter shared/markdown/inline/code region -match-capture (`+) (`+) fill mono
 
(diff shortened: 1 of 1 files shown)


rc/filetype/markdown.kak
```diff
@@ -59,9 +59,20 @@ add-highlighter shared/markdown/codeblock region -match-capture \
...
+add-highlighter shared/markdown/url region -recurse \( ([a-z]+://|(mailto|magnet|xmpp):) (?!\\).(?=\))|\s fill link
+add-highlighter shared/markdown/listblock/angle_bracket_url region (?<=<)([a-z]+://|(mailto|magnet|xmpp):) (?!\\).(?=>)|\n fill link
+add-highlighter shared/markdown/listblock/url region -recurse \( ([a-z]+://|(mailto|magnet|xmpp):) (?!\\).(?=\))|\s fill link
+
+require-module html
+add-highlighter shared/markdown/tag region </?(a|abbr|acronym|address|applet|area|article|aside|audio|b|base|basefont|bdi|bdo|big|blockquote|body|br|button|canvas|caption|center|cite|code|col|colgroup|data|datalist|dd|del|details|dfn|dialog|dir|div|dl|dt|em|embed|fieldset|figcaption|figure|font|footer|form|frame|frameset|h[1-6]|head|header|hr|html|i|iframe|img|input|ins|kbd|label|legend|li|link|main|map|mark|meta|meter|nav|noframes|noscript|object|ol|optgroup|option|output|p|param|picture|pre|progress|q|rp|rt|ruby|s|samp|script|section|select|small|source|span|strike|strong|style|sub|summary|sup|svg|table|tbody|td|template|textarea|tfoot|th|thead|time|title|tr|track|tt|u|ul|var|video|wbr)\b > ref html/tag
```
krobelus, line 75: I wonder if these should be in a `try` block, in case the user doesn't load html.kak, though that seems unlikely.

lenormf: Out of consistency, this `require-module` probably shouldn't be there, to keep in line with the bit above:

https://github.com/mawww/kakoune/blob/df7b33bc7b9da897d1d0127b54ae1e629d2334ff/rc/filetype/markdown.kak#L43-L54

Now it's possible that this loop *should* be calling `require-module` itself, but it's not really self-evident, so maybe we should take care of that in a subsequent issue/PR (or not).

lenormf:
I'm not convinced this is the right fix, after all regions exist so we don't have to embed logic for one language into another's highlighters. Having a (long) list of hardcoded HTML attributes in `markdown.kak` doesn't seem right.

rc/filetype/markdown.kak
Ordoviz: The `require-module html` needs to be there for `ref html/tag` to work. Sure, the `require-module` command that gets run when you type <code>```html</code> into a Markdown document is redundant but there does not seem to be any issue with running `require-module html` twice (the description for the `require-module` command is “ensures that <var>name</var> module has been loaded”).

We could replace `ref html/tag` with `group` but this results in poorer highlighting and makes debugging harder if the `markdown/tag` highlighter causes a false positive (I am not aware of such a case but you never know).

Ordoviz:
> Having a (long) list of hardcoded HTML attributes in markdown.kak doesn't seem right.

I agree. It turns out that pandoc and GitHub even treat non-existing HTML elements as HTML elements: If you type `Let x<y and x>1` then `<y and x>` is treated as a HTML tag, meaning the `<` does not get escaped to `&lt;`.

Thus, taking the restrictions that [CommonMark](https://spec.commonmark.org/0.29/#raw-html) imposes on HTML tags into account, the hardcoded list should be replaced with `(?i)</?[a-z][a-z0-9-]*\s*([a-z_:]|(?=>))`.

Pandoc does not permit HTML attribute names to start with  `_` or `:` but GitHub does.

mawww:
Wanted to merge that, howerver it breaks the existing markdown tests, could you check those ?

Ordoviz:
The `indent/markdown` tests were failing because markdown.kak now requires the html module. I put the relevant part in a `try %{ }` block to make it optional.

Issue #3970, opened 2020-12-21
`man.kak` declares a user mode


lenormf:
Hi,

I see that `man.kak` declares a user mode:

https://github.com/mawww/kakoune/blob/74b6c20a0d19c9fcf2b30a93de1ff3b715151e65/rc/tools/man.kak#L133-L142

This is surprising to me, as *user* modes seem like they should remain in the *user* configuration. The name of this mode is also fairly simple, which could potentially lead to confusion on the user's side when they try to declare a mode with the same name, but are told it's already taken.

So my question is: are user modes not expected to be controlled entirely by the user? If not should we not declare more user modes in built-in scripts (as an intermediary between mappings and commands), if yes should we move this one to the wiki (or better, `man.asciidoc`)?

Thanks!

Screwtapello:
As a third-party plugin author, I like being able to include a user-mode, since it lets me present the plugin's functionality in an accessible way without stomping on the user's normal mappings, and without requiring them to read all the documentation and figure out for themselves which commands are most directly useful, and how to wire them together.

For example, kakoune-cargo [includes a user mode](https://gitlab.com/Screwtapello/kakoune-cargo/-/blob/a9047d698658930cd57537ccc39b65546b4a4136/cargo.kak#L142-153) that exposes its most important functionality, as does [kak-lsp](https://github.com/kak-lsp/kak-lsp/blob/34e888eea109ad1b2fbaee3b0b7afab46b38f1c8/rc/lsp.kak#L1287-L1304).

It seems reasonable that first-party plugins might want to do the same thing, although most of them aren't complex enough to need it... or maybe they are, and if they included user-modes I'd have learned about their useful extra features by now.

mawww:
I did not remember `man.kak` did define a user mode, but I dont see a big issue with it, I think plugins should not map a user mode to a key themselves, however they are welcome to define a user mode that can be easily bound to a key by the user.

lenormf:
Alright, so should this be spread over to other scripts as well? If #3698 gets merged, that's going to create a lot of commands, and indexing them into a single user mode might be convenient.

lenormf:
#4016 also creates a user mode, it follows the same naming nomenclature (`XXX-mode`), but I'm thinking it should drop the `-mode` to make the command to enter it more elegant:

`enter-user-mode man-mode` → `enter-user-mode man`
`enter-user-mode clipboard-mode` → `enter-user-mode clipboard`

lenormf:
[...]

Issue #1555, opened 2017-09-07
Set `termcmd` to distribution specific terminal launchers when possible


lenormf:
Hi,

Debian provides with the `x-terminal-emulator` wrapper script that automatically picks the terminal configured by the user, or auto-detect it otherwise. The `i3` window manager also has the `i3-sensible-terminal` utility that basically does the same thing.

Maybe those two (along with other similar utilities I don't know about) should be placed at the top of the list in `x11.kak` to follow the user's system configuration?

mawww:
I dont know about that, `x-terminal-emulator` would probably be fine as its distribution provided, and hopefully various distributions standardize on that name, but using `i3-sensible-terminal` seems a bit far fetched.

lenormf:
What about [xdg-terminal](https://cgit.freedesktop.org/xdg/xdg-utils/tree/scripts/xdg-terminal.in) ? It seems to handle the default terminals for the most popular desktop environments, but I don't know if this script is available by default on those systems to warrant using it.

`x-terminal-emulator` seems to be more well known (at least in the Debian community), I think it's a good fit.

mawww:
I'd accept a PR adding x-terminal-emulator and/or xdg-terminal to `x11.kak` terminal detection code. I expect these would probably need to be the highest priority one. One thing that needs to be checked is that these commands take care of unifying the "command to execute" syntax for the underlying terminal.

PR #4016, opened 2021-01-08
rc clipboard: Support system clipboard ineractions

### LINKED ISSUE #3935: Built-in clipboard integration
### Feature 

A substantial amount of users complains that the editor doesn't integrate with their system's clipboard out of the box.

If that's not going to happen, is there anything we can do to meet them halfway? I was thinking that the editor hardcodes multiple commands already (for example, to interact with window managers like Tmux, Kitty…), and has some selection logic (`termcmd`).

Could we not provide users with a command (or user mode? other?) based API that works along the same lines, so that they at least wouldn't have to look for the right `xclip`/`xsel`/… command? It would then be up to them to declare mapping in their user configuration.

HTH.

lenormf:
This commit implements an API to interact with the system clipboard.

Similarly to how terminals are picked by `x11.kak`, the
`clipboardcmd_copy` and `clipboardcmd_paste` options detect which tool
should be used to interact with the system clipboard at startup. Only
`xsel` has been tested, but `pbcopy`/`pbpaste` and `wl-copy`/`wl-paste`
will be used if they're installed. These options can also be overridden
by the user to use `xclip` or any other tool they want.

Commands allow copying, pasting before/after, replacing
selections. Data to copy to the clipboard can also be passed directly
to `clipboard-copy`.

A convenience alias forwards `clipboard-paste` to
`clipboard-paste-after` to save users, who expect pasting to always
append, some brain cycles.

The `clipboard-mode` user mode is also declared to accommodate user
mappings, and assigned predictable key mappings.

Savvier users can enable register integration with
`clipboard-enable-register`, which currently only works one way: all
contents assigned to the `c` register is automatically forwarded to
the system clipboard. Vim uses the `*` and `+` registers, but
punctuation signs are generally associated with “core” primitives
in Kakoune, so using a letter register for this opt-in layer of
abstraction seemed more fitting.

Implementing the register integration in the opposite direction
in a portable way is not self-evident, and thus not implemented by
this patch.

Fixes #3935.


lenormf:
I've been using this for some time (albeit occasionally), if everyone is happy with the API I would write some documentation for the plugin.

It would also be great if some people could test the code, to make sure it's not broken and is not too awkward to use!

rc/tools/clipboard.kak
```diff
@@ -0,0 +1,94 @@
+declare-option -docstring %{
+    The shell command to which data will be piped to copy to the system clipboard
+} str clipboardcmd_copy %sh{
+    for cmd in "xsel -i" "pbcopy" "wl-copy -p"; do
+        readonly program="${cmd%% *}"
```
Screwtapello, line 5: I don't think `readonly` is POSIX (at least, it's not in `dash`).

rc/tools/clipboard.kak
```diff
@@ -0,0 +1,94 @@
+declare-option -docstring %{
+    The shell command to which data will be piped to copy to the system clipboard
+} str clipboardcmd_copy %sh{
+    for cmd in "xsel -i" "pbcopy" "wl-copy -p"; do
```
Screwtapello, line 4: You might also want to check for [vis-clipboard](https://github.com/martanne/vis/blob/v0.6/vis-clipboard), a tool from the `vis` editor that abstracts over all those other tools. It's *probably* not going to be installed anywhere Kakoune is, but if it *is* installed, it's pretty nice.

rc/tools/clipboard.kak
```diff
@@ -0,0 +1,94 @@
...
+'
+        printf %s "$*"
+    }
+
+    if [ $# -eq 0 ]; then
+        eval "set -- $kak_quoted_selections"
+    fi
+
+    if ! join "$@" | eval "${kak_opt_clipboardcmd_copy}"; then
+        fail "Unable to copy data to the system clipboard"
```
Screwtapello, line 44: Wait, where is `fail` defined?

rc/tools/clipboard.kak
```diff
@@ -0,0 +1,94 @@
...
+    fi
+} }
+
+define-command -docstring %{
+    Paste the contents of the system clipboard before the selection
+} clipboard-paste-before %{
+    execute-keys ! %opt{clipboardcmd_paste} <ret>
```
Screwtapello, line 51: Add `<space>` here to prevent the paste command being added to the history register?

rc/tools/clipboard.kak
```diff
@@ -0,0 +1,94 @@
...
+}
+
+define-command -docstring %{
+    Automatically copy the contents of the `c` register to the clipboard when it is modified
+} clipboard-enable-register %{
+    try clipboard-disable-register
+    hook -group clipboard-bind-register global RegisterModified c %{
+        clipboard-copy %reg{c}
+    }
+}
```
Screwtapello, line 76: Why not take the register to hook as a parameter? I imagine a lot of people would want to `:clipboard-enable-register dquote` — that probably shouldn't be the default, but it should be easy to do.

rc/tools/clipboard.kak
```diff
@@ -0,0 +1,94 @@
...
+
+# NOTE: Users who discover the command interactively might not want to
+# have to think about the before/after side effects, so assume they expect
+# pasting to always be performed after the selection
+alias global clipboard-paste clipboard-paste-after
+
+declare-user-mode clipboard-mode
+
+map global clipboard-mode -docstring "Copy selection to clipboard" y ": clipboard-copy<ret>"
```
Screwtapello, line 91: It's not strictly necessary, but people might expect to find "Cut selection to clipboard" here. I guess the implementation would be `: clipboard-copy; exec d<ret>` (since `exec` saves and restores the default register, `exec d` would delete the selection without disturbing any other registers).

rc/tools/clipboard.kak
lenormf: Hmm I can see it in the [OpenGroup publication](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_23), but I suppose if it doesn't work on Dash it will have to go.

I'll have to remove it where there too, since I thought this keyword was portable and started using it everywhere:

https://github.com/mawww/kakoune/blob/df7b33bc7b9da897d1d0127b54ae1e629d2334ff/rc/detection/modeline.kak#L63-L64

rc/tools/clipboard.kak
lenormf: I didn't know about this, I'll probably set the `clipboardcmd_*` options to it in my configuration!

I'm not sure this should be in the built-in list though, I envisioned that it would only contain tools that are well-known and ubiquitous.

A similar solution was discussed for `terminalcmd` and `x-terminal-emulator` in #1555.

rc/tools/clipboard.kak
lenormf: I forgot to add `echo` ;)

rc/tools/clipboard.kak
lenormf: I considered it, but then thought that it'd be better to stay in line with the “pick a good default” line the editor has followed.

I guess that Vi users would rather the register be `+` or `*`, but as the commit message says, punctuation signs are generally used by core primitives.

I also wanted to have the register documented in the docstring so that users wouldn't have to read the code itself, which is difficult to implement if the register is arbitrary.

rc/tools/clipboard.kak
SolitudeSF: whats the point of wrapping a wrapper that does the same thing?

lenormf: Only have the right tool installed on a given system (or in a container), and not worry whether it's supported by the editor.

rc/tools/clipboard.kak
occivink: You could also accept a parameter, but default to `c` if there are none.

lenormf: That was the idea, but how to insert the given register into the docstring? It was simpler to hardcode `c` and let users override the command from `autoload`.

rc/tools/clipboard.kak
Screwtapello: I'm sorry, I've made a mistake - I'm sure `readonly` wasn't in the dash manpage when I checked yesterday, but it's definitely there today, and testing it does seem to work as intended:

    /bin/dash: 3: foo: is read only

rc/tools/clipboard.kak
occivink: I'm not sure what is the problem, the docstring can simply mention that the register passed as argument is used, or 'c' if there are none?

Screwtapello: Well, Kakoune doesn't have `+` or `*` registers, so it doesn't really matter whether punctuation signs are used by core primitives, they're just not available. Besides, Vim also has the `set clipboard += unnamed` option which makes yank/delete/change/put interact with the clipboard register instead of the unnamed register, and I suspect *that's* how a lot of people interact with the clipboard from Vim rather than `"*y`.

rc/tools/clipboard.kak
Screwtapello: > whats the point of wrapping a wrapper that does the same thing?

For one, `vis-clipboard` wraps more backends than just xsel, pbcopy and wl-copy. For two, it's smarter about detecting which one to use - for example, it only tries xsel if the X11 `$DISPLAY` variable is set. That's all logic users don't have to re-implement in Kakoune.

krobelus: > for example, it only tries xsel if the X11 $DISPLAY variable is set.

We should probably also do that. Many users have both `xsel` and `wl-{copy,paste}`.
We should first check if `WAYLAND_DISPLAY` is nonempty.

As the logic gets more refined, we should probably use just one shell scope to initialize both `clipboardcmd_copy` and `clipboardcmd_paste`.

rc/tools/clipboard.kak
```diff
@@ -0,0 +1,95 @@
...
+} clipboard-disable-register %{
+    remove-hooks global clipboard-bind-register
+}
+
+# NOTE: Users who discover the command interactively might not want to
+# have to think about the before/after side effects, so assume they expect
+# pasting to always be performed after the selection
+alias global clipboard-paste clipboard-paste-after
+
+declare-user-mode clipboard-mode
```
krobelus, line 89: So currently all user-modes defined by Kakoune's rc scripts have a ``-mode``  suffix. I wonder if it's better to drop it, or maybe use a prefix like `kak-clipboard`.

lenormf: I mentioned that in [another issue](https://github.com/mawww/kakoune/issues/3970#issuecomment-756757800), I'm leaving that for a future PR.

rc/tools/clipboard.kak
lenormf: Came up with this, not sure if this is really better than what we already have:

```
define-command -params ..1 -docstring %{
    clipboard-enable-register [<register>]: automatically copy the contents of the given register to the clipboard, when it is modified

    If no register is passed as argument, the `c` register will be used
    The built-in copy register can be referred to as `dquote`, instead of `"`
} -shell-script-candidates %{
    echo dquote
} clipboard-enable-register %{ evaluate-commands %sh{
    reg="${1:-c}"

    # Make sure the register is valid
    printf 'nop %%reg{%s}\n' "${reg}"

    case "${reg}" in
        dquote) reg='"';;
    esac

    # NOTE: The following will break with the closing-brace register
    printf '
        try clipboard-disable-register
        hook -group clipboard-bind-register global RegisterModified %%{%s} %%{
            clipboard-copy %%reg{%s}
        }
    ' "${reg}" "${reg}"
} }
```

rc/tools/clipboard.kak
lenormf: I don't think verifying whether the tool selected can actually work is necessary, we don't want to re-implement a robust abstraction in what I see as a convenience.

It would make the code heavier, and the editor can reasonably expect an X-less environment not to have `xsel` installed in the first place. I'd leave all that logic to a wrapper like `vis-clipboard` if it's important to the user.

krobelus: I think the additional check is worth it. Consider `nvim` and `pass` who both support copy-to-clipboard on wayland out of the box, even when xsel/xclip are installed.

>  the editor can reasonably expect an X-less environment not to have xsel installed

I think it's also not unreasonable to have both installed.

rc/tools/clipboard.kak
lenormf: I pushed this new implementation. One user indicated that they would like to have seamless integration with the system clipboard when writing prose, but use a letter register for code, which is a good reason to have this more complicated implementation.

lenormf:
PING

lenormf:
Drop the `-mode` suffix on the user mode.

Screwtapello (approved):
As much as I'd love to have real native clipboard integration in Kakoune, or the hooks required to implement such a thing (a `RegisterReadPre` hook, for example, or some kind of "paste" and "paste request" events in the JSON UI protocol), that would take a bunch of thought and design and implementation work.

On the other hand, this PR solves 80% of the problem and is here now.

Given how often new users get confused or frustrated with clipboard integration, I think this is worth merging.

rc/tools/clipboard.kak
```diff
@@ -0,0 +1,112 @@
...
+    reg="${1:-c}"
+
+    # Make sure the register is valid
+    printf 'nop %%reg{%s}\n' "${reg}"
+
+    case "${reg}" in
+        dquote) reg='"';;
+    esac
+
+    # NOTE: The following will break with the closing-brace register
```
krobelus, line 86: no need to worry, there is no closing-brace register

rc/tools/clipboard.kak
```diff
@@ -0,0 +1,112 @@
...
+
+    if ! join "$@" | eval "${kak_opt_clipboardcmd_copy}"; then
+        echo fail "Unable to copy data to the system clipboard"
+    fi
+} }
+
+define-command -docstring %{
+    Paste the contents of the system clipboard before the selection
+} clipboard-paste-before %{
+    execute-keys ! <space> %opt{clipboardcmd_paste} <ret>
```
krobelus, line 51: the space is not necessary, everything that's run inside a command or mapping does not add to history.
Same for the mappings at the bottom of the file.

rc/tools/clipboard.kak
```diff
@@ -0,0 +1,112 @@
...
+            printf %s "${cmd}"
+            exit
+        fi
+    done
+}
+
+declare-option -docstring %{
+    The shell command which will paste the system clipboard into the current buffer
+} str clipboardcmd_paste %sh{
+    for cmd in "xsel -o 2>/dev/null" "pbpaste" "wl-paste -p"; do
```
krobelus, line 16: wl-paste needs the `--no-newline` switch to print the clipboard as-is

Kratacoa:
bump?

stacyharper:
I think Kakoune should, as Tmux does, support `set-clipboard` first. And then, eventually fallback to an auto-detected program. It make clipboard over ssh to works, and make things easier for Kakoune, as this rely to the terminal to do the connection with the wm clipboard.

ref: https://github.com/tmux/tmux/wiki/Clipboard

Book-reader:
After trying this I have found that the `-p`  flag on `wl-copy` and `wl-paste` with Hyprland causes `wl-copy` to hang forever (only in kakoune, not anywhere else) and `wl-paste` to paste the last text selected in Hyprland, not the last copied text.
removing the `-p` flag from `wl-copy` and `wl-paste` fixes it and does not seem to cause any other issues, at least on Hyprland.

PR #4023, opened 2021-01-14
Improve graphql syntax highlighters

gf3:
Add more advanced highlighters for graphql syntax. There are still some small issues, however I believe this is mostly due to the use of regexps for syntax highlighting.

### Before
2021-01-13T21:21:20,304983500-05:00

### After
2021-01-13T21:36:36,912971421-05:00

### DIFF
--- a/rc/filetype/graphql.kak
+++ b/rc/filetype/graphql.kak
@@ -4,7 +4,7 @@
 # Detection
 # ‾‾‾‾‾‾‾‾‾
 
-hook global BufCreate .*[.](graphql) %{
+hook global BufCreate .*[.](graphqls?) %{
     set-option buffer filetype graphql
 }
 
--- a/rc/filetype/graphql.kak
+++ b/rc/filetype/graphql.kak
@@ -43,7 +43,7 @@ add-highlighter shared/graphql/object region -recurse \{ [{] [}] regions
 add-highlighter shared/graphql/object/line-description region '#' '\n' fill comment
 add-highlighter shared/graphql/object/block-description region '"""' '"""' fill comment
 add-highlighter shared/graphql/object/field default-region group
-add-highlighter shared/graphql/object/field/ regex ([A-Za-z][A-Za-z0-9_-]*)(?:\([\w\d\h,:$=]*\))?\h*[:{] 1:attribute
+add-highlighter shared/graphql/object/field/ regex ([A-Za-z][A-Za-z0-9_-]*)(?:\([\w\d\h,:$=!"_]*\))?\h*[:{] 1:attribute
 add-highlighter shared/graphql/object/field/ regex ^\h*([A-Za-z][A-Za-z0-9_-]*)\h*$ 1:attribute
 
 # Values
--- a/rc/filetype/graphql.kak
+++ b/rc/filetype/graphql.kak
@@ -49,7 +49,8 @@ add-highlighter shared/graphql/object/field/ regex ^\h*([A-Za-z][A-Za-z0-9_-]*)\
 # Values
 add-highlighter shared/graphql/object/field/values regex \b(true|false|null|\d+(?:\.\d+)?(?:[eE][+-]?\d*)?)\b 0:value
 add-highlighter shared/graphql/object/field/variables regex \$[a-zA-Z0-9]+\b 0:variable
-add-highlighter shared/graphql/object/field/string regex '"[^"]*"' 0:string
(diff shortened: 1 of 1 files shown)


gf3:
someone more expierienced may know the answer to this, but for some reason i was unable to get negated groups in regexs to work for certain patterns 

e.g. i had to do the following for fields with arguments:

```
add-highlighter shared/graphql/object/field/ regex ([A-Za-z][A-Za-z0-9_-]*)(?:\([\w\d\h,:$=!"_]*\))?\h*[:{] 1:attribute
```

when ideally i'd like to do (see `\([^)]\)`):

```
add-highlighter shared/graphql/object/field/ regex ([A-Za-z][A-Za-z0-9_-]*)(?:\([^)]*\))?\h*[:{] 1:attribute
```

it should work: regexr.com/5k7cb

Ordoviz:
> when ideally i'd like to do (see \([^)]\)):

That line of code is inside a `%( )` block. The closing parenthesis in `[^)]` has no opening equivalent so the block ends at that parenthesis. I recommended switching to a `%§ §` block to avoid issues like this. See `:doc expansions` for details.

gf3:
thanks @Ordoviz! works as expected, i've updated the PR

rc/filetype/graphql.kak
```diff
@@ -27,22 +27,46 @@ hook -group graphql-highlight global WinSetOption filetype=graphql %{
...
+add-highlighter shared/graphql/object/line-description region '#' '\n' fill comment
+add-highlighter shared/graphql/object/block-description region '"""' '"""' fill comment
+add-highlighter shared/graphql/object/field default-region group
+add-highlighter shared/graphql/object/field/ regex ([A-Za-z][A-Za-z0-9_-]*)(?:\([^)]*\))?\h*[:{] 1:attribute
+add-highlighter shared/graphql/object/field/ regex ^\h*([A-Za-z][A-Za-z0-9_-]*)\h*$ 1:attribute
+
+# Values
+add-highlighter shared/graphql/object/field/values regex \b(true|false|null|\d+(?:\.\d+)?(?:[eE][+-]?\d*)?)\b 0:value
+add-highlighter shared/graphql/object/field/variables regex \$[a-zA-Z0-9]+\b 0:variable
+add-highlighter shared/graphql/object/field/string regex '"[^"]*"' 0:string
```
eraserhd, line 52: This seems like it would not catch escaped quotes as before.  Maybe something like `'"([^"]|\\.)*"'`?

rc/filetype/graphql.kak
```diff
@@ -27,22 +27,46 @@ hook -group graphql-highlight global WinSetOption filetype=graphql %{
...
 # ‾‾‾‾‾‾‾‾‾‾‾‾
 
 add-highlighter shared/graphql regions
 add-highlighter shared/graphql/code default-region group
-add-highlighter shared/graphql/string region '"' (?<!\\)(\\\\)*" fill string
-add-highlighter shared/graphql/comment region '#' \n fill comment
+add-highlighter shared/graphql/line-description region '#' '\n' fill comment
+add-highlighter shared/graphql/block-description region '"""' '"""' fill comment
+add-highlighter shared/graphql/description region '"' '"\s*\n' fill comment
+add-highlighter shared/graphql/object region -recurse \{ [{] [}] regions
```
eraserhd, line 40: I haven't checked, but won't this get confused by `{ foo(s: "-- } ") { bar } }`?

rc/filetype/graphql.kak
gf3: good catch! fixed

rc/filetype/graphql.kak
gf3: yes you're absolutely right, i'm not sure this can be accounted for with just regexes or without supporting a additional nested string region?

luckily string literals within field arguments are pretty uncommon directly in graphql, the values are typically passed in programmatically from the host language. so while this is a problem, i think it's unlikely to surface in the wild

m0ar: > yes you're absolutely right, i'm not sure this can be accounted for with just regexs or without supporting an additional nested string region?
> 
> luckily string literals within field arguments are pretty uncommon directly in graphql, the values are typically passed in programmatically from the host language. so while this is a problem, i think it's unlikely to surface in the wild

I just hit this in the wild fwiw, and have no idea how to get around it :monocle_face:

PR #4031, opened 2021-01-19
rc spell: Remove `:spell-replace` default value

lenormf:
This commit removes the default prompt value from the `spell-replace`
command.

Currently, running the command after selecting a misspelled word
might not allow the editor to propose alternative spellings because
it completes upon whatever is inserted into the prompt. If the words
returned by `aspell` are too different from the currently misspelled
word, no candidates are shown.

For example, selecting “unanymously” and running `:spell-replace`
will not show any candidates under the current implementation (the
‘y’ probably trips the fuzzy-matcher).

The user develops a habit of clearing the prompt every time, because
that's the only way to make sure all suggestions from `aspell` are
visible, so the editor might as well not have any default value for
`:spell-replace`.

### DIFF
--- a/rc/tools/spell.kak
+++ b/rc/tools/spell.kak
@@ -148,7 +148,6 @@ define-command \
     -docstring "Suggest replacement words for the current selection, against the last language used by the spell-check command" \
     spell-replace %{
     prompt \
-        -init %val{selection} \
         -shell-script-candidates %{
             options=""
             if [ -n "$kak_opt_spell_last_lang" ]; then


Screwtapello:
Here's my reasoning for defaulting to the original spelling:

- If you accidentally trigger a `:spell-replace` mapping, `<ret>` will leave the buffer unmodified
- The correct spelling is probably pretty close to the original spelling, so it makes sense to let the user edit the original rather than have to retype most of it to narrow down the completion list
- The correct spelling might not be in the completion list at all, because it's an unusual word and the suggestion engine considers five other words more likely, and again the user shouldn't have to retype the whole thing

As a comparison, the LibreOffice spell-checker gives you the original spelling in an editable text box, and a separate list of suggested replacements:

image

A dialog like this isn't possible with the user-interface components Kakoune provides (`:prompt`, `:menu`, `:info`, etc.) but I figured `:prompt` with a default was the closest thing. Of course, it still may not be close *enough*, and maybe removing the default is an overall improvement. I'd be interested to hear what other people think.

For myself, I generally just hold backspace until I see the word I want in the completion list, rather than clearing the entire prompt every time. I feel like that lets me find the correct word faster than starting from an empty prompt and tabbing through the options, but I confess I haven't timed it or anything.

lenormf:
I figured that was the intent behind the current behaviour.

However:

- hitting <kbd>ret</kbd> does modify the buffer, the selected word is simply replaced by itself
- <kbd>esc</kbd> on the prompt will cancel the replacing, it doesn't matter what the default value is in that particular scenario
- the correct spelling is often different enough from the default value that it will trip the completion engine, resulting in no candidates being displayed at (as if Aspell didn't return any suggestions)
- if the correct spelling isn't suggested, the user can simply hit <kbd>esc</kbd>, no need to re-type the word

In fact, I suppose that my problem is that the new implementation of `:spell-replace` (f75f484) doesn't behave like a menu more (allowing the user to type in characters to refine the list candidates is good though).

Relying on the editor's completion engine to *only* provide the best spelling suggestion is not robust, as we can see with my example that only has a one-letter mistake in it but results in no candidates being shown at all.

PR #4032, opened 2021-01-19
 doc: clarification in regards to define-command containing multiple commands

Kratacoa:


lenormf:
I don't think that description is accurate, as commands are not just split using the space character as a separator.

Kratacoa:
> I don't think that description is accurate, as commands are not just split using the space character as a separator.

You're right, I've rethought it after making the PR.
However, I am not sure where it is documented what characters are used as separators; I assume they are the same as for shell scripts, but I'm not sure how to put that succinctly (and whether it is true in the first place). Any ideas?

Kratacoa:
Nevermind, I'm just clueless. It's described [here](https://github.com/mawww/kakoune/blob/master/doc/pages/commands.asciidoc#multiple-commands). Sorry.

PR #4035, opened 2021-01-21
Respect KITTY_LISTEN_ON when launching terminal

kayhide:
When connecting to an existing session from another shell and executing `terminal`, new window appears in the original shell, not in the currently working shell.

This PR relaxes this problem by taking account of `KITTY_LISTEN_ON` environment variable of the client side.

----

Here is a background why I want this.

Using `connect` plugin, I often open a shell and connect to the current session by:

```
map global space-terminal a ': $<space>kitty :a<ret>' -docstring 'open new client'
```

I rather do this than letting kitty to open a window in the same shell because I want to let the window manager handle whole layout in a centralized way.

This command works fine but, `terminal` command does not work as expected and targets on the original shell.

I found this can be solved by starting kitty with `--listen-on` and targeting on that socket when remotely controlling.

The line above can be updated as:

```
map global space-terminal a ': $<space>kitty --listen-on "unix:@%sh{date +%s}" :a<ret>' -docstring 'open new client'
```

And with this PR, things will work fine.

### DIFF
--- a/rc/windowing/kitty.kak
+++ b/rc/windowing/kitty.kak
@@ -15,7 +15,7 @@ kitty-terminal <program> [<arguments>]: create a new terminal as a kitty window
 The program passed as argument will be executed in the new terminal' \
 %{
     nop %sh{
-        if [[ -z ${kak_client_env_KITTY_LISTEN_ON:-} ]]; then
+        if [ -z "$kak_client_env_KITTY_LISTEN_ON" ]; then
             kitty @ new-window --no-response --window-type "$kak_opt_kitty_window_type" --cwd "$PWD" "$@"
         else
             kitty @ --to "$kak_client_env_KITTY_LISTEN_ON" new-window --no-response --window-type "$kak_opt_kitty_window_type" --cwd "$PWD" "$@"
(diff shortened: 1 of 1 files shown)


rc/windowing/kitty.kak
```diff
@@ -15,7 +15,11 @@ kitty-terminal <program> [<arguments>]: create a new terminal as a kitty window
 The program passed as argument will be executed in the new terminal' \
 %{
     nop %sh{
-        kitty @ new-window --no-response --window-type $kak_opt_kitty_window_type --cwd "$PWD" "$@"
+        if [[ -z ${kak_client_env_KITTY_LISTEN_ON:-} ]]; then
```
Screwtapello, line 18: `[[` is a bashism, for portability just use `[`. Also, trying to "default" (`:-`) to an empty string is redundant, because an unset environment variable already expands to an empty string.

```suggestion
        if [ -z "$kak_client_env_KITTY_LISTEN_ON" ]; then
```

kayhide:
@Screwtapello Hey, thanks for your review.
Okay, I applied your suggestion.

kayhide:
I noticed that the same thing is happening to the other `kitty-*` functions as well.
Updated those functions in a same way.

Issue #4034, opened 2021-01-20
regex character class \s does not match NO-BREAK SPACE (Unicode A0).


avalenn:
### Steps 
`%s\s+<enter>`
### Outcome 
Only ASCII whitespaces selected.
### Expected 
no-break space characters to be selected too, possibly also applicable to all characters in Unicode Zs General Category.

### Additionnal information
From @Screwtapello on IRC
> `:doc regex` doesn't seem to mention that character-class escapes are only defined for ASCII characters, so it seems reasonable to file an issue.
> Kakoune usually follows the ECMAScript regex syntax, but that says \s "Matches a single white space character, including space, tab, form feed, line feed, and other Unicode spaces. Equivalent to [ \f\n\r\t\v\u00a0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]."

7h3w1zz:
I have created a pull request for a tentative fix to this.  I can write some tests if the implementation looks fine.

PR #4045, opened 2021-02-03
Follow ECMA specification for regex whitespace

7h3w1zz:
Tentative fix for #4034.


src/unicode.hh
```diff
@@ -20,12 +20,42 @@ inline bool is_eol(Codepoint c) noexcept
 
 inline bool is_horizontal_blank(Codepoint c) noexcept
 {
-    return c == ' ' or c == '\t';
+    // Characters considered whitespace by ECMA Regex Spec
+    // <https://262.ecma-international.org/11.0/#sec-white-space>
+    return c == '\t'      or
+           c == '\v'      or
```
krobelus, line 26: A vertical tab is not a horizontal blank, see

`printf '\v' | perl -ne 'print if m/\h/'`

7h3w1zz: Fair enough, fixed.
Test added.
I looked through the rest of the characters, but I did not see any other vertical space characters that are also not line terminators.

7h3w1zz:
Any other fixes needed for this?

PR #4047, opened 2021-02-04
Fix editorconfig incorrectly matching properties with prefixes

vktec:
A custom editorconfig property such as `ij_continuation_indent_size=8` is currently matched as `indent_size=8` by editorconfig.kak. This PR fixes that bug.

### DIFF
--- a/rc/detection/editorconfig.kak
+++ b/rc/detection/editorconfig.kak
@@ -20,13 +20,13 @@ define-command editorconfig-load -params ..1 -docstring "editorconfig-load [file
             /*) # $kak_buffile is a full path that starts with a '/'
                 printf %s\\n "remove-hooks buffer editorconfig-hooks"
                 editorconfig "$file" | awk -v file="$file" -F= -- '
-                    /indent_style=/             { indent_style = $2 }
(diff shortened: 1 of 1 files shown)


rc/detection/editorconfig.kak
```diff
@@ -20,13 +20,13 @@ define-command editorconfig-load -params ..1 -docstring "editorconfig-load [file
             /*) # $kak_buffile is a full path that starts with a '/'
                 printf %s\\n "remove-hooks buffer editorconfig-hooks"
                 editorconfig "$file" | awk -v file="$file" -F= -- '
-                    /indent_style=/             { indent_style = $2 }
-                    /indent_size=/              { indent_size = $2 == "tab" ? 4 : $2  }
```
lenormf, line 24: Do we not need the ternary comparison any more?

vktec: It's duplicated below

lenormf: From what I understand, if `indent_size="tab"` and `tab_width` is not set, then line 41 in the `END` scope will print out `set-option buffer tabstop tab`.

vktec: Good point, I'll remove the other one instead

PR #4051, opened 2021-02-07
Update julia.kak

clad26:
Enabling features like auto-indent for julia files.

### DIFF
--- a/rc/filetype/julia.kak
+++ b/rc/filetype/julia.kak
@@ -4,7 +4,7 @@
 # Detection
 # ‾‾‾‾‾‾‾‾‾
 
-hook global BufCreate .*\.(jl) %{
+hook global BufCreate .*[.](jl) %{
     set-option buffer filetype julia
 }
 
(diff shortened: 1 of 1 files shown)


clad26:
@mawww I hope this gets merged. I just changed the julia.kak file, so this isn't a breaking change.

mawww:
Hello,

Sorry for the huge delay in revewing that, can you add a copyright waiver commit to it ? (see the CONTRIBUTING file) ?

VarLad:
@mawww Sorry for responding late :sweat_smile: 
Added the copyright waiver

VarLad:
@mawww Uhhh.... any update?

mbarbar:
Thanks for this! some quick playing around and it seems quite nice. I've updated my local installation with these changes.

Having a `function` line followed by a `begin` then some statements then an `end` results in all indentation for that last line being removed. For example:
```
function cool()
    begin
        println("cool")
end
```
Adding `begin` to line 87 seems to fix it? But I'm not sure, I'm a bit unfamiliar with a lot of these commands.

Also, `end` sometimes gets indented (not deindented)
```
function cool()
    if true
        if false
        end
    end # This was manually deindented before typing the next end.
        end
```
This one I'm not really sure of the reason why.

PR #4059, opened 2021-02-11
Move space from :comment-line to comment_line

SeerLite:
Move the space put after the `comment_line` the `comment-line` command to each of the `comment_line` options in comment.kak.
This allows me to finally just do:
```kak
hook global BufSetOption 'comment_line=(.+) +' %{
	set buffer comment_line %val{hook_param_capture_1}
}
```
to get rid of the space.

Why:
I never use `:comment-line` to write comments, but I do use it to comment out code temporarily. The space inserted after the comment character makes it hard for me to discern actual comments (that I wrote with a space) from  commented out code (no space).


SeerLite:
I could also add a `comment_line_sep` option that's set to a single space by default. That way it's just a matter of emptying it to get it the way I want, and I don't have to set up a hook for it.

Not sure if such an option is too specific though, I can't think of many usecases. What do you think?

PR #4074, opened 2021-02-23
Have C-u scroll cursor up even when window is at top

bollu:
Here are [two gifs](https://imgur.com/a/66OYfiJ) that compares the current and proposed behaviour. The proposed behaviour causes `C-u` to move the cursor upward when the window position is already on the first line. This behaviour is consistent with vim's `C-d/C-u` motions.

My code is terrible, unfortunately. I literally copy-pasted the cursor adjustment shenanigans since I don't see a clean way to integrate this part of the cursor manipulation with the rest of the calculations that take the scrolling of the window position into account.

### DIFF
--- a/src/input_handler.cc
+++ b/src/input_handler.cc
@@ -1815,7 +1815,30 @@ void scroll_window(Context& context, LineCount offset, bool mouse_dragging)
     DisplayCoord win_pos = window.position();
     DisplayCoord win_dim = window.dimensions();
 
-    if ((offset < 0 and win_pos.line == 0) or (offset > 0 and win_pos.line == line_count - 1))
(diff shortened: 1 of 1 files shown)


lenormf:
I don't think this change makes sense. <kbd>c-u</kbd> moves the *view* half a screen upward. The editor requires the main selection to be always visible, that's the reason why the anchor consequently moves sometimes, but not always — it's a side effect.

The proposed patch makes <kbd>c-u</kbd> manage the selection directly, effectively acting like <kbd>g</kbd><kbd>t</kbd> in cases where the above side effect would not have applied.

Maybe you could use a mapping that executes <kbd>g</kbd><kbd>t</kbd>, <kbd>v</kbd><kbd>v</kbd> instead?

PR #4077, opened 2021-02-26
add fennel filetype

andreyorst:
syntax highlighting, indentation and static words for [Fennel language](http://fennel-lang.org).

Largely based on clojure.kak/perl.kak. Seems to work well on Fennel's own codebase

### DIFF
--- a/rc/filetype/fennel.kak
+++ b/rc/filetype/fennel.kak
@@ -1,0 +1,109 @@
+# http://fennel-lang.org
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+# Detection
+# ‾‾‾‾‾‾‾‾‾
+
+hook global BufCreate .*[.]fnl %{
+    set-option buffer filetype fennel
+}
+
+# Initialization
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+hook global WinSetOption filetype=fennel %{
+    require-module fennel
+    fennel-configure-window
+}
+
+hook -group fennel-highlight global WinSetOption filetype=fennel %{
+    add-highlighter window/fennel ref fennel
+    hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/fennel }
+}
+
+provide-module fennel %{
+
+require-module lisp
+
+# Highlighters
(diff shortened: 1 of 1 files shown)


rc/filetype/fennel.kak
```diff
@@ -0,0 +1,113 @@
...
+    set-option buffer extra_word_chars '_' . / * ? + - < > ! : "'"
+    hook -once -always window WinSetOption filetype=.* %{ remove-hooks window fennel-.+ }
+}
+
+define-command -hidden fennel-trim-indent lisp-trim-indent
+
+declare-option \
+    -docstring 'regex matching the head of forms which have options *and* indented bodies' \
+    regex fennel_special_indent_forms \
+    '(?:if|when|each|for|fn|lambda|λ|partial|while|local|var|doto|let)'
```
lenormf, line 89: We try to be conservative with options, I'm not sure this regex warrants one.

The idiomatic way of overriding the behaviour of a core script is to modify it in `autoload`, so runtime configuration of the indent hook is not necessary.

andreyorst: This is how it is done in clojure.kak

lenormf: I see, it looks like that slipped through the review process in #2447.

Ignore me then, this should probably be fixed (in all files) in a subsequent PR.

rc/filetype/fennel.kak
```diff
@@ -0,0 +1,113 @@
...
+}
+
+hook -group fennel-highlight global WinSetOption filetype=fennel %{
+    add-highlighter window/fennel ref fennel
+    hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/fennel }
+}
+
+hook -group fennel-insert global BufNewFile .*[.]fnl %{
+    require-module fennel
+}
```
mawww, line 25: I dont quite understand the point of this hook, its unusual to just require a filetype module without using something from it.

andreyorst: oh, perhaps a copy mistake. Clojure.kak did something there, but I removed everything not related to fennel and forgot to clean it up

andreyorst: @mawww, should be fine now

PR #4083, opened 2021-03-03
Add `ncurses_ascii` to `ui_options`

6112:
When set to `yes` or `true`, the NCurses UI uses ASCII characters
instead of Unicode characters in the following places:

* Assistant "drawings" become ASCII art
* Info box uses `+`, `-` and `|` instead of Box-Drawing Unicode
  characters (e.g., `╭───┤foo├───╮` becomes `+---|foo|---+`)
* Menus use 3 dots (`...`) instead of the ellipsis character in
  Unicode (`…`)
* Menus use background/foreground colors to draw their scrollbar,
  instead of Box Elements Unicode characters.

This fixes an issue with certain fonts, where the Box-Drawing Unicode
characters are absent from the font, and aren't rendered with the same
width as other characters. This caused mis-alignment in the UI.

By using only ASCII characters instead, we fix issues when the
characters are missing from the font (at the cost of the UI being a
little bit uglier).

`ncurses_ascii` is disabled by default.

Screenshot with ncurses_ascii=no: https://imgur.com/a/c5dEMGP
Screenshot with ncurses_ascii=yes: https://imgur.com/a/MJEbUZH

### DIFF
--- a/doc/pages/options.asciidoc
+++ b/doc/pages/options.asciidoc
@@ -335,6 +335,10 @@ are exclusively available to built-in options.
             if *yes*, or *true* the status line will be placed
             at the top of the terminal rather than at the bottom
 
+        *ncurses_ascii*:::
+            if *yes*, or *true*, use ASCII instead of Unicode characters
+            for info boxes, assistant, etc.
+
         *ncurses_assistant*:::
             specify the nice assistant displayed in info boxes,
             can be *clippy* (the default), *cat*, *dilbert* or *none*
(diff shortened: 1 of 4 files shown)


Screwtapello:
Out of curiosity, is the terminal that's giving you trouble QTerminal, Konsole, or some other Qt/KDE-based tool?

6112:
No, this is the [SSH client Chrome application](https://chrome.google.com/webstore/detail/secure-shell-app/pnhechapfaindjhompbnflcldabbghjo)

mawww:
There are various other places where Kakoune uses non-ascii characters, and I think its okay to rely on those nowadays. Most terminal emulator have had no issue with it, I would think this is more an issue of the chrome ssh extension that should be fixed there.

mawww:
This is now out of date with the builtin_terminal_ui being merged, closing.

PR #4085, opened 2021-03-04
kitty command: use launch, simplify, unify

basbebe:
`new-window` is a legacy command (see kovidgoyal/kitty#2708) and could be replaced by `launch`.
`KITTY_WINDOW_ID` should be set for any kitty window anyways so a check is redundant.
add `--keep-focus` to repl command

### DIFF
--- a/rc/windowing/kitty.kak
+++ b/rc/windowing/kitty.kak
@@ -8,23 +8,24 @@ evaluate-commands %sh{
     [ -z "${kak_opt_windowing_modules}" ] || [ "$TERM" = "xterm-kitty" ] || echo 'fail Kitty not detected'
 }
 
-declare-option -docstring %{window type that kitty creates on new and repl calls (kitty|os)} str kitty_window_type kitty
+declare-option -docstring %{window type that kitty creates on new and repl calls (window|os-window)} str kitty_window_type window
 
 define-command kitty-terminal -params 1.. -shell-completion -docstring '
 kitty-terminal <program> [<arguments>]: create a new terminal as a kitty window
 The program passed as argument will be executed in the new terminal' \
 %{
     nop %sh{
-        local match=""
+        match=""
         if [ -n "$kak_client_env_KITTY_WINDOW_ID" ]; then
             match="--match=id:$kak_client_env_KITTY_WINDOW_ID"
         fi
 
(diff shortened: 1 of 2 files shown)


rc/windowing/kitty.kak
```diff
@@ -54,13 +46,12 @@ If no client is passed then the current one is used' \
...
-                kitty @ focus-window --no-response --match=id:$kak_client_env_KITTY_WINDOW_ID
-            else
-                kitty @ --to "$kak_client_env_KITTY_LISTEN_ON" focus-tab --no-response --match=id:$kak_client_env_KITTY_WINDOW_ID
-                kitty @ --to "$kak_client_env_KITTY_LISTEN_ON" focus-window --no-response --match=id:$kak_client_env_KITTY_WINDOW_ID
+            listen=""
+            if [ -n "$kak_client_env_KITTY_LISTEN_ON" ]; then
+                listen="--to=$kak_client_env_KITTY_LISTEN_ON"
             fi
+
+            kitty @ "$listen" focus-window --no-response --match=id:"$kak_client_env_KITTY_WINDOW_ID"
```
mawww, line 54: I think `"$listen"` should not be quoted here, as we want an empty listen not-to be passed as an empty parameter

rc/windowing/kitty.kak
```diff
@@ -8,23 +8,19 @@ evaluate-commands %sh{
...
-declare-option -docstring %{window type that kitty creates on new and repl calls (kitty|os)} str kitty_window_type kitty
+declare-option -docstring %{window type that kitty creates on new and repl calls (window|os-window)} str kitty_window_type window
 
 define-command kitty-terminal -params 1.. -shell-completion -docstring '
 kitty-terminal <program> [<arguments>]: create a new terminal as a kitty window
 The program passed as argument will be executed in the new terminal' \
 %{
     nop %sh{
-        local match=""
-        if [ -n "$kak_client_env_KITTY_WINDOW_ID" ]; then
```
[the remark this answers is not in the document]
basbebe: yes, you're right. I'll readd the check.

rc/windowing/kitty.kak
basbebe: yes!

PR #4092, opened 2021-03-08
Performance improvements for update-option and set-option

### LINKED ISSUE #4084: O(N^2) case for updating ranges
This can be triggered by kak-ansi.  `ansi-render` filters out all ANSI escapes, which ends up being a large number of changes, then updates a ranges option with a close-to-equal number of ranges, using the prior timestamp.

When the option is updated, every range is iteratively updated with every "relevant" change, resulting in (I think) N^2/2 iterations.  Long runtimes are reported in eraserhd/kak-ansi#6, sometimes six or eleven minutes.

## Solution

We are clearly duplicating work.  If we know the ranges are sorted and the changes are sorted, we could solve this in O(N) with no new memory, though I don't think we can fit the algorithm to the ForwardChangesTracker interface.

If one or the other isn't sorted, then I think we need O(N) memory for cached intermediate adjustments.

1. Sort changes O(N log N)
2. Allocate N buckets for total insertions before change N
3. Allocate N buckets for total deletions before change N
4. Populate 2 and 3 in O(N)
5. Iterate through ranges, binary search for the last relevant change for that range, then update it with the buckets O(N log N).

Back-of-the-envelope math says the [profile given to me](https://gist.github.com/malte-v/574e9fcc55808a9f768ed36201493be5) by @malte-v will run in 177 seconds instead of 2134 seconds.

[Here](https://github.com/mawww/kakoune/blob/master/src/changes.hh#L46-L58) is the code in question.

EDIT: Fix iteration count.

eraserhd:
Closes #4084

The use of `std::inplace_merge()` improves the performance of `set-option -add` much more than expected, given that it is theoretically linear, so I suspect the actual implementation binary searches for the insertion place, or checks if the lists are already sorted, or something?

Before this, `set-option` grew to 341548 us per invocation.  After, it stayed under 4000 us per invocation.

Not happy with the abstractions or the duplication of the comparator from the option_list_postprocess(), but I already spent a day on this instead of my Real Job. :D  Perhaps we need some kind of OptionTraits<>?  Let me know and I'll fix.

### DIFF
--- a/src/changes.hh
+++ b/src/changes.hh
@@ -53,8 +53,10 @@ void update_forward(ConstArrayView<Buffer::Change> changes, RangeContainer& rang
             changes_tracker.update(*it++);
     };
 
-    for (auto& range : ranges)
-        update_range(changes_tracker, range, advance_while_relevant);
+    auto range_it = std::lower_bound(ranges.begin(), ranges.end(), changes.front(),
+                                     [](auto& range, const Buffer::Change& change) { return get_last(range) < change.begin; });
+    for (auto end = ranges.end(); range_it != end; ++range_it)
+        update_range(changes_tracker, *range_it, advance_while_relevant);
 }
 
 template<typename RangeContainer>
--- a/src/highlighters.cc
+++ b/src/highlighters.cc
@@ -1523,19 +1523,34 @@ private:
 BufferCoord& get_first(RangeAndString& r) { return std::get<0>(r).first; }
 BufferCoord& get_last(RangeAndString& r) { return std::get<0>(r).last; }
 
+bool option_element_compare(RangeAndString const& lhs, RangeAndString const& rhs)
+{
+    return std::get<0>(lhs).first == std::get<0>(rhs).first ?
+        std::get<0>(lhs).last < std::get<0>(rhs).last
+      : std::get<0>(lhs).first < std::get<0>(rhs).first;
+}
+
 void option_list_postprocess(Vector<RangeAndString, MemoryDomain::Options>& opt)
 {
-    std::sort(opt.begin(), opt.end(),
-              [](auto& lhs, auto& rhs) {
-        return std::get<0>(lhs).first == std::get<0>(rhs).first ?
-            std::get<0>(lhs).last < std::get<0>(rhs).last
-          : std::get<0>(lhs).first < std::get<0>(rhs).first;
-    });
+    std::sort(opt.begin(), opt.end(), option_element_compare);
 }
 
 void option_update(RangeAndStringList& opt, const Context& context)
 {
     update_ranges(context.buffer(), opt.prefix, opt.list);
+}
+
+bool option_add_from_strings(Vector<RangeAndString, MemoryDomain::Options>& opt, ConstArrayView<String> strs)
+{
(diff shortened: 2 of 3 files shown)


src/option_types.cc
```diff
@@ -21,4 +21,21 @@ UnitTest test_option_parsing{[]{
     check(DebugFlags::Keys | DebugFlags::Hooks, {"hooks|keys"});
 }};
 
+bool option_add_from_strings(Vector<RangeAndString, MemoryDomain::Options>& opt, ConstArrayView<String> strs)
+{
+    auto vec = option_from_strings(Meta::Type<Vector<RangeAndString, MemoryDomain::Options>>{}, strs);
+    size_t original_size = opt.size();
+    opt.insert(opt.end(),
```
mawww, line 28: This returns an iterator to the first inserted element, so no need for original_size

src/option_types.hh
```diff
@@ -105,6 +106,8 @@ Vector<T, domain> option_from_strings(Meta::Type<Vector<T, domain>>, ConstArrayV
     return res;
 }
 
+bool option_add_from_strings(Vector<RangeAndString, MemoryDomain::Options>& opt, ConstArrayView<String> strs);
```
mawww, line 109: I think this should be in highlighters.hh next to the other RangeAndString methods.

src/option_types.cc
```diff
@@ -21,4 +21,21 @@ UnitTest test_option_parsing{[]{
     check(DebugFlags::Keys | DebugFlags::Hooks, {"hooks|keys"});
 }};
 
+bool option_add_from_strings(Vector<RangeAndString, MemoryDomain::Options>& opt, ConstArrayView<String> strs)
```
mawww, line 24: This should be in highlighters.cc

src/option_types.cc
```diff
@@ -21,4 +21,21 @@ UnitTest test_option_parsing{[]{
...
+               std::make_move_iterator(vec.begin()),
+               std::make_move_iterator(vec.end()));
+    auto compare = [](RangeAndString const& lhs, RangeAndString const& rhs) -> bool {
+        return std::get<0>(lhs).first == std::get<0>(rhs).first ?
+            std::get<0>(lhs).last < std::get<0>(rhs).last
+          : std::get<0>(lhs).first < std::get<0>(rhs).first;
+    };
+    std::sort(opt.begin() + original_size, opt.end(), compare);
+    std::inplace_merge(opt.begin(), opt.begin() + original_size, opt.end(), compare);
+    return not vec.empty();
```
mawww, line 38: Maybe early-out on vec.empty() ?

mawww (changes requested):

eraserhd:
This should be good.  A sample file went from 12m27s to 1m3s with non-debug build.

Notes:
* The first display update takes about 5 seconds.  I haven't looked into that.
* The most significant time per iteration now is `execute-keys '<a-:><a-;>Gh'`, used to ensure each iteration processes any partial escape sequences from the last iteration.... Although this probably doesn't actually work, and it would be a good idea to track partial state in an option anyway.

lenormf:
Will this help with the performance issues in #3438 ?

PR #4095, opened 2021-03-09
Allow profiling of commands

eraserhd:
A hacky version of this was useful for figuring out what was happening with kak-ansi, so maybe this is a good idea to make permanent?


occivink:
Can be closed as per da9a196fa0e08463265534fecd82e12dae3ce45a (typo in the issue reference)

PR #4099, opened 2021-03-11
Permit usage on wayland environments.

Eluminae:
Some wayland wm will not setup a DISPLAY environment variable. Kakoune
should allow x11 windowing toolkit for those wm also.

### DIFF
--- a/rc/windowing/detection.kak
+++ b/rc/windowing/detection.kak
@@ -23,7 +23,7 @@ declare-option -docstring \
 "Ordered list of windowing modules to try and load. An empty list disables
 both automatic module loading and environment detection, enabling complete
 manual control of the module loading." \
-str-list windowing_modules 'tmux' 'screen' 'kitty' 'iterm' 'x11'
+str-list windowing_modules 'tmux' 'screen' 'kitty' 'iterm' 'x11' 'wayland'
 
 hook -group windowing global KakBegin .* %{
 
--- a/rc/windowing/wayland.kak
+++ b/rc/windowing/wayland.kak
@@ -1,0 +1,56 @@
+# wayland
+
+provide-module wayland %{
+
+# ensure that we're running in the right environment
+evaluate-commands %sh{
+    [ -z "${kak_opt_windowing_modules}" ] || [ -n "$WAYLAND_DISPLAY" ] || echo 'fail WAYLAND_DISPLAY is not set'
+}
+
+# termcmd should be set such as the next argument is the whole
+# command line to execute
+declare-option -docstring %{shell command run to spawn a new terminal
+A shell command is appended to the one set in this option at runtime} \
+    str termcmd %sh{
+    for termcmd in 'alacritty      -e sh -c' \
+                   'kitty             sh -c' \
+                   'termite        -e      ' \
+                   'wterm          -e sh -c' \
+                   'gnome-terminal -e      ' \
+                   'xfce4-terminal -e      ' \
+                   'konsole        -e      '; do
+        terminal=${termcmd%% *}
+        if command -v $terminal >/dev/null 2>&1; then
+            printf %s\\n "$termcmd"
+            exit
+        fi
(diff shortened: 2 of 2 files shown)


alexherbo2:
Shouldn’t we have a `wayland` module instead?

[`x11-focus`][`x11.kak`] and [`x11-repl`][`repl/x11.kak`] are likely to not work.

[`x11.kak`]: https://github.com/mawww/kakoune/blob/master/rc/windowing/x11.kak
[`repl/x11.kak`]: https://github.com/mawww/kakoune/blob/master/rc/windowing/repl/x11.kak

Screwtapello:
Also, a lot of the `termcmd` candidates won't work without `$DISPLAY` set (`xterm`, for example, will almost certainly never be Wayland native), so it's reasonable to require it. A separate Wayland plugin for Wayland-native terminals would be a better solution, I think.

Eluminae:
I think you right ! I added a new wayland.kak

mawww:
Hello,

I dont think I have a copyright waiver commit from you yet, could you add that ? (see the CONTRIBUTING file)

rc/windowing/wayland.kak
```diff
@@ -0,0 +1,58 @@
...
+    }
+}
+
+define-command wayland-focus -params ..1 -client-completion -docstring '
+wayland-focus [<kakoune_client>]: focus a given client''s window
+If no client is passed, then the current client is used' \
+%{
+    evaluate-commands %sh{
+        printf "There is no way to focus another window on Wayland"
+    }
```
krobelus, line 52: Eh, no need for `evaluate-commands` here 
we probably want

```
fail "There is no way to focus another window on Wayland"
```

Eluminae:
> I dont think I have a copyright waiver commit from you yet, could you add that ? (see the CONTRIBUTING file)

done !

rc/windowing/wayland.kak
Eluminae: Done :D

rc/windowing/detection.kak
```diff
@@ -23,7 +23,7 @@ declare-option -docstring \
 "Ordered list of windowing modules to try and load. An empty list disables
 both automatic module loading and environment detection, enabling complete
 manual control of the module loading." \
-str-list windowing_modules 'tmux' 'screen' 'kitty' 'iterm' 'x11'
+str-list windowing_modules 'tmux' 'screen' 'kitty' 'iterm' 'x11' 'wayland'
```
mawww, line 26: I think we might want to move wayland in front of x11, as Xwayland will likely make loading the x11 module succeed.

Eluminae: I'm not sure. This will disallow xwayland compatible terminals completly right ?

mawww: I dont have wayland here to test (nvidia...) but I would expect the x11 module to load fine thanks to DISPLAY being set which would then prevent loading of the wayland module.

I guess this is a deeper issue that might ultimately warrant a more complex solution (for example, it seems here wayland-terminal and x11-terminal could be shared), I imagine this could be fixed later on.

Eluminae: We could still add an xwayland module to permit that. I dont see the point at the moment

alexherbo2:
@Eluminae Can we load the Wayland module before X11?

I’m using Sway with XWayland, and as `$DISPLAY` is set, the Wayland module is never loaded.

Eluminae:
@alexherbo2 If we do that, someone will come and ask to move X11 before wayland.

I think we should find a configurable way to choose between one or the other

alexherbo2:
I don’t think it will hurt to have Wayland before X11, because `$WAYLAND_DISPLAY` is unlikely to be set when running X11.

Eluminae:
No but you might want to be in sway + xwayland and use an x11 terminal by default

PR #4101, opened 2021-03-12
improve keys mapping documentation

naringas:
Update mapping.asciidoc

Add explanation about mapping some characters lifted from https://github.com/mawww/kakoune/issues/2553

### DIFF
--- a/doc/pages/mapping.asciidoc
+++ b/doc/pages/mapping.asciidoc
@@ -159,19 +159,37 @@ In normal mode:
 Shift version of those mappings exist as well
 (for example `<s-left>` maps to `H`).
 
+
 == Additional notes
 
-Some keys cannot be mapped directly but they can still mapped by `c-v` to directly insert their keycode as an argument to map.
+Some keys cannot be mapped directly but they can still mapped by either: typing `c-v` before the key to insert the keycode directly or by using an alias.
 
-  * `<c-space>`
-  * `<c-_>`
-  * `<c-/>` (`<c-7>`)
-  * `<c-^>` (`<c-6>`, `<c-~>`)
+Here some of the aliases that can be used:
 
-In order to map these keys type `<c-v>` before the key in the command. For example:
+*<c-space>*::
+  <c-2>, <c-@>, <c-`>
 
----------------------------------------
-map global normal '�' ': comment-line<ret>' -docstring "<c-/> to comment/uncomment selection"
-                   ^
-                   Ctrl+v Ctrl+/ were used here to insert this
----------------------------------------
+*<esc>*::
+  <c-[>, <c-3>, <c-{>
+
+*<c-\>*::
+  <c-4>
+
+*<c-]>*::
+  <c-5>, <c-}>
+
+*<c-^>*::
+  <c-6>, <c-s-6>, <c-~>
+
+*<c-_>*::
+  <c-7>, <c-/>
+
+*<backspace>*::
+  <c-8>, on some terminals <c-h>
+
+*<tab>*::
+  <c-i>
+
+*<ret>*::
+  <c-j>, <c-m>
+
--- a/doc/pages/mapping.asciidoc
+++ b/doc/pages/mapping.asciidoc
@@ -158,3 +158,20 @@ In normal mode:
 
 Shift version of those mappings exist as well
 (for example `<s-left>` maps to `H`).
+
+== Additional notes
+
+Some keys cannot be mapped directly but they can still mapped by `c-v` to directly insert their keycode as an argument to map.
+
+  * `<c-space>`
+  * `<c-_>`
+  * `<c-/>` (`<c-7>`)
+  * `<c-^>` (`<c-6>`, `<c-~>`)
+
+In order to map these keys type `<c-v>` before the key in the command. For example:
+
+---------------------------------------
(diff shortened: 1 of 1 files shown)


Screwtapello:
Actually, all those keys can be mapped without `<c-v>` now. However, it may be worth adding a section describing some of the redundancies commonly encountered via the terminal interface:

| Mappable key | Aliases                     |
|--------------|-----------------------------|
| `<c-space>`  | `<c-2>`, `<c-@>`, ``<c-`>`` |
| `<esc>` | `<c-[>`, `<c-3>`, `<c-{>` |
| `<c-\>` | `<c-4>`, `<c-|>` |
| `<c-]>` | `<c-5>`, `<c-}>` |
| `<c-^>` | `<c-6>`, `<c-s-6>`, `<c-~>` |
| `<c-_>` | `<c-7>`, `<c-/>` |
| `<backspace>` | `<c-8>`, on some terminals `<c-h>` |
| `<tab>` | `<c-i>` |
| `<ret>` | `<c-j>`, `<c-m>` |

naringas:
Hi, I added the updated notes based on https://github.com/mawww/kakoune/pull/4101#issuecomment-797861857

doc/pages/mapping.asciidoc
```diff
@@ -159,19 +159,37 @@ In normal mode:
 Shift version of those mappings exist as well
 (for example `<s-left>` maps to `H`).
 
+
 == Additional notes
 
-Some keys cannot be mapped directly but they can still mapped by `c-v` to directly insert their keycode as an argument to map.
+Some keys cannot be mapped directly but they can still mapped by either: typing `c-v` before the key to insert the keycode directly or by using an alias.
```
Screwtapello, line 165: You cannot map these keys via `<c-v>`, you have to use their actual names.

doc/pages/mapping.asciidoc
```diff
@@ -159,19 +159,37 @@ In normal mode:
...
+  <c-7>, <c-/>
+
+*<backspace>*::
+  <c-8>, on some terminals <c-h>
+
+*<tab>*::
+  <c-i>
+
+*<ret>*::
+  <c-j>, <c-m>
```
Screwtapello, line 194: For consistency with the rest of this document, you should use backticks around key-names.

doc/pages/mapping.asciidoc
naringas: what do you mean "You cannot map these keys via `<c-v>`..."? 

I have mapped `<c-/>` exactly as described here https://github.com/mawww/kakoune/issues/2553#issuecomment-437600134 i.e. I used `<c-v><c-/>` to input a symbol in to `map global normal...`

this worked for me in Kakoune v2020.09.01

doc/pages/mapping.asciidoc
naringas: should I enforce this across the entire document? (it is not consistent right now)

doc/pages/mapping.asciidoc
Screwtapello: In the current Git version of Kakoune, you can't map these keys with `<c-v>`, it's changed since v2020.09.01.

doc/pages/mapping.asciidoc
Screwtapello: It isn't? Well, in that case never mind. If it needs to be consistent, it can be cleaned up all at once pretty easily.

naringas: when the keys are listed as an itemization they are not surrounded by backticks. I've left this alone. I removed the comment about `<c-v>`,

mawww:
Hello,

It looks to me like those are artifacts of how some terminals encode keys, and we are getting close to having that fixed thanks to CSI u and the recent push in some terminals to add a enable/disable escape sequence for it.

I am not very keen on adding this inside the documentation as it relates more to terminal emulators than to Kakoune itself.

Screwtapello:
While these keyboard encodings are out of Kakoune's control and therefore not it's responsibility, it's more likely that a user will hit a problem with Kakoune than with other tools because Kakoune's key-name syntax is so flexible. And it's not immediately obvious where to look for such documentation: the fact that `<c-3>` is associated with the byte `\x1b` (ESC) is implemented in the X server, so it works in a terminal like `st` even though that mapping does not appear in `st`'s source code, never mind its documentation.

As terminals that support CSI u become more widespread, this behaviour will become more surprising to the people who bump into it, so if the documentation acknowledges that these encodings exist in old/weird terminals, the resulting user happiness will outweigh the documentation maintenance cost.

naringas:
> While these keyboard encodings are out of Kakoune's control and therefore not it's responsibility, it's more likely that a user will hit a problem with Kakoune than with other tools because Kakoune's key-name syntax is so flexible

This has been the case in my experience, I had a weird issue where `alt-;` wouldn't work unless I chaned my keyboard layout, but then it would break again (it would work for a few minutes after resetting it). I wound up changing to a completely different layout.

> Also, it gives us a place to say "if these restrictions annoy you, go pester your terminal emulator to support CSI u, instead of filing issues against Kakoune".

I think it would also be more helpful to also say a few words about terminal emulators and the way keyboards work within them and within Xorg. For example, I had never heard of CSI u (but I have been reading up about terminfo)

PR #4109, opened 2021-03-20
line-number: Add option to use a different separator on the current line

hristost:
I thought that the `number-lines` highlighter could use a different separator for the currently selected line, like so:
Current line indicator

This commit introduces a new optional argument to the `number-lines` highlighter:
```kak
hook global WinCreate ^[^*]+$ %{ add-highlighter window/ number-lines -hlcursor -separator ' ⎸ ' -separator-cursor '  ' }
```

### DIFF
--- a/src/highlighters.cc
+++ b/src/highlighters.cc
@@ -1107,14 +1107,13 @@ struct LineNumbersHighlighter : Highlighter
             throw runtime_error("separator length is limited to 10 bytes");
 
         Optional<StringView> separator_cursor = parser.get_switch("separator-cursor");
-        if (separator_cursor.value_or(separator).length() != separator.length()) {
+        if (separator_cursor && (*separator_cursor).length() != separator.length()) {
             // Throw runtime error instead?
-            write_to_debug_buffer("line-numbers: Separator for active line should have the same length as seperator");
+            write_to_debug_buffer("number-lines: Separator for active line should have the same length as `separator`");
             separator_cursor.reset();
         }
 
-        // Effectively separator_cursor?.str(), except idk how to code in C++
-        Optional<String> separator_cursor_str = separator_cursor ? separator_cursor->str() : Optional<String>{};
+        auto separator_cursor_str = separator_cursor.map([](auto&& t){ return t.str(); });
 
         int min_digits = parser.get_switch("min-digits").map(str_to_int).value_or(2);
         if (min_digits < 0)
--- a/src/highlighters.cc
+++ b/src/highlighters.cc
@@ -1082,7 +1082,7 @@ private:
 
 struct LineNumbersHighlighter : Highlighter
 {
-    LineNumbersHighlighter(bool relative, bool hl_cursor_line, String separator, Optional<String> separator_cursor, int min_digits)
+    LineNumbersHighlighter(bool relative, bool hl_cursor_line, String separator, String separator_cursor, int min_digits)
       : Highlighter{HighlightPass::Move},
         m_relative{relative},
         m_hl_cursor_line{hl_cursor_line},
@@ -1103,15 +1103,13 @@ struct LineNumbersHighlighter : Highlighter
         ParametersParser parser(params, param_desc);
 
         StringView separator = parser.get_switch("separator").value_or("│");
-        Optional<StringView> separator_cursor = parser.get_switch("separator-cursor");
+        StringView separator_cursor = parser.get_switch("separator-cursor").value_or(separator);
 
         if (separator.length() > 10)
             throw runtime_error("separator length is limited to 10 bytes");
 
-        if (separator_cursor && (*separator_cursor).column_length() != separator.column_length())
+        if (separator_cursor.column_length() != separator.column_length())
             throw runtime_error("separator for active line should have the same length as 'separator'");
-
-        auto separator_cursor_str = separator_cursor.map([](auto&& t){ return t.str(); });
 
         int min_digits = parser.get_switch("min-digits").map(str_to_int).value_or(2);
         if (min_digits < 0)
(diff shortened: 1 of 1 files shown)


src/highlighters.cc
```diff
@@ -1142,7 +1153,10 @@ struct LineNumbersHighlighter : Highlighter
             const auto atom_face = last_line == current_line ? face_wrapped :
                 ((m_hl_cursor_line and is_cursor_line) ? face_absolute : face);
             line.insert(line.begin(), {buffer, atom_face});
-            line.insert(line.begin() + 1, {m_separator, face});
+
+            const auto separator = is_cursor_line && m_separator_cursor ? *m_separator_cursor
```
mawww, line 1157: this should be a const-ref.

src/highlighters.cc
```diff
@@ -1104,13 +1106,22 @@ struct LineNumbersHighlighter : Highlighter
...
+        auto separator_cursor_str = separator_cursor.map([](auto&& t){ return t.str(); });
+
         int min_digits = parser.get_switch("min-digits").map(str_to_int).value_or(2);
         if (min_digits < 0)
             throw runtime_error("min digits must be positive");
         if (min_digits > 10)
             throw runtime_error("min digits is limited to 10");
 
-        return std::make_unique<LineNumbersHighlighter>((bool)parser.get_switch("relative"), (bool)parser.get_switch("hlcursor"), separator.str(), min_digits);
+        return std::make_unique<LineNumbersHighlighter>((bool)parser.get_switch("relative"), (bool)parser.get_switch("hlcursor"), separator.str(), separator_cursor_str, min_digits);
```
mawww, line 1124: move separator_cursor_str here

src/highlighters.cc
```diff
@@ -1104,13 +1106,22 @@ struct LineNumbersHighlighter : Highlighter
         if (separator.length() > 10)
             throw runtime_error("separator length is limited to 10 bytes");
 
+        Optional<StringView> separator_cursor = parser.get_switch("separator-cursor");
+        if (separator_cursor && (*separator_cursor).length() != separator.length()) {
+            // Throw runtime error instead?
```
mawww, line 1111: Yep, I think that would make sense, also you could use `separator_cursor->column_length()` as its really the display length that we want to compare, not the byte one.

mawww (changes requested):
I am leaning slightly in favor of merging this because the implementation is really simple, but it is still tingling my feature-bloat sense.

hristost:
> I am leaning slightly in favor of merging this because the implementation is really simple, but it is still tingling my feature-bloat sense.

I just made the changes you recommended. 

I was also thinking whether this is feature-bloat, especially since there are so many ways this highlighter can be expanded. But we already have the separator, and a different face for the current line, so why not have a different separator too?

src/highlighters.cc
hristost: I am not much familiar with C++. I changed the line to
```cpp
const auto& separator { /* ... */ };
```
Is this what you meant?

src/highlighters.cc
```diff
@@ -1141,8 +1150,12 @@ struct LineNumbersHighlighter : Highlighter
             snprintf(buffer, 16, format, std::abs(line_to_format));
             const auto atom_face = last_line == current_line ? face_wrapped :
                 ((m_hl_cursor_line and is_cursor_line) ? face_absolute : face);
+
+            const auto& separator {is_cursor_line && m_separator_cursor && last_line != current_line
```
mawww, line 1154: its a bit weird to use a brace base init here, at least its not very consistent with the codebase, we typically use `=` to assign references.

src/highlighters.cc
```diff
@@ -1174,6 +1187,7 @@ struct LineNumbersHighlighter : Highlighter
     const bool m_relative;
     const bool m_hl_cursor_line;
     const String m_separator;
+    const Optional<String> m_separator_cursor;
```
mawww, line 1190: Thinking a bit more about it, I think we could avoid the Optional here and just duplicate the separator is -separator-cursor was not provided. Separators are very likely to fit into the small string optimization and it would simplify the highlighting logic by not having to care about whether that optional is engaged or not.

hristost: That's how I actually implemented everything initially. I thought that it's a bit weird to have two strings that will have the same value for most use cases, but the implementation is indeed a lot simpler. Should we revert back to using just `String`?

mawww: Yeah, I think we can go with the simplest implementation, especially as the Optional does not buy us any perf/memory.

src/highlighters.cc
```diff
@@ -1082,35 +1082,42 @@ struct ShowWhitespacesHighlighter : Highlighter
...
         m_separator{std::move(separator)},
+        m_separator_cursor{std::move(separator_cursor)},
         m_min_digits{min_digits} {}
 
     static std::unique_ptr<Highlighter> create(HighlighterParameters params, Highlighter*)
     {
         static const ParameterDesc param_desc{
             { { "relative", { false, "" } },
               { "separator", { true, "" } },
+              { "separator-cursor", { true, "" } },
```
hristost, line 1098: @mawww what name is do you think appropriate for this? I was thinking that `cursor-separator` works equally well

mawww: I think both are fine, `-cursor-separator` seems more natural, but `-separator-cursor` seems more consistent with the fact that its kind-of a sub-option of `-separator`.

hristost: I changed it to `-cursor-separator`, but I'm personally fine with anything. We could also consider "active", "current", or "selected" separator.

Other than that, I consider this to be ready to merge if you decide.

hristost:
@mawww I forgot to update the documentation in `./doc/`. I just did that on the same branch (https://github.com/hristost/kakoune/tree/hristost/line-number-active-separator).

PR #4110, opened 2021-03-21
rc markdown: Fix HTML highlighting in inline code

### LINKED ISSUE #4091: Incorrect highlighting of code for Markdown
`example.md`

``` markdown
Run a command as `<command> sh -c {connect} -- [arguments]`.
```

It spans until the end of the buffer.

Ordoviz:
Fixes #4091

Because the HTML highlighter was higher up in the hierarchy than the code highlighter, it took precedence. I fixed it by making HTML an inline region. Using my new knowledge of "inline" I was able to remove one line of code.

The angle_bracket_url region cannot be made inline because otherwise HTML takes precedence and `<http://example.org>` is highlighted as a HTML tag.


alexherbo2:
@Ordoviz Thanks for fixing it

Issue #4118, opened 2021-03-24
<s-ret> not mappable (hard-mapped in code)


yerlaser:
It appears that <s-ret>, i.e. Shift+Enter is hard-mapped to insert a new line above in normal mode and so doesn't allow to remap.
It's not that the terminal doesn't identify the key - pressing just Enter does nothing in normal mode.
On my terminal the key sequence being sent is ^[OM while for plain Enter it is ^M.

yerlaser:
Actually with a bit more testing I found out that Kakoune recognizes Shift+Enter as Alt+Shift+O and that is mappable.
Still looks like a bug to me.

yerlaser:
More findings.
Alt+Shift+O corresponds to ^[O
Shift+Enter corresponds to ^[OM
So, by coincidence above combinations work identically at list in normal mode, but last M gets eaten.
So, the issue seems to be more general how escape sequences with modifiers are handled by Kakoune.
Feel free to close this as not-a-bug if you feel that it's expected.

Screwtapello:
What terminal are you using that produces `^[OM` for Shift+Enter? gnome-terminal and urxvt seem to send plain `^M` for Enter, Shift+Enter, and Ctrl+Enter. xterm does too, unless the `modifyOtherKeys` option is set, in which case it sends `^[[13;2u` which Kakoune understands as `<s-ret>`.

yerlaser:
It's KDE Plasma Konsole. I was thinking it's the default.
Fish, e.g., identifies it as 'enter' key.
But, of course, there's no correct answer.

I just tried your suggestion with ^[[13;2u
Yes, that works with Kakoune. Messes up with other CLI tools, but I think I can redo them.

Screwtapello:
Apparently Konsole has sent `^[OM` for Shift+Enter [since 2001](https://github.com/KDE/konsole/commit/d65cb69a55f96e45c639dde46e2b222894f26025), but it's not clear to me why.

According to [the xterm docs](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h2-PC-Style-Function-Keys), `^[OM` (written as "SS3 M" there) is the tradition code for the *numeric keypad* Enter key, but only in "application keypad" mode.

[...]

PR #4119, opened 2021-03-26
Support keypad keys

Screwtapello:
Kakoune now knows about all the keypad keys listed in:

https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h2-PC-Style-Function-Keys
https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h2-VT220-Style-Function-Keys

The VT220-style encodings are used to for modified numeric keys when NumLock is off. For example, consider the 8/Up key:

| Modifiers       | Sequence    | Notes             |
|-----------------|-------------|-------------------|
| Unmodified      | CSI A       | Ordinary up arrow |
| Shift           | SS3 2 x     | Shift-8           |
| NumLock         | 8           | Ordinary 8        |
| Shift + NumLock | CSI 1 ; 2 A | Shift-Up          |

Note that even though the terminal distinguishes between keypad and regular keys, Kakoune maps keypad keys onto regular keys - keypad Enter is still <ret>, it just supports more modifiers than the regular Enter key.

Since I worked from the xterm docs, unsurprisingly this works best in xterm.

gnome-terminal never sends modified SS3 sequences when NumLock is on: it only ever sends plain ASCII, CSI sequences (with modifiers where appropriate),  or nothing at all. When NumLock is off, it sends SS3 sequences when the `+`, `-`, `*`, `/` or Enter keys are pressed with a modifier, otherwise it's still plain ASCII, CSI sequences (with modifiers) or nothing.

st sends SS3 x instead of CSI A, so you can't type `<up>` on the keypad at all. It also doesn't support modifiers, except that adding Shift sends `<8>`, adding Shift+Alt sends `<a-8>`, and adding Ctrl+Shift sends `<backspace>`.

Konsole does not seem to send SS3 sequences at all, with the exception that Shift+Enter (the regular Enter, not the keypad Enter) send SS3 M which is supposed to mean unmodified keypad Enter (see #4118). This patch makes Shift+Enter indistinguishable from Enter in Konsole, but previously it opened a new line above the cursor and typed M, which wasn't very useful behaviour either.

With putty's default settings, it only ever sends SS3 sequences, regardless of Shift or NumLock. The only modifier supported is Alt (by Esc-prefixing), except that Numpad Plus sends SS3 l (the code for Numpad Comma) while Shift+Numpad Plus sends SS3 m (the code for Numpad Minus). Numlock, Numpad Slash, Numpad Star and Numpad Minus send SS3 P, SS3 Q, SS3 R, and SS3 S respectively, the codes for F1-F4. It kind of makes sense, that's where those keys are placed on a VT100. You can configure Putty to use "Xterm R6" encoding for the keypad, in which case it behaves a little more like xterm, but doesn't support all the modifiers.

rxvt has an entirely different encoding, as usual, and I can't quickly find documentation about it, so I haven't implemented it.

### DIFF
--- a/src/ncurses_ui.cc
+++ b/src/ncurses_ui.cc
@@ -726,6 +726,7 @@ Optional<Key> NCursesUI::get_next_key()
         case 'B': return masked_key(Key::Down);
         case 'C': return masked_key(Key::Right);
         case 'D': return masked_key(Key::Left);
+        case 'E': return masked_key('5');        // Numeric keypad 5
         case 'F': return masked_key(Key::End);   // PC/xterm style
         case 'H': return masked_key(Key::Home);  // PC/xterm style
         case 'P': return masked_key(Key::F1);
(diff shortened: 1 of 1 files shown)


src/ncurses_ui.cc
```diff
@@ -792,19 +793,46 @@ Optional<Key> NCursesUI::get_next_key()
...
+        case 'p': return Key{parse_mask(raw_mask), '0'};
+        case 'q': return Key{parse_mask(raw_mask), '1'};
+        case 'r': return Key{parse_mask(raw_mask), '2'};
+        case 's': return Key{parse_mask(raw_mask), '3'};
+        case 't': return Key{parse_mask(raw_mask), '4'};
+        case 'u': return Key{parse_mask(raw_mask), '5'};
+        case 'v': return Key{parse_mask(raw_mask), '6'};
+        case 'w': return Key{parse_mask(raw_mask), '7'};
+        case 'x': return Key{parse_mask(raw_mask), '8'};
+        case 'y': return Key{parse_mask(raw_mask), '9'};
```
craigbarnes, line 835: For the range `'j' ... 'y'` you can just use the value `code - 64` instead of all these separate statements. Same for `'I'`.

I wonder if the compiler figures out that this can be optimized? Probably not...

craigbarnes: Hmm, it seems like GCC and Clang are both smart enough to see it: https://godbolt.org/z/s8h1xEMhf

PR #4131, opened 2021-04-03
Improved links in man pages

greenfierydragon:
# Problem

Sometimes in man pages links to other man pages would get wrapped over multiple lines. Previously, only the second part of the link *(the section with the brackets)* would be highlighted, and upon pressing `<ret>` this part would not be included and you would get an error about invalid man pages.

# Solution

I modified the regex so that it would recognise these multiline links. I also changed the `man_jump` function so that these multiline links will be recognised, and therefore you can quickly browse the man pages. An example of the changes to the regex is below:
 
image

The links now work and upon pressing `ret` it will take you to the respective man page.

### DIFF
--- a/rc/tools/man.kak
+++ b/rc/tools/man.kak
@@ -12,7 +12,7 @@ hook -group man-highlight global WinSetOption filetype=man %{
     # Command line options
     add-highlighter window/man-highlight/ regex '^ {7}-[^\s,]+(,\s+-[^\s,]+)*' 0:list
     # References to other manpages
-    add-highlighter window/man-highlight/ regex [-a-zA-Z0-9_.]+\([a-z0-9]+\) 0:header
+    add-highlighter window/man-highlight/ regex '(\w|\d|-|_|((-|‐)\n +))+\(\d\)' 0:header
 
     map window normal <ret> ': man-jump<ret>'
 
@@ -33,18 +33,13 @@ define-command -hidden -params ..3 man-impl %{ evaluate-commands %sh{
         exit
     fi
     shift
-    manout=$(mktemp "${TMPDIR:-/tmp}"/kak-man.XXXXXX)
-    manerr=$(mktemp "${TMPDIR:-/tmp}"/kak-man.XXXXXX)
-    colout=$(mktemp "${TMPDIR:-/tmp}"/kak-man.XXXXXX)
+    manout=$(mktemp "${TMPDIR:-/tmp}"/kak-man-XXXXXX)
+    manerr=$(mktemp "${TMPDIR:-/tmp}"/kak-man-XXXXXX)
+    colout=$(mktemp "${TMPDIR:-/tmp}"/kak-man-XXXXXX)
     env MANWIDTH=${kak_window_range##* } man "$@" > "$manout" 2> "$manerr"
     retval=$?
-    if command -v col >/dev/null; then
-        col -b -x > ${colout} < ${manout}
-    else
-        sed 's/.//g' > ${colout} < ${manout}
-    fi
+    col -b -x > ${colout} < ${manout}
     rm ${manout}
-
     if [ "${retval}" -eq 0 ]; then
         printf %s\\n "
                 edit -scratch %{*$buffer_name ${*}*}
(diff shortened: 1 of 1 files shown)


rc/tools/man.kak
```diff
@@ -33,18 +33,13 @@ define-command -hidden -params ..3 man-impl %{ evaluate-commands %sh{
         exit
     fi
     shift
-    manout=$(mktemp "${TMPDIR:-/tmp}"/kak-man.XXXXXX)
-    manerr=$(mktemp "${TMPDIR:-/tmp}"/kak-man.XXXXXX)
-    colout=$(mktemp "${TMPDIR:-/tmp}"/kak-man.XXXXXX)
+    manout=$(mktemp "${TMPDIR:-/tmp}"/kak-man-XXXXXX)
+    manerr=$(mktemp "${TMPDIR:-/tmp}"/kak-man-XXXXXX)
+    colout=$(mktemp "${TMPDIR:-/tmp}"/kak-man-XXXXXX)
```
mawww, line 38: Why those changes ?

rc/tools/man.kak
```diff
@@ -33,18 +33,13 @@ define-command -hidden -params ..3 man-impl %{ evaluate-commands %sh{
...
     env MANWIDTH=${kak_window_range##* } man "$@" > "$manout" 2> "$manerr"
     retval=$?
-    if command -v col >/dev/null; then
-        col -b -x > ${colout} < ${manout}
-    else
-        sed 's/.//g' > ${colout} < ${manout}
-    fi
+    col -b -x > ${colout} < ${manout}
```
mawww, line 41: Some platforms lack the `col` command, we actually could always use sed here

rc/tools/man.kak
```diff
@@ -130,18 +125,51 @@ define-command man-link %{ evaluate-commands -save-regs / %{
 
 define-command -docstring 'Try to jump to a man page' \
 man-jump %{
-  try %{ man-link } catch %{ man-link-here } catch %{ fail 'Not a valid man page link' }
-  try %{ man } catch %{ fail 'No man page link to follow' }
+  try %{
```
mawww, line 128: I think we could just expand the man_link1 option to contain that line-break alternative, and tweak the `man` command implementation to strip them from the selection, should just be sed oneliner.

rc/tools/man.kak
```diff
@@ -130,18 +125,51 @@ define-command man-link %{ evaluate-commands -save-regs / %{
...
+  
 # Suggested keymaps for a user mode
 declare-user-mode man-mode
 
 define-command man-mode-map -params 3 %{
   map global man-mode %arg[1] %arg[2] -docstring %arg[3]
 } -hidden
 
-man-mode-map 'g' ': man-jump<ret>' 'Jump to a man page using selected man page link'
+  man-mode-map 'g' ': man-jump<ret>' 'Jump to a man page using selected man page link'
```
mawww, line 172: ?

rc/tools/man.kak
```diff
@@ -130,18 +125,51 @@ define-command man-link %{ evaluate-commands -save-regs / %{
...
+  # put you at the start of a word though but
+  # close as i can get)
+  } catch %{
+    execute-keys "u<a-:><a-;>;"
+    # Error Message
+    fail "No man page link to follow"
+  }
 }
 
+define-command -hidden man-jump-raw %{
```
mawww, line 161: ?

mawww (changes requested):
I think this is a nice fix, but I would like the implementation to be simpler.

greenfierydragon:
I agree with you the previous solution was too complex.
I have made a couple of changes:
 * Support for letters in the brackets.
 * Pipes to perl instead of trying to *'unwrap'* the newline within the kakoune.
 * Removes some of the weird changes, these happened because instead of basing my changes of the one is the kakoune repo, i based them of the one which comes in arch linux which I think has been slightly modified/was old.
 * Made it so if you press enter on something other than a link it doesn't completely freak out.

greenfierydragon:
Using sed instead of perl now, just to remove an added dependency :).

rc/tools/man.kak
mawww: Any issue with that alternaive approach ? I think it would avoid all this quite complex logic.

PR #4140, opened 2021-04-11
fix line completion with prefix

tw4452852:
There's a bug in current line completion, fix it.

Signed-off-by: Tw <tw19881113@gmail.com>


test/compose/line-completion/in
```diff
@@ -1,4 +1,4 @@
             w111111
 
     w222222
-                          
+                          w2
```
krobelus, line 4: This test case also passes without your change. I'm not sure what your change is supposed to do?
(Maybe it's better to make a separate test case because this one remains valid)

krobelus (changes requested):

test/compose/line-completion/in
tw4452852: The original test case will match both `w222222` and `w111111`, but the current line completer has the prefix filter functionality which is not covered in the original test case. So I just amend the test case to cover this.

krobelus: > BTW, if we complete the line with prefix w3, the previous implementation will also provide both w222222 and w111111 as candidates.

Sorry I can't reproduce - when I replace `w2` with `w3` I correctly get no completions on `<c-x>l` on master - can you provide a test case?

tw4452852: It's weird, I can reproduce in my side with `5696ed02`.  For the test case, you can just use this amended test case which I add `w2` as the prefix, in the `cmd` you can see if you trigger line completion, you will get two candidates which are `w222222` and `w111111` w/o this patch,  actually, I expect only `w111111` should show.

src/insert_completer.cc
```diff
@@ -348,7 +348,7 @@ InsertCompletion complete_line(const SelectionList& sels,
     const ColumnCount tabstop = options["tabstop"].get<int>();
     const ColumnCount column = get_column(buffer, tabstop, cursor_pos);
 
-    StringView prefix = trim_indent(buffer[cursor_pos.line].substr(0_byte, cursor_pos.column));
+    String prefix = trim_indent(buffer[cursor_pos.line].substr(0_byte, cursor_pos.column));
```
mawww, line 351: Those bugs are quite bad, I really wish there would be a good way in C++ to avoid that with some kind of life time annotations.

tw4452852: Same as you, I really wish C++ could trigger a warning about type mismatch in this case...

mawww: Well, the different types are fine, Strings can convert to StringView, but I wish we could have something like this [proposal](http://wg21.link/p0936r0) to specify that using that StringView past the lifetime of the original String is incorrect.

PR #4145, opened 2021-04-13
Unprefix client name

alexherbo2:
Rationale: Clients are easier to reach.


Screwtapello:
Easier to "reach"? In what way?

Without a prefix, client IDs and session IDs are both just integers. It's nice to be able to look at an ID and guess what kind of thing it identifies.

Delapouite:
Related, when the we went to `unnamed` → `client` prefix in order to be less ambiguous in the modeline: https://github.com/mawww/kakoune/commit/3afa32935f93d1b56c21e1800a5b006d31d2bc5a

alexherbo2:
@Delapouite I wonder if we can be less ambiguous without having to put `client` in the name of the client.

It is true seeing:

```
*scratch* 1:1 [scratch] 1 sel - client0@[30548] - Kakoune
```

is clearer than:

```
*scratch* 1:1 [scratch] 1 sel - 0@[30548] - Kakoune
```

krobelus:
In what situations would you type out client names (without completion available)?
I don't really see a string use case but if we do this, we can still keep the `client123@[456]` style formatting independently.

alexherbo2:
@krobelus With [kakoune.cr] I like to open files in specific clients.  For example, if I have a terminal open in a directory of another project, I can `kcr edit -s kanto -c 1 file.txt` or `kcr shell -s kanto -c 1 broot` if I need to open multiple files (usually I do that instead of `kcr edit`).

[kakoune.cr]: https://github.com/alexherbo2/kakoune.cr

mawww:
client names are still arbitrary (one can always rename them using `rename-client`) so I see only minimal value in making the default client names faster to type.

I think this change could lead to increased confusion about clients as it makes it unclear if this is some kind of index or if it is an arbitrary name. I dont think the session has the same issue because pids are rarely single digit which makes them unlikely to be mistaken for some kind of index.

krobelus:
@alexherbo2 you can add shell completion that offers completions based on `%val{client_list}`.
Or just rename your clients in a `ClientCreate` hook.

PR #4150, opened 2021-04-17
Add hash unit tests.

Icantjuddle:
Long time user wanted to start contributing so figured I could start with some low hanging fruit like this to learn the codebase.

Unit test cases are generated from https://github.com/PeterScott/murmur3

### DIFF
--- a/src/hash.cc
+++ b/src/hash.cc
@@ -55,10 +55,8 @@ size_t hash_data(const char* input, size_t len)
     uint32_t key = 0;
     switch (len & 0b11)
     {
-        case 3: key ^= tail[2] << 16;
-            [[fallthrough]];
-        case 2: key ^= tail[1] << 8;
-            [[fallthrough]];
+        case 3: key ^= tail[2] << 16; [[fallthrough]];
+        case 2: key ^= tail[1] << 8;  [[fallthrough]];
         case 1: key ^= tail[0];
                 key *= c1;
                 key = rotl(key,15);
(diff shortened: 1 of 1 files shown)


src/hash.cc
```diff
@@ -67,4 +72,16 @@ size_t hash_data(const char* input, size_t len)
     return hash;
 }
 
+UnitTest test_murmer_hash{[] {
```
mawww, line 75: murmer ?

src/hash.cc
```diff
@@ -50,10 +53,12 @@ size_t hash_data(const char* input, size_t len)
...
     uint32_t key = 0;
-    switch (len & 3)
+    switch (len & 0b11)
     {
-        case 3: key ^= tail[2] << 16; /* fallthrough */
-        case 2: key ^= tail[1] << 8;  /* fallthrough */
+        case 3: key ^= tail[2] << 16;
+            [[fallthrough]];
+        case 2: key ^= tail[1] << 8;
+            [[fallthrough]];
```
mawww, line 61: This alignment is inconsistent with how `case 1` aligns statements, which itself is inconsistent with how the code is usually formatted in Kakoune.

Could you make all cases use consistent indentation (either by moving the fallthrough attirbutes back onto the previous line, keeping all operations clearly aligned, or by introducing a line break after all `case N:` and just indenting by 4 spaces as usual.

mawww (changes requested):
Thanks for that pull request, can you fix the small style issues ?

src/hash.cc
Icantjuddle: Yeah, my b used to there just being a clang-format I can blindly apply :laughing:

PR #4152, opened 2021-04-18
[POC] Shared scopes. Initial implementation of shared scopes for hooks

topisani:
Shared scopes is an idea I saw mentioned by @mawww somewhere some time ago, and I think it is a great kakoune-esque alternative to emacs minor mode, and may even make sense for some things that would be major modes in emacs. It is also basically the concept of shared highlighters extended to hooks, options, key maps, aliases and faces.

Basic idea:
```kak
declare-shared-scope testscope
hook window ScopeLinked testscope %{
    // ...
}
hook window ScopeUnlinked testscope %{
    // ...
}

hook testscope InsertChar .* %{
    // Example hook in testscope
}

set testscope some_option true

link-scope window testscope
// Now hooks, options etc from testscope exist in window scope
unlink-scope window testscope
```

In this proof of concept i've implemented hooks from linked scopes only to get some feedback on idea and implementation before continuing.

The way i've done it here, `HookManager`, `OptionManager` etc, will all get `m_linked` vectors of other managers that they are linked to. Then `Scope::add_linked` and `remove_linked` add and remove the linked managers. This might not be the optimal way to implement it, but it seems fairly self contained this way, which is nice. I am very open to suggestions on this.

I also would like to discuss design options around the naming/linking of scopes. Currently shared scopes can be linked into any "local" scope (window, buffer, global), but I see two other ways to go from there:
 1. Only link a shared scope to a window, instead of linking it to another scope. This would probably especially simplify the `Scope[Un]Linked` hooks, but it might be very useful to link shared scopes to the global scope also. 
 2. Allow shared scopes to be linked to each other, i.e. link `manual-indent` scope into a `c++` scope. The current implementation would allow this (if the commands were modified) and should handle it fine. But use cases might need to be explored a bit further.
 
 Depending on some of these design choices some of the added functionality could be split out of `Scope` also
 
 I would like to hear everyone's feedback before I continue, on the feature itself as well as the code.


topisani:
For some further testing of my implementation, I decided to implement keymaps as well. They are implemented the exact same way as the hooks are described above.

src/scope.cc
```diff
@@ -4,6 +4,35 @@
...
 
+void Scope::add_linked(SharedScope& scope, Context& context) {
+    if (contains(m_linked, &scope)) {
+        throw runtime_error("scope already linked");
+    }
+    m_linked.emplace_back(&scope);
+    
+    m_hooks.add_linked(scope.hooks());
+    
+    m_hooks.run_hook(Hook::ScopeLinked, scope.get_name(), context);
```
topisani, line 16: Should these hooks contain information about which scope the shared scope was linked into?
That would require `get_name()` to be implemented for all scopes, but that shouldn't be a big issue I think. I am not entirely sure on the use cases here however, and would probably need some feedback on that.

src/hook_manager.hh
```diff
@@ -133,6 +140,7 @@ private:
     struct HookData;
 
     SafePtr<HookManager> m_parent;
+    Vector<SafePtr<HookManager>> m_linked;
```
topisani, line 143: I am unsure about exactly how and when `SafeCountable`/`SafePtr` should be used. Here it is just done like this to match `m_parent`

src/hook_manager.hh
```diff
@@ -125,6 +129,9 @@ public:
     CandidateList complete_hook_group(StringView prefix, ByteCount pos_in_token);
     void run_hook(Hook hook, StringView param, Context& context);
 
+    void add_linked(HookManager& hook_manager);
```
topisani, line 132: I am considering whether these should be moved out into some `LinkableManager<T>` base class, but on the other hand it is not that much code duplication to give all five manager types the `(add|remove|m)_linked` members.

mawww:
Argh, I should have comunicated a bit more about this...

I actually did a proof of concept implementation back in september 2020, and ended up unconvinced it would be a significant enough gain compared to what we already have when I started adapting filetype support scripts to it.

I must say I dont fully recall what led me to discard the idea at the time, so I am open to re-explore this.

I'll push what I had to the shared-scopes branch, it is surprisingly similar to what you came up with.

src/hook_manager.hh
topisani: Ha, I see that is exactly what you did with the `ScopeMember<T>` class in your implementation

topisani:
Wow yes, that _is_ surprisingly similar!

I do understand the reasoning in not pursuing it further initially, however, I think it becomes very valuable especially when used more for minor-mode-style features.

I see a lot of plugins with `X-enable` and `X-disable` commands that contain a lot of `map`/`unmap` etc, and I think being able to do this in a more structured way would help a lot. And importantly, here the `unmap` would actually return the key to what it was mapped to before.

Another issue that often comes up with these "minor modes" is that people are unsure which scopes to add the features to. This is what leads `kak-lsp` to have separate `lsp-enable` and `lsp-enable-window` commands for setting up the feature in global or window scope. With shared scopes this can be left fully up to the user. This is also what leads to options that should always be set in the window scope because they trigger a `WinSetOption` hook, and similar undocumented dependencies on the scope.

And finally, even in the case of features that can also be implemented currently with just `WinSetOption` hooks or `-enable`/`-disable` commands, I think just unifying shared hooks, options, aliases, highlighters and keymaps into one system instead of relying on hook groups, shared highlighters, and manual save/restore for options, keys, and aliases, is worth pursuing.

topisani:
I would also like to hear some thoughts on this from plugin authors, I hope you guys don't mind the tag: @alexherbo2 @andreyorst @occivink @robertmeta

andreyorst:
I like the idea, but most of my plugins, which added keybindings, either did it via user modes, or had a dedicated buffer for the plugin where I was free to remap everything, so I'm not the best candidate to ask :) One exception would be my [langmap.kak](https://github.com/andreyorst/langmap.kak), but it is meant to replace insert mode keys everywhere, thus not a good candidate for scopes, I guess.

Although I also use Emacs a lot and I understand the concept of minor modes good enough to say, that in non-modal editor this idea is great and I miss it a lot in other editors like VS Code or Atom, because there every keybinding basically is always present.

But in modal editor I don't know how valuable it is. Yes there are some plugins that map and unmap keys, but I think these are mostly specific mode keys (like object mode), not normal mode keys, which is already handled pretty nice. Scoped user modes seem like an interesting idea, like when you can extend object mode with some S-Expressions for LISP languages only though, and I think well matches the idea of minor modes from Emacs.

topisani:
As an example of where this would help:
https://gitlab.com/listentolist/kakoune-table/-/blob/master/table.kak#L191
`table-enable` (like `lsp-enable` in kak-lsp) enables hooks and mappings in the global scope, because that does not require the user to set up a hook to enable it in all windows of a filetype or similar. It also contains a `table-remove-mappings` function that removes these mappings again. This means there's no way to enable table mode only in a certain window, which would require a separate set of `table-enable-window` commands, like [kak-lsp has done](https://github.com/kak-lsp/kak-lsp/blob/master/rc/lsp.kak#L1309-L1415).

> Yes there are some plugins that map and unmap keys, but I think these are mostly specific mode keys (like object mode), not normal mode keys, which is already handled pretty nice.

I agree that key binds are less of an issue in a modal editor, however I think the power of shared scopes comes mainly in tying together key maps, hooks, highlighters etc in a scope, and letting the user decide when and how to include the collection of those, along with a proper way to unregister them again.

As a more subjective point, I feel like the shared scope model is a lot easier to understand, both when implementing shared scopes and when using them, compared to the current variety of solutions that exist, i.e. `X-(enable|disable)(-window)?`, `hook window WinSetOption X=.*` and the quirks that follow these solutions.

mawww:
One thing I remember not liking was that we end-up with two ways to share highlighters: through shared scopes and through shared highlighters, and we cannot really get rid of shared highlighters because we still want to have say the markdown highlighter refer to the other languages for code blocks (a shared scope would pull in all hooks for that filetype which we dont want).

I am not sure the `lsp-enable` case is such a strong argument, I think it could be easily solved by just taking the scope as an argument and using `%arg{1}`. The fact that it ends with calling a call to `lsp-did-change-config` suggests it would not go away with shared scopes which would still just be an implementation detail.

I agree the mental model of linking a scope feels higher level and cleaner.

I think we should try to convert a few of the bundled scripts to see what they would look like using scopes, I believe its doing that that led me to think it was not such a good idea after all.

PR #4166, opened 2021-04-26
improve colors for rust/solarized

dead10ck:
The colors for the Rust language seemed very different than other text editors and seemed suboptimal, so I did a bit of tinkering. It turns out there were two issues.

* The solarized color scheme has some deviations from other text editors, including the reference implementation for vim. Types were being colored red, instead of yellow; and operators were being colored yellow instead of green. This change fixes the deviations.
* The Rust syntax files:
    * Have a lot missing compared to vim's rust.vim plugin, namely operators, some key words, and a few traits and types from the Rust stdlib.
    * There were certain highlighters that were overly broad and highlighted basically all tokens. This included the `user_defined_type` highlighter, which had the effect of coloring all alphanumeric symbols that start with a capital letter as a "type", which is visually distracting. Others had similar problems, like `field_or_parameter`.

Here is a screenshot of a Rust file as highlighted by the standard files shipped with `2020.09.01`:

kakoune-rust-before

And after:

kakoune-rust-after

### DIFF
--- a/colors/solarized-dark.kak
+++ b/colors/solarized-dark.kak
@@ -21,13 +21,13 @@ evaluate-commands %sh{
    echo "
         # code
         face global value              ${cyan}
-        face global type               ${red}
+        face global type               ${yellow}
         face global variable           ${blue}
         face global module             ${cyan}
         face global function           ${blue}
         face global string             ${cyan}
         face global keyword            ${green}
-        face global operator           ${yellow}
+        face global operator           ${green}
         face global attribute          ${violet}
         face global comment            ${base01}
         face global documentation      comment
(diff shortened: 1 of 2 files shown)


mawww:
Hello,

That indeed looks better, could you add a copyright waiver commit as this looks to be your first contribution (see the CONTRIBUTING file)

Thanks,

Maxime.

dead10ck:
Okay, I figured out how to fix it: I moved the key words after the function call highlighter. I'm still not quite sure why it didn't work before, but this way also has the benefit of not being a special highlighter for let statements.

dead10ck:
Hmm, it was hard for me to figure out what the errors were from the GitHub actions because of the raw ascii codes, but after running it locally, it looks like there are regression tests for rust highlighting that this change breaks. The tests look like they are expecting reach individual token to be colored exactly how they were before. I'm not actually sure how to generate those json rpc messages; presumably this is not done by hand? Is there a better way to generate "correct" output other than looking at each test case in there and visually verifying it looks "right", and just copying the new output from the failing test case into the expected output?

Also, just as a general comment, I do like the idea of regression testing highlighters, but comparing for exact equality with specific colors seems a bit aggressive.

mawww:
Argh, I should have run the tests again before pushing, fixed now.

PR #4171, opened 2021-04-27
Improve prompt handling.

markus-oberhumer:
As a long time vi user I find it highly irritating that
you cannot backspace out of the command prompt.

Similar changes should be made to other prompts such
as the search prompt.

### DIFF
--- a/src/input_handler.cc
+++ b/src/input_handler.cc
@@ -807,7 +807,8 @@ public:
             m_callback(line, PromptEvent::Validate, context());
             return;
         }
-        else if (key == Key::Escape or key == ctrl('c'))
+        else if (key == Key::Escape or key == ctrl('c') or
+                 ((key == Key::Backspace or key == ctrl('h')) and line.empty()))
         {
             history_push(line);
             context().print_status(DisplayLine{});


src/input_handler.cc
```diff
@@ -810,6 +810,7 @@ class Prompt : public InputMode
         else if (key == Key::Escape or key == ctrl('c'))
         {
             history_push(line);
+quit:
```
krobelus, line 813: The behavior seems reasonable. (The implementation could be done by simply extending the condition above since `history_push("")` is a nop.)

src/input_handler.cc
```diff
@@ -821,6 +822,10 @@ class Prompt : public InputMode
             m_callback(line, PromptEvent::Abort, context());
             return;
         }
+        else if ((key == Key::Backspace or key == ctrl('h')) and line.empty())
```
mawww, line 825: Merging this into line 810 condition should do the trick, `goto` is quite tricky in C++ because it cannot span any object lifetime starting/finishing, which adds constraints on how th code can be structured. As @krobelus said history_push will check that the line is empty (we dont want to add empty lines even if we press escape) so the behaviour should be fine.

markus-oberhumer:
I've pushed a new version now ready for merge, if this feature is accepted.

Issue #1753, opened 2017-12-08
Suggestion: Store ncurses_assistant(s) in text files?


whereswaldon:
Today I discovered the ncurses_assistant ui_option, and I love it. I've been using kakoune for almost a year now, but I didn't realize that there were options beyond clippy.

I was wondering what @mawww  and others would think about storing the assistants in text files somewhere in the existing kakoune file hierarchy. Maybe in `/usr/share/kak/assistants` or something like that. This would make it easy to create new ones without bloating this repo. I personally might import a bunch of asciiart and then randomly select one whenever kakoune started, but perhaps this would only be amusing to me.

Clearly this isn't an important feature. I'd be happy to try to implement it myself if there's interest, but I don't want to start down that road if nobody wants it.

lenormf:
This was considered when I added the Dilbert assistant, but it's just too much work for a simple ASCII art.

danr:
This spurred a bit of conversation on #kakoune@irc.freenode.net, pasting it here for those who don't follow that channel:

    10:47 < danr> lol @ the two issues related to the ncurses_assistant
    10:47 < alex``> I noticed another bug, is that file-completion don’t work with `:` characters
    10:47 < danr> maybe I should start using clippy again
    10:49 < ironzorg> mawww   I'd say from now on you have to be a long time contributor for your assistant pull request to be considered :D
    10:49 < occivink> clippy draws too much attention when you want to show something to somebody
    10:49 < danr> occivink: agreed
    10:50 < occivink> the fairy IS pretty cool though
    10:51 * alex`` needs a Kirby assistant
    10:52 < danr> kirby is cute :)
    10:52 < alex``> If someone knows how to draw it
    10:52 < Franciman> Hi
    10:52 < Franciman> alex``, https://www.facebook.com/unk.org/posts/10153778718727072
    10:52 < Franciman> here you go
    10:52 < alex``> <('-'<)
    10:52 < occivink> that's the bad kind of ascii art though
    10:52 < alex``> (>'-')> !
    10:52 < occivink> I had found a website with thousands of nice art recently
    10:53 < danr> the linux tux is a reasonable assistant too I guess
    10:53 < Franciman> https://www.gamefaqs.com/boards/949625-kirby-super-star-ultra/47058915 this one i cute
    10:53 < alex``> Wow
    10:53 < alex``> This Kirby lol
    10:53 < occivink> ah there it is
[...]

PR #4180, opened 2021-05-01
Add the duck assistant discussed on the discourse

CharlesGueunet:
Add the duck assistant discussed [here](https://discuss.kakoune.com/t/future-assistants/66/3)

result:

duck

Best,
C.

### DIFF
--- a/doc/pages/options.asciidoc
+++ b/doc/pages/options.asciidoc
@@ -337,7 +337,7 @@ are exclusively available to built-in options.
 
         *ncurses_assistant*:::
             specify the nice assistant displayed in info boxes,
-            can be *clippy* (the default), *cat*, *dilbert* or *none*
+            can be *clippy* (the default), *cat*, *dilbert*, *duck* or *none*
 
         *ncurses_enable_mouse*:::
             boolean option that enables mouse support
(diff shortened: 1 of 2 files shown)


purxiz:
I'm not sure if this is the right place for this dicussion, but would it make more sense to be able to set the assistants dynamically? I.e. in kakrc somehow?

CharlesGueunet:
I agree with the idea.  
Two remarks:
* I think this would requires its own MR
* I think some assistants should still be available in the source code for performance reasons.

somasis (changes requested):
`doc/pages/options.asciidoc` ought to have this assistant added to the list in this commit as well.

CharlesGueunet:
@somasis you are right, done

CharlesGueunet:
@mawww if you have the time for a quick review :) this work should be ready

lenormf:
From what I can tell, this assistant comes from [this page](https://www.asciiart.eu/animals/birds-water), and the site [requires attribution](https://web.archive.org/web/20181216203205/http://asciiartist.com/wp/respect-ascii-artists-campaign) if any ASCII art is to be used.

That means the PR cannot be merged into an UNLICENSE’d project, unless you draw the assistant yourself and place it under the UNLICENSE yourself. But then having [more than three assistants is considered overkill](https://github.com/mawww/kakoune/issues/1753#issuecomment-350225838).

CharlesGueunet:
@lenormf you are completely right about the origin of this art, I found it on the discourse and did not look further. I close the PR as the licenses are not compatible.

Issue #1192, opened 2017-02-05
Improve show_matching highlighter


antoyo:
Hello.

In vim, while in insert mode, the parenthesis matched will be the one **before** the cursor.
It would be nice that kakoune behaves the same because this make it easy to see if I need to type another parenthesis or not.
Because, currently, when adding characters at the end of the line, for instance, no matching parenthesis will be shown.

By the way, is there any way to enable this by default in every buffer?
Also, is there a way to make the highlighting more apparent? Like having another background color.

I tried the following, but it does not work:

```
hook global BufSetOption add-highlighter show_matching
```

Thanks to fix this issue.

ekie:
hello,
 
to enable the show_matching highlighter by default add the following snippet to your kakrc:

```
hook global WinCreate .* %{
    addhl show_matching
}
```
To change the face of the show_matching highlighter, and let it pop out visually, you can modify the MatchingChar face. You can read up on faces in the readme or use `:doc faces` from within kakoune. For example you could add `set-face MatchingChar red,white+b` to the previous mentioned hook.

antoyo:
@ekie Thanks. The only thing that I miss now is the behaviour in insert mode.

antoyo:
@lenormf Thanks for the explanation.

I think I don't quite understand kakoune, currently (I'm a new user).
I like the idea of using selections to have an immediate visual feedback of what you're doing, but I feel like some tasks are slower to do with kakoune than with vim.

This issue is one example.
Another example is that we don't have (and it does not seem ideomatic to have) a `<c-w>` binding to delete the previous word (I know I could add this binding, but it seems not to be the kakoune way).
In both of these examples, the idiomatic way (as I understand) is to go back to normal mode.
But I feel it is slower that what I'd do with vim.
I can understand that editing the code (instead of inserting new code) is powerful in normal mode (hence why `<c-w>` might not be needed), but one process that I do a lot is using a command from the history (in command mode), changing the last word and running it again.
In this case, a shortcut like `<c-w>` is really useful.
Once again, I know that I could add such a shortcut, but I want to understand how to use kakoune in the most efficient/idiomatic way.

As a new kakoune user, I am at a loss.
Could you give me some explanations?

Thanks.

lenormf:
From `:doc faq`

>   Why aren’t widely known command line shortcuts such as <c-w> or <c-u> available in Kakoune ?
[...]

PR #4187, opened 2021-05-04
Show matching offset

purxiz:
This PR adds an offset parameter to the show-matching highlighter. This is similar to the request in #1192, which for some reason github won't let me reference ([direct link](https://github.com/mawww/kakoune/issues/1192)). 

In short, the offset parameter makes the show-matching check the character at `cursor_position - offset` within a line, instead of only allowing checking directly under the cursor. This functionality would likely most commonly be used with `-offset 1` to show the matching char of the character before the cursor. This makes it much easier to see (while in insert mode) if you need to continue adding closing brackets or parens to finish the line correctly. The original issue was rejected, seemingly due to conflicts w/ the m/M operators. I thought that exposing it as an option rather than a default would help alleviate that confusion. This is my first commit to the project, and my first time working with c++ in a long time, so let me know if I need to change anything.

Thanks!

### DIFF
--- a/src/highlighters.cc
+++ b/src/highlighters.cc
@@ -998,6 +998,89 @@ struct TabulationHighlighter : Highlighter
     }
 };
 
+struct ShowMatchingHighlighter : Highlighter
+{
+    ShowMatchingHighlighter(int offset)
+      : Highlighter{HighlightPass::Move}, m_offset{std::move(offset)}
+      {}
+
+    static std::unique_ptr<Highlighter> create(HighlighterParameters params, Highlighter*)
+    {
+        static const ParameterDesc param_desc{
+            { { "offset", { true, "" } } },
+            ParameterDesc::Flags::None, 0, 0
+        };
+        ParametersParser parser(params, param_desc);    
+
+        int offset = parser.get_switch("offset").map(str_to_int).value_or(0);
+
+        return std::make_unique<ShowMatchingHighlighter>(
+            offset);
+    }
+private:
+        void do_highlight(HighlightContext context, DisplayBuffer& display_buffer, BufferRange) override
+        {
+            const Face face = context.context.faces()["MatchingChar"];
+            const auto& matching_pairs = context.context.options()["matching_pairs"].get<Vector<Codepoint, MemoryDomain::Options>>();
+            const auto range = display_buffer.range();
+            const auto& buffer = context.context.buffer();
+            for (auto& sel : context.context.selections())
+            {
+                auto pos = sel.cursor();
+                pos.column = pos.column-m_offset;
+                if (pos < range.begin or pos >= range.end)
+                    continue;
+
+                Utf8Iterator it{buffer.iterator_at(pos), buffer};
+                auto match = find(matching_pairs, *it);
+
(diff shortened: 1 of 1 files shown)


src/highlighters.cc
```diff
@@ -998,6 +998,89 @@ struct TabulationHighlighter : Highlighter
...
+        void do_highlight(HighlightContext context, DisplayBuffer& display_buffer, BufferRange) override
+        {
+            const Face face = context.context.faces()["MatchingChar"];
+            const auto& matching_pairs = context.context.options()["matching_pairs"].get<Vector<Codepoint, MemoryDomain::Options>>();
+            const auto range = display_buffer.range();
+            const auto& buffer = context.context.buffer();
+            for (auto& sel : context.context.selections())
+            {
+                auto pos = sel.cursor();
+                pos.column = pos.column-m_offset;
```
mawww, line 1030: This can get to negative columns which would have issues. It also wont work once you migrate to a CharCount for offset as we dont know how many bytes back we have to go.

src/highlighters.cc
```diff
@@ -998,6 +998,89 @@ struct TabulationHighlighter : Highlighter
...
+        static const ParameterDesc param_desc{
+            { { "offset", { true, "" } } },
+            ParameterDesc::Flags::None, 0, 0
+        };
+        ParametersParser parser(params, param_desc);    
+
+        int offset = parser.get_switch("offset").map(str_to_int).value_or(0);
+
+        return std::make_unique<ShowMatchingHighlighter>(
+            offset);
```
mawww, line 1018: No need to break that line here, its not that long.

src/highlighters.cc
```diff
@@ -998,6 +998,89 @@ struct TabulationHighlighter : Highlighter
     }
 };
 
+struct ShowMatchingHighlighter : Highlighter
+{
+    ShowMatchingHighlighter(int offset)
+      : Highlighter{HighlightPass::Move}, m_offset{std::move(offset)}
```
mawww, line 1004: std::move on an int is unecessary and looks a bit weird. Also I think the offset should not be an int but a CharCount: we want to look at the previous codepoint.

mawww, line 1004: HighlightPass::Move looks incorrect here, this is purely colorization

src/highlighters.cc
```diff
@@ -998,6 +998,89 @@ struct TabulationHighlighter : Highlighter
...
+            const auto range = display_buffer.range();
+            const auto& buffer = context.context.buffer();
+            for (auto& sel : context.context.selections())
+            {
+                auto pos = sel.cursor();
+                pos.column = pos.column-m_offset;
+                if (pos < range.begin or pos >= range.end)
+                    continue;
+
+                Utf8Iterator it{buffer.iterator_at(pos), buffer};
```
mawww, line 1034: I think its on this iterator that the offset should apply.

mawww (changes requested):

mawww:
I think this is a reasonable approach to that issue with a pretty small change to the existing code.

purxiz:
Sorry I never came back to this, seemingly it's been implemented by the -previous flag though!

Issue #4127, opened 2021-04-01
Native support for coprocesses/daemons/long-running helpers


Screwtapello:
### Feature 

Kakoune should natively support managing coprocesses, long-running external processes that communicate bi-directionally with Kakoune. This would involve:

- A way to tell Kakoune to launch a shell as a coprocess, holding onto its stdin, stdout, and stderr.
- A way to send a string to the coprocess' stdin
- Routing the coprocess' stderr to `*debug*` like any other external process
- A way for the coprocess to send commands back to Kakoune (such as writing to stdout)
- A way to kill a misbehaving/stuck coprocess

### Rationale

Kakoune's `%sh{}` expansions provide excellent support for interacting with short-running external processes. It's more difficult, but still reasonable, to interact with long-running one-shot processes (for example, the way [grep.kak](https://github.com/mawww/kakoune/blob/212242cc602d386fe1b7c1ce10c55ad0430c4960/rc/tools/grep.kak#L17) launches a grep process in the background, plumbed to a FIFO buffer, or [lint.kak](https://github.com/mawww/kakoune/blob/212242cc602d386fe1b7c1ce10c55ad0430c4960/rc/tools/lint.kak) launches a subshell in the background that pipes to `kak -p` when its work is done).

However, every plugin that wants to interact with a long-running external process (a "coprocess") has to solve the same problems:

- launching the external process, and shutting it down when the Kakoune session ends
- de-multiplexing individual messages from Kakoune into per-session event streams
- communicating events back to Kakoune
- finding a way to report errors to the user

Not only is this effort duplicated in every plugin that wants a coprocess, but it's also a substantial increase in complexity over just running a process in the background:

- launching a process from Kakoune is easy, but shutting it down requires a timeout (short timeouts make the coprocess lose state easily, long timeouts risk having a heavy coprocess hang around long after the session exits)
- De-multiplexing messages requires using a FIFO for rendezvous (tricky, since FIFOs are supposed to be one-shot) or a socket (you have to write a client as well as a server, since POSIX provides no command-line socket-connection tool)
- Communicating events back to Kakoune is straightforward enough with `kak -p`, but expensive
- You can write errors to a log-file easily enough, but who's going to rotate it? How do you keep errors for different sessions separate? What if the error is "cannot open log file for writing"?

If these things were provided by Kakoune, it would make writing coprocess-based plugins much, much easier.

### Concrete use-cases

[...]

PR #4200, opened 2021-05-16
Add %val{bufstr} with buffer content as a string

danr:
Add the expansion `%val{bufstr}` with buffer content as a string. Having this avoids having to write the roundabouts using `|` or registers, such as:

    eval -save-regs b %{
        exec -draft '%"by'
        eval %sh{
             # something with $kak_reg_b
        }
    }

This is now simply:

	eval %sh{
         # something with $kak_bufstr
	}

For large buffers that are too big to send as an execve environment this `%val{bufstr}` expansion is still useful together with `echo -to-file`.

Included as an example how `spell.kak` can be simplified using this expansion.

### DIFF
--- a/doc/pages/expansions.asciidoc
+++ b/doc/pages/expansions.asciidoc
@@ -204,6 +204,10 @@ The following expansions are supported (with required context _in italics_):
     _in buffer, window scope_ +
     name of the current buffer
 
+*%val{bufstr}*::
+    _in buffer, window scope_ +
+    the contents of the entire buffer, as a string
+
 *%val{client_env_X}*::
     _in window scope_ +
     value of the `$X` environment variable in the client displaying the current
--- a/rc/tools/spell.kak
+++ b/rc/tools/spell.kak
@@ -1,6 +1,5 @@
 declare-option -hidden range-specs spell_regions
 declare-option -hidden str spell_last_lang
-declare-option -hidden str spell_tmp_file
 
 declare-option -docstring "default language to use when none is passed to the spell-check command" str spell_lang
 
@@ -14,15 +13,9 @@ define-command -params ..1 -docstring %{
     } spell %{
     try %{ add-highlighter window/ ranges 'spell_regions' }
     evaluate-commands %sh{
-        file=$(mktemp -d "${TMPDIR:-/tmp}"/kak-spell.XXXXXXXX)/buffer
-        printf 'eval -no-hooks write -sync %s\n' "${file}"
-        printf 'set-option buffer spell_tmp_file %s\n' "${file}"
-    }
-    evaluate-commands %sh{
         use_lang() {
             if ! printf %s "$1" | grep -qE '^[a-z]{2,3}([_-][A-Z]{2})?$'; then
                 echo "fail 'Invalid language code (examples of expected format: en, en_US, en-US)'"
-                rm -rf "$(dirname "$kak_opt_spell_tmp_file")"
                 exit 1
             else
                 options="-l '$1'"
@@ -37,7 +30,7 @@ define-command -params ..1 -docstring %{
         fi
 
         {
-            sed 's/^/^/' "$kak_opt_spell_tmp_file" | eval "aspell --byte-offsets -a $options" 2>&1 | awk '
+            printf '%s' "$kak_bufstr" | sed 's/^/^/' | eval "aspell --byte-offsets -a $options" 2>&1 | awk '
                 BEGIN {
                     line_num = 1
                     regions = ENVIRON["kak_timestamp"]
@@ -83,7 +76,6 @@ define-command -params ..1 -docstring %{
                     close(server_command)
                 }
             '
-            rm -rf $(dirname "$kak_opt_spell_tmp_file")
         } </dev/null >/dev/null 2>&1 &
     }
 }
--- a/src/main.cc
+++ b/src/main.cc
@@ -190,6 +190,10 @@ static const EnvVarDesc builtin_env_vars[] = { {
         "buffile", false,
         [](StringView name, const Context& context) -> Vector<String>
         { return {context.buffer().name()}; }
+    },  {
+        "bufstr", false,
+        [](StringView name, const Context& context) -> Vector<String>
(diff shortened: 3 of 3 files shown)


doc/pages/expansions.asciidoc
```diff
@@ -200,9 +200,14 @@ The following expansions are supported (with required context _in italics_):
     quoted list of the names of currently-open buffers (as seen in
     `%val{bufname}`)
 
-*%val{bufname}*::
+*%val{buffile}*::
+    _in buffer, window scope_ +
+    full path of the file or same as `%val{bufname}` when there’s no
+    associated file
```
Screwtapello, line 206: I assume you didn't intend to replace the documentation for `bufname` with a second copy of the `buffile` description?

danr: hehe nope, will fix right away

danr: fixed now, thanks

mawww:
I must say I am not very keen to add another way to get the buffer content (we already can get it in the env via `%` and `$kak_selection`) and I am far from convinced with the `:echo -to-file` argument that sounds like a weird way to `:write`.

The PR did however trigger my curiosity (kudos on including the `spell.kak` example as a motivating use case) and that led me to do 6e15a86737acfad2124820eb4b073d6d488f0b7c on the command-fd branch. That approach is more general does not have issue with the environment size limit (although its arguably more complex to wrap one's head around).

I guess my point here is that I'd be more inclined to go with command-fd than to add `%val{bufstr}`, but command-fd has been there as a POC for a while, and I did not convince myself (yet) it is worth it.

occivink:
Could you explain the `command-fd` approach? I encounter the environment size issue occasionally but I've not found a good solution to it. I did not see any documentation on the branch.

Screwtapello:
The way Kakoune is designed at the moment, Kakoune has all the context (it has the buffer contents, it knows what expansions expand to, etc.) but the shell has the business logic (loops, conditionals, etc.). Thus, most plugins follow this pattern:

- request all the context they need as environment variables
- process that information in the shell
- produce Kakoune commands as output

Unfortunately, that doesn't always work. For example, a plugin might want context (like the buffer content) that's not available as an environment variable, or it might need to do some processing *before* requesting context from Kakoune (such as creating a temporary directory, or making a FIFO). In such cases, there's a couple of things you can do:

- you can do some shell-processing first, stick the intermediate results into a Kakoune register or a throwaway option, then do more work in Kakoune and then launch another shell context that grabs the temporary register (or option) contents and continue processing
    - This means saving/restoring custom registers, which is a pain, and things like child processes or network connections can't be stored in registers
- you can have one shell block that does some processing, calls `kak -p` to make Kakoune do the middle steps, then continues
    - This means messing with session IDs, and often buffer names and client IDs to make sure the commands execute in the same context as the shell block was invoked from

The command-fd branch adds a new, third approach. Instead of sending commands to `kak -p`, you can send them to file-descriptor 3.i

- Unlike writing commands to stdout (for `eval %sh{}`) or to `kak -p`, commands are executed as they arrive, not just when the file-descriptor is closed
- Unlike `kak -p`, the commands are executed in the expected context by default, because the command fd is associated with the place where the shell block was launched from, it's not a brand new client connection 
- Unlike `kak -p`, the only overhead is writing to a file-descriptor, not launching another process and connecting to a socket

There's nothing you can do with the command-fd branch you couldn't already do with Kakoune today, but some things are much nicer.

mawww:
@Screwtapello gave a good overview of what command-fd provides, one additional precision is that `kak -p` would not execute in parallel with the current shell block. This is necessary to avoid scripts having unforseen interactions with each other as we could end-up with Kakoune state changing at arbitrary points if `kak -p` was allowed to execute at any moment.

command-fd does not have that issue because it is only available during the execution of a `%sh{...}` block, so we can assume any side effects from commands sent through that fd are expected by the running script.

I view command-fd as a promising proof-of-concept because it solves some hard problems:
* messages boundaries are clear if we use a fifo to return data because we close it after each message
* combined with `echo -quoting ... -to-file ...` provides access to all exposed values in whatever quoting style we require
* Its basically as fast as it gets, no forks, just a fifo, only shared memory would beat that

It still has some drawbacks that are unsolved
* Only available during the execution of a `%sh{...}` block, cannot currently detach and keep a command fd around (that would have the same issue `kak -p` has)
* Working with fifo makes deadlock quite easy (opening for reading blocks until something opens the fifo for writing, small buffer size)
* Its a pretty arcane pattern

danr:
`mawww[m]` on IRC 01 Jun 2021:
> danr, andreyorst : I have been thinking about the command-fd branch recently and I think I have a design that should fix the env size issue while being  relatively easy to to use
Creating and cleaning up the response fifo is inconvenient boilerplate so the initial idea was to move that responsibility to kakoune
Then I figured that instead of having fd 3 as a command fifo, we could have Kakoune create a named fifo for requests as well, leading to this way of  requesting data from kakoune:
`echo "echo -to-file $kak_response_fifo %opt{my_big_option}" >$kak_request_fifo; my_big_option=$(cat $kak_response_fifo)`
Similar to other $kak_..., kakoune will create those fifo on-demand if it sees their name appear in the %sh{...} content
Currently this only works for the lifetime of a %sh{...} block (as with the command fd), but this is close to a communication channel that can stay alive  for detached co-process as well, we just need a way to defer clean-up of those fifo to when they are not needed anymore (detecting that this is the case  is the tricky part)
Using a named fifo for requests instead of fd 3 as in the current command-fd looks a bit nicer: Its more clear whats going on (`echo "commands" >  $kak_request_fifo` vs `echo "commands" >&3`), and the fifo can be opened/closed multiple times so we can just execute when the fifo is closed instead of  trying to execute partial commands every time we receive some data on it
That also means in the case of a detached co-process, we can defer execution to when we come back to the main loop, at which point its fair game to  change Kakoune's state because no scripts are being executed

This might also partially solve #4127 (though not the restarting part)

mawww:
I do not intend to merge this, I will merge the command-fifo branch that provides a more robust solution to getting the buffer content (and much more)

PR #4202, opened 2021-05-17
Kotlin: Add highlight for character literals and string interpolation

yerlaser:
Also adding character literals to java.kak

### DIFF
--- a/rc/filetype/kotlin.kak
+++ b/rc/filetype/kotlin.kak
@@ -56,6 +56,7 @@ provide-module kotlin %§
 add-highlighter shared/kotlin regions
 add-highlighter shared/kotlin/code default-region group
 add-highlighter shared/kotlin/string region %{(?<!')"} %{(?<!\\)(\\\\)*"} fill string
+add-highlighter shared/kotlin/character region %{(?<!")'} %{(?<!\\)(\\\\)*'} fill string
 add-highlighter shared/kotlin/comment region /\* \*/ fill comment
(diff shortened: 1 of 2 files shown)


rc/filetype/kotlin.kak
```diff
@@ -56,6 +56,7 @@ provide-module kotlin %§
 add-highlighter shared/kotlin regions
 add-highlighter shared/kotlin/code default-region group
 add-highlighter shared/kotlin/string region %{(?<!')"} %{(?<!\\)(\\\\)*"} fill string
+add-highlighter shared/kotlin/character region %{(?<!")'} %{(?<!\\)(\\\\)*'} fill string
```
krobelus: I guess this can be simplified a bit: we don't need the check for `"` because the string region is checked earlier, and there can be at most one backslash.

```suggestion
add-highlighter shared/java/character region %{'} %{(?<!\\)'} fill string
```

I don't know if that's really better, it's simpler anyway

krobelus: could you add the same change to `java.kak`? I think they are the same here

yerlaser:
Intentionally avoiding string interpolation of $var as it's considered bad practice.

yerlaser:
Hi @krobelus 
I updated the line as you suggested with the only difference that I think characters should be highlighted differently than strings.

I didn't understand your comment about Java though.
Do you mean Java strings inside a Kotlin file?

krobelus:
Looks good thanks!
I'd probably squash some of the commits, but that's not necessary.

Issue #4196, opened 2021-05-13
Markdown syntax highlighting is broken for list items that span multiple lines


dead10ck:
### Steps 

Write any markdown list where the bullet item spans multiple lines, where the continuation of the line has at least 4 spaces before the text starts. e.g.

```markdown
* this is a list item
  * this is a sub list item
    that spans multiple lines
```

This produces

![image](https://user-images.githubusercontent.com/846275/118160298-24717080-b3ec-11eb-9a60-452fab91914b.png)

Notice line 3 is highlighted.

### Outcome 

The text is highlighted

### Expected 

It shouldn't be

eraserhd:
I'm pretty sure this happens for asciidoc, too.

Frojdholm:
I think this is due to your indentation being 4 spaces, which then conflicts with the "indented code blocks" that at least exists in [Pandoc's version of Markdown](https://pandoc.org/MANUAL.html#indented-code-blocks).

kj:
Here's what I get with Pandoc:

```
$ pandoc -f markdown -t html test.md >test.html
```

test.md:

```markdown
* Test
  * Test
    Test
```

test.html:

```html
<ul>
<li>Test
<ul>
<li>Test Test</li>
</ul></li>
</ul>
```

Same result with `-f commonmark` and `-f gfm`. I think it's pretty standard to write list items like this, and I've only ever seen it treated as a code block in Kakoune. Kakoune's highlighting just isn't aware of the context.

Frojdholm:
The problematic highlighter is this one in `rc/filetype/markdown.kak:95`
```
add-highlighter shared/markdown/inline/text/ regex "^( {4}|\t)+([^\n]+)" 2:meta
```

PR #4208, opened 2021-05-24
Fix highlighting of indented bullet text

Frojdholm:
This tries to solve #4196 by:
1. Changing the `listblock` region to end on an empty line or a line starting with text (no whitespace). Previously the `listblock` region ended on the last bullet point line.
2. Move the indented code block highlighter outside of `shared/markdown/inline` so it's not included in the `listblock` region.

I'm not super confident in how the highlighting works so please let me know if I need to fix something.

### DIFF
--- a/rc/filetype/markdown.kak
+++ b/rc/filetype/markdown.kak
@@ -55,12 +55,13 @@ evaluate-commands %sh{
   done
 }
 
-add-highlighter shared/markdown/codeblock region -match-capture \
+add-highlighter shared/markdown/inline/codeblock region -match-capture \
     ^(\h*)```\h* \
     ^(\h*)```\h*$ \
     fill meta
+add-highlighter shared/markdown/codeblock region "^( {4}|\t)" "$" fill meta
 
-add-highlighter shared/markdown/listblock region ^\h*[-*]\s ^\h*((?=[-*])|$) regions
+add-highlighter shared/markdown/listblock region ^\h*[-*]\s ^(?=\S) regions
 add-highlighter shared/markdown/listblock/g default-region group
 add-highlighter shared/markdown/listblock/g/ ref markdown/inline
 add-highlighter shared/markdown/listblock/g/marker regex ^\h*([-*])\s 1:bullet
@@ -90,9 +91,6 @@ add-highlighter shared/markdown/inline/text/ regex (?<!\*)(\*\*([^\s*]|([^\s*](\
 add-highlighter shared/markdown/inline/text/ regex (?<!_)(__([^\s_]|([^\s_](\n?[^\n_])*[^\s_]))__)(?!_) 1:+b
 add-highlighter shared/markdown/inline/text/ regex ^\h*(>\h*)+ 0:comment
 add-highlighter shared/markdown/inline/text/ regex "\H( {2,})$" 1:+r@meta
-
-# Inline code
-add-highlighter shared/markdown/inline/text/ regex "^( {4}|\t)+([^\n]+)" 2:meta
 
 # Commands
 # ‾‾‾‾‾‾‾‾


Frojdholm:
Here is the test file that I used, [test.md](https://github.com/mawww/kakoune/files/6535386/test.md), content shown below:
```
asdf
```
    asdf <- should be highlighted
    asdf

- asdf
- asdf
  asdfasdf
  - asdf
    asdfasdf <- should not be highlighted
- asdfasdf
asdf
asdf
    asdf <- should be highlighted
    asdf

        asdf <- should be highlighted

dead10ck:
I tested this change locally with v2020.09.01 and it did not fix the issue. :/ If I am not mistaken, removing it from the inline group will also make it so that indented code blocks don't work from within a bullet list item.

I tried to take a crack at it myself, and I'm just a bit lost on how to properly express the regions. The best I understand it at an informal level, a bulleted list item starts a new embedded "indentation level", so a top-level bullet starts a region that stops at either:

* another bullet "at the same indentation level as the first bullet" or 
* any non-bullet text that is at a lower "indentation level", excepting blank lines

And the region can be recursive with an embedded bullet that starts at a higher indentation level.

This just seems really difficult to express with kakoune's highlighter expressions.

Screwtapello:
Breaking up a stream of characters into meaningful chunks (like "a list item" or "a variable declaration" or "a function definition") is called "parsing". There's lots of ways to go about parsing, but for a few decades now humans have known that [context-free grammars](https://en.wikipedia.org/wiki/Context-free_grammar) are a convenient middle-ground: CFGs can't describe *every* syntax or file-format, but a tool that can handle parsing a CFG can handle *any* CFG without any additional special-case hacks, and there's a lot of battle-hardened, high-performance tools for parsing CFGs.

Kakoune's syntax highlighting system is *not* a full CFG parser, but some parts of it (especially the region system) overlap the design space, and so it's pretty good at highlighting languages that conform to the limitations of CFGs.

Markdown was *not* designed to conform to the limitations of CFGs. It wasn't even designed to be portable or implemented by other people: originally, the only authoritative definition of "what is valid Markdown" was "whatever John Gruber's hacky Perl script accepts". These days there's [the CommonMark spec](https://spec.commonmark.org/) but because it mostly *describes* the behaviour of existing Markdown implementations, it's not any easier to parse. If you look at [the bit that describes indented code blocks](https://spec.commonmark.org/0.29/#indented-code-blocks), you'll see it doesn't provide a solid way to distinguish the two, it just says "when you hit this ambiguity, resolve it this way".

So Markdown will probably always be impossible to describe with Kakoune's highlighter expressions, because it's designed to be impossible to describe with any generic tool. Probably the best approach to getting Markdown highlighting working in Kakoune would be hooking into an external tool like [Tree-Sitter](https://tree-sitter.github.io/tree-sitter/) to calculate the formatting (except not Tree-Sitter specifically, because it too is a generic tool and therefore cannot parse Markdown on its own).

Frojdholm:
@dead10ck This is how it looks on kakoune compiled from this PR for me
kak-indented-bullets

@Screwtapello Thank you for your thorough explanation. I realize that this can't be done perfectly, however I think this change slightly improves the highlighting. There's still a problematic case when there are multiple paragraphs under a single nested bullet point (see above image), in the other cases it looks like it's matching GitHub now.

For me personally, since I never use the indented version of the code block, I might just remove the highlighter all together though..

dead10ck:
@Frojdholm see line 27. That's the case that #4196 was about. I'm unsure how to fix it. I also considered just removing the highlighter for indented code blocks, although I realized that that would mean any markdown syntax elements would still be highlighted within what's supposed to be an unformatted code block. That might still be better than the current situation though, as this is a pretty common case in Markdown.

Frojdholm:
I think line 17 is unfixable unfortunately. Line 11 and 20 are working now, they were highlighted previously.

For me, this still covers a lot of use cases since I rarely have multiple paragraphs under a single bullet point.

dead10ck:
It's very common for me personally, which makes the bug very prevalent and makes most of my markdown files much harder to read. Does anyone have any opinions on just removing the highlighter? Unless someone has an idea about how to fix it.

SeerLite:
> Does anyone have any opinions on just removing the highlighter?

Yeah, to me this sounds like the best solution. Makes it much more simple, considering that indentation in Markdown can mean so many different things in different contexts, and also different Markdown flavours understand it differently.

Neovim doesn't highlight indented blocks either and I assume it's because of the same reason.

Also I actually find the current indented code block highlighting kinda annoying, all with the same color. So personally I wouldn't miss it if it was removed.

If I were to write a code block I'd use the fenced syntax (\`\`\`) and specify the language so I get some actual syntax highlighting for it. If I were to find an indented code block in someone else's markdown file I think the indentation and the context would be enough for me to understand that it's code.

Screwtapello:
Some people use Markdown as a way to write HTML with fewer keypresses, other people use Markdown the way it was originally intended: as a way to write nicely-formatted plain-text documents with the added bonus that they can be automatically converted to HTML if necessary.

If you're writing a nicely-formatted plain-text document, making a code-block indented just looks nice, and is probably what you'd do even if you'd never heard of Markdown before. Indented code-blocks are the only syntax supported by older Markdown parsers.

Fenced code-blocks were invented by GitHub because it's difficult to reliably indent things by four spaces in a primitive browser text box like this one (unlike a Real Text Editor such as Kakoune). Because GitHub got so popular, a lot of people assume that fenced code blocks are "the way to make code blocks", and so many Markdown implementations support it, but not all (or at least not by default) — [a simple test](https://johnmacfarlane.net/babelmark2/?normalize=1&text=%60%60%60%0Aabc%0A%60%60%60) revealed that 10/29 tested Markdown implementations didn't actually produce a `<pre/>` element.

Frojdholm:
To clarify, I think removing the highlighter entirely should be something that is controlled in the user `kakrc`. With this change the highlighter can be remove using the following hook in your `kakrc`
```
hook global WinSetOption filetype=markdown %{
    remove-highlighter shared/markdown/codeblock
}
```

If this is not accepted I still think the inline code highlighter should be given a name so that it can be removed separately in the user's config.

Frojdholm:
Can we come to a conclusion on how this should be handled? My opinion is that it should be possible to disable this highlighter. Either by removing it entirely or by requiring a change in the user config.

As you said @Screwtapello some people, me included, use markdown as a way of writing nicely formatted plain text. This highlighter breaks the readability of that text in some cases and should be optional.

@dead10ck would a solution using an option be better than manually removing it from your perspective?

dead10ck:
Is there any reason we shouldn't just delete the highlighter?

Ordoviz:
@dead10ck You gave a reason yourself:

> any markdown syntax elements would still be highlighted within what's supposed to be an unformatted code block

We could of course now investigate what’s more likely to occur: Markdown formatting in indented code blocks or 4-indented lines in list blocks. I would err on the side of keeping the status quo, though.

dead10ck:
@Ordoviz you might read my next sentence though.

> That might still be better than the current situation though, as this is a pretty common case in Markdown.

I can't back this up with any data, but I would bet money that bullet points with 4+ whitespaces are going to be far more common than indented code blocks. I think it's already been mentioned that fenced code blocks have become much more favored.

Markdown highlighting happening within an indented code block is the norm in other text editors, I imagine for the same reason. I'm actually kind of shocked that I am having to push so hard for this.

sclu1034:
On `master`, **every line** in indented list blocks is broken (if you're using tabs or 4 spaces):
image

I agree that the current iteration of this PR is not perfect, but it certainly is better. Indented code is still highlighted for those who want it, but its overzealousness is restricted to ambiguous edge cases, rather than obvious list items.
And since the highlighter has a unique name, it can easily be removed by those who don't want/need it.

So I'd love to see this conversation be moved to a new issue and the improvements be merged, rather than blocking the PR in - what looks to me - an attempt to seek perfection.

dead10ck:
"ambiguous edge case"? Multi-line bullet list items are not an edge case, this is a far cry from agonizing over minute details. I'm talking about removing a single line, and that single line is for a feature that is far less common than the alternative syntax. Why is there such resistance to removing a broken highlighter?

I'm pretty frustrated here with the lack of care for user experience. Making users have to remove a highlighter in their kakrc to make Markdown readable at a basic level is kind of ridiculous. Maybe next we should just turn all highlighters off by default, and make users add the highlighters they want in their kakrc?

dead10ck:
If there were a bug that made kakoune crash 80% of the time you hit the `a` key, and you could fix it by turning off an option, would you accept a PR that made it crash only 50% of the time and tell users "oh you just need to set this option in your kakrc to make it stop crashing"?

sclu1034:
First and foremost: I don't care if that one highlighter makes it into `master` or not. I'm fine with adding three lines to my `kakrc`, regardless of whether that's adding or removing highlighters.
What I do care about is the fact that this PR, even in its current state, is a lot better than `master`, but it has been stuck in a discussion on that one line for more than a week.

To make your analogy more accurate to the current situation:

```
You have a bug that will crash Kakoune for every user with a chance of 80%.
You also have a PR that can reduce it to 0% for one part of the user base (size unknown) and reduce it to 20% for the others. Translation: everyone would win already.
But you are currently debating on an option that would flip these chances by introducing a different bug. I.e. (0%, 20%) vs (20%, 0%).

Would you rather
- keep debating for two weeks while all users are experiencing crashes at 80% chance?
- publish, so that at most some users crash with 20% chance, while debating which bug might have less impact?
```

I'd certainly take 20% over 80% and choose option #2.

> Multi-line bullet list items are not an edge case

Multi-line bullet points are fine with this PR. Only multi-paragraph (i.e. interrupted by blank line) bullet points in indented list blocks are still highlighted as code. Compared to the current status where _every single line_ in indented list blocks is broken in some way, that's a huge improvement.

And this is also the case where

- the linked CommonMarc Spec talks about ambiguity
- Kakoune's "slightly better than regular"-grammar might not be enough to parse what the spec asks for, since it would likely require keeping track of indentation levels

> feature that is far less common than the alternative syntax

I'd love to see some statistically relevant data on this. Fenced code blocks as an alternative to indented code blocks certainly don't look great:
"10/29 tested Markdown implementations didn't actually produce a `<pre/>` element [for fenced code blocks]" ([by Screwtapello](https://github.com/mawww/kakoune/pull/4208#issuecomment-849678258))

> Making users have to remove a highlighter in their kakrc to make Markdown readable at a basic level

If the highlighter were to be removed, that sentence could also be turned into "now you have to add a highlighter in your `kakrc` to get a basic feature like indented code blocks to work, even outside of list items".
And as for good user experience, it is still better to keep one broken thing that users had time to adapt to, than to exchange it for some other, new broken thing.

Frojdholm:
@mawww if you think this can be merged I can rebase and squash this so we can get it merged.

dead10ck:
> First and foremost: I don't care if that one highlighter makes it into `master` or not. I'm fine with adding three lines to my `kakrc`, regardless of whether that's adding or removing highlighters.
> What I do care about is the fact that this PR, even in its current state, is a lot better than `master`, but it has been stuck in a discussion on that one line for more than a week.

I'd like to point out that the last release for kakoune was September 2020. It's not like there's a big rush to get this fix into the next release, whenever that happens to be. Judging by your response, it seems like you're someone who pulls and compiles kakoune yourself locally. This is not the typical way non-diehard fans who want the bleeding edge get their updates. And if you do pull and compile yourself, then you should have no problem pulling this fix yourself and using it in the mean time.

I'd also like to point out that I'm not someone with authority to merge. Any author can feel free to render their own judgment about merging or not.

> And this is also the case where
> 
>     * the linked CommonMarc Spec talks about ambiguity
> 
>     * Kakoune's "slightly better than regular"-grammar might not be enough to parse what the spec asks for, since it would likely require keeping track of indentation levels

It seems to me that if there's an ambiguity in the language itself, then the right response is not to try to highlight it. It will just be broken. This is what most other editors do.

> > feature that is far less common than the alternative syntax
> 
> I'd love to see some statistically relevant data on this. Fenced code blocks as an alternative to indented code blocks certainly don't look great:
> "10/29 tested Markdown implementations didn't actually produce a `<pre/>` element [for fenced code blocks]" ([by Screwtapello](https://github.com/mawww/kakoune/pull/4208#issuecomment-849678258))

I'm not really sure what to make of this. Looking at this web site that @Screwtapello linked to, it's hard to actually interpret the results of this web page. It seems to be a table of some kind, but the renderers listed on the left don't match up with the rendered text boxes on the right, so I'm not even sure this web site is working. Some of the API calls it makes even 500. It also looks like some of the "29 renderers" are actually the same renderer with different flavors. So of the "29" you're including renderers that intentionally don't render that.

Furthermore, if you want numbers on how many users and/or Markdown documents you can expect this to affect, this is a rather poor metric. It's not weighted by how popular the renderer is, for example. The renderer that GitHub uses will affect tons and tons more users than many others, for example. GitHub might not capture a representative sample of everyone who uses markdown, but I would wager that it will have a large intersection with users of kakoune, which is primarily a code editor.

What you'd really need if you wanted accurate metrics are how many occurrences of the fenced block vs the indented block there are. Given the parsing ambiguity, this seems impossible anyway. But if you'd like to see how many people use fenced code blocks, you can take a look on [SourceGraph](https://sourcegraph.com/search?q=context:global+lang:Markdown+%60%60%60&patternType=literal).

> And as for good user experience, it is still better to keep one broken thing that users had time to adapt to, than to exchange it for some other, new broken thing.

Judging by the fact that kakoune is not nearly as popular as other editors, and that other editors don't try to highlight indented code blocks, I don't think very many people will be upset by what passes for status quo everywhere else.

basbebe:
to unnecessarily also chime in here: 
[I also came across this, creating a duplicate issue](https://github.com/mawww/kakoune/issues/4247), while I was trying to add some latex and html syntax highlighting to markdown files when using a pandoc plugin.

Highlighted nested lists have been annoying me for a long time now as this (as pointed out in other comments) drastically reduces readability in what would otherwise be a clean and easy-to-read format.

The behavior shown by this PR seems to be exactly what I personally would a text editor expect to do and it seems to be in line with the recommendations of CommonMark.

Looking forward to pulling this from master

mawww:
Hello, I dont use markdown enough to really have an opinion on what behaviour should be followed, and I was hoping the discussion here would get to some kind of consensus on whether this should be merged.

It looks to me as there is overall a stronger preference towards this getting merged and that it generally gives better result than the status quo. I guess we could merge it for now and see how it goes ?

dead10ck:
I'm still of the opinion that the indented code block highlighter should just get removed, but between status quo and this PR, I'd rather have this PR merged.

krobelus:
> I think line 17 is unfixable

I think we can fix this as well by disabling the indented-block highlighter inside lists:
just make a list region span until the next line that has no indentation.
This means that the indented-block highlighter will still work fine in many plain-text-style documents.

```Markdown
- asdf

    asdf <- should not be highlighted, because it's inside a list

asdf asdf asdf

    asdf <- should be highlighted because it's not inside a list
```

Diff to master:

```diff
diff --git a/rc/filetype/markdown.kak b/rc/filetype/markdown.kak
index f0337ffc..eddf45d4 100644
--- a/rc/filetype/markdown.kak
+++ b/rc/filetype/markdown.kak
@@ -57,3 +57,3 @@ evaluate-commands %sh{
 
-add-highlighter shared/markdown/codeblock region -match-capture \
+add-highlighter shared/markdown/inline/codeblock region -match-capture \
     ^(\h*)```\h* \
@@ -61,4 +61,5 @@ add-highlighter shared/markdown/codeblock region -match-capture \
     fill meta
+add-highlighter shared/markdown/codeblock region "^( {4}|\t)" $ fill meta
 
-add-highlighter shared/markdown/listblock region ^\h*[-*]\s ^\h*((?=[-*])|$) regions
+add-highlighter shared/markdown/listblock region ^\h*[-*]\s ^(?=\S) regions
 add-highlighter shared/markdown/listblock/g default-region group
@@ -93,5 +94,2 @@ add-highlighter shared/markdown/inline/text/ regex "\H( {2,})$" 1:+r@meta
 
-# Inline code
-add-highlighter shared/markdown/inline/text/ regex "^( {4}|\t)+([^\n]+)" 2:meta
-
 # Commands
```

I think this will work well in practise. Of course there are pathological cases, and there are limits to what we can do.

(The [spec](https://spec.commonmark.org/0.29/#lists) says that "list items may be separated by any number of blank lines" so we'd even improve on that point, though not for nested lists.)

Frojdholm:
I went with @krobelus suggestion for when to end the list region since I think that solves more problems than it introduces.

An obvious edge case that is problematic is a list followed by an indented code block. With this change at least it's possible to work around the missing highlighting by adding fences.

@mawww this should be ready to merge now.

PR #4211, opened 2021-06-02
fix: iterm focus client from different tab or window

gkrrr:
currently focus client doesn't work if target pane of the client is in different tab or window from current one.

**select window, select tab should be triggered in order to select a pane on a currently not focused tab or window.**


mawww:
Thanks !

PR #4220, opened 2021-06-09
Clarify which parts of Kakoune the extra_word_chars option affects.

Screwtapello:
Also, document the "empty option implies underscore" quirk. I do wonder why that quirk exists, instead of just making underscore the global default value.

### DIFF
--- a/doc/pages/options.asciidoc
+++ b/doc/pages/options.asciidoc
@@ -273,7 +273,11 @@ are exclusively available to built-in options.
 
 *extra_word_chars* `codepoint-list`::
     a list of all additional codepoints that should be considered
-    as word character. This must be set on the buffer, not the window,
+    part of a word, for the purposes of the `w`, `b`, and `e` commands
+    (See <<keys#movement,`:doc keys movement`>>).
+    If this option is empty, Kakoune pretends it contains an
+    underscore, otherwise the value is used as-is.
+    This must be set on the buffer, not the window,
     for word completion to offer words containing these codepoints.
 
 *matching_pairs* `codepoint-list`::
--- a/doc/pages/regex.asciidoc
+++ b/doc/pages/regex.asciidoc
@@ -56,10 +56,11 @@ characters, the `-` character, and the `+` character.
 
 Supported character class escapes are:
 
-* `\d` which matches all digits.
-* `\w` which matches all word characters.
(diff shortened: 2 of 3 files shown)


doc/pages/options.asciidoc
```diff
@@ -273,7 +273,11 @@ are exclusively available to built-in options.
 
 *extra_word_chars* `codepoint-list`::
     a list of all additional codepoints that should be considered
-    as word character. This must be set on the buffer, not the window,
+    part of a word, for the purposes of the `w`, `b`, and `e` commands
+    (See <<keys#movement,`:doc keys movement`>>).
+    If this option is empty, Kakoune pretends it contains an
+    underscore, otherwise the value is used as-is.
```
krobelus, line 279: > Kakoune pretends it contains an underscore, otherwise the value is used as-is.

Maybe this is so `kak -f` works as expected (and I guess also `kak -n`, even though that seems less important).
If my guess is correct we might want ot add it:

```
If this option is empty, Kakoune pretends it contains an underscore. This
allows conventional word movement without loading configuration, like in `kak -f`.
```

(The "otherwise" part seems redundant.)

doc/pages/options.asciidoc
```diff
@@ -273,7 +273,11 @@ are exclusively available to built-in options.
...
+    This must be set on the buffer, not the window,
```
krobelus, line 280: Interesting. If there are more options like this, it might be worth to add an error message instead.

doc/pages/regex.asciidoc
```diff
@@ -56,10 +56,11 @@ characters, the `-` character, and the `+` character.
 
 Supported character class escapes are:
 
-* `\d` which matches all digits.
-* `\w` which matches all word characters.
-* `\s` which matches all whitespace characters.
-* `\h` which matches all horizontal whitespace characters.
+* `\d` which matches digits 0-9.
+* `\w` which matches word characters A-Z, a-z, 0-9 and underscore
```
krobelus, line 60: It actually matches everything that is accepted by `iswalnum` (which is locale-dependent).
Giving some examples (but not the full list) might still be a good idea:

```
... word characters like `A`, `é`, and `_` (ignoring the `extra_word_chars` option).
```

krobelus (changes requested):

PR #4221, opened 2021-06-10
Document how to escape special characters in completions.

Screwtapello:
[Rendered](https://github.com/Screwtapello/kakoune/blob/document-completion-escaping/doc/pages/options.asciidoc#types) (scroll down to the "completions" definition)


krobelus:
> Rendered (scroll down to the "completions" definition)

Looks like GitHub can now show a somewhat useful diff of the rendered version (the button saying "Display the rich diff" on the top right).

PR #4224, opened 2021-06-13
Fix typos

alexherbo2:


mawww:
I am not sure that those are typos, its unclear to me which form is more correct, any native speaker around ?

CharlesGueunet:
I am not a native, but these changes seems to be in accordance with : https://practicaltypography.com/parentheses-brackets-and-braces.html

raiguard:
I'm a native speaker. It definitely works both ways, but the original "feels" more right to me.

eraserhd:
I'm a native (American) english speaker, and "curly brace block" sounds more natural.  I don't have a good reason, but we also say "ice cube tray" and not "ice cubes tray", for example.  So, I guess I think the reason is that adjectives are not generally pluralized?

So let us [look it up](https://learningenglish.voanews.com/a/everyday-grammar-when-nouns-act-like-adjectives/2998821.html).  Oh god, english is a mess.  "Sports car", which sounds odd to me if I think about it, except that I've heard it my whole life and say it that way.

Seems the rule is that nouns used as adjectives are singular by default, and exceptions are accidental plural phrases we seem to have already accepted for no particular reason.

To make matters worse, the phrase "a bracketed block" works nicely, but "a braced block" does not, because "braced" means something else.

raiguard:
And this is the problem with English - you have to run on feeling! It's such a mess of inconsistencies.

eraserhd did actual research, so their answer is probably more well-founded than my off-the-cuff opinion.

alexherbo2:
@mawww I also found occurrences for [hoooks].

[hoooks]: https://github.com/mawww/kakoune/search?q=hoooks

PR #4231, opened 2021-06-15
doc/kak.1: Convert to mdoc(7) semantic markup

somasis:
This aids in making it more uniform in style with most manuals on *nix systems in terms of formatting.
Plus, the semantic markup will make it easier to maintain and less reliant on writing for the renderer.

I've rewritten some phrasing and lists at points to make it take better advantage of formatting seen in tools like [ssh(1)](https://manpages.debian.org/ssh(1)), [tmux(1)](https://manpages.debian.org/tmux(1)), and so on; unfortunately this couldn't really be untangled from the syntax conversion without being annoying. Hopefully these changes will make it more readable when reading through `man`.

### DIFF
--- a/doc/kak.1
+++ b/doc/kak.1
@@ -1,251 +1,276 @@
-.TH KAK 1
-
-.SH NAME
-
-kak \- a vim inspired, selection oriented code editor
-
-.SH SYNOPSIS
-
-.PP
-.B kak \-help
-
-.PP
-.B kak \-version
-
-.PP
-.B kak \-l
-
-.PP
-.B kak \-clear
-
-.PP
-.B kak \-f
-.I keys
-[\fB\-q\fR] [\fB\-i\fR]
-.IR file ...
-
-.PP
-.B kak \-p
-.I session_id
-
-.PP
-.B kak \-s
-.I session_id
-.B \-d
-[\fB\-n\fR] [\fB\-ro\fR]
-[\fB\-E\fR \fIcommand\fR]
-[\fB+line\fR[\fB:column\fR]|\fB+:\fR]
-.IR file ...
-
-.PP
-.B kak
-[\fB\-c\fR \fIsession_id\fR|\fB\-s\fR \fIsession_id\fR]
-[\fB\-n\fR] [\fB\-ro\fR]
-[\fB\-ui\fR \fIui_type\fR] [\fB\-e\fR \fIcommand\fR]
-[\fB\-E\fR \fIcommand\fR]
-[\fB+line\fR[\fB:column\fR]|\fB+:\fR]
-.IR file ...
-
-.SH DESCRIPTION
-
-Kakoune is a code editor heavily inspired by Vim. As such, most of its commands
-are similar to Vi's and it shares Vi's "keystrokes as a text editing
-language" model.
-
-Kakoune can operate in two modes, normal and insertion. In insertion mode,
-keys are directly inserted into the current buffer. In normal mode, keys
-are used to manipulate the current selection and to enter insertion mode.
-
-Kakoune has a strong focus on interactivity, most commands provide immediate
-and incremental results, while still being competitive (as in keystroke
-count) with Vim.
-
-Kakoune works on selections, which are oriented, inclusive ranges of
-characters. Selections have an anchor and a cursor. Most commands
-move both of them except when extending selection, where the anchor character
-stays fixed and the cursor moves around.
-
-For more information, use the \fB:doc\fR command after starting Kakoune,
-the Kakoune wiki at https://github.com/mawww/kakoune/wiki
-or the main Kakoune web site: https://kakoune.org/
-
-.SH OPTIONS
-
-.TP
-.BR \-help
-display a help message and quit
-
-.TP
-.BR \-version
-display kakoune version and quit
-
-.TP
-.BR \-n
-do not load resource files on startup (\fIkakrc\fR, \fIautoload\fR, \fIrc\fR etc)
-
-.TP
-.BR \-l
-list existing sessions
-
-.TP
-.BR \-d
-run as a headless session (requires \fB\-s\fR)
-
-.TP
-.BR \-e " " \fIcommand\fR
-execute \fIcommand\fR after the client initialization phase
-
-.TP
-.BR \-E " " \fIcommand\fR
-execute \fIcommand\fR after the server initialization phase
-
-.TP
-.BR \-f " " \fIkeys\fR
-enter in \fIfilter mode\fR: select the whole file, then execute \fIkeys\fR
-
-.TP
-.BR \-i " " \fIsuffix\fR
-backup the files on which a filter is applied using the given suffix
-
-.TP
-.BR \-q
-when in \fIfilter mode\fR, don't print any errors
-
-.TP
-.BR \-p " " \fIsession_id\fR
-send the commands written on the standard input to session \fIsession_id\fR
-
-.TP
-.BR \-c " " \fIsession_id\fR
-connect to the given session
-
-.TP
-.BR \-s " " \fIsession_id\fR
-set the current session name to \fIsession_id\fR
-
-.TP
-.BR \-ui " " \fItype\fR
-select the user interface, can be one of \fIncurses\fR, \fIdummy\fR or \fIjson\fR
-
-.TP
-.BR \-clear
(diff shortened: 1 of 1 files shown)


somasis:
For a feel of how it looks now:
```text
KAK(1)                      General Commands Manual                     KAK(1)

NAME
     kak – a vim-inspired, selection oriented code editor

SYNOPSIS
     kak [-d] [-n] [-ro] [-c session_id | -s session_id] [-ui ui_type]
         [-e command] [-E command] [+line[:column] | +:] [file ...]
     kak -f keys [-q] [-i suffix] [file ...]
     kak -p session_id
     kak -l
     kak -clear
     kak -version
     kak -help

DESCRIPTION
     Kakoune is a code editor heavily inspired by vim(1) and vi(1).  As such,
     most of its commands are similar to vi(1)'s, sharing its “keystrokes as a
     text editing language” model.

     Kakoune operates in two modes, normal and insertion.  In insertion mode,
     keys are directly inserted into the current buffer.  In normal mode, keys
     are used to manipulate the current selection and to enter insertion mode.

     Kakoune has a strong focus on interactivity.  Most commands provide
     immediate and incremental results, while still being competitive in
     keystroke count with vim.

     Kakoune works on selections, which are oriented, inclusive ranges of
     characters.  Selections have an anchor and a cursor.  Most commands move
     both of them except when extending selection, where the anchor character
     stays fixed and the cursor moves around.

     The options are as follows:

     -help   Display a help message and quit.

     -version
             Display Kakoune version and quit.

     -n      Do not load the system's kakrc.  (often, /usr/share/kak/kakrc)

     -l      List existing sessions.

     -d      Run as a headless session (requires -s).

     -e command
             Execute command after the client initialization phase.

     -E command
             Execute command after the server initialization phase.

     -f keys
             Enter in ‘filter mode’: select the whole file, then execute keys.

     -i suffix
             Backup the files on which a filter is applied, using the given
             suffix.

     -q      When in ‘filter mode’, don't print any errors

     -p session_id
             Send the commands written on the standard input to session
             session_id.

     -c session_id
             Connect to the given session session_id.

     -s session_id
             Set the current session name to session_id.

     -ui type
             Select user interface type, which can be ncurses, dummy, or json.

     -clear  Remove sessions that were terminated in an incorrect state (e.g.
             after a crash).

     -ro     Begin in readonly mode, all the buffers opened will not be
             written to disk.

     +line[:column] | +:
             Specify a target +line and :column for the first file.  When the
             plus sign is followed by only a colon, then the cursor is sent to
             the last line of the file.

     file ...
             One or more file(s) to edit.

ENVIRONMENT
     KAKOUNE_POSIX_SHELL
             Overrides the POSIX shell used for %sh{...} expansion, which is
             /bin/sh (sh(1)) if unset.

     KAKOUNE_CONFIG_DIR
             Overrides the location of the directory containing Kakoune user
             configuration.  If unset, $XDG_CONFIG_HOME/kak is used.

     XDG_CONFIG_HOME
             Path to the user's configuration directory.  If unset,
             $HOME/.config is used.

     XDG_RUNTIME_DIR
             Path to the user's session's sockets.  If unset, $TMPDIR/kakoune
             is used.

FILES
     In the paths documented below, <rtdir> refers to the runtime directory,
     whose value is determined in relation to the path to kak's binary
     location: <rtdir> = <path_to_kak_binary>/../share/kak.

     If not started with the -n switch, kak will first load <rtdir>/kakrc,
     which will in turn load the following additional files:

     •   If the $KAKOUNE_CONFIG_DIR/autoload directory exists, recursively
         load every *.kak file in it, and its sub-directories.

     •   If it does not exist, fall back to the system-wide autoload directory
         in <rtdir>/autoload, and recursively load all files in a similar way.

     •   <rtdir>/kakrc.local, if it exists; this is a user-defined system-wide
         configuration.

     •   $KAKOUNE_CONFIG_DIR/kakrc, if it exists; this is the user
         configuration.

     Consequently, if the $KAKOUNE_CONFIG_DIR/autoload directory exists, only
     scripts stored within that directory will be loaded — the built-in *.kak
     files will not be.

     Users who still want the built-in scripts to be loaded along their own
     can create a symbolic link to <rtdir>/autoload (or to individual scripts
     in it) in their user-configuration directory:

           ln -s <rtdir>/autoload ${XDG_CONFIG_HOME:-$HOME/.config}
           /kak/autoload

EXAMPLES
     Edit a file:
             kak /path/to/file

     Edit multiple files (multiple buffers will be created):
             kak ./file1.txt /path/to/file2.c

     Insert a modeline that sets the tabstop variable at the beginning of
             several source code files:
             kak -f "ggO// kak: tabstop=8<esc>" *.c

SEE ALSO
     For the complete on-line documentation, use the :doc command after
     starting kak.

     The Kakoune wiki: https://github.com/mawww/kakoune/wiki.

     The main Kakoune web site: https://kakoune.org.

     vi(1), vim(1), sam(1plan9).

AUTHORS
     Maxime Coste <frrrwww@gmail.com> and many others.

Linux 5.12.9-arch1-1             June 15, 2021            Linux 5.12.9-arch1-1
```

doc/kak.1
```diff
@@ -1,251 +1,282 @@
...
+.Sh SYNOPSIS
+.Nm
+.Op Fl d
+.Op Fl n
+.Op Fl ro
+.Op Fl c Ar session_id | Fl s Ar session_id
+.Op Fl ui Ar ui_type
+.Op Fl e Ar command
+.Op Fl E Ar command
+.Op Ar +line Ns Oo Ar :column Oc | +:
```
Ordoviz, line 17: The plus sign and the colon should be bold since they need to be typed literally: `.Op Sy + Ns Ar line Ns Oo Sy \&: Ns Ar column Oc | Sy +:`. mdoc(7) has this to say about the .Sy macro: “When none of the semantic macros fit, it is also adequate for syntax elements that have to be given or that appear verbatim.” The `\&` inserts a zero-width space used to escape the colon, which is treated as a delimiter otherwise.

doc/kak.1
```diff
@@ -1,251 +1,282 @@
...
+.It Fl clear
+Remove sessions that were terminated in an incorrect state
+.Pq e.g. after a crash .
+.
+.It Fl ro
+Begin in
+.Em readonly mode ,
+all the buffers opened will not be written to disk.
+.
+.It Ar +line Ns Oo Ar :column Oc | Ar +:
```
Ordoviz, line 142: This should be `.It Sy + Ns Ar line Ns Oo Sy \&: Ns Ar column Oc | Sy +:`. See above comment.

doc/kak.1
```diff
@@ -1,251 +1,282 @@
...
+Connect to the given session
+.Ar session_id .
+.
+.It Fl s Ar session_id
+Set the current session name to
+.Ar session_id .
+.
+.It Fl ui Ar type
+Select user interface
+.Ar type ,
```
Ordoviz, line 126: *type* here is not a placeholder for the argument to the `-ui` flag so no macros needed.

doc/kak.1
```diff
@@ -1,251 +1,282 @@
...
+.Pq e.g. after a crash .
+.
+.It Fl ro
+Begin in
+.Em readonly mode ,
+all the buffers opened will not be written to disk.
+.
+.It Ar +line Ns Oo Ar :column Oc | Ar +:
+Specify a target
+.Pf + Ar line
```
Ordoviz, line 144: Similar to *type*, *line* and *column* do not refer to arguments. They refer to English words so no macros needed.

doc/kak.1
```diff
@@ -1,251 +1,282 @@
...
+.It Edit a file:
+.Nm
+.Pa /path/to/file
+.
+.It Edit multiple files (multiple buffers will be created):
+.Nm
+.Pa ./file1.txt
+.Pa /path/to/file2.c
+.
+.It Insert a modeline that sets the tabstop variable at the beginning of several source code files:
```
Ordoviz, line 259: If the screen is too narrow for this example to fit on one line, the second line will be indented, which looks odd.

Ordoviz (changes requested):

doc/kak.1
somasis: I noticed this as well and wasn't quite sure what was preferrable-- rewriting the sentence further, or trying to work around the wrapping in some way, since I couldn't find a way to have it wrap around to the start of the `.It` line anyway. I've rewritten it further.

PR #4239, opened 2021-06-20
Update color scheme template

alexherbo2:


colors/default.kak
```diff
@@ -1,50 +1,57 @@
 # Kakoune default color scheme
+#
+# Template: https://github.com/mawww/kakoune/blob/master/colors/default.kak
+# Documentation: https://github.com/mawww/kakoune/blob/master/doc/pages/faces.asciidoc
```
mawww, line 4: I would prefer no references to github in the codebase, I cannot see any good reason to direct people towards the WWW version of files that they already have on their filesystem.

alexherbo2: It was for people browsing the file and persons writing color schemes.

I find having the URLs at hand make it easy to notice changes and where the custom color schemes come from.

mawww: I still dont think the code should refer to wherever my Kakoune git repository happens to be hosted at the moment, I´d like to keep as low as possible coupling with github.

PR #4240, opened 2021-06-21
Fix windowing detection priority

alexherbo2:


mawww:
Merged, I agree we should give wayland priority over x11, but can you put some rationale for the change next time please.

PR #4243, opened 2021-06-24
Kakoune passed environment variables in shell invocations are repeated

sidkshatriya:
```
    Fix: Kakoune passed environment variables in shell invocations are repeated
    
    If a %sh{} script refers to any variables multiple times they are all multiply
    included in the environment. Example: if a %sh{} invocation refers to
    ${kak_buffile} 5 times, the environment will have "kak_buffile=..." repeated 5
    times and so on. This repetition happens for each multiply used variable that
    is passed into the environment.
    
    The variable should, of course, be only passed into the environment once. This
    commit should fix this issue.
```

### DIFF
--- a/src/shell_manager.cc
+++ b/src/shell_manager.cc
@@ -144,8 +144,9 @@ Vector<String> generate_env(StringView cmdline, const Context& context, const Sh
         StringView name{match[2].first, match[2].second};
 
         auto match_name = [&](const String& s) {
-            return s.substr(0_byte, name.length()) == name and
-                   s.substr(name.length(), 1_byte) == "=";
+            // 4_byte because of the initial `kak_` prefix
+            return s.substr(4_byte, name.length()) == name and
+                   s.substr(4_byte + name.length(), 1_byte) == "=";
         };
         if (any_of(env, match_name))
             continue;
--- a/src/shell_manager.cc
+++ b/src/shell_manager.cc
@@ -142,11 +142,11 @@ Vector<String> generate_env(StringView cmdline, const Context& context, const Sh
     for (auto&& match : RegexIterator{cmdline.begin(), cmdline.end(), re})
     {
         StringView name{match[2].first, match[2].second};
+        StringView shell_name{match[0].first, match[0].second};
 
         auto match_name = [&](const String& s) {
-            // 4_byte because of the initial `kak_` prefix
-            return s.substr(4_byte, name.length()) == name and
-                   s.substr(4_byte + name.length(), 1_byte) == "=";
+            return s.substr(0_byte, shell_name.length()) == shell_name and
+                   s.substr(0_byte + shell_name.length(), 1_byte) == "=";
         };
         if (any_of(env, match_name))
             continue;
--- a/src/shell_manager.cc
+++ b/src/shell_manager.cc
@@ -146,7 +146,7 @@ Vector<String> generate_env(StringView cmdline, const Context& context, const Sh
 
         auto match_name = [&](const String& s) {
             return s.substr(0_byte, shell_name.length()) == shell_name and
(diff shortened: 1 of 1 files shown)


occivink:
I cannot reproduce the issue. `StringView name{match[2].first, match[2].second}` contains the second capture group of the regex, which should not include the kak_ prefix.

Screwtapello:
Looking at `shell_manager.cc`, I can see why Kakoune provides redundant environment variable definitions to `execve()`, as described in this commit message. I don't understand why this change fixes the issue, though it's late and I may be missing something important.

I can't find any documentation stating what happens when `execve()` is given redundant definitions of an environment variable, but it's *probably* harmlessly deduplicated by the kernel. It'd be nice to not require the kernel to do that extra work, though.

sidkshatriya:
The problem definitely exists. (I notice that @occivink has edited his comment and he also notices the issue). My current fix _partially_ fixes the problem and does not deal with quoted case as he correctly points out. I will push out a fix that deals with the quoted case also shortly.

To see the issue in action, for the following shell script (as a random example) invocation by kak:

```
sh -c 
    autoload_directory() {
        find -L "$1" -type f -name '*\.kak' \
            | sed 's/.*/try %{ source "&" } catch %{ echo -debug Autoload: could not load "&" }/'
    }

    echo "colorscheme default"

    if [ -d "${kak_config}/autoload" ]; then
        autoload_directory ${kak_config}/autoload
    elif [ -d "${kak_runtime}/autoload" ]; then
        autoload_directory ${kak_runtime}/autoload
    fi

    if [ -f "${kak_runtime}/kakrc.local" ]; then
        echo "source '${kak_runtime}/kakrc.local'"
    fi

    if [ -f "${kak_config}/kakrc" ]; then
        echo "source '${kak_config}/kakrc'"
    fi
```

If you debug `/usr/bin/dash` in `rr`

```
(rr) b main
Breakpoint 1 at 0x564fdc2d8af0: file main.c, line 91.
(rr) c
Continuing.

Breakpoint 1, main (argc=3, argv=0x7ffd6c0b0768) at main.c:91
91      main.c: No such file or directory.
(rr) print environ[61]
$1 = 0x7ffd6c0b1ec8 "kak_config=/home/sidk/.config/kak"
(rr) print environ[62]
$2 = 0x7ffd6c0b1eea "kak_config=/home/sidk/.config/kak"
(rr) print environ[63]
$3 = 0x7ffd6c0b1f0c "kak_runtime=/home/sidk/.local/share/kak"
(rr) print environ[64]
$4 = 0x7ffd6c0b1f34 "kak_runtime=/home/sidk/.local/share/kak"
(rr) print environ[65]
$5 = 0x7ffd6c0b1f5c "kak_runtime=/home/sidk/.local/share/kak"
(rr) print environ[66]
$6 = 0x7ffd6c0b1f84 "kak_runtime=/home/sidk/.local/share/kak"
(rr) print environ[67]
$7 = 0x7ffd6c0b1fac "kak_config=/home/sidk/.config/kak"
(rr) print environ[68]
$8 = 0x7ffd6c0b1fce "kak_config=/home/sidk/.config/kak"
(rr) print environ[69]
$9 = 0x0
(rr) 
```
Notice how the `kak_runtime` and `kak_config` environment variables are needlessly repeated in the environment.

sidkshatriya:
I think this should be ready now. Please let me know what the convention is and I'll be happy to squash the last three small commits into a single commit and force push on this branch.

Screwtapello:
Ah, I get it now. Kakoune was trying to dedupe environment variables itself, but failing because it was comparing the basename (`selections_desc`) against the shell name (`kak_quoted_selections_desc`) and therefore never noticing the duplicates.

Issue #3355, opened 2020-02-09
Add support for range in %arg{…}


alexherbo2:
**Example**

``` kak
echo %arg{1..}
```

lenormf:
This particular example is already possible: `echo %arg{@}`.

As for allowing arbitrary splicing indexes, that's typically handled in a shell scope with `shift` and other control statements provided by the shell language.

Issue #2353, opened 2018-09-02
No way to access %arg(@) from shell


alexherbo2:

mawww:
I dont get what you want ? `"$@"` in shell is guaranteed to expand as multiple arguments, you have `$#`,  `$1..N`, `shift` and various other tools to manipulate them, whats your use case ?

alexherbo2:
https://github.com/alexherbo2/auto-pairs.kak

mawww:
What do you use that for ? To iterate over both `$kak_opt_auto_pairs` and the positional arguments ? Then how about

```
iterate() { ...iterate on $@... }
eval "iteratate $kak_opt_pairs"
iteratate "$@"
```

alexherbo2:
I wish I could avoid to define `iterate`, but simply use `set` to append to the argument list.

Screwtapello:
Is this like #3307?

alexherbo2:
Yes.

PR #4249, opened 2021-06-30
expansions: %arg{@:N} (args from N onward)

mralusw:
Some extra expansions. For now

### `%arg{@:N}`
```
1-based, like %arg{N}; shell-like syntax
test:
def -params .. echodbg %{ echo -debug -quoting shell -- %arg{@} } -override
def -params .. tst %{ echodbg %arg{@:3} } -override
tst 9 8 7 6 5
tst 1 2 3
tst 1
tst
def -params .. tst %{ echodbg "%arg{@:3}" } -override # and repeat
```

It's costly to call the shell to execute another command with the first N args removed.

Possible enhancement: `%arg{@:N:count}` (again, as already common in shells).

### DIFF
--- a/src/command_manager.cc
+++ b/src/command_manager.cc
@@ -374,6 +374,16 @@ expand_token(const Token& token, const Context& context, const ShellContext& she
             else
                 return Vector<String>{params.begin(), params.end()};
         }
+        else if (prefix_match(content, "@:")) {
+            const int arg1 = str_to_int(content.substr (2_byte)) - 1;
+            if (arg1 < 0)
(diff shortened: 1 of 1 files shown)


Delapouite:
Related issues about args : 

- https://github.com/mawww/kakoune/issues/3355
- https://github.com/mawww/kakoune/issues/2353

mawww:
I am really not keen to expand the command language to add more scripting logic support, I know this is just a small step, but one of many proposed. I would really need a strong use case for that, some actual script thats being used where the cost of forking a shell proves too expensive and where this feature would fix the problem.

PR #4256, opened 2021-07-09
Add `markdown` comment region

razcore-rad:
I saw `comment_block_*` entries are defined for `markdown`, but not the highlight region. This adds that region.

image


razcore-rad:
Actually this crosses over into the next lines and I'm not sure if that's valid, but I'd need some help figuring out the closing regex :)

image

razcore-rad:
I updated this to include HTML-type comments and restrict the comment hack ([see stackoverflow solution](https://stackoverflow.com/questions/4823468/comments-in-markdown#comment33676653_20885980)) to one line:

```markdown
[//]: # "some comment"

<!--
HTML-type comment
-->
```

PR #4261, opened 2021-07-13
Lua: insert comment prefix *after* the indent

hugomg:
This fixes a bug in how the Lua scripts handle new comment lines. Currently if we have a comment that is indented, when we add a new line it inserts the `--` prefix before the automatic indentation.

```
  --ABC
--  XYZ
```

After the fix, it correctly inserts the comment prefix after the indentation:

```
  --ABC
  --XYZ
```

The solution I used is inspired by the ruby.kak script.


hugomg:
Added a copyright waiver, following the CONTRIBUTING document.

PR #4263, opened 2021-07-13
Feature: In `git diff` buffer navigate to exact source line by pressing `<ret>` within a hunk 

sidkshatriya:
When viewing `:git diff` within a kakoune buffer, this commit adds the ability to press `<ret>` within a hunk and navigate to the source code at the exact line in the hunk

This makes it easy to go back and forth between the source code and `git diff` in kakoune

(Edited)
- Now also works for :git show

### DIFF
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -74,9 +74,12 @@ define-command -params 1.. \
 
     show_git_cmd_output() {
         local filetype
+        local map_diff_goto_source
+
         case "$1" in
-           diff) filetype=diff ;;
-           log|show)  filetype=git-log ;;
+           diff) map_diff_goto_source=true; filetype=diff ;;
+           show) map_diff_goto_source=true; filetype=git-log ;;
+           log)  filetype=git-log ;;
            status)  filetype=git-status ;;
            *) return 1 ;;
         esac
@@ -84,10 +87,17 @@ define-command -params 1.. \
         mkfifo ${output}
         ( git "$@" > ${output} 2>&1 & ) > /dev/null 2>&1 < /dev/null
 
+        # We need to unmap in case an existing buffer changes type,
+        # for example if the user runs "git show" and "git status".
+        map_diff_goto_source=$([ -n "${map_diff_goto_source}" ] \
+          && printf %s "map buffer normal <ret> %[: git-diff-goto-source<ret>] -docstring 'Jump to source from git diff'" \
+          || printf %s "unmap buffer normal <ret> %[: git-diff-goto-source<ret>]")
+
         printf %s "evaluate-commands -try-client '$kak_opt_docsclient' %{
                   edit! -fifo ${output} *git*
                   set-option buffer filetype '${filetype}'
                   hook -always -once buffer BufCloseFifo .* %{ nop %sh{ rm -r $(dirname ${output}) } }
+                  ${map_diff_goto_source}
               }"
     }
 
@@ -326,3 +336,108 @@ define-command -params 1.. \
             ;;
     esac
 }}
+
+# Options needed by git-diff-goto-source command
+declare-option -hidden str git_diff_hunk_filename
+declare-option -hidden int git_diff_hunk_line_num_start
+declare-option -hidden int git_diff_go_to_line_num
+declare-option -hidden str git_diff_git_dir
+declare-option -hidden str git_diff_section_heading
+declare-option -hidden int git_diff_cursor_column
+
+# Works within :git diff and :git show
+define-command git-diff-goto-source \
+    -docstring 'Navigate to source by pressing the enter key in hunks when git diff is displayed. Works within :git diff and :git show' %{
+    try %{
+        set-option global git_diff_git_dir %sh{
+           git rev-parse --show-toplevel
+        }
+        # We will need this later. Need to subtract 1 because a diff has an initial column
+        # for -,+,<space>
+        set-option global git_diff_cursor_column %sh{ echo $(($kak_cursor_column-1)) }
+
+        # This function works_within a hunk or in the diff header.
+        #   - On a context line or added line, it will navigate to that line.
+        #   - On a deleted line, it will navigate to the context line or added line just above.
+        #   - On a @@ line (i.e. a "hunk header") this will navigate to section heading (see below).
+        #   - A diff header contains lines starting with "diff", "index", "+++", and "---".
+        #     Inside a diff header, this will navigate to the first line of the file.
+        execute-keys -draft 'x<a-k>^[@ +-]|^diff|^index<ret>'
+
+        # Find the source filename for the current hunk (reverse search)
+        evaluate-commands -draft %{
+            # First look for the "diff" line. because "+++" may be part of a diff.
+            execute-keys 'x<semicolon><a-/>^diff<ret></>^\+\+\+ \w([^\n]*)<ret>'
+            set-option global git_diff_hunk_filename %reg{1}
+        }
+
+        try %{
+            # Are we inside the diff header? If so simply go to the first line of the file.
+            # The diff header is everything before the first hunk header.
+            execute-keys -draft 'x<semicolon><a-?>^diff<ret><a-K>^@@<ret>'
+            edit -existing "%opt{git_diff_git_dir}%opt{git_diff_hunk_filename}" 1
+        } catch %{
+            # Find the source line at which the current hunk starts (reverse search)
+            evaluate-commands -draft %{
+                execute-keys 'x<semicolon><a-/>^@@ -\d+(?:,\d+)? \+(\d+)(?:,\d+)? @@<ret>'
+                set-option buffer git_diff_hunk_line_num_start %reg{1}
+            }
+            # If we're already on a hunk header (i.e. a line that starts with @@) then
+            # our behavior changes slightly: we need to go look for the section heading.
+            # For example take this hunk header:  @@ -123,4 +123,4 @@ fn some_function_name_possibly
+            # Here the section heading is "fn some_function_name_possibly". Please note that the section
+            # heading is NOT necessarily at the hunk start line so we can't trivially extract that.
+            try %{
+                # First things first, are we on a hunk header? If not, head to the nearest `catch`
+                execute-keys -draft 'x<a-k>^@@<ret>'
+                evaluate-commands -try-client %opt{jumpclient} %{
+                    # Now attempt to find the section heading!
+                    try %{
+                        # First extract the section heading.
+                        evaluate-commands -draft %{
+                            execute-keys 'xs^@@ -\d+(?:,\d+)? \+(\d+)(?:,\d+)? @@ ([^\n]*)<ret>'
+                            set-option global git_diff_section_heading %reg{2}
+                        }
+                        # Go to the hunk start in the source file. The section header will be above us.
+                        edit -existing "%opt{git_diff_git_dir}%opt{git_diff_hunk_filename}" %opt{git_diff_hunk_line_num_start}
+                        # Search for the raw text of the section, like "fn some_function_name_possibly". That should work most of the time.
+                        set-register / "\Q%opt{git_diff_section_heading}"
+                        # Search backward from where the cursor is now.
+                        # Note that the hunk line number is NOT located at the same place as the section heading.
+                        # After we have found it, adjust the cursor and center the viewport as if we had directly jumped
+                        # to the first character of the section header with and `edit` command.
+                        execute-keys "<a-/><ret><a-semicolon><semicolon>vc"
+                    } catch %{
+                        # There is no section heading, or we can't find it in the source file,
+                        # so just go to the hunk start line.
+                        # NOTE that we DONT go to the saved cursor column,
+                        # because our cursor column will be fixed to the start of the section heading
+                        edit -existing "%opt{git_diff_git_dir}%opt{git_diff_hunk_filename}" %opt{git_diff_hunk_line_num_start}
+                    }
+                }
+           } catch %{
+                # This catch deals with the typical case. We're somewhere on either:
+                # (a) A context line i.e. lines starting with ' '
+                # or (b) On a line removal i.e. lines starting with '-'
+                # or (c) On a line addition i.e. lines starting with '+'
+                # So now try to figure out a line offset + git_diff_hunk_line_num_start that we need to go to
+                # Ignoring any diff lines starting with `-`, how many lines from where we
+                # pressed <ret> till the start of the hunk?
+                evaluate-commands -draft %{
+                   execute-keys '<a-?>^@@<ret>J<a-s><a-K>^-<ret>'
+                   set-option global git_diff_go_to_line_num %sh{
+                       set -- $kak_reg_hash
+                       line=$(($#+$kak_opt_git_diff_hunk_line_num_start-1))
+                       echo $line
+                   }
+                }
+                evaluate-commands -try-client %opt{jumpclient} %{
+                    # Open the source file at the appropriate line number and cursor column
+                    edit -existing "%opt{git_diff_git_dir}%opt{git_diff_hunk_filename}" %opt{git_diff_go_to_line_num} %opt{git_diff_cursor_column}
+                }
+            }
+        }
+    } catch %{
+        fail "git-diff-goto-source: unable to navigate to source. Use only inside a diff"
+    }
+}


sidkshatriya:
This will work with even things like `git diff --staging`, `git diff <some-sha1>`, `git diff HEAD^^`

rc/tools/git.kak
```diff
@@ -326,3 +334,59 @@ define-command -params 1.. \
             ;;
     esac
 }}
+
+# Options needed by git-diff-goto-source command
+declare-option str git_diff_hunk_filename
+declare-option int git_diff_hunk_line_num_start
+declare-option int git_diff_go_to_line_num
+declare-option str git_diff_git_dir
```
occivink, line 342: You should make these options hidden, since they're just implementation details

occivink:
This could also be enabled for `git show` buffers, with the same caveat that invocations with certain arguments (or with local changes) can lead to erroneous jumps. I think that even if the jump is not always perfect, it's useful to have.

sidkshatriya:
@occivink Thank for your review! Just wanted to know if you had the chance to play around with this feature? Did things work as you expected?

occivink:
I only tested briefly so far, and it seemed to work. I did encounter one case where it did not: working within a submodule of another repository. `git rev-parse --absolute-git-dir` returns `/path-to-enclosing-git-dir/.git/modules/submodule-name` which is not what is expected. 
`git rev-parse --show-toplevel` seems to print the directory we want in both cases.

occivink:
Another suggestion: when using `goto-source` on a line starting with `-`, instead of erroring out it could take you to the previous valid line

sidkshatriya:
> Another suggestion: when using `goto-source` on a line starting with `-`, instead of erroring out it could take you to the previous valid line

Yes, this is definitely easy to do. However, I thought about this long and hard earlier: this error has the effect of making the user understand what exactly the hunk corresponds to in the current file. But I'm open to suggestions and can definitely change the behavior to go to the nearest valid line depending on what further opinions we get.

rc/tools/git.kak
sidkshatriya: Done!

krobelus (commented):
Seems to work well, thanks! I used something very similar for a while and it's really handy.
I think we should definitely enable it for show/log buffers, (and also for the diff filetype but that can be done later).

>  this error has the effect of making the user understand what exactly the hunk corresponds to in the current file

Kind of but this binding will still be a heuristic in many cases, so the user should be able to deal with it if it's off by a few lines.
I prefer enabling it for deleted lines and hunk boundaries, for convenience.

---

For the record here's my implementation. I believe it could be simplified a bit with the new command-fifo feature.

 <details>

```
define-command git-diff-goto-source \
    -docstring 'Navigate to source by pressing <ret> in hunks when git diff is displayed' %{
    evaluate-commands -save-regs c %{
        evaluate-commands -draft -save-regs | %{
            set-register c %val{cursor_column}
            execute-keys <a-l><semicolon><a-?>^diff<ret><a-x>
            set-register | %{
                column="$kak_reg_c" perl -wne '
                    use Env;
                    $file = $1 if m{^diff --git a/\S* b/(.*)$} or m{^diff --cc (.*)$};
                    $line = $1 - 1 if m{^@@ -\d+(?:,\d+)? \+(\d+)(?:,\d+)? @@};
                    $line++ if defined $line and m{^[ +]};
                    END {
                        $root = qx(git rev-parse --show-toplevel);
                        chomp $root;
                        $column--; # Account for prefix character from diff.
                        print "edit -existing -- $root/$file $line $column"
                    }
                '
            }
            execute-keys '|<ret>'
            set-register c %val{selection}
            execute-keys u
        }
        evaluate-commands -try-client %opt{jumpclient} %reg{c}
        try %{ focus %opt{jumpclient} }
    }
}
```

</details>

sidkshatriya:
@krobelus @occivink 
I've implemented the following **new** features:
-  By popular demand: Allow user to press `<ret>` on deleted lines (`-`). Navigation will be to the nearest valid line. Approximate but still useful
- Add the ability to navigate to the range line section heading.
    
    e.g. If you press <ret> on anywhere on a range line:
    ```
    @@ -123,4 +123,4 @@ fn some_function {
    ```
    
    The code will try to navigate to the section heading `fn some_function {`
    
    **Note** the section heading is _not_ necessarily located at the range line so we can't trivially extract using a regex. It needs to be searched backwards _from_ range line (in above example range line is 123). **This feature is quite useful.** Please give me feedback. Try this out on git diff, say, on a rust/C++ code base
- Preserve column number during navigation (requested by @krobelus)
- Navigate in `%opt{jumpclient}` (requested by @krobelus)

rc/tools/git.kak
```diff
@@ -326,3 +334,58 @@ define-command -params 1.. \
...
+           execute-keys '<a-?>^@@<ret>J<a-s><a-K>^-<ret>'
+           set-option buffer git_diff_go_to_line_num %sh{
+               set -- $kak_reg_hash
+               line=$(($#+$kak_opt_git_diff_hunk_line_num_start-1))
+               echo $line
+           }
+        }
+
+        # Open the source file at the appropriate line number!
+        edit -existing "%opt{git_diff_git_dir}%opt{git_diff_hunk_filename}" %opt{git_diff_go_to_line_num}
```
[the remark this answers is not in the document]
sidkshatriya: Done!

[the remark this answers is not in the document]
sidkshatriya: Done!

rc/tools/git.kak
```diff
@@ -326,3 +334,58 @@ define-command -params 1.. \
...
 }}
+
+# Options needed by git-diff-goto-source command
+declare-option -hidden str git_diff_hunk_filename
+declare-option -hidden int git_diff_hunk_line_num_start
+declare-option -hidden int git_diff_go_to_line_num
+declare-option -hidden str git_diff_git_dir
+
+define-command git-diff-goto-source \
+    -docstring 'Navigate to source by pressing <ret> in hunks when git diff is displayed' %{
```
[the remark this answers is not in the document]
sidkshatriya: Done!

sidkshatriya:
@krobelus @occivink In addition to the new features mentioned in my comment [above](https://github.com/mawww/kakoune/pull/4263#issuecomment-880437261) here are some additional features I've pushed up.

- You can press `<ret>` on a `+++` line. This navigates to the first line of the file
- Git goto source navigation is now available in `:git show` also

rc/tools/git.kak
```diff
@@ -326,3 +336,117 @@ define-command -params 1.. \
...
+
+        try %{
+            # Are we on a +++ line? If so simply go to the first line of the file
+            execute-keys -draft 'x<a-k>^\+\+\+<ret>'
+            # Goto line 1 as we pressed on +++
+            edit -existing "%opt{git_diff_git_dir}%opt{git_diff_hunk_filename}" 1
+        } catch %{
+            # Find the source line at which the current hunk starts (reverse search)
+            evaluate-commands -draft %{
+                execute-keys 'x<semicolon><a-/>^@@ -\d+,\d+ \+(\d+),\d+ @@ ([^\n]*)<ret>'
```
occivink, line 382: There is not necessarily a 'section heading',  so the space after the second `@@` should be optional. Currently this takes you to an erroneous line

sidkshatriya: @occivink Thanks for the feedback -- I've pushed up a fresh commit. I think it should have dealt with the problem. Let me know if it fixed it!

occivink: Not quite. If there is no section heading he line `execute-keys 'x<semicolon><a-/>^@@ -\d+,\d+ \+(\d+),\d+ @@ ([^\n]*)<ret>'` will reverse-search (which might wrap around) to the previous section heading, which is a different hunk altogether. 
We're already on a range line, so I think the search can be transformed into a `<a-x><a-k>` and then distinguish the two cases: with a section heading and without. Something like that
```
try %{
    # test @@ line
    try %{
        # test @@ line with section heading
    } catch %{
        # no section heading
    }
} catch %{
    # typical case
}
```

sidkshatriya: @occivink Good point about the potential for a wrap around. I did a `xs` to prevent that and made related changes. I'm very sure that I've fixed the issue now -- I've pushed up another commit.

(BTW The code just needs a `try %{ }` for the section header case. We don't need a catch as we already have a default location from the range line to navigate to)

krobelus:
Looks great, he behavior seems fairly complete. The section heading behavior is nice.

Here's my version with the command_fifo approach.
It should be mostly equivalent, but it doesn't have the section heading feature.

```kak
define-command -override git-diff-goto-source \
    -docstring 'Given the cursor inside a diff, navigate to the source' %{
    evaluate-commands -draft -save-regs | %{
        # Select everything from the last "diff" entry up to the cursor. 
        execute-keys <a-l><semicolon><a-?>^diff<ret><a-x>
        set-register | %{
            perl -wne '
                # Compute file and line number of the diff line.
                $file = $1 if m{^diff --git a/\S* b/(.*)$} or m{^diff --cc (.*)$};
                $line = $1 - 1 if m{^@@ -\d+(?:,\d+)? \+(\d+)(?:,\d+)? @@};
                $line++ if defined $line and m{^[ +]};
                END {
                    $root = qx(git rev-parse --show-toplevel);
                    chomp $root;
                    use Env;
                    $column = $kak_cursor_column - 1; # Account for diff prefix.
                    $client = $kak_opt_jumpclient ? $kak_opt_jumpclient : $kak_client;
                    print "evaluate-commands -client $client %{
                        edit -existing -- $root/$file $line $column
                    }"
                }
            ' > "$kak_command_fifo"
        }
        execute-keys '<a-|><ret>'
    }
}
```

Maybe you should squash your commits, since some commits rewrite a whole lot. It will be easier for future readers to read with fewer intermediate states.
I think rebasing on master instead of merging master into your branch would make the history easier to read.
I'd avoid tagging people in commit messages because GitHub sends them a useless notification when such commits are merged to random forks (though that's GitHub's fault of course).

sidkshatriya:
As per suggestion from @krobelus I have squashed all the commits into a single commit and rebased against master.


Here is the single commit message that summarizes the current state of the feature. @mawww I would be ecstatic if you were able to check out the feature and provide any comments/feedback!

```
    When viewing `:git diff` or `:git show` within kakoune, this commit
    adds the ability to press <ret> within a hunk and navigate to the original
    source code. This can be useful because one often needs to go back and forth
    between the diff and the full source code.
    
    - You can press <ret> anywhere _within_ a hunk i.e. lines that start with
      ` `, `+`, `-`. You will be taken to the exact line in the source that corresponds
      to where you pressed <ret> in the hunk. It actually does not make sense
      to press <ret> on a `-` line because that does not exist anymore but
      in that case you are taken to a nearby line in the hope this is still useful.
    
    - You can also press <ret> on a range line (lines that
      look like @@ ... @@). If you press <ret> on anywhere on a range line e.g.
    
      ```
      @@ -120,3 +123,4 @@ fn some_function {
      ```
      The code will try to navigate to the section heading "fn some_function {"
      Note that the section heading is _not_ necessarily located at the
      range line (in the above example the range line is 123).
    
    - You can press <ret> on a +++ line also and you will be taken the first
       line of the file
    
    Caveats:
    - Navigation to the original source file will be accurate only if any edits to
      the original source file have been saved to disk, because otherwise
      they will not be detected by the `:git diff` or `:git show` commands
    - This feature should work well for most typical uses e.g. `:git diff`, `:git diff HEAD^`
      `:git diff <some-sha1>`. In fact this feature should work in all scenarios when
      the *current files* on disk are being compared _with_ some arbitrary git revision/staging.
      It will be less useful in other scenarios when two arbitrary revisions are being
      compared to each other or when you are trying to compare staging to some revision.
      For example when you invoke `:git diff --staged` you are trying to compare staging
      with HEAD but are navigating to what is currently on disk (which may be different
      from staging).
```

sidkshatriya:
Just made a really small change -- changed an option type from a `str` to `int`

rc/tools/git.kak
```diff
@@ -326,3 +337,124 @@ define-command -params 1.. \
...
+                # (a) section heading could not be found for some reason (b) section heading does not exist
+                set-option global git_diff_go_to_line_num %opt{git_diff_hunk_line_num_start}
+
+                try %{
+                    # Now search for the section heading!
+                    #
+                    # We're doing something a bit weird here: we're editing the file in a draft context!
+                    # The reason is that first we open up the file and _then_ we try to navigate to the
+                    # section heading. This adds two locations to the jump list. We ideally should have gone
+                    # to the section heading in one step but didn't know its line number in advance.
```
mawww, line 404: This would indicate a bug in Kakoune, it should collapse multiple jumps done in `evaluate-commands` to a single one.

rc/tools/git.kak
```diff
@@ -326,3 +337,124 @@ define-command -params 1.. \
...
     esac
 }}
+
+# Options needed by git-diff-goto-source command
+declare-option -hidden str git_diff_hunk_filename
+declare-option -hidden int git_diff_hunk_line_num_start
+declare-option -hidden int git_diff_go_to_line_num
+declare-option -hidden str git_diff_git_dir
+declare-option -hidden str git_diff_section_heading
+declare-option -hidden int git_diff_cursor_column
```
mawww, line 347: Not terribly happy about the amount of hidden options necessary to make this work, but not a big deal either.

mawww (commented):
Hey, Sorry for the long delay in reviewing this, it would be a really nice feature.

krobelus (commented):
Here is my updated version with the section header feature.
I think this is the better approach if the logic gets any more complex, but either one seems fine to me.

<details>
```kak
define-command -override git-diff-goto-source \
    -docstring 'Given the cursor inside a diff, navigate to the source' %{
    evaluate-commands -draft -save-regs c| %{
        # Remember the column because we will move the cursor.
	set-register c %val{cursor_column}
        # We'll use the diff line to find the filename, and the hunk headers
        # and the +/- and context lines to compute the line number.
        # Select everything from the last "diff" line up to the line at cursor.
        try %{
            execute-keys <a-l><semicolon><a-?>^diff<ret><a-x>
        } catch %{
            fail 'git-diff-goto-source: failed to find "diff" line above cursor'
        }
        set-register | %{
            client="${kak_opt_jumpclient:-$kak_client}" column="$kak_reg_c" perl -wne '
                # Compute file and line number of the line after application of the diff.
                if (m{^diff --git a/\S* b/(.*)$} or m{^diff --cc (.*)$}) {
                    $file = $1;
                    $line = 0;
                }
                $line = $1 - 1 if m{^@@ -\d+(?:,\d+)? \+(\d+)(?:,\d+)? @@};
                $line++ if defined $line and m{^[ +]};
                $last_line = $_;
                END {
                    $root = qx(git rev-parse --show-toplevel);
                    chomp $root;
                    use Env;
                    $column = $column - 1; # Account for [ +-] diff prefix.

                    # If the cursor was on a hunk header, go to the section header if possible.
                    if ($last_line =~ m/^(@@ -\d+(?:,\d+)? \+\d+(?:,\d+) @@ )([^\n]*)/) {
                        $hunk_header_prefix = $1;
                        $hunk_header_from_userdiff = $2;
                        open FILE, "<", $file or die "Failed to open file: $!: $file";
                        @lines = <FILE>;
                        for (my $i = $line - 1; $i >= 0 && $i < scalar @lines; $i--) {
                            next unless $lines[$i] =~ m/\Q$hunk_header_from_userdiff/;
                            $line = $i + 1;
                            # Re-add 1 because the @@ line does not have a [ +-] diff prefix.
                            $column = $column + 1 - length $hunk_header_prefix;
                            last;
                        }
                    }

                    print "evaluate-commands -client $client %{
                        edit -existing -- $root/$file $line $column
                    }"
                }
            ' > "$kak_command_fifo"
        }
        execute-keys '<a-|><ret>'
    }
}
```

</details>

krobelus:
@sidkshatriya I have addressed review comments in your patch and fixed some other edge cases. WDYT?

```diff
diff --git a/rc/tools/git.kak b/rc/tools/git.kak
index 5b12c7a6..9fae5cae 100644
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -74,12 +74,11 @@ define-command -params 1.. \
 
     show_git_cmd_output() {
         local filetype
-        local apply_goto_line_map
-        local maybe_goto_line_map
+        local map_diff_goto_source
 
         case "$1" in
-           diff) apply_goto_line_map='true'; filetype=diff ;;
-           show) apply_goto_line_map='true'; filetype=git-log ;;
+           diff) map_diff_goto_source=true; filetype=diff ;;
+           show) map_diff_goto_source=true; filetype=git-log ;;
            log)  filetype=git-log ;;
            status)  filetype=git-status ;;
            *) return 1 ;;
@@ -88,17 +87,17 @@ define-command -params 1.. \
         mkfifo ${output}
         ( git "$@" > ${output} 2>&1 & ) > /dev/null 2>&1 < /dev/null
 
-        if [ "$apply_goto_line_map" = 'true' ]; then
-            maybe_goto_source_map="map -docstring 'Jump to source from git diff' buffer normal <ret> %{: git-diff-goto-source<ret>}"
-        else
-            maybe_goto_source_map=''
-        fi
+        # We need to unmap in case an existing buffer changes type,
+        # for example if the user runs "git show" and "git status".
+        map_diff_goto_source=$([ -n "${map_diff_goto_source}" ] \
+          && printf %s "map buffer normal <ret> %[: git-diff-goto-source<ret>] -docstring 'Jump to source from git diff'" \
+          || printf %s "unmap buffer normal <ret> %[: git-diff-goto-source<ret>]")
 
         printf %s "evaluate-commands -try-client '$kak_opt_docsclient' %{
                   edit! -fifo ${output} *git*
                   set-option buffer filetype '${filetype}'
                   hook -always -once buffer BufCloseFifo .* %{ nop %sh{ rm -r $(dirname ${output}) } }
-                  ${maybe_goto_source_map}
+                  ${map_diff_goto_source}
               }"
     }
 
@@ -357,84 +356,68 @@ define-command git-diff-goto-source \
         # for -,+,<space>
         set-option global git_diff_cursor_column %sh{ echo $(($kak_cursor_column-1)) }
 
-        # Allowed to press <ret> only _within_ a hunk
-        # Note: - Pressing <ret> on a @@ line (i.e. a "range line") is allowed.
-        #         This action will result in navigating to section heading (see below).
-        #       - Pressing <ret> on a +++ line is allowed. This action will result in
-        #         navigating to the first line of the file
-        execute-keys -draft 'x<a-K>^\w<ret>'
-        execute-keys -draft 'x<a-K>^$<ret>'
-        execute-keys -draft 'x<a-K>^---<ret>'
+        # This function works_within a hunk or in the diff header.
+        #   - On a context line or added line, it will navigate to that line.
+        #   - On a deleted line, it will navigate to the context line or added line just above.
+        #   - On a @@ line (i.e. a "hunk header") this will navigate to section heading (see below).
+        #   - A diff header contains lines starting with "diff", "index", "+++", and "---".
+        #     Inside a diff header, this will navigate to the first line of the file.
+        execute-keys -draft 'x<a-k>^[@ +-]|^diff|^index<ret>'
 
         # Find the source filename for the current hunk (reverse search)
         evaluate-commands -draft %{
-            execute-keys 'x<semicolon><a-/>^\+\+\+ b([^\n]*)<ret>'
+            # First look for the "diff" line. because "+++" may be part of a diff.
+            execute-keys 'x<semicolon><a-/>^diff<ret></>^\+\+\+ \w([^\n]*)<ret>'
             set-option global git_diff_hunk_filename %reg{1}
         }
 
         try %{
-            # Are we on a +++ line? If so simply go to the first line of the file
-            execute-keys -draft 'x<a-k>^\+\+\+<ret>'
-            # Goto line 1 as we pressed on +++
+            # Are we inside the diff header? If so simply go to the first line of the file.
+            # The diff header is everything before the first hunk header.
+            execute-keys -draft 'x<semicolon><a-?>^diff<ret><a-K>^@@<ret>'
             edit -existing "%opt{git_diff_git_dir}%opt{git_diff_hunk_filename}" 1
         } catch %{
             # Find the source line at which the current hunk starts (reverse search)
             evaluate-commands -draft %{
-                execute-keys 'x<semicolon><a-/>^@@ -\d+,\d+ \+(\d+),\d+ @@<ret>'
+                execute-keys 'x<semicolon><a-/>^@@ -\d+(?:,\d+)? \+(\d+)(?:,\d+)? @@<ret>'
                 set-option buffer git_diff_hunk_line_num_start %reg{1}
             }
-            # If we're already on a range line (i.e. a line that starts with @@) when
-            # <ret> was pressed our behavior slightly alters: we need to go look for the section heading.
-            # For example for range line:  @@ -123,4 +123,4 @@ fn some_function_name_possibly
+            # If we're already on a hunk header (i.e. a line that starts with @@) then
+            # our behavior changes slightly: we need to go look for the section heading.
+            # For example take this hunk header:  @@ -123,4 +123,4 @@ fn some_function_name_possibly
             # Here the section heading is "fn some_function_name_possibly". Please note that the section
-            # heading is NOT at line range line number necessarily so we can't trivially extract that.
+            # heading is NOT necessarily at the hunk start line so we can't trivially extract that.
             try %{
-                # First things first, are we on a range line? If not, head to the nearest `catch`
+                # First things first, are we on a hunk header? If not, head to the nearest `catch`
                 execute-keys -draft 'x<a-k>^@@<ret>'
-                # Sane default. This is where we will go if:
-                # (a) section heading could not be found for some reason (b) section heading does not exist
-                set-option global git_diff_go_to_line_num %opt{git_diff_hunk_line_num_start}
-
-                try %{
-                    # Now search for the section heading!
-                    #
-                    # We're doing something a bit weird here: we're editing the file in a draft context!
-                    # The reason is that first we open up the file and _then_ we try to navigate to the
-                    # section heading. This adds two locations to the jump list. We ideally should have gone
-                    # to the section heading in one step but didn't know its line number in advance.
-                    #
-                    # Once we get to the section heading, save the line number. Then in a *non-draft* context
-                    # navigate again in a single step so that only one entry is saved to the jump list.
-                    evaluate-commands -draft %{
-                        execute-keys 'xs^@@ -\d+,\d+ \+(\d+),\d+ @@ ([^\n]*)<ret>'
-                        set-option global git_diff_section_heading %reg{2}
-
-                        # Open the source file at line number found from the range.
-                        # As discussed above we need to now search for the section heading
-                        edit -existing "%opt{git_diff_git_dir}%opt{git_diff_hunk_filename}" %opt{git_diff_go_to_line_num}
-                        # Try to search for the section heading. If for some reason that fails
-                        # then we automatically fall back to the the currently saved git_diff_go_to_line_num
-                        try %{
-                            # We also need to navigate to the context!
-                            # Note the use of \Q so we don't need to quote special regex characters
-                            set-register / "\Q%opt{git_diff_section_heading}"
-                            # This searches backward from where the cursor is now
-                            # Note that the hunk line number is NOT located at the same place as the section heading
-                            execute-keys "<a-/><ret>"
-
-                            set-option global git_diff_go_to_line_num %val{cursor_line}
+                evaluate-commands -try-client %opt{jumpclient} %{
+                    # Now attempt to find the section heading!
+                    try %{
+                        # First extract the section heading.
+                        evaluate-commands -draft %{
+                            execute-keys 'xs^@@ -\d+(?:,\d+)? \+(\d+)(?:,\d+)? @@ ([^\n]*)<ret>'
+                            set-option global git_diff_section_heading %reg{2}
                         }
+                        # Go to the hunk start in the source file. The section header will be above us.
+                        edit -existing "%opt{git_diff_git_dir}%opt{git_diff_hunk_filename}" %opt{git_diff_hunk_line_num_start}
+                        # Search for the raw text of the section, like "fn some_function_name_possibly". That should work most of the time.
+                        set-register / "\Q%opt{git_diff_section_heading}"
+                        # Search backward from where the cursor is now.
+                        # Note that the hunk line number is NOT located at the same place as the section heading.
+                        # After we have found it, adjust the cursor and center the viewport as if we had directly jumped
+                        # to the first character of the section header with and `edit` command.
+                        execute-keys "<a-/><ret><a-semicolon><semicolon>vc"
+                    } catch %{
+                        # There is no section heading, or we can't find it in the source file,
+                        # so just go to the hunk start line.
+                        # NOTE that we DONT go to the saved cursor column,
+                        # because our cursor column will be fixed to the start of the section heading
+                        edit -existing "%opt{git_diff_git_dir}%opt{git_diff_hunk_filename}" %opt{git_diff_hunk_line_num_start}
                     }
                 }
-                # Whether or not we found the section heading we now have a line to go to
-                evaluate-commands -try-client %opt{jumpclient} %{
-                    # Open the source file at the appropriate line number! NOTE that we DONT go to the saved cursor column
-                    # because our cursor column will be fixed to the start of the section heading
-                    edit -existing "%opt{git_diff_git_dir}%opt{git_diff_hunk_filename}" %opt{git_diff_go_to_line_num}
-                }
            } catch %{
-                # This catch deals with the typical case. We're somewhere within either:
-                # (a) The context area i.e. lines starting with ' '
+                # This catch deals with the typical case. We're somewhere on either:
+                # (a) A context line i.e. lines starting with ' '
                 # or (b) On a line removal i.e. lines starting with '-'
                 # or (c) On a line addition i.e. lines starting with '+'
                 # So now try to figure out a line offset + git_diff_hunk_line_num_start that we need to go to
@@ -455,7 +438,7 @@ define-command git-diff-goto-source \
             }
         }
     } catch %{
-        fail "Unable to navigate to source! You can press the `enter` key only inside a hunk, on a range line (`@@ ... @@`) and on a `+++` line"
+        fail "git-diff-goto-source: unable to navigate to source. Use only inside a diff"
     }
 }
 ```

**Comment by sidkshatriya** [comment]
@krobelus Thanks for being persistent on this. I will check this and get back to your by the weekend.

**Comment by sidkshatriya** [comment]
@mawww @krobelus 
This is a useful feature and I don't think we should let it fall by the wayside. It is very useful to go from the diff to the code and all of the kakoune community should be able to enjoy this feature
- I simply incorporated the diff posted by by krobelus on https://github.com/mawww/kakoune/pull/4263#issuecomment-920391916 where he says:

> I  have addressed review comments in your patch and fixed some other edge cases. WDYT? [...]

- I appreciate the effort Johannes put in this feature and I have credited him with `Co-authored-by: Johannes Altmanninger <aclopte@gmail.com>` on the commit message
- The commit is now rebased off the current kakoune master branch. However, in case you want to see what exactly has been changed since my last commit and the one today you can do:

```bash
$ git diff ec67e67f8efdf2311c769c238620cb0de89c904a..2825ca5f73e3947c408f0eb36468438cbe486275 -- rc/tools/git.kak
```

sidkshatriya:
@krobelus If you're OK with everything mentioned in my above comment, please reply on this thread so that mawww can merge (if he too is satisfied with the code, of course).

sidkshatriya:
BTW The following command invocation might be a more convenient way of visualizing the difference between the original patch and the latest patch (which incorporates @krobelus changes):
```
git difftool ec67e67f8efdf2311c769c238620cb0de89c904a 2825ca5f73e3947c408f0eb36468438cbe486275 rc/tools/git.kak
```

I prefer the `meld` tool but `git difftool` accepts many others.

PR #4264, opened 2021-07-15
Check if anything is staged before allowing `:git commit`

sidkshatriya:
I seem to run into this issue occasionally: 
- Let us assume I've NOT staged anything for commit into `git`
- I then (mistakenly) issue a `:git commit` in kakoune, write a commit message and then `:w`. This will result in a cryptic kakoune error message `Error running hooks for 'BufWritePost' '/some/path/.git/COMMIT_EDITMSG', see *debug* buffer`

According to me I should not be allowed to proceed to the `:git commit` buffer _if_ I have nothing staged. 

This change simply produces a helpful `fail` message
`Nothing to commit! Please stage some changes first e.g. :git add <some-file>`

### DIFF
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -267,6 +267,13 @@ define-command -params 1.. \
 			$@
 		EOF
 
+        # Is there anything to commit?
+        staged_files=$(git diff --name-only)
+        if [ "$staged_files" = '' ]; then
+            echo "fail 'Nothing to commit! Please stage some changes first e.g. :git add <some-file>'"
+            exit
+        fi
+
         # fails, and generate COMMIT_EDITMSG
         GIT_EDITOR='' EDITOR='' git commit "$@" > /dev/null 2>&1
         msgfile="$(git rev-parse --git-dir)/COMMIT_EDITMSG"


sidkshatriya:
@occivink @krobelus Thanks for your feedback on #4263 . I was wondering if you could give your opinion on this separate (small) feature. If it does not make sense, I'd like to know why -- perhaps I'm missing something obvious?

rc/tools/git.kak
```diff
@@ -267,6 +267,13 @@ define-command -params 1.. \
 			$@
 		EOF
 
+        # Is there anything to commit?
+        staged_files = $(git diff --name-only)
+        if [ $staged_files = '']; then
```
occivink, line 272: `$staged_files` should be in double-quotes for shell-safety, and there should be a space between `''` and `]`.

occivink:
I don't have much opinion on this, I think it makes sense as the git CLI does not allow it.

This does remove one feature where you can do `git commit`, add files and then finish writing the commit message but this is not possible on the CLI either.

rc/tools/git.kak
sidkshatriya: Thanks -- cleaned it up!

sidkshatriya:
Rebased on top of master and squashed commits to a single one to clean this small feature up in case it is deemed fit to merge.

(Thanks for your review on this earlier @occivink to make this more shell compliant)

rc/tools/git.kak
```diff
@@ -267,6 +267,13 @@ define-command -params 1.. \
 			$@
 		EOF
 
+        # Is there anything to commit?
+        staged_files=$(git diff --name-only)
+        if [ "$staged_files" = '' ]; then
```
mawww, line 272: Wouldnt this break `:git commit <explicit file list>` ?

krobelus:
In additionto the two edgecases mentioned, `git commit` can also fail in different ways, for example if a hook fails.
The present error message already points to the `*debug*` buffer, so how about we just forward the error from git-commit there? It looks like that doesn't happen automatically in the hook context.

```diff
diff --git a/rc/tools/git.kak b/rc/tools/git.kak
index 9fae5cae..9ac11c5c 100644
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -282,9 +282,10 @@ define-command -params 1.. \
         msgfile="$(git rev-parse --git-dir)/COMMIT_EDITMSG"
         printf %s "edit '$msgfile'
               hook buffer BufWritePost '.*\Q$msgfile\E' %{ evaluate-commands %sh{
-                  if git commit -F '$msgfile' --cleanup=strip $* > /dev/null; then
+                  if stderr=\$(git commit -F '$msgfile' --cleanup=strip $* 2>&1 > /dev/null); then
                      printf %s 'evaluate-commands -client $kak_client echo -markup %{{Information}Commit succeeded}; delete-buffer'
                   else
+                     printf \"echo -debug -- '%s'\n\" \"\$(printf %s \"\$stderr\" | sed s,\',\'\',g)\"
                      printf 'evaluate-commands -client %s fail Commit failed\n' "$kak_client"
                   fi
               } }"
```

PR #4265, opened 2021-07-15
Implement customizable terminal title

### LINKED ISSUE #2217: Separate ncurses title from mode line

Gskartwii:
This pull request revives #3829 which in turn revived #2249. It fixes #2217.

The `terminal_set_title` key-value-pair in the `ui_options` option is expanded to allow values beyond booleans. When a non-boolean value is given in the option, that value becomes the terminal title. The documentation is updated to reflect this.

### DIFF
--- a/src/terminal_ui.cc
+++ b/src/terminal_ui.cc
@@ -611,13 +611,22 @@ void TerminalUI::draw_status(const DisplayLine& status_line,
         constexpr char suffix[] = " - Kakoune\007";
         writer.write("\033]2;");
         // Fill title escape sequence buffer, removing non ascii characters
-        for (auto& atom : mode_line)
+        if (m_title_line.empty())
         {
-            const auto str = atom.content();
-            for (auto it = str.begin(), end = str.end(); it != end; utf8::to_next(it, end))
-                writer.write((*it >= 0x20 and *it <= 0x7e) ? *it : '?');
+            for (auto& atom : mode_line)
+            {
+                const auto str = atom.content();
+                for (auto it = str.begin(), end = str.end(); it != end; utf8::to_next(it, end))
+                    writer.write((*it >= 0x20 and *it <= 0x7e) ? *it : '?');
+            }
+            writer.write(suffix);
         }
-        writer.write(suffix);
+        else
+        {
+            for (auto it = m_title_line.begin(), end = m_title_line.end(); it != end; utf8::to_next(it, end))
+                writer.write((*it >= 0x20 and *it <= 0x7e) ? *it : '?');
+            writer.write("\007");
+        }
     }
 
     m_dirty = true;
@@ -1518,7 +1527,17 @@ void TerminalUI::set_ui_options(const Options& options)
     auto to_bool = [](StringView s) { return s == "yes" or s == "true"; };
 
     m_status_on_top = find("terminal_status_on_top").map(to_bool).value_or(false);
-    m_set_title = find("terminal_set_title").map(to_bool).value_or(true);
+
+    auto title_config = find("terminal_set_title").value_or("yes");
+    if (title_config != "yes" and title_config != "no" and title_config != "true" and title_config != "false")
+    {
+        m_set_title = true;
+        m_title_line = String(title_config);
+    }
+    else {
+        m_set_title = to_bool(title_config);
+        m_title_line = "";
+    }
 
     auto synchronized = find("terminal_synchronized").map(to_bool);
     m_synchronized.set = (bool)synchronized;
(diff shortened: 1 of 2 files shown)


Gskartwii:
There is a limitation in that the str-to-str-map doesn't allow the terminal_set_title option to have any spaces in it (without a nbsp or similar Unicode tricks). Would it be better to introduce a new option for this, distinct from ui_options?

src/terminal_ui.cc
```diff
@@ -555,16 +555,29 @@ void TerminalUI::draw_status(const DisplayLine& status_line,
...
+
+        if (not m_title_line.empty())
         {
-            const auto str = atom.content();
-            for (auto it = str.begin(), end = str.end(); it != end; utf8::to_next(it, end))
-                writer.write((*it >= 0x20 and *it <= 0x7e) ? *it : '?');
+            for (auto ch: m_title_line)
+            {
+                writer.write(ch);
+            }
```
mawww, line 565: writing arbitrary bytes is dangerous, that string could for example contain a `\007` character that would end the sequence. The filtering of all non basic ascii characters is here to avoid this kind of issues and should be preserved.

src/terminal_ui.cc
```diff
@@ -1410,7 +1423,18 @@ void TerminalUI::set_ui_options(const Options& options)
     auto to_bool = [](StringView s) { return s == "yes" or s == "true"; };
 
     m_status_on_top = find("terminal_status_on_top").map(to_bool).value_or(false);
-    m_set_title = find("terminal_set_title").map(to_bool).value_or(true);
+
+    auto title_config = find("terminal_set_title").value_or("yes");
+    if (title_config != "yes" && title_config != "no" && title_config != "true" && title_config != "false")
```
mawww, line 1428: In the Kakoune codebase we use the alternate tokens for logical operator (`or`, `and` instead of `||`, `&&`)

mawww (commented):
Sorry for the delay in looking at this, I guess it makes sense to support arbitrary titles although it still feels a bit like feature creep. I'd be keen to get a bit of feedback to see who would actually want to use this features.

Gskartwii:
Hi, no problem about the delay! I have been quite preoccupied too. I've resolved the code style issue, and I re-implemented the filtering too (I wonder if it could be more lenient given UTF-8, since in that encoding, control characters should be easy to filter out).

As for the feature creep issue, I agree it would be good to collect some feedback. For me, the presence of previous PRs for this feature indicates at least some interest, and the feature seems to be impossible to achieve without at least some help from the editor. Of course, a more orthogonal feature idea might be possible.

danr:
I'm strongly in favor of the terminal title being customizable. As you know, the default title copies the modeline, and I put the cursor position in the modeline and I think many others do too. Having the (frequently updated) position reflected in the title too introduces an unneccessary distraction. Either we make the default title less invasive, or we simply accept this PR. Let's :)

src/terminal_ui.cc
```diff
@@ -1477,7 +1486,17 @@ void TerminalUI::set_ui_options(const Options& options)
...
+
+    auto title_config = find("terminal_set_title").value_or("yes");
+    if (title_config != "yes" and title_config != "no" and title_config != "true" and title_config != "false")
+    {
+        m_set_title = true;
+        m_title_line = String(title_config);
+    }
+    else {
+        m_set_title = to_bool(title_config);
+	    m_title_line = "";
```
danr, line 1498: nitpick: Indentation is off by one character here

Gskartwii: Fixed that, thanks!

ricochet1k:
> The documentation is updated to reflect this.

There are no documentation changes in this PR.

ryaninvents:
To provide a bit of feedback: I'm using kakoune inside tmux, and it would be nice to have the tmux title be simply the filename instead of the entire modeline. Right now my tmux `status-right` is essentially `${pane_title} ${date} ${time}`, but when I open kakoune, the date and time get run off of the side of the screen.

Lunarmagpie:
Hello,
I would like this feature to be part of my workflow. I use sway tabs to organize my files. At the moment they show the entire status bar which makes it hard for me to focus on what I want to read. If I could only show the title and cursor position I would find the tabs to be much more readable.

Gskartwii:
I have a similar use case to that which has been mentioned above. The ideal terminal title for me would be `~/projects/foo/bar.cpp - Kakoune`, while still displaying more information on the statusline.

mawww:
Thanks for feedback on this and for opening that pull request, ended up implementing it slightly differently.

PR #4268, opened 2021-07-17
make pascal filetype compatible with bash

listentolist:
When using bash, the whitespaces before the delimiting identifier `EOF`
cause an error as well as a missing `\` before `$`.


Ordoviz (approved):
I can confirm that the changes are necessary for Bash and that Dash still works with them. I would prefer to revert 3ab2b98ec38018159f323919cae0a344e71046dd instead of removing the whitespace, though.

Issue #4148, opened 2021-04-15
Kitty integration (both REPL and terminal) aren't working


egolep:
**Steps** 
- open Kakoune on Kitty
- use kitty-repl or kitty-terminal (in my case I need  to use it for a Julia REPL, but no command is actually working)
- 
**Outcome**
Nothing happens, and no output in *debug* buffer either

**Expected**
A REPL window or a terminal window respectively to open and be usable (n particular, I need to use send-text with the REPL one)

OS: Manjaro
Kak version: 2020.09.01
Kitty version: 0.19.3
WM: i3-gaps 4.19.1

basbebe:
Hi @egolep!

Do you have `allow_remote_control` set in `kitty.conf`?

Do you have `listen_on` set or do you use `--listen-on=[…]` at startup of kitty?

`:kitty-terminal bash` works for me.

egolep:
Thanks @basbebe,
after your comment I started to dig because I actually had set the two variables you said and I found out that kitty was reading another .conf file then the one it was supposed to.
Now it actually works.
Then, my question shifts to the x11 integration: I'm able to open a terminal or a REPL using it, but the x11-send-text does nothing.

Screwtapello:
The X11 REPL plugin works by:

- using `xclip` or `xsel` to copy the selection to the clipboard
- using `xdotool` to switch the keyboard focus to the REPL terminal
- using `xdotool` to send a synthetic Shift-Insert keystroke to the terminal, hoping it will paste from the clipboard

Any or all of these steps can fail for a variety of reasons - you might not have the relevant helper tools installed, your window manager might prevent applications from forcibly switching the keyboard focus, your terminal might not interpret Shift-Insert as a paste command, or it might paste from a different clipboard than the one Kakoune copies to.

If you can't get the X11 REPL working, you might try:

- running inside tmux, so you can use the tmux version of the REPL plugin instead
- use the third-party [xmux plugin](https://github.com/forbesmyester/xmux.kak), which launches the REPL process inside tmux inside a separate X11 terminal, so it's still running in a terminal but Kakoune can send text more reliably
- use my own [kakoune-repl-buffer plugin](https://gitlab.com/Screwtapello/kakoune-repl-buffer), which displays the REPL output in a Kakoune buffer, so sending text is reliable, and it's easy to copy/paste output back into Kakoune

egolep:
Hi @Screwtapello,
thanks for your suggestions.
[...]

PR #4272, opened 2021-07-20
add module dtach-repl

listentolist:
There are still some issues that the `x11-send-text` does not work with some window managers and terminal emulators. It seems that using xdotool for sending text to a REPL is not too reliable (https://github.com/mawww/kakoune/issues/4148#issuecomment-824882799). Using dtach for REPL interaction should solve these issues. One drawback is that it introduces a new dependency but on the other hand xsel is not needed any more.

### DIFF
--- a/rc/windowing/repl/dtach.kak
+++ b/rc/windowing/repl/dtach.kak
@@ -1,0 +1,38 @@
+provide-module dtach-repl %{
+
+# test if dtach is installed
+evaluate-commands %sh{
+    [ -n "$(command -v dtach)" ] || echo 'fail dtach not found'
+}
+
+declare-option -docstring "id of the REPL" str dtach_repl_id
+
+define-command -docstring %{
+    dtach-repl [<arguments>]: create a new terminal window for repl interaction
+    All optional parameters are forwarded to the new terminal window
+} \
+    -params .. \
+    -shell-completion \
+    dtach-repl %{ terminal sh -c %{
+        file="$(mktemp -u -t kak_dtach_repl.XXXXX)"
+        trap 'rm -f "${file}"' EXIT
+        printf "evaluate-commands -try-client $1 \
+            'set-option current dtach_repl_id ${file}'" | kak -p "$2"
+        shift 2
+        dtach -c "${file}" -E sh -c "${@:-$SHELL}" || "${@:-$SHELL}"
+    } -- %val{client} %val{session} %arg{@}
+}
+
+define-command dtach-send-text -params 0..1 -docstring %{
+        dtach-send-text [text]: Send text to the REPL.
+        If no text is passed, then the selection is used
+        } %{
+    nop %sh{
+        printf "%s" "${@:-$kak_selection}" | dtach -p "$kak_opt_dtach_repl_id"
+    }
+}
+
+alias global repl-new dtach-repl
+alias global repl-send-text dtach-send-text
+
+}


Screwtapello:
I have two concerns.

The first concern is that it's not obvious what tools you need to have installed for the X11 REPL feature to work. That's equally a problem with the existing code, but if we're going to change anything it would be nice to improve this aspect. For example, adding a `repl-x11.asciidoc` file beside `repl/x11.kak` that mentions the required tools and links the official homepage for them, or (at the very least) doing something like:

    fail "Required tool XYZ not found"

...so the error message shows up on the status line and gets logged to the `*debug*` buffer.

The second concern is that `dtach` is obscure. It's impossible to know how widely available it is among Kakoune's users (without surveying them) but Debian's [Popularity Contest statistics](https://popcon.debian.org/) suggest that of the ~200k Debian users who submit statistics, [4,160 have `dtach` installed, and 331 use it regularly](https://qa.debian.org/popcon.php?package=dtach) while [8,707 have `xsel` installed, and 1,907 use it regularly](https://qa.debian.org/popcon.php?package=xsel).

As a more relevant comparison, [xmux](https://github.com/forbesmyester/xmux.kak) is a Kakoune REPL plugin that is similar to this, but uses tmux to host the REPL instead of `dtach`. According to popcon, [28,396 people have `tmux` installed, and 7,104 use it regularly](https://qa.debian.org/popcon.php?package=tmux), so tmux would be an even more convenient dependency than `dtach` and `xsel` combined.

ifreund:
`dtach` is a *far* simpler and lighter weight dependency than tmux. This approach looks potentially very useful on wayland where changing the users's focus for them and sending arbitrary keystrokes isn't possible in a portable way (at least not without wide adoption of some ass-backwards protocol extension that ruins one of the nicer things about wayland). It also seems like much less of a hack on X11 IMO.

ifreund:
Regarding package availability for dtach: according to repology it is more widely packaged than kakoune so I don't see that as much of an issue. It's also near-trivial to package

https://repology.org/project/dtach/versions
https://repology.org/project/kakoune/versions

occivink:
Adding this as a separate `repl/dtach.kak` seems like it would be the best approach.

listentolist:
As @ifreund said, dtach is smaller and simpler than tmux and it stays more out of the way. I always had colour issues when running tmux without configuring it. With dtach I hardly even notice that it is running.
I like @occivink 's idea. Having a separate `repl/dtach.kak` makes more sense because it isn't specific to X11.

mawww:
Agreed, this should be a separate repl implementation, it should rely on `terminal` alias instead of the `x11-terminal`command to support various windowing system.

listentolist:
Now it is a separate module and it doesn't change x11-repl.

rc/windowing/repl/dtach.kak
```diff
@@ -0,0 +1,38 @@
...
+        dtach -c "${file}" -E sh -c "${@:-$SHELL}" || "${@:-$SHELL}"
+    } -- %val{client} %val{session} %arg{@}
+}
+
+define-command dtach-send-text -params 0..1 -docstring %{
+        dtach-send-text [text]: Send text to the REPL.
+        If no text is passed, then the selection is used
+        } %{
+    nop %sh{
+        printf "%s" "${@:-$kak_selection}" | dtach -p "$kak_opt_dtach_repl_id"
```
mawww, line 31: If I remember correctly how dtach works, this will pass through any escape sequences. I am not sure what behaviour we have with the other repl implementations, nor if its worth trying to fix that.

PR #4279, opened 2021-07-26
Recognize ".rockspec" file extension as a Lua file

hugomg:
rockspec files are used by [Luarocks](https://luarocks.org/), the most prominent package manager for Lua. Despite the different file extension, these files are actually Lua files and should be syntax highlighted as such.

For what it is worth, Neovim also does the same thing that I am doing in this commit. They recognize both ".lua" and ".rockspec" as being Lua files (and no other extensions, as far as I know).

### DIFF
--- a/rc/filetype/lua.kak
+++ b/rc/filetype/lua.kak
@@ -4,7 +4,7 @@
 # Detection
 # ‾‾‾‾‾‾‾‾‾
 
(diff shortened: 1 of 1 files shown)


hugomg:
In case that matters, my other pull request (#4261) includes a copyright waiver statement.

TeddyDD:
#3966

PR #4280, opened 2021-07-28
Some improvements to Lua syntax highlighting

hugomg:
In this PR, I correct a couple of mistakes in the Lua syntax highlighting. Each point is in a separate commit.

1. When returning a table literal, highlight the `return` as a keyword. Currently is is highlighted as a function call.

```lua
return { x = 10 }
```

2. Highlight the `local` keyword using the same color as the other keywords. Currently it is highlighted as an "attribute".

```lua
local x = 12
```

3. The special environment variable is called `_ENV`, not `_E`.

```lua
_ENV = { print = print }
```

4. Highlight variable attributes. This is a new language feature, introduced in Lua 5.4.

```lua
local y <const> = 42
```

---------------

5. Highlight `and` and `or` as operators instead of as regular keywords

```lua
while x and y do
end
```

### DIFF
--- a/rc/filetype/lua.kak
+++ b/rc/filetype/lua.kak
@@ -48,12 +48,11 @@ add-highlighter shared/lua/comment       region '--'  $                 fill com
 
 add-highlighter shared/lua/code/function_declaration regex \b(?:function\h+)(?:\w+\h*\.\h*)*([a-zA-Z_]\w*)\( 1:function
 add-highlighter shared/lua/code/function_call regex \b([a-zA-Z_]\w*)\h*(?=[\(\{]) 1:function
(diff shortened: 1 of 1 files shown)


raiguard:
Ah, I was going to make a similar PR. There are a few things I've noticed:
- `function` literals are colored as function names instead of keywords
image
(`function` on the top line should be purple)
- `and` and `or` are colored as keywords or functions depending on the context, instead of as operators
image
(`and` and `or` should be red)

hugomg:
I liked that idea of highlighting `and` and `or` as operators. There's now another commit to implement that.

That issue with the anonymous function should already be fixed, by the same commit that fixed the `return {`.

raiguard:
Thank you! That saves me the trouble of making my own PR. :D

hugomg:
For reference, I took some screenshots of the before and after, in the solarized-dark colorscheme

**BEFORE**
colors-before

**AFTER**
colors-after

PR #4283, opened 2021-07-31
README: Add Nix installation instructions

waldyrious:
What the title says :)


eraserhd (approved):
Nix user here.  This is good.  @mawww

PR #4305, opened 2021-08-20
Make sure the ParsedRegex structure has the right size.

ppentchev:
Hi,

Thanks a lot for your work on kakoune!

While working on the Debian package, I noticed that 2020.09.01 failed to build on the m68k processor. It turns out that the definition of the ParsedRegex structure does not necessarily guarantee that it will be packed and padded up to 16 bytes. What yo you think about this change that ensures that?

Thanks for your time, and keep up the great work!

G'luck,
Peter

### DIFF
--- a/src/regex_impl.cc
+++ b/src/regex_impl.cc
@@ -81,7 +81,8 @@ struct ParsedRegex
         NodeIndex children_end;
         Codepoint value;
         Quantifier quantifier;
-    };
+        uint16_t filler = 0;
+    } __attribute__((packed));
     static_assert(sizeof(Node) == 16, "");
 
     Vector<Node, MemoryDomain::Regex> nodes;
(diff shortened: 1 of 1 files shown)


src/regex_impl.cc
```diff
@@ -81,7 +89,8 @@ struct ParsedRegex
         NodeIndex children_end;
         Codepoint value;
         Quantifier quantifier;
-    };
+        uint16_t filler = 0;
+    } __packed;
```
mawww, line 93: `[[gnu::packed]]` should do the trick and does not necessitate the conditional macros as unrecognized attributes must be ignored.

ppentchev: Thanks! I must admit I may have not quite kept up-to-speed with the C++ standards in the last, hm, twenty years... thanks for pointing the attribute syntax out to me!

I just pushed another commit to the branch that minimizes the changes. Let me know if you'd like me to squash/force-push a single commit instead.

PR #1144, opened 2017-01-19
[WIP] Tutorial proposal

lisael:
Here's a proposal to address #1122. This adds a `:tutorial` command that opens a tutorial where one can navigate page by page.

There's a mechanism to chose the tutorial language, a draft of a few lessons, added as simple .md file.

I show it now to validate the design, not the content (there must be a lot of gramar and syntax oddities as, I'm not a native English speaker).

I think, if the design is validated, we should resolve wether the tutorial must remain in the main kakoune repo or live in its own repository to avoid poluting the issues with ToC choices, wording, translations and such.

Thank you for the feedback.

PR #4312, opened 2021-08-25
possibly resolve #1122

### LINKED ISSUE #1122: Add a tutorial like `vimtutor`
As mentioned in https://github.com/mawww/kakoune/issues/249

MaxCan-Code:
Direct people to a place they can onboard


contrib/TRAMPOLINE
```diff
@@ -9,6 +9,11 @@
                   Mawww's experiment for a better code editor
 +=----------------------------------------------------------------------------=+
 
+To view this inside Kakoune, map it to some key like so:
```
mawww, line 12: I am not sure this is the best place to put that advice, and in general would avoid suggesting to pull files from the web.

MaxCan-Code: > I am not sure this is the best place to put that advice,

What would be a better place to put this? If it would resolve #1122

> and in general would avoid suggesting to pull files from the web.

What's wrong with pulling files from the web? If downloading repeatedly is a problem the user can save the `TRAMPOLINE` file somewhere

Screwtapello:
I'm not mawww, but:

- I'm not sure this will resolve #1122, since this isn't exactly a tutorial like `vimtutor` is, even though it's the closest thing Kakoune has
- putting instructions for how to obtain the tutorial inside the tutorial itself doesn't make it easier to find
- Telling people to set up a button that automatically downloads a blob from the internet and pipes it into Kakoune is *probably* safe, but it feels pretty sketchy and a bad idea.

I would imagine the easiest path to a `vimtutor`-like experience would be something like:

- move `contrib/TRAMPOLINE` to something like `rc/tools/tutorial.txt`
- add a `rc/tools/tutorial.kak` script that defines a `:tutorial` command which pipes the `tutorial.txt` file into a scratch buffer
- Add a mention of the `:tutorial` command in the online documentation (the thing you see when you type `:help<ret>`)
- Add a mention of the `:tutorial` command (and maybe `:doc` too) in the default scratch buffer text
- Add a mention of the `:tutorial` command (possibly in the form `kak -e tutorial`) to the manual page
- rewrite the tutorial to be an actual tutorial (like vimtutor, where it gives you sample exercises for each command instead of just describing them)

MaxCan-Code:
@Screwtapello good ideas, there's a tutorial in the works in #1144

contrib/TRAMPOLINE
krobelus: > What would be a better place to put this?

As @Screwtapello said, move it somewhere inside `rc` or `doc`, so it will be installed by default

Issue #4307, opened 2021-08-21
Cursor not visible on blank lines with Plain colorscheme


jpe90:
Hi,

I'm not sure if this is a bug or a quirky interaction between my terminals and kakoune, but when I place my cursor on a blank line with the "Plain" colorscheme, it disappears completely. This is happening both in Kitty and Alacritty terminals on X11 and Wayland. 

I figured I could workaround the problem by modifying one of the values [here](https://github.com/mawww/kakoune/blob/master/colors/plain.kak) while in the Plain colorscheme, but I wasn't quite sure how or which one to change.

I love this program by the way!! And the colorscheme is otherwise fantastic! Thanks in advance.

Below  is a screenshot with two Kakoune windows, one with the Plain colorscheme and the other with Tomorrow Night, with the cursor in the same position (blank line on line 13)

![20210821_16h05m52s_grim](https://user-images.githubusercontent.com/9307830/130333759-301f67b6-4b91-450e-82e0-aff76fb90124.png)

caksoylar:
The cursor is highlighted with `PrimaryCursorEol` face for empty lines, since the only character the cursor can sit on on those lines is the end-of-line character. The problem is that `plain.kak` uses the `default` colors for those:
```kak
face global PrimaryCursorEol default
face global SecondaryCursorEol default
```
However `Default` face also has the same colors, which is just your terminal background and foreground, so EOL characters make the cursor disappear. `Eol` faces have to have a different background defined to be visible.

I don't know what the best solution to fix this would be given such a limited palette I imagine it would be hard to be able to make everything distinguishable (but fixing at least the EOL cursor should be doable). For instance I usually aim to have all of `(Primary|Secondary)Cursor(Eol)?` distinct from each other, since otherwise multiple selections also become very hard to work with. I would also be curious to hear if anyone actually uses `plain.kak` day to day.

jpe90:
Closing issue as a fix has been merged, thank you @mawww  :^)

PR #4315, opened 2021-08-28
Usability changes to plain colorscheme

jpe90:
This commit aims to address two usability issues when using the plain colorscheme. 

**1) Moving over blank lines renders the cursor invisible:**

kak-plain-cursor

**Fix:** highlight the `PrimaryCursorEol` character to make it visible:

kak-fix-cursor

**2) It is visually difficult to distinguish autocomplete text from surrounding program text, and selections are hard to see:**

kak-plain-autocomplete

**Fix:** Change the autocomplete (and accompanying information text) to blue to contrast with regular program text. Invert colors for selected items.

kak-fix-autocomplete

Addresses #4307.

### DIFF
--- a/colors/plain.kak
+++ b/colors/plain.kak
@@ -1,3 +1,4 @@
+
 # Kakoune simple colors, mostly default
 
 # For default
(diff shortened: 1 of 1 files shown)


colors/plain.kak
```diff
@@ -46,3 +47,4 @@ face global StatusCursor default+r
 face global Prompt default
 face global MatchingChar default
 face global BufferPadding default
+ace global BufferPadding default
```
mawww, line 50: I think this is a typo, fixed in the merged version.

PR #4327, opened 2021-09-02
Fix unterminated string in rc/filetype/php.kak

mujo-hash:
Edit a php file and insert a new line to see messages in *debug* buffer regarding unterminated string in rc/filetype/php.kak

This commit fixes that issue, just terminate the %{ ... } string.


mujo-hash:
Failed linux_gcc test seems unrelated to this PR.

Issue #3369, opened 2020-02-14
Add a new "Warning" face


Screwtapello:
Currently, Kakoune has two faces for giving messages to the user:

- `Error` is used for things that have absolutely gone wrong and need attention, like runtime errors, or the `:fail` command.
- `Information` is used for extra information the user might want to refer to, but can be safely ignored, like the docstrings shown while typing a command.

For consistency's sake, these faces are ([or should be](https://github.com/mawww/kakoune/issues/3155)) used by plugins for similar purposes. For example, `spell.kak` uses `Error` to highlight spelling errors, and PR#3347 makes `lint.kak` use `Error` to flag errors reported by the linter.

However, many tools that report errors have at least *three* levels of diagnostic:

- errors are absolutely problems that need to be fixed
- warnings might be problems in certain circumstances the tool can't determine on its own
- other information, often used to point to other places in the code relevant to the previously-emitted error or warning (for example, a "duplicate definition" error might be followed by a "previous definition was here" info diagnostic)

If we use the `Error` face to represent errors, and the `Information` face to represent informational diagnostics, what do we do for warning messages? If we display them as errors, we risk inducing error-fatigue, but if we display them as informational, we're hiding potentially useful information. Also, while the default appearance of `Information` is striking black-on-yellow (appropriate for warnings) many colour schemes make that face a *lot* more subdued, making it much less appropriate for warnings.

I recommend:

- adding a new Warning face, defaulting to black-on-magenta (if the default colour scheme must be compatible with 8-color terminals) or black-on-bright-red (if we can assume a 16-color terminal)
- Teach `lint.kak` and any similar script to use it where necessary
- Add a sensible definition of the face to each colour scheme that ships with Kakoune

mawww:
I agree, and I'll happily merge a PR that implements that.

Delapouite:
A related demand on kak-lsp: https://github.com/ul/kak-lsp/issues/317

evanrelf:
> A related demand on kak-lsp: ul/kak-lsp#317

_Demand_ is a bit strong 🙂

PR #4330, opened 2021-09-07
Add standard DiagnosticError and DiagnosticWarning faces.

Screwtapello:
kak-lsp uses these faces to mark errors inside the buffer, instead of the Error face which is much more jarring, and which does not have an associated warning face. Since the :spell command marks errors inside the buffer, it's also updated
to use this new face.

Adding these faces to Kakoune makes it more likely that colorschemes will automatically do the right thing when used with kak-lsp, and makes it possible to use a subtle appearance (like curly underlines) for in-buffer errors while keeping Kakoune errors bold and jarring as they should be.

### DIFF
--- a/colors/base16.kak
+++ b/colors/base16.kak
@@ -60,6 +60,8 @@ evaluate-commands %sh{
         face global MenuInfo ${cyan_light}
         face global Information ${black_light},${cyan_light}
         face global Error ${grey_light},${magenta_light}
+        face global DiagnosticError ${magenta_light}
+        face global DiagnosticWarning ${cyan_light}
         face global StatusLine ${grey_light},${black_lighterer}
         face global StatusLineMode ${orange_dark}
         face global StatusLineInfo ${cyan_light}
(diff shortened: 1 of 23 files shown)


krobelus:
This is nice, it makes colorschemes work better with kak-lsp out-of-the box.

For example with the base16 colorscheme before/after
image
image

kak-lsp learned a few new faces recently but by default it should try to
only use builtin Kakoune faces.

`Information` and `Error` are meant to be used in the status line and info
box. Contrast that to `DiagnosticError` and `DiagnosticWarning`, which are
meant to used in-buffer (and maybe in a `flag-lines` highlighter) so they
should be more subtle.

We could probably also use more general names, for example if we used `Warning`,
then plugins could delegate to that using `face DiagnosticWarning Warning`.
Maybe `DiagnosticError` and `DiagnosticInformation` for symmetry
with `Error` and `Information`?

Screwtapello:
Rather than align with `Error` and `Information`, I'd rather add a new `Warning` face: https://github.com/mawww/kakoune/issues/3369

PR #4343, opened 2021-09-11
Better lua indentation

hugomg:
Over the last weeks I've been using Kakoune to program in Lua... I felt that some parts of the automatic indentation could be improved, so here is a PR for it :)

## Indentation

- Indent inside **do** **end**
- Indent inside **repeat** **until**
- Indent after a line ending with `{` or `(`
- Do not indent if the keyword is inside a string or comment
- More accurate un-indentation for the **end** keyword

 ```lua
-- Previously, the second "end" was
-- indented at the same level as the "if"
 while true do
      if false then
      end
      end
```

## Automatic insertion of end keyword

- Also insert after **do**, **else** and **elseif**
- Do not insert if the keyword is inside a string or comment
- Do not insert if the block already contains statements

```lua
-- No longer does this:
-- If we open a new line at the start of an unclosed block, adding
-- an "end" right after the cursor is probably not what we want.
if true then
    @
end
    x = 1
    y = 2
```

### DIFF
--- a/rc/filetype/lua.kak
+++ b/rc/filetype/lua.kak
@@ -92,9 +92,12 @@ define-command -hidden lua-trim-indent %[
 
 define-command -hidden lua-indent-on-char %[
     evaluate-commands -no-hooks -draft -itersel %[
-        # align middle and end structures to start and indent when necessary, elseif is already covered by else
-        try %[ execute-keys -draft <a-x><a-k>^\h*(else)$<ret><a-semicolon><a-?>^\h*(if)<ret>s\A|.\z<ret>)<a-&> ]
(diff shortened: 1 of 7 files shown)


mawww:
This seems to break an existing test case for lua indentation, can you take a look and either update the test or fix the behaviour ?

hugomg:
I hadn't noticed that there were test cases for the indentation. Clever!

> can you take a look and either update the test or fix the behaviour ?

The last two commits do both things. I fixed a a bug that was caught by the test cases and I also added more test cases to document the improvements I made.

mawww:
Excellent, thanks for the updated version, will merge soon.

PR #4345, opened 2021-09-11
doc keys: Explain `<c-s>` better

lenormf:
Not sure about this entire section, the pre-amble talks about “push[ing] the previous selections to the client's jump list” which doesn't make the definition of `<c-s>` better with this commit.

### DIFF
--- a/doc/pages/keys.asciidoc
+++ b/doc/pages/keys.asciidoc
@@ -552,7 +552,8 @@ to skim through the jump list using:
     jump backward
 
 *<c-s>*::
-    save selections
+    create a jump step, but also save the current selection to have it be
+    restored when the step is subsequently cycled through
 
 == Multiple selections
 


doc/pages/keys.asciidoc
```diff
@@ -552,7 +552,8 @@ to skim through the jump list using:
     jump backward
 
 *<c-s>*::
-    save selections
+    create a jump step, but also save the current selection to have it be
```
krobelus, line 555: Instead of the first part I'd rather say `push current selections to the jump list` because "push" is already used in the existing documentation in the paragraph above.

doc/pages/keys.asciidoc
```diff
@@ -552,7 +552,8 @@ to skim through the jump list using:
...
+    restored when the step is subsequently cycled through
```
krobelus, line 556: > but also save the current selection to have it be restored

I think that's also the behavior for other jump list entries.. Kakoune might fail restore all selections if the buffer has been modified but that's also independent of `<c-s>`, right?

PR #4346, opened 2021-09-11
src: Allow `ga` to skip removed buffers

### LINKED ISSUE #1840: ga: goto last buffer fails if previous buffer was deleted
My fuzzy file finder will preview documents as you search. I do this by creating and copying the file contents to a `*preview*` buffer.

After using the file finder, `ga` does not work because the `*preview*` buffer has been deleted. It'd be great if there was a way to tell kakoune the buffer is only temporary or if `ga` could just look at a buffer MRU history list.

lenormf:
This commit prevents `ga` from returning a “no last buffer” error
when the previously displayed buffer was removed.

Since the jumps list keeps track of the order in which buffers were
displayed already, handling arbitrary `delete-buffer`s as well,
cycle through it to implement `ga` instead of storing a pointer.

Note that this commit doesn't take into account buffer flags that
might exclude some buffers from being cycled over by commands.

Fixes #1840


krobelus (commented):
Thanks, I've always wanted that behavior - I think this makes `ga` strictly better

> Note that this commit doesn't take into account buffer flags that
might exclude some buffers from being cycled over by commands.

I prefer that for the time being, I got used to use `ga` with the scratch buffer. But we can change that in future.

src/context.cc
```diff
@@ -197,16 +190,41 @@ void Context::change_buffer(Buffer& buffer)
...
         input_handler().reset_normal_mode();
 
-    change_buffer(m_last_buffer ? *m_last_buffer : BufferManager::instance().get_first_buffer());
+    auto last_buffer = this->last_buffer();
+    change_buffer(last_buffer ? *last_buffer : BufferManager::instance().get_first_buffer());
+}
+
+Buffer* Context::last_buffer() const
+{
+    if (not m_jump_list.size())
```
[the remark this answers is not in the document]
lenormf: Thanks, fixed.

src/context.cc
```diff
@@ -197,16 +190,41 @@ void Context::change_buffer(Buffer& buffer)
...
+        return nullptr;
+
+    const auto next_buffer = std::find_if(
```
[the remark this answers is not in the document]
lenormf: I understand, I think the same way. But then I thought that `JumpList` *is* a vector, and the private data it uses (such as the `m_current` index) is only used by the extra methods that interact with the storage.

Since the intent behind this structure is not to override a container but only to add state to it, I think it's not that bad if it makes its use simpler.

@mawww just drop this commit if needed, I'm fine either way.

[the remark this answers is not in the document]
krobelus: Yeah, probably both variants are fine. A compromise would be to expose a conversion operator (or a getter) returning `const Vector<Buffer*> &`, then this can stay roughly the same, and we can't accidentally modify the jumplist.

src/context.hh
```diff
@@ -28,16 +28,14 @@ struct JumpList
 
     friend bool operator==(const JumpList& lhs, const JumpList& rhs)
     {
-        return lhs.m_jumps == rhs.m_jumps and lhs.m_current == rhs.m_current;
+        return lhs == rhs and lhs.m_current == rhs.m_current;
```
mawww, line 31: This looks like its going to be infinitely recursive, you'd need to cast to the base class or call vector::operator== explicitely, but either way I think it shows inheriting from std::vector might not be that nice here.

src/context.cc
```diff
@@ -197,16 +191,51 @@ void Context::change_buffer(Buffer& buffer)
...
         input_handler().reset_normal_mode();
 
-    change_buffer(m_last_buffer ? *m_last_buffer : BufferManager::instance().get_first_buffer());
+    auto last_buffer = this->last_buffer();
+    change_buffer(last_buffer ? *last_buffer : BufferManager::instance().get_first_buffer());
+}
+
+Buffer* Context::last_buffer() const
+{
+    // FIXME: convert JumpList to a vector and use find_if() with reverse iterators
```
mawww: If you just expose a ConstArrayView from the jump list you can, you could even do `find_if(jump_list.list() | reverse(), <predicate>)` with the toolset exposed in ranges.hh

mawww (changes requested):
I like that this gets rid of the last buffer field and use the jump list instead, I would not want the jump list to inherit from Vector but you can just expose either begin/end iterators or a ConstArrayView to the data.

lenormf:
Rebased on master, dropped the commit that makes `JumpList` inherit its inner storage container, implement `JumpList::get_as_list()` to allow using regular algorithm functions.

src/context.cc
```diff
@@ -197,16 +191,36 @@ void Context::change_buffer(Buffer& buffer)
...
+    const auto jump_list = m_jump_list.get_as_list();
+    if (jump_list.empty())
+        return nullptr;
+
+    auto predicate = [this](const auto& sels) {
+        return &sels.buffer() != &this->buffer();
+    };
+
+    auto next_buffer = std::find_if(jump_list.begin() + (m_jump_list.current_index()-1),
+                                    jump_list.end(), predicate);
```
mawww, line 216: `find_if(jump_list.subrange(m_jump_list.current_index() -1), predicate)` ?

src/context.cc
```diff
@@ -197,16 +191,36 @@ void Context::change_buffer(Buffer& buffer)
...
+    if (next_buffer != jump_list.end())
+        return &next_buffer->buffer();
+
+    auto previous_buffer = std::find_if(jump_list.rbegin() + (jump_list.size() - m_jump_list.current_index()),
+                                        jump_list.rend(), predicate);
```
mawww, line 221: `find_if(jump_list.subrange(0, m_jump_list.current_index() | reverse(), predicate)` ?

mawww (changes requested):
Looks good, a couple nitpick on style.

lenormf:
Code factorised. Quite satisfied with how it looks now!

PR #4352, opened 2021-09-14
src: Make `g` and `v` case sensitive

### LINKED ISSUE #3976: The `g` goto menu should be case sensitive
Hi,

All mappings added to the goto menu (<kbd>g</kbd>) are case insensitive.

I don't think it makes sense, since users could want to declare an alternative mapping to the default ones using <kbd>shift</kbd> (more practical than <kbd>alt</kbd>).

The shift modifier also has a significant meaning in the context of mappings (it generally means "extend"), despite that meaning being superseded by the <kbd>G</kbd> menu.

HTH.

lenormf:
The current implementation allows users to declare mappings in the
`goto` and `view` modes with uppercase letters, but doesn't consider
case to be meaningful.

This quirk is also inconsistent as hitting an unmapped key will quit
the mode menu without complaints, but hitting a key that isn't in it
will work if it has a lowercase mapping equivalent.

Fixes #3976

### DIFF
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -228,7 +228,7 @@ void goto_commands(Context& context, NormalParams params)
             if (not cp or key == Key::Escape)
                 return;
             auto& buffer = context.buffer();
-            switch (to_lower(*cp))
+            switch (*cp)
             {
             case 'g':
             case 'k':
@@ -363,7 +363,7 @@ void view_commands(Context& context, NormalParams params)
 
         const BufferCoord cursor = context.selections().main().cursor();
         Window& window = context.window();
-        switch (to_lower(*cp))
+        switch (*cp)
         {
         case 'v':
         case 'c':


krobelus:
I'd only do this if we add mappings like `GK` in addition to the default `Gk`. As mentioned in the issue, this is friendlier when typing quickly.

lenormf:
As usual, if users want to be able to type <kdb>GK</kdb>, they can add the appropriate mapping to their user configuration. This shortcut of questionable usefulness is not worth the drawbacks documented in the commit message.

mawww:
I am not sure this adds much value, I am happy to eventually break upper case mapping, but I'd do it to gain something in return, here it does not simplify code much, I guess it prevents scripts from depending on a likely to be removed eventually feature. Not sure its worth removing yet.

lenormf:
I know this PR looks like a feature request, but it's a bugfix.

<kbd>g</kbd><kbd>h</kbd> is **not** <kbd>g</kbd><kbd>H</kbd>. Users should be able to map the latter without the editor defaulting to executing the former.

PR #4355, opened 2021-09-15
src: Error out on unmapped key in `g` and `v`

### LINKED ISSUE #4353: Hitting an unmapped key in the `g` goto or `v` view menus is a nop
Related to #3976, #2278

### Steps 

* Open the goto or view menu, with <kbd>g</kbd> or <kbd>v</kbd>
* Hit a key that isn't mapped

### Outcome 

* The menu closes as if the key was valid input

### Expected 

An error is printed.

lenormf:
Fixes #4353.

### DIFF
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -327,6 +327,8 @@ void goto_commands(Context& context, NormalParams params)
                 select_coord<mode>(buffer, *pos, context.selections());
                 break;
             }
+            default:
+                throw runtime_error("key not mapped");
             }
         }, (mode == SelectMode::Extend ? "goto (extend to)" : "goto"),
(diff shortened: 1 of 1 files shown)


src/normal.cc
```diff
@@ -391,6 +393,8 @@ void view_commands(Context& context, NormalParams params)
         case 'l':
             window.scroll( std::max<ColumnCount>(1, count));
             break;
+        default:
+            throw runtime_error("key not mapped");
```
krobelus, line 397: Good idea, thanks for the patch.
I just have one quibble: if I hit a wrong key in lock view mode (`V`),
then the error message will not go away even after I press valid keys. I think it should.

mawww: Would be good to fix yeah, but can merge as-is and fix later.

PR #4358, opened 2021-09-16
rc doc: Hide errors on missing directories

### LINKED ISSUE #4273: Completing `:doc` topics reports an error in the debug buffer when `autoload` doesn't exist
### Steps 

* `:doc ` <kbd>tab</kbd>
* `:b *debug*`

### Outcome 

An error from `find` is printed that indicates that `autoload` doesn't exist.

### Expected 
No error.

lenormf:
Fixes #4273.

### DIFF
--- a/rc/tools/doc.kak
+++ b/rc/tools/doc.kak
@@ -141,7 +141,7 @@ define-command -params 0..2 \
                 "${kak_config}/autoload/" \
                 "${kak_runtime}/doc/" \
                 "${kak_runtime}/rc/" \
-                -type f -name "*.asciidoc" |
+                -type f -name "*.asciidoc" 2>/dev/null |
                 sed 's,.*/,,; s/\.[^.]*$//'
         elif [ "$kak_token_to_complete" -eq 1 ]; then
             page=$(
@@ -149,7 +149,7 @@ define-command -params 0..2 \
                     "${kak_config}/autoload/" \
                     "${kak_runtime}/doc/" \
                     "${kak_runtime}/rc/" \
-                    -type f -name "$1.asciidoc" |
+                    -type f -name "$1.asciidoc" 2>/dev/null |
                     head -1
             )
             if [ -f "${page}" ]; then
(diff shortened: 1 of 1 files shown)


rc/tools/doc.kak
```diff
@@ -136,29 +136,30 @@ define-command -params 1 -hidden doc-render %{
...
 define-command -params 0..2 \
     -shell-script-candidates %{
-        if [ "$kak_token_to_complete" -eq 0 ]; then
-            find -L \
-                "${kak_config}/autoload/" \
-                "${kak_runtime}/doc/" \
-                "${kak_runtime}/rc/" \
-                -type f -name "*.asciidoc" |
-                sed 's,.*/,,; s/\.[^.]*$//'
-        elif [ "$kak_token_to_complete" -eq 1 ]; then
```
krobelus, line 146: So `[` spawns a process and `case` doesn't?
`[` is a builtin so I guess this depends on the (quality of) implementation

krobelus (approved):

rc/tools/doc.kak
lenormf: I know that it's a binary on Busybox, so I suppose it might also be the case (pun intended) for `case`.

It's a cosmetic commit, so @mawww may drop it if needed, but I think it makes the code more readable.

Screwtapello: `[` was originally an external tool (a symlink to `/usr/bin/test`), although it's probably a builtin in most shells you're likely to interact with today.

`case` is a flow-control construct, so it can't be an external tool (at least, not since the Thompson shell was replaced by the Bourne shell and `goto` was no longer an external tool).

occivink: `case` is a shell language construct, so I'm pretty sure it can't be a separate executable. `[` (which is just an alias to `test`) is indeed by default a separate process, but a lot of shells also implement it as builtin, I suppose for performance reasons.

PR #4359, opened 2021-09-16
src: Split implementations of -l and -clear

lenormf:
Output:

```
$ kak -L
test | clients: 2 | buffers: 5 | pwd: /home/fle
```


occivink:
I believe `pwd` means 'print working directory'. Maybe `cwd` for 'current working directory', or better yet just `directory`

lenormf:
The value of this field is `$PWD` so I just lowercase'd it. I should probably change it to simply `directory` though.

PR #4365, opened 2021-09-21
src: Differenciate file removal and access errors

lenormf:


krobelus:
I like using `set global autoreload yes` but I don't want it to automatically delete buffers if the file is deleted.
The reason is that deleting buffers can cause data loss, unlike other external changes which I can always recoup with Kakoune's undo.

So I think we should either have separate options for modification/deletion (perhaps via new BufChangedExternally hook), or only hide deleted buffers.

I'm wondering if we should add a modeline indicator for buffers where the file has been deleted. We already have `[+]` for dirty buffers, so maybe `[!]`. Not sure if that fits since the things are different.

PR #4370, opened 2021-09-21
src: Hint at lines above/below with `number_lines`

### LINKED ISSUE #1573: (re)Hint at lines above/below the current selection on the `number_lines` highlighter
Hi,

A while ago, the `number_lines` highlighter was modified to never show a minus `-` sign in front of the relative line numbers that were above the current selection (#1068). This change has taken away a feature that reduced eye strain, and I couldn't find anything to work around that.

When I scroll around a file and the current selection gets slammed around the window left-right&center and I find the line I want to send the anchor too, I look at the number in the gutter to see what offset I should jump to (with `j` or `k`). In the past I knew I had to use `k` when a minus sign was before the line number or `j` otherwise, but now I need to look at the numbers around that relative line offset and actively think about whether they're greater or lesser than the offset I want to send the line to.

This extra brain effort is annoying, as it sometimes causes some misses (aka brain farts) where the selection would move in the opposite direction because I hit `j` instead of `k`, or the other way around.

What's a way to fix that? Using faces on positive/negative line offsets would make things confusing, and I don't see a good way to pass this option as a flag.

Thanks.

lenormf:
This commit allows users to change the separator used by `number-lines`
for the current line, for the others, the ones before/after.

As handling either arbitrary behaviour would prompt for the creation
of two more parameters of the highlighter, totalling four in total,
the values have been unified being a single `-separators` parameter.

It's not idiomatic of the codebase (it makes the highlighter parse
this value as a tuple, backslash escaping included), but only a single
string needs to be specified by the user for basic highlighting
functionality — the others have default values set automatically.

Examples:

	addhl window/ number-lines -separators |
	addhl window/ number-lines -separators " :|"
	addhl window/ number-lines -separators "→:↓:↑"

Fixes #1573.


lenormf:
Rebased on master.

mawww:
This is cute, but its a bit of an unnecessary feature, `-hlcursor` should already make it easy to see if a line is above/below the cursor, so while this makes it a tad easier I am unconvinced its good enough to warrant adding even more knobs to number-lines. (If you want to jump to a line without having to even know if its above/below, you can already use non-relative line numberings)

PR #4374, opened 2021-09-23
rc markdown: Fix fenced code blocks

### LINKED ISSUE #4351: Markdown syntax highlighting: closing ``` ignored after indented code
When opening the following Markdown file in Kakoune:

    ```
        a
    ```

    b

The closing ``` is not detected, causing `b` to be highlighted like a codeblock.
This seems to happen whenever there is indented code inside a codeblock.

The relevant highlighters from `markdown.kak` are: 

```
add-highlighter shared/markdown/inline/codeblock region -match-capture \
    ^(\h*)```\h* \
    ^(\h*)```\h*$ \
    fill meta
add-highlighter shared/markdown/codeblock region "^( {4}|\t)" "$" fill meta
```

Ordoviz:
The closing ``` in the following example was not detected because the indented code block highlighter was higher up in the hierarchy than the fenced code block highlighter.

    ```
        indented
    ```

Fixes #4351

### DIFF
--- a/rc/filetype/markdown.kak
+++ b/rc/filetype/markdown.kak
@@ -55,16 +55,20 @@ evaluate-commands %sh{
   done
 }
 
-add-highlighter shared/markdown/inline/codeblock region -match-capture \
-    ^(\h*)```\h* \
-    ^(\h*)```\h*$ \
-    fill meta
-add-highlighter shared/markdown/codeblock region "^( {4}|\t)" "$" fill meta
-
 add-highlighter shared/markdown/listblock region ^\h*[-*]\s ^(?=\S) regions
 add-highlighter shared/markdown/listblock/g default-region group
 add-highlighter shared/markdown/listblock/g/ ref markdown/inline
 add-highlighter shared/markdown/listblock/g/marker regex ^\h*([-*])\s 1:bullet
+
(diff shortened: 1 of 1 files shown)


mawww:
This seems to revert part of e0731b70cf04adb728bb6bb7848d2c67aef24860. How does that impact markdown lists highlighting ?

Ordoviz:
All testcases at #4208 and #4196 are still highlighted correctly. I'm not sure why @krobelus  [suggested ](https://github.com/mawww/kakoune/pull/4208#issuecomment-872510907) to make the fenced codeblock highlighter inline.

krobelus:
I kept the inline bit from the original patch of that PR.
I just realized a behavior change here that explains why it exists: on

````markdown
- foo
  ```
  echo asdf
  ```
````

on master, the indented fenced code block is purple (face `meta`)
with this patch, it is green (face `mono`), which is intended for `` `inline code` ``

How about we just define the region twice; once as `shared/markdown/codeblock` to fix #4351, and once as `shared/markdown/inline/codeblock` to keep recognizing fenced code blocks inside lists.
(We define highlighters for ```` ```sh ```` and friends only outside the list region, but they also default to the `meta` face, so that would only increase consistency.)

PR #4382, opened 2021-10-04
Fix OCaml line comments

newsch:
Hi, when trying the new-ish OCaml support I found that the the default line comment `#` was still enabled.
To my knowledge OCaml does not have explicit line comments, and as far as I can tell neither does Coq.

Based on the markdown and html settings it seems like the pattern is to set `comment_line` to `''` which throws an error when `comment-line` is used.


mawww:
Thanks !

PR #4404, opened 2021-10-25
Remove scratch message

alexherbo2:


sidkshatriya:
@eraserhd I would make a small modification to your useful snippet.

Instead of:
```kak
hook global BufCreate '\*scratch(-\d+)?\*' %{ execute-keys '%<a-d>' }
```

I suggest:
```kak
hook global BufCreate '\*scratch(-\d+)?\*' %{ execute-keys -draft '%<a-d>' }
```

Though this happens almost instantaneously, it should happen without explicitly executing in the "live" user interface.

P.S. I am a fan of your kak-ansi plugin! :-)

Issue #4159, opened 2021-04-22
[builtin-terminal-ui] <a-O> not working on kitty


Parasrah:
### Steps 

on `builtin-terminal-ui` branch commit `d5eb98f7fc4c0f065f96bec247504e99d67ffb54`

1) open up kak with default autoload, or with no config `kak -n`, in kitty
2) `<a-O>`

### Outcome 

nothing

### Expected 

should insert line above selection. Seems to also be a problem with all other alt+<capital_letter> combinations.

Parasrah:
tried this on my other computer running `kitty 0.19.3` and it seems to work fine, whereas earlier I was using `0.20.1` I believe

Screwtapello:
If you execute (inside Kakoune):

    :set global debug keys
    :b *debug*

...then press `<a-O>`, what does it display?

Parasrah:
output:
```
Client 'client0' got key '<a-s-o>'
```

And yet I tried `map global normal <a-s-o> ':echo testing<ret>'  to no avail.

PS: thanks for that debug trick, I had no idea that was a thing

Screwtapello:
And I bet if you try the previous commit on the builtin-terminal-ui branch (the one before "Opt-in kitty CSI u support"), it works correctly, right?

In the traditional terminal encoding, the terminal always sends the actual character typed; for example, the encoding for `<a-!>` is literally 0x1b 0x21 (or `ESC !`) without any indication of the Shift key, or what that symbol would be without Shift. As a result, Kakoune doesn't support the shift modifier for printable characters; `<a-o>` and `<a-O>` are different keys and we don't care about whether Shift was involved.

Kitty 0.20(.1?) added support for a new key-encoding (also supported by various other terminals) that provides more information about pressed keys. It seems like it includes the Shift key state in a way that Kakoune wasn't expecting. I'm not sure whether that's a bug in Kakoune, a bug in Kitty, or just a miscommunication between what Kakoune expects and what Kitty implemented.

Parasrah:
Sure enough I went back a single commit and rebuilt and the problem is no longer occurring. I appreciate the explanation, is this new key-encoding protocol why I had to manually map keys like `<c-[>` to `<esc>`, and why I can map [ambiguous keys](https://github.com/mawww/kakoune/wiki/Ambiguous-Keys) now?

[...]

PR #4415, opened 2021-10-30
Use shifted key codes on kitty (ready)

### LINKED ISSUE #4242: mapping percent w/ kitty
I used to use a mapping which used `<a-percent>` (something like `kakoune.cr`'s `map -docstring 'select all occurrences of current selection' global normal <a-percent> ':select-highlights<ret>'`)

It used to work in the past but since one of the last updates of either kakoune (building from HEAD) or kitty it doesn't do anything anymore.

I'm on
kitty 0.21.1
kakoune 8fdda6d
macOS 11.4

`:buffer *debug*; set -add window debug keys<ret>` reports:
```
Client 'client0' got key '<a-s-5>'
```

changing the shortcut to `<a-s-5>` throws this error on startup:
```
Shift modifier only works on special keys and lowercase ASCII, not '5'
```

`kitty @kitten show_key` reports:
```
UNIX        send_text
%        %
^[%        \x1b%
```

### LINKED ISSUE #4333: Forward jump no longer working in kitty (regression)
### Steps 
- Open a scratch buffer
- Jump to end of buffer (`ge`)
- Jump back (`<c-o>`)
- Attempt to jump forward again (`<c-i>`)

### Outcome 
Nothing happens.

### Expected 
The selection should jump back to the end of the buffer.

### Comments
This appears to be a regression that was introduced somewhere between v2020.09.01 and v2021.08.28.

eraserhd:
Adjusts #4159.
Closes #4242.
Closes #4333.

## How this works

This implements Kitty's [keyboard progressive enhancement](https://sw.kovidgoyal.net/kitty/keyboard-protocol/).  Basically, after enabling the CSI u mode, Kitty's progressive enhancement code is sent, which should be ignored by other terminals.

The progressive enhancement protocol uses "sub-arguments" separated by colons, so that instead of sending just the codepoint of the key pressed, it will also send the shifted key code and the layout key code.  We only use the shifted key code.

This means that we now have enough information to make, say `<a-(>` as the key, instead of `<a-s-9>`.  Without progressive enhancement, we can't do that unless we know the user's physical keyboard layout (or worse, assume it is ANSI).  In this PR, when a shifted character is given, we always use it and remove the shift modifier.

The previous work-around for alphabetic characters is removed, since this one works more generally.

## Testing

* In kitty, `<a-(>`, `<a-)>`, `<a-%>` etc. now work
* In kitty, `J` `K` etc. still work as they should and aren't reported as `<s-j>`, `<s-k>`.
* In kitty, `<s-pageup>` and `<s-pagedown>` are reported as they should, because those keys do not have shifted variants.
* In tmux, all the above still works.  (I assume tmux doesn't support CSI u, right?)

## Refactoring

I started refactoring the CSI key parsing code, but the refactor ended up big and I'm not so certain @mawww will like, so I'll make that a separate PR.

### DIFF
--- a/src/terminal_ui.cc
+++ b/src/terminal_ui.cc
@@ -754,7 +754,7 @@ Optional<Key> TerminalUI::get_next_key()
 
     auto parse_csi = [this, &convert, &parse_mask]() -> Optional<Key> {
         auto next_char = [] { return get_char().value_or((unsigned char)0xff); };
-        int params[16] = {};
+        int params[16][4] = {};
         auto c = next_char();
         char private_mode = 0;
         if (c == '?' or c == '<' or c == '=' or c == '>')
@@ -762,12 +762,17 @@ Optional<Key> TerminalUI::get_next_key()
             private_mode = c;
             c = next_char();
         }
-        for (int count = 0; count < 16 and c >= 0x30 && c <= 0x3f; c = next_char())
+        for (int count = 0, subcount = 0; count < 16 and c >= 0x30 && c <= 0x3f; c = next_char())
         {
             if (isdigit(c))
-                params[count] = params[count] * 10 + c - '0';
+                params[count][subcount] = params[count][subcount] * 10 + c - '0';
+            else if (c == ':' && subcount < 3)
+                ++subcount;
             else if (c == ';')
+            {
                 ++count;
+                subcount = 0;
+            }
             else
                 return {};
         }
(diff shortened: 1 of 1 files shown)


jtrees:
You. Are. Da. Man. Seriously, this will boost my productivity immensely. Thanks so much!

P.S.: It works. :) I didn't test every possible key combination but the ones that matter to me look good.

eraserhd:
Tested under tmux (as a non-xterm-like terminal) and kitty and Terminal.app, and `<a-)>` works on all three.

Started to refactor key parsing - extracting a CSIKey class and breaking up the large `parse_csi` function.  A little bit more to go yet.

raiguard:
I too can confirm that this has fixed the problem! Thank you @eraserhd :heart:

jtrees:
I just found a key combo that still doesn't seem to work with this PR: <c-i>

Can anyone confirm?

eraserhd:
> I just found a key combo that still doesn't seem to work with this PR:

The key combo was eaten by GitHub's markdown processing.  Can you paste it in backticks or use HTML entities?

jtrees:
Oops, sorry. I updated the comment.

PR #4418, opened 2021-11-01
Support dedicated filter text in "completions" options

krobelus:
This allows to resolve https://github.com/kak-lsp/kak-lsp/issues/444, via https://github.com/kak-lsp/kak-lsp/pull/551.

For other LSP completions it doesn't seem to matter much. For example clangd sends this:

```json
{
        "label": " someFunction() const volatile",
        "filterText": "someFunction",
        "insertText": "someFunction"
}
```

We are supposed to filter by `filterText`, but it's also fine (and more predictable) to use `insertText` here.

### DIFF
--- a/doc/pages/options.asciidoc
+++ b/doc/pages/options.asciidoc
@@ -142,22 +142,22 @@ are exclusively available to built-in options.
     escaped as `\|` or `\\`.
 
 *completions*::
-    a list of `<text>|<select cmd>|<menu text>` candidates,
+    a list of `<text>[|<filter text>]|<select cmd>|<menu text>` candidates,
     except for the first element which follows the
     `<line>.<column>[+<length>]@<timestamp>` format to define where the
     completion apply in the buffer.
     Any `|` or `\` characters that occur within the `<text>`,
-    `<select cmd>`, or `<menu text>` fields should be escaped as `\|`
-    or `\\`.
+    `<filter text>`, `<select cmd>`, or `<menu text>` fields should be
+    escaped as `\|` or `\\`.
 
     Options of this type are are meant to be added to the `completers`
     option to provide insert mode completion. Candidates are shown if the
     text typed by the user (between `<line>.<column>` and the cursor) is a
-    subsequence of `<text>`.
+    subsequence of `<filter text>` (if provided) or `<text>`.
 
-    For each remaining candidate, the completion menu displays
-    `<text>`, followed by `<menu text>`, which is a Markup string (see
-    <<faces#markup-strings,`:doc faces markup-strings`>>).
+    For each remaining candidate, the completion menu displays `<filter text>`
+    (if provided) or `<text>`, followed by `<menu text>`, which is a Markup
+    string (see <<faces#markup-strings,`:doc faces markup-strings`>>).
 
     As the user selects items from the completion menu, the text they typed
     will be replaced with `<text>`, and the Kakoune command in
--- a/src/insert_completer.cc
+++ b/src/insert_completer.cc
@@ -291,8 +291,15 @@ InsertCompletion complete_option(const SelectionList& sels,
         using RankedMatch::operator==;
         using RankedMatch::operator<;
 
+        Optional<StringView> completion;
         StringView on_select;
         DisplayLine menu_entry;
+
+        const StringView &candidate() const {
+            if (completion)
+                return *completion;
+            return RankedMatch::candidate();
+        }
     };
 
     StringView query = buffer.substr(coord, cursor_pos);
@@ -300,14 +307,33 @@ InsertCompletion complete_option(const SelectionList& sels,
 
     for (auto& candidate : opt.list)
     {
-        if (RankedMatchAndInfo match{std::get<0>(candidate), query})
+        StringView completion;
+        Optional<StringView> filter_text;
+        StringView on_select;
+        StringView menu;
+        if (std::holds_alternative<std::tuple<String, String, String, String>>(candidate))
         {
-            match.on_select = std::get<1>(candidate);
-            auto& menu = std::get<2>(candidate);
-            match.menu_entry = not menu.empty() ?
-                parse_display_line(expand_tabs(menu, tabstop, column), faces)
+            auto& cand = std::get<std::tuple<String, String, String, String>>(candidate);
+            completion = std::get<0>(cand);
+            filter_text = {std::get<1>(cand)};
+            on_select = std::get<2>(cand);
+            menu = std::get<3>(cand);
+        }
+        else
+        {
+            auto& cand = std::get<std::tuple<String, String, String>>(candidate);
+            completion = std::get<0>(cand);
+            on_select = std::get<1>(cand);
+            menu = std::get<2>(cand);
+        }
+        if (RankedMatchAndInfo match{filter_text.value_or(completion), query})
+        {
+            if (filter_text)
+                match.completion = completion;
+            match.on_select = on_select;
+            match.menu_entry = not menu.empty()
+              ? parse_display_line(expand_tabs(menu, tabstop, column), faces)
               : DisplayLine{String{}, {}};
-
             matches.push_back(std::move(match));
         }
     }
--- a/src/insert_completer.hh
+++ b/src/insert_completer.hh
@@ -8,6 +8,8 @@
 
 #include "optional.hh"
 
+#include <variant>
+
 namespace Kakoune
 {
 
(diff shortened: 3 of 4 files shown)


krobelus:
Ok that didn't work at all (because it messes with the buffer while the completion is active). Changed now to just add an optional filter text to completions, which allows to resolve the motivating issue - a completion that inserts `ξ` which can be triggered by typing `\xi`.

doc/pages/options.asciidoc
```diff
@@ -139,13 +139,16 @@ are exclusively available to built-in options.
...
     `<line>.<column>[+<length>]@<timestamp>` format to define where the
     completion apply in the buffer.
     Any `|` or `\` characters that occur within the `<text>`,
-    `<select cmd>`, or `<menu text>` fields should be escaped as `\|`
-    or `\\`.
+    `<filter text>`, `<select cmd>`, or `<menu text>` fields should be
+    escaped as `\|` or `\\`.
+
+    Kakoune will filter away candidates that do not match what the user types.
+    Candidates are matched against the optional `<filter text>`, or `<text>`.
```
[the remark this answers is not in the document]
krobelus: yep, I added a commit that tries to explain this

> - the completion menu displays "filter text" (if provided) and/or "text"

Currently that's an "or", but at least for the Julia completion of Unicode symbols it would definitely work well to display `\xrat (℞)` here.

I did not yet find other examples where the text-to-insert is different from the filter text.
If the filter text can be very similar to the text-to-insert, it may not be wise to display both.

doc/pages/options.asciidoc
```diff
@@ -139,20 +139,27 @@ are exclusively available to built-in options.
...
-    to display element specific documentation.
-
-    Markup can be used in the menu text.
-    (see <<faces#markup-strings,`:doc faces markup-strings`>>)
+    `<filter text>`, `<select cmd>`, or `<menu text>` fields should be
+    escaped as `\|` or `\\`.
+
+    Options of this type are are meant to be added to the `completers`
+    option to provide insert mode completion.  Candidates are shown if the
+    text typed by the user (between `<line>.<colum>` and the cursor) is a
```
Screwtapello, line 152:
```suggestion
    text typed by the user (between `<line>.<column>` and the cursor) is a
```

doc/pages/options.asciidoc
```diff
@@ -139,20 +139,27 @@ are exclusively available to built-in options.
...
+    prefix of `<filter text>` (if provided) or `<text>`.
```
Screwtapello, line 153: I'm not sure there's a concise word that describes Kakoune's fuzzy matching, but "substring" might be better than "prefix".

doc/pages/options.asciidoc
```diff
@@ -139,20 +139,27 @@ are exclusively available to built-in options.
...
+
+    For each remaining candidate, the completion menu displays `<filter text>`
+    (if provided) or `<text>`, followed by `<menu text>`, which can contain
+    Markup (see <<faces#markup-strings,`:doc faces markup-strings`>>).
+
+    If the user selects an item from the completion menu, the text they typed
```
Screwtapello, line 159:
```suggestion
    As the user selects items from the completion menu, the text they typed
```

With the original wording, it sounded like these actions occurred once the user had accepted a completion, rather than just while they cycle through the completione.

doc/pages/options.asciidoc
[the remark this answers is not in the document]
Screwtapello: The more general form of completions is "type the name of something, get something else". It could be something shorter (like typing "Smiling Face with Open Mouth and Tightly-Closed Eyes" to get 😆), or it could be something much longer (like typing "defprop" to get a snippet that defines a property access method — I know @alexherbo2 has been working on something like this).

If I had to pick a convention for people to follow, it would probably be that the completion menu itself displays the text that I need to type to narrow down the list of completions, while the menu text (if short) or `info -menu` (if long) is used to preview the thing that will be inserted when I accept an option.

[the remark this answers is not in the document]
krobelus: Right, clients that use filter-text can use the menu-text and the on-select handler to display what would be inserted.

doc/pages/options.asciidoc
krobelus: I changed it to [subsequence](https://en.wikipedia.org/wiki/Subsequence) which also matches the term used by the implementation (see `subsequence_match_smart_case`)

koehlerson:
Me and colleagues use this branch on a daily basis due to the completion possibilities for Julia. So, just want to ping and emphasize that there is some interest in merging it, since currently the branch cannot be compiled by gcc-12.2 due to the constexpr issue https://github.com/mawww/kakoune/pull/4549

mawww:
@koehlerson Thanks for bumping this, I took a look when it was proposed but did not take the time to give feedback, I am in favor of adding this feature but am not too keen on the variant based implementation. I'll try to go through and give more specific feedback to @krobelus in the coming days.

krobelus:
> So, just want to ping and emphasize that there is some interest in merging it

I think the Julia LSP Unicode input use case is not so strong because there are probably better alternatives:
1. the OS could provide Unicode input
2. the language server could send snippets; so the completion would just be
   the literal text `\xi` but when accepting it it will be replaced by ξ.
   It is conceivable that we can teach kak-lsp to rewrite those Julia
   completions that have filterText to equivalent snippets. (Snippet behavior
   has been fixed recently.) Better yet, the Julia language server could
   send snippets, maybe only if the LSP client announces that it doesn't
   support filterText (might need a new LSP capability). I didn't test this,
   will try soon.

There is another use case with PHP/intelephense. The commit message tries
to explain it. I haven't spend much ciritcal thought on that use case,
maybe it can also be satisfied by telling the LSP server that we don't
support filterText?

I think the main downside of this patch is that it makes the completion
option harder to understand (for users reading the docs), which is tragic
as the filterText is only useful to relatively few users.

I'm not against adding filterText support to Kakoune but if we do, we
probably want to implement custom completion sorting the same way (can't
find the issue right now). One proposal is to add a "key" field to completions
tuple. But having 5 tuple elements, most of which are optional is pretty messy,
something like named keys would probably be easier to work with.

> since currently the branch cannot be compiled by gcc-12.2 due to the constexpr issue

It should be easy for you to rebase this patch on latest master yourself.
I've also pushed a rebased version, just in case.

koehlerson:
> It should be easy for you to rebase this patch on latest master yourself. I've also pushed a rebased version, just in case.

Sorry, I formulated it too harsh. I point to this PR whenever someone is asking what needs to be done. Thanks for rebasing it, but it wasn't meant to push you in that direction even though it's nice to be able again to point to this specific PR or respectively the issue in `kak-lsp`.

doppioandante:
Having developed a small plugin that uses julia REPL library for completion (without going through the complexity of the LSP) a feature like this would be helpful for my use case, but I don't have a strong opinion, I agree that simpler may be better.

krobelus:
I took a quick look, AFAICT the difference between this patch and a snippet-based solution is that snippets can only expand the value (`\xi` to `ξ`) *after* the completion has been accepted (that is, the completion menu has been hidden).

The snippet solution would use InsertCompletionHide (like kak-lsp does nowadays).

krobelus:
I've implement the LSP filterText support via snippet expansion in https://github.com/kak-lsp/kak-lsp/commit/25076db47cd50e209823adb8c59e968aff57ba68

There are two small UI differences to using this patch.
To test, either run Kakoune without this patch or add `set-option global lsp_have_kakoune_feature_filtertext false`.
@koehlerson how do you feel about this one? I think it works just as well although it might not be as intuitive.
I think the Julia language sever could also improve things (see commit message).

Screwtapello:
> However this is something the language server can fix easily: they could simply set the "completionItem.label" property to something like "\xi → ξ".

Is this something kak-lsp could add to the completions' menu text by default, if the LSP doesn't specify a label? Or is the problem that the LSP already sets the label to something unhelpful?

krobelus:
Good point, yeah kak-lsp could add a label if missing (although I'd prefer not to, an LSP client shouldn't get creative here).
Indeed the Julia language server already sets `label` and relies on the fact that `filterText` defaults to `label`:

```json
{
  "label": "\\xi",
  "kind": 11,
  "detail": "ξ",
  "documentation": "ξ",
  "textEdit": {
    "range": {
      "start": {
        "line": 25,
        "character": 0
      },
      "end": {
        "line": 25,
        "character": 2
      }
    },
    "newText": "ξ"
  }
},
```

reported as https://github.com/julia-vscode/LanguageServer.jl/issues/1157

krobelus:
Nevermind I don't think we can safely make up a label, mostly because that property is mandatory in LSP.

koehlerson:
Ok tried it with kakoune#master and the only UI difference is the additional whitespace press to confirm the selection from the options, isn't it? From my perspective it's totally fine and I actually like the additional confirmation via whitespace press

krobelus:
Yeah that and the labels are the only differences.
Thanks for the feedback, let's close since there seems to be no strong motivation left.

PR #4420, opened 2021-11-02
WIP: CSI parsing code refactor

eraserhd:
This is an attempt to split parsing and handling of CSI codes, and to make a useful data structure storing the parsed key sequence.

This is on top of #4415 .

If this isn't a good idea let me know.  I want to test mouse parsing a bit more before it is ready.


lenormf:
This is good because it exposes parsing functions to the root namespace, which is accessible to fuzzers.

mawww (commented):
I am really not a big fan of introducing this class, its interface is very coupled with the parse_csi function to the point where it is clearly an implementation detail of it. Overall, I dont think the resulting code is easier to read or cleaner.

It could make sense to extract the existing lambdas as functions, but as it stands this refactor either does too much, or not enough change.

PR #4421, opened 2021-11-04
Miscellaneous fixes for `man.kak`

### LINKED ISSUE #3970: `man.kak` declares a user mode

lenormf:
Closes #3970.

### DIFF
--- a/rc/tools/man.kak
+++ b/rc/tools/man.kak
@@ -108,8 +108,12 @@ declare-option -hidden regex man_link2 \
 # Define a useful command sequence for searching a given regex
 # and a given sequence of search keys.
 define-command man-search -params 2 %{
-  set-register / %arg[1]
-  execute-keys %arg[2]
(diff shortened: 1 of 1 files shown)


lenormf:
With #3221 the mode can get a docstring as well.

PR #4423, opened 2021-11-04
OCaml: Add a command to shift between .ml and .mli files and vice versa

sidkshatriya:
C has header and source files and you need to often switch between them.
Similarly OCaml has .ml (implementation) and .mli (interface files) and
one often needs to switch between them.

This commit provides a simple functionality that allows you to accomplish this.

### DIFF
--- a/rc/filetype/ocaml.kak
+++ b/rc/filetype/ocaml.kak
@@ -18,7 +18,11 @@ hook global WinSetOption filetype=ocaml %{
 
 hook -group ocaml-highlight global WinSetOption filetype=ocaml %{
     add-highlighter window/ocaml ref ocaml
-    hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/ocaml }
+    alias window alt ocaml-alternative-file
+    hook -once -always window WinSetOption filetype=.* %{
+        unalias window alt ocaml-alternative-file
+        remove-highlighter window/ocaml
+    }
 }
 
 provide-module ocaml %{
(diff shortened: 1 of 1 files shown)


sidkshatriya:
cc: @alexherbo2 @aecepoglu @Guest0x0 

I notice that you have made contributions to `ocaml.kak` so I thought you might be interested in this small PR! Please add any comments if you find this PR useful / have any suggestions.

Thank you!

rc/filetype/ocaml.kak
```diff
@@ -74,4 +74,22 @@ evaluate-commands %sh{
...
 
+# Conveniences
+# ‾‾‾‾‾‾‾‾‾‾‾‾
+
+# C has header and source files and you need to often switch between them.
+# Similarly OCaml has .ml (implementation) and .mli (interface files) and
+# one often needs to switch between them.
+#
+# This command provides a simple functionality that allows you to accomplish this.
+define-command ocaml-ml-switch-mli -docstring 'Switch between .ml and .mli file or vice versa' %{
```
krobelus, line 85: This should probably be called `ocaml-alternative-file`, and set the `alt` alias at window scope.
See `go-alternative-file` for an example, their error messages are also nice.

rc/filetype/ocaml.kak
```diff
@@ -74,4 +74,22 @@ evaluate-commands %sh{
...
+    evaluate-commands %sh{
+        if [ "${kak_buffile##*.}" = 'ml' ]; then
+            printf "edit -existing '%s'\n" "${kak_buffile}i"
+        elif [ "${kak_buffile##*.}" = 'mli' ]; then
+            printf "edit -existing '%s'\n" "${kak_buffile%i}"
```
krobelus, line 90: I guess we can protect against weird filenames

```suggestion
            printf "edit -existing -- '%s'\n" "$(printf %s "${kak_buffile%i}" | sed "s/'/''/g")"
```

Guest0x0:
My two cents here.. Switching between `.ml` and `.mli` files is indeed a common action when writing OCaml code. I remember when I was using Vim, the OCaml filetype script there has a similar command, too

sidkshatriya:
Thanks for your comments @Guest0x0 !

The vscode ocaml [plugin](https://github.com/ocamllabs/vscode-ocaml-platform) also allows you to switch between ml and mli files. (I'm sure there is something in emacs too.)

This is definitely a common requirement when developing in OCaml...  There are some small changes requested in this PR which are easily accomplished, I'm going to get to it soon.

sidkshatriya:
@krobelus 
Made the changes so that this feature is consistent with `go.kak`'s similar feature i.e. added the `alt` alias, escaped any `'` characters in the filename, renamed the command...

Note that `edit` does not have the `-existing` flag anymore. If a `mli`/`ml` alternate file does not exist a new empty one will be created by kakoune. This is consistent with the behavior in the [vscode plugin](https://github.com/ocamllabs/vscode-ocaml-platform) also.

rc/filetype/ocaml.kak
sidkshatriya: Done! 

I don't think any error messages are needed for OCaml though...

rc/filetype/ocaml.kak
sidkshatriya: Done and made consistent with `go.kak` (e.g. `-existing` removed, see below)

PR #4424, opened 2021-11-04
src highlighters: Factorise docstrings

### LINKED ISSUE #4367: Describe highlighter parameters in the autoinfo
### Feature 

Parameters for highlighters are hardcoded:

https://github.com/mawww/kakoune/blob/b3a1017a264cfea7e9ae1a374208e9b02efbda5a/src/highlighters.cc#L2283-L2383

They are also declared in a private list in the instance of each highlighter, with empty docstring parameters:

https://github.com/mawww/kakoune/blob/b3a1017a264cfea7e9ae1a374208e9b02efbda5a/src/highlighters.cc#L1095-L1102

These docstrings should be set, and the static list of parameters used to generate the highlighters' autoinfo docstring at runtime.

### Usecase 

For example, type in `:addhl buffer/ number-lines ` and understand what the `-min-digits` parameter does without opening `:doc highlighters` in a separate client.

lenormf:
Fixes #4367.

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -987,7 +987,19 @@ const CommandDesc add_highlighter_cmd = {
             HighlighterRegistry& registry = HighlighterRegistry::instance();
             auto it = registry.find(params[1]);
             if (it != registry.end())
-                return format("{}:\n{}", params[1], indent(it->value.docstring));
+            {
+                auto docstring = it->value.description->docstring;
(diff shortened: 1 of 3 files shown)


lenormf:
The one highlighter I use for the screenshot had some docstring info missing (the parameters), so here it is again:

2021-11-05-114715_1920x590_scrot

PR #4430, opened 2021-11-10
change-directory: echo the changed directory on the status line

sidkshatriya:
change-directory: echo the changed directory on the status line

Sometimes you are not sure about which exact directory you chose to `cd` to. This provides just a bit of feedback.

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -2525,11 +2525,14 @@ const CommandDesc change_directory_cmd = {
                                         context.options()["ignored_files"].get<Regex>(),
                                         cursor_pos, FilenameFlags::OnlyDirectories) };
         }),
-    [](const ParametersParser& parser, Context&, const ShellContext&)
+    [](const ParametersParser& parser, Context& context, const ShellContext&)
     {
         StringView target = parser.positional_count() == 1 ? StringView{parser[0]} : "~";
-        if (chdir(parse_filename(target).c_str()) != 0)
+        auto dir = parse_filename(target);
+        if (chdir(dir.c_str()) != 0)
             throw runtime_error(format("unable to change to directory: '{}'", target));
+
+        context.print_status({dir, context.faces()["StatusLine"]});
         for (auto& buffer : BufferManager::instance())
             buffer->update_display_name();
     }


sidkshatriya:
@mawww I would very much appreciate if you chimed in on this small feature. I'm OK if you're not in favor -- we can park it then.

sidkshatriya:
@lenormf Interesting point: scripts that use `cd` in a non-draft context would echo the directory on the command line. That may not be desirable.

But don't scripts that want to do things without affecting the user interface need to run the commands in a draft context? Isn't the draft context already the way to suppress output?

In https://github.com/mawww/kakoune/issues/4449#issue-1060886393 you link to a specific example https://github.com/lenormf/kakmerge/blob/105893428ad6fd875f30b538d63a3454e76abf47/kakmerge#L21-L23 but there eval/exec are choosing specific clients, hence the `echo ""` workaround that comes after that.

So:
- If you `cd` within a draft context you will never see an echo
- Most usage `cd` should be in `%sh{}` blocks anyways

If there are other scenarios where `cd` echo-ing by default will generally be a problem, please let me know.

lenormf:
I’m in favour of this PR, and I also forgot about draft contexts when I created the issue linked.

It’s so cumbersome to wrap commands in `eval -draft` that I didn’t consider it at all, I was hoping for something equivalent to and as simple as <kbd>\\</kbd>, or a special register that would act as a flag.

mawww:
I am not fully convinced we need that, if you cd'd to a directory without knowing what it was, it sounds like its already too late and I think the magic `<a-!>` keystroke in the prompt would be a better solution.

I see two case: either you use cd with and explicit directory name and then you already know the directory you are moving to, in which case I see little value in echoing it again, or you cd into some directory generated by an expansion statement, and in that case hitting `<a-!>` will evaluate that expansion in the prompt and give you the opportunity to drop the cd command if its not correct.

lenormf:
A usecase for this change would be my old [cdmenu](https://github.com/lenormf/kakoune-extra/blob/master/cdmenu.asciidoc) script that allows jumping to a dynamic directory.

For example, `:cdmenu<ret>` <kbd>g</kbd> would jump to the root of the current Git repository with `cd`, which could spontaneously indicate to the user that the CWD was changed without them interactively deciding what path.

krobelus:
doesn't cdmenu already use `echo` for that?

sidkshatriya:
> I see two case: either you use cd with and explicit directory name and then you already know the directory you are moving to, in which case I see little value in echoing it again, or you cd into some directory generated by an expansion statement, and in that case hitting <a-!> will evaluate that expansion in the prompt and give you the opportunity to drop the cd command if its not correct.

There is another use case than the above for me and that is the menu that pops up when you type in `:cd` and you can tab through different options. That can be unwieldy when there are a lot of choices and often you don't know if you ended up chosing `some_dir_a` or `some_dir_aa` when you pressed `<ret>`. The terminal UI can sometimes be a bit clunky and that is the nature of the interface. This just helps to give you some visual feedback.

I guess there are two philosophies here: can we help the user a bit? Or should be totally orthogonal and non-redundant?

Though to be honest, I don't see how just echoing the new directory would be redundant. Because it serves an important function: user feedback. And user feedback should not be considered redundant.

Take git. 
```
$ git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
$
```

One could argue that git should _not_ say that I have switched to branch master. It should be quiet and only give me an error if it was _not_ able to checkout master. But yet, git does it. It gives the reassurance to the user that the current branch is indeed master. Similarly `cd` in kakoune should convey to the user the new directory the editor has switched to. For someone who switches directories often in kakoune this would be a useful behavior in the default kakoune installation.

Take `cd` itself in a shell prompt.

```
sid@some-computer ~> cd some-directory
sid@some-computer ~/some-directory>
```

You don't miss the fact `cd` in a shell prompt does not echo you the new directory because this important feedback is fulfilled by any decent shell prompt that includes the directory. Using any shell without knowing what is your current directory is, and where the prompt is always `$` is a headache because then we always need to type in `pwd`. 

Interestingly the lack of `pwd` in default kakoune is another problem for me but that is entirely another issue.

---
BTW `<a-!>` is cool.

sidkshatriya:
@mawww Any chance you were convinced by the above argument? I'm happy for us to park it otherwise.

mawww:
Not really convinced, if the issue is that you are not sure what completion entry was selected when you pressed enter, I'd rather find a way to fix that at the completion system level, not only for the `cd` command.

The feedback you get from git/shell is tad more convincing to me, and you could point to some existing feedback mechanisms in Kakoune, such as echoing the new search pattern on `*`. However, at least in the shell case, you get that information after every command, not just after cd-ing, which is probably the moment you need it least. It might be due to how I use Kakoune, but I very rarely wonder what my cwd is, and if I do it will not be just after cd-ing into a directory.

I have looked at how this could be implemented as user configuration and we hit an issue that we cannot avoid aliases (`def -params 1 cd-echo %{ cd %arg{1}; echo %arg{1} }; alias global cd cd-echo` ends up in an infinite recursion). This is something that needs to be fixed, we need to have a way to call the command itself, probably as something like `eval -noalias -verbatim cd %arg{1}`

krobelus:
> I have looked at how this could be implemented as user configuration and we hit an issue that we cannot avoid aliases (`def -params 1 cd-echo %{ cd %arg{1}; echo %arg{1} }; alias global cd cd-echo` ends up in an infinite recursion). This is something that needs to be fixed, we need to have a way to call the command itself, probably as something like `eval -noalias -verbatim cd %arg{1}`

This is trivial to implement in user configuration, just don't call the alias but the actual `change-directory`:

```kak
define-command -override -params 1 verbose-cd %{
    change-directory %arg{1}
    echo -- "%sh{pwd -P}"
} -file-completion
alias global cd verbose-cd
```

(or add a command with a different name)
The infinite recursion is interesting, maybe we want a cycle detection with a low limit on how often each alias is called.

> Take `cd` itself in a shell prompt.
>  where the prompt is always `$` is a headache because then we always need to type in `pwd`.

That's purely a matter of preference. I use `$` as my shell prompt because I usually know my cwd, plus my terminal title also has it.

I appreciate the attention to detail but I think we have other, more exciting changes to Kakoune coming up.
Print the cwd after cd is easy to configure and it probably depends on the workflow whether the user prefers that to silence.

We can add `-directory-completion` to the above alias to get only directories, so the diff below.
I haven't checked whether the same is possible with `-shell-script-completion`, but I also have this command to quickly go to frequently used directories

```kak
define-command my-cd -params ..1 %{ cd %arg{@} } -shell-script-candidates %{
	# My shell's $CDPATH is .:~:~/git but exporting CDPATH can break scripts, so duplicate it here.
	find . ~ ~/git -maxdepth 1 -type d -o -type l | sed s,^$HOME,~,
}
```

```diff
diff --git a/src/commands.cc b/src/commands.cc
index 4ca9bf9e..c8c64169 100644
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -1214,9 +1214,12 @@ void define_command(const ParametersParser& parser, Context& context, const Shel
     }
 
     CommandCompleter completer;
-    if (parser.get_switch("file-completion"))
+    if (parser.get_switch("file-completion") || parser.get_switch("directory-completion"))
     {
-        completer = [=](const Context& context, CompletionFlags flags,
+        auto flags = parser.get_switch("file-completion")
+            ? FilenameFlags::Expand
+            : FilenameFlags::OnlyDirectories;
+        completer = [=](const Context& context, CompletionFlags completion_flags,
                        CommandParameters params,
                        size_t token_to_complete, ByteCount pos_in_token)
         {
@@ -1224,7 +1227,7 @@ void define_command(const ParametersParser& parser, Context& context, const Shel
              const auto& ignored_files = context.options()["ignored_files"].get<Regex>();
              return Completions{0_byte, pos_in_token,
                                 complete_filename(prefix, ignored_files,
-                                                  pos_in_token, FilenameFlags::Expand),
+                                                  pos_in_token, flags),
                                 completions_flags};
         };
     }
@@ -1297,6 +1300,7 @@ const CommandDesc define_command_cmd = {
           { "docstring",                { true,  "define the documentation string for command" } },
           { "menu",                     { false, "treat completions as the only valid inputs" } },
           { "file-completion",          { false, "complete parameters using filename completion" } },
+          { "directory-completion",     { false, "complete parameters using directory completion" } },
           { "client-completion",        { false, "complete parameters using client name completion" } },
           { "buffer-completion",        { false, "complete parameters using buffer name completion" } },
           { "command-completion",       { false, "complete parameters using kakoune command completion" } },
@@ -2100,6 +2104,7 @@ const CommandDesc prompt_cmd = {
         { { "init", { true, "set initial prompt content" } },
           { "password", { false, "Do not display entered text and clear reg after command" } },
           { "file-completion", { false, "use file completion for prompt" } },
+          { "directory-completion", { false, "use directory completion for prompt" } },
           { "client-completion", { false, "use client completion for prompt" } },
           { "buffer-completion", { false, "use buffer completion for prompt" } },
           { "command-completion", { false, "use command completion for prompt" } },
@@ -2119,15 +2124,17 @@ const CommandDesc prompt_cmd = {
         auto initstr = parser.get_switch("init").value_or(StringView{});
 
         PromptCompleter completer;
-        if (parser.get_switch("file-completion"))
-            completer = [](const Context& context, CompletionFlags,
+        if (parser.get_switch("file-completion") || parser.get_switch("directory-completion")) {
+            auto flags = parser.get_switch("file-completion")
+                ? FilenameFlags::Expand
+                : FilenameFlags::OnlyDirectories;
+            completer = [flags](const Context& context, CompletionFlags,
                            StringView prefix, ByteCount cursor_pos) -> Completions {
                 auto& ignored_files = context.options()["ignored_files"].get<Regex>();
                 return { 0_byte, cursor_pos,
-                         complete_filename(prefix, ignored_files, cursor_pos,
-                                           FilenameFlags::Expand) };
+                         complete_filename(prefix, ignored_files, cursor_pos, flags) };
             };
-        else if (parser.get_switch("client-completion"))
+        } else if (parser.get_switch("client-completion"))
             completer = [](const Context& context, CompletionFlags,
                            StringView prefix, ByteCount cursor_pos) -> Completions {
                  return { 0_byte, cursor_pos,
```

krobelus:
Having said that, this may be useful with `cd ..`

Issue #2330, opened 2018-08-24
Add kak_session_env_NAME


alexherbo2:

mawww:
I believe this would be exactly the same as `$NAME`, wouldnt it ?

alexherbo2:
Yep

lenormf:
Why do you need this redundant expansion, in that case?

alexherbo2:
To be able to:

``` kak
echo %val(session_env_NAME)
```

Instead of:

``` kak
echo %sh(echo $NAME)
```

lenormf:
How about `%val{env_NAME}` directly? The fact that there is a `%val{client_env_NAME}` hints that the former is expanded on the server.

alexherbo2:
It’s a good suggestion.  I’m not sure which is best.

lenormf:
The case I was thinking of was when I spawn a server in the root of a project, in daemon mode, I can easily get the path to the root of the project from subsequent clients with `%val{env_PWD}`.

It allows me not to rely on a VCS, and is more practically than spawning a shell to execute `pwd`.

alexherbo2:
@mawww It is useful in the kakrc, because there is no client yet and to not spawn a shell.  I also have some use with plugins to save shell calls.

krobelus:
Why not use a single shell block to define Kakoune variables for the environment variables you care about?

```
eval %sh{
  echo declare-option str github $GIT_HUB
  ...
}
```

mawww:
Avoiding a shell call in the kakrc is not a strong motivating use case. If you need control flow or to access to some env vars, going through a `%sh{...}` block is the idiomatic way.

Issue #2091, opened 2018-05-30
Expose working directory via option interface


ul:
I propose to make working directory an option with usual scoping rules. This is more or less backward compatible (`change-directory ...` could be alias for `set global cwd ...`), makes things like `autochdir.kak` trivial in implementation and more robust (one-line hook!), and things like providing relative paths in `filetype=grep` buffers non-intrusive (`set buffer cwd ...`).

I don't see immediate downsides or caveats except of complexity of implementation which could be high enough to outweigh benefits.

Screwtapello:
Where is the current directory used, outside of `%sh{}` blocks?

- filename completion in insert mode?
- filename completion in the command-prompt?
- anything else?

I guess a Kakoune client would call `getcwd(2)` when it starts, and set it in the window-scope? If a client is editing a file, and the user types `:edit some/relative/path` then somehow the current window's working directory would need to be copied to the new window.

What happens if you open a file in one client, and then use `:b` to show that buffer in a different client with a different (POSIX) working directory?

It'd probably work quite nicely for an auto-chdir style setup, but I think it would be weird for anything less aggressive than that.

ul:
I imagine that Kakoune server would set `getcwd(2)` call result in a global scope. After that if no overrides are provided in window or buffer scope everything works the same way as it does now. If user types `:edit some/relative/path` or selects buffer using `:b` in another window then destination shares the same global value as it shares server's cwd now. In case if cwd option get's overridden in a window or buffer scope then the same inheritance rules apply as to other options. As far as I can see it should work fine with non-auto-chdir style, especially if you touch that cwd option only when needed, like for `*references*` buffer.

mawww:
> * filename completion in insert mode?

Those already respect the path option, which gives a list of directory to look in for insert mode path completion. This option already support '%/' as meaning current buffer directory.

I am not sure I understand the use case for such a feature, likely due to difference in workflow. I tend to run one Kakoune session per "project" and it stays at the project root, using `find`, `ctags` and `lsp-goto-definition` to jump around.

Could you describe what kind of workflow works better with this feature ?

ul:
[...]

PR #4431, opened 2021-11-10
pwd: Add a new command that prints out the server's current working directory

sidkshatriya:
```
pwd: Add a new command that prints out the server's current working directory
```

I have lost count of the times I have had to do `:echo %sh{pwd}` . I think we should have this command, especially because we have `change-directory` . 

Since we can `cd` we should also have the capability of checking our `pwd`.

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -31,6 +31,7 @@
 
 #include <functional>
 #include <utility>
+#include <cerrno>
 
 #include <sys/types.h>
 #include <sys/stat.h>
@@ -2535,6 +2536,24 @@ const CommandDesc change_directory_cmd = {
     }
 };
 
+const CommandDesc present_working_directory_cmd = {
+    "print-working-directory",
+    "pwd",
+    "print-working-directory: print the server's current working directory to the status line",
+    no_params,
+    CommandFlags::None,
+    CommandHelper{},
+    CommandCompleter{},
+    [](const ParametersParser&, Context& context, const ShellContext&)
+    {
+        char cwd[PATH_MAX+1];
+        if (!::getcwd(cwd, sizeof(cwd)))
+            throw runtime_error(format("unable to get the current working directory (errno: {})", ::strerror(errno)));
+
+        context.print_status({cwd, context.faces()["StatusLine"]});
+    }
+};
+
 const CommandDesc rename_session_cmd = {
     "rename-session",
     nullptr,
@@ -2739,6 +2758,7 @@ void register_commands()
     register_command(set_register_cmd);
     register_command(select_cmd);
     register_command(change_directory_cmd);
+    register_command(present_working_directory_cmd);
     register_command(rename_session_cmd);
     register_command(fail_cmd);
     register_command(declare_user_mode_cmd);


src/commands.cc
```diff
@@ -2535,6 +2536,24 @@ const CommandDesc change_directory_cmd = {
     }
 };
 
+const CommandDesc present_working_directory_cmd = {
+    "present-working-directory",
```
lenormf, line 2540: I think the `p` stands for `print`.

src/commands.cc
```diff
@@ -2535,6 +2536,24 @@ const CommandDesc change_directory_cmd = {
...
+    "pwd",
+    "present-working-directory: get the server's current working directory",
+    no_params,
+    CommandFlags::None,
+    CommandHelper{},
+    CommandCompleter{},
+    [](const ParametersParser&, Context& context, const ShellContext&)
+    {
+        char cwd[1024];
```
lenormf, line 2549: Better use `PATH_MAX` here, 1024 seems short.

src/commands.cc
```diff
@@ -2535,6 +2536,24 @@ const CommandDesc change_directory_cmd = {
...
+        if (!::getcwd(cwd, 1024))
```
lenormf, line 2550: Instead of hardcoding the size, you could use `sizeof(cwd)`.

lenormf:
I can’t find the previous issue/PR that requested that change, I only remember that it was turned down because it’s so easily implemented in the user configuration that those who need it can just create a command in `kakrc`.

Maybe enough users have had to implement this locally that it would warrant a *laissez-passer* into the core?

sidkshatriya:
FWIW, neovim, vim and helix-editor all come with the `pwd` command.

occivink:
I don't think a simple convenience command belongs in the core, but maybe a `current_directory` value could be useful, to avoid the `%sh{pwd}` dance. Just my opinion ofc

sidkshatriya:
To me "core" also signifies _essential_ functionality. Some essential functionality _can_ be built outside the core but in my view it is best available in core because it is frequently used. We shouldn't expect new users to write a `define-command` for `pwd` when they are just starting out.

As a new user I found it very frustrating that I could not find out my pwd. Eventually I figured out `:echo %sh{pwd}` and I regularly use it to this day (though I will probably need to define something in my `kakrc` if this PR is rejected  ;-) ) 

Regarding @occivink's idea to have a value called `current_directory` -- it is is a good idea also but this can be done in addition to this PR. It need not be a mutually exclusive addition to kakoune.

src/commands.cc
sidkshatriya: Done.

I've also made some additional PATH_MAX related changes in rest of the code base. See https://github.com/mawww/kakoune/pull/4431/commits/76f925e9d4dfb974fcbd0c0a5299b1d1420e27b3 . Whether or not this PR gets accepted I think [that commit](https://github.com/mawww/kakoune/pull/4431/commits/76f925e9d4dfb974fcbd0c0a5299b1d1420e27b3 ) should be cherry-picked into the code? cc: @mawww

src/commands.cc
sidkshatriya: Changed!

sidkshatriya:
@lenormf I noticed a downvote to my previous comment. I am cool with it because there will always be disagreements on aesthetics, core values and code in any project. 

I am curious about what exactly you disagree with in my comment -- a reason for the downvote is always useful for the other party (i.e. me). It helps me understand -- and perhaps even come around to _your point of view_!

I also notice that you frequently downvote issues on Kakoune -- that is again totally cool because it one should also oppose inclusion of features that may not be good. But sometimes this has an effect of discouraging new users / first time contributors especially. 

May I suggest that instead of just a down vote, please explain your opposition to the suggested change. You are free to do whatever you choose ultimately -- this is just a humble suggestion.

Delapouite:
Related discussion on exposing the cwd https://github.com/mawww/kakoune/issues/2091

src/commands.cc
sidkshatriya: @lenormf Thanks for your comments related to PATH_MAX on https://github.com/mawww/kakoune/commit/76f925e9d4dfb974fcbd0c0a5299b1d1420e27b3#r60092366 -- useful!

(I am shifting the conversation back here as it is easier to refer back to a conversation on a PR than a loose commit coming from my fork)

I had a glance at the link you provided -- https://stackoverflow.com/questions/9449241/where-is-path-max-defined-in-linux

The headline take away for me is that PATH_MAX is a dynamic system quantity that can vary. So for instance different file systems can have different values on Linux. Since kakoune can edit files on different file systems the effective PATH_MAX will vary. So its not really even possible to define something like this at compile time.

Regarding the kakoune code base. We have three scenarios currently on the master branch:
- Usage of PATH_MAX
- Usage of PATH_MAX+1
- Usage of hard coded number like 1024

I changed all of them to option 1 for consistency in https://github.com/mawww/kakoune/commit/76f925e9d4dfb974fcbd0c0a5299b1d1420e27b3 .

So if I understand correctly you are advocating using PATH_MAX+1 in case there is an OS that defines PATH_MAX to be non-inclusive of nul? Linux BTW very clearly defines that it includes `nul`.

PATH_MAX has the advantage that it looks cleaner in code. PATH_MAX+1 only has the minor advantage if:

(a) We are using an operating system that `#defines` PATH_MAX as _not_ including nul
_and_
(b) You have a file that is PATH_MAX characters long. In that case the last character will get truncated unnecessarily. (Anything greater than PATH_MAX will get truncated anyways) 

Not too say the whole concept of PATH_MAX is flawed anyways. I can live with both PATH_MAX and PATH_MAX+1.

---
So finally:
- I simply reverted my changes to other people's code. So the hardcoded 1024 stays and usages of PATH_MAX and PATH_MAX+1 stays. Making this consistent can come later.
- I changed my own use of PATH_MAX to PATH_MAX+1 for the `pwd` feature.

lenormf: Got it, I created #4434 for this specific topic.

danr:
I see no reason why this should be in the c++ code since, as you point out, it's easy to access from kakscript:

```kak
define-command print-working-directory -docstring %{
  print the server's current working directory to the status line
} %{
  echo %sh{echo $PWD}
}
alias global pwd print-working-directory
```

Do we have any such utilities in the `rc` scripts where it fits next to?

sidkshatriya:
@danr I would very much love it if `pwd` was available to users. It does not matter at this point if it defined via C++ code or as a kakscript. I would suggest that you submit your snippet as a fresh PR in the hope that it might be accepted.

Some people use `kak` on a per-project basis. They never change their directory so don't see the need for `pwd`. I may `cd` to various directories on single session so I feel the need for this command.

---

I defined `pwd` in C++ for performance reasons. `pwd` is simple enough that it can be implemented in C++ once and then forgotten forever. I don't forsee it's implementation ever changing. If something is likely to never/rarely change, its good to implement in native code -- just my opinion.
 
I don't know if you have followed the history of `git`. Initially some git commands were built using Perl. Later they got translated into C, mostly for performance reasons.

Performance is about making hundred's of small decisions. They add up eventually to give you a fast system.

However, I don't know if @mawww and some others are interested in having `pwd` in C++ core or as part of the default kakscripts. I can only hope!

alexherbo2:
If we were to add the current working directory to the status line, it could be nice to avoid shell calls.

It shouldn’t be costly as a one-shot command, but for example in the `modelinefmt` option.

See #2330.

mawww:
I have not made my mind on whether we need to expose the current working directory, but if we do, it wont be through a pwd command that only caters to a single use case, it will be with a `%val{cwd}` that can be used for many other purposes.

One shell call at each redraw should be unnoticeable to users, so I am not sure the modeline example is very convincing: Accepting this means we could justify any new expension by saying we do not want to pay for a shell call to put it in the modeline.

sidkshatriya:
> One shell call at each redraw should be unnoticeable to users

One must also worry about kakoune debuggability and overall efficiency.

I've seen it generally mentioned in the kakoune community that shell calls are cheap and generally not a problem. If you run `kakoune -debug profile` you are correct. I haven't done any statistics but I'm noticing shell calls can be in the low single digit milliseconds and many times even below 1 ms.

But eventually the shell calls do add up. Thousands upon thousands of syscalls are being made by kakoune just because we don't have important environment information available to us. The Linux kernel needs to execute a _whole_ program just to get access to a small piece of information.

If you record kakoune thus: `rr record kak` you will see how slow kakoune is under the rr debugger. This is because rr is recording the whole execution tree including the shell programs. But then do `rr record kak -n` you will suddenly see the recording much faster. Many of these shell calls could be avoided with more information available within kakoune itself, improving kakoune's debuggability over longer sessions.

I find neovim's recordability under rr much better. Try `rr record nvim` and see how faster things are. This gives me a hint of the underlying syscall load.

Shell calls are part of kakoune's interesting interoperability story and we will always have a lot of them. But when we can, lets reduce them, is my thought...

sidkshatriya:
P.S. Presence of `%val{cwd}` would be great!

PR #4438, opened 2021-11-14
JSON UI only shows stdin when connecting to an existing session #4432

### LINKED ISSUE #4432: [BUG] JSON UI only shows stdin when connecting to an existing session
### Version of Kakoune

v2021.11.08

### Reproducer

Create a script that will print text without exiting:

```sh
# make_commands.sh
while :
do
    echo '{"jsonrpc":"2.0","method":"keys","params":["e"]}'
    sleep 1
done
```

Then, create a named kak session

```sh
kak -ui "json" -s "test"
```

Then, try to connect with a second client that has a non-tty stdin:

```sh
sh ./make_commands.sh | kak -ui "json" -c "test"
```

### Outcome

the second client shows a FIFO buffer where the contents of stdin are displayed as text rather than interpreted as commands.

### Expectations

If you pipe input into a normal kak instance, it will display the text from stdin in a fifo buffer. 
```sh
sh ./make_commands.sh | kak 
```
(The status line shows ` *stdin* 1:1 [scratch][fifo] 1 sel - client0@[1608978]`.)

If you pipe text to stdin of a kak instance created with `-ui json`, it will know to treat the input as JSON RPC commands and open the normal scratch buffer.
```sh
sh ./make_commands.sh | kak -ui "json"
```
```json
{ "jsonrpc": "2.0", "method": "draw_status", "params": [[], [{ "face": { "fg": "default", "bg": "default", "underline": "default", "attributes": [] }, "contents": "*scratch* 1:1 " }, { "face": { "fg": "black", "bg": "yellow", "underline": "default", "attributes": [] }, "contents": "[scratch]" }, { "face": { "fg": "default", "bg": "default", "underline": "default", "attributes": [] }, "contents": " " }, { "face": { "fg": "blue", "bg": "default", "underline": "default", "attributes": [] }, "contents": "1 sel" }, { "face": { "fg": "default", "bg": "default", "underline": "default", "attributes": [] }, "contents": " - client0@[1610796]" }], { "fg": "cyan", "bg": "default", "underline": "default", "attributes": [] }] }
```

But, if you pipe into the stdin of a JSON UI client that is connecting to an existing session, it goes back to the original behaviour where the input is interpreted as text, not commands, and is displayed (the status line is `[fifo]`).

```sh
sh ./make_commands.sh | kak -ui "json" -c "test"
```
```json
{ "jsonrpc": "2.0", "method": "draw_status", "params": [[], [{ "face": { "fg": "default", "bg": "default", "underline": "default", "attributes": [] }, "contents": "*stdin-1* 1:1 " }, { "face": { "fg": "black", "bg": "yellow", "underline": "default", "attributes": [] }, "contents": "[scratch][fifo]" }, { "face": { "fg": "default", "bg": "default", "underline": "default", "attributes": [] }, "contents": " " }, { "face": { "fg": "blue", "bg": "default", "underline": "default", "attributes": [] }, "contents": "1 sel" }, { "face": { "fg": "default", "bg": "default", "underline": "default", "attributes": [] }, "contents": " - client1@[test]" }], { "fg": "cyan", "bg": "default", "underline": "default", "attributes": [] }] }
```

This makes multiple clients useless for JSON UI because you cannot interact with them by sending commands. Instead, it should interpret stdin as RPC commands and not display them to the user

### Additional information

I am using Ubuntu 20.04.2 LTS.

I first encountered this error running kak from a Rust program with 

```rust
Command::new("kak")
    .args(&["-ui", "json"])
    .args(&["-c", "test"])
    .stdout(Stdio::piped())
    .stdin(Stdio::piped())
    .spawn()
    .unwrap()
```
but I was able to reproduce it in the terminal (Alacritty) as well.

This error seems to be a regression. A very old version of kak (probably from a couple years ago -- I don't remember the version number) did not have this issue but a late 2020 or early 2021 version introduced it.

sidkshatriya:
Fixes #4432

json-ui is not intended to be a user interactive session
so don't worry about making the tty your stdin if fd 0
is not a tty.

This allows json-rpc commands sent via stdin to be acted up rather
than sent to a fifo (which is the default behavior for kakoune).

Does not change the behavior for non-json ui sessions

cc: @Screwtapello @QuinnFreedman

P.S.  This approach is simple and does the job. There probably are more approaches

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -692,7 +692,11 @@ int run_client(StringView session, StringView name, StringView client_init,
     try
     {
         Optional<int> stdin_fd;
-        if (not isatty(0))
+        // json-ui (or dummy) is not intended to be user interactive.
+        // So only worry about making the tty your stdin if:
+        // (a) ui_type is Terminal, *and*
+        // (b) fd 0 is not interactive.
+        if (ui_type == UIType::Terminal && not isatty(0))
         {
             // move stdin to another fd, and restore tty as stdin
             stdin_fd = dup(0);


sidkshatriya:
Detailed analysis on https://github.com/mawww/kakoune/issues/4432#issuecomment-968367128

src/main.cc
```diff
@@ -692,7 +692,10 @@ int run_client(StringView session, StringView name, StringView client_init,
     try
     {
         Optional<int> stdin_fd;
-        if (not isatty(0))
+        // json-ui is not intended to be a user interactive session
+        // so don't worry about making the tty your stdin if fd 0
+        // is not a tty.
+        if (not isatty(0) && ui_type != UIType::Json)
```
Screwtapello, line 698: In `create_local_ui()` Kakoune only reads stdin into a buffer for `UIType::Terminal`, rather than "everything but `UIType::Json`:

https://github.com/mawww/kakoune/blob/e7100dc87434d933fd8268e5bf70080b627750c5/src/main.cc#L644-L645

...so we should copy the same logic here too.

Screwtapello:
I think there's probably a much cleaner refactoring of `main.cc` which would slurp stdin in one place, instead of having it happen at different places depending on whether we're starting a client or a server. However, this is a simple change that works, we'll see what mawww says.

src/main.cc
sidkshatriya: To be consistent with the if condition you pointed out,

I changed:
```c++
if (not isatty(0) && ui_type != UIType::Json)
```

to:
```
if (ui_type == UIType::Terminal && not isatty(0))
```

sidkshatriya:
Improved the `if` condition so that client and server both are consistent i.e. have separate code paths for Terminal ui type / non-Terminal ui type.

Ready to merge if the approach that this fix takes is deemed acceptable.

Issue #2110, opened 2018-06-06
Consistency of identifier symbols in commands and aliases


lenormf:
Hi,

Command names have to follow the alpha-num (plus hyphen `-` and underscore `_`) alphabet, but aliases don't.

A user asked today, on IRC, if they could create a command whose name would be trailed by an exclamation point `!`, and seeing that it wasn't possible, proceeded to create an alias.

Is this discrepancy there on purpose, for instance for commands to have a name spelled out in plain english, or is it an oversight? Note that most (all?) punctuation signs are allowed in an alias' name.

HTH.

PR #4441, opened 2021-11-16
Allow for periods in commands, modes, modules names, and session names

### LINKED ISSUE #4381: Allow dots in session names

fennewald:
Closes #4381 

This allows for dots in commands, modes, modules, and sessions. The original issue only asked for this feature in session names. I can amend the commit if desired to be limited to only session names.

### DIFF
--- a/src/unicode.hh
+++ b/src/unicode.hh
@@ -92,7 +92,7 @@ inline bool is_basic_digit(Codepoint c) noexcept
 inline bool is_identifier(Codepoint c) noexcept
 {
     return is_basic_alpha(c) or is_basic_digit(c) or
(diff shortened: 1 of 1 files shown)


lenormf:
Why only dots and not any punctuation signs (except maybe slashes)? After all the reasons to have a dot in e.g. a session name are probably similar to the ones to have colons for example.

In any case I don’t think the changes should be done to `is_identifier`, as these don’t have dots in them.

Delapouite:
lenormf, it reminds me of one your previous discussion on the topic of allowed characters: https://github.com/mawww/kakoune/issues/2110

PR #4442, opened 2021-11-16
Fix display of lint messages without columns

### LINKED ISSUE #4336: Linting doesn't work if no column is provided by linting tool
### Steps 
1. Set the linting tool to something that uses the [flycheck](https://www.flycheck.org/) format (e.g. `set-option window lintcmd "mix credo list --config-file=.credo.exs --format=flycheck"` for an elixir project with credo)
1. Provoke a linting error that produces a linting error with no specified column (e.g. `foo.ex:42: error: use rust instead`)
1. Lint the buffer

### Outcome 
The Linting output is ignored completely.

### Expected 
The error should be handled, despite the absence of a column.

jtrees:
Fixes #4336.

### DIFF
--- a/rc/tools/lint.kak
+++ b/rc/tools/lint.kak
@@ -145,7 +145,7 @@ define-command \
 
                 # Mention the column where this problem occurs,
                 # so that information is not lost.
-                msg = msg "(col " $3 ")"
+                msg = msg " (col " $3 ")"
 
                 # Messages will be stored in a line-specs option,
                 # and each record in the option uses "|"
--- a/rc/tools/lint.kak
+++ b/rc/tools/lint.kak
@@ -89,12 +89,14 @@ define-command \
                     '
                         BEGIN { OFS=":"; FS=":" }
 
-                        /:[1-9][0-9]*:[1-9][0-9]*:/ {
+                        /^[^:]+:([1-9][0-9]*:){1,2} / {
                             $1 = ENVIRON["kak_bufname"]
-                            if ( $2 == 1 ) {
+                            $2 += line_offset
+                            if ( $3 ~ /^ / ) {
+                                $2 = $2 FS
+                            } else if ( $2 == 1 + line_offset ) {
                                 $3 += first_line_byte_offset
                             }
-                            $2 += line_offset
                             print $0
                         }
                     ' >>"$dir"/result
@@ -119,7 +121,7 @@ define-command \
                 warning_count = 0
             }
 
-            /:[1-9][0-9]*:[1-9][0-9]*:/ {
+            /^[^:]+:[1-9][0-9]*:([1-9][0-9]*)?: / {
                 # Remember that an error or a warning occurs on this line..
                 if ($4 ~ /[Ee]rror/) {
                     # We definitely have an error on this line.
@@ -145,7 +147,9 @@ define-command \
 
                 # Mention the column where this problem occurs,
                 # so that information is not lost.
-                msg = msg " (col " $3 ")"
+                if ($3 != "") {
+                  msg = msg " (col " $3 ")"
+                }
 
                 # Messages will be stored in a line-specs option,
                 # and each record in the option uses "|"


rc/tools/lint.kak
```diff
@@ -89,7 +89,16 @@ define-command \
                     '
                         BEGIN { OFS=":"; FS=":" }
 
-                        /:[1-9][0-9]*:[1-9][0-9]*:/ {
+                        # without column
+                        /^[^:]+:[1-9][0-9]*: / {
+                            $1 = ENVIRON["kak_bufname"]
+                            $2 += line_offset
+                            $2 = $2 FS "x";
```
mawww, line 96: can we get away with an empty string instead of a magic 'x' here ?

rc/tools/lint.kak
```diff
@@ -89,7 +89,16 @@ define-command \
...
+                            print $0
+                        }
+
+                        # with column
+                        /^[^:]+:[1-9][0-9]*:[1-9][0-9]*: / {
```
mawww, line 101: could not capture groups do the trick to mostly share the code between with/without column ?

mawww:
@lenormf Can you expand on the issue you see with this ?

lenormf:
I’m not convinced this error format is ubiquitous enough to warrant the complexity in the plugin.

Plus our stance has been that if the output format is different than the “emacs” one, the linter should be wrapped in a shell script or post-processed in `lintcmd` directly.

The first option is not convenient because it makes users create random local scripts, the second requires knowledge of how the plugin runs `lintcmd` (`run() { … } && run`) — so I’d rather see #2336 merged so we can avoid handling corner cases like the `flycheck` format in Awk.

jtrees:
> Plus our stance has been that if the output format is different than the “emacs” one, the linter should be wrapped in a shell script or post-processed in `lintcmd` directly.

Is this really the case though? I tried to find a specification for the format but couldn't find one. In fact, I think I found a snippet somewhere in the docs of the "emacs one" that implied the format without a column is also valid. I am simply recalling from memory though, don't have a source ready.

If it's not, I might simply try to get this changed upstream (in [credo](https://github.com/rrrene/credo)) instead.

rc/tools/lint.kak
jtrees: Yes, good idea: b83b576028685b6d160f1fb2e7ca2aae5c0f5a6f

rc/tools/lint.kak
jtrees: I'm not too familiar with awk and it looked like working with capture groups is either non-trivial or non-portable.

Here's my attempt at making the code more DRY: b83b576028685b6d160f1fb2e7ca2aae5c0f5a6f

Does this work for you?

rc/tools/lint.kak
```diff
@@ -89,12 +89,14 @@ define-command \
                     '
                         BEGIN { OFS=":"; FS=":" }
 
-                        /:[1-9][0-9]*:[1-9][0-9]*:/ {
+                        /^[^:]+:([1-9][0-9]*:){1,2} / {
                             $1 = ENVIRON["kak_bufname"]
-                            if ( $2 == 1 ) {
+                            $2 += line_offset
+                            if ( $3 ~ /^ / ) {
+                                $2 = $2 FS
```
mawww, line 96: why do we need to add that `:` to $2 ? cant we use this logic (`$3 ~ /^ /`) in the subsequence awk script to detect if we had a column ?

rc/tools/lint.kak
```diff
@@ -89,12 +89,14 @@ define-command \
...
+                            } else if ( $2 == 1 + line_offset ) {
```
mawww, line 97: I dont like adding line offset then comparing with line_offset + 1, it seems convoluted, surely we can just store that information in a boolean like variable if we need to check it before modifying $2.

mawww (commented):
A few additional changes i'd like.

jtrees:
Sorry for abandoning this. I have moved on from Kakoune (to Helix) and lost interest.

PR #4451, opened 2021-11-24
Add `-debug shell-trace` to add `set -x` to every shell script kak executes

sidkshatriya:
_Note_: This is a separate feature from #4450 and can be merged independently.

This commit teaches Kakoune to trace every shell script it
executes if you start Kakoune as `kak -debug shell-trace`.
The results of the trace of each script can be seen in the `*debug*`
buffer.

The alternative is to explictly add `set -x` inside `%sh{}` that
we are interested in. However, this just makes things a bit
more convenient.

Most likely you will want to use this with `shell` i.e.
`kak -debug 'shell|shell-trace'`.

(The `shell` option shows the script that Kakoune sent for execution
while the `shell-trace` option traces the execution via `set -x`
feature)

### DIFF
--- a/src/option.hh
+++ b/src/option.hh
@@ -74,12 +74,13 @@ using TimestampedList = PrefixedList<size_t, T>;
 
 enum class DebugFlags
 {
-    None     = 0,
-    Hooks    = 1 << 0,
-    Shell    = 1 << 1,
-    Profile  = 1 << 2,
-    Keys     = 1 << 3,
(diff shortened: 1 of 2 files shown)


mawww:
Not sure how useful this is, wanting to trace every single shell call sounds like it would be pretty rare. If we really need it I'd probably rather go with some way of specifying some preamble text to include in shell commands than hard code support for `set -x`.

PR #4453, opened 2021-11-25
readme: update build instructions

71:
After a small hour trying to build Kakoune (mostly due to [`compare`](https://en.cppreference.com/w/cpp/header/compare) being very recent), I think I got the requirements right.

Not sure about the required `libc++`, I couldn't get Kakoune to build with it no matter which version I tried  ¯\\\_(ツ)\_/¯

Also used https://godbolt.org/ to find which version of GCC and Clang managed to compile `#include <compare>`.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -103,8 +103,8 @@ Building
 
 Kakoune dependencies are:
 
- * A {cpp}17 compliant compiler (GCC >= 8 or clang >= 6) along with its
-   associated {cpp} standard library (libstdc{pp} or libc{pp})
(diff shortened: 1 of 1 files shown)


Screwtapello:
`libstdc++` is GCC's C++ standard library implementation.

`libc++` is Clang's C++ standard library implementation, but it generally prefers to re-use the most common C++ library on a platform for compatibility with other compilers. So, on Linux it'll use `libstdc++` and on Windows it'll use whatever MSVC uses.

71:
Thanks for the explanation, it does make things clearer! I've also heard that `libc++` is the default on Mac which is why I wanted to get its minimum supported version, but as I don't have a Mac I can't do that.

PR #4455, opened 2021-11-27
Added basic indent guides to show-whitespaces highlighter.

Calum-J-I:
This adds support for basic indent guides in the show-whitespaces highlighter. The implementation is pretty simple. Just replaces every the whitespace hightlighter token with a special for each column that is a multiple of indentwidth and is not proceeded by a non-whitespace character.
indent-guides


Calum-J-I:
There is a quirk in this implemenation in that it only checks what's on screen for the "only_whitespace" condition. So if you have no word wrap and a wide screen of text with some whitespace then it will show the alignment lines. I personally consider that acceptable but opinions may differ. I'm sure the off screen lines can be checked since the regex highlighters would need to be able to do that.

In pracice the scenario would be rare but I can think of a few places it could happen, e.g. SQL query outputs to ascii tables which often contain a lot of whitespace.

Some users might also want the indent guidelines at the end of each tab character and RHS of the column before the next indent width. This would have the added benefit of preserving the tab character. Could add this behaviour as an option. Maybe `-lguide` and `-rguide`.

src/highlighters.cc
```diff
@@ -1034,15 +1034,16 @@ const HighlighterDesc show_whitespace_desc = {
         { "tabpad", { true, "append as many of the given character as is necessary to honor `tabstop`" } },
         { "spc",    { true, "replace spaces with the given character" } },
         { "lf",     { true, "replace line feeds with the given character" } },
-        { "nbsp",   { true, "replace non-breakable spaces with the given character" } } },
+        { "nbsp",   { true, "replace non-breakable spaces with the given character" } } ,
+        { "guide",  { true, "use given character an whitespace indent guide" } } },
```
Screwtapello, line 1038:
```suggestion
        { "guide",  { true, "use given character as whitespace indent guide" } } },
```

doc/pages/highlighters.asciidoc
```diff
@@ -51,6 +51,9 @@ highlighter is replaced with the new one.
     *-tabpad* <separator>:::
         a one character long separator that will be appended to tabulations to honor the *tabstop* option
 
+    *-guide* <separator>:::
+        a one character long separator that, if provided, will replace every *indentwidth* whitespace char
```
Screwtapello, line 55: Does it replace *every* whitespace char, or only consecutive leading spaces?

Calum-J-I: consecutive leading spaces/tabs. Although I mentioned in a comment that there's a caveat. Current implementation only iterates over what's visible on screen.

src/highlighters.cc
```diff
@@ -1088,28 +1100,45 @@ struct ShowWhitespacesHighlighter : Highlighter
                         if (it != end)
                             atom_it = line.split(atom_it, it.coord());
 
+                        const int is_indent_guide = only_whitespace and whitespace_col % indentwidth == 0;;
```
Calum-J-I, line 1103: bool*

src/highlighters.cc
```diff
@@ -1056,27 +1057,43 @@ struct ShowWhitespacesHighlighter : Highlighter
...
         const auto& buffer = context.context.buffer();
-        auto win_column = context.setup.window_pos.column;
+        const auto win_column = context.setup.window_pos.column;
+        auto line_index = context.setup.window_pos.line;
+        const LineCount buffer_end = buffer.end().coord().line;
+
         for (auto& line : display_buffer.lines())
         {
+            bool only_whitespace = m_add_guide and
+                                   line_index < buffer_end ? all_whitespace(buffer[line_index++], win_column) : true;
```
Calum-J-I, line 1086: was getting segfaults on buffer[line_index++] until I added this check. Feel like their should be a cleaner way to do this though.

Calum-J-I:
Now checks off screen whitespace.

src/highlighters.cc
Calum-J-I: Actually, this is probably because of line wrapping isn't it. I think current implementation is looking at the wrong buffer lines for all lines after a wrapped one.

src/highlighters.cc
```diff
@@ -1056,27 +1057,43 @@ struct ShowWhitespacesHighlighter : Highlighter
...
+
         const auto& buffer = context.context.buffer();
-        auto win_column = context.setup.window_pos.column;
+        const auto win_column = context.setup.window_pos.column;
+        auto line_index = context.setup.window_pos.line;
+        const LineCount buffer_end = buffer.end().coord().line;
+
         for (auto& line : display_buffer.lines())
         {
+            bool only_whitespace = m_add_guide and
```
mawww, line 1085: I think this variable should be called `in_indent`, and I dont think it needs to depend on m_add_guide.

src/highlighters.cc
```diff
@@ -1056,27 +1057,43 @@ struct ShowWhitespacesHighlighter : Highlighter
...
+                                   (line_index < buffer_end ? all_whitespace(buffer[line_index++], win_column) : true);
```
mawww, line 1086: This looks wrong, a display line does not 1:1 map to a buffer line, it might contain multiple buffer lines, or only part of one. I suspect this should be done at the atom level, as we know the atom buffer range at that point and can check if we are inside the indent part of the buffer line.

src/highlighters.cc
```diff
@@ -1088,28 +1105,55 @@ struct ShowWhitespacesHighlighter : Highlighter
...
 
+                        const bool is_indent_guide = only_whitespace and whitespace_col % indentwidth == 0;
+
                         if (cp == '\t')
                         {
                             const ColumnCount column = get_column(buffer, tabstop, coord);
                             const ColumnCount count = tabstop - (column % tabstop) -
                                                       std::max(win_column - column, 0_col);
-                            atom_it->replace(m_tab + String(m_tabpad[(CharCount)0], count - m_tab.column_length()));
+                            if (is_indent_guide)
```
mawww, line 1115: a ternary to select m_guide or m_tab would work here.

src/highlighters.cc
```diff
@@ -1088,28 +1105,55 @@ struct ShowWhitespacesHighlighter : Highlighter
...
+        for (auto it = line.begin(); it != line.end() and it - line.begin() <= window_lhs; ++it)
+        {
+            if (*it != '\t' and *it != ' ')
+                return false;
+        }
+        return true;
+    }
+
+    const String m_tab, m_tabpad, m_spc, m_lf, m_nbsp, m_guide;
+    const bool m_add_guide;
```
mawww, line 1156: I would prefer to use `m_guide.empty()` to check if we want to use the guide or not.

mawww (commented):
The implementation seems like it will break when interacting with wrapping and folding. It also cannot fully draw the indent guide that we would expect on empty lines (as we do not have any buffer content to replace there).

Maybe a separate highlighter would be better for this, but it would need to interact correctly with  show_whitespace and in particular with tab handling.

Calum-J-I:
@mawww I think it _might_ be possible to have two seperate show_whitespace and indent_guide highlighters because indent_guide should use the underlying buffer contents while show_whitespace mutates the display atoms. I'll have a play around. I do think if you want a full implementation with "virtual text" to cover empty lines it would be preferable to have to seperate highlighters.

As an aside is there an idiomatic way to reserve and mutate space for a String? I couldn't see an obvious way to write something like
```
const codepoint[] unicode_array = codepoint[n];
for (int i =0; i < something; i += 4) {
    unicode_array[i] = "|"
}
const string = String{unicode_array};
```

maybe correct constructor is
```
const string = String{unicode_array.begin(), unicode_array.end()}
```
?

krobelus:
> As an aside is there an idiomatic way to reserve and mutate space for a String? I couldn't see an obvious way to write something like

I believe that Kakoune always uses UTF-8 internally (could be wrong though).
So we never construct an array of codepoints, but always just an array of UTF-8-encoded bytes.
To mutate string data, you'd allocate a large enough buffer and then loop over that with
utf8::codepoint, and then directly replace the bytes.

Calum-J-I:
Bit of a delay on this, I'm happy enough with the end result, have been using it a fair bit myself. Only thing I'm personally not too sure about is that extra DisplayAtom type. It's not strictly necessary but I like to know that the atoms I'm replacing are ReplacedRanges from the whitespace highlighters and not something else.

My preferred highlighter style is

```
add-highlighter window/ indent-guide -spcpad ' ' -use-actual
add-highlighter window/ show-whitespaces
```

haven't run into any crashes with this version yet but after merging to master I got reliable crashes in the ShowWhitespace highlighter with offscreen tab chars. Don't think it's this code's fault since I saw similar crashes on fresh clone of master but you never know.

Calum-J-I:
https://asciinema.org/a/HLFq0aygkijPG4MPF64Ty1WtS

PR #4456, opened 2021-11-27
define `%val{jumplist}` and `%val{jumplist_pos}`

### LINKED ISSUE #4452: [REQUEST] Jump list access in scripting (like `%val{jumplist}`)
### Feature

[Jump list](https://github.com/mawww/kakoune/blob/716f1f967a2d2fc22390b05d23c62c444f68ebff/README.asciidoc#jump-list) is good for digging code - it gives 'go back' in browsing history. I sometimes want to see full list of it, so I can get quick view of where I am and why I'm here. Unfortunately I can't find `%val{jumplist}` or similar.

### Usecase

_No response_

sidkshatriya:
`echo %val{jumplist}` will return a list of jump locations. Fixes #4452

e.g.
```
.git/COMMIT_EDITMSG 7.1 *grep* 1.47 *grep* 1.1 src/main.cc 231.1 *debug* 4.65
```

Each jump location is `<filename> <line>.<column>`

`%val{jumplist_pos}` reports the currently active jumplist position as a 0 based index.

This is consistent with indexing adopted by the user interface messages when doing `<c-i>`, `<c-o>`

*Motivation*
- I am not aware of a way of obtaining a jumplist in Kakoune. This
  information should be exposed by Kakoune so that plugins/scripts can show
  jump history. You would conceivably want to see a jumplist and
  navigate to a particular entry in it. There could be other interesting
  use cases also.
- It is not programmatically possibly to obtain the jumplist even
  if you want. If you try doing `eval -draft %{ exec <c-o> ; echo %val{buffname} }`
  this will give an error. Newly created draft contexts start off
  with an empty jumplist. They don't inherit the jumplist of the
  current client context.
- This is borne out of a user request on #4452

### DIFF
--- a/src/context.hh
+++ b/src/context.hh
@@ -126,6 +126,7 @@ public:
     Flags flags() const { return m_flags; }
 
     JumpList& jump_list() { return m_jump_list; }
+    const JumpList& jump_list_readonly() const { return m_jump_list; }
     void push_jump(bool force = false)
     {
         if (force or not (m_flags & Flags::Draft))
--- a/src/main.cc
+++ b/src/main.cc
@@ -214,6 +214,28 @@ static const EnvVarDesc builtin_env_vars[] = { {
         "history_id", false,
         [](StringView name, const Context& context) -> Vector<String>
         { return {to_string((size_t)context.buffer().current_history_id())}; }
+    },{
+        "jumplist", false,
+        [](StringView name, const Context& context)  -> Vector<String>
+        {
+            Vector<String> vec{};
+            auto jumplist = context.jump_list_readonly().get_as_list();
+            for (auto item = jumplist.begin(); item != jumplist.end(); item++) {
+                vec.push_back(item->buffer().display_name());
+                auto cur = item->main().cursor();
+                vec.push_back(format("{}.{}", cur.line + 1, cur.column + 1));
+            }
+            return vec;
+        }
+    },{
+        "jumplist_pos", false,
+        [](StringView name, const Context& context)  -> Vector<String>
+        {
+            Vector<String> vec;
+            // To be consistent with user facing <c-i> <c-o>, the index is 0 based.
+            auto current = context.jump_list_readonly().current_index();
+            return {format("{}", current)};
+        }
     }, {
         "selection", false,
         [](StringView name, const Context& context) -> Vector<String>


Calum-J-I:
There's a list of all `%val` expansions you'd need to update
https://github.com/mawww/kakoune/blob/master/doc/pages/expansions.asciidoc

sidkshatriya:
@Calum-J-I Agree -- documentation would need to be updated. However, I'm first curious about how the kakoune community feels about this change. Once the feature is approved, I can update the documentation. What are your views BTW?

One design point is that instead of two variables `%val{jumplist}` and `%val{jumplist_pos}` we could have only one `%val{jumplist}` with the jumplist position as the very first item in the jumplist.

```
2 .git/COMMIT_EDITMSG 7.1 *grep* 1.47 *grep* 1.1 src/main.cc 231.1 *debug* 4.65
```

Note the `2` -- this is the index of the current jump. This avoids having two vals.

But its also simpler to understand the jumplist without the initial 2 and have a separate val.

Screwtapello:
Marks (as saved to a register with `<Z>`) store the buffer name and the buffer timestamp so they can be restored after a buffer has been edited; if the jumplist has something similar it might be useful to expose it.

Some precedents for "which is the current/active item of a list":

- Marks store the main selection index as a number in the header item
- `%val{selections_desc}` puts the main selection as the first item in the list, with no special header, but the selections are otherwise in document order so it's easy to figure out where the main selection "should" be
- The `#` register does not indicate which is the main selection
- `%val{buflist}` has `%val{buffile}` to indicate which is the active buffer (although `%val{buffile}` was first and `%val{buflist}` was added later, I think)

Some precedents for item formatting:

- Most list-of-records expansions use the same character for every field (Mark headers use `name@timestamp@selection_index`, completions use `text|command|desc`) or different characters based on precedence (most things with a buffer coordinate use `.` to separate the line and column, even if they use different characters for other fields)
- `%val{window_range}` is the only expansion I can find that separates fields with whitespace
- Mark headers put the buffer filename first and the other fields at the end
- Most other syntaxes put the fixed-format fields first and the free-text field at the end (`%val{uncommitted_modifications}`, `line-specs` options, `range-specs` options)

sidkshatriya:
@Screwtapello 
Thanks for this comprehensive summary/overview on textual representations of kakoune structures. This is worthy of a wiki entry if you have the time. It will be useful in future discussions of a similar nature.

So there are two aspects here:
- Do we want to expose this as a value? @mawww ? I'll hold off on any further code changes unless we decide we do.
- If we want it, how do we represent it?

 > Marks store the main selection index as a number in the header item

If we adopt this approach we would have a single `%val{jumplist}` at the cost of additional script inconvenience.

```
.git/COMMIT_EDITMSG@1 7.1 *grep* 1.47 *grep* 1.1 src/main.cc 231.1 *debug* 4.65
```
Note the `@1` to indicate the active jumplist item (0 based index).

> %val{selections_desc} puts the main selection as the first item in the list,

Order is important in a jump list so is not applicable to our current use-case.

> %val{window_range} is the only expansion I can find that separates fields with whitespace

bash naturally splits on spaces in loops and other places. I like the separated by space representation which is why I choose it. I prefer it to `|`. 

> Most other syntaxes put the fixed-format fields first and the free-text field at the end (%val{uncommitted_modifications}, line-specs options, range-specs options)

The jumplist is pretty structured and I think a list of `<buffer-name> <row>.<column>` feels quite natural.

`<row>.<column>` is consistent with `%val{selection_desc}`, `%val{selections_desc}`. However `<row> <column>` (with the space) would make things marginally easier in bash? We actually don't need a `.` because there will be only two numbers and not a variable number of entries (in which it becomes difficult to visually keep pairing things together and a `.` becomes useful)

> %val{buflist} has %val{buffile} to indicate which is the active buffer (although %val{buffile} was first and %val{buflist} was added later, I think)

I actually like 2 val solution `%val{jumplist}` and `%val{jumplist_pos}` because you could have a jumplist that remains the same but only the jumplist position varies. Basically your two variable representation gives you an array (which can remain same) _and_ a separate  current position (which is an index into it) that can change. It may not be a good idea to combine these two things which are different.

PR #4457, opened 2021-11-29
Add syntax highlighting for `git-revise`

evanrelf:
> `git revise` is a `git` subcommand to efficiently update, split, and rearrange commits. It is heavily inspired by `git rebase`, however it tries to be more efficient and ergonomic for patch-stack oriented workflows.

https://github.com/mystor/git-revise

### DIFF
--- a/rc/filetype/git.kak
+++ b/rc/filetype/git.kak
@@ -14,7 +14,11 @@ hook global BufCreate .*git-rebase-todo %{
     set-option buffer filetype git-rebase
 }
 
-hook global WinSetOption filetype=git-(commit|notes|rebase) %{
+hook global BufCreate .*git-revise-todo %{
+  set-option buffer filetype git-revise
+}
+
+hook global WinSetOption filetype=git-(commit|notes|rebase|revise) %{
     require-module "git-%val{hook_param_capture_1}"
 }
 
@@ -33,6 +37,10 @@ hook -group git-rebase-highlight global WinSetOption filetype=git-rebase %{
     hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/git-rebase }
 }
 
+hook -group git-revise-highlight global WinSetOption filetype=git-revise %{
+  add-highlighter window/git-revise ref git-revise
+  hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/git-revise }
+}
 
 provide-module git-commit %{
 add-highlighter shared/git-commit regions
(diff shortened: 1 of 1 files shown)


lenormf:
As this is a third party command, support for it should probably remain in a separate plugin or as part of your user configuration.

evanrelf:
@lenormf 

> As this is a third party command, support for it should probably remain in a separate plugin or as part of your user configuration.

All the files under [`rc/filetype/`](https://github.com/mawww/kakoune/tree/master/rc/filetype) are for third-party tools from Kakoune's perspective. Do you mean third-party from Git's perspective?

If that's the case, I'd be happy to move this into its own file at `rc/filetype/git-revise.kak` instead of bundling it with the existing Git support.

sidkshatriya:
Initially I thought that @lenormf was correct in suggesting a separate plugin. However, this seems to be just syntax highlighting and _not_ some sort of generic integration plugin. And we have syntax highlighting for all sorts of things bundled with kakoune. So I think a case can be made to include this with Kakoune.

> If that's the case, I'd be happy to move this into its own file at rc/filetype/git-revise.kak instead of bundling it with the existing Git support.

Yes it would be better if this change did not disturb `git.kak`.

(Interestingly @krobelus who is active in the kakoune community is also an active contributor to `git-revise`)

lenormf:
Yea, I mean it’s not available with Git out the box, and so shouldn’t be supported by `git.kak`.

There are so many third party extensions out there that supporting one out of the box would provide with a reason for all others to be also supported (on account of the precedent), burdening the codebase with every extension under the sun that users might not have installed.

The idiomatic way to handle that situation is via user configuration or plugin, which are generally quite simple to write, and don’t put the responsibility of maintaining them on the project owner.

If `git-revise` was so popular that it could be expected to be installed by most Git users, maybe it could be considered for inclusion under its own `git-revise.kak` script (or maybe `git.kak` directly), but I don’t think we’re there yet.

sidkshatriya:
We have the [exherbo filetype](https://github.com/mawww/kakoune/blob/master/rc/filetype/exherbo.kak) which is related to the linux distribution [exherbo](https://distrowatch.com/table.php?distribution=exherbo).

As a distribution, exherbo is very [niche](https://distrowatch.com/dwres.php?resource=popularity) so I think there is already a precedent for niche filetypes. (Disclaimer: I don't know much about exherbo).

As Kakoune is a cutting edge editor in some respects we should also consider a syntax highlighting for inclusion if the project is excellent/worthy in some way rather than simply just popular.

* Since this is just syntax highlighting I don't think there will be much of a maintenance burden
* I think author has already agreed to break out to `git-revise.kak` if necessary so it will be self contained

P.S. I don't know about `git-revise` myself so I can't vouch for its inclusion on grounds of being worthy/excellent.

evanrelf:
There's a lot of factors that would discourage me from upstreaming things:

- Is the code complicated, so it would burden the larger project?
- Is the code very opinionated, and not everyone might share my opinions?
- Does the code change existing functionality?
- Is the code in an incomplete state, or would it benefit from more changes?

The good news is, this change does none of those things! 🙂 I'm just mirroring the existing `git-rebase` syntax highlighting. And no further development is required; this is feature complete.

I also agree with @sidkshatriya that there is precedent for supporting less popular tools/languages/systems.

evanrelf:
I'm still interested in this being included in Kakoune by default, but for now I've made it available as a plugin: https://github.com/evanrelf/git-revise.kak

PR #4458, opened 2021-12-01
Add more patterns to .gitignore

sidkshatriya:
Add more patterns to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,11 @@
 *.pyc
 *.1-r
 .*.kak.*
+.cache
+compile_commands.json
+.idea
+**/.gdb_history
+**/.kak_history
 src/kak
 src/kak.debug
 src/kak.opt


lenormf:
Are those files generated as a result of cloning/building Kakoune on any supported system?

sidkshatriya:
This is my rationale for adding the 4 new entries to `.gitignore`

- `.cache` folder is generated when you're using clangd to work with kakoune code. clangd is going to be a popular way to look at kakoune C++ code
- Similarly `.idea` folder is generated when using the clion IDE
- You often need `compile_commands.json` when wanting to use clangd. You can generate it automatically often. But it should not be committed to the repo, hence putting it in gitignore
- `.kak_history` was mistakenly committed to the repo sometime ago (not by me :-) ). Here is my PR that removed it #4419 . It would be nice if our gitignore protected us from that!

sidkshatriya:
Added 1 more pattern to `.gitignore` i.e. `**/.gdb_history`

Issue #3751, opened 2020-09-20
Control-C does not cancel some shells running backgrounded eval to fifo buffer


krobelus:
I'm having trouble canceling my `:make` commands when using `dash` as shell.
For example if I do `set global makecmd 'ps aux | grep sleep; sleep 10'` and call `:make` repeatedly, or press <kbd>Control+C</kbd>. Old make jobs are not killed either way.

### Steps 

Run kak like below and press Control+C. When using bash this cancels the
background command, while other shells like dash or zsh it run for the full
five seconds.

```sh
# repro.kak
nop %sh{
  rm -f /tmp/fifo
  mkfifo /tmp/fifo
  (
    eval 'sleep 5' >/tmp/fifo 2>&1 &
  ) >/dev/null 2>&1 </dev/null
}
edit -fifo /tmp/fifo *my-fifo*
```

```sh
# can cancel bash
KAKOUNE_POSIX_SHELL=/bin/bash kak -n -e 'source repro.kak'
# cannot cancel dash
KAKOUNE_POSIX_SHELL=/bin/dash kak -n -e 'source repro.kak'
```

### Outcome 

`dash` keeps running

### Expected 

`dash` should be canceled on <kbd>Control+C</kbd>

Not sure if this is a bug in Kakoune, or the shell.

## Reduced program

The reduced C program below tries to run the shell the same way Kakoune does
and sends SIGINT to its process group. If it finishes in 0.5 seconds, then
the cancellation worked, otherwise it takes 1.5 seconds.

The commented line labeled `possible fix?` fixes cancellation for all shells,
but I'm not sure if it is safe: I imagine we could lose data from a finished
process whose file descriptor is not ready yet. Is that possible?
A similar fix would be possible somewhere in `ShellManager::eval`.

### pselect-deadlock.c

```c
#include <assert.h>
#include <errno.h>
#include <sys/types.h>
#include <sys/wait.h>
#include <unistd.h>

static void nop() {}

int main(int argc, char *argv[]) {
  // Pass the absolute path of a shell, like /bin/bash
  assert(argv[1] && "expect path to shell as argument");
  const char *shell = argv[1];

  // Launch the shell to run a background sleep for 1.5 seconds.
  //
  // When using bash the SIGINT below will stop the sleep and we terminate
  // in 0.5 seconds.
  //
  // With dash, zsh, busybox the sleep is not terminated, and pselect blocks
  // for 1.5 seconds.
[...]

PR #4473, opened 2021-12-14
Testing harness and compatibility fixes

sidkshatriya:
While trying to get kakoune `make test` to succeed on OpenBSD (See #4417), I found some issues. Here are the bugs and fixes (see commits).

**Some bugs are related to all systems and _not_ only OpenBSD.**

* **Bug fix: use only iswlower() in RankedMatch::is_word_boundary()**
  **Applicability**: All systems.

  This manifests itself in OpenBSD when you try to run kak compiled with debug=yes.

  Symptom:
  ```
  Fatal error: assert failed "RankedMatch{"find(1p)", "find"} < RankedMatch{"findfs(8)", "find"}" at    ranked_match.cc:255
  ...
  ```
  The fix is to simply use `iswlower()` instead of `islower()`. I think it `islower()` was used by mistake as the code around it does use `iswlower()`.

* **Use grep -E as OpenBSD grep does not like | in regexp otherwise**
  **Applicability**: OpenBSD

  This only manifests itself itself in OpenBSD as it does not use gnu grep by default. Fix by using `grep -E`.

* **Bug: The enabled test checks, though they exist are never actually run**
  **Applicability:** All systems
  
  We can check whether we want to run a test or not by executing the `enabled` script. However, the tests are run in the `/tmp` directory and the `enabled` file needs to be copied over in order to run. We were not copying the file so the `enabled` check was never actually carried out. Fix by copying the file over.

* **Bug: A `enabled` script had a bug in it**
  **Applicability:** All systems

  The enabled check was actually was not being run by the test infrastructure.

  Now that it is, we get the following error when trying to execute
`test/compose/select-display-columns`

  ```sh
  ./enabled: 2: [: 4: unexpected operator
  ```
  
  Fix the script

------
**OpenBSD specific issues**

* ```diff
  -MAÏS MÉLANGE BIENTÔT
  +MAïS MéLANGE BIENTôT
  ```
  (and many other test failures)
  **Resolution:**
  ```bash
  # I'm using bash in OpenBSD 7.0
  $ export LC_ALL=C.UTF-8
  # export LC_ALL=en_US.UTF-8
  # export LC_ALL=en_IN.UTF-8 
  # export LC_ALL=<your-favorite-one>.UTF-8
  # All these will work also. Basically ensure you have the suffix UTF-8
  # man locale in OpenBSD is helpful
  ```
* **There is a single test that still fails (more accurately it does not fail but hangs the test suite) on OpenBSD 7.0**. It's probably going to be complicated as its fifo related on openbsd. And that is `test/regression/3398-readonly-fifo-failure`. I didn't have time to investigate.

  **Temporary Workaround**: Pull this branch, (or if this branch is merged to master, the latest master) and then add the following file to the `test/regression/3398-readonly-fifo-failure` directory:
  ```bash
  $ cat > enabled <<-EOT
  #!/bin/sh
  false
  EOT
  $ chmod u+x enabled
  ```
  This will cause the test to be skipped.
 
  **Final Result of `make test` on OpenBSD 7.0** :
  ```
  ...
  3398-readonly-fifo-failure (disabled)
  ...
  Summary: 508 tests, 0 failures
  ```

### DIFF
--- a/test/compose/select-display-columns/enabled
+++ b/test/compose/select-display-columns/enabled
@@ -1,2 +1,2 @@
 #!/bin/sh
-[ $(echo -n "😄😊😉😍" | wc -m) == 4 ] && locale | grep LC_CTYPE | grep -qi 'utf-*8'
+[ $(echo -n "😄😊😉😍" | wc -m) = '4' ] && locale | grep LC_CTYPE | grep -qi 'utf-*8'
--- a/src/ranked_match.cc
+++ b/src/ranked_match.cc
@@ -174,7 +174,7 @@ RankedMatch::RankedMatch(StringView candidate, StringView query)
 static bool is_word_boundary(Codepoint prev, Codepoint c)
 {
     return (iswalnum((wchar_t)prev)) != iswalnum((wchar_t)c) or
-           (iswlower((wchar_t)prev) != islower((wchar_t)c));
+           (iswlower((wchar_t)prev) != iswlower((wchar_t)c));
 }
 
 bool RankedMatch::operator<(const RankedMatch& other) const
--- a/test/normal/keep-cmd-reg/rc
+++ b/test/normal/keep-cmd-reg/rc
@@ -1,1 +1,1 @@
-set-register a %{grep 'foo\|bar'}
+set-register a %{grep -E 'foo|bar'}
--- a/test/normal/keep-cmd/cmd
+++ b/test/normal/keep-cmd/cmd
@@ -1,1 +1,1 @@
-%<a-s>H$grep 'foo\|bar'<ret>
+%<a-s>H$grep -E 'foo|bar'<ret>
--- a/test/regression/3398-readonly-fifo-failure/rc
+++ b/test/regression/3398-readonly-fifo-failure/rc
@@ -1,2 +1,2 @@
-nop %sh{ mkfifo test-fifo; ( printf 'blah' > test-fifo ) </dev/null >/dev/null 2>&1 & }
+nop %sh{ mkfifo test-fifo; ( printf 'blah' > test-fifo & ) </dev/null >/dev/null 2>&1 }
 edit -readonly -fifo test-fifo out
--- a/test/run
+++ b/test/run
@@ -33,7 +33,7 @@ main() {
   for dir in $(find "${@:-.}" -type d | sort); do
     cd $root/$dir;
     mkdir -p $work/$dir
-    for file in in cmd rc; do
+    for file in in cmd rc enabled; do
       [ -f $file ] && cp $file $work/$dir/
     done
     cd $work/$dir;


sidkshatriya:
P.S. Tip for @fredg0 : If you need to run `make test` again, be sure to do `$ kak -clear` to clear out things before you run again.

test/compose/select-display-columns/enabled
```diff
@@ -1,2 +1,2 @@
 #!/bin/sh
-[ $(echo -n "😄😊😉😍" | wc -m) == 4 ] && locale | grep LC_CTYPE | grep -qi 'utf-*8'
+[ $(echo -n "😄😊😉😍" | wc -m) = '4' ] && locale | grep LC_CTYPE | grep -qi 'utf-*8'
```
Screwtapello, line 2: This baffles me. According to everything I know about shell-script parsing, this change should make no difference — `[` is an ordinary command, `4` is an ordinary word, and quoting words doesn't change their meaning unless they include metacharacters like `$` or `\` and `4` isn't one. But apparently in OpenBSD's shell this is a thing?

sidkshatriya: This is a not a OpenBSD thing but a Linux thing or at least a Ubuntu thing. 

On Ubuntu 20.04 which uses the dash shell as the `sh` shell, I get this: 

```
$ ./enabled
./enabled: 2: [: 4: unexpected operator
$
```

krobelus:
> -[ $(echo -n "😄😊😉😍" | wc -m) == 4 ] && locale | grep LC_CTYPE | grep -qi 'utf-*8'
> +[ $(echo -n "😄😊😉😍" | wc -m) = '4' ] && locale | grep LC_CTYPE | grep -qi 'utf-*8'

it's the == vs =

test/compose/select-display-columns/enabled
sidkshatriya: Interestingly, this is OK in OpenBSD!

In OpenBSD 7.0
```sh
$ [ 4 == '4' ]
$ echo $? 
0
```

But using `sh` (which is dash on Ubuntu) you get an error as mentioned in above comment. So the `enabled` script will be fine in OpenBSD. 

Fedora is fine because it uses bash as its `sh` and bash is quite forgiving in general.

sidkshatriya:
**Update:** It turns out the single test that is hanging is `3398-readonly-fifo-failure` (after applying the commits on this PR) is due to a `sh` incompatibility.

If I do:
```
$ export KAKOUNE_POSIX_SHELL=/usr/local/bin/bash
```

**And then when run the tests, I get zero failures.** 

The issue is with the `rc` of the test.
```
nop %sh{ mkfifo test-fifo; ( printf 'blah' > test-fifo ) </dev/null >/dev/null 2>&1 & }
edit -readonly -fifo test-fifo out
```

The OpenBSD's `sh` does not like: 
```sh
mkfifo test-fifo; ( printf 'blah' > test-fifo ) </dev/null >/dev/null 2>&1 &
```

This causes a hang in Kakoune. I've tried reducing the script (e.g. removing /dev/null redirections etc. etc.) to see where the problem lies. Since there is a hang, I suspect it has something to do with the `&` but I don't know what the workaround/fix is for OpenBSD's `sh`.

_Anyways, running the test with bash as the KAKOUNE_POSIX_SHELL (and the commits in this PR) allows all tests to pass._

@mawww when you get the chance please review! I think the commits we have so far can be merged.

krobelus:
> mkfifo test-fifo; ( printf 'blah' > test-fifo ) </dev/null >/dev/null 2>&1 &
> This causes a hang in Kakoune.

Yeah, non-bash shells seem to have this behavior where signals are blocked in backgrounded builtins.
There's a report at https://github.com/mawww/kakoune/issues/3751
I do not know which shell is correct here; I haven't investigated further.

sidkshatriya:
This seems to be a OpenBSD specific `sh` issue because 

(a) The regression test 3398 succeeds on Ubuntu 20.04. Ubuntu uses `dash` as `sh`
(b) The regression test 3398 will succeed on OpenBSD 7 with `dash` if I do `pkg_add dash; export KAKOUNE_POSIX_SHELL=/usr/local/bin/dash`

Screwtapello:
I don't think that's a "non-bash shells" issue, it looks like the "OpenBSD's shell does not like mixing FIFOs, subshells, redirections, and job control" issue which Kakoune has [dealt with before](https://github.com/mawww/kakoune/pull/2955).

sidkshatriya:
@Screwtapello Thank you! That was very helpful. 

I've pushed out a workaround commit to deal with the `&` issue in OpenBSD (consistent with the ones in #2955).

**So now, all tests pass with the default `sh` in OpenBSD**

PR #4478, opened 2021-12-20
modeline.kak: fix multiple kak options

throwawayaccount12345-1:
`modeline-parse` fails with the following file:
```
# kak: set indentwidth=0 tabstop=8:
```

### DIFF
--- a/rc/detection/modeline.kak
+++ b/rc/detection/modeline.kak
@@ -60,8 +60,8 @@ define-command -hidden modeline-parse-impl %{
 
         # Pass a few whitelisted options to kakoune directly
         translate_opt_kakoune() {
-            readonly key="$1"
-            readonly value="$2"
+            key="$1"
+            value="$2"
 
             case "${key}" in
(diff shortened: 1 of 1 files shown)


Screwtapello:
OK, I see what's going on here.

- `:modeline-parse` splits the given modeline into `[("indentwidth", 0), ("tabstop", 8)]`
- It invokes `translate_opt_kakoune indentwidth 0`
- That function creates a global, readonly variable named "key", initialised to "indentwidth"
- Later, `:modeline-parse` invokes `translate_opt_kakoune tabstop 8`
- This time, that function tries to set "key" to "tabstop", but since it's already set to "indentwidth" and since it's marked as readonly, the attempt fails

Removing "readonly" fixes the problem. Making the variables *local* readonly variables would also solve the problem in shells that support `local`, but unfortunately that behaviour [isn't part of POSIX yet](https://www.austingroupbugs.net/view.php?id=767).

PR #4486, opened 2022-01-01
modeline-parse: vim `shiftwidth=0`, `expandtab` and `noexpandtab`

throwawayaccount12345-1:
- Accept `expandtab` and `noexpandtab` in `translate_opt_vim`.
- Recognize `shiftwidth=0`

Vim's behaviour of `expandtab` is to convert to spaces when indenting
(`<<` and `>>`) _and_ on insert mode. This patch does _not_ convert tabs
to spaces in insert mode, it only does the indenting part (`<lt>` and
`<gt>`) by setting `indentwidth` to a proper value.

When `noexpandtab` is given, it will set `indentwidth` to zero.

When `expandtab` is given and shiftwidth is zero, it will set indentwidth
to `%opt{tabstop}`.

The following files are recognized properly:

```
// vim: set noet ts=4 sw=4:
// tabstop=4, indent with tabs
```
```
// vim: set noexpandtab ts=8 sw=4 expandtab:
// tabstop=8, indent with 4 spaces
```
```
// vim: set sw=0 ts=16 et:
// tabstop=16, indent with 16 spaces
```

### DIFF
--- a/rc/detection/modeline.kak
+++ b/rc/detection/modeline.kak
@@ -14,6 +14,7 @@ define-command -hidden modeline-parse-impl %{
     evaluate-commands %sh{
         kakquote() { printf "%s" "$*" | sed "s/'/''/g; 1s/^/'/; \$s/\$/'/"; }
 
+        last_vim_et="" last_vim_sw="" last_vim_ts=""
         # Translate a vim option into the corresponding kakoune one
         translate_opt_vim() {
             key="$1"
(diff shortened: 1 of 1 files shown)


rc/detection/modeline.kak
```diff
@@ -86,13 +106,18 @@ define-command -hidden modeline-parse-impl %{
...
                 -e 's/[ \t]*set\{0,1\}[ \t]//'      \
                 -e 's/:[^a-zA-Z0-9_=-]*$//'         \
                 -e 's/:/ /g'                        \
                 | tr ' ' '\n'                       \
-                | while read -r option; do
+                | while read -r option || stop_looping=1; do
+            if [ "${stop_looping}" = 1 ]; then
+                after_loop_ends
+                break
+            fi
```
mawww, line 120: This confuses me (might be due to still being tired from NYE), why cannot the after_loop_ends logic be just written after the loop end (i.e. after the done on line 108 ?)

throwawayaccount12345-1: The variables `last_vim_et`, `last_vim_sw` and `last_vim_ts` are modified during the loop, which is a subshell because of the pipe. If that code was written after the `done`, they would retain their initial values before the subshell of the `while` loop started.

PR #4487, opened 2022-01-04
haiku: use BPathFinder to find runtime and settings dirs 

t6:
Haiku packages put data files under /boot/system/data and
settings should be under ~/config/settings. On Haiku we can use
BPathFinder to look them up instead of hardcoding the paths.

While here also mention the Haiku package in the README.


mawww:
Hello,

sorry for the delay, I am not quite decided on this yet, my main issue is that this deviates quite a bit from what modern posix systems do, and I am a bit relunctant to add platform specific code.

I am not a Haiku user, but looking at this https://marc.info/?l=haiku-commits&m=140664638010142&w=4 suggests it might already support XDG configs dirs fine, so the value of the change is unclear.

Can you comment on how Kakoune currently behave on Haiku ?

PR #4489, opened 2022-01-04
markdown: allow additional highlighters in code blocks

jeroendehaas:
This PR adds the ability to add additional highlighters for fenced code blocks.

It does so by introducing the option `markdown_supported_languages`. This is a map from identifiers (keys) to highlighters (values). If a value is omitted, the key will be used as the name of a highlighter.

What this does, is enable plugins to register their language by adding a key-value pair to the variable. Any pairs added before the markdown module is required, will be highlighted. 

This is how I intend to use this mechanism in idris2.kak:

```kak
hook global KakBegin .* %{
    set -add global markdown_supported_languages idris=idris
}
```

### DIFF
--- a/rc/filetype/markdown.kak
+++ b/rc/filetype/markdown.kak
@@ -139,7 +139,16 @@ define-command -hidden markdown-indent-on-new-line %{
 define-command -hidden markdown-load-languages %{
     evaluate-commands -draft %{ try %{
         execute-keys 'gtGbGls```\h*\{?[.=]?\K[^}\s]+<ret>'
-        evaluate-commands -itersel %{ require-module %val{selection} }
+        evaluate-commands -itersel %sh{
+          for kv in ${kak_opt_markdown_supported_languages}; do
+            ref=${kv##*=}
+            lang=${kv%%=*}
+            [ -n "$ref" ] || ref=$lang
+            if [ "${kak_selection}" = "${lang}" ]; then
+                printf 'require-module %s\n' "${ref}"
+            fi
+          done
+        }
     }}
 }
 
--- a/rc/filetype/markdown.kak
+++ b/rc/filetype/markdown.kak
@@ -67,7 +67,7 @@ define-command -hidden markdown-build-fenced-highlighters %{
 }
 
 markdown-build-fenced-highlighters
-hook -group markdown-update-highlighers global GlobalSetOption markdown_supported_languages=.* %{
+hook -group markdown-update-highlighters global GlobalSetOption markdown_supported_languages=.* %{
     markdown-build-fenced-highlighters
 }
 
--- a/rc/filetype/markdown.kak
+++ b/rc/filetype/markdown.kak
@@ -31,6 +31,15 @@ hook -group markdown-highlight global WinSetOption filetype=markdown %{
     hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/markdown }
 }
 
+declare-option \
+    -docstring "map of supported languages in fenced code blocks. Keys determine which languages are recognised. The value names the corresponding highligher, if empty the key is used instead." \
+    str-to-str-map markdown_supported_languages \
+    awk= c= cabal= clojure= coffee= cpp= crystal= css= cucumber= d= diff= dockerfile= elixir= erlang= fish= \
+    gas= go= haml= haskell= html= ini= java= javascript= json= julia= kak=kakrc kickstart= \
+    latex= lisp= lua= makefile= markdown= moon= objc= ocaml= perl= pug= python= ragel= \
+    ruby= rust= sass= scala= scss= sh= swift= toml= tupfile= typescript= yaml= sql=
+
+
 
 provide-module markdown %{
 
@@ -42,16 +51,12 @@ add-highlighter shared/markdown/inline default-region regions
 add-highlighter shared/markdown/inline/text default-region group
 
 evaluate-commands %sh{
-  languages="
-    awk c cabal clojure coffee cpp crystal css cucumber d diff dockerfile elixir erlang fish
-    gas go haml haskell html ini java javascript json julia kak kickstart
-    latex lisp lua makefile markdown moon objc ocaml perl pug python ragel
-    ruby rust sass scala scss sh swift toml tupfile typescript yaml sql
-  "
-  for lang in ${languages}; do
+  for kv in ${kak_opt_markdown_supported_languages}; do
+    ref=${kv##*=}
+    lang=${kv%%=*}
+    [ -n "$ref" ] || ref=$lang
     printf 'add-highlighter shared/markdown/%s region -match-capture ^(\h*)```\h*(%s\\b|\\{[.=]?%s\\})   ^(\h*)``` regions\n' "${lang}" "${lang}" "${lang}"
     printf 'add-highlighter shared/markdown/%s/ default-region fill meta\n' "${lang}"
-    [ "${lang}" = kak ] && ref=kakrc || ref="${lang}"
     printf 'add-highlighter shared/markdown/%s/inner region \A```[^\\n]*\K (?=```) ref %s\n' "${lang}" "${ref}"
   done
 }
--- a/rc/filetype/markdown.kak
+++ b/rc/filetype/markdown.kak
@@ -35,7 +35,7 @@ declare-option \
     -docstring "map of supported languages in fenced code blocks. Keys determine which languages are recognised. The value names the corresponding highligher, if empty the key is used instead." \
     str-to-str-map markdown_supported_languages \
     awk= c= cabal= clojure= coffee= cpp= crystal= css= cucumber= d= diff= dockerfile= elixir= erlang= fish= \
-    gas= go= haml= haskell= html= ini= java= javascript= json= julia= kak=kakrc kickstart= \
+    gas= go= haml= haskell= hs=haskell html= ini= java= javascript= js=javascript json= julia= kak=kakrc kickstart= \
     latex= lisp= lua= makefile= markdown= moon= objc= ocaml= perl= pug= python= ragel= \
     ruby= rust= sass= scala= scss= sh= swift= toml= tupfile= typescript= yaml= sql=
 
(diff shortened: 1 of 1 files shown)


sidkshatriya:
I haven't tested this but this is a useful addition, no doubt.

> This PR adds the ability to add additional highlighters for fenced code blocks.

For those who are not immediately familiar, author means things like ` ```haskell ` and so forth.

> This is how I intend to use this mechanism in idris2.kak

Do you intend to keep `idris2.kak` outside the kakoune tree or contribute it to the kakoune source? I'm guessing that you intend to keep it outside, hence this feature?

jeroendehaas:
> I haven't tested this but this is a useful addition, no doubt.
> 
> > This PR adds the ability to add additional highlighters for fenced code blocks.
> 
> For those who are not immediately familiar, author means things like `` ```haskell `` and so forth.
> 
> > This is how I intend to use this mechanism in idris2.kak
> 
> Do you intend to keep `idris2.kak` outside the kakoune tree or contribute it to the kakoune source? I'm guessing that you intend to keep it outside, hence this feature?

I am open to the idea of contributing support for Idris 2 to Kakoune! Nevertheless, I believe this feature is useful not only for plugin developers but also for regular users. For instance, to have proper syntax highlighting for Haskell code blocks, users are required to spell out "haskell". GitHub and Pandoc also accept the shorter `hs`.

CleanShot 2022-01-05 at 18 56 09@2x

With this change, a user could add the following to her kakrc and get support for both:

```kak
hook global KakBegin .* %{
    set -add global markdown_supported_languages hs=haskell
}
provide-module hs %{ require-module haskell }
```

rc/filetype/markdown.kak
```diff
@@ -30,6 +30,15 @@ hook -group markdown-highlight global WinSetOption filetype=markdown %{
     hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/markdown }
 }
 
+declare-option \
```
mawww, line 33: This could still be in the shell block (using the map syntax). I am not really keen on that option because its only when the module is loaded that it gets used, so changing it after will not work and could be confusing.

mawww (changes requested):

rc/filetype/markdown.kak
jeroendehaas: I agree that this is more a pragmatic rather than an elegant solution. I was looking for a way to allow third-party highlighters in code blocks. 

However, I will adjust the code to at least allow for the second use case, i.e., to allow synonyms for languages.

alexherbo2: You can add a group in `markdown`.

When option changes, build the group.

jeroendehaas: Now, in the summer holidays I have taken another shot at this., see the new forced push

jeroendehaas:
I'd like to request your critical attention to my second attempt at this. The markdown module now installs a hook to update the highlighter when `markdown_supported_languages` is altered. I intend this to be a global only variable, but I don't know if that (a) fits with kakoune's module and (b) if I have implemented it sensibly (in particular my use of the `GlobalSetOption` hook)

rc/filetype/markdown.kak
```diff
@@ -31,6 +31,15 @@ hook -group markdown-highlight global WinSetOption filetype=markdown %{
...
+    -docstring "map of supported languages in fenced code blocks. Keys determine which languages are recognised. The value names the corresponding highligher, if empty the key is used instead." \
+    str-to-str-map markdown_supported_languages \
+    awk= c= cabal= clojure= coffee= cpp= crystal= css= cucumber= d= diff= dockerfile= elixir= erlang= fish= \
+    gas= go= haml= haskell= hs=haskell html= ini= java= javascript= js=javascript json= julia= kak=kakrc kickstart= \
+    latex= lisp= lua= makefile= markdown= moon= objc= ocaml= perl= pug= python= ragel= \
+    ruby= rust= sass= scala= scss= sh= swift= toml= tupfile= typescript= yaml= sql=
```
krobelus, line 40: Here's an idea - since you mainly want to use this for

    set -add global markdown_supported_languages idris=idris

how about we do this automagically by assuming that a language identifier is the name of a filetype by default.


so when a markdown file with this contents is opened

	```foo
	```

we run `require-module foo` and give it a filetype=foo highlighter.
No user configuration needed.
This also means that we can remove the long list of filetypes from markdown.kak, which is currently a pain to keep in sync with supported languages anyway. 

Some more comments before I had that idea below

rc/filetype/markdown.kak
```diff
@@ -31,6 +31,15 @@ hook -group markdown-highlight global WinSetOption filetype=markdown %{
     hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/markdown }
 }
 
+declare-option \
+    -docstring "map of supported languages in fenced code blocks. Keys determine which languages are recognised. The value names the corresponding highligher, if empty the key is used instead." \
+    str-to-str-map markdown_supported_languages \
+    awk= c= cabal= clojure= coffee= cpp= crystal= css= cucumber= d= diff= dockerfile= elixir= erlang= fish= \
+    gas= go= haml= haskell= hs=haskell html= ini= java= javascript= js=javascript json= julia= kak=kakrc kickstart= \
```
krobelus, line 38: I initially wondered why the `str-to-str-map` instead of a `str-list` (and just keep our hack for `kakrc`).
The second commit reveals why: this allows to add filetype aliases like hs=haskell.
Prior discussion on that is in https://github.com/mawww/kakoune/pull/4592
where I thought it's ugly to add a new highlighter for each separate alias (there can be many)
but with your solution, users are responsible for adding the aliases
they need, so this seems okay.

rc/filetype/markdown.kak
```diff
@@ -41,19 +50,25 @@ add-highlighter shared/markdown regions
...
-  for lang in ${languages}; do
-    printf 'add-highlighter shared/markdown/%s region -match-capture ^(\h*)```\h*(%s\\b|\\{[.=]?%s\\})   ^(\h*)``` regions\n' "${lang}" "${lang}" "${lang}"
-    printf 'add-highlighter shared/markdown/%s/ default-region fill meta\n' "${lang}"
-    [ "${lang}" = kak ] && ref=kakrc || ref="${lang}"
-    printf 'add-highlighter shared/markdown/%s/inner region \A```[^\\n]*\K (?=```) ref %s\n' "${lang}" "${ref}"
-  done
+# Used for the pluggable infrastructure of languages
+define-command -hidden markdown-build-fenced-highlighters %{
+    add-highlighter -override shared/markdown/fenced region -match-capture ^(\h*)``` ^(\h*)``` regions
+    add-highlighter shared/markdown/fenced/default default-region fill meta
```
krobelus, line 56: It seems backwards to first parse the value and then the key, how about changing the order?

rc/filetype/markdown.kak
```diff
@@ -41,19 +50,25 @@ add-highlighter shared/markdown regions
...
+        [ -n "$ref" ] || ref=$lang
+        printf 'add-highlighter shared/markdown/fenced/%s region -match-capture ^(\h*)```\h*(%s\\b|\\{[.=]?%s\\})   ^(\h*)``` regions\n' "${lang}" "${lang}" "${lang}"
+        printf 'add-highlighter shared/markdown/fenced/%s/ default-region fill meta\n' "${lang}"
+        printf 'add-highlighter shared/markdown/fenced/%s/inner region \A```[^\\n]*\K (?=```) ref %s\n' "${lang}" "${ref}"
+      done
+    }
+}
+
+markdown-build-fenced-highlighters
+hook -group markdown-update-highlighers global GlobalSetOption markdown_supported_languages=.* %{
```
krobelus, line 70: typo:

```suggestion
hook -group markdown-update-highlighters global GlobalSetOption markdown_supported_languages=.* %{
```

rc/filetype/markdown.kak
```diff
@@ -124,8 +139,19 @@ define-command -hidden markdown-indent-on-new-line %{
...
     evaluate-commands -draft %{ try %{
         execute-keys 'gtGbGls```\h*\{?[.=]?\K[^}\s]+<ret>'
-        evaluate-commands -itersel %{ require-module %val{selection} }
+        evaluate-commands -itersel %sh{
+          for kv in ${kak_opt_markdown_supported_languages}; do
+            ref=${kv##*=}
+            lang=${kv%%=*}
+            [ -n "$ref" ] || ref=$lang
+            if [ "${kak_selection}" = "${lang}" ]; then
+                printf 'require-module %s\n' "${ref}"
```
krobelus, line 148: A shame we can't do an actual map lookup.

rc/filetype/markdown.kak
krobelus, line 40: Here's an idea - since you mainly want to use this for

    set -add global markdown_supported_languages idris=idris

how about we do this automagically by assuming that a language identifier is the name of a filetype by default.


so when a markdown file with this contents is opened

	```foo
	```

we run `require-module foo` and give it a filetype=foo highlighter.
No user configuration needed.
This also means that we can remove the long list of filetypes from markdown.kak, which is currently a pain to keep in sync with supported languages anyway. 

Some more comments before I had that idea below

rc/filetype/markdown.kak
krobelus, line 38: I initially wondered why the `str-to-str-map` instead of a `str-list` (and just keep our hack for `kakrc`).
The second commit reveals why: this allows to add filetype aliases like hs=haskell.
Prior discussion on that is in https://github.com/mawww/kakoune/pull/4592
where I thought it's ugly to add a new highlighter for each separate alias (there can be many)
but with your solution, users are responsible for adding the aliases
they need, so this seems okay.

rc/filetype/markdown.kak
krobelus, line 56: It seems backwards to first parse the value and then the key, how about changing the order?

rc/filetype/markdown.kak
krobelus, line 70: typo:

```suggestion
hook -group markdown-update-highlighters global GlobalSetOption markdown_supported_languages=.* %{
```

rc/filetype/markdown.kak
krobelus, line 148: A shame we can't do an actual map lookup.

rc/filetype/markdown.kak
jeroendehaas: The decision to choose a map was motivated by the use of the `kakrc` module for blocks marked `kak`. But indeed, this allows for a much more general approach to managing recognised languages.

rc/filetype/markdown.kak
```diff
@@ -41,19 +50,25 @@ add-highlighter shared/markdown regions
...
+# Used for the pluggable infrastructure of languages
+define-command -hidden markdown-build-fenced-highlighters %{
+    add-highlighter -override shared/markdown/fenced region -match-capture ^(\h*)``` ^(\h*)``` regions
+    add-highlighter shared/markdown/fenced/default default-region fill meta
+    evaluate-commands %sh{
+      for kv in ${kak_opt_markdown_supported_languages}; do
+        ref=${kv##*=}
+        lang=${kv%%=*}
+        [ -n "$ref" ] || ref=$lang
+        printf 'add-highlighter shared/markdown/fenced/%s region -match-capture ^(\h*)```\h*(%s\\b|\\{[.=]?%s\\})   ^(\h*)``` regions\n' "${lang}" "${lang}" "${lang}"
```
krobelus, line 62: I think we need `add-highlighter -override` or else it will error if we modify the option.

rc/filetype/markdown.kak
krobelus: restructuredtext.kak has a similar hardcoded list of filetypes, we should use the same solution there (also we should add an equivalent to `markdown-load-languages`)

there are security considerations when interpreting "```foo" as `add-highlighter shared/markdown/foo [..]` but as long as we escape foo properly, all a malicious markdown file can do is to add a bunch of highlighters and make calls to `require-module`, which *should* be fine..

krobelus: https://github.com/mawww/kakoune/pull/4715 implements the behavior of loading arbitrary modules+highlighters, which also allows users to add their own aliases fairly easily

jeroendehaas:
I am guessing, we should prefer #4715 to this PR. My solution does not handle the scope of the custom setting particularly well.

PR #4493, opened 2022-01-09
Do not show custom completions when autocomplete is off

krobelus:
As reported in [1], completions provided by "set global completers
option=my_completion" activate insert mode autocompletion, even when
the autocomplete option does not have the insert mode flag.

This happens because InsertCompleter::on_option_changed() calls
InsertCompleter::setup_ifn(), which shows the completion pager.
Fix this by computing whether the completion pager is enabled;
otherwise we can return early from setup_ifn().
The completion pager is enabled if the autocompletion bit is set,
or if the user has requested explicit completion.

[1]: https://github.com/kak-lsp/kak-lsp/issues/585


sidkshatriya:
This is a very minor point but the directory `test/regression/0-autocomplete-overrules-completers` should be renamed as `test/regression/4493-autocomplete-overrules-completers`. This would make it consistent with the rest of the regression suite.

I think that was the original intention to name it `0-` but it was somehow forgotten.

krobelus:
On Mon, Jan 10, 2022 at 11:53:31PM -0800, Sidharth Kshatriya wrote:
> This is a very minor point but the directory `test/regression/0-autocomplete-overrules-completers` should be renamed as `test/regression/4493-autocomplete-overrules-completers`. This would make it consistent with the rest of the regression suite.
> 
> I think that was the original intention to name it `0-` but it was somehow forgotten.

Thanks for the review; I think this is fine as-is, because there is no meaningful extra info on the github PR. There are a couple others prefixed with 0.
(also since there is no issue, the patch would need to reference its PR, which is a weird chicken and egg problem that righfully suggests that we're doing something wrong)

mawww:
The number in front is usually the github issue that this test reproduces, but here we are not fixing any existing github issue so I am happy to use 0- as a prefix.

stacyharper:
I think it may be related to this. I had `set global autocomplete 'prompt'` and was able to <c-n> to start autocompletion while in insert mode. Now I had to enable insert also to have the same feature but I dont really want the popup by default. 

Is it a regression ? Or need I some other config ?

krobelus:
@stacyharper right that's a regression; fix in https://github.com/mawww/kakoune/pull/4532

PR #4494, opened 2022-01-09
Allow to jump from an email's inline diff to source file

krobelus:
* rc git: skip email quotes in git-diff-goto-source

  When reading and writing emails that contain patches (possibly
  email-quoted), it can be convenient to the jump to the source file.
  Allow this by making git-diff-goto-source (bound to <ret> in git-diff
  buffers) ignore leading email quotes ("> "). A line that starts with
  "> " does not occur in a unified diff, so it should not affect other
  use cases.
    
  Observe that git-diff-goto-source even works around interleaved
  replies; it will just ignore lines that don't start with (> )*[ +-].

* rc/filetype//mail: enable jumping from inline diff to source file
    
  In combination with the previous patch, this allows to jump from
  a mail buffer with inline diff to the corresponding source files
  (most accurate when the patch has been applied locally).


mawww:
I like the feature, but I dont quite like that mail.kak would depend on the git module, I wonder if any of the git logic to jump to the file is actually git specific and if it could move to diff.kak. It would make much more sense then to have mail.kak rely on diff.kak.

krobelus:
Yeah it sounds good to move the logic to diff.kak and load it from git.kak and mail.kak

To port this to non-Git unified diffs (do we care about other formats?), we probably want to parse the --- and +++ lines instead of the ^diff line because the ^diff line varies across implementations, for example `git diff`:

	diff --git a/file b/file
	--- a/file
	+++ a/file

and `hg diff`

	diff -r <revision1> -r <revision2> file
	--- a/file	<timestamp>
	+++ b/file	<timestamp>

GNU `diff -u` doesn't add a ^diff line (yet diff.kak uses that) but it also doesn't add the `a/` prefix.

	--- oldfile	<timestamp>
	+++ newfile	<timestamp>

To sum up:
- Git specific: `git rev-parse --show-toplevel` (can be factored out easily)
- VCS-specific a/ b/ prefixes. We can probably ignore the case if these are not present, then for `diff -u` we'll jump to `newfile`. Adding a symmetric command that jumps to the old file is also trivial.

Screwtapello:
> VCS-specific a/ b/ prefixes.

Technically these aren't specific to VCS tools, they're specific to the "comparing two directories" use-case. Back before VCSs became widely popular, you'd download a tarball, extract it, try to build it, fix it, and then want to contribute your fixes back upstream so you'd do something like:

```console
$ tar xZf libfoo-1.2.tar.Z
$ cd libfoo-1.2
$ # ...build, fix, repeat...
$ cd ..
$ mv libfoo-1.2 libfoo-1.2-patched # the copy with my fixes in it
$ tar xZf libfoo-1.2.tar.Z # extract a clean copy of the original version
$ diff -urw libfoo-1.2 libfoo-1.2-patched > ~/libfoo-fixes.diff
```

So each hunk in the resulting diff would start off like:

```diff
--- libfoo-1.2/foo.c
+++ libfoo-1.2-patched/foo.c
```

Since the first path segment of each filename in a diff was unpredictable and pretty meaningless, it was standard practice to apply patches with `patch -p1`, meaning "strip 1 path segment from every filename before trying to use it". When git came along later and tried to formalise the unified-diff file-format, they deliberately added a bogus prefix to each filename so that people without `git apply` could continue to use `patch -p1` out of habit, and things would Just Work.

To make Kakoune handle unified diffs in general, it'd be nice to do something like "always strip the first path segment unless there's only one path segment".

krobelus:
Right that sounds like an obvious and sensible workflow (I wonder if anyone still uses it).

> To make Kakoune handle unified diffs in general, it'd be nice to do something like "always strip the first path segment unless there's only one path segment".

Yeah that looks like the cleanest solution.
I wonder if we can also accomodate the "compare two files (with multiple path segments)" use case:

	$ diff -u src/main.cc src/.main.cc.kak.123456 
	--- src/main.cc	<timestamp>
	+++ src/.main.cc.kak.123456	<timestamp>

here I expect to not strip the src/ prefix.

I guess we can first try without the prefix (common case), and then try again with the prefix. A bit too magic.
Or parameterize the new diff.kak command, and by default only strip the prefix in a context where we expect a patch (git.kak + mail.kak).

krobelus:
OK I've found a reasonable criteria: `diff` only prints the lines starting with `diff` if given `-r`.
So we can just strip the prefix if there is a `diff` line.

PR #4502, opened 2022-01-13
Add support for PHP heredoc in syntax highlighting

sidkshatriya:
This functionality allows PHP heredocs (the `<<<SOMETHING` ... `SOMETHING;` pattern) to be properly syntax highlighted in Kakoune. Example:

```php
    if (!function_exists('proc_open')) {
        echo <<<NO_PROC_OPEN_ERROR

+-----------------------------------------------------------+
|                       ! ERROR !                           |
| The test-suite requires that proc_open() is available.    |
| Please check if you disabled it in php.ini.               |
+-----------------------------------------------------------+

NO_PROC_OPEN_ERROR;
        exit(1);
    }
```

PR #4503, opened 2022-01-13
Add support for PHP heredoc in syntax highlighting

sidkshatriya:
This functionality allows PHP heredocs (the `<<<SOMETHING` ... `SOMETHING;` pattern) to be properly syntax highlighted in Kakoune. Example:

```php
    if (!function_exists('proc_open')) {
        echo <<<NO_PROC_OPEN_ERROR

+-----------------------------------------------------------+
|                       ! ERROR !                           |
| The test-suite requires that proc_open() is available.    |
| Please check if you disabled it in php.ini.               |
+-----------------------------------------------------------+

NO_PROC_OPEN_ERROR;
        exit(1);
    }
```

P.S. Sorry for the open/close in #4502 -- I am happy with this (resubmitted) PR now. I've tested it more thoroughly and it should be ready to merge!

### DIFF
--- a/rc/filetype/php.kak
+++ b/rc/filetype/php.kak
@@ -39,7 +39,12 @@ add-highlighter shared/php/doc_comment2  region /\*\*  \*/             ref php/d
 add-highlighter shared/php/comment1      region //     '$'             fill comment
 add-highlighter shared/php/comment2      region /\*    \*/             fill comment
 add-highlighter shared/php/comment3      region '#'    '$'             fill comment
-
+# Example PHP heredoc:
+# echo <<<SOMETHING
+#  some text
+#  more text etc.
+# SOMETHING;
+add-highlighter shared/php/heredoc       region -match-capture '<<<(.*?)$' '^\h*(.*?);' fill string
 
 
 add-highlighter shared/php/code/ regex &?\$\w* 0:variable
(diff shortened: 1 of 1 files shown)


sidkshatriya:
And while I'm here, added support for syntax highlighting in PHP Test files (`.phpt`)

For some examples of these kind of files see here https://github.com/php/php-src/blob/master/tests/basic

rc/filetype/php.kak
```diff
@@ -39,7 +39,12 @@ add-highlighter shared/php/doc_comment2  region /\*\*  \*/             ref php/d
 add-highlighter shared/php/comment1      region //     '$'             fill comment
 add-highlighter shared/php/comment2      region /\*    \*/             fill comment
 add-highlighter shared/php/comment3      region '#'    '$'             fill comment
-
+# Example PHP heredoc:
```
mawww, line 42: I do not think this comment is very valuable, this is not really the place to document php's syntax, and the highlighter name (shared/php/heredoc) should be enough to understand what this is supposed to highlight.

sidkshatriya: Done -- removed the comment

sidkshatriya: @mawww ^

PR #4509, opened 2022-01-17
Don't escape markup strings in modeline expansions

### LINKED ISSUE #4507: [BUG] Markup returned from expansions is not parsed in the modeline
### Version of Kakoune

v2021.11.08

### Reproducer

Execute the following commands:

```kak
declare-option str modeline_test "{green}testing"
set-option global modelinefmt '%opt{modeline_test}'
```

### Outcome

The modeline is printed with markup strings included, and the markup strings are not parsed.

image

### Expectations

The markup strings would be parsed. This would require a second pass on the modeline, but would be worth it in order to facilitate easier customization. As it is now, you have to hardcode the markup strings and duplicate logic in multiple places if you want things to be dynamically added/removed to/from the modeline.

### Additional information

_No response_

raiguard:
Fixes #4507

The documentation for `modelinefmt` states that expansions are processed, then markup strings are applied. This brings the actual behavior in line with the documentation.

**Before**

image

**After**

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,6 @@
 *.pyc
 *.1-r
 .*.kak.*
-.ccls-cache
-compile_commands.json
 src/kak
 src/kak.debug
 src/kak.opt
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@
 *.pyc
 *.1-r
 .*.kak.*
+.ccls-cache
+compile_commands.json
 src/kak
 src/kak.debug
 src/kak.opt
--- a/src/client.cc
+++ b/src/client.cc
@@ -157,8 +157,7 @@ DisplayLine Client::generate_mode_line() const
         HashMap<String, DisplayLine> atoms{{ "mode_info", context().client().input_handler().mode_line() },
                                            { "context_info", {generate_context_info(context()),
                                                               context().faces()["Information"]}}};
-        auto expanded = expand(modelinefmt, context(), ShellContext{},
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -3,6 +3,8 @@
 *.pyc
 *.1-r
 .*.kak.*
+.ccls-cache
+compile_commands.json
```
caksoylar, line 7: These two seem to be added by mistake?

raiguard: Actually, it was intentional, but I will remove those from the main .gitignore and put them `.git`.

caksoylar: Right, I meant adding them to this PR.

mawww:
I think the reason we escape potential '{' is that there is not guarantee that what you expand in modelinefmt is markup aware, and there is no easy was to escape it yourself. The buffer name for example could contain a '{' and with this change applied it will be treated as a face description.

raiguard:
For what it's worth, I have been using this in a custom branch for nine months, and have had no issues with it. Sure, if you put braces in a filename, it'll be treated as markup, but braces in a filename is extraordinarily rare. At least, I've never run into it in my workflow.

Without this change, getting my modeline to look how it does requires significantly more shell logic, which executes on every keystroke and murders performance.

I'm not sure what a proper solution for this issue would look like. I made my filename `{Error}foo` just for fun:
image

raiguard:
I have drastically simplified my modeline, so I no longer need this PR. Since it has unresolved issues, I'll go ahead and close it.

PR #4516, opened 2022-01-26
Git blame wrapper: Pass blame flags all at once

Qeole:
The wrapper for `git blame` creates flags for each line of the buffer. It parses the output from git and would send a flag (or a series of flags) each time the commit to blame for a line differs from the previous one. For files that were touched by a large number of commits, this results in a high number of connections to the kakoune session, and may take some time. This is visible in the session through the flags for the different commits appearing on the lines one by one, possibly during several seconds.

Instead, it is faster to build a single series of flags while parsing the output from git, and to send it all at once to kakoune.

[EDIT: Latest version sends one batch of flags every 1s or so, instead of passing them all at once.]

### DIFF
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -109,21 +109,27 @@ define-command -params 1.. \
                       set-option buffer=$kak_bufname git_blame_flags '$kak_timestamp'
                   }" | kak -p ${kak_session}
                   git blame "$@" --incremental ${kak_buffile} | awk '
-                  function send_flags(text, flag, i) {
+                  function send_flags(flush,    text, i) {
                       if (line == "") { return; }
                       text=substr(sha,1,8) " " dates[sha] " " authors[sha]
                       # gsub("|", "\\|", text)
                       gsub("~", "~~", text)
-                      flag="%~" line "|" text "~"
-                      for ( i=1; i < count; i++ ) {
-                          flag=flag " %~" line+i "|" text "~"
+                      for ( i=0; i < count; i++ ) {
+                          flags = flags " %~" line+i "|" text "~"
+                      }
+                      now = systime()
+                      # Send roughly one update per second, to avoid creating too many kak processes.
+                      if (!flush && now - last_sent < 1) {
+                          return
                       }
                       cmd = "kak -p " ENVIRON["kak_session"]
-                      print "set-option -add buffer=" ENVIRON["kak_bufname"] " git_blame_flags " flag | cmd
+                      print "set-option -add buffer=" ENVIRON["kak_bufname"] " git_blame_flags " flags | cmd
                       close(cmd)
+                      flags = ""
+                      last_sent = now
                   }
                   /^([0-9a-f]+) ([0-9]+) ([0-9]+) ([0-9]+)/ {
-                      send_flags()
+                      send_flags(0)
                       sha=$1
                       line=$3
                       count=$4
@@ -134,7 +140,7 @@ define-command -params 1.. \
                        cmd | getline dates[sha]
                        close(cmd)
                   }
-                  END { send_flags(); }'
+                  END { send_flags(1); }'
         ) > /dev/null 2>&1 < /dev/null &
     }
 


rc/tools/git.kak
```diff
@@ -134,7 +131,12 @@ define-command -params 1.. \
                        cmd | getline dates[sha]
                        close(cmd)
                   }
-                  END { send_flags(); }'
+                  END {
+                      update_flags()
+                      cmd = "kak -p " ENVIRON["kak_session"]
+                      print "set-option -add buffer=" ENVIRON["kak_bufname"] " git_blame_flags " flags | cmd
+                      close(cmd)
```
krobelus, line 138: This is probably fine for now, but do note that makes us lose the incremental updates (some line flags are already shown before git-blame exits) which can be noticeable on large files in a large repo.

Since we are a long-running process (`> /dev/null 2>&1 < /dev/null &`) our only way to communicate with Kakoune is `kak -p`.
(although we could also write to Kakoune's command socket directly, like [kakoune-smooth-scroll]( https://github.com/caksoylar/kakoune-smooth-scroll/blob/master/smooth-scroll.py#L15)).

I think between running `kak -p` for every new blame and it onlly once, a reasonable middle ground would be to run it once per second.
We just need an atomic buffer to keep our `set-option -add` commands; I'm not sure how to do that with shell/awk. I guess we could create a bunch of temporary files but that might defeat the point of this patch.

krobelus:
> This is visible in the session through the flags for the different commits appearing on the lines one by one, possibly during several seconds.

Well this is intentional, because `git-blame` can be slow! Are you sure this is because of `kak -p`? That should be really fast, like 5-10ms. Does your patch decrease the runtime of the `awk` process significantly?

Qeole:
> Well this is intentional, because `git-blame` can be slow! Are you sure this is because of `kak -p`? That should be really fast, like 5-10ms. Does your patch decrease the runtime of the `awk` process significantly?

I haven't measured the duration of `kak -p` itself, but I'm not sure what else can cause a variation in the runtime duration for `awk`. So I've measured the time for `awk` over a few runs, by logging the date right before and after. I used a file within the Linux kernel repository (I tried `include/uapi/linux/bpf.h`). I get:

- Without the patch: 4.5s
- With the patch applied: 1.0s

As you mentioned, we loose the incremental updates, so with the patch, the buffer remains 1s without any flag appearing when running :git blame`. But the complete blame info lands faster.

But then the PR is just a suggestion. If the regular updates _are_ intentional, maybe it's not a good idea to change.

Qeole:
Yes, this works for me. The duration for `awk` is the same, about 1.0s average on my tests.

For what it's worth, I don't think we have to create several lines with the `set-option ...` before sending, it looks like a single one with all the flags retrieved works just as well, i.e.:

```
  sendbuffer = sendbuffer "set-option -add buffer=" ENVIRON["kak_bufname"] " git_blame_flags " flag "\n"
  ...
  print sendbuffer | cmd
```
can be:
```
  sendbuffer = sendbuffer " " flag
  ...
  print "set-option -add buffer=" ENVIRON["kak_bufname"] " git_blame_flags " sendbuffer "\n" | cmd
```

But it doesn't change much anyway.

Do you want to open a separate PR with your code? Or shall I update the current one?

krobelus:
> sendbuffer = sendbuffer " " flag
> ...
> print "set-option -add buffer=" ENVIRON["kak_bufname"] " git_blame_flags " sendbuffer "\n" | cmd

Yep that looks better. Maybe also rename `sendbuffer` -> `flags`.

> Do you want to open a separate PR with your code? Or shall I update the current one?

You go ahead if you have time. (The coding part is easy, the hard part is to write a good summary for the commit message ;)

Qeole:
Updated now, @krobelus please take another look when you can.

rc/tools/git.kak
```diff
@@ -109,21 +109,27 @@ define-command -params 1.. \
                       set-option buffer=$kak_bufname git_blame_flags '$kak_timestamp'
                   }" | kak -p ${kak_session}
                   git blame "$@" --incremental ${kak_buffile} | awk '
-                  function send_flags(text, flag, i) {
+                  function send_flags(flush,    text, i) {
```
krobelus, line 112: there is extra whitespace here

```suggestion
                  function send_flags(flush, text, i) {
```

LGTM otherwise, nice work!

BTW I haven't used the `git blame` wrapper much since I mostly use a binding that [launches `tig blame`](https://git.sr.ht/~krobelus/dotfiles/tree/f0fbf9c77bd413973b82899c73820034430e502b/item/.config/kak/kakrc#L433)

krobelus (approved):

rc/tools/git.kak
qmonnet: > there is extra whitespace here

Yes, this is on purpose. From awk's man page:
```
Since functions were not originally part of the AWK language, the provision
for local variables is rather clumsy: They are declared as extra parameters
in the parameter list.  The convention is to separate local variables from
real parameters by extra spaces in the parameter list.  For example:

      function  f(p, q,     a, b)   # a and b are local
      {
           ...
      }

      /abc/     { ... ; f(1, 2) ; ... }

```

(Although I'm fine with removing the extra spaces if you prefer, it's not like I'd expect a lot of people to be familiar with that convention anyway.)

> BTW I haven't used the `git blame` wrapper much since I mostly use a binding that [launches `tig blame`](https://git.sr.ht/~krobelus/dotfiles/tree/f0fbf9c77bd413973b82899c73820034430e502b/item/.config/kak/kakrc#L433)

Never tried tig before, I've mostly worked with the fugitive plugin for Vim so far. I'll check it out, thank you for the pointer!

Qeole: > there is extra whitespace here

Yes, this is on purpose. From awk's man page:
```
Since functions were not originally part of the AWK language, the provision
for local variables is rather clumsy: They are declared as extra parameters
in the parameter list.  The convention is to separate local variables from
real parameters by extra spaces in the parameter list.  For example:

      function  f(p, q,     a, b)   # a and b are local
      {
           ...
      }

      /abc/     { ... ; f(1, 2) ; ... }

```

(Although I'm fine with removing the extra spaces if you prefer, it's not like I'd expect a lot of people to be familiar with that convention anyway.)

> BTW I haven't used the `git blame` wrapper much since I mostly use a binding that [launches `tig blame`](https://git.sr.ht/~krobelus/dotfiles/tree/f0fbf9c77bd413973b82899c73820034430e502b/item/.config/kak/kakrc#L433)

Never tried tig before, I've mostly worked with the fugitive plugin for Vim so far. I'll check it out, thank you for the pointer!

krobelus:
> Yes, this is on purpose. From awk's man page:

I see, this is in `man gawk` on my system, but it works the same way on POSIX awk. Seems fine.

> I've mostly worked with the fugitive plugin for Vim so far.

Yeah, fugitive (and Magit) are great. There's nothing as comprehensive for Kakoune (yet) but it's easy to use an external tool.

Issue #3801, opened 2020-10-13
Allow closing client with :q instead of :q!


GavinSmith0123:
If I start a session with `kak -s foo -d`, and then a client with `kak -c foo tmp`, then make changes to the `tmp` buffer, and then try to quit with `:q`, I get an error: ` 'q' 1 modified buffers remaining`

However, if I quit anyway with `:q!`, and then go into the `tmp` buffer in another Kakoune client, then the changes are still there, even though I hadn't saved them. Hence, it would be harmless if you could close a client with `:q` even if there were unsaved changes.

This would be useful when editing with lots of clients at once and you decide you don't need some of them.

mawww:
We already have this behaviour in general, all but the last clients can quit with modfiied buffers. It seems here what happens is that you quit the last client, but because your server is in daemon mode you'd like to be able to do that regardless of modified buffers because you can still connect another client to it later on.

I think this would be reasonable to do.

GavinSmith0123:
Another way around the issue is to not have any unsaved buffers.

In Vim, there is the `set nohidden` setting which blocks you from changing the buffer while you have unsaved changes. This helps you to remember to save buffers (otherwise, "out of sight, out of mind").

Is it possible to have the same feature in Kakoune? I couldn't use the `WindowDisplay` hook to do this; is there some other hook that could be used?

PR #4520, opened 2022-01-29
Add commands `quit-all` (alias to `kill`), `write-all-quit-all`

Qeole:
Add two commands `quit-all` (in fact an alias to `kill`) and `write-all-quit-all` to write all buffers then `kill` Kakoune.

I wish there was a simple way to add `qa` as an alias for `quit-all`, but commands do not support more than one alias, so I saw the following options:

- Alias `quit-all` to `kill` and don't have a `qa` alias (what I did)
- Register a separate (but identical) command `quit-all` and add a `qa` alias to it; but it looked awful to have duplicated commands just for that
- Change `struct CommandDesc` to handle list of aliases instead of a single `char *`; I could work on that, but this certainly needs discussion/approval first, and might be overkill for the current case?
- We could also rename `kill` as `quit-all` and make `kill` an alias, to have `quit-all` the main command name, more in line with `quit`, `write-all`, _et cetera_, although it still won't provide a way to alias `qa`.

As a side note, I would also understand if none of those commands are desirable at all. I do think they are a decent addition, but mostly I wanted to pick some low-hanging fruit amongst the open issues and help fix something in the code :).


alexherbo2:
I like the idea of `quit-all`.

I’m not sure, but maybe `quit-all` is not the same as `kill` in case of running Kakoune in server mode.

To my understanding, `quit` closes the client, and when no client remains, terminates the server unless in daemon mode.

So I see `quit-all` as closing all clients, but not necessarily killing the server in daemon mode.

To my observation, `kill` is related to the server and yells with unsaved buffers, while `quit` is more related to closing clients: it does not necessarily yell with unsaved buffers.  If you are in headless mode or have 2 clients for example, `quit` will allow you to close the client, only yelling if the action will kill the server.

So maybe `quit` (client) could inherit the `kill` (server) action.

Related to #3801

occivink:
To avoid all these `-all` variants, maybe we could have a special `*` value to the `eval -client` switch, much like with `-buffer`. 

So `quit-all` would become `eval -client * 'quit'` 
and `write-all-quit-all` would become `write-all ; eval -client * 'quit'`

doc/pages/commands.asciidoc
```diff
@@ -142,9 +142,15 @@ of the file onto the filesystem
     set current session name
 
 *kill[!]* [<exit status>]::
+    *alias* quit-all +
```
krobelus, line 145: As @alexherbo2 pointed out, there is a subtle difference between `:kill` and closing all clients, so this alias might cause confusion. So the inconsisten naming might be somewhat justified. 

For discoverability, maybe we should cross-reference  `kill` in the `quit*` docstrings... but we don't do that anywhere else so I'm not sure.

> So maybe `quit` (client) could inherit the `kill` (server) action.

I probably misunderstood but do you mean that `quit` should yell if it's closing the last client with a dirty buffer backed by a daemonized server? That would regress 2cd323b314e9f2b2cdf11bbf2974e9390914d5ea.

> I wish there was a simple way to add qa as an alias for quit-all

I personally rarely ever use `:kill`, but I use `:quit` a lot.
So I wouldn't care much for a short convenience alias for `:kill`.

alexherbo2:
I agree with @occivink.

`quit-all` is more or less `evaluate-commands -client * quit`, but taking the action as a whole: it could be nice to prevent any actions instead of closing each client one by one and hit a potential yelling message for unsaved buffers in the last one.

`write-all` is more or less `evaluate-commands -buffer * write`.

Qeole:
Thank you all for the feedback!

I do understand the difference between closing all clients (`quit-all`) and killing the server as well (`kill`). I'm not entirely sure what the consensus is, though. Should I:
1. Update the PR to have a separate `quit-all` command which checks that no buffer is left unsaved (unless we have a daemonized server running), and then close all clients?
2. Or instead, focus on a `"*"` argument to pass to `eval -client "*" quit` (which means, as I understand it, some work on `context_wrap()` in commands.cc to support running `eval` on several clients)?

Qeole:
So I went for the first option, and implemented `quit-all[!]` as a separate command which does not kill the session. I dropped `write-all-quit-all` from the PR.

Qeole:
Seems like interest has dropped, I haven't worked myself on this in a while. I'm closing the PR but leaving my branch, in case someone else wishes to pick it up.

PR #4537, opened 2022-02-09
rc ini: use # over ; for Unix-style config files

krobelus:
We set the "ini" filetype for files ending
repo|ini|cfg|properties|desktop.  AFAIK, most of these are actually
use Unix style comments (#) instead of DOS INI comments (;).

These files are using # comments:
*.repo -- search for # in
    https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/sec-configuring_yum_and_yum_repositories
*.desktop: per spec
    https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s03.html#comments
*.cfg: don't know much about this one, but at least the motivating file uses #
    https://github.com/buildout/buildout/blob/master/buildout.cfg
*.properties files:
    https://en.wikipedia.org/wiki/.properties

Set # via a hook. This might look a bit weird but systemd.kak uses
the same hook.

*.ini files traditionally use ; but for example the "foot" terminal's
foot.ini uses #. Let's use ; only if there is no existing # in the
file; this way we should support many files out of the box.


mawww:
This introduces a dependency from ini.kak onto comments.kak (which declares and uses the comment_line option), an alternate approach would be to split ini filetype between regular ini file, and hash commented ones.

krobelus:
> This introduces a dependency from ini.kak onto comments.kak (which declares and uses the comment_line option), an alternate approach would be to split ini filetype between regular ini file, and hash commented ones.

Right, I considered using a separate filetype but discarded it too soon.
Added the "conf" filetype now. 

We'll never realistically detect a significant fraction of all Unix
configuration files but filetype="" is fine for them too.

The second commit adds the hack to work with rogue *.ini files that only
accept #.

PR #4543, opened 2022-02-10
Faces: Allow alpha-blending with non-RGB base colours as well

Qeole:
In `merge_faces()`, we only blend alpha values if the following conditions are met:

```cpp
    if ((base.*color).isRGB() and (face.*color).isRGB() and (face.*color).a != 255)
        return alpha_blend(base.*color, face.*color);
```

The alpha value different from 255 makes sense, otherwise there is nothing to blend. Similarly, if the face is not an `rgba:` (hence RGB) colour, it does not have an alpha value set, so there is no need to proceed.

However, I could find no compelling reason to skip blending if the base colour is not specified as `rgba:`. In that case, blending can still happen, the base colour is simply considered as a plain colour with its alpha value at 255.

The current condition makes it difficult to use a transparent background, since we must have an RGB base first, so we have to proceed in two steps:

```
:set-face global foo ,rgb:ff0000
:add-highlighter global/bar column 80 ,rgba:0000ff44@foo
```

If we allow alpha-blending with all bases, we could instead have something like:

```
:add-highlighter global/bar column 80 ,rgba:ffffff88
```

In particular, this allows merging with the default background. This is useful to blend some background colour (white for a dark background, typically) with the default background colour for the buffer. This is a nice way, for example, to implement a monochrome theme for the different segments of a Powerline modeline, by using a single background colour, and merging it with various level of transparency with the default background.

To achieve this, this PR removes the condition on the base colour being RGB for the alpha-blend.

### DIFF
--- a/src/face.hh
+++ b/src/face.hh
@@ -71,7 +71,7 @@ inline Face merge_faces(const Face& base, const Face& face)
             return base.*color;
         if (face.*color == Color::Default)
             return base.*color;
-        if ((base.*color).isRGB() and (face.*color).isRGB() and (face.*color).a != 255)
+        if ((face.*color).isRGB() and (face.*color).a != 255)
             return alpha_blend(base.*color, face.*color);
         return face.*color;
     };


Screwtapello:
`.isRGB()` doesn't mean "colour was defined with `rgba:`", it means "colour was defined with `rgb:` or `rgba:`". Your use-case of blending an `rgba:` colour on top of an `rgb:` colour should work fine with the code as-is.

However, there are some colours where Kakoune simply does not know the RGB(A) values, and therefore blending is impossible: the named colours (`black`, `red`, `blue`, etc.) and `default` which could be *anything*. That's why Kakoune requires both the base and face colours to be RGB, so that it has enough information to do the alpha calculation at all.

Qeole:
Yes you're right. I looked a bit more and did more testing, and my PR is indeed incorrect. I was initially under the impression that using any base would work because the face foreground would successfully blend with my dark background, when it would in fact blend with `#000000` since no RGB value was found.

I haven't found a good solution for my use case yet (using degrees of transparency for Powerline's segments), but I'll look deeper and see if I can find a way to pass a RGB base. No point in keeping this PR open, I'm closing it. Thank you for the review, and apologies for the noise!

PR #4552, opened 2022-02-13
Add brief modules explanation

dontlaugh:
In lieu of adding a whole docs page for modules, a brief introduction can
be added above the provide-module and require-module command docs.

### DIFF
--- a/doc/pages/commands.asciidoc
+++ b/doc/pages/commands.asciidoc
@@ -423,9 +423,10 @@ but not really useful in that context.
 
 == Module commands
 
-In Kakoune, modules are groups of commands that are evaluated all at once. This allows
-lazy loading of complex configurations with a single invocation of *require-moule*.
-The builtin filetype handling for Kakoune is implemented via modules.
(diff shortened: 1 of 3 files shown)


doc/pages/commands.asciidoc
```diff
@@ -423,6 +423,10 @@ but not really useful in that context.
 
 == Module commands
 
+In Kakoune, modules are groups of commands that are evaluated all at once. This allows
+lazy loading of complex configurations with a single invocation of *require-moule*.
```
Screwtapello, line 427: The important thing about modules is not that they are groups of commands, or that they're all at once (so are the arguments to `eval`), but that they're evaluated lazily. Perhaps something like:

```suggestion
In Kakoune, modules are groups of commands that are stored,
to be executed the first time they are needed.
This allows complex configurations to be evaluated lazily,
and for plugins to ensure their dependencies have been loaded before they execute.
```

dontlaugh: Thanks for the suggestion. I've posted an update.

PR #4558, opened 2022-02-18
Fix elvish highlighter

SolitudeSF:
@xiaq PTAL
adjusted single quote string, to not break on empty string `''`.
highlight keywords as keywords instead of `builtin`.
terminate keyword/builtin detection on `-`/`:`/`\b`, so partial matches dont get highlighted.
misc reordering.

### DIFF
--- a/rc/filetype/elvish.kak
+++ b/rc/filetype/elvish.kak
@@ -22,7 +22,7 @@ provide-module elvish %§
 add-highlighter shared/elvish regions
 add-highlighter shared/elvish/code default-region group
 add-highlighter shared/elvish/double_string region '"' (?<!\\)(\\\\)*" fill string
-add-highlighter shared/elvish/single_string region "'" (?<!')('')*' fill string
+add-highlighter shared/elvish/single_string region "'" ('')*' fill string
 add-highlighter shared/elvish/comment region '#' $ fill comment
 
 add-highlighter shared/elvish/code/variable regex \$[\w\d-_:~]+ 0:variable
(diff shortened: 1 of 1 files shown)


rc/filetype/elvish.kak
```diff
@@ -22,16 +22,17 @@ provide-module elvish %§
 add-highlighter shared/elvish regions
 add-highlighter shared/elvish/code default-region group
 add-highlighter shared/elvish/double_string region '"' (?<!\\)(\\\\)*" fill string
-add-highlighter shared/elvish/single_string region "'" (?<!')('')*' fill string
+add-highlighter shared/elvish/single_string region "'" ('')*' fill string
```
xiaq, line 25: The end pattern can actually be just `"'"` here. String literals that contain two single quotes (e.g. `'I''m'`) can just be highlighted as two separate string literals joined together.

rc/filetype/elvish.kak
```diff
@@ -22,16 +22,17 @@ provide-module elvish %§
...
-add-highlighter shared/elvish/code/keyword regex \}\h+(elif|else|except|finally) 1:keyword
+add-highlighter shared/elvish/code/builtin regex (?:^|\{\s|\(|\||\;)\h*(!=|!=s|%|\*|\+|-gc|-ifaddrs|-log|-override-wcwidth|-stack|-|/|<|<=|<=s|<s|==|==s|>|>=|>=s|>s|all|assoc|base|bool|break|call|cd|compare|constantly|continue|count|defer|deprecate|dissoc|drop|each|eawk|echo|eq|eval|exact-num|exec|exit|external|fail|fg|float64|from-json|from-lines|from-terminated|get-env|has-env|has-external|has-key|has-value|is|keys|kind-of|make-map|multi-error|nop|not-eq|not|ns|num|one|only-bytes|only-values|order|peach|pprint|print|printf|put|rand|randint|range|read-line|read-upto|repeat|repr|resolve|return|run-parallel|search-external|set-env|show|sleep|slurp|src|styled|styled-segment|take|tilde-abbr|time|to-json|to-lines|to-string|to-terminated|unset-env|use-mod|wcswidth)(?![-:])\b 1:builtin
+add-highlighter shared/elvish/code/keyword regex (?:^|\{\s|\(|\||\;)\h*(use|var|set|tmp|del|and|or|coalesce|pragma|while|for|try|fn|if)(?![-:])\b 1:keyword
```
xiaq, line 34: Is there a non-stylistic reason to put special commands in the `keyword` category rather than the `builtin` category?

If it comes down to stylistic choice, I'd prefer my original categorization of putting special commands in `builtin`:

- Special commands are parsed in the same way as ordinary commands in Elvish.
- `and` (special) and `not` (ordinary) should be in the same category; as should `for` (special) and `each` (ordinary).
- Finally Elvish itself highlight special commands like ordinary commands (for the reasons above).

xiaq (commented):
Sorry for the late feedback.

rc/filetype/elvish.kak
SolitudeSF: i put them into keyword since they (to me) are part of the language, not "stdlib", that should be consistent with how other languages highlighted.

rc/filetype/elvish.kak
SolitudeSF: im pretty sure i made it like that because the simplest one was breaking on some edgecases. Cant check now.

PR #4562, opened 2022-03-08
Fix buffer overflow in parse_quoted

tomKPZ:
This fixes a crash when using kak-lsp with bash-language-server.  The
issue is that the second read() in parse_quoted may read past the end of
the string.  If this happens and the condition on line 126 is false,
then the loop on line 119 will continue to read past the end of the
buffer since it checks for state.pos != end instead of state.pos < end,
which will likely result in a crash.  The fix is to add a check for the
buffer end before the second read. The added test fails without the
change and passes with the change.


src/command_manager.cc
```diff
@@ -132,7 +132,7 @@ ParseResult parse_quoted(ParseState& state, Delimiter delimiter)
         if (c == delimiter)
         {
             auto next = state.pos;
-            if (read(next, end) != delimiter)
+            if (next == end || read(next, end) != delimiter)
```
krobelus, line 135: Often, the last read() will return the terminating zero-byte, but that breaks if the input is not zero-terminated.
I wonder when that happens. Anyway, looks good.

PR #4566, opened 2022-03-17
Improved support for front-end languages

in0ni:
I have been using the following files for 4 months now, while actively working on front-end projects. I would like to note the following:

# Highly improved CSS support
CSS support is very much lacking, tons of syntax for a lot of what the spec covers simply does not work. This of course is the base for other languages, like sass. This updated version of the file fully covers all examples. At the very end of this comment I include my test.css file that I used to ensure everything works well. This is where most of my time/learning was spent, it's important to note I'm definitely no expert and have done my best between sample code and reading the docs.

## Assistance with fixing one issue
I have only one thing I was unable to solve, and that is support for the following syntax example `a[attr='value]`, where I cannot get the tag to be properly highlighted. Help to fix this would be greatly appreciated -- again the test.css file below can be used for testing this.

## Improvements on SCSS
Minor improvements, mainly for keywords and variables -- inherits all from css.

# Better commenting support for Javascript
Most of the code is simply pulled from c-family.kak

## Assistance fixing single line comments
This is commented in the code, I could not get it to work... but again, been working for 4 months, and it's a minor issue

# Initial Twig support
Also been working with this for 4 months, works great. I will be making improvements as I go along... been pressed with time and issues I have encountered are so minor. But now that I am finally creating this pull request, I will be sure to update fixes more proactively.

# test.css sample file
I find it very important to include this file as it includes various examples from the CSS spec taken from MDN to ensure the best support possible

```
/* A one-line comment */

/*
span {
  color: blue;
  font-size: 1.5em;
}
*/

/* @note
 * @see https://developer.mozilla.org/en-US/docs/Web/CSS/Reference
 */

@charset "utf-8";

@import url("fineprint.css") print;
@import url("bluish.css") speech;

/* some shit here */
@import "common.css" screen;
@import url("landscape.css") screen and (orientation: landscape);

body {
  background-image: url("https://mdn.mozillademos.org/files/11983/starsolid.gif");
}

#thisIsAnId,
.withAClass,
[data-field] {
  display: none;
}

div:not([lang]) {
  font-style: italic !important;
}

/* Links with "insensitive" anywhere in the URL,
   regardless of capitalization */
a[href*="insensitive" i] {
  color: cyan;
}

/* Links with "cAsE" anywhere in the URL,
   with matching capitalization */
a[href*="cAsE" s] {
  color: pink;
}

a[href^="example"] {
  font-size: 2em;
}

a[href$=".org"] {
  font-style: italic;
}

a[class~="logo"] {
  padding: 2px;
}

span[data-descr]:hover::after,
span[data-descr]:focus::after {
  content: attr(data-descr);
  position: absolute;
  left: 0;
  top: 24px;
  min-width: 200px;
  border: 1px #aaaaaa solid;
  border-radius: 10px;
  background-color: #ffffcc;
  padding: 12px;
  color: #000000;
  font-size: 14px;
  z-index: 1;
}

::selection {
  color: gold;
  background-color: red;
}

h2 + p::first-letter {
  color: white;
  background-color: black;
  border-radius: 2px;
  box-shadow: 3px 3px 0 red;
  font-size: 250%;
  padding: 6px 3px;
  margin-right: 6px;
  float: left;
}

img ~ p {
  color: red;
}

strong {
  color: red;
  background: #000 url(images/bg.gif) no-repeat left top;
  font: italic bold 0.8em/1.2 Arial, sans-serif;
}

div#test span {
  color: green;
}
html div span b a li ul {
  color: blue;
}
span {
  color: red;
}

.foo[style*="color: red"] {
  color: firebrick !important;
}

[id="someElement"] p {
  color: blue;
}

div:not(.outer) p {
  color: blueviolet;
}

div:where(.outer) p {
  color: orange;
}

*#foo {
  color: green;
}

*[id="foo"] {
  color: purple;
  margin: inherit;
}

div.menu-bar li:hover > ul {
  display: block;
}

@keyframes slidein {
  from {
    transform: translateX(0%);
  }

  to {
    transform: translateX(100%);
  }
}

.ani-linear-fn {
  animation-timing-function: linear;
}

.ani-ease-fns {
  animation-timing-function: ease;
  animation-timing-function: ease-in;
  animation-timing-function: ease-out;
  animation-timing-function: ease-in-out;
  animation-timing-function: cubic-bezier(0.1, -0.6, 0.2, 0);
}

.ani-step-fns{
  animation-timing-function: step-start;
  animation-timing-function: step-end;
  animation-timing-function: steps(1, end);
}
/*
 * @note - media types
 * all
 * print
 * screen
 * speech
 */
@media screen {
  li {
    margin-left: 3px;
  }
}

@media print {
  li {
    margin-left: 1px;
  }
}

@media (400px <= width <= 700px) {
  body {
    line-height: 1.4;
  }
}

@media only screen and (min-width: 320px) and (max-width: 480px) and (resolution: 150dpi) {
  body {
    line-height: 1.4;
  }
}
/*
 * @note - logical operators
 */
@supports (display: flex) {
  @media screen and (min-width: 900px) {
    article {
      display: flex;
    }
  }
}

```

### DIFF
--- a/rc/filetype/css.kak
+++ b/rc/filetype/css.kak
@@ -46,10 +46,6 @@ evaluate-commands %sh{
   # includes elements that cannot be styled, which is fine.
   # 
   # https://developer.mozilla.org/en-US/docs/Web/HTML/Element
-  # <CODE>
-  # // selector might change on site updates
-  # var els; document.querySelectorAll('tr td:first-child a[href^="/en-US/docs/Web/HTML/Element/"]').forEach((el) => {els += el.childNodes[0].innerText + " ";}); console.log(els);
-  # </CODE>
   html_tags='html body address article aside footer header h1 h2 h3 h4 h5 h6 main nav section blockquote dd div dl dt figcaption figure hr li ol p pre ul a abbr b bdi bdo br cite code data dfn em i kbd mark q rp rt ruby s samp small span strong sub sup time u var wbr area audio img map track video embed iframe object param picture portal source canvas noscript script del ins caption col colgroup table tbody td tfoot th thead tr button datalist fieldset form input label legend meter optgroup option output progress select textarea details dialog menu summary slot template acronym applet basefont bgsound big blink center content dir font frame frameset hgroup image keygen marquee menuitem nobr noembed noframes plaintext rb rtc shadow spacer strike tt xmp'
 
   # Units
@@ -58,10 +54,6 @@ evaluate-commands %sh{
   # includes #rgb, #rrggbb, #rrggbbaa as color values {3,8}
   # 
   # https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Values_and_Units
-  # <CODE>
-  # // selector might change on site updates
-  # var units; document.querySelectorAll('tr td:first-child').forEach((el) => {els += el.childNodes[0].innerText + " ";}); console.log(units)
-  # </CODE>
   units='% cap ch cm deg dpcm dpi dppx em ex grad Hz ic in kHz lh mm ms pc pt px Q rad rem rlh s turn vb vh vi vmax vmin vw x'
 
   logical_ops='and not only from to'
@@ -84,34 +76,26 @@ evaluate-commands %sh{
 
   # order below matters
   printf %s "
-  # html tag selectors
-  add-highlighter shared/css/code/ regex \b($(join "${html_tags}"))((:[a-z:])|[\h.#]) 1:keyword
+  add-highlighter shared/css/code/tag_selectors regex \b($(join "${html_tags}"))((:[a-z:])|[\h.#]) 1:keyword
 
-  #functional notation
-  add-highlighter shared/css/code/ regex ([a-zA-Z0-9-_]+[a-zA-Z0-9])\( 1:keyword
+  add-highlighter shared/css/code/functional_notation regex ([a-zA-Z0-9-_]+[a-zA-Z0-9])\( 1:keyword
 
-  # logical operators, and other keywords
-  add-highlighter shared/css/code/ regex (\b($(join "${logical_ops}"))\b|$(join "${keywords}")) 1:keyword 1:+i
+  add-highlighter shared/css/code/logical_operators regex (\b($(join "${logical_ops}"))\b|$(join "${keywords}")) 1:keyword 1:+i
 
-  # media types
-  add-highlighter shared/css/code/ regex \b($(join "${media_types}"))\b 1:+i
+  add-highlighter shared/css/code/media_types regex \b($(join "${media_types}"))\b 1:+i
 
-  # pseudo (after functional notation as they may contain paranthesis)
-  add-highlighter shared/css/code/ regex (:{1,2})([a-z-]+) 2:attribute 2:+a
+  # (after functional notation as they may contain paranthesis)
+  add-highlighter shared/css/code/psudo regex (:{1,2})([a-z-]+) 2:attribute 2:+a
 
-  # at-rules
-  add-highlighter shared/css/code/ regex @[a-z-]+ 0:function 
+  add-highlighter shared/css/code/at_rules regex @[a-z-]+ 0:function 
 
-  # css property
-  add-highlighter shared/css/code/ regex ([A-Za-z][A-Za-z0-9_-]*)\h*:\h 1:operator 1:+a
+  add-highlighter shared/css/code/css_property regex ([A-Za-z][A-Za-z0-9_-]*)\h*:\h 1:operator 1:+a
 
-  # universal, class, id selectors
-  add-highlighter shared/css/code/ regex (\*|[*]?[.][A-Za-z][A-Za-z0-9_-]+) 1:type
-  add-highlighter shared/css/code/ regex (\*|[*]?[#][A-Za-z][A-Za-z0-9_-]+) 1:type 1:+i
+  add-highlighter shared/css/code/selectors regex (\*|[*]?[.][A-Za-z][A-Za-z0-9_-]+) 1:type
+  add-highlighter shared/css/code/selectors_id regex (\*|[*]?[#][A-Za-z][A-Za-z0-9_-]+) 1:type 1:+i
 
-  # hex values
-  add-highlighter shared/css/code/ regex (#[0-9A-Fa-f]{3,8})\b 0:value 0:+a
-  add-highlighter shared/css/code/ regex \b(\d*\.)?\d+($(join "${units}"))?\b 0:value 0:+a
+  add-highlighter shared/css/code/hex_values regex (#[0-9A-Fa-f]{3,8})\b 0:value 0:+a
+  add-highlighter shared/css/code/units regex \b(\d*\.)?\d+($(join "${units}"))?\b 0:value 0:+a
   "
 }
 
--- a/rc/filetype/css.kak
+++ b/rc/filetype/css.kak
@@ -92,8 +92,8 @@ evaluate-commands %sh{
 
   add-highlighter shared/css/code/css_property regex ([A-Za-z][A-Za-z0-9_-]*)\h*:\h 1:operator 1:+a
 
-  add-highlighter shared/css/code/selectors regex (\*|[*]?[.][A-Za-z][A-Za-z0-9_-]+) 1:type
-  add-highlighter shared/css/code/selectors_id regex (\*|[*]?[#][A-Za-z][A-Za-z0-9_-]+) 1:type 1:+i
+  add-highlighter shared/css/code/selectors regex (\*|[*]?[.][A-Za-z][A-Za-z0-9_-]+) 0:type
+  add-highlighter shared/css/code/selectors_id regex (\*|[*]?[#][A-Za-z][A-Za-z0-9_-]+) 0:type 0:+i
 
   add-highlighter shared/css/code/hex_values regex (#[0-9A-Fa-f]{3,8})\b 0:value 0:+a
   add-highlighter shared/css/code/units regex \b(\d*\.)?\d+($(join "${units}"))?\b 0:value 0:+a
--- a/rc/filetype/css.kak
+++ b/rc/filetype/css.kak
@@ -86,7 +86,7 @@ evaluate-commands %sh{
   add-highlighter shared/css/code/media_types regex \b($(join "${media_types}"))\b 1:+i
 
   # (after functional notation as they may contain paranthesis)
-  add-highlighter shared/css/code/pseudo regex (:{1,2})([a-z-]+) 2:attribute
+  add-highlighter shared/css/code/pseudo regex (:{1,2})([a-z-]+) 0:attribute
 
   add-highlighter shared/css/code/at_rules regex @[a-z-]+ 0:function 
 
--- a/rc/filetype/css.kak
+++ b/rc/filetype/css.kak
@@ -19,6 +19,7 @@ hook global WinSetOption filetype=css %[
     hook window InsertChar \n -group css-indent css-indent-on-new-line
     hook window InsertChar \} -group css-indent css-indent-on-closing-curly-brace
     set-option buffer extra_word_chars '_' '-'
+    set-face global cssLogicalOperator +i@keyword
 
     hook -once -always window WinSetOption filetype=.* %{ remove-hooks window css-.+ }
 ]
@@ -76,11 +77,11 @@ evaluate-commands %sh{
 
   # order below matters
   printf %s "
-  add-highlighter shared/css/code/tag_selectors regex \b($(join "${html_tags}"))((:[a-z:])|[\h.#]) 1:keyword
+  add-highlighter shared/css/code/tag_selectors regex \b($(join "${html_tags}"))((:[a-z:])|[\h.#,]) 0:keyword
 
   add-highlighter shared/css/code/functional_notation regex ([a-zA-Z0-9-_]+[a-zA-Z0-9])\( 1:keyword
 
-  add-highlighter shared/css/code/logical_operators regex (\b($(join "${logical_ops}"))\b|$(join "${keywords}")) 1:keyword 1:+i
+  add-highlighter shared/css/code/logical_operators regex (\b($(join "${logical_ops}"))\b|$(join "${keywords}")) 1:cssLogicalOperator
 
   add-highlighter shared/css/code/media_types regex \b($(join "${media_types}"))\b 1:+i
 
--- a/rc/filetype/css.kak
+++ b/rc/filetype/css.kak
@@ -86,7 +86,7 @@ evaluate-commands %sh{
   add-highlighter shared/css/code/media_types regex \b($(join "${media_types}"))\b 1:+i
 
   # (after functional notation as they may contain paranthesis)
-  add-highlighter shared/css/code/pseudo regex (:{1,2})([a-z-]+) 2:attribute 2:+a
+  add-highlighter shared/css/code/pseudo regex (:{1,2})([a-z-]+) 2:attribute
 
   add-highlighter shared/css/code/at_rules regex @[a-z-]+ 0:function 
 
--- a/rc/filetype/css.kak
+++ b/rc/filetype/css.kak
@@ -77,7 +77,7 @@ evaluate-commands %sh{
 
   # order below matters
   printf %s "
-  add-highlighter shared/css/code/tag_selectors regex \b($(join "${html_tags}"))((:[a-z:])|[\h.#,]) 0:keyword
+  add-highlighter shared/css/code/tag_selectors regex \b($(join "${html_tags}"))((:[a-z:])|[\h.#,]) 1:keyword
 
   add-highlighter shared/css/code/functional_notation regex ([a-zA-Z0-9-_]+[a-zA-Z0-9])\( 1:keyword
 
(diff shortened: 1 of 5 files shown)


rc/filetype/twig.kak
```diff
@@ -0,0 +1,95 @@
...
+# Initialization
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+hook global WinSetOption filetype=twig %[
+    require-module twig
+
+    hook window ModeChange pop:insert:.* -group twig-trim-indent  twig-trim-indent
+    hook window InsertChar \n -group twig-insert twig-insert-on-new-line
+    hook window InsertChar \n -group twig-indent twig-indent-on-new-line
+    hook window InsertChar '>' -group twig-indent twig-indent-on-grater-than
```
Screwtapello, line 20: There's a typo in this command name.

```suggestion
    hook window InsertChar '>' -group twig-indent twig-indent-on-greater-than
```

rc/filetype/twig.kak
```diff
@@ -0,0 +1,95 @@
...
+# delimiters
+add-highlighter shared/twig/delim/base/ regex (\{[%]|[%]\}) 0:function
+add-highlighter shared/twig/delim/base/ regex (\{\{|\}\}) 0:operator
+
+# Commands
+# ‾‾‾‾‾‾‾‾
+
+define-command -hidden twig-trim-indent                   html-trim-indent
+define-command -hidden twig-indent-on-new-line            html-indent-on-new-line
+define-command -hidden twig-indent-on-grater-than         html-indent-on-greater-than
```
Screwtapello, line 74: Typo here too.

```suggestion
define-command -hidden twig-indent-on-greater-than         html-indent-on-greater-than
```

in0ni: Good eye, thanks!

in0ni:
@Screwtapello I am wondering.... if you have the time, to help me out with the issue I cannot figure out with the one CSS example I outline in my comment? (the second heading). It's quite simple, just have not been able to get anywhere -- have tried so many things. If not no worries, figured I'd ping you as you were quick to do a review.

Thanks!

rc/filetype/twig.kak
```diff
@@ -0,0 +1,95 @@
...
+add-highlighter shared/twig/delim region \{([%]-?|\{) (-?[%]|\})\} regions
+
+add-highlighter shared/twig/core/ ref html
+
+add-highlighter shared/twig/delim/base default-region group
+add-highlighter shared/twig/delim/double_string region '"'  (?<!\\)(\\\\)*" fill string
+add-highlighter shared/twig/delim/single_string region "'"  (?<!\\)(\\\\)*' fill string
+add-highlighter shared/twig/delim/base/ regex (\w+)\h= 1:variable
+
+# functions
```
mawww, line 53: same, name this explicitely `shared/twig/delim/base/functions` and drop the comment

rc/filetype/css.kak
```diff
@@ -32,28 +33,94 @@ provide-module css %[
...
+  # Selectors
+  # ‾‾‾‾‾‾‾‾‾
+  # universal: *, ns|*, *|*, |*
+  # class/id: .class, #id
+  # type: element
+  # attr: [attr=val]
+
+  # order below matters
+  printf %s "
+  # html tag selectors
```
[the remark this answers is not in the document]
in0ni: Sounds good, will get to this change soon, test, and resolve when done.

rc/filetype/twig.kak
in0ni: Sounds good, will do same, test and resolve when done.

rc/filetype/css.kak
```diff
@@ -32,28 +33,94 @@ provide-module css %[
...
+evaluate-commands %sh{
+  # html tags
+  # generated from the URL & <code> below
+  # includes elements that cannot be styled, which is fine.
+  # 
+  # https://developer.mozilla.org/en-US/docs/Web/HTML/Element
+  # <CODE>
+  # // selector might change on site updates
+  # var els; document.querySelectorAll('tr td:first-child a[href^="/en-US/docs/Web/HTML/Element/"]').forEach((el) => {els += el.childNodes[0].innerText + " ";}); console.log(els);
+  # </CODE>
```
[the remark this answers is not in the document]
in0ni: Ok, saw similar comments on another filetype -- and yes, can grow out of date, but I have no problem maintaining and making sure it's up to date. I am using this as a main editor, and once changes are in I will continue contributing (for example already seeing I need to update some sass stuff. I will remove this for now.

rc/filetype/css.kak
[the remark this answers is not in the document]
in0ni: Ok, have added explicitly named highlighters for each

rc/filetype/twig.kak
in0ni: Have dropped the comments for all, and explicitly named.

in0ni:
@mawww, requested changes have been made.

I am wondering if you have the time, could you point me in the right direction in solving the issue outlined in my original comment regarding the CSS? I have tried multiple things and I am unable to figure it out. I am adding a image to better see the issue, it has to do with html tags followed by attributes... they are not highlighted as they should be (in red, as the DIV shown in screenshot). I have included the test.css file I use in my original comment as well :)

image

rc/filetype/css.kak
```diff
@@ -32,28 +33,78 @@ provide-module css %[
...
+  # Selectors
+  # ‾‾‾‾‾‾‾‾‾
+  # universal: *, ns|*, *|*, |*
+  # class/id: .class, #id
+  # type: element
+  # attr: [attr=val]
+
+  # order below matters
+  printf %s "
+  add-highlighter shared/css/code/tag_selectors regex \b($(join "${html_tags}"))((:[a-z:])|[\h.#]) 1:keyword
```
Screwtapello, line 79: This change makes HTML tags get highlighted as "keyword" wherever they appear, if that helps. It looks OK for me with your `test.css`, but I assume there was a *reason* you tried to make HTML tags only recognised when followed by a psuedo-class, class, ID, or space.
```suggestion
  add-highlighter shared/css/code/tag_selectors regex \b$(join "${html_tags}")\b 0:keyword
```

rc/filetype/css.kak
```diff
@@ -32,28 +33,78 @@ provide-module css %[
...
 
-add-highlighter shared/css/declaration/base default-region group
-add-highlighter shared/css/declaration/double_string region '"' (?<!\\)(\\\\)*" fill string
-add-highlighter shared/css/declaration/single_string region "'" "'"             fill string
-add-highlighter shared/css/declaration/comment region /[*] [*]/ fill comment
+  add-highlighter shared/css/code/functional_notation regex ([a-zA-Z0-9-_]+[a-zA-Z0-9])\( 1:keyword
 
-# https://developer.mozilla.org/en-US/docs/Web/CSS/length
-add-highlighter shared/css/declaration/base/ regex (#[0-9A-Fa-f]+)|((\d*\.)?\d+(ch|cm|em|ex|mm|pc|pt|px|rem|vh|vmax|vmin|vw)) 0:value
+  add-highlighter shared/css/code/logical_operators regex (\b($(join "${logical_ops}"))\b|$(join "${keywords}")) 1:keyword 1:+i
```
Screwtapello, line 83: Shouldn't "logical operators" get the "operator" face? Also, are there other highlighters that add extra decoration to defined faces? Unless somebody else chimes in to correct me, I'd expect the preference to be:

- use an existing face as-is
- define a new face based on an existing face, so themes can still override it if they want

For an example of the latter, it would be something like:

```kak
set-face global cssLogicalOperator +i@keyword
```

...and then you can use `cssLogicalOperator` anywhere a face-name is needed.

rc/filetype/css.kak
```diff
@@ -32,28 +33,78 @@ provide-module css %[
...
 
-add-highlighter shared/css/declaration/base/ regex ([A-Za-z][A-Za-z0-9_-]*)\h*: 1:keyword
-add-highlighter shared/css/declaration/base/ regex :(before|after) 0:attribute
-add-highlighter shared/css/declaration/base/ regex !important 0:keyword
+  add-highlighter shared/css/code/media_types regex \b($(join "${media_types}"))\b 1:+i
```
Screwtapello, line 85: As above, this should be a standard face, or a custom face that defaults to a standard face (possibly with some alterations)

rc/filetype/css.kak
```diff
@@ -32,28 +33,78 @@ provide-module css %[
...
 
-# element#id element.class
-# universal selector
-add-highlighter shared/css/selector/ regex         [A-Za-z][A-Za-z0-9_-]* 0:keyword
-add-highlighter shared/css/selector/ regex [*]|[#.][A-Za-z][A-Za-z0-9_-]* 0:variable
+  # (after functional notation as they may contain paranthesis)
+  add-highlighter shared/css/code/psudo regex (:{1,2})([a-z-]+) 2:attribute 2:+a
```
Screwtapello, line 88: Typo: "pseudo"

You might as well include the `:` or `::` in the highlighting, since the punctuation is highlighted in `.class` and `#id` selectors.

Why `+a`? That means other styles won't be able to override attributes (bold, underline, etc.) but you don't know what the attributes of the `attribute` face are, so this seems brittle.

rc/filetype/css.kak
```diff
@@ -32,28 +33,78 @@ provide-module css %[
...
+
+  add-highlighter shared/css/code/at_rules regex @[a-z-]+ 0:function 
+
+  add-highlighter shared/css/code/css_property regex ([A-Za-z][A-Za-z0-9_-]*)\h*:\h 1:operator 1:+a
+
+  add-highlighter shared/css/code/selectors regex (\*|[*]?[.][A-Za-z][A-Za-z0-9_-]+) 1:type
+  add-highlighter shared/css/code/selectors_id regex (\*|[*]?[#][A-Za-z][A-Za-z0-9_-]+) 1:type 1:+i
```
Screwtapello, line 95: I feel like a simpler definition of the selectors rule might be:

```suggestion
  add-highlighter shared/css/code/selectors regex \*|[.#][A-Za-z][A-Za-z0-9_-]+ 0:type
```

That is, a `*`, or a `.` or `#` followed by alphanumerics. If somebody chains selectors together like `***.class*#id` then each selector in the list should be identified individually, and still be displayed the same.

Screwtapello (commented):
Hi, sorry it's taken a while to get around to looking at your code. I'm not a front-end developer myself, so I wasn't sure exactly what behaviour you were expecting, but once you posted a screenshot I became more confident I could figure out what was going on.

rc/filetype/css.kak
in0ni: Yes, there is a reason. Simple html tags (like one letter ones) will cause problems, here on the screenshot I am unsure why it's not respecting word boundaries... but I did have it working and was causing problems when a simple tag (like a, b, i) would be preceded or followed by say a hyphen.

Here is what will happen if I follow your suggestion (copied your suggestion exactly as you have it). Actually I have added a comma recently, and just need to figure out how to have it be followed by a '[' which I somehow can't escape.
image

rc/filetype/css.kak
in0ni: There are few, and not widely used logical operators and I decided to follow similar styling to vscode (which uses same color as html tags. Extra decorations help and I don't see any issues w/ being able to use them if they are available.

I could definitely create a new face if that makes things cleaner, and will commit.

rc/filetype/css.kak
in0ni: Again, using simply italics here and following vscode styles mostly -- with default face. Not sure what the issue or concern is.

rc/filetype/css.kak
in0ni: Other editors do not include the `:` and `::` highlighted, again tried to follow existing styles, rather than just create my own.

Regarding `+a`... there must of been a reason I added this initially, have removed it and don't see it making a difference so, I'll commit that as well.

rc/filetype/css.kak
in0ni: As other editors, I wanted to highlight an ID selector (#) with italics -- as this is somewhat important and unique.

Also, the chaining you speak of is invalid, and would not even work, so it's not a valid use case as far as I know.

I will change the `1:type` to `0:type` -- in other cases above your suggestion of using  `0` is problematic, as it will highlight items not wanted.

rc/filetype/css.kak
in0ni: This still needs the question, and help of determining how to have a tag be followed by '['... I just can't find how to escape it! Could you help me sort this out :)?

in0ni:
@mawww -- I am showing that there are changes requested which I have taken care of... am I missing something here? Would like to see if we can commit these changes if possible to improve support especially for CSS.

rc/filetype/css.kak
Screwtapello: > Simple html tags (like one letter ones) will cause problems, here on the screenshot I am unsure why it's not respecting word boundaries...

Ah, right, I see. `\b$(join "$html_tags")\b` will expand to something like:

    html|b|body

which turns the regex into:

    \bhtml|b|body\b

...which will match `html` at the beginning of a word, `body` at the end of a word, or `b` anywhere. I removed one too many grouping brackets, so you need to put one back:

    \b($(join "$html_tags"))\b

...which will expand to:

    \b(html|b|body)\b

...so that HTML tags are always required to be standalone words, not appearing within some other word.

rc/filetype/css.kak
Screwtapello: Kakoune's model is that filetype plugins (like this one) apply named faces to the buffer text, and colorscheme plugins decide what each named face looks like. If somebody wants to make a VS Code colorscheme, they can choose to style operators the same way VS Code does; if somebody's using Kakoune it's probably more helpful if the CSS keywords and operators in Kakoune look like the Python or Rust keywords and operators in Kakoune, rather than the CSS keywords and operators in some other editor.

rc/filetype/css.kak
Screwtapello: I figure `:` and `::` are as much a part of the "spelling" of a (pseudo)class as `.` or `#` are part of the "spelling" of class or ID selectors. It may be that there's a good syntactical reason to treat them separately, but I think Kakoune being consistent with itself is more important than being consistent with some other editor.

rc/filetype/css.kak
in0ni: That still causes issues in cases like this, the word boundary I am afraid is not enough:
image

rc/filetype/css.kak
in0ni: Ok, keeping them as part of spelling -- committed and pushed.

rc/filetype/javascript.kak
```diff
@@ -55,21 +55,112 @@ define-command -hidden javascript-indent-on-char %<
 
 define-command -hidden javascript-insert-on-new-line %<
     evaluate-commands -draft -itersel %<
-        # copy // comments prefix and following white spaces
-        try %{ execute-keys -draft k <a-x> s ^\h*\K/{2,}\h* <ret> y gh j P }
+    execute-keys <semicolon>
+    # NOT working, copies single line comment "//"
```
mawww, line 59: I missed those commented blocks when I merged, this needs to be fixed

rc/filetype/javascript.kak
```diff
@@ -55,21 +55,112 @@ define-command -hidden javascript-indent-on-char %<
...
+            ]
+        ]
+
+        # trim trailing whitespace on the previous line
+        try %[ execute-keys -draft s\h+$<ret> d ]
+        # align the new star with the previous one
+        execute-keys K<a-x>1s^[^*]*(\*)<ret>&
+    ]
+        # # copy // comments prefix and following white spaces
+        # try %{ execute-keys -draft k <a-x> s ^\h*\K/{2,}\h* <ret> y gh j P }
```
mawww, line 106: same

rc/filetype/javascript.kak
```diff
@@ -55,21 +55,112 @@ define-command -hidden javascript-indent-on-char %<
...
+        # Go to opening parenthesis and opening brace, then select the most nested one
+        try %< execute-keys [c [({],[)}] <ret> >
+        # Validate selection and get first and last char
+        execute-keys <a-k>\A[{(](\h*\S+)+\n<ret> <a-K>"(([^"]*"){2})*<ret> <a-K>'(([^']*'){2})*<ret> <a-:><a-semicolon>L <a-S>
+        # Remove possibly incorrect indent from new line which was copied from previous line
+        try %< execute-keys -draft <space> <a-h> s\h+<ret> d >
+        # Now indent and align that new line with the opening parenthesis/brace
+        execute-keys 1<a-&> &
+     > >
+        # # preserve previous line indent
```
mawww, line 156: this block as well

mawww (commented):
I merged this thinking it had been reviewed already, but I missed the introduction of commented blocks of code, that code needs to either be removed, or fixed, I would like to avoid commented code in the published support files.

rc/filetype/javascript.kak
in0ni: sorry, I was unable to get this working and overlooked it -- took the time to troubleshoot and it's now fixed. Actually the commented blocks mentioned below are what made it work.

rc/filetype/javascript.kak
in0ni: has been removed

rc/filetype/javascript.kak
in0ni: same, removed -- commenting works as expected now.

in0ni:
@mawww do you merge the latest changes, or do I create a new merge request?

PR #4568, opened 2022-03-17
Fix cmake comment

CharlesGueunet:
This MR allows for comment to be indented and inside functions:

image

Before:
image

This is my first contribution regarding highlighter so it may not be the best way of doing this.


rc/filetype/cmake.kak
```diff
@@ -19,12 +19,13 @@ provide-module cmake %{
 
 add-highlighter shared/cmake regions
 add-highlighter shared/cmake/code default-region group
-add-highlighter shared/cmake/comment  region '#' '$' fill comment
+add-highlighter shared/cmake/comment  region '\s*#' '$' fill comment
```
mawww, line 22: That does not look necessary, the only effect of this would be to make the comment region start earlier but I dont expect any other regions to start in those whitespaces.

CharlesGueunet: you are right, fixed.

PR #4575, opened 2022-03-26
Make Shift+Space insert a space in insert mode

### LINKED ISSUE #4534: [BUG] Pressing space has no effect while holding the shift key
### Version of Kakoune

v2021.11.08

### Reproducer

Open a document, press `i` to enter insert mode. Hold shift and type `# HELLO THERE`.

### Outcome

The text appears as `#HELLOTHERE`.

### Expectations

The text should appear as `# HELLO THERE`.

### Additional information

Delete also does not work while holding shift. In the terminal or in other programs, holding shift and pressing space inserts a space.

macOS 12.1

krobelus:
Terminals that support CSI u escape codes (like iTerm2, Kitty and foot)
allow us to map `<s-space>` independently of `<space>`.

Users expect that `<s-space>` inputs a space character; make it so.

Fixes #4534
Also reported in https://discuss.kakoune.com/t/shift-space-doesnt-send-space-character/2004

### DIFF
--- a/src/keys.cc
+++ b/src/keys.cc
@@ -49,7 +49,7 @@ Optional<Codepoint> Key::codepoint() const
         return '\n';
     if (*this == Key::Tab)
         return '\t';
-    if (*this == Key::Space)
+    if (*this == Key::Space or *this == shift(Key::Space))
         return ' ';
     if (*this == Key::Escape)
         return 0x1B;
--- a/src/input_handler.cc
+++ b/src/input_handler.cc
@@ -1003,7 +1003,7 @@ public:
         }
         else
         {
-            if (key == Key::Space and
+            if ((key == Key::Space or key == shift(Key::Space)) and
                 not (m_completions.flags & Completions::Flags::Quoted) and // if token is quoted, this space does not end it
                 can_auto_insert_completion())
                 m_line_editor.insert_from(line.char_count_to(m_completions.start),


src/keys.cc
```diff
@@ -49,7 +49,7 @@ Optional<Codepoint> Key::codepoint() const
         return '\n';
     if (*this == Key::Tab)
         return '\t';
-    if (*this == Key::Space)
+    if (*this == Key::Space or *this == shift(Key::Space))
```
krobelus, line 52: I guess with the same logic `<s-ret>` should also insert a newline in insert mode, as should `<s-tab>`.
Vim behaves that  way.

Screwtapello: But `<s-tab>` is backtab, and people expect it to unindent.

krobelus: > But `<s-tab>` is backtab, and people expect it to unindent.

Makes sense although it would be weird (for Kakoune) if <s-tab> in insert mode unindented.

---

Some other keys that don't work OOTB in terminals like foot are
<c-[> for Escape
<c-j> or <c-m> for Return
<c-i> for Tab

I'm not sure where to draw the line.

I guess <c-[> doubles as escape purely for historical reasons,
whereas <c-s> inserting a space makes sense even if it wasn't historical behavior.

PR #4578, opened 2022-04-05
rc/filetype/go: add new go 1.18 predeclared identifiers

inahga:
rc/filetype/go: add new go 1.18 predeclared identifiers

Go 1.18 introduces the `any` and `comparable` predeclared identifiers. Modify
the list of identifiers here, so syntax highlighting will catch these new
identifiers. See https://go.dev/ref/spec#Predeclared_identifiers.


inahga:
I also just noticed `uintptr` is missing, went ahead and added that as well.

PR #4589, opened 2022-04-11
[Docs: doc] Amendments to the language

Pound-Hash:
I hope the changes are self-explanatory.

### DIFF
--- a/rc/tools/doc.asciidoc
+++ b/rc/tools/doc.asciidoc
@@ -2,33 +2,35 @@
 
 This is Kakoune's online documentation system.
 
-To see what documentation topics are available, type `:doc` and look at the
-completion menu. To view the a particular documentation topic, type the topic
-name or choose it from the completion menu and hit Enter.
+To see what documentation topics are available,
+type `:doc` and look at the completion menu.
+To view a particular topic,
+type its name or select it from the completion menu.
+Then hit Enter.
 
 Documentation will be displayed in the client named in the `docsclient` option.
 
 == Using the documentation browser
 
-Documentation buffers are like any other buffer, so you can scroll through them
-as normal, search within a topic with `/`, etc. However, they can also contain
-links, <<doc#demonstration-target,like this>>. When you see a link, you can
-follow it by moving the cursor onto it and pressing Enter. If the link takes you
-to a different documentation topic, you can get back by using the `:buffer`
-command.
+Documentation buffers are like any other buffer,
+so you can scroll through them as normal,
+search within a topic with `/`, etc.
+However, they can also contain links: <<doc#demonstration-target,like this>>.
+Links can be followed by moving the cursor onto them and pressing Enter.
+If a link takes you to a different documentation topic,
+you can return to the original by using the `:buffer` command.
 
 == Writing documentation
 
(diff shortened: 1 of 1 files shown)


rc/tools/doc.asciidoc
```diff
@@ -2,33 +2,35 @@
...
 This is Kakoune's online documentation system.
 
-To see what documentation topics are available, type `:doc` and look at the
-completion menu. To view the a particular documentation topic, type the topic
-name or choose it from the completion menu and hit Enter.
+To see what documentation topics are available,
+type `:doc` and look at the completion menu.
+To view a particular topic,
+type its name or select it from the completion menu.
+Then hit Enter.
```
krobelus, line 9: Ok there are two changes in this paragraph
1. convert to one-sentence-per-line
2. grammar fixes (which look good)

I think converting to one-sentence-per-line should be a separate PR, and if
we want it, we should probably change everything under doc/ to maintain
a consistent style. I personally don't think it's worth it. Granted,
one-sentence-per-line makes future diffs less noisy but do we actually
have so many doc changes that it warrants making sources less pleasing to read?

Separating style from content changes would also make this easier to review.

rc/tools/doc.asciidoc
```diff
@@ -2,33 +2,35 @@
...
+you can return to the original by using the `:buffer` command.
 
 == Writing documentation
 
-Documentation must be in AsciiDoc format, with the extension `.asciidoc`,
-and stored somewhere within <<doc#sources,the documentation search path>>.
-Kakoune's built-in documentation renderer does not necessarily support every
-feature, so don't go overboard with formatting.
+Documentation must be in AsciiDoc format, with the extension `.asciidoc`.
+It must be stored somewhere within <<doc#sources,the documentation search path>>.
```
krobelus, line 26: Possible alternative:

	Documentation must be in AsciiDoc files with an `.asciidoc` extension within <<doc#sources,the documentation search path>>.

rc/tools/doc.asciidoc
Pound-Hash: I'm new to this. Another developer directed me to some article explaining that we should try to break lines across clauses. So that's what I was aiming for, not at sentences. I can start this revision over again form the beginning.

rc/tools/doc.asciidoc
Pound-Hash: > Possible alternative:
> 
> ```
> Documentation must be in AsciiDoc files with an `.asciidoc` extension within <<doc#sources,the documentation search path>>.
> ```

Sure, concision is always a goal.

rc/tools/doc.asciidoc
krobelus: Every project has a different style; the style of this project's *.asciidoc files appears to be to hard-wrap lines at 80 columns (using `|fmt -w 80` for example).

Pound-Hash: > Docs should be easy to read unrendered as well, so I'd like to get back to nicely wrapped paragraphs, can you fix that please ?

I've wrapped all the lines constituting paragraphs at column 80.

How's that?

PR #4590, opened 2022-04-12
Disable external diff tools when diffing buffer

AndrewVos:
There is a bug that causes `:git show-diff` to fail when using an external diff, for example `difftastic`.

This change ensures that we don't use an external diff tool when diffing the current buffer.

To replicate the bug, try changing your difftool to something else (`git config --global diff.external difft
`) and then use `:git show-diff`. Nothing will show.

### DIFF
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -157,7 +157,7 @@ define-command -params 1.. \
     update_diff() {
         (
             cd_bufdir
-            git --no-pager diff -U0 "$kak_buffile" | perl -e '
+            git --no-pager diff --no-ext-diff -U0 "$kak_buffile" | perl -e '
             $flags = $ENV{"kak_timestamp"};
             foreach $line (<STDIN>) {
                 if ($line =~ /@@ -(\d+)(?:,(\d+))? \+(\d+)(?:,(\d+))?/) {


Screwtapello:
> There is a bug...

Is this a bug in Kakoune, or in git?

AndrewVos:
In kakoune imho

Screwtapello:
Is that bug filed? It'd be nice to fix the underlying problem, rather than just patching over the consequences.

AndrewVos:
This is fixing the underlying problem

krobelus:
This is simply a bug in the git.kak script.
We must not make any assumptions about the output format of `git diff` without `--no-ext-diff`, because a user might have set `diff.external` (globally).
For example, difftastic will print a side-by-side diff, which obviously breaks our parser that expects a unified diff:

```
git -c diff.external=difft diff origin/master origin/pr-4590
rc/tools/git.kak --- Text
157     update_diff() {                                                                                                                      157     update_diff() {
158         (                                                                                                                                158         (
159             cd_bufdir                                                                                                                    159             cd_bufdir
160             git --no-pager diff -U0 "$kak_buffile" | perl -e '                                                                           160             git --no-pager diff --no-ext-diff -U0 "$kak_buffile" | perl -e '
161             $flags = $ENV{"kak_timestamp"};                                                                                              161             $flags = $ENV{"kak_timestamp"};
162             foreach $line (<STDIN>) {                                                                                                    162             foreach $line (<STDIN>) {
163                 if ($line =~ /@@ -(\d+)(?:,(\d+))? \+(\d+)(?:,(\d+))?/) {                                                                163                 if ($line =~ /@@ -(\d+)(?:,(\d+))? \+(\d+)(?:,(\d+))?/) {
```

Git porcelain commands like `diff` are optimized for interactive use, not for portable scripting.
OTOH, there are plumbing commands like `diff-index`, `diff-files` and `diff-tree` that are more predictable; they ignore `diff.external`.
We could use diff-files but I'm not sure if there's a need to.

AndrewVos:
> This is simply a bug in the git.kak script. We must not make any assumptions about the output format of `git diff` without `--no-ext-diff`, because a user might have set `diff.external` (globally). For example, difftastic will print a side-by-side diff, which obviously breaks our parser that expects a unified diff:
> 
> ```
> $ git -c diff.external=difft diff origin/master origin/pr-4590
> rc/tools/git.kak --- Text
> 157     update_diff() {                                                          157     update_diff() {
> 158         (                                                                    158         (
> 159             cd_bufdir                                                        159             cd_bufdir
> 160             git --no-pager diff -U0 "$kak_buffile" | perl -e '               160             git --no-pager diff --no-ext-diff -U0 "$kak_buffile" | perl -e '
> 161             $flags = $ENV{"kak_timestamp"};                                  161             $flags = $ENV{"kak_timestamp"};
> 162             foreach $line (<STDIN>) {                                        162             foreach $line (<STDIN>) {
> 163                 if ($line =~ /@@ -(\d+)(?:,(\d+))? \+(\d+)(?:,(\d+))?/) {    163                 if ($line =~ /@@ -(\d+)(?:,(\d+))? \+(\d+)(?:,(\d+))?/) {
> ```
> 
> Git porcelain commands like `diff` are optimized for interactive use, not for portable scripting. OTOH, there are plumbing commands like `diff-index`, `diff-files` and `diff-tree` that are more predictable; they ignore `diff.external`. We could use diff-files but I'm not sure if there's a need to.

Thanks for explaining way better than I did!

Screwtapello:
If we need machine-readable output in a specific format, and `git diff-files` is the reliable way to produce that format, that's probably what we should use. It seems like we already have to provide a few command-line arguments to bend `git diff` output into our preferred shape, and I bet we'll have to add more in future as `git diff` gets friendlier.

That's definitely something we can do later, though — this PR is good enough for now.

krobelus:
Right, probably it'd be best practice to replace

	git --no-pager diff --no-ext-diff

with the equivalent

	git diff-files

AndrewVos:
Not sure of what the next step is here, is this okay as-is or should I move to `diff-files`?

mawww:
This is a step in the good direction so merging, we can move to diff-files as a separate change

PR #4591, opened 2022-04-13
[README] Fixes to grammar and punctuation

Pound-Hash:
- Small simple fixes
- Rendered constructions parallel on lines 382,383
- Line 404, do we need "in multiple steps"?
- Lines 394 & 397, removed the period becuase the other key definitions don't have them

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -29,7 +29,7 @@ See https://github.com/mawww/golf for Kakoune solutions to vimgolf challenges,
 regularly beating the best Vim solution.
 
 See the link:doc/design.asciidoc[design document] for more information on
-Kakoune philosophy and design.
+Kakoune's philosophy and design.
 
 :numbered:
 
@@ -37,16 +37,15 @@ Introduction
 ------------
 
 Kakoune is a code editor that implements Vi's "keystrokes as a text editing
-language" model. As it's also a modal editor, it is somewhat similar to the
+language" model. As it is also a modal editor, it is somewhat similar to the
 Vim editor (after which Kakoune was originally inspired).
 
-Kakoune can operate in two modes, normal and insertion. In insertion mode,
+Kakoune can operate in two modes: *normal* and *insertion*. In insertion mode,
 keys are directly inserted into the current buffer. In normal mode, keys
 are used to manipulate the current selection and to enter insertion mode.
 
-Kakoune has a strong focus on interactivity, most commands provide immediate
-and incremental results, while still being competitive (as in keystroke count)
-with Vim.
+Kakoune has a strong focus on interactivity. Most commands provide immediate
+and incremental results, while being competitive with Vim in terms of keystroke count.
 
 Kakoune works on selections, which are oriented, inclusive ranges of characters.
 Selections have an anchor and a cursor. Most commands move both of
@@ -101,7 +100,7 @@ Getting started
 Building
 ~~~~~~~~
 
-Kakoune dependencies are:
+Kakoune's dependencies are:
 
  * A {cpp}20 compliant compiler (GCC >= 10.3 or clang >= 11) along with its
    associated {cpp} standard library (libstdc{pp} >= 10 or libc{pp})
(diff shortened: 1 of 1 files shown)


README.asciidoc
```diff
@@ -29,24 +29,23 @@ See https://github.com/mawww/golf for Kakoune solutions to vimgolf challenges,
...
-Kakoune can operate in two modes, normal and insertion. In insertion mode,
+Kakoune can operate in two modes: *normal* and *insertion*. In insertion mode,
 keys are directly inserted into the current buffer. In normal mode, keys
 are used to manipulate the current selection and to enter insertion mode.
 
-Kakoune has a strong focus on interactivity, most commands provide immediate
-and incremental results, while still being competitive (as in keystroke count)
-with Vim.
+Kakoune has a strong focus on interactivity. Most commands provide immediate
+and incremental results, while having nearly the same keystroke count as vim.
```
krobelus, line 48: I think the "competitive" was pretty strong, we should probably keep it.
(also Vim over vim is consistent with how we spell Kakoune).

README.asciidoc
```diff
@@ -393,16 +392,16 @@ selection's cursor. Some additional keys are recognised in insert mode:
...
- * `<c-u>`: commit changes up to now as a single undo group.
+ * `<c-u>`: commit changes up to now as a single undo group
 
  * `<a-;>`: escape to normal mode for a single command
 
 Movement
 ~~~~~~~~
 
-See <<Appending>> below for instructions on extending (appending to) the current selection, in order to select more text in multiple steps.
+See <<Appending>> below for instructions on extending (appending to) the current selection in order to select more text.
```
krobelus, line 404: "in multiple steps" could also be "incrementally" but it also seems fine to drop it

README.asciidoc
Pound-Hash: Okay, changes made in new commit.

README.asciidoc
Pound-Hash: Unless someone wants it otherwise, I'm inclined to dropping it.

README.asciidoc
krobelus: `git commit --amend` is fine but it looks like you didn't push an actual content change but only a redundant merge commit.
When you push a rewritten history you need `git push --force`.
If you don't use `--force`, git will tell you to pull (the old version) first.
That's actually misleading advice here, if you follow it, git will create a merge commit that combines the local and the remote branches.
For this simple change we don't need merge commits, a single commit with all changes is enough. A quick `git rebase -i` drops merge commits by default.

Pound-Hash: Okay! Done. How's it look?

I still want to amend the rest of the README. I sent this PR early because I wasn't sure how the response would be. I could amend this PR or I can make a new one for the second half of the README.

krobelus: Thanks, the changes look good to me (except for one thing I'll comment on).

> I could amend this PR or I can make a new one for the second half of the README.

If you'll add similar changes, then amending is a good idea because it makes the next reviewer more efficient.

README.asciidoc
```diff
@@ -339,13 +338,13 @@ directory. It will first look for an `autoload` directory at
...
-`${runtime}/kakrc.local` which is expected to contain distribution provided
+`${runtime}/kakrc.local`, which is expected to contain distribution provided
 configuration.
 
-And finally, the user configuration will be loaded from `${userconf}/kakrc`.
+Finally, the user configuration will load from `${userconf}/kakrc`.
 
 NOTE: If you create a user `autoload` directory in `${userconf}/autoload`,
-the system one at `${runtime}/autoload` will not be loaded anymore. You can
+the system one at `${runtime}/autoload` will not load anymore. You can
```
krobelus, line 347: Not sure about "loads" vs. "is loaded" (same above).
I think of the editor as the subject ("the editor loads the configuration file") that's why passive voice makes more sense to me.

Pound-Hash: I had thought that as well. There's a terrible tendency for people to use the passive in contexts that don't warrant it. However, in this case, I think you're right. I'll change it back.

PR #4592, opened 2022-04-14
markdown: Add inline typescript highlighting with ```ts construction

sidkshatriya:
Currently only ` ```typescript ` gets inline highlighted when viewing typescript code in a markdown file in Kakoune.

GitHub for instance accepts both ` ````ts ` and ` ````typescript `

For usage in the wild, see https://github.com/denoland/deno/blob/main/README.md which uses ` ```ts ` . This will fail to inline highlight in Kakoune without this PR.

### DIFF
--- a/rc/filetype/markdown.kak
+++ b/rc/filetype/markdown.kak
@@ -45,12 +45,16 @@ evaluate-commands %sh{
     awk c cabal clojure coffee cpp crystal css cucumber d diff dockerfile elixir erlang fish
     gas go haml haskell html ini java javascript json julia kak kickstart
     latex lisp lua makefile markdown moon objc ocaml perl pug python ragel
-    ruby rust sass scala scss sh swift toml tupfile typescript yaml sql
+    ruby rust sass scala scss sh swift toml ts tupfile typescript yaml sql
   "
   for lang in ${languages}; do
     printf 'add-highlighter shared/markdown/%s region -match-capture ^(\h*)```\h*(%s\\b|\\{[.=]?%s\\})   ^(\h*)``` regions\n' "${lang}" "${lang}" "${lang}"
     printf 'add-highlighter shared/markdown/%s/ default-region fill meta\n' "${lang}"
-    [ "${lang}" = kak ] && ref=kakrc || ref="${lang}"
+    case "${lang}" in
+      kak) ref="kakrc" ;;
+      ts)  ref="typescript" ;;
+      *)   ref="${lang}" ;;
+    esac
     printf 'add-highlighter shared/markdown/%s/inner region \A```[^\\n]*\K (?=```) ref %s\n' "${lang}" "${ref}"
   done
 }
--- a/rc/filetype/markdown.kak
+++ b/rc/filetype/markdown.kak
@@ -45,13 +45,24 @@ evaluate-commands %sh{
     awk c cabal clojure coffee cpp crystal css cucumber d diff dockerfile elixir erlang fish
     gas go haml haskell html ini java javascript json julia kak kickstart
     latex lisp lua makefile markdown moon objc ocaml perl pug python ragel
-    ruby rust sass scala scss sh swift toml ts tupfile typescript yaml sql
+    ruby rust sass scala scheme scss sh swift toml ts tupfile typescript yaml sql
   "
   for lang in ${languages}; do
     printf 'add-highlighter shared/markdown/%s region -match-capture ^(\h*)```\h*(%s\\b|\\{[.=]?%s\\})   ^(\h*)``` regions\n' "${lang}" "${lang}" "${lang}"
     printf 'add-highlighter shared/markdown/%s/ default-region fill meta\n' "${lang}"
     case "${lang}" in
+      clj) ref="clojure" ;;
+      js)  ref="javascript" ;;
+      cr)  ref="crystal" ;;
+      c++) ref="cpp" ;;
+      ex)  ref="elixir" ;;
+      exs) ref="elixir" ;;
       kak) ref="kakrc" ;;
+      pl)  ref="perl" ;;
+      py)  ref="python" ;;
+      py3) ref="python" ;;
+      python3) ref="python" ;;
+      scm) ref="scheme" ;;
       ts)  ref="typescript" ;;
       *)   ref="${lang}" ;;
     esac


rc/filetype/markdown.kak
```diff
@@ -45,12 +45,12 @@ evaluate-commands %sh{
...
     gas go haml haskell html ini java javascript json julia kak kickstart
     latex lisp lua makefile markdown moon objc ocaml perl pug python ragel
-    ruby rust sass scala scss sh swift toml tupfile typescript yaml sql
+    ruby rust sass scala scss sh swift toml ts tupfile typescript yaml sql
   "
   for lang in ${languages}; do
     printf 'add-highlighter shared/markdown/%s region -match-capture ^(\h*)```\h*(%s\\b|\\{[.=]?%s\\})   ^(\h*)``` regions\n' "${lang}" "${lang}" "${lang}"
     printf 'add-highlighter shared/markdown/%s/ default-region fill meta\n' "${lang}"
-    [ "${lang}" = kak ] && ref=kakrc || ref="${lang}"
+    [ "${lang}" = kak ] && ref=kakrc || [ "${lang}" = ts ] && ref=typescript || ref="${lang}"
```
krobelus, line 53: Unfortunately this will cause unexpected results if `$lang = kak` (it will set ref=typescript). Consider

	$ true && echo kakrc || false && echo typescript || echo fallthrough
	kakrc
	typescript

Looks like the jobs separated by && and || are evaluated one-by-one, left-to-right.
How about using a `case` statement instead?

rc/filetype/markdown.kak
```diff
@@ -45,12 +45,12 @@ evaluate-commands %sh{
     awk c cabal clojure coffee cpp crystal css cucumber d diff dockerfile elixir erlang fish
     gas go haml haskell html ini java javascript json julia kak kickstart
     latex lisp lua makefile markdown moon objc ocaml perl pug python ragel
-    ruby rust sass scala scss sh swift toml tupfile typescript yaml sql
+    ruby rust sass scala scss sh swift toml ts tupfile typescript yaml sql
```
krobelus, line 48: There are a bunch of other aliases listed by `pygmentize -L`, like "js" for "javascript" or the various flavours of "sh".
Probably we should add the ones that are used in practice. No strong opinion here.

rc/filetype/markdown.kak
sidkshatriya: Added a case statement -- Done!

rc/filetype/markdown.kak
sidkshatriya: The `case` statement now exists and the aliases can be easily added from now. I set out to fix the typescript case and now that is done. 

Thanks for the `pygmentize -L` tip. Let me review and see if we can get some easy wins like js for javascript and so on.

sidkshatriya: @krobelus I did a quick pass and added some common aliases. Also, `scheme` now has inline highlighting -- it was simply not listed earlier.

I debated added the `bash` => `sh` mapping but strictly speaking `bash` is a super-set of `sh` so its not exactly an alias. There are similar concerns with zsh, ksh etc. 

Anyways, if there are some more aliases I think this can be done as part of another PR. I think this is a good start for now...

rc/filetype/markdown.kak
```diff
@@ -45,12 +45,27 @@ evaluate-commands %sh{
...
+      js)  ref="javascript" ;;
+      cr)  ref="crystal" ;;
+      c++) ref="cpp" ;;
+      ex)  ref="elixir" ;;
+      exs) ref="elixir" ;;
+      kak) ref="kakrc" ;;
+      pl)  ref="perl" ;;
+      py)  ref="python" ;;
+      py3) ref="python" ;;
+      python3) ref="python" ;;
```
krobelus, line 64: For the record, I feel that using these short aliases for markdown code blocks is pretty poor taste.
I guess the ones that match the file extension make sense on some level.

rc/filetype/markdown.kak
```diff
@@ -45,12 +45,27 @@ evaluate-commands %sh{
...
+    ruby rust sass scala scheme scss sh swift toml ts tupfile typescript yaml sql
   "
   for lang in ${languages}; do
     printf 'add-highlighter shared/markdown/%s region -match-capture ^(\h*)```\h*(%s\\b|\\{[.=]?%s\\})   ^(\h*)``` regions\n' "${lang}" "${lang}" "${lang}"
     printf 'add-highlighter shared/markdown/%s/ default-region fill meta\n' "${lang}"
-    [ "${lang}" = kak ] && ref=kakrc || ref="${lang}"
+    case "${lang}" in
+      clj) ref="clojure" ;;
+      js)  ref="javascript" ;;
+      cr)  ref="crystal" ;;
```
krobelus, line 56: I think you forgot to add the new aliases to the list above, so this is dead code.
Actually, it seems ugly and a bit wasteful to add a separate highlighter for each language alias.
How about adding just one highlighter for each supported language:

```kak
for lang in ${languages}; do
  case "${lang}" in
      (clojure) aliases='(clojure|clj)' ;;
      (javascript) aliases='(javascript|js)' ;;
      # ...
      *) aliases=${lang}
  esac
  printf 'add-highlighter shared/markdown/%s region -match-capture ^(\h*)```\h*(%s\\b|\\{[.=]?%s\\})   ^(\h*)``` regions\n' "${lang}" "${aliases}" "${aliases}"
  printf 'add-highlighter shared/markdown/%s/ default-region fill meta\n' "${lang}"
  printf 'add-highlighter shared/markdown/%s/inner region \A```[^\\n]*\K (?=```) ref %s\n' "${lang}" "${lang}"
```

rc/filetype/markdown.kak
krobelus: > I did a quick pass and added some common aliases.

Thanks :)

> I debated added the `bash` => `sh` mapping but strictly speaking `bash` is a super-set of `sh` so its not exactly an alias. There are similar concerns with zsh, ksh etc.

IMO we should definitely support the "shell" alias and probably all others too. We also use filetype=sh for all these shells.

rc/filetype/markdown.kak
sidkshatriya: Yes, this is a cleaner approach indeed. Will explore and check that everything will work as expected.

sidkshatriya:
Sorry I don't have the cycles to work on the new, principled approach. I just wanted to fix this for typescript but it does make sense to fix this for more languages in one go.

Closing for now, if I get the chance I may re-submit a new PR later. Anyone can submit a fresh PR if they wish in the meanwhile.

Issue #3670, opened 2020-08-20
Unable to remove hooks with no group


alexherbo2:
It could be nice to auto-generate a group, so it can be completed with `remove-hooks`.

mawww:
Yeah, I wonder what the best approach would be, hook groups are a bit hacky, maybe we could improve that with a hierarchical structure, similar to how highlighters are specified.

PR #4594, opened 2022-04-21
rc file-detection hook group

throwawayaccount12345-1:
Add a group to the `file-detection` hooks.

There's no way to remove hooks without a group. With this patch, you'll be able to remove those
`file-detection` hooks manually. There's no need for two separate groups since if you wanted to
remove only one, you could run `remove-hooks` and then only add one again.

Related: #3670

### DIFF
--- a/rc/detection/file.kak
+++ b/rc/detection/file.kak
@@ -22,5 +22,5 @@ define-command -hidden file-detection %{ evaluate-commands %sh{
     fi
 } }
 
-hook global BufOpenFile .* file-detection
-hook global BufWritePost .* file-detection
+hook -group file-detection global BufOpenFile .* file-detection
+hook -group file-detection global BufWritePost .* file-detection


rc/detection/file.kak
```diff
@@ -22,5 +22,5 @@ define-command -hidden file-detection %{ evaluate-commands %sh{
     fi
 } }
 
-hook global BufOpenFile .* file-detection
-hook global BufWritePost .* file-detection
+hook -group file-detection global BufOpenFile .* file-detection
+hook -group file-detection global BufWritePost .* file-detection
```
krobelus, line 26: Thanks for this patch. I sometimes edit plaintext files that are wrongly classified as C++ or Ruby by file(1). It'll be convenient to work around this via disabled_hooks.

`file-detection` is a good name IMO. Looking at our other hooks, alternatives are `file-hooks` or `file` but both sound pretty generic

PR #4596, opened 2022-04-23
Golang filetype handling improvements

ttttcrngyblflpp:
When indenting on newline in Go files, only remove trailing whitespace on the previous line and copy indentation of the previous line if in comment context.


rc/filetype/go.kak
```diff
@@ -101,16 +101,24 @@ define-command -hidden go-indent-on-new-line %~
...
-        # indent after a switch's case/default statements
-        try %[ execute-keys -draft k<a-x> <a-k> ^\h*(case|default).*:$ <ret> j<a-gt> ]
-        # deindent closing brace(s) when after cursor
-        try %[ execute-keys -draft <a-x> <a-k> ^\h*[})] <ret> gh / [})] <ret> m <a-S> 1<a-&> ]
+        try %{
+            try %{ # line comment
+                execute-keys -draft k<a-x> s ^\h*// <ret>
+            } catch %{ # block comment
+                execute-keys -draft <a-?> /\* <ret> <a-K>\*/<ret>
+            }
```
krobelus, line 111: Why? This breaks an important feature. Given that `%()` is the cursor

```go
package main

func main() {
	if true {%()
	}
}
```

Today if I press o I get

```go
	if true {
		%()
	}
```

with your change I get

```go
	if true {
	%()
	}
```

which seems much worse.

ttttcrngyblflpp: That's not what I'm observing locally. I even tried running `XDG_CONFIG_HOME=blah kak` to not source my personal kak configuration and your example still works fine for me.

krobelus: Sorry my bad, I used tabs to indent but had indentwidth=4 so two indents look like one.

Looks good to me then, thanks! I guess an example would have helped. Starting from

```go
func main() {
    // comment {  %()
}
```

old behavior of `o`:

```go
func main() {
    // comment {
        //
    }
}
```

new behavior of `o`:

```go
func main() {
    // comment {
    //
    }
}
```

We could copy this behavior over to other languages but that can be a follow-up.

This example shows that we should also disable the insertion of } from the go-insert-on-new-line hook.
(I have disabled all .*-insert hooks because I think they make things less predictable.)

ttttcrngyblflpp: Glad we hashed out the indent width thing.

I will include examples in future in the PR description. The behavior that drove me to put this up is:

```
// This is a comment (stuff inside parentheses %()
```

Hitting newline yields:

```
// This is a comment (stuff inside parentheses %()
                      //
```

The above example also shows the "align to open paren" behavior which is not how `gofmt` formats code and should probably be changed as well (argument could be made to make it configurable but almost certainly shouldn't be default). I had a go but kept getting got by edge cases. Will address in a follow-up change.

+1 to disabling the auto-insertion stuff for Go. I've had a lot of issues with it.

I will probably follow up by applying this to some other languages as well (IIRC kakscript and c-family are the languages I care about that also have this problem). There are a lot of languages that I don't write so I wouldn't necessarily feel comfortable applying the fix, however.

krobelus: Right, the "align to open paren" was probably copied over from some other language (C?), we should probably remove it.

ttttcrngyblflpp: I pushed a commit to remove aligning to open paren and brace; and add a layer of indentation instead.

I also pushed a commit to add a hook group for inserting closing paren and brace, because it doesn't work reliably for me and I don't know of a way to turn it off without also turning off inserting the comment characters `//`. If there is please let me know and I can remove that commit.

PR #4611, opened 2022-05-09
OCaml: Automatically insert closing comment when user starts a comment

sidkshatriya:
An OCaml comment is `(* Some comment *)`. Like the C-family this can be
a multiline comment.

Recognize when the user is trying to commence a comment when they type `(*` and
then automatically insert `*)` on behalf of the user. A small convenience.

### DIFF
--- a/rc/filetype/ocaml.kak
+++ b/rc/filetype/ocaml.kak
@@ -97,3 +97,16 @@ define-command ocaml-alternative-file -docstring 'Switch between .ml and .mli fi
 }
 
 }
+
+# The OCaml comment is `(* Some comment *)`. Like the C-family this can be a multiline comment.
+#
+# Recognize when the user is trying to commence a comment when they type `(*` and
+# then automatically insert `*)` on behalf of the user. A small convenience.
+hook global WinSetOption filetype=ocaml %{
+    hook window InsertChar '\*' %{
+        try %{
+            execute-keys -draft 'HH<a-k>\(\*<ret>'
+            execute-keys '  *)<left><left><left>'
+        }
+    }
+}
--- a/rc/filetype/ocaml.kak
+++ b/rc/filetype/ocaml.kak
@@ -32,7 +32,7 @@ provide-module ocaml %{
 
 add-highlighter shared/ocaml regions
 add-highlighter shared/ocaml/code default-region group
-add-highlighter shared/ocaml/string region (?<!')" (?<!\\)(\\\\)*" fill string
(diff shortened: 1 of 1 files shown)


rc/filetype/ocaml.kak
```diff
@@ -97,3 +97,19 @@ define-command ocaml-alternative-file -docstring 'Switch between .ml and .mli fi
...
+# then automatically insert `*)` on behalf of the user. A small convenience.
+hook global WinSetOption filetype=ocaml %{
+    hook window InsertChar '\(' %{
+        hook -once -group ocaml-easy-comment window InsertChar '\*' %{
+            execute-keys ' *)<left><left>'
+            remove-hooks window ocaml-easy-comment
+        }
+        hook -once -group ocaml-easy-comment window InsertChar '[^\*]' %{
+            remove-hooks window ocaml-easy-comment
+        }
```
mawww, line 113: This logic feels a bit convoluted compared to hooking on inserting '*' and checking if the previous character is a '('. They have different corner cases (your version breaks if the cursor moves between the two inserts, mine does not care if '(' and '*' are inserted in sequence) but it seems simpler to have a single hook than a hook that adds a hook that removes itself.

sidkshatriya: Do you mean something like this?

```kak
hook global WinSetOption filetype=ocaml %{
    hook window InsertChar '(?<=\()\*' %{
        execute-keys ' *)<left><left>'
    }
}
```

I actually tried this before -- for some reason this does not work for me.

sidkshatriya: @mawww gentle ping ^^

mawww: InsertChar does not give you anything more for the filter string than the inserted char, but you can test with the `<a-k>` command, something like (off the top of my head)

```kak
hook global WinSetOption filetype=ocaml %{
    hook window InsertChar '\*' %{ try %{ 
        execute-keys -draft 'h<a-k>\(<ret>'
        execute-keys ' *)<left><left>' }
    }
}
```

sidkshatriya: @mawww Thanks -- This overall approach works, with a bit of a modification.

I've force pushed a single commit and given you `Co-Author` status on the commit. Please merge if this looks OK now!

sidkshatriya:
Also, while we're on OCaml make a small fix in the highlighting code. (See commit just pushed)

Issue #3743, opened 2020-09-17
Add wayland.kak


alexherbo2:

Hardy7cc:
Hi, for the record at the moment I copied the x11-terminal command from here: 
[https://github.com/mawww/kakoune/blob/master/rc/windowing/x11.kak](url)
Within sway this works fine for generating new terminal instances. However I do not use the focus thing and I'm honest I'm not sure if it is that easy to implement some kind of focus switch.

alexherbo2:
We can add commands such as `sway-focus` and alias it to `focus` when in [Sway].

[Sway]: https://swaywm.org

PR #4631, opened 2022-05-26
Support focus in sway

AnthonyYoManz:
While Wayland offers nothing general to help us support `focus` on all
window managers, WM-specific implementations are generally possible.
Sway is a tiling window manager that mimics i3, and has a reasonably
powerful CLI that can help us achieve this.

In addition to supporting `focus` for Sway, this change paves the way
for additional WM-specific Wayland functionality by adding a detection
step to wayland.kak, in a similar fashion to detection.kak.

Somewhat relevant to https://github.com/mawww/kakoune/issues/3743, 
although the headline for that issue is already resolved.

### DIFF
--- a/rc/windowing/sway.kak
+++ b/rc/windowing/sway.kak
@@ -1,0 +1,51 @@
+provide-module sway %{
+
+# Ensure we're actually in Sway
+evaluate-commands %sh{
+    [ -z "${kak_opt_windowing_modules}" ] || 
+    [ -n "$SWAYSOCK" ] ||
+    echo 'fail SWAYSOCK is not set'
+}
+
+require-module 'wayland'
+
+define-command sway-focus-pid -hidden %{
+    evaluate-commands %sh{
+        pid=$kak_client_pid
+
+        # Try to focus a window with the current PID, walking up the tree of 
+        # parent processes until the focus eventually succeeds
+        while ! swaymsg [pid=$pid] focus > /dev/null 2> /dev/null ; do
+            # Replace the current PID with its parent PID
+            pid=$(ps -p $pid -o ppid=)
+
+            # If we couldn't get a PPID for some reason, or it's 1 or less, we 
+            # should just fail. 
+            if [ -z $pid ] || [ $pid -le 1 ]; then
+                echo "fail Can't find PID for Sway window to focus"
+                break
+            fi
+        done
+    }
+}
+
+define-command sway-focus -params ..1 -docstring '
+sway-focus [<kakoune_client>]: focus a given client''s window.
+If no client is passed, then the current client is used' \
+%{
+    # Quick branch to make sure we're calling sway-focus-pid from the client 
+    # the user wants to focus on.
+    evaluate-commands %sh{
+        if [ $# -eq 1 ]; then
+            printf "evaluate-commands -client '%s' sway-focus-pid" "$1"
+        else
+            echo sway-focus-pid
+        fi
+    }
+}
+complete-command sway-focus client
+
+unalias global focus
+alias global focus sway-focus
+
+}
--- a/rc/windowing/wayland.kak
+++ b/rc/windowing/wayland.kak
@@ -49,7 +49,7 @@ define-command wayland-focus -params ..1 -docstring '
 wayland-focus [<kakoune_client>]: focus a given client''s window
 If no client is passed, then the current client is used' \
 %{
-    fail There is no way to focus another window on Wayland
+    fail 'Focusing specific windows in most Wayland window managers is unsupported'
 }
 complete-command wayland-focus client
 
--- a/rc/windowing/detection.kak
+++ b/rc/windowing/detection.kak
@@ -23,7 +23,7 @@ declare-option -docstring \
 "Ordered list of windowing modules to try and load. An empty list disables
 both automatic module loading and environment detection, enabling complete
 manual control of the module loading." \
-str-list windowing_modules 'tmux' 'screen' 'kitty' 'iterm' 'wayland' 'x11'
+str-list windowing_modules 'tmux' 'screen' 'kitty' 'iterm' 'sway' 'wayland' 'x11'
 
 hook -group windowing global KakBegin .* %{
 


AnthonyYoManz:
I should mention probably the major downside of this PR is it introduces dependence on `jq` for parsing the JSON that `swaymsg -t get_tree` spits out. This only applies to Sway users though, and I'd argue they're prepared to run a quick install command should they see an error about it in \*debug\*.

rc/windowing/wayland.kak
```diff
@@ -45,15 +45,25 @@ The program passed as argument will be executed in the new terminal' \
...
-wayland-focus [<kakoune_client>]: focus a given client''s window
-If no client is passed, then the current client is used' \
-%{
-    fail There is no way to focus another window on Wayland
+
+# Detect specific window manager functionality, similarly to how detection.kak
+# peruses different environments.
+declare-option -docstring '
+Ordered list of Wayland window manager modules to try and load.' \
+str-list wayland_wm_modules 'wayland-sway' 'wayland-unsupported'
```
mawww, line 53: I am not really keen on that nested approach to wayland wm, I think a simpler alternative would be to just introduce a `sway.kak` windowing module, put it before `wayland.kak` in the checking order, and have it `require-module wayland` if detection succeeds (then overwrite the focus alias with its own implementation).

mawww (changes requested):
This is a really nice PR, I'd just like to amend the approach slightly as described in the other comment.

AnthonyYoManz:
Woo, fast review :smile: 
I've made the changes you suggested. Wasn't sure if you'd be squashing things in this PR since it has the copyright waiver stuff, so I've reordered amended my commits instead of making a new one.

rc/windowing/sway.kak
```diff
@@ -0,0 +1,67 @@
+provide-module sway %{
+
+# Ensure we're actually in Sway
+evaluate-commands %sh{
+    [ -z "${kak_opt_windowing_modules}" ] || 
+    [ "$DESKTOP_SESSION" = "sway" ] || 
+    [ "$XDG_SESSION_DESKTOP" = "sway" ] ||
```
krobelus, line 7: I'm running sway version 1.7 and neither of those `*DESKTOP*` variables is set.
How about using `[ -n "$SWAYSOCK" ]` instead?

rc/windowing/sway.kak
```diff
@@ -0,0 +1,67 @@
...
+                # If we couldn't get a PPID for some reason, or it's 1 or 
+                # less, we should just fail. Otherwise, we call step into 
+                # our parent.
+                if [ ! -z $ppid ] && [ $ppid -gt 1 ]; then
+                    focus_window_ancestor $ppid
+                else
+                    echo "fail Can't find ID for Sway window to focus"
+                fi
+            else
+                # The bit where we actaully tell Sway to focus on a window
```
krobelus, line 38: typo: actaully

rc/windowing/sway.kak
```diff
@@ -0,0 +1,67 @@
...
+        # that's running bash that's running kak.
+
+        focus_window_ancestor() {
+            # Get the container ID from the tree of windows
+            con_id=`swaymsg -t get_tree | jq ".. | (.nodes? // empty)[] | select(.pid == $1) | .id"`
+
+            if [ -z "$con_id" ]; then
+                # If we couldn't find a matching window, then we need to do
+                # the same check for our parent, so we grab our PPID
+                ppid=`ps -p $1 -o ppid=`
```
krobelus, line 27: Nit: I'd  use `$()` instead of backticks for consistency with other support scripts (same above)

rc/windowing/sway.kak
```diff
@@ -0,0 +1,66 @@
...
+define-command sway-focus-pid -hidden %{
+    evaluate-commands %sh{
+        # We need to walk our way up a tree of process IDs to find out which
+        # window a specific client is associated with. This can be further up
+        # than you might expect, as a window could be for a terminal emulator 
+        # that's running bash that's running kak.
+
+        focus_window_ancestor() {
+            # Get the container ID from the tree of windows
+            con_id=`swaymsg -t get_tree | jq ".. | (.nodes? // empty)[] | select(.pid == $1) | .id"`
```
mawww, line 21: I think we can get rid of the jq dependency pretty easily with something like

```
if swaymsg -t get_tree | grep "\"pid\": $1"; then
    swaymsg [pid=$1] focus
else
   recursion to parent pid
fi
```

mawww: Or even better:

```
pid=$1
while ! swaymsg [pid=$pid] focus; do
    pid=get parent pid, exit if invalid
done
```

AnthonyYoManz: That's much better, thanks!

rc/windowing/sway.kak
```diff
@@ -0,0 +1,51 @@
...
+
+require-module 'wayland'
+
+define-command sway-focus-pid -hidden %{
+    evaluate-commands %sh{
+        pid=$kak_client_pid
+
+        # Try to focus a window with the current PID, walking up the tree of 
+        # parent processes until the focus eventually succeeds
+        while ! swaymsg [pid=$pid] focus > /dev/null ; do
```
sidkshatriya, line 18: If `swaymsg` sends an error to stderr it will appear in kakoune's `*debug*` buffer. So as you traverse the hierarchy looking for the correct pid you might see a lot of these messages generated by `swaymsg` which are not really errors but rather us hunting for the correct pid. So, should we have something like:

```
 while ! swaymsg [pid=$pid] focus > /dev/null 2> /dev/null; do
...
```

here instead?

rc/windowing/sway.kak
```diff
@@ -0,0 +1,51 @@
...
+%{
+    # Quick branch to make sure we're calling sway-focus-pid from the client 
+    # the user wants to focus on.
+    evaluate-commands %sh{
+        if [ $# -eq 1 ]; then
+            printf "evaluate-commands -client '%s' sway-focus-pid" "$1"
+        else
+            echo sway-focus-pid
+        fi
+    }
```
krobelus, line 44: totally not important but you can replace the shell block with

```kak
evaluate-commands -try-client %arg{1} sway-focus-pid
```

PR #4633, opened 2022-05-29
make clean: also clean up generated version files

### LINKED ISSUE #4619: Cannot compile different OS’s due to `.version.*` not cleaned up
If I compile Kakoune on Linux, and switch on macOS with the same local repository, I cannot compile Kakoune—even after `make clean`—due to `.version.cc` and `.version.o` not being cleaned up.

krobelus:
We also discussed using "git clean -dXf" but that could remove files
that were not generated by make.

Closes #4619

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -128,7 +128,7 @@ tags:
 	ctags -R
 
 clean:
-	rm -f $(objects) $(deps)
+	rm -f $(objects) $(deps) .version.cc .version.o
 
 dist:
 	@if ! [ -d ../.git ]; then echo "make dist can only run from a git repo";  false; fi


src/Makefile
```diff
@@ -128,7 +128,7 @@ tags:
 	ctags -R
 
 clean:
-	rm -f $(objects) $(deps)
+	rm -f $(objects) $(deps) .version.cc .version.o
```
mawww, line 131: I am not sure we want to remove .version.cc as this is distributed in releases and wont get regenerated out of a git repository, I think we only want to remove .version.o

mawww, line 131: I am not sure we want to remove .version.cc as this is distributed in releases and wont get regenerated out of a git repository, I think we only want to remove .version.o

krobelus: > I am not sure we want to remove .version.cc as this is distributed in releases

No, `.version` is added to the release tarball, but `.version.cc` isn't. It can always be generated from `.version` (if available) or `git decribe`.
My main motivation is to make `bear -- make clean all debug=yes` behave as expected, so I don't care so much about `.version.cc`, we can also keep it

mawww: Oh, right, my bad I misremembered how we managed version.cc then this commit is correct, no need to change anything.

PR #4637, opened 2022-06-09
rc: add hare language module

c7skasku:
I've had this for a while in https://git.sr.ht/~c7s/hare.kak, since before the language was public.

A few other changes by other people exist but aren't in my tree, so I'm making this PR so we can coordinate that in public, and so that people can comment on any other useful integrations beyond the rc/filetype module.

Currently this is missing:
* ~~good whitespace trimming~~ (done)
* ~~autoindentation~~ (done)
* ~~integration with any rc/tools~~ (done?)

### DIFF
--- a/rc/filetype/hare.kak
+++ b/rc/filetype/hare.kak
@@ -9,6 +9,8 @@ hook global WinSetOption filetype=hare %{
     hook window ModeChange pop:insert:.* -group hare-trim-indent hare-trim-indent
     hook window InsertChar \n -group hare-indent hare-indent-on-new-line
     hook window InsertChar \n -group hare-insert hare-insert-on-new-line
+    hook window InsertChar \{ -group hare-indent hare-indent-on-opening-curly-brace
+    hook window InsertChar \} -group hare-indent hare-indent-on-closing-curly-brace
 }
 
 hook -group hare-highlight global WinSetOption filetype=hare %{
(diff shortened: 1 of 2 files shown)


raiguard:
Thanks for tackling this! I have some feedback on the highlighting:

- Functions and function calls should be highlighted as `function`
- `export` and `use` should be highlighted as `keyword`
- `::`, `<`, `=`, `!`, `?`, etc should be highlighted as `operator` (arguably `,` and `;` as well, but those I'm not 100% on)

image

Here's some Lua code for comparison:

image

raiguard:
Ah, my bad, I didn't realize that those were highlighted as `meta` because I made meta white in my colorscheme for some reason. I should have tested against the default colorscheme. I changed it to match `keyword` and it gave me the results I wanted.

Function calls being highlighted is how highlighting behaves in every editor I have ever used. Of the languages that I use in kak, both the Rust and Lua modules highlight them consistently. Checking against vscode, every language I have tried has them highlighted. IMO all language highlighters should highlight function calls, but that is out of scope for this issue.

sidkshatriya:
Thanks for this PR!

> Not sure about function calls being highlighted.

I'm also in favor of function calls being highlighted because they enhance readability of the code. When you look at some code, highlighted functions give you an easy overview of what subroutines a function is calling without having to AST parse source with your brain implicitly. This is an easy win and we shouldn't worry about consistency if some kak languages don't have this (though many probably do).

Issue #4495, opened 2022-01-10
[BUG] 


RehMaar:
### Version of Kakoune

v2021.11.08

### Reproducer

Save the following as a `.hs` file, open in Kakoune:
```haskell
--
main = undefined
```

### Outcome

The second line is highlighted as a comment.
![2022-01-10-205842_412x85_scrot](https://user-images.githubusercontent.com/11967506/148815318-bad2cb83-6e57-4b41-8dfe-5cae9c201a6c.png)

### Expectations

The second line is highlighted as code.

Delapouite:
I confirm the highlighting is now correct 

<img width="206" height="56" alt="Image" src="https://github.com/user-attachments/assets/5b546500-3c1d-4a4c-ab8c-318a47abd242" />

This issue can be closed

PR #4641, opened 2022-06-15
Fix syntax hightlighting issue with comments

astaugaard:
as mentioned in #4495
I only tested it a tiny bit.


astaugaard:
ok fixed formating

PR #4644, opened 2022-06-17
git-show-branch added to show graph with branches and commits

nonumeros:


rc/tools/git.kak
```diff
@@ -81,6 +90,7 @@ define-command -params 1.. \
         case "$1" in
            diff) map_diff_goto_source=true; filetype=diff ;;
            show) map_diff_goto_source=true; filetype=git-log ;;
+           show-branch) map_diff_goto_source=true; filetype=git-show-branch ;;
```
krobelus, line 93: Looks good overall.
I'm not very familiar with `show-branch` but I don't think it shows content changes, so `map_diff_goto_source` here is redundant.

krobelus: > When you said content changes, you meant as if source and target of git-diff

exactly

I guess if users want to do `!git show <c-r>.` (after selecting a commit ID) then the mapping could be useful but I'd just run `:git show <c-r>.`

Issue #4648, opened 2022-06-24
highlighting latex3 functions variables and list


eburghar:
### Feature

[latex3](https://www.latex-project.org/latex3/) brings a lot of [conventions](https://www.alanshawn.com/tech/2020/05/25/latex-3.html) to simulate types on top of TeX macro expansion language. It is now possible to easily differentiate functions from variables in latex3 code.

### Usecase

`filetypes/latex.kak` is mainly for highlighting document not for code. basic support for latex3 syntax would be nice for packages authors.

PR #4649, opened 2022-06-24
feat: highlight latex3 functions, variables and lists

eburghar:
See #4648

### DIFF
--- a/rc/filetype/latex.kak
+++ b/rc/filetype/latex.kak
@@ -38,9 +38,9 @@ add-highlighter shared/latex/content default-region group
 add-highlighter shared/latex/cs region '(?<!\\)(?:\\\\)*\K\\[@\w]' '\n|(?<![@:\\{}\[\]*])(?![@:{}\[\]*])\b' group
 add-highlighter shared/latex/comment region '(?<!\\)(?:\\\\)*\K%' '\n' fill comment
 
-# Document and latex2e control sequence
+# Document and LaTeX2e control sequence
 add-highlighter shared/latex/cs/ regex '(?:\\[a-zA-Z@]+)' 0:keyword
-## Options passed to latex2e control sequences, between brackets
+## Options passed to LaTeX2e control sequences, between brackets
 add-highlighter shared/latex/cs/ regex '\\[a-zA-Z@]+\b\[([^\]]+)\]' 1:value
 ## Emphasized text
 add-highlighter shared/latex/cs/ regex '\\(?:emph|textit|textsl)\{([^}]+)\}' 1:default+i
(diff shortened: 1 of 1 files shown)


rc/filetype/latex.kak
```diff
@@ -35,8 +35,23 @@ add-highlighter shared/latex regions
...
+# latex3 functions (expl3 doc)
+add-highlighter shared/latex/content/ regex '\\(?:__|@@_)?[a-zA-Z@]+_\w+(:[nNpTFDwcVvxefo]+)?\b' 0:function 1:+db@type
+# latex3 variable (expl3 doc)
+add-highlighter shared/latex/content/ regex '\\([lgc]_)[a-zA-Z]+_[a-zA-Z]+\b' 0:variable 1:+db
+# latex3 type (expl3 doc)s
+add-highlighter shared/latex/content/ regex '_(bool|box|cctab|clist|coffin|dim|fp|ior|iow|int|muskip|prop|seq|skip|str|tl)' 0:+db
+ # latex3 l3kernel modules (l3kernel/doc/l3prefixes.csv)
+add-highlighter shared/latex/content/ regex '\\(alignment|alloc|ampersand|atsign|backslash|bitset|bool|box|catcode|cctab|char|chk|circumflex|clist|code|codedoc|coffin|colon|color|cs|debug|dim|document|dollar|driver|e|else|empty|etex|exp|expl|false|fi|file|flag|fp|group|hash|hbox|hcoffin|if|inf|initex|insert|int|intarray|ior|iow|job|kernel|keys|keyval|left|log|lua|luatex|mark|marks|math|max|minus|mode|msg|muskip|nan|nil|no|novalue|one|or|other|parameter|pdf|pdftex|peek|percent|pi|prg|prop|ptex|quark|recursion|ref|regex|reverse|right|scan|seq|skip|sort|space|stop|str|sys|tag|term|tex|text|tilde|tl|tmpa|tmpb|token|true|underscore|uptex|use|utex|vbox|vcoffin|xetex|zero)_' 0:+db
```
krobelus, line 49: I guess the space is not needed since we already have `\s` 

I'm not an expert in latex but this regex seems prone to false positives.
Is it meant to match something like `\begin{document}`?

Also we should probably make sure it doesn't trigger for escaped braces by prepending `(?<!\\)(?:\\\\)*\K`
I guess this applies to a bunch of other regexes in this file, like the one for macro arguments and the ones match commands like `\foo`

eburghar: Thanks for the valuable input. I've made the fixes.

In the spirit of lowering the number of false positive, I would like to restrict the module regex (alignment_, alloc_, ...) so that it applies only in the context of a latex3 function. If it is a recognized module dim it otherwise keep the style from the generic latex3 function.

Same thing for the type (_bool) inside variable regex.

Is there a simple way to do that ? Define a new region ?

eburghar: The regex for grouped list is for highlighting list you can provide to tikz macros or latex3 clist related ones `{item1,item2:option,item31/item32}` also it catches `\begin{document}` or `\usepackage{expl3,xparse}` which is ok

eburghar:
The modifications so far:

Fixes:
 - proper highlighting of escaped comments, control sequences (cs) and math mode

Features:
 - underline face for \underline{} cs
 - highlight of latex3 cs, cs arguments, variables, macros arguments and lists
 - module names, scopes and types are dimmed to help focus on macro function and variable main name

mawww:
@eburghar Could you add a copyright waiver commit on that PR as I do not have any for you yet ?

eburghar:
I added `.ins` extension to latex filetype as it is not used by other filetypes. `.ins` is `.dtx` companion which already was in the extensions list.

PR #4650, opened 2022-06-24
Fix murmurhash for big-endian architectures.

ppentchev:
Hi,

Thanks for writing and maintaining kakoune!

What do you think about this fix for the murmurhash implementation's assumption that uint32 values are stored in memory in little-endian byte order?

Thanks again, and keep up the great work!

G'luck,
Peter

### DIFF
--- a/src/hash.cc
+++ b/src/hash.cc
@@ -41,7 +41,11 @@ size_t hash_data(const char* input, size_t len)
     for (ptrdiff_t i = -nblocks; i; ++i)
     {
         uint32_t key;
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
         memcpy(&key, blocks + 4*i, 4);
+#else
+        key = blocks[4*i] + (blocks[4*i + 1] << 8) + (blocks[4*i + 2] << 16) + (blocks[4*i + 3] << 24);
+#endif
         key *= c1;
         key = rotl(key, 15);
         key *= c2;
--- a/src/hash.cc
+++ b/src/hash.cc
@@ -41,11 +41,7 @@ size_t hash_data(const char* input, size_t len)
     for (ptrdiff_t i = -nblocks; i; ++i)
     {
         uint32_t key;
-#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
-        memcpy(&key, blocks + 4*i, 4);
-#else
-        key = blocks[4*i] + (blocks[4*i + 1] << 8) + (blocks[4*i + 2] << 16) + (blocks[4*i + 3] << 24);
-#endif
+        key = (blocks[4*i + 3] << 24) | (blocks[4*i + 2] << 16) | (blocks[4*i + 1] << 8) + blocks[4*i];
         key *= c1;
         key = rotl(key, 15);
         key *= c2;


Screwtapello:
Is the hash value ever written out to disk, or sent between client and server, so it might move from a host with one endianness to a host with another? Has hash-endianness caused a problem for you that this patch fixes, or is this patch just being careful?

krobelus:
> Is the hash value ever written out to disk, or sent between client and server

I don't think so but it fixes a valid test

Screwtapello:
OK, apparently this patch doesn't fix any actual runtime behaviour of Kakoune, but Kakoune has a debug assertion that checks for the little-endian hash value:

> Fatal error: assert failed "hash_data(data, strlen(data)) == 0xf816f95b" at hash.cc:76


...as encountered [on Debian's s390x port](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1011978#5).

Rather than use conditional compilation, could we just use the endian-safe version on every platform instead of memcpy? I bet they both compile down to the same instructions anyway.

mawww:
Yeah, I am not too keen on having `#ifdefs` arounds for that, we do not care about the actual generated hash values so we could get rid of the test, but I think its good to have it just to ensure we do apply the expected murmur3 hash. Replacing that memcpy with an endianness independent version would have my preference.

ppentchev:
Thanks for weighing in! How would you feel about an inline function that uses either ifdefs or a `__builtin_constexpr()` check for the byte order constants and then either uses `__builtin_memcpy()` or `_builtin_bswap32()`? Or would that be too GCC-specific?
...of course, it could be made conditional on "is the compiler GCC" and, if not, then use the non-optimized version (read bytes and shift them to the left).

Screwtapello:
The implementation you provided for big-endian platforms seems to work just fine on little-endian platforms, so why not use it unconditionally? If there's only one code path, that makes it much less likely for big-endian platforms to break without anyone noticing.

Screwtapello:
Comparing `memcpy` versus the endian-agnostic version [in the Godbolt compiler explorer](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:5,endLineNumber:8,positionColumn:5,positionLineNumber:8,selectionStartColumn:5,selectionStartLineNumber:8,startColumn:5,startLineNumber:8),source:'%23include+%3Cstdint.h%3E%0A%23include+%3Cstring.h%3E%0A%23include+%3Cstdio.h%3E%0Aint+main()+%7B%0A++++uint8_t+asbytes%5B%5D+%3D+%7B0,+1,+2,+3%7D%3B%0A++++uint32_t+asint+%3D+0%3B%0A%0A++++memcpy(%26asint,+asbytes,+4)%3B%0A++++//asint+%3D+asbytes%5B0%5D+%2B+(asbytes%5B1%5D+%3C%3C+8)+%2B+(asbytes%5B2%5D+%3C%3C+16)+%2B+(asbytes%5B3%5D+%3C%3C+24)%3B%0A%0A++++printf(%22%25%23x%22,+asint)%3B%0A%7D%0A'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:output,i:(editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+11.2+(Compiler+%231)',t:'0')),header:(),l:'4',m:25.442477876106196,n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:g112,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'0',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-O1',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1,tree:'1'),l:'5',n:'0',o:'x86-64+gcc+11.2+(C%2B%2B,+Editor+%231,+Compiler+%231)',t:'0')),header:(),k:50,l:'4',m:74.5575221238938,n:'0',o:'',s:0,t:'0')),k:50,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4), both versions are evaluated at compile-time and the result hard-coded into the executable even at `-O1`, so I don't expect there to be any practical performance difference between the two.

src/hash.cc
```diff
@@ -41,7 +41,11 @@ size_t hash_data(const char* input, size_t len)
     for (ptrdiff_t i = -nblocks; i; ++i)
     {
         uint32_t key;
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
         memcpy(&key, blocks + 4*i, 4);
+#else
+        key = blocks[4*i] + (blocks[4*i + 1] << 8) + (blocks[4*i + 2] << 16) + (blocks[4*i + 3] << 24);
```
krobelus, line 47: I'd place the most significant bit first:

	const uint8_t *block = blocks + 4*i;
	uint32_t key = (block[3] << 24) + (block[2] << 16) + (block[1] << 8) + block[0];

mawww: Yep, always using that line would be my preference, on little endian machines this is just a 32 byte load and as @Screwtapello shown it gets compiled as such.

ppentchev:
Unfortunately I am not sure that this is necessarily true. The Compiler Explorer example showed loading a constant value; the hash function will most commonly not be called on constant input known at compile time. At least on my Debian testing machine with gcc 11.2.0 and g++ 11.2.0, the following program:

```
#include <err.h>
#include <inttypes.h>
#include <stdio.h>

int main(void) {
        char buf[4];
        if (fread(buf, 1, 4, stdin) != 4)
                err(1, "whee fread");
        uint32_t value = buf[0] + (buf[1] << 8) + (buf[2] << 16) + (buf[3] << 24);
        printf("%X\n", value);
        return 0;
}
```

...produced the following assembly output when built with `gcc -O2 -ggdb`:

```(gdb) disass main                                                                                                      
Dump of assembler code for function main:                                                                              
   0x0000000000001070 <+0>:     sub    $0x18,%rsp
   0x0000000000001074 <+4>:     mov    0x2fc5(%rip),%rcx        # 0x4040 <stdin@GLIBC_2.2.5>
   0x000000000000107b <+11>:    mov    $0x4,%edx  
   0x0000000000001080 <+16>:    mov    $0x1,%esi                                                                       
   0x0000000000001085 <+21>:    lea    0xc(%rsp),%rdi
   0x000000000000108a <+26>:    call   0x1030 <fread@plt>
   0x000000000000108f <+31>:    cmp    $0x4,%rax
   0x0000000000001093 <+35>:    jne    0x10cd <main+93>
   0x0000000000001095 <+37>:    movsbl 0xd(%rsp),%esi
   0x000000000000109a <+42>:    movsbl 0xc(%rsp),%eax
   0x000000000000109f <+47>:    lea    0xf69(%rip),%rdi        # 0x200f
   0x00000000000010a6 <+54>:    shl    $0x8,%esi
   0x00000000000010a9 <+57>:    add    %eax,%esi
   0x00000000000010ab <+59>:    movsbl 0xe(%rsp),%eax
   0x00000000000010b0 <+64>:    shl    $0x10,%eax
   0x00000000000010b3 <+67>:    add    %eax,%esi
   0x00000000000010b5 <+69>:    movsbl 0xf(%rsp),%eax
   0x00000000000010ba <+74>:    shl    $0x18,%eax
   0x00000000000010bd <+77>:    add    %eax,%esi
   0x00000000000010bf <+79>:    xor    %eax,%eax
   0x00000000000010c1 <+81>:    call   0x1040 <printf@plt>
```

I think that this is not exactly a 32-bit load there :( And compiling the same source code with `g++ -O3 -ggdb` produced pretty much the same sequence of "load, shift, add, load, shift, add, ..." instructions.

ppentchev:
So I just spotted something stupid there, but no, changing the type of buf to unsigned char did not make a difference.

krobelus:
> changing the type of buf to unsigned char did not make a difference.

it does for versions higher than 11.2.
Both [gcc 12.1 with -O2](https://godbolt.org/z/hxbnsbxbW) as well as clang 13. can optimize this.

mawww:
Using bitwise or instead of plus triggers the optimization with gcc 10.3 as well (for unsigned char)

src/hash.cc
```diff
@@ -41,7 +41,7 @@ size_t hash_data(const char* input, size_t len)
     for (ptrdiff_t i = -nblocks; i; ++i)
     {
         uint32_t key;
-        memcpy(&key, blocks + 4*i, 4);
+        key = (blocks[4*i + 3] << 24) | (blocks[4*i + 2] << 16) | (blocks[4*i + 1] << 8) + blocks[4*i];
```
mawww, line 44: Missed one `+`

PR #4652, opened 2022-06-26
Rename Go auto-insertion hooks to make it easier to disable them

krobelus:
The canonical way to disable all auto-insertion hooks is

	set-option global disabled_hooks .*-insert

A recent change allowed to disable hooks that insert ) and }
independent of hooks that insert // (a step in the right
direction, we should do it for more filetypes).

Since the new hook ("go-insert-closing-delimiter") doesn't match
.*-insert, it broke the above snippet.  Fix this by renaming it to
"go-closing-delimiter-insert".

This makes it a bit less obvious how to disable only comment insertion.
Not sure if there's interest in that, but make it easier by renaming
"go-insert" to "go-comment-insert".

cc @ttttcrngyblflpp, I missed this in the review.
I'm sure we can improve the UX further but I think this is a good intermediate step.


ttttcrngyblflpp:
Sorry about that, thanks for the fix!

Issue #1709, opened 2017-11-19
Remove sorting of user-supplied insert completions


aver-d:
Hi, for a plugin I'm writing I need insert-mode completions to be displayed in a pre-determined order.

The implementation for displaying user-generated completions in `complete_option` includes a sort

https://github.com/mawww/kakoune/blob/079b006cdab6f0ae594f80fff18ad61b71b0754d/src/insert_completer.cc#L315

As far as I'm aware, there's no other way I can show the custom-generated completions without this sorting.

Would it be possible to either make the sort optional or simply remove this one line and leave sorting to the user if it's required?

mawww:
It would be tricky to remove/disable this sort, because this is not simply an alphabetical sort, its sorting candidates according to their fuzzy matching relevance. When no text has been written by the user to filter out candidates yet, this will be alphabetically sorted as that is the last condition when comparing two ranked matches, but as soon as some text has been written, the sorting will be based on the number of matching word boundaries, whether the user entered text is a prefix or just a subsequence, wether the matched part is contiguous...

Disabling this behaviour would be possible, after all we could just display candidates that matches the entered text without sorting them at all, but it would require a very strong motivating use case to add that complexity inside the insert completion code.

Can you describe our use case in more details ?

aver-d:
Hi, and thanks for the response.

In case I wasn't clear, I didn't mean for there to be any change in the way that kakoune currently operates – only that it be useful to have a choice to display a static list of completions candidates (provided by an external process) to be shown as originally supplied.

So, for example, `set window my_completions ${header}:d||d:b||b:a||a:c||c`  could show in order `d b a c`, not `a b c d`.

You mention there are two stages: (1) an initial lexicographic sort and (2) subsequent fuzzy matching. 

I had been considering the problem in a way that had seemed to me would result in less work for kakoune. Neither stage 1 nor 2 would occur. The user is presented with a list of candidates and can either select an item or make a change to the buffer. Either way, the list is then immediately cleared, so there's no opportunity for stage 2 fuzzy matching to occur – the external process instead just sends a new list which reflects the new buffer state. I'd been able to get this to work inside a few milliseconds, so just recompiling and commenting out that sort line seemed enough to resolve what appears to be the only problem of this extra sort (Though I appreciate you're more familiar with kakoune's implementation and there may be further issues I hadn't encountered during my brief trial).

[...]

PR #4653, opened 2022-06-26
Use optimal subsequence match to rank completions

### LINKED ISSUE #3806: Fuzzy finder accuracy
The behavior of the builtin fuzzy finder puzzles me. Consider this example:

image

As you can see, the file I actually want, `common_startup.rb`, is listed dead last, even though I have entered its filename almost verbatim. As a user coming from Sublime Text, and having written my own fuzzy finder implementations in the past, I don't understand the sorting of results. I've tried to read through the source code of the fuzzy search in Kakoune, but couldn't immediately make sense of its sorting algorithm. Could anyone provide more information on how it prioritizes results, and what I should have entered here instead to get the file I wanted?

krobelus:
When using fuzzy completion with candidates from "git ls-files" in
a project with many files I often get unexpected ordering even when
one candidate is clearly the best.
This usually happens if both candidates are the same kind of match
(typically only a subsequence match), which means that the leftmost
match wins (via max_index), which is often wrong.
Also there are some other matching heuristics like FirstCharMatch
that can lead to unexpected ordering by dominating everything else
(for example with query "remcc" we rank "README.asciidoc" over
"src/remote.cc").

Fix these two issues by
1. switching to a restricted variant[*] of the Needleman-Wunsch
algorithm to find an optimal match instead of the leftmost one.
2. Dropping some of the old heuristics in favor of ones based on the
   optimal match.

Optimality is defined by a new matching heuristic which favors longer
subsequences by adding a penalty to each new gap.
On top of this, when comparing two matches, we first compare by the
number of gaps. This makes sure that short queries can prefer long
candidates over short (but worse) candidates. Alternatively, we could
just increase the penalty. I didn't do that yet because I didn't
encounter a real-world use case where the proposed heuristic fails.

Some heuristics are kept because they enable some test cases;
for example "higher number of matched word boundaries is better".
Since we now compute this heuristic based on an optimal match rather
than a greedy match, this metric should be more accurate too -
though we could probably achieve even better matches by including
word boundaries in the optimal-matching heuristic. Again I didn't
do this because I don't have a concrete example to improve.

The algorithm needs quadratic time/space. I believe this should be
mostly okay because we expect small inputs.
Insert completion candidates are usually below max_word_len=50.
The exception is line completion, (which is off in the default
completers).  With lines that are abnormally large, this can cause lag.
For example, if I have a 2 line with 2 million characters and a query
with 5, then `<c-x>l` takes like half a second. No harm I think since
other things are also slow given that line length.

Of course, there's plenty of room to optimize, for example we can
skip matching prefixes.

Every successful match adds a temporary heap allocation.  This feels
bad but it's still much lower than the number of string allocations,
so I didn't optimize that yet.

Closes #3806

[*]: Since we require a (smartcase) subsequence match, we need no
deletions or mismatches, which reduces the search space.

Alternative solution: another fuzzy-matching algorithm
with public domain implementations is described in
https://github.com/tajmone/fuzzy-search/tree/8a0e231f756cccd20f7c57e9593df08fd2bb79ef
but I didn't really look into it.

### DIFF
--- a/gdb/kakoune.py
+++ b/gdb/kakoune.py
@@ -235,5 +235,5 @@ def build_pretty_printer():
     pp.add_printer('ByteCount', '^Kakoune::ByteCount$', ByteCount)
     pp.add_printer('Color', '^Kakoune::Color$', Color)
     pp.add_printer('Regex', '^Kakoune::Regex$', Regex)
-    pp.add_printer('SubsequenceDistance', '^Kakoune::SubsequenceDistance$', SubsequenceDistance)
+    pp.add_printer('SubsequenceDistance', '^Kakoune::SubsequenceDistance<true>$', SubsequenceDistance)
     return pp
--- a/gdb/kakoune.py
+++ b/gdb/kakoune.py
@@ -206,6 +206,16 @@ class Regex:
     def to_string(self):
         return "regex%s" % (self.val["m_str"])
 
+class SubsequenceDistance:
+    """Print a SubsequenceDistance"""
+
+    def __init__(self, val):
+        self.val = val
+
+    def to_string(self):
+        reference = "*(%s*)%s" % (self.val.type, self.val.address)
+        return gdb.parse_and_eval("Kakoune::to_string(%s)" % reference)
+
 
 def build_pretty_printer():
     pp = gdb.printing.RegexpCollectionPrettyPrinter("kakoune")
@@ -225,4 +235,5 @@ def build_pretty_printer():
     pp.add_printer('ByteCount', '^Kakoune::ByteCount$', ByteCount)
     pp.add_printer('Color', '^Kakoune::Color$', Color)
     pp.add_printer('Regex', '^Kakoune::Regex$', Regex)
+    pp.add_printer('SubsequenceDistance', '^Kakoune::SubsequenceDistance$', SubsequenceDistance)
     return pp
--- a/src/array_view.hh
+++ b/src/array_view.hh
@@ -9,7 +9,7 @@ namespace Kakoune
 
 // An ArrayView provides a typed, non owning view of a memory
 // range with an interface similar to std::vector.
-template<typename T>
+template<typename T, typename SizeType = std::size_t>
 class ArrayView
 {
 public:
@@ -21,7 +21,7 @@ public:
     constexpr ArrayView(T& oneval)
         : m_pointer(&oneval), m_size(1) {}
 
-    constexpr ArrayView(T* pointer, size_t size)
+    constexpr ArrayView(T* pointer, SizeType size)
         : m_pointer(pointer), m_size(size) {}
 
     constexpr ArrayView(T* begin, T* end)
@@ -44,10 +44,10 @@ public:
         : m_pointer(v.begin()), m_size(v.size()) {}
 
     constexpr T* pointer() const { return m_pointer; }
-    constexpr size_t size() const { return m_size; }
+    constexpr SizeType size() const { return m_size; }
 
     [[gnu::always_inline]]
-    constexpr T& operator[](size_t n) const { return *(m_pointer + n); }
+    constexpr T& operator[](SizeType n) const { return *(m_pointer + (size_t)n); }
 
     constexpr T* begin() const { return m_pointer; }
     constexpr T* end()   const { return m_pointer+m_size; }
@@ -61,27 +61,28 @@ public:
 
     constexpr bool empty() const { return m_size == 0; }
 
-    constexpr ArrayView subrange(size_t first, size_t count = -1) const
+    constexpr ArrayView subrange(SizeType first, SizeType count = -1) const
     {
-        auto min = [](size_t a, size_t b) { return a < b ? a : b; };
+        auto min = [](SizeType a, SizeType b) { return a < b ? a : b; };
         return ArrayView(m_pointer + min(first, m_size),
                          min(count, m_size - min(first, m_size)));
     }
 
 private:
     T* m_pointer;
-    size_t m_size;
+    SizeType m_size;
 };
 
 template<typename It>
     requires std::contiguous_iterator<It>
 ArrayView(It begin, It end) -> ArrayView<std::iter_value_t<It>>;
 
-template<typename T>
-using ConstArrayView = ArrayView<const T>;
+template<typename T, typename SizeType = std::size_t>
+using ConstArrayView = ArrayView<const T, SizeType>;
 
-template<typename T>
-bool operator==(ArrayView<T> lhs, ArrayView<T> rhs)
+
+template<typename T, typename SizeType>
+bool operator==(ArrayView<T, SizeType> lhs, ArrayView<T, SizeType> rhs)
 {
     if (lhs.size() != rhs.size())
         return false;
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -166,6 +166,7 @@ static Completions complete_buffer_name(const Context& context, CompletionFlags 
     };
 
     StringView query = prefix.substr(0, cursor_pos);
+    RankedMatchQuery q{query};
     Vector<RankedMatchAndBuffer> filename_matches;
     Vector<RankedMatchAndBuffer> matches;
     for (const auto& buffer : BufferManager::instance())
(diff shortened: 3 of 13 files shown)


Screwtapello:
Does this new ranking algorithm affect https://github.com/mawww/kakoune/issues/1709#issuecomment-1053541569 ?

krobelus:
> Does this new ranking algorithm affect https://github.com/mawww/kakoune/issues/1709#issuecomment-1053541569 ?

No, we still sort all completions.

I wonder what the best solution is, maybe a `static-option=<option-name>` is enough, or perhaps a user-defined sorting function.
Maybe a priority field in the completions options, but it's not clear how the priority should interact with other things in `RankedMatch::operator<`.
Custom filtering is a similar issue although less important.

mawww:
Thats a big one, I need to take some time to review but I am happy to improve the fuzzy matching algorithm. I'll try to dedicate some time to go throug that soon.

krobelus:
I'm fairly happy with this now, though the heuristic could still be fine-tuned.
The last couple of commits are not so important / experimental, I can drop them.

krobelus:
Fixed a regression where we failed to prioritize exact matches over smartcase matches (for some reason other fuzzy finders don't do this).

krobelus:
Recent versions also optimize the number of gaps and matched word boundaries.
I found some weird orderings which made me realize that the gaps/word boundaries need a slightly [different algorithm](https://en.wikipedia.org/wiki/Smith%E2%80%93Waterman_algorithm#Affine), I'll work on that.

nonumeros:
awesome job @krobelus.  

I've read tidbits from all the sources/related issues you have referenced and I have to admit to have found fascinating, the whole heuristics rank that  kakoune  implements.  Truly impressive. 

So, given a number of files to edit under the `src` dir , the string match would be, say, a ` ha.c` 

So, assuming that the files runder the `src` dir are as tollow:

hash.cc
hash_map.cc
hook_manager.cc
changes.cc
input_hadler.cc
shared_string.cc
shell_manager.cc

 Prety much a before-and-after scenarios would result in the following

**After the changes with this PR**

hash.cc
hash_map.cc
input_handler.cc
changes.cc
shared_string.cc
hook_manager.cc
shell_manager.cc

and with the same token if I were to have a list of  txt files named accordingly `chattr.txt`, `chgrp.txt`, `chmod.txt`, and `chown.txt` and the string  to match is `:e ch` then,

**Before**
Screenshot_2022-12-23_20-03-11

and

**After**
Screenshot_2022-12-23_20-03-30

would just provide me with an uncertaintly  and whether in fact, the algorithm, is working as it should. 

I'm not versed on math, and I guess that could be a problem. A major problem. So to speak. And paraphrasing Bernie, — who once mentored Wittgenstein — would have said: upon the preface of the latter's work: «that leaves me in intelectual discomfort»  A perpetual discomfort. 

When you take a look at the list of text files such as those imaginary files such as, (sorry to sound redundant now) `chattr.txt`, `chgrp.txt`, `chmod.txt`, and `chown.txt` one can't stop wondering why  it would have chosen `chgrp` over `chattr` or `chmod` over `chown` .  Perhaps the latter is self-explanatory and it follows the sequence. But I can't stop but wondering why it would have picked `chgrp` over `chattr` for example. 

The letter `a` follows `ch` which makes `chattr` perhaps more appropriate than `chgrp` from  the list of completion candidates. 

Wasn't that one of the first problems that came up with the Roman numerals as opposed with the introduction later on, of the Arabic numeral system? 

The current proposal from your pull request is pretty much based on ordinal sequence matching. Not cardinal. Or so I think. I may be wrong here. But I have nothing to lose either way. Heck. I'm not writing a paper or trying to impress anyone. And I'm fine to sound like a fool. 

I also read about the idea to implement the matching similarly to whatever algorithm Google uses on its n gram viewer. 

I mean… it's more than obvious that a simple query for `chattr`, `chgrp`, and `chmod` would result in the latter showing the highest match.  The first string is not as widely used as the rest. So I don't think is the most adequate indicator  to pick an algorithm based on the pervasive usage of a particular term. 

But. As I said. I find it fascinating nonetheless. 


Screenshot_2022-12-23_21-27-05

krobelus:
yeah, given query `ch`, this change sorts `chgrp` before `chattr` because the first is shorter.
If the query is empty we fall back to lexicographical order, which is the other way round here.

nonumeros:
> yeah, given query `ch`, this change sorts `chgrp` before `chattr` because the first is shorter. If the query is empty we fall back to lexicographical order, which is the other way round here.

Is that the weight by which @Screwtapello  was referring to? 

I just have a genuine curiosity about this. It  does make sense what you're implementing with the changes from the pull request. But on the other hand, and since I find it fascinating…. 

If, as you said, and I quote 

> this change sorts `chgrp` before `chattr` because the first is shorter.

then it follows that given, say, two (2) files named 

101.txt
12.txt

**After the pull request** , sort matching them would yield (in that order):

12.txt
101.txt

But the above — according to some sources —  would not be considered a valid lexicographical ordered set. At least for matching a string of numbers. 

According to the following  and allow me to copy-paste:

* The phrase lexicographic order means in alphabetical order. It is a common term in computer science. It may indicate that numbers are to be treated as [strings](https://en.wiktionary.org/wiki/string) so 10 comes before 2 because there is no positional weighting and the symbol 1 comes before 2.

https://en.wiktionary.org/wiki/lexicographic 

and although I haven't checked for sources to warrant the above statement, let's assume it's correct, then:

given a set of files

* 12.txt
* 101.txt
* 12abc.txt
* 101abc.txt

before the pull request it would yield (in that order):

* 101abc.txt
* 101.txt
* 12abc.txt
* 12.txt

Screenshot_2022-12-24_05-43-07


and **after the pull request** it'd yield: 

* 101.txt
* 101abc.txt
* 12.txt
* 12abc.txt


Screenshot_2022-12-24_05-40-05

Wrapping it up and again, going solely by 

* The phrase lexicographic order means in alphabetical order. It is a common term in computer science. It may indicate that numbers are to be treated as [strings](https://en.wiktionary.org/wiki/string) so 10 comes before 2 because there is no positional weighting and the symbol 1 comes before 2.

a **before the pull request** a string match `.txt` would return (in that order):

* 2.txt
* 10.txt
* 12.txt
* 101.txt
* 12abc.txt
* 101abc.txt

Screenshot_2022-12-24_06-18-31


and **after the pull request** a string match `.txt` would yield:

* 2.txt
* 10.txt
* 12.txt
* 101.txt
* 12abc.txt
* 101abc.txt

Screenshot_2022-12-24_06-19-00


which is not convincing enough it does follow a lexicographical order of sorts (no pun intended).

My closest sources I can rely on, are not versed enough on string theory so I can't elaborate on that. 

And as I said before,it seems that as long as it's a combination of string of numbers and characters, perhaps it corroborates its validity, but based solely on numbers alone, it doesn't.

krobelus:
> Is that the weight by which Screwtapello was referring to? 

No, this PR makes the fuzzy match scoring algorithm more robust and general.
The weight would be a way to let users influence the scoring algorithm, that's https://github.com/mawww/kakoune/pull/4813

> **After the pull request** , sort matching them would yield (in that order):
> 
> 12.txt
> 101.txt

This exposes the same quality as the "chgrp"/"chattr" case.
1. if the query is empty, the scores of "12.txt" and "101.txt" are equal. So  we fall back to lexicographic order.
2. if the query is non-empty, say "1"
   - without this PR, the scores of "12.txt" and "101.txt" are equal, so we still use lexicographic order.
   - with this PR, "12.txt" gets a better score than "101.txt" because the first is shorter.

This change is a consequence of the new fuzzy scoring algorithm.
Every mismatched character adds a small penalty.
Both `fzf` and `fzy` have the same behavior (empty query means lexicographic order while "1" picks the shorter one).

I think this behavior is good in general but I have no strong opinion about this specific case where the query is a prefix of the candidate.
If we want to change this case, the way to do so would be to add a preprocessing step that removes matching prefix.
Then "match 1 against 101.txt and 12.txt" becomes "match the empty string against 01.txt and 2.txt" which would use lexicographic order as before.

raiguard:
This algorithm is undoubtedly an improvement, but for file finding it still isn't as good as fzf. In my usecase it keeps ranking the test file ahead of the src file.

Kak (bad):
image

Fzf (good):
image

Here is the prompt command I am using in kak:
```kak
# Go to file (basic fuzzy finder)
prompt -shell-script-candidates %{ fd --type f --hidden } file: %{ edit %val{text} }
```

krobelus:
@raiguard thanks for the test case, I have tweaked the heuristic to make it work.

FWIW for buffer type completion it already worked this way (also without this PR) because Kakoune knows that the completions are file paths. For those we prioritize matches on the basenames (see 2df7b1f6d (In buffer name completion, give priority to filename (not dirname) matches, 2016-03-02)).
For your prompt command, we don't know if the input is a file path or some other token, so it seems wrong to prioritize basenames.
What I did was change the word boundary bonus so matching `H` against `/Here` is worth twice as much as matching it against `OverHere`.

krobelus:
latest master works fine for almost all test cases, it seems to be good enough for my current use cases

mawww:
Thanks again for exploring this and pushing for fuzzy matching to be improved.

PR #4668, opened 2022-07-11
Assign janet group to janet.kak global hooks

amano-kenji:
If groups are assigned to janet.kak global hooks, the hooks can be overridden by third party janet modules.

### DIFF
--- a/rc/filetype/janet.kak
+++ b/rc/filetype/janet.kak
@@ -4,12 +4,11 @@
 # Detection
 # ‾‾‾‾‾‾‾‾‾
 
-hook global BufCreate .*[.](janet|jdn) %{
+hook -group janet global BufCreate .*[.](janet|jdn) %{
     set-option buffer filetype janet
 }
 
-
-hook global WinSetOption filetype=janet %{
+hook -group janet global WinSetOption filetype=janet %{
     require-module janet
 
     hook window ModeChange pop:insert:.* -group janet-trim-indent janet-trim-indent


rc/filetype/janet.kak
```diff
@@ -4,12 +4,12 @@
...
 # ‾‾‾‾‾‾‾‾‾
 
-hook global BufCreate .*[.](janet|jdn) %{
+hook -group janet global BufCreate .*[.](janet|jdn) %{
     set-option buffer filetype janet
 }
 
 
-hook global WinSetOption filetype=janet %{
+hook -group janet global WinSetOption filetype=janet %{
```
krobelus, line 12: The traditional way of excluding bundled kak files is to create your own [autoload directory](https://github.com/mawww/kakoune#configuration-autoloading).

Your solution seems fine too though I don't immediately see why a third party janet.kak would want to suppress these benign hooks.

rc/filetype/janet.kak
```diff
@@ -19,7 +19,7 @@ hook global WinSetOption filetype=janet %{
     hook -once -always window WinSetOption filetype=.* %{ remove-hooks window janet-.+ }
 }
 
-hook -group janet-highlight global WinSetOption filetype=janet %{
+hook -group janet global WinSetOption filetype=janet %{
```
krobelus, line 22: I think this should keep the `janet-highlight` group, for consistency with other filetypes

rc/filetype/janet.kak
amano-kenji: I want third party janet.kak moduels to start with a clean plate. The simpler, the better.

amano-kenji:
The problem with creating my own autoload directory is that I can't just block one specific module. It can only work as a whitelist.

amano-kenji:
Can you merge this if there is no problem?

amano-kenji:
Since this is probably not going to be merged, I'm going to close it.

PR #4672, opened 2022-07-16
[faces.asciidoc] Amended English mechanics and styling for correctness and clarity.

Pound-Hash:
1) Fixed grammar & punctuation
2) Made formatting consistent
3) Replaced quoted words for bolded/italicized words, which style guides prefer (if we have computers capable of rendering boldness and italics, then why not?).
4) For the color specs, I swapped the explanans for the explananda (the explanations of words for the words being explained). This is more logical than listing a series of color names and then telling the reader that these color names are color names.
5) Capitalized beginning of long phrases, clauses, or sentences. Similarly, ended such passages with periods.
6) Removed redundancies or pleonasms.
7) Tried to rephrase certain passages for clarity.
8) Replaced certain quoted characters with backticks
9) Changed the wording of "elements" in menus to "items" in menus. The latter term has a meaning specific to lists, enumerations, and menus. "Elements", on the contrary, is ambiguous.

### DIFF
--- a/doc/pages/faces.asciidoc
+++ b/doc/pages/faces.asciidoc
@@ -2,183 +2,188 @@
 
 == Declaration
 
-A 'face' determines how text is displayed, it has a foreground
-color, a background color, and some attributes. The value of a face has the
+A *face* determines how text is displayed. It has a foreground color,
+a background color, and some attributes. The value of a face has the
 following format:
 
 -----------------------------------------------------------
 [fg_color][,bg_color[,underline_color]][+attributes][@base]
 -----------------------------------------------------------
 
-'fg_color', 'bg_color', 'underline_color'::
-    a color whose value can be expressed in the following formats:
-        *black*, *red*, *green*, *yellow*, *blue*, *magenta*, *cyan*, *white*:::
-        *bright-black*, *bright-red*, *bright-green*, *bright-yellow*:::
-        *bright-blue*, *bright-magenta*, *bright-cyan*, *bright-white*:::
-            a named color
-        *default*:::
-            keep the existing color
-        *rgb:RRGGBB*:::
-            hexadecimal value
-        *rgba:RRGGBBAA*:::
-            hexadecimal value
-    if unspecified or empty, *default* is used.
-    Alpha values are used to blend the face onto either its base or the existing
-    color. For technical reasons alpha values must be >16 if specified.
+fg_color, bg_color, underline_color::
+    A color whose value can be:
+        A named color:::
+            *black*, *red*, *green*, *yellow*, *blue*, *magenta*, *cyan*,
+            *white* *bright-black*, *bright-red*, *bright-green*,
+            *bright-yellow* *bright-blue*, *bright-magenta*, *bright-cyan*,
+            *bright-white*
+       The existing color:::
+            *default*
+        A hexadecimal value:::
+            *rgb:RRGGBB*, *rgba:RRGGBBAA*
+    If left unspecified, the value of *default* is used.
+    Alpha values are used to blend the face onto either its base (see below) or
+    else onto whatever color happens to be used at the moment. For technical
+    reasons, alpha values must be >16.
 
-'attributes'::
-    string whose individual letters set an attribute:
+attributes::
+    A string whose individual letters each set an attribute:
         *u*:::
-            underline
+            Underline
         *c*:::
-            curly underline, takes precedence over underline if both are
-            specified
+            Curly underline
+            Note: This takes precedence over underline if both are specified.
         *r*:::
-            reverse
+            Reverse
         *b*:::
-            bold
+            Bold
         *B*:::
-            blink
+            Blink
         *d*:::
-            dim
+            Dim
         *i*:::
-            italic
+            Italic
         *s*:::
-            strikethrough
+            Strikethrough
         *F*:::
-            final, override the previous face instead of merging with it,
-            and this face will only be replaced if another face with
-            the final attribute is applied
+            Final::::
+                Override the previous face instead of merging with
+                it. Can only be replaced if another face with the final
+                attribute is applied.
         *f*:::
-            final foreground, as final but only applies to face's
-            foreground color
+            Final foreground::::
+                Same as final, but only applies to a face's foreground color.
         *g*:::
-            final background, as final but only applies to face's
-            background color
+            Final background::::
+                Same as final, but only applies to a face's background color.
         *a*:::
-            final attributes, as final but only applies to face's
-            attributes
+            Final attributes::::
+                Same as final, but only applies to a face's attributes.
 
-'base'::
-    The base face on which this face applies, which can be any face name,
-    as long as no cycle is introduced. A face can reference itself, in
-    which case it will apply on top of the parent scope version.
+base::
+    The face onto which the rest of the face applies. Its value can be any
+    face name, as long as no cycle is introduced. A face can reference itself,
+    in which case it will apply on top of the parent scope.
 
-== Builtin faces
+== Built-in faces
 
-The following default faces are used by color schemes to highlight certain
-areas of the user interface:
+The following faces are used by color schemes to highlight certain areas of
+the user interface:
 
 *Default*::
-    default colors
+    Default colors:::
+        The default foreground and background colors. If the value of *default*
+        is used here, then the colors used will be your terminal's defaults.
 
 *PrimarySelection*::
-    main selection face for every selected character except the cursor
+    Main selection face for every selected character except the cursor.
 
 *SecondarySelection*::
-    secondary selection face for every selected character except the cursor
+    Secondary selection face for every selected character except the cursor.
 
 *PrimaryCursor*::
-    cursor of the primary selection
+    Cursor of the primary selection.
 
 *SecondaryCursor*::
-    cursor of the secondary selection
+    Cursor of the secondary selection.
 
 *PrimaryCursorEol*::
-    cursor of the primary selection when it lies on an end of line character
+    Cursor of the primary selection when it lies on an EOL (end of line)
+    character.
 
 *SecondaryCursorEol*::
-    cursor of the secondary selection when it lies on an end of line character
+    Cursor of the secondary selection when it lies on an EOL (end of line)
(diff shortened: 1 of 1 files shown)


doc/pages/faces.asciidoc
```diff
@@ -2,183 +2,188 @@
...
+            Blink
         *d*:::
-            dim
+            Dim
         *i*:::
-            italic
+            Italic
         *s*:::
-            strikethrough
+            Strikethrough
```
krobelus, line 47: I don't think we should capitalize this, it looks worse IMHO

doc/pages/faces.asciidoc
```diff
@@ -2,183 +2,188 @@
...
+                Same as final, but only applies to a face's attributes.
 
-'base'::
-    The base face on which this face applies, which can be any face name,
-    as long as no cycle is introduced. A face can reference itself, in
-    which case it will apply on top of the parent scope version.
+base::
+    The face onto which the rest of the face applies. Its value can be any
+    face name, as long as no cycle is introduced. A face can reference itself,
+    in which case it will apply on top of the parent scope.
```
krobelus, line 66: `parent scope` sounds weird because a face can't apply to a scope.
The old phrasing is `parent scope version` which is better - it means the base is the face of the same name but in the parent scope.

doc/pages/faces.asciidoc
```diff
@@ -2,183 +2,188 @@
...
 
 *Default*::
-    default colors
+    Default colors:::
+        The default foreground and background colors. If the value of *default*
+        is used here, then the colors used will be your terminal's defaults.
 
 *PrimarySelection*::
-    main selection face for every selected character except the cursor
+    Main selection face for every selected character except the cursor.
```
krobelus, line 79: I agree with adding capitalization/punctuation to full sentences but this is an ellipsis, where we usually don't do either.
There's a similar style in `doc expansions`, `doc commands` and docstrings in autoinfo (`:echo`). They are "definition lists" where the verb is left out, because it's implied by the context

doc/pages/faces.asciidoc
```diff
@@ -2,183 +2,188 @@
...
-    face used to highlight the line number of the main selection
+    Face used to highlight the line number of the main selection.
 
 *LineNumbersWrapped*::
-    face used to highlight the line number of wrapped lines
+    Face used to highlight the line number of wrapped lines.
 
 *MatchingChar*::
-    face used by the `show-matching` highlighter
+    Face used by the *show-matching* highlighter.
```
krobelus, line 155: yeah, making the literal names bold seems good

doc/pages/faces.asciidoc
```diff
@@ -2,183 +2,188 @@
...
+*Default* face, and "error" in the *Error* face:
 
 ----
 echo -markup 'default {Error}error{Default} default'
 ----
 
-Inside a markup string, a literal `{` character is written `\{`, and a
-literal backslash (`\`) that precedes a '{' character is escaped as well
+Inside a markup string, a literal `{` character is written `\{`, and a literal
+backslash `\` character that precedes a `{` character is escaped as well
```
krobelus, line 178: unrelated to your changes, so feel free to ignore: this documentation is misleading.
All backslashes in markup strings must be escaped.
For example,

	echo -markup \\

outputs only one backslash.

So I'd write something like

	Inside a markup string, a literal `{` character is written as `\{`, and a literal
	backslash (`\`) character is written as `\\`.

doc/pages/faces.asciidoc
```diff
@@ -2,183 +2,188 @@
...
-    if unspecified or empty, *default* is used.
-    Alpha values are used to blend the face onto either its base or the existing
-    color. For technical reasons alpha values must be >16 if specified.
-
-'attributes'::
-    string whose individual letters set an attribute:
+fg_color, bg_color, underline_color::
+    A color whose value can be:
+        A named color:::
+            *black*, *red*, *green*, *yellow*, *blue*, *magenta*, *cyan*,
```
krobelus, line 16: Yeah, this order makes more sense

doc/pages/faces.asciidoc
```diff
@@ -2,183 +2,188 @@
...
+            *white* *bright-black*, *bright-red*, *bright-green*,
+            *bright-yellow* *bright-blue*, *bright-magenta*, *bright-cyan*,
+            *bright-white*
+       The existing color:::
```
krobelus, line 20: This is misaligned with the two other lines that describe the other variants (missing one space here)

Also maybe write "The color of the base face"?
Though I'm not sure what happens if there is no base face?

doc/pages/faces.asciidoc
Pound-Hash: Okay, after checking with the OED and [MW](https://www.merriam-webster.com/dictionary/editor), I agree with you that the definitions don't need to be capitalized or punctuated at the end. I'll revert these changes.

doc/pages/faces.asciidoc
Pound-Hash: The ellipsis is the omitted part. It's still a sentence - an elliptical sentence. Regardless, as mentioned in the other comment, these definitions don't need capitalization or punctuation at the end. I'll revert these changes.

Pound-Hash: I see now the original issue. The problem is that some definitions require more than one clause to explain. Writing such clauses is messy without the normal delimiters of capitalization on the left end and punctuation on the right end. For example, the definition of the `final` attribute:

> Override the previous face instead of merging with it. Can only be replaced if another face with the final attribute is applied.

If such multi-clause definitions are required, I would think it best to adopt a uniform look for all definitions, regardless of the number clauses, if any.

The alternative would be to begin and end certain definitions one way, and other definitions a different way, which lacks uniformity and hence is jarring.

doc/pages/faces.asciidoc
Pound-Hash: We can continue this line of thought in the other comment thread below.

doc/pages/faces.asciidoc
Pound-Hash: Sorry, I don't know what to make of this. The word **version** seems pleonastic to me:

If I have two books called "Moby Dick", each being different in some way (published year, for example), I can refer to one of them by saying either: a) The older one is in better shape, or b) The older version is in better shape.

The word **version** doesn't add anything meaningful. It's a noun redundantly affixed to another noun. The object is the same regardless. Consider, "Roses are pretty" vs "Rose flowers are pretty".

doc/pages/faces.asciidoc
Pound-Hash: I'll make your suggested change. But why the parentheses around the first backslash (`\`)? Aren't the backticks enough?

doc/pages/faces.asciidoc
Pound-Hash: Thanks.

I've made your changes.

I asked about the base face a few months ago. I think the answer is that it uses the terminal's defaults.

doc/pages/faces.asciidoc
krobelus: > But why are there parentheses around the first backslash (\)? Aren't the backticks enough?

Backticks are not rendered by `:doc`, and on GitHub translate to grey background. The resulting "backslash \" is a pleonasm.

doc/pages/faces.asciidoc
krobelus: To me 

	A face can reference itself, in which case it will apply on top of the parent scope.

sounds a bit like

	Do you want to read this copy of "Moby Dick" or the public library?

From that perspective, adding "version" at the end makes a difference.

I also see your point that it's obvious to humans but it trips up my internal type checker.
Maybe we can find an elegant rephrasing.

doc/pages/faces.asciidoc
krobelus: You're right, thanks for your tenacity. I agree that we should achieve consistency.

How about a compromise: keep the unabbreviated attribute lowercase and put everything beyond that in proper sentences. Rough sketch:

```diff
        *F*:::
            final::::
                Override the previous face instead of merging with
                it. Can only be replaced if another face with the final
                attribute is applied.

        *c*:::
            curly underline::::
                Note: This takes precedence over underline if both are specified.
```

doc/pages/faces.asciidoc
Pound-Hash: I doubt that it's obvious to even 1% of native English speakers, let alone humans.

Anyhow, as I was just reverting the wording to "parent scope version", I realized from your comments that what we may want to say is, "version of the parent scope" or "parent scope's version". The preposition in the former, and the possessive in the latter, do much in the way of disambiguation. Would one of these serve us well? Or shall I revert to the original "parent scope version", thus betraying my own linguistic integrity (joking).

doc/pages/faces.asciidoc
Pound-Hash: Okay, will do.

P.S. I suspect it's more stubbornness than tenacity...

doc/pages/faces.asciidoc
Pound-Hash: Haha, now you're talking like me.

Note, pleonasms are about words, not symbols: "It _may_ be possible" instead of "It _is_ possible"; the **may** is already entailed by the **possible**, hence redundant.

Anyhow, shall I just use your suggestion for this passage?

> Inside a markup string, a literal `{` character is written as `\{`, and a literal
> backslash (`\`) character is written as `\\`.

doc/pages/faces.asciidoc
krobelus: Good idea, I think both "version of the parent scope" and "parent scope's version" are better.

doc/pages/faces.asciidoc
krobelus: Yeah I think that's a good improvement

doc/pages/faces.asciidoc
Pound-Hash: Done

doc/pages/faces.asciidoc
Pound-Hash: Done

PR #4677, opened 2022-07-21
Fix autoinfo for "set-option -remove" not showing option-specific info

krobelus:


src/ranges.hh
```diff
@@ -574,6 +575,13 @@ bool contains(Range&& range, const T& value)
     return find(range, value) != end(range);
 }
 
+template<typename T, typename U>
+bool contains(std::initializer_list<T> range, const U& value)
```
mawww, line 579: I would rather avoid having to introduce a copy of each algorithm for initializer_list, so until we find a solution that does not requires duplication I'd rather explicitely pass Array{x, y}

krobelus: Makes sense, I went with the safe choice of `... == ... or ... == ...` now, which we use in lots of places.

Issue #2402, opened 2018-09-18
Standardization of plugin file-structure layout


andreyorst:
Hello. I'm developing a plugin manager for Kakoune (for personal use, but if anyone interested, you're welcome). Since there's no currently de facto standard of Kakoune plugin project layout, I'd like to propose Vim-like one (with some exceptions, because of structure), because it works really well.

### Vim's plugin structure
Owerall structure of most plugins can be described like this:
```
.
├── after
│   └── plugin
├── autoload
├── doc
├── ftplugin
├── indent
├── plugin
└── syntax
```

Not all folders are necessary, it is highly dependent on the plugin's task. However, the possibilities that come with abilities off putting different modules of the plugin to the supported methods of loading via different directories are huge.

In Vim plugin we have essential `plugin` folder, which stores all the plugin scripts (I'm dropping autoload part, because there's no equivalent in Kakoune for it, which is sort of a shame, but whatever), `doc` folder, which contains documentation, and sometimes `ftplugin` folder, which acts the same way as plugin except it allows loading plugin only for a certain `filetype` (which again isn't implemented in Kakoune, AFAIK).

This sturcture allows easy unified mechanics for all plugins, to be loaded, loaded with specific `filetype`, parts of plugin can be `autoload`ed only when really needed, which boosts overall load time, adding `syntax` as a part of plugins functions (like Rust plugin adds command interface to `cargo` through Vim, plus adds advanced syntax highlighting on top of Vim's one), `indent`ation settings for this filetipe, and more. All of this leads to ability of developing really great and advanced plugins, that can turn Vim into full featured IDE.

### Documentation
One of the most powerful Vim feature is help system, and one of the most powerful feature of vim's help system is that it can be extended easily. As a plugin creator responsible of writing a comprehensive documentation, editor is responsible of making it easy to refer to that documentation.

Plugins tend grow, and it is natural for them to have a documentation. For example, look at the documentation of [LanguageClient-Neovim](https://github.com/autozimu/LanguageClient-neovim/blob/next/doc/LanguageClient.txt) or [UltiSnips](https://github.com/SirVer/ultisnips/blob/master/doc/UltiSnips.txt). It's kinda huge. And it is meaningless to store all that in the README.md, since this file should contain only essential part about a plugin (although some plugin maintainers do that). 

[...]

PR #4688, opened 2022-07-25
Add docsextrapaths option for extra doc search folders

razcore-rad:
This PR adds the `str-list` `docsextrapaths` option so we can add extra search paths, for example when using `plug.kak`. For example.

```kak
set-option global docsextrapaths "%val{config}/plugins"
```


mawww:
Realised I never commented on this. I wonder if this is the best approach. With the existing design `plug.kak` could just symlink the pages somewhere in `${kak_config}/autoload/` hierarchy and things should just work. Is there an issue with that approch ?

razcore-rad:
The discussion on docs was covered to some extent in https://github.com/mawww/kakoune/issues/2402. As I don't know the internals of `plug.kak`, to me adding `docsextrapaths` was the easiest approach. I don't really have a strong opinion on either option.

For me, this setup works pretty well. One major issue is when we have multiple plugins with `README.asciidoc` and there's no way to distinguish between them in kakoune. Perhaps `plug.kak` should be intelligent enough at handling symlinks by linking with the name of the plugin if it finds a `README.asciidoc` perhaps. But I don't really have the time to look into this approach.

PR #4690, opened 2022-07-28
Add closing delimiter to completions of %opt/%file/%reg/%val expansion

krobelus:
This looks a bit weird but it's exactly what the user needs to type anyway.

This means we can type

    echo %val{bufn<ret>

and it will do the right thing.


src/command_manager.hh
```diff
@@ -52,13 +52,15 @@ struct Token
...
-        CommandSeparator
+        CommandSeparator,
+        UnknownExpand
     };
 
     Type type;
     ByteCount pos;
     String content;
     bool terminated = false;
+    Optional<Codepoint> missing_closing_delimiter{};
```
mawww, line 63: It feels like terminated and missing_closing_delimiter are quite coupled, I wonder if there would be a way to have a single field for both information.

krobelus: valid configurations are:

	terminated=false missing_closing_delimiter=None
	terminated=false missing_closing_delimiter='}'
	terminated=true missing_closing_delimiter=None

two separate fields allow this configuration, but it's impossible in practice:

	terminated=true missing_closing_delimiter='}'

I'll push an attempt that gets rid of that configuration by using `std::variant`.
Unfortunately, the naive approach (pasted below) is quite ugly so I added a separate class. 

```diff
diff --git a/src/command_manager.cc b/src/command_manager.cc
index 2b24965fb..12137c20a 100644
--- a/src/command_manager.cc
+++ b/src/command_manager.cc
@@ -286,7 +286,7 @@ Token parse_percent_token(ParseState& state, bool throw_on_unterminated)
 
     auto start = state.pos;
     const ByteCount byte_pos = start - state.str.begin();
-    Optional<Codepoint> missing_closing_delimiter{};
+    decltype(Token::terminated) terminated{Token::Unterminated{}};
 
     if (auto it = find_if(matching_pairs, [=](const CharPair& cp) { return opening_delimiter == cp.opening; });
         it != std::end(matching_pairs))
@@ -300,9 +300,9 @@ Token parse_percent_token(ParseState& state, bool throw_on_unterminated)
                                      it->opening, it->closing)};
         }
         if (not quoted.terminated)
-            missing_closing_delimiter = it->closing;
+            terminated = Token::Unterminated{ it->closing };
 
-        return {type, byte_pos, std::move(quoted.content), quoted.terminated, missing_closing_delimiter};
+        return {type, byte_pos, std::move(quoted.content), terminated};
     }
     else
     {
@@ -317,9 +317,9 @@ Token parse_percent_token(ParseState& state, bool throw_on_unterminated)
                                      opening_delimiter, opening_delimiter)};
         }
         if (not quoted.terminated)
-            missing_closing_delimiter = opening_delimiter;
+            terminated = Token::Unterminated{ opening_delimiter };
 
-        return {type, byte_pos, std::move(quoted.content), quoted.terminated, missing_closing_delimiter};
+        return {type, byte_pos, std::move(quoted.content), terminated};
     }
 }
 
@@ -428,7 +428,9 @@ Optional<Token> CommandParser::read_token(bool throw_on_unterminated)
         return Token{c == '"' ? Token::Type::RawEval
                               : Token::Type::RawQuoted,
                      start - line.begin(), std::move(quoted.content),
-                     quoted.terminated};
+                     quoted.terminated
+                         ? decltype(Token::terminated)(Token::Terminated{})
+                         : decltype(Token::terminated)(Token::Unterminated{ c })};
     }
     else if (c == '%')
     {
@@ -697,7 +699,7 @@ static Completions complete_expansion(const Context& context, CompletionFlags fl
         if ((token.type == Token::Type::ValExpand and c.ends_with("_")) // opt_, reg_, client_env_
             or (token.type == Token::Type::FileExpand and c.ends_with("/")))
             continue;
-        c += to_string(*token.missing_closing_delimiter);
+        c += to_string(*std::get<Token::Unterminated>(token.terminated).missing_closing_delimiter);
     }
 
     auto completions_flags = token.type == Token::Type::RegisterExpand
@@ -719,7 +721,7 @@ static Completions complete_raw_eval(const Context& context, CompletionFlags fla
             else
             {
                 auto token = parse_percent_token(state, false);
-                if (token.terminated)
+                if (std::holds_alternative<Token::Terminated>(token.terminated))
                     continue;
                 if (token.type == Token::Type::Raw or token.type == Token::Type::RawQuoted)
                     return {};
@@ -764,7 +766,7 @@ Completions CommandManager::complete(const Context& context,
     kak_assert(not tokens.empty());
     const auto& token = tokens.back();
 
-    if (token.terminated) // do not complete past explicit token close
+    if (std::holds_alternative<Token::Terminated>(token.terminated)) // do not complete past explicit token close
         return Completions{};
 
     auto requote = [](Completions completions, Token::Type token_type) {
diff --git a/src/command_manager.hh b/src/command_manager.hh
index 04867d6bc..54fc6354a 100644
--- a/src/command_manager.hh
+++ b/src/command_manager.hh
@@ -13,6 +13,7 @@
 
 #include <functional>
 #include <initializer_list>
+#include <variant>
 
 namespace Kakoune
 {
@@ -59,8 +60,9 @@ struct Token
     Type type;
     ByteCount pos;
     String content;
-    bool terminated = false;
-    Optional<Codepoint> missing_closing_delimiter{};
+    struct Terminated {};
+    struct Unterminated { Optional<Codepoint> missing_closing_delimiter; };
+    std::variant<Terminated, Unterminated> terminated = Unterminated{};
 };
 
 struct ParseState

```

krobelus: the we might be able to simplify this by solving 0cecea659 (Fix completion past explicitely closed tokens, 2021-05-18) in a different way. Essentially, just like with this patch, the completions should add the closing delimiter. Then the `terminated`  bit should be redundant.
I'll investigate.

krobelus: ok I found a way to exclude the invalid configuration by design.

---

Now adding completions after explicitly closed tokens is easy but probably not useful:

```diff
diff --git a/src/command_manager.cc b/src/command_manager.cc
index c7fe460d8..a23cba5e5 100644
--- a/src/command_manager.cc
+++ b/src/command_manager.cc
@@ -760,10 +760,7 @@ Completions CommandManager::complete(const Context& context,
     kak_assert(not tokens.empty());
     const auto& token = tokens.back();
 
-    if (token.terminator and token.terminator->present) // do not complete past explicit token close
-        return Completions{};
-
-    auto requote = [](Completions completions, Token::Type token_type) {
+    auto requote = [&token](Completions completions, Token::Type token_type) {
         if (completions.flags & Completions::Flags::Quoted)
             return completions;
 
@@ -779,7 +776,14 @@ Completions CommandManager::complete(const Context& context,
             }
         }
         else if (token_type == Token::Type::RawQuoted)
+        {
+            if (token.terminator and token.terminator->present)
+            {
+                for (auto& candidate : completions.candidates)
+                    candidate.push_back(token.terminator->character);
+            }
             completions.flags |= Completions::Flags::Quoted;
+        }
         else
             kak_assert(false);
```

PR #4692, opened 2022-07-30
Do not record prompt history when executing user mode mappings

krobelus:
<details>

Commit 217dd6a1d (Disable history when executing maps, 2015-11-10)
made it so with

	map global normal X %{:echo 123<ret>}

X does not add to prompt history (%reg{:}).

Unfortunately this behavior was not extended to mappings in the "user"
keymap, nor to mappings in custom user modes.
In my experience, not adding to history is almost always the expected
behavior for mappings. Most users achieve this by adding a leading space:

	map global user X %{: echo 123<ret>}

but that's awkward. We should have good defaults (no nnoremap)
and map should work the same way across all modes.

Fix this by also disabling history when executing user mappings. This
is a breaking change but I think it only breaks exotic use cases.

I found some uses where user mappings add to history but none of them
looks intentional.

## Introduce "map -with-history" to record prompt history

My terminal allows to map `<c-[>` and `<esc>` independently.  I like
to use `<c-[>` as escape so I use

	map global insert <c-[> <esc>
	map global prompt <c-[> <esc>

Unfortunately, the prompt mapping is not equivalent to `<esc>`: since
mappings are run with history disabled, `<c-[>` will not add the
unfinished command to the prompt history.

Enable this by adding a -with-history switch.

This is also useful when mapping something to `<ret>`.

</details>

### DIFF
--- a/doc/pages/changelog.asciidoc
+++ b/doc/pages/changelog.asciidoc
@@ -18,6 +18,8 @@ released versions.
 * History registers `%reg{colon}`, `%reg{slash}` and `%reg{pipe}` now
   have reverse chronological order
 
+* Executing user mode mappings no longer adds to prompt history registers by default.
+
 == Kakoune 2021.11.07
 
 * Support for curly and separately colored underlines (undocumented in 2021.10.28)
--- a/doc/pages/mapping.asciidoc
+++ b/doc/pages/mapping.asciidoc
@@ -59,12 +59,8 @@ It's common to use a normal-mode or user-mode mapping to trigger a command,
 like this:
 
 ----
-map global user n ': make-next-error<ret>'
+map global user n :make-next-error<ret>
 ----
-
-Note the space between the `:` and the command. This prevents Kakoune from
-adding this command to the prompt history, so the user won't have to scroll
-past it to review commands they actually typed.
 
 If you make a normal-mode mapping, you can prefix it with a count or a register
 name like any other normal-mode key. You can forward this information to the
--- a/doc/pages/mapping.asciidoc
+++ b/doc/pages/mapping.asciidoc
@@ -68,7 +68,7 @@ command you invoke with the `%val{count}` and `%val{register}` expansions
 (See <<expansions#`:doc expansions`>>). For example:
 
 ----
-map global normal = ': echo Got count %val{count} and reg %val{register}<ret>'
+map global normal = ':echo Got count %val{count} and reg %val{register}<ret>'
 ----
 
 == Mappable keys
--- a/rc/filetype/diff.kak
+++ b/rc/filetype/diff.kak
@@ -4,7 +4,7 @@ hook global BufCreate .*\.(diff|patch) %{
 
 hook global WinSetOption filetype=diff %{
     require-module diff
-    map buffer normal <ret> %{: diff-jump<ret>}
+    map buffer normal <ret> :diff-jump<ret>
 }
 
 hook -group diff-highlight global WinSetOption filetype=diff %{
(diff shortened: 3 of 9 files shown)


Screwtapello:
Hm. Does `exec :info<space>hello<ret>` add to the command history? Should it?

How difficult would it be to make "add to history" something triggered by the UI, above the layer where `exec` and `map` operate?

Alternatively, how difficult would it be to teach `map` a `-save-regs` switch like `exec -save-regs` and put `/|:` in it by default?

krobelus:
> Hm. Does `exec :info<space>hello<ret>` add to the command history? Should it?

Interesting. It does not, because the default for -save-regs is '/"|^@:'
whereas evalute-commands defaults to saving no registers, should probably document that.

> How difficult would it be to make "add to history" something triggered by the UI, above the layer where `exec` and `map` operate?

You mean so exec/map never add to history? That would be trivial but lack support for things like `map global prompt <c-j> <ret>`. (Only relevant for `<esc>` and `<ret>` but still.)

> Alternatively, how difficult would it be to teach `map` a `-save-regs` switch like `exec -save-regs` and put `/|:` in it by default?

Yeah that sounds better.
It feels a bit inelegant to copy history registers every time we execute a mapping but I don't think it matters, plus we can avoid the copy for read-only registers.

Screwtapello:
> It does not, because the default for `-save-regs` is `/"|^@:`

Hmm. I admit I didn't test it myself, but the docs [say](https://github.com/mawww/kakoune/blob/master/doc/pages/execeval.asciidoc#execute-keys-and-evaluate-commands):

> Without the -save-regs switch, execute-keys saves the following registers, which are then restored when the keys have been executed: /, ", |, ^, @.

If `:` has been added to the list of saved registers, I guess that should be added to the docs too.

krobelus:
implemented -save-regs now. Not pretty but it works for me.

> > > Does `exec :info<space>hello<ret>` add to the command history?
> > It does not, because the default for `-save-regs` is `/"|^@:`
> 
> Hmm. I admit I didn't test it myself, but the docs [say](https://github.com/mawww/kakoune/blob/master/doc/pages/execeval.asciidoc#execute-keys-and-evaluate-commands):
> 
> > Without the -save-regs switch, execute-keys saves the following registers, which are then restored when the keys have been executed: /, ", |, ^, @.

Turns out even without -save-regs exec/eval never write to prompt history registers (unless they run set-register). Seems very reasonable.

> 
> If `:` has been added to the list of saved registers, I guess that should be added to the docs too.

good catch, done

krobelus:
Hmm the current version breaks

	map global normal x %{:set-register / ^@@<ret>}

perhaps the default should be only `map -save-regs :` because that's rarely every written to.
Though it might be better to go with all prompt history registers for consistency..

Screwtapello:
When I suggested `-save-regs` I was thinking of the "set a register, use the default" idiom to avoid key-quoting arbitrary text (for example, `set-register / %val{selection} ; exec /<ret>` to avoid having to "type" the selection content in the exec command. Setting a register then immediately using it within the same mapping should be fine.

I'm not sure why you'd make a mapping to push things into a history register without immediately using it, though. Is that a mapping you found in the wild, or just a potential compatibility issue you thought of?

krobelus:
The set-register scenario is mainly hypothetical but it is an indicator that my attempted solution is a bad idea.
In the meantime I have realized I don't care for making this configurable; I just want to add to history whenever I close an interactive prompt by default.
(Making it configurable might make the average experience worse actually, because it complicates the `map` autoinfo.)
Your first suggestion is the correct solution:

> How difficult would it be to make "add to history" something triggered by the UI, above the layer where `exec` and `map` operate?

I'll do that in a separate PR, so the consistency fix can live here.

PR #4695, opened 2022-08-01
Disable history only for prompts that are never shown in the UI

krobelus:
My terminal allows to map `<c-[>` and `<esc>` independently.  I like
to use `<c-[>` as escape key so I have this mapping:

        map global prompt `<c-[>` `<esc>`

Unfortunately, this is not equivalent to `<esc>`.  Since mappings are
run with history disabled, `<c-[>` will not add the command to the
prompt history.

So disabling command history inside mappings is wrong in case the
command prompt was created before mapping execution. The behavior
should be: "a prompt that is both created and closed inside a
noninteractive context does not add to prompt history", where
"noninteractive" means inside a mapping, hook, command, execute-keys
or evaluate-commands.

Implement this behavior, it should better meet user expectations.
Scripts can always use "set-register" to add to history.

Here are my test cases:

1. Basic regression test (needs above mapping):

        :nop should be added to history<c-[>

---

2. Create the prompt in a noninteractive context:

        :exec %{:}

now we're back in the interactive context, so we can type:

        nop should be added to history<ret>

---

3. To check if it works for nested prompts, first set up this mapping.

        map global prompt <c-g> '<a-semicolon>:nop should NOT be added to history<ret>'
        map global prompt <c-h> '<a-semicolon>:nop should be added to history first'

Then type

        :nop should be added to history second<c-g><c-h><ret><ret>

the inner command run by `<c-g>` should not be added to history because
it only existed in a noninteractive context.

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -1993,7 +1993,7 @@ void context_wrap(const ParametersParser& parser, Context& context, StringView d
                                        Context::Flags::Draft};
             Context& c = input_handler.context();
 
-            ScopedSetBool disable_history(c.history_disabled());
+            ScopedSetBool noninteractive(c.noninteractive());
 
             func(parser, c);
         };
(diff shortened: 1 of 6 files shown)


mawww:
Merged, along with a follow up refactoring in e3122ab2c19248b590cb0143610439ec050766b7

PR #4697, opened 2022-08-01
Document prompt history registers

krobelus:
So far they have only been talked about in source code
(HistoryRegister), not in documentation.

Let's give them an official name so users can find them better;
"prompt history register" seems better than "history register" since
it's easier to grok. OTOH, in future other registers (like @) could
grow into history registers, so I'm not sure.

State that the history registers are only changed by _interactive_
prompts; that's not quite true yet for some edge cases but it will be with
https://github.com/mawww/kakoune/pull/4695

### DIFF
--- a/doc/pages/registers.asciidoc
+++ b/doc/pages/registers.asciidoc
@@ -34,7 +34,9 @@ All normal-mode commands using a register default to a specific one if not speci
     default search / regex register, used by:
     */*, *<a-/>*, *?*, *<a-?>*, *n*, *<a-n>*, *N*, *<a-N>*, ***, *<a-*>*,
     *s*, *S*, *<a-k>* and *<a-K>*
-    (see <<keys#searching, `:doc keys searching`>>)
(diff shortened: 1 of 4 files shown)


doc/pages/registers.asciidoc
```diff
@@ -34,7 +34,9 @@ All normal-mode commands using a register default to a specific one if not speci
     default search / regex register, used by:
     */*, *<a-/>*, *?*, *<a-?>*, *n*, *<a-n>*, *N*, *<a-N>*, ***, *<a-*>*,
     *s*, *S*, *<a-k>* and *<a-K>*
-    (see <<keys#searching, `:doc keys searching`>>)
+    (see <<keys#searching, `:doc keys searching`>>).
+    This is a prompt history register, holding all commands entered at an
```
Screwtapello, line 38: If we're going to document this, we might as well be specific

```suggestion
    This is a prompt history register, holding the last 100 commands entered at an
```

PR #4699, opened 2022-08-03
Add desktop entry

### LINKED ISSUE #4698: [REQUEST] Desktop entry file.
### Feature

I would like to add a desktop entry file.  See https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-1.1.html.

### Usecase

This would allow listing of kakoune in application menus in a standard way across desktop environments.

7h3w1zz:
Adds a desktop entry and icons.

Closes #4698 

Currently, does not actually install the files, trying to figure out how best to fit them into the makefile.


7h3w1zz:
I've taken the liberty of creating a simple desktop entry file and icons (the fixed version of which is a bit messy, as it's just the current logo scaled down to a size it was not intended to be).

At the moment, and unless/until there's a positive reception for this, I have just added the files to the project, they are not actually installed by `make install` yet.

share/applications/kakoune.desktop
```diff
@@ -0,0 +1,13 @@
+[Desktop Entry]
+Type=Application
+Version=
+Name=Kakoune
+GenericName=Text Editor
+Comment=Edit text files
+Icon=kakoune
+TryExec=kak
+Exec=kak %F
```
Screwtapello, line 9: You should use `--` to prevent Kakoune from interpreting filenames as command-line options.
```suggestion
Exec=kak -- %F
```

share/applications/kakoune.desktop
```diff
@@ -0,0 +1,13 @@
+[Desktop Entry]
+Type=Application
+Version=
```
Screwtapello, line 3: The spec says:

> Version of the Desktop Entry Specification that the desktop entry conforms with. Entries that confirm with this version of the specification should use `1.0`. Note that the version field is not required to be present.

So this key should probably set to `1.0`, or removed entirely, rather than being left blank.

Screwtapello:
> Currently, does not actually install the files, trying to figure out how best to fit them into the makefile.

It should probably go in the `installdirs`/`install`/`uninstall` targets at the bottom of `src/Makefile`.

7h3w1zz:
@Screwtapello I've fixed the oversights you pointed out (thanks!).  I'll probably have a go at the makefile tomorrow.

PR #4707, opened 2022-08-15
Add Face names to json-ui

mspielberg:
Currently each Face communicated to the json-ui carries only appearance information, leaving semantic ambiguity due the inevitable overloading of the limited number of color combinations available, e.g. PrimaryCursor vs. SecondaryCursor. This PR adds the original face name to the representation of each Face and Atom, and outputs them to the json-ui. This allows for more intelligent presentation by a front-end, especially when language-specific syntax highlighters utilize additional faces. This has a slight increase in memory footprint, kept to a minimum through interning of the face names.

### DIFF
--- a/doc/json_ui.asciidoc
+++ b/doc/json_ui.asciidoc
@@ -16,7 +16,7 @@ Here are the data structures used:
 
 * Color: a string, either a named color, or #rrggbb, or 'default'
 * Attribute: one of {underline, reverse, blink, bold, dim, italic, final_fg, final_bg, final_attr}
-* Face { Color fg; Color bg; Array<Attribute> attributes; }
+* Face { String name; Color fg; Color bg; Array<Attribute> attributes; }
 * Atom { Face face; String contents; }
 * Line : Array of Atom
 * Coord { int line; int column }
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -2411,7 +2411,7 @@ static Completions complete_face(const Context& context, CompletionFlags flags,
 {
     return {0_byte, cursor_pos,
             complete(prefix, cursor_pos, context.faces().flatten_faces() |
-                     transform([](auto& entry) -> const String& { return entry.key; }))};
+                     transform([](auto& entry) -> StringView { return entry.key; }))};
 }
 
 static String face_doc_helper(const Context& context, CommandParameters params)
--- a/src/face.hh
+++ b/src/face.hh
@@ -3,7 +3,7 @@
 
 #include "color.hh"
 #include "flags.hh"
-#include "string.hh"
+#include "shared_string.hh"
 
 namespace Kakoune
 {
@@ -33,7 +33,30 @@ struct Face
     Color bg = Color::Default;
     Attribute attributes = Attribute::Normal;
     Color underline = Color::Default;
-    String name = {};
+    StringDataPtr name = {};
+
+    Face() { }
+
+    Face(const Color& fg, const Color& bg)
+    : fg(fg), bg(bg)
+    {
+    }
+
+
+    Face(const Color& fg, const Color& bg, const Attribute& attributes)
+    : fg(fg), bg(bg), attributes(attributes)
+    {
+    }
+
+    Face(const Color& fg, const Color& bg, const Attribute& attributes, const Color& underline, const StringDataPtr name)
+    : fg(fg), bg(bg), attributes(attributes), underline(underline), name(name)
+    {
+    }
+
+    Face(const Color& fg, const Color& bg, const Attribute& attributes, const Color& underline, const StringView& name)
+    : fg(fg), bg(bg), attributes(attributes), underline(underline), name(name.empty() ? StringDataPtr{} : intern(name))
+    {
+    }
 
     friend constexpr bool operator==(const Face& lhs, const Face& rhs)
     {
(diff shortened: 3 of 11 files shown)


Screwtapello:
Looking at the changes to the tests, it looks like they're mostly adding `"name": ""` to each Face record in the JSON output, and that doesn't seem very... helpful to a front-end. If you actually want to extract *semantic* information rather than just visual representation, it seems like you'd need to collect a list of all the faces applied to each span, not just the final result.

mspielberg:
> Looking at the changes to the tests, it looks like they're mostly adding `"name": ""` to each Face record in the JSON output, and that doesn't seem very... helpful to a front-end. If you actually want to extract _semantic_ information rather than just visual representation, it seems like you'd need to collect a list of all the faces applied to each span, not just the final result.

I located a bug causing the standard semantic faces (e.g. variable, string, keyword) to not show up properly in the input. Have a look at the revised version of e.g. [this test](https://github.com/mawww/kakoune/pull/4707/files#diff-5228f172be90e331e6109d3012005c056dc4dee3ad4ba3fe2f8deea67909aeb6R2) and you'll see a lot more useful information present.

I agree it would be even nicer to get access to the full set of faces applied to a given atom, but for performance reasons I wanted to keep the size increase of DisplayAtom to a minimum, as well as keeping it a fixed-size structure. Keeping the last named face seems to be me be a reasonable tradeoff between memory footprint and utility.

Screwtapello:
Kakoune has the idea of a "face" (as used in commands like `:set-face` and `:debug faces`), and the JSON-UI has the idea of a "face", but although they have the same name and a bunch of the same fields, they're not really the same thing. The `:set-face` face is like eggs or flour, and the JSON-UI face is like a cake - it's incorrect to say that a cake is "eggs" even though a cake does *contain* eggs, and eggs may very well have been the most recently added ingredient.

So, adding a "name" field to the JSON-UI "Face" object is misleading in two directions:

- if the Face object includes a non-empty name such as "Menu" (or whatever), the front-end might mistakenly try to render that span as some kind of menu, even though the effective foreground/background/attributes are completely different from those defined in the "Menu" face, just because the span was assembled differently
- if the Face object includes an empty name (as many of the test-cases in this PR still do), the front-end might mistakenly assume that span is *not* a menu (or whatever) even though the foreground/background/attributes exactly match the Menu face, just because the span was assembled differently

Those limitations themselves are not necessarily fatal flaws, however, depending on the use-case. If I were writing a front-end that absolutely had to know which characters on-screen represented cursors (perhaps for some accessibility reason), I wouldn't trust this PR to provide sufficiently reliable information. On the other hand, if you wanted to collect statistics about what phrases are considered "keywords" in typical editing sessions over the course of a month, and unreliable data was not a problem as long as it was consistently unreliable, then sure, it might be fine.

What actual concrete problem are you trying to solve?

mspielberg:
I don't think the "eggs in cake" analogy is a particularly good one, because a cake batter is blended into a homogeneous mixture without any sense of ordering. Instead, Kakoune's faces are combined in a stable, predictable structure. In general, the layer added on-top is the most specific information that is most relevant to a user, as it overrides formatting applied by previous layers. A better analogy might be a "pizza," where a given bite might include (bottom-to-top) crust, sauce, cheese and pepperoni, but it's the pepperoni that has the final, critical influence on how that bite is perceived.

In particular, I have found that cursor and selection information is completely reliable, as it is added last. On the other hand, the only way a face with an empty name can appear in the JSON-UI output is if that span has *no* named faces applied to it anywhere throughout the processing flow. If you look at the test cases, these are identifiers and punctuation that have no matching highlighter pattern.

As for the specific problem I am trying to solve:

It is difficult to distinguish actual buffer content in a line from information that logically is in the "gutter," such as line numbers, and output from `flag-lines` highlighters such as from [`kak-lsp` diagnostics](https://github.com/kak-lsp/kak-lsp/blob/v13.0.0/rc/lsp.kak#L1934-L1936), [`git blame` information](https://github.com/mawww/kakoune/blob/master/rc/tools/git.kak#L122), and so on. In particular, when these elements are present display columns no longer match up with buffer columns.

So far in my experimental frontend I have been able to address this by comparing the cursor position as given in the status line and `selections_desc`, which are given in units relative to the buffer, and the position of a DisplayAtom with color information matching the PrimaryCursor or PrimaryCursorEol face. This is unfortunately not robust since depending on the theme it may not be possible to unambiguously find the cursor by color. Being able to directly identify the PrimaryCursor within a line, or better yet directly identify "gutter" content with the Information and/or LineNumber faces would be a significant improvement.

mawww:
I think the use case you describe is something worth supporting, but I do not think this is the approach I would want to go with.

We could easily know what part of the display buffer if the "gutter" area because we actually compute the number of column used by "widgets" in the `widget_columns` field of `DisplaySetup`. We could pass that to the UI as part of the `draw` json-rpc call.

Regarding the cursor postion or other semantic information, this could be provided as a set of flags on the DisplayAtom, not on the Face.

That said, I would not be too keen to do such changes without understanding what you actually want to do in your frontend with this information.

mspielberg:
The issue I've been working around is diagnostic information from kak-lsp being incorrectly displayed in the main text area of my frontend (a VSCode extension). For my current purposes, knowing the number of display columns dedicated to widgets would be sufficient.

PR #4710, opened 2022-08-16
Correct autoinfo of write commands

krobelus:


src/commands.cc
```diff
@@ -460,11 +460,24 @@ const CommandDesc force_edit_cmd = {
...
+    {
+        { "sync", { false, "force the synchronization of the file onto the filesystem" } },
+        { "method", { true, "explicit writemethod (replace|overwrite)" } },
+        { "force", { false, "Allow overwriting existing file with explicit filename" } }
+    },
+    ParameterDesc::Flags::SwitchesOnlyAtStart, 0, 1
+};
+
+template<>
+const ParameterDesc write_params<false>{
```
mawww, line 477: I dont see much value in making those specializations of a common template, they can just be two separate globals, for example write_params and write_no_force_params. (Not fully satisfied with those names, but I think this is still better than introducing a template global and a boolean parameter).

krobelus: yeah that was a remnant of me trying to share most parameter descriptions, fixed

Issue #4685, opened 2022-07-23
[BUG] Syntax highlighting is slow and inefficient.


amano-kenji:
### Version of Kakoune

v2021.11.08

### Reproducer

* Try to open any file with 20,000 or more lines.
* Try multi selection editing in the file.

### Outcome

For files that have 20,000 lines or more, syntax highlighting is slow.

The file size is still less than 1 megabytes with a bit more than 20,000 lines.

Multi selection editing is also slow due to syntax highlighting. Perhaps, syntax highlighting is updated as I type each letter in editing mode.

neovim doesn't seem to apply syntax highlighting to the whole file. It seems to apply syntax highlighting to just visible parts. But, neovim's syntax highlighting is buggy because it applies syntax highlighting to visible parts only.

3D engines are efficient because they render only what they have to. Recent 3D engines even tweak level of detail dynamically to make rendering more efficient.

### Expectations

Syntax highlighting should be efficient.

mawww:
Would you have an example source file to reproduce this on ? Kakoune does only apply highlighting to the currently displayed part of the buffer and tries to aggressively cache data to avoid reparsing. The only highlighting logic that uses the full buffer is region handling but this should only parse the whole buffer once, then only apply to modified lines between highlighting updates.

amano-kenji:
[example.journal.txt](https://github.com/mawww/kakoune/files/9175121/example.journal.txt)

This is an example ledger journal file.

* Try to select `Yeti` with `sYeti[Earth]` and edit it.

With ledger filetype, doing this takes very long. With plain filetype, doing this is very fast.

mawww:
Took a look, this seems to be an issue in `ledger.kak` that defines highly costly highlighting (it seems to define regions that match every single character of the buffer)

asb:
I'm seeing the same issue with a very large Markdown file (I keep all my notes in one big file). The ~8.5MiB file takes > 10s to display in Kakoune, though displays instantly if I rename it to .txt.

mawww:
@asb would you have an example file you could provide for the markdown case ?

asb:
It looks like there's nothing in particular special about my notes file. The following script generates a ~5MiB file that takes >5 seconds to open with Kakoune.

```sh
#!/bin/sh

for i in $(seq 10000); do
  cat <<EOF >> big_markdown.md
# Title
## Subtitle
Some notes about things.

[...]

PR #4716, opened 2022-08-16
Allow empty region end in region highlighter

krobelus:
Some of our highlighters abuse the regions highlighter by using one
of these patterns as region end:

	'\K'
	'.\K'
	(?!\\).(?=>)|\n
	(?!\\).(?=\))|\s
	\b

These highlighters should be regex highlighters (because they have
no children) but are probably required to be region highlighters to
compete with other region highlighters.

A region highlighter works by first computing *all* matches of both
region start and region end (which faciliates caching).  Unfortunately
this means that these oddball region-end-patterns create loads of
matches, most of which are useless.

Let's standardize these hacks by allowing the region end to be empty.
This allows us to match the entire region in the region begin pattern,
and do no extra work if the region end is empty.

It's still hacky (the region highlighter feels like the wrong tool
for some of these cases) but I don't think we have a good alternative
today.

On the [5MB Markdown file] this gives a decent speedup for the initial
highlighting pass:

	$ HOME=$PWD hyperfine -w 1 'git checkout HEAD'{~,}' -- :/rc/filetype/markdown.kak && ./kak.opt big_markdown.md -e "hook global NormalIdle .* quit" -ui dummy'
	Benchmark 1: git checkout HEAD~ -- :/rc/filetype/markdown.kak && ./kak.opt big_markdown.md -e "hook global NormalIdle .* quit" -ui dummy
	  Time (mean ± σ):      1.969 s ±  0.033 s    [User: 1.935 s, System: 0.035 s]
	  Range (min … max):    1.906 s …  2.036 s    10 runs
	 
	Benchmark 2: git checkout HEAD -- :/rc/filetype/markdown.kak && ./kak.opt big_markdown.md -e "hook global NormalIdle .* quit" -ui dummy
	  Time (mean ± σ):      1.111 s ±  0.022 s    [User: 1.093 s, System: 0.020 s]
	  Range (min … max):    1.079 s …  1.145 s    10 runs
	 
	Summary
	  'git checkout HEAD -- :/rc/filetype/markdown.kak && ./kak.opt big_markdown.md -e "hook global NormalIdle .* quit" -ui dummy' ran
	    1.77 ± 0.05 times faster than 'git checkout HEAD~ -- :/rc/filetype/markdown.kak && ./kak.opt big_markdown.md -e "hook global NormalIdle .* quit" -ui dummy'

(The numbers depend on a other optimizations I'm using.)

---

Regarding markdown.kak :

To avoid the empty-region-end hack I considered using regex
highlighters (instead of region highlighters) for Markdown
URLs.  However, we probably want to keep highlighting bare
URLs (without enclosing <>).  This violates the [CommonMark
spec](https://spec.commonmark.org/0.29/#example-607) but both GitHub
Markdown and editors like VSCode and Emacs highlight bare URLs here.

The problem with highlighting bare URLs is that
they may contain Markdown metacharacters, see
https://spec.commonmark.org/dingus/?text=https%3A%2F%2Fexample.com%2F(1)_2_(3)

So unless we keep using a region highlighter, it will be very difficult
to avoid spurious Markdown highlighting in bare URLs.

So for the above performance improvement we need the empty-region-end hack.
This means we need to match the entire URL in the region-start pattern.
This is slightly challenging because a URL can contain balanced
parentheses but we must not match a trailing ")".
CommonMark [mandates](https://spec.commonmark.org/0.30/#link-destination)
support for at least 3 levels of nested parentheses, so let's just
unroll the regex to implement that.

---

[5MB Markdown file]: https://github.com/mawww/kakoune/issues/4685#issuecomment-1208129806

### DIFF
--- a/doc/pages/highlighters.asciidoc
+++ b/doc/pages/highlighters.asciidoc
@@ -224,6 +224,11 @@ add-highlighter <path>/<name>/<region_name> region \
     A highlighter type, and associated params, as they would be passed
     to `add-highlighter` if they were not applied as a region.
 
+A region spans from the beginning of an *opening* match to the end of the
+next *closing* match that does not overlap with the *opening* match. Hence if
+*closing* is the empty regex, the region will be congruent with the *opening*
+match. If there is no *closing* match the region extends until buffer end.
+
 If the *-match-capture* switch is passed, then region *closing* and *recurse*
 regex matches are considered valid for a given region opening match only if they
 matched the same content for the capture 1 in the *opening* regex.
--- a/rc/filetype/crystal.kak
+++ b/rc/filetype/crystal.kak
@@ -86,7 +86,7 @@ add-highlighter shared/crystal/quoted-symbol region ':"' '(?<!\\)(\\\\)*"' fill 
 # https://crystal-lang.org/reference/syntax_and_semantics/literals/regex.html#modifiers
 add-highlighter shared/crystal/regex region '/' '(?<!\\)(\\\\)*/[imx]*' regions
 # Avoid unterminated regular expression
-add-highlighter shared/crystal/division region ' / ' '.\K' group
+add-highlighter shared/crystal/division region ' / .' '' group
 
 # Percent regex literals
 # https://crystal-lang.org/reference/syntax_and_semantics/literals/regex.html#percent-regex-literals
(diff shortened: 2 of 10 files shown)


Screwtapello:
Please also update `doc/pages/highlighters.asciidoc` to describe how this new feature works, and when it's appropriate to use!

krobelus:
OK, added some docs (which are also applicable to the old behavior, since this is really just an optimization)

krobelus:
this still gives a decent speed up of 1.8 (even 2.6 when applied after another regex optimization that finds required substrings).
But adding the special case is a bit ugly (even though it's good that we canonicalize various hacks that use things like `\K` as region end).
I think the better solution is to allow regex highlighters as children of the regions highlighter.

krobelus:
This current patch has a bad performance regression when navigating very long lines. Didn't look into it yet. Probably it messes with cache keys.

PR #4717, opened 2022-08-16
Fix quadratic runtime when updating region highlighter matches

krobelus:


src/highlighters.cc
```diff
@@ -2210,7 +2210,7 @@ struct RegionsHighlighter : public Highlighter
         m_regexes.insert({key, Regex{str, flags}});
     }
 
-    void add_matches(const Buffer& buffer, LineRange range, Cache& cache) const
+    void add_matches(const Buffer& buffer, const LineRangeList& ranges, Cache& cache) const
```
mawww, line 2213: Did you consider avoiding having to build up that temporary LineRangeList by builiding and passing the Matchers externally ? add_matches would then just contain the loop at line 2229 and we would not need to refactor line_modifications. (We could introduce a Matchers struct that would own the Vector<Matcher> and could provide a method to do the inplace_merge, add_matches could actually also be a method of that struct).

Not sure if I missed something here, but if we can avoid building a temporary vector and refactoring line_modifications I'd prefer that.

krobelus: yeah I want to avoid the vector by using coroutines with [this commit](https://github.com/krobelus/kakoune/commit/d2922398c212f01852891fa5806d39048601af66 on top but that made CI fail for clang<14), so I'm waiting with that commit

a domain-specific solution with a custom struct sounds possible as well, I'll check later

krobelus: added the BatchMatchAdder to avoid the temporary vector

mawww: Thanks ! Sorry for how long this took to review.

PR #4725, opened 2022-09-01
Allow to undo and redo selection changes

### LINKED ISSUE #898: Undo selection change
Hi,

It often happens to me that I carefully craft a selection with multiple cursors, ready to make changes elegantly, only to completely mess it up by pressing a wrong key (by merging the cursors for example). Being able to undo the last selection change (even if only until the previous buffer change) would make this much less painful.

krobelus:
Current status: there are some open questions for the future UI but basic usage
works well and will hopefully not need change (except maybe the key binding).

From the issue:

> It often happens to me that I carefully craft a selection with multiple
> cursors, ready to make changes elegantly, only to completely mess it
> up by pressing a wrong key (by merging the cursors for example). Being
> able to undo the last selection change (even if only until the previous
> buffer change) would make this much less painful.

Fix this by recording selection changes and allowing simple linear
undo/redo of selection changes.

The preliminary key bindings are `<c-h>` and `<c-k>`.
Here are some other vacant normal mode keys I considered

	X Y
	<backspace> <minus>
	# ^ =
	<plus> '

unfortunately none of them is super convenient to type.  Maybe we
can kick out some other normal mode command?

---

This feature has some overlap with the jump list (`<c-o>`/`<c-i>`) and
with undo (u) but each of the three features have their moment.

Currently there's no special integration with either peer feature;
the three histories are completely independent.  In future we might
want to synchronize them so we can implement Sublime Text's "Soft
undo" feature.

Note that it is possible to restore selections that predate a buffer
modification. Depending on the buffer modification, the selections
might look different of course. (When trying to apply an old buffer's
selection to the new buffer, Kakoune computes a diff of the buffers
and updates the selection accordingly. This works quite well for
many practical examples.)

This makes us record the full history of all selections for each
client. This seems wasteful, we could set a limit. I don't expect
excessive memory usage in practice (we also keep the full history of
buffer changes) but I could be wrong.

Closes #898

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -26,7 +26,7 @@ Buffer& Context::buffer() const
 {
     if (not has_buffer())
         throw runtime_error("no buffer in context");
-    return const_cast<Buffer&>((*m_selections).buffer());
+    return const_cast<Buffer&>(selections(false).buffer());
 }
 
 Window& Context::window() const
@@ -223,24 +223,23 @@ Buffer* Context::last_buffer() const
     return previous_buffer != jump_list.rend() ? &previous_buffer->buffer() : nullptr;
 }
 
-SelectionList& Context::selections()
+SelectionList& Context::selections(bool update)
 {
     if (not m_selections)
         throw runtime_error("no selections in context");
-    (*m_selections).update();
+    if (update)
+        (*m_selections).update();
     return *m_selections;
 }
 
 SelectionList& Context::selections_write_only()
 {
-    if (not m_selections)
-        throw runtime_error("no selections in context");
-    return *m_selections;
+    return selections(false);
 }
 
-const SelectionList& Context::selections() const
+const SelectionList& Context::selections(bool update) const
 {
-    return const_cast<Context&>(*this).selections();
+    return const_cast<Context&>(*this).selections(update);
 }
 
 Vector<String> Context::selections_content() const
@@ -277,7 +276,7 @@ void Context::end_edition()
 
 StringView Context::main_sel_register_value(StringView reg) const
 {
-    size_t index = m_selections ? (*m_selections).main_index() : 0;
+    size_t index = has_buffer() ? selections(false).main_index() : 0;
     return RegisterManager::instance()[reg].get_main(*this, index);
 }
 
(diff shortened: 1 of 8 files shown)


src/normal.cc
```diff
@@ -553,7 +559,8 @@ void pipe(Context& context, NormalParams params)
         prompt, {}, default_command, context.faces()["Prompt"],
         PromptFlags::DropHistoryEntriesWithBlankPrefix, '|',
         shell_complete,
-        [default_command](StringView cmdline, PromptEvent event, Context& context)
+        [default_command, selection_edition=std::make_shared<ScopedSelectionEdition>(context)]
```
mawww, line 562: ~That make_shared looks suspicious, I guess that means we end-up copying this lambda around, is that something we can avoid ?~

Looked it up, this is due to std::function requiring a copyiable target, isn't it... Well maybe its going to be time to replace that with our own, or to use std::move_only_function once we migrate to C++23. I guess for now we can live with the shared_ptr here.

src/context.cc
```diff
@@ -164,7 +164,147 @@ void JumpList::forget_buffer(Buffer& buffer)
...
+    auto select_parent = [&, parent] {
+        HistoryId before_undo = m_history_id;
+        m_history_id = parent;
+        current_history_node().redo_child = before_undo;
+        m_staging = current_history_node();
+    };
+    if (&history_node(parent).selections.buffer() == &m_context.buffer())
+        select_parent();
+    else
+        m_context.change_buffer(history_node(parent).selections.buffer(), { std::move(select_parent) });
```
mawww, line 241: Was there an issue with letting change_buffer move the saved selections and overwrite them here ? Its unclear to me why we need to pass that lambda to change_buffer.

krobelus: `Context::change_buffer` calls `Client::change_buffer`, which creates a new selection in the destination buffer and then runs the `WinDisplay` hooks.
The original problem was that it would crash inside some WinDisplay hook because of broken invariants in the selection undo-stack.
I think those problems only occurred in a previous version, as I can't reproduce a problem anymore. Will try some more later.
Still, there is a behavior difference that is probably correct:

If we overwrite selections later, then the WinDisplay hooks will be run with whatever selection is returned from `client_manager.get_free_window()`. By passing the lambda we make sure the selection inside the hook is the same that the user will see. Though the lambda is awkward, would be nice to simplify this somehow.

src/normal.cc
krobelus: yeah the commit message tries to mention that but didn't give enough detail.
We could backport std::move_only_function (or perhaps add a template parameter to Prompt but that's probably not worth it)

PR #4731, opened 2022-09-10
Fix crash when trying to display the menu in a tiny window

### LINKED ISSUE #4728: [BUG][CRASH] Crash when typing “:” in a window that has only one line
### Version of Kakoune

v2021.11.08-409-g1a8f379a

### Reproducer

Steps to reproduce
- open kakoune with `kak -n`
- resize the terminal window so that it only contains one line
- type `:`

### Outcome

Kakoune crashes with this message:
```
Received SIGFPE, exiting.
Pid: 43840
Callstack:
kak(_ZN7Kakoune9BacktraceC1Ev+0x18) [0x5654eec3cf38]
kak(_ZN7Kakoune14signal_handlerEi+0x6f) [0x5654eed45fdf]
/usr/lib/libc.so.6(+0x38a40) [0x7fc35d451a40]
kak(_ZN7Kakoune10TerminalUI9draw_menuEv+0x38b) [0x5654eee1d4bb]
kak(_ZN7Kakoune10TerminalUI9menu_showENS_9ArrayViewIKNS_11DisplayLineEEENS_12DisplayCoordENS_4FaceES6_NS_9MenuStyleE+0x28d) [0x5654eee21c6d]
kak(_ZN7Kakoune6Client10redraw_ifnEv+0x55a) [0x5654eec5904a]
kak(_ZNK7Kakoune13ClientManager14redraw_clientsEv+0x24) [0x5654eec5f784]
kak(_ZN7Kakoune10run_serverENS_10StringViewES0_S0_NS_8OptionalINS_11BufferCoordEEENS_11ServerFlagsENS_6UITypeENS_10DebugFlagsENS_9ArrayViewIKS0_EE+0x6fc) [0x5654eed585fc]
kak(main+0x161d) [0x5654eec3a97d]
/usr/lib/libc.so.6(+0x232d0) [0x7fc35d43c2d0]
/usr/lib/libc.so.6(__libc_start_main+0x8a) [0x7fc35d43c38a]
kak(_start+0x25) [0x5654eec3b2f5]
```

### Expectations

It should not crash, although who would open a 1-line-height editor and then try to see the menu?

### Additional information

* Arch Linux
* master build
* it happened in gnome-terminal, with and without tmux.

arrufat:
Fixes #4728

### DIFF
--- a/src/terminal_ui.cc
+++ b/src/terminal_ui.cc
@@ -1073,6 +1073,9 @@ void TerminalUI::menu_show(ConstArrayView<DisplayLine> items,
     const LineCount height = is_search ?
         1 : (min<LineCount>(max_height, div_round_up(item_count, m_menu.columns)));
 
+    if (height == 0)
+        return;
+
     const ColumnCount maxlen = (m_menu.columns > 1 and item_count > 1) ?
         max_width / m_menu.columns - 1 : max_width;
 


src/terminal_ui.cc
```diff
@@ -968,7 +968,7 @@ void TerminalUI::draw_menu()
 {
     // menu show may have not created the window if it did not fit.
     // so be tolerant.
-    if (not m_menu)
+    if (not m_menu or m_menu.size.line == 0)
```
mawww, line 971: This does not look right to me, that condition would mean that m_menu.size.line is 0 but we still have allocated  the m_menu.lines arry, I think we should fix the logic that might lead to this inconsistency instead.

mawww (changes requested):

src/terminal_ui.cc
arrufat: I see, thank you for the feedback. After looking a bit, it seems that maybe the `m_menu.create()` should not create a menu if it won't fit in the window?
I will keep checking for a right solution :)

mawww:
Thanks.

arrufat:
That should do it: avoid doing anything as soon as we realize the height will be 0.

PR #4737, opened 2022-09-14
Provide context from DisplaySetup in json-ui draw message

mspielberg:
Spawned from discussion in #4707.

This patch embeds context to help front-ends establish the relationship between the text content in the draw message and the originating buffer. In particular: the position in buffer coordinates of the top-left corner of the content of the draw message, the position of the cursor within the drawn content, and the number of widget_columns (columns at the start of each line in the drawn content that were generated by flag-lines highlighters, including the built-in number-lines highlighter).


mawww (commented):
Looking back at the code, we already provide the cursor position to the UI with the set_cursor RPC call that also notifies which cursor (buffer or prompt) is currently active. From the discussion it seems you only need to expose widget_columns for your use case.

We have a bunch of potential extra data we could want to provide the UI and while they mostly are in DisplaySetup at the moment I do not think we should expose DisplaySetup itself but introduce a new struct specifically for this pupose (as we can see here some data from DisplaySetup is not really relevant to the UI). I would prefer not to add it as a field in DisplayBuffer but just pass it to the draw(...) call as an extra parameter if that is not too inconvenient to do.

mspielberg:
Unless I am missing something (not at all out of the question), DisplaySetup is created in Window::update_display_buffer() and goes out of scope at the end of this function, whose only return value is DisplayBuffer. Do you envision this new struct as a member of Window that would be updated as a side effect of update_display_buffer()?

mawww:
Finally back from travelling around, if you still require that data I think for now we could expose it directly in DisplayBuffer as an extra field (widget_columns). We might get to refactor that eventually. I realize from your comment that it would be a bit weird to pull DisplayBuffer and whatever extra data we generate in Window::update_display_buffer separately.

I would like to avoid exposing all of DisplaySetup so as it seems all you need is the widget_columns we could start with exposing just that.

Issue #4747, opened 2022-10-06
[BUG] Normal mode C key has wrong description in the doc


vsyl:
### Version of Kakoune

v2021.11.08

### Reproducer

`:doc keys multiple-selections`

### Outcome

Description of `C` and `<a-C>` doesn't match the actual behavior.

The current description is:
```
*C*::
    copy the main selection to the next line
```

### Expectations

Description of `C` and `<a-C>` should probably be the same as in `README.asciidoc` or `src/normal.cc`

### Additional information

Description from `README.asciidoc`:
```
`C` copies the current selection to the next line (or lines if a count is given)
```

Description from `src/normal.cc`:
```
    { {'C'}, {"copy selection on next lines", copy_selections_on_next_lines<Forward>} },
```

Note that those two does not match either (but are equivalent).

vsyl:
Creating related PR...

PR #4748, opened 2022-10-06
Correct and unify descriptions of normal mode C key behavior

vsyl:
Fixes https://github.com/mawww/kakoune/issues/4747

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -635,8 +635,8 @@ use `n` followed by `,`, in order to remove a selection, use `<a-,>`.
 contains a match for this regex. Using `<a-K>` you can keep the selections
 not containing a match.
 
-`C` copies the current selection to the next line (or lines if a count is given)
-`<a-C>` does the same to previous lines.
+`C` duplicates selections on the lines that follow them.
+`<a-C>` does the same but on the preceding lines.
 
 `$` allows you to enter a shell command and pipe each selection to it.
 Selections whose shell command returns 0 will be kept, other will be dropped.
--- a/doc/pages/keys.asciidoc
+++ b/doc/pages/keys.asciidoc
@@ -570,10 +570,10 @@ to skim through the jump list using:
     select first and last characters of each selection
 
 *C*::
-    copy the main selection to the next line
+    duplicate selections on the lines that follow them
 
 *<a-C>*::
-    copy the main selection to the previous line
+    duplicate selections on the lines that precede them
 
 *,*::
     clear selections to only keep the main one
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -2364,8 +2364,8 @@ static constexpr HashMap<Key, NormalCmd, MemoryDomain::Undefined, KeymapBackend>
 
     { {'_'}, {"trim selections", trim_selections} },
 
-    { {'C'}, {"copy selection on next lines", copy_selections_on_next_lines<Forward>} },
-    { {alt('C')}, {"copy selection on previous lines", copy_selections_on_next_lines<Backward>} },
+    { {'C'}, {"duplicate selections on the lines that follow them.", copy_selections_on_next_lines<Forward>} },
+    { {alt('C')}, {"duplicate selections on the lines that precede them.", copy_selections_on_next_lines<Backward>} },
 
     { {Key::Space}, {"user mappings", exec_user_mappings} },
 


Screwtapello:
If we're trying to be correct and specific, maybe the description should be something like:

```
`C`: for each selection, also select the same region of the next non-blank line
`<a-C>`: the same, but on the previous non-blank line
```

"copy the selection" sounds more like what `y` does than `C`. "copy the selection region" is a little better, but to be really specific you'd have to say something like "copy the region of the selection (not the content)" and that's just cumbersome.

vsyl:
I just noticed the behavior is even more subtle than that
if you have buffer like this:
```
some text | lets say cursor is here
some text |
some text | the selection will copy here
```
And your cursor is in the first line after the `|` then `C` will copy the selection to the third line, so saying that it copies the region to the next `non-blank line` won't be 100% accurate either.

How about a completely different change:

I agree that in the current state, with the "copy" word being used, people might misunderstood it as something that has to do with yanking and pasting, how about using "duplicate" here, so something like: "duplicates selections to the lines that follow them"?

Right now  "duplicate" is used to describe <+> which does something in some sense similar to `C` but "in-place".

doc/pages/keys.asciidoc
```diff
@@ -570,10 +570,10 @@ to skim through the jump list using:
     select first and last characters of each selection
 
 *C*::
-    copy the main selection to the next line
+    duplicates selections on the lines that follow them
```
caksoylar, line 573: This doc uses imperative mood (?, I am not a native speaker) so it should be "duplicate selections" etc. Same for the docstrings in `normal.cc`

README.asciidoc
```diff
@@ -635,8 +635,8 @@ use `n` followed by `,`, in order to remove a selection, use `<a-,>`.
 contains a match for this regex. Using `<a-K>` you can keep the selections
 not containing a match.
 
-`C` copies the current selection to the next line (or lines if a count is given)
-`<a-C>` does the same to previous lines.
+`C` duplicates selections on the lines that follow them.
+`<a-C>` does the same but on the preceding lines.
```
krobelus, line 639: nice, I never realized it worked for multiple selections on the same line.
Actually it's even more complicated than that. Looking at the source code it's something like roughly

```
`C`: for each selection that spans n lines, consider the following chunks of n lines each. Add a new selection on the first *count* chunks where the first/last line of the chunk are long enough to accomodate the column of the selection start/end
`<a-C>`: the same, but on preceding lines
```

which actually doesn't seem right. Below patch has the behavior I expect, what do you think?

It simplifies the description, which is a good sign

```
`C`: for each selection, add *count* selections with the same height starting at the next line where the tentative start/end lines are long enough to accomodate the column of the selection start/end.
`<a-C>`: the same, but on preceding lines
```

whether it's a good idea to go into such detail in the README is another question.

```patch
Author: Johannes Altmanninger <aclopte@gmail.com>
Date:   Sun Oct 9 03:29:17 2022 +0200

    Allow multiline C/<a-C> to match at offsets that are not multiples of the selection height
    
    Select the first <> pair in below file and press C. Nothing happens
    because Kakoune only tries create selections at lines that are
    multiples of 3 (the selection height). Fix this by considering every
    line, as long as it doesn't overlap with another selection we create.
    
             <
            A
            >
    
             <
            B
            >

diff --git a/src/normal.cc b/src/normal.cc
index 1608e77e4..93f3cce99 100644
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -1456,10 +1456,9 @@ void copy_selections_on_next_lines(Context& context, NormalParams params)
         const LineCount height = std::max(anchor.line, cursor.line) - std::min(anchor.line, cursor.line) + 1;
         const size_t max_lines = std::max(params.count, 1);
 
+        LineCount offset = direction * height;
         for (size_t i = 0, nb_sels = 0; nb_sels < max_lines; ++i)
         {
-            LineCount offset = direction * (i + 1) * height;
-
             const LineCount anchor_line = anchor.line + offset;
             const LineCount cursor_line = cursor.line + offset;
 
@@ -1478,8 +1477,10 @@ void copy_selections_on_next_lines(Context& context, NormalParams params)
                 result.emplace_back(BufferCoord{anchor_line, anchor_byte},
                                     BufferCoordAndTarget{cursor_line, cursor_byte, cursor.target});
 
+                offset += direction * height;
                 nb_sels++;
-            }
+            } else
+                offset += direction;
         }
     }
     selections.set(std::move(result), main_index);
```

vsyl: In the old behavior `CC` does the same as `2C`, I'm not sure if this is true for this patch, however probably it would be easy to fix.

krobelus: > In the old behavior `CC` does the same as `2C`,

should still be the case with this patch

PR #4757, opened 2022-10-16
rc modeline: fix error trying to write readonly global variable

krobelus:
When running modeline-parse on this file:

	# kakoune: filetype=ledger:indentwidth=4

I get this error from dash (and a similar one from bash):

	sh: 53: readonly: key: is read only

This is because the readonly variable "key" is used elsewhere, both
times as global. Fix this by making both variables local. While
at it, remove an unused variable.

### DIFF
--- a/rc/detection/modeline.kak
+++ b/rc/detection/modeline.kak
@@ -16,9 +16,8 @@ define-command -hidden modeline-parse-impl %{
 
         # Translate a vim option into the corresponding kakoune one
         translate_opt_vim() {
-            key="$1"
-            value="$2"
-            tr=""
+            local key="$1"
+            local value="$2"
 
             case "${key}" in
                 so|scrolloff)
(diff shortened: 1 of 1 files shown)


Screwtapello:
It turns out that [the `readonly` attribute is part of POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_23) but although `local` is in [the list of reserved command names](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_09_01_01) it is not guaranteed to be implemented, although it's present in `dash` and `mksh` and `posh` and even `busybox sh`, never mind the bigger and more complete implementations like bash and zsh.

krobelus:
no strong opinion on whether to use `local readonly` or plain old globals.

Screwtapello (approved):
On IRC, mawww said he prefers "practically POSIX" over "technically POSIX", so this looks good to me.

Issue #2760, opened 2019-03-03
Using a non-balanced { in a mapping or execute-keys inside a hook


aidam38:
While setting up my LaTeX hooks, I encountered a weird problem, where the `{` and `}` characters inside a snippet interfere with the `{` and `}` characters used for expansions in kakoune configuration. It only happens when the mapping is inside a hook. Here, I provide a Minimal Working Example:
```
hook global WinSetOption filetype=latex %{ 
    map global user e 'i\{'
}
```
The error I get is
```
kakrc_test:parse error: 0:41: unterminated string '%{...}'
Error: 1:1: 'source' parse error: 0:41: unterminated string '%{...}'
```

eraserhd:
The `%{ ... }` quoting syntax, which you are using here, requires inside braces to be balanced.  You could use `%( ... )` or `%< ... >` instead, or you can add a comment at the end of the map line like so:

```
hook global WinSetOption filetype=latex %{ 
    map global user e 'i\{' # } <-- balance braces
}
```

Kakoune's configuration language is much like shell, although a bit saner because of the ability to nest.  Kakoune doesn't know the inside brace is quoted at the time it's collecting the hook commands because it doesn't know yet that it is looking at more configuration language.

andreyorst:
> or you can add a comment at the end of the map line like so

Another workaround:
```
hook global WinSetOption filetype=latex %{ evaluate-commands %sh{
    opener='{'; closer='}'
    printf "%s\n" "map global user e 'i${opener}'"
}}
```

mawww:
You can also use non nestable punctionation characters, which can be escaped inside the string by doubling them up:
```
hook global WinSetOption filetype=latex %~ evaluate-commands %sh`
    printf "%s\n" "map global user e 'i{'"
`~
# or (but a bit confusing)
hook global WinSetOption filetype=latex %~ evaluate-commands %sh~~
    printf "%s\n" "map global user e 'i{'"
~~~
```

alexherbo2:
How about a doubling syntax?

``` kak
evalutate-commands %{{
  echo '{'
}}
```

eraserhd:
On Tue, Mar 12, 2019 at 10:23 PM Alex Leferry 2> <notifications@github.com>
wrote:

> How about a doubling syntax?
>
> evalutate-commands %{{
>   echo '{'
> }}
>
>
That would prevent quoting arguments with surrounding braces.

andreyorst:
[...]

PR #4763, opened 2022-10-20
Balanced strings cannot be escaped

Screwtapello:
Originally the page said:

> No other escaping takes place in balanced strings.

...but in the course of a general readability rewrite (56287da) this was
changed to:

> Characters may be escaped in the same manner as those for quoted strings.

...which is not actually true; probably this change made it in because there
were a lot of good changes in those commits and we didn't read them all closely
enough.

Now the documentation is correct again, and I've added some examples covering
the problems people occasionally ask about (see #2760).

Also, as a bonus, I added a link from the Quoted Strings section to Balanced Strings, just to be clear about how far down "below" was.

### DIFF
--- a/doc/pages/command-parsing.asciidoc
+++ b/doc/pages/command-parsing.asciidoc
@@ -51,33 +51,7 @@ delimiter matches that of its opening delimiter (respectively, `)`, `]`,
 `}`, and `>`).
 
 There is no way to escape the opening and closing characters, even if they
-are nested inside some other kind of string. For example, this will **not**
-work, because the `{` in the map command interferes with the delimiters for the
-hook's command block:
-
-----
-# DOES NOT WORK
-hook global WinSetOption filetype=latex %{
-    map window user b 'o\begin{'
-}
-----
-
-You can solve this by using a different outer delimiter:
-
-----
-hook global WinSetOption filetype=latex %[
-    # Note different delimiter used -----^
-    map window user b o\begin{
-]
-----
-
-...or just by including matching delimiters inside comments:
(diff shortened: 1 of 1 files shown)


doc/pages/command-parsing.asciidoc
```diff
@@ -49,7 +50,34 @@ of `(`, `[`, `{` and `<`), it is parsed as a balanced string whose closing
...
-Characters may be escaped in the same manner as those for quoted strings.
+There is no way to escape the opening and closing characters, even if they
+are nested inside some other kind of string. For example, this will **not**
+work, because the `{` in the map command interferes with the delimiters for the
+hook's command block:
+
+----
+# DOES NOT WORK
+hook global WinSetOption filetype=latex %{
+    map window user b 'o\begin{'
```
krobelus, line 61: Looks fine.

It is a bit weird to place the more complex examples before the simple ones, not sure how to best fix that.

mawww: agreed, I'd rather add simple not working examples in the Examples sections such as:

```
- %{ foo\{ } is an unterminated balanced string wheras %[ bar\{ ] is terminated
```

I would not mention the comment trick in here as I do not think its something that needs to be encouraged by the docs (I am not against using that trick, but putting it in the docs feels like an endorsement)

Screwtapello: I actually forgot there was a "Balanced String Examples" section right there. I've updated the PR to include examples more in line (in style and location) with the existing ones.

PR #4765, opened 2022-10-26
fix highlighting for undetected config files

nonumeros:
consider the following

```
# Includes files with directives to load dynamic modules.
include /etc/nginx/modules/*.conf

# Uncomment to include files with config snippets into the root context.
```

anything after and including `*`, except `#` where applicable,, is not commented out.

### DIFF
--- a/rc/filetype/conf.kak
+++ b/rc/filetype/conf.kak
@@ -1,4 +1,4 @@
-hook global BufCreate .+\.(conf|repo|cfg|properties|desktop) %{
+hook global BufCreate .+\.(repo|cfg|properties|desktop) %{
     set-option buffer filetype conf
 }
 
--- a/rc/filetype/conf.kak
+++ b/rc/filetype/conf.kak
@@ -1,4 +1,4 @@
-hook global BufCreate .+\.(repo|cfg|properties|desktop) %{
+hook global BufCreate .+\.(conf|repo|cfg|properties|desktop) %{
     set-option buffer filetype conf
 }
 
--- a/rc/filetype/nginx.kak
+++ b/rc/filetype/nginx.kak
@@ -1,0 +1,30 @@
+hook global BufCreate .*\.(conf) %{
+    set-option buffer filetype nginx
+}
+
+hook global WinCreate .+\.ini %{
+    try %{
+        execute-keys /^\h*#<ret>
+        set-option buffer filetype nginx
+    }
+}
+
+hook global WinSetOption filetype=nginx %{
+    require-module nginx
+}
+
+hook -group nginx-highlight global WinSetOption filetype=nginx %{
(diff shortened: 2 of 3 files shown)


rc/filetype/conf.kak
```diff
@@ -1,4 +1,4 @@
-hook global BufCreate .+\.(repo|cfg|properties|desktop) %{
+hook global BufCreate .+\.(conf|repo|cfg|properties|desktop) %{
```
mawww, line 1: This will conflict with various other highlighters, such as the resolv.conf one in etc.kak or the exherbo ones, we would need some way to give priority to one or the other.

nonumeros: @mawww  is there a use case in particular for say the exherbo files needed for config or the resolve.conf in etc.kak? Does ti have anything to do with the mime encoding at the time it gets written? I just don't see it. And if I don't see it, it becomes a problem.  Is there another outstanding issue related to this that I can read about?

nonumeros: @mawww  undone. 

And I went ahead and added nginx.kak. Although I agree with you in that if by any chance there's a problem, conflict with whatever other files from the list, nginx however, needs highlighting no less than the rest. Or so I think.

src/commands.cc
```diff
@@ -2553,9 +2553,7 @@ const CommandDesc change_directory_cmd = {
              return { 0_byte, cursor_pos,
                       complete_filename(prefix,
                                         context.options()["ignored_files"].get<Regex>(),
-                                        cursor_pos, FilenameFlags::OnlyDirectories),
-                      Completions::Flags::Menu };
-        }),
+                                        cursor_pos, FilenameFlags::OnlyDirectories)};}),
```
nonumeros, line 2556: @mawww  and this one here I don't like it. Do you want to hear my honest opinion? (how the saying goes).  We'd be better off by wait it out I think. But since when has become an impediment to use the <backspace> freely? Heck. The completions have to work and have to take me to the right dir no matter how many times I press the backspace key or any other key for that matter.

krobelus: this is an unrelated change, don't include it here.
I have a different fix prepared, will submit it soon.

rc/filetype/nginx.kak
```diff
@@ -0,0 +1,30 @@
+hook global BufCreate .*\.(conf) %{
+    set-option buffer filetype nginx
```
krobelus, line 2: this has the same problem of false positives.
Probably better to restrict this to files under /etc/nginx/

krobelus (changes requested):

rc/filetype/nginx.kak
nonumeros: that's what I thought after mawww's comment with the other conf files.

PR #4766, opened 2022-10-27
Add Discord server link to Kakoune page

natsukagami:
The Discord server has been alive and active for a while and so I think
it should be given some spotlight on the homepage as well.

The invite link is taken from the original post (https://discuss.kakoune.com/t/i-made-a-kakoune-focused-discord-server/1058),
but is set up as a permanent invite link.


raiguard:
The discord is not official, so I don't think it should be on the official website.

krobelus:
> The discord is not official, so I don't think it should be on the official website.

How do we define "official"?

raiguard:
I just noticed that the subreddit is there, so I suppose the Discord server wouldn't be out of place.
I was thinking "official" as in the things Mawww uses (which AFAIK is community.kakoune.com and IRC).

natsukagami:
I had some debate about this as well. But the Discord server has been getting a slow but steady influx of users even though the only public point of entry was a link on the community forum.
I think this is a good sign that there is demand for one. And given that both the subreddit and the forum (which iirc was not created by mawww?) are linked to from the main page, I think the Discord server deserves a spot too :P

mawww:
Personally I'd rather not rely on non-free software for the community but I also want to take a hands off approach to it (because I do not have the time nor the interest in managing a community myself). I happen to use discord for video games so I did join the Kakoune server recently.

My main hesitation here is that I'd rather encourage the use of the IRC channel (or even the Matrix one) than the Discord one, but I guess its up to users to decide where they want to go.

krobelus:
Occasionally, users have reported kak-lsp bugs on the various channels (IRC/Discord/Matrix) instead of opening an issue. I guess because the barrier of chatting is lower and also it may not be obvious that an odd behavior is a bug.
My point is that we don't wanna lose this signal but amplify it (as long as there's not too much noise).

So I think it's useful to advertise a chat channel that has a low barrier of entry. I'm sure many people are put off by IRC not providing offline messaging.
TIL there is a Matrix channel, we should advertise it. The Helix one has a whopping 2k members.

PR #4777, opened 2022-11-12
Switch undo storage from a tree to a plain list

occivink:
See commit message for details.
I've not tested too extensively, but I figured I would put a PR out already so that others can test if they wish. From my brief experimentation it seems like good change.

I was wondering if it might not be interesting to optimize the 'reverse' node. Instead of computing ^C at the moment that we commit a new group, we could have nodes be some kind of union between a full node with the full content, or a 'reference' node with just an node index + a boolean flag (since a reference might be a ^C, or a ^^C). The inverse would be computed on demand, which I think would save some memory.
edit: after reading a bit about the storage of the undo modification, `StringDataPtr` looks to be reference counted and so there should be no memory duplication in case significant buffer modifications (for example pasting large text). 
There might still be some benefit in using the approach described above, but not as large I thought.

~~If this is accepted, I will adapt the documentation later.~~


TeddyDD:
I've been running this branch locally for two weeks. The linear history is great. Less mental overhead when undoing.

arrufat:
If this gets merged, maybe we can pick up the newly freed <a-u> and <a-U> for undoing and redoing selection changes (https://github.com/mawww/kakoune/pull/4725). What do you think @krobelus?

mawww:
I'll run with this merged for a while to get a feel for it.

krobelus:
I've implemented the equivalent behavior for selection undo and the jump list.
If anyone wants to try, it's in 03bfbe0333db139b68cc04f66f75d5f6d04b0dc8 where I've also included this PR.

krobelus:
Ok for selection undo this behavior is probably not right because I often want to do
1. jump to A
2. jump to B
3. jump to C
4. use `<c-o>` to jump back to B
5. jump to D
6. use `<c-o>` to jump back to B
7. use `<c-o>` to jump back to A

with the list behavior, step 7 jumps to D first; so it would required two extra `<c-o>`.
For selection history it's important to make excursions cheap. My new approach to navigate the undo tree is in https://github.com/mawww/kakoune/pull/4816 (still experimental).

Not sure if this applies to content undo. I find that tree navigation more intuitive than current `<c-u>`.. will see which one is easy to use in practice

mawww:
Heads up: I've been running with this for a while, and I think I'll merge it soon.

Issue #4764, opened 2022-10-26
[QUESTION] highlighting after * comments out the remainder code


nonumeros:
## Steps to reproduce

```
# comment 

include /etc/module/*.conf
```

With the above in place , and  regardless of the filetype, results in the remainder code block/s to be highlighted as comments instead of code 

So anything after `*` is treated as comments' sections

In this particular case this is a conf file but it makes no difference what the file extension happens to be, It could be any other such as d, kak, c 

### system 

Alpine v3.16

nonumeros:
I'm not familiar with Alpine. I installed ncurses, ncuses-devel and a few other packages (can't remember which ones now) that presumably handle the terminfo databases, but… I can't get it to work. 

I also checked and double checked the config file and the settings match the kak settings I have been running on other systems like Slackware, Arch, Void 

If anyone has any idea/suggestion and whether I'm missing a package with Alpine that handles this issue with kakoune, please let me know.

krobelus:
Probably your buffer is assigned filetype cpp. What's the output of `:echo %opt{filetype}`?
If no filetype is set based on the file extension, then `file.kak`will set the filetype based on `file -b --mime-type -L "${kak_buffile}"`

If that's true, we should fix file/libmagic because this is surely not a C++ file

As a workaround I disable file detection for some files

```
hook global BufCreate .*/todo %{
	set-option buffer disabled_hooks "%opt{disabled_hooks}|file-detection"
}
```

Alternatively, if there's a well-known extension you can add it to `rc/filetype/conf.kak` but it's impossible to match all Unix configuration files.

nonumeros:
> Probably your buffer is assigned filetype cpp. What's the output of `:echo %opt{filetype}`? If no filetype is set based on the file extension, then `file.kak`will set the filetype based on `file -b --mime-type -L "${kak_buffile}"`
> 

I owe you that one. Let me get back to you later on with the output. I'm not on that system now


> If that's true, we should fix file/libmagic because this is surely not a C++ file
> 
> As a workaround I disable file detection for some files
> 
> ```
> hook global BufCreate .*/todo %{
> 	set-option buffer disabled_hooks "%opt{disabled_hooks}|file-detection"
> }
> ```

thanks. 

> Alternatively, if there's a well-known extension you can add it to `rc/filetype/conf.kak` but it's impossible to match all Unix configuration files.

[...]

Issue #2147, opened 2018-06-23
kak always writes files with a newline at the end


shachaf:
```
$ printf test > file
$ xxd file
00000000: 7465 7374                                test
$ kak -e 'wq' file
$ xxd file
00000000: 7465 7374 0a                             test.
```

Opening a file and writing immediately shouldn't modify the file.

Vim keeps track of whether a file ended with a newline on opening with an "endofline" flag. Kakoune could presumably do the same thing.

mawww:
Kakoune is a code editor, not an arbitrary data editor. Adding an option to control that would be relatively simple, but I think it would go a bit against Kakoune's principles. Unix text files are expected to end with an end of line character, its usually an error when they dont.

The motivating example is not strong enough to convince me we need to support that. Kakoune does not guarantee that opening a file and writing it should be a no-op (should it touch the file modification date ? should it change the inode number as suggested in another issue ? I dont think Kakoune should make any promises about any of those).

That said, you could argue that BOM and end of line format support are already precedents in your requested direction, which is true, but I added them because I had to work on source files where those *needed* to be preserved. I am not sure we have real source code where the last line needs not to end with an EOL.

shachaf:
There's plenty of source code that doesn't end with \n, though. I'm not sure which editors do that -- and they probably shouldn't -- but I run into it regularly. The files should probably be fixed but that's a separate issue.

More generally, even though this isn't primarily a code feature, I'd like to be confident that if I open a file and make no changes, or only make changes in a specific area, that the rest of the file doesn't get changed unexpectedly. Also note that writing empty files isn't supported right now (though looking at vim's handling for empty files, it seems like a special case anyway).

It's also a useful feature of vim that you can edit or examine binary data with `:%!xxd` and `:%!xxd -r`, though this is obviously not the primary use case.

This seems like a small feature -- and the only change necessary? -- to be able to open and modify an arbitrary file without causing unexpected changes.

tmccombs:
fwiw, editorconfig does have an option `insert_final_newline`which enforces either always ending with newlines _or not_.

lenormf:
Related #3669. Seems like once the choices have been made, we'll have to document them.

JJK96:
I often have changed files in git where the only change is a newline at the end as a result of this behaviour.
I would prefer to change this behaviour as suggested by this issue.

lenormf:
[...]

PR #4778, opened 2022-11-15
add nginx.kak for highlighting detection filetype

nonumeros:


rc/filetype/nginx.kak
```diff
@@ -0,0 +1,30 @@
+hook global BufCreate .*\.(nginx) %{
+    set-option buffer filetype nginx
```
krobelus, line 2: I don't think `*.nginx` files are a thing

nonumeros: I couldn't think of anything else here. You'e right. I'll add `params`, remove `nginx` and replace `*` by `+`. I guess that'll do it for now. I guess.

nonumeros: @krobelus  if you were to tell me now that say, the fastcgi_params and uwscgi_params are considered to be configuration files and as such such be included on the `conf.kak` instead, I would disagree. I mean. What would be the point of having those *files* there in the first place?

nonumeros: I'll go over it again @krobelus. On a system with gcc 11.2 works fine, whereas on 12.2 doesn't.

rc/filetype/nginx.kak
```diff
@@ -0,0 +1,30 @@
+hook global BufCreate .*\(_params) %{
```
krobelus, line 1: Thanks but overall I'm not convinced this is worth merging.
It does not add anything, except on systems where `file(1)` is broken.
If there is a good reason for this change, please note it down.

Also you copied over parts of `conf.kak` that are definitely not valid in nginx files.

nonumeros: > Also you copied over parts of `conf.kak` that are definitely not valid in nginx files.

Yes. Perhaps some of it or most of it, is not valid for nginx files, but for the most part,  the few tidbits from it,  matched the syntax needed for nginx files accordingly.  Namely, as outlined on https://github.com/mawww/kakoune/pull/4765 which had as a premise that  "anything after and including *, except # where applicable,, is not commented out" still holds 

> Thanks but overall I'm not convinced this is worth merging. It does not add anything, except on systems where `file(1)` is broken. If there is a good reason for this change, please note it down.
> 

no problem. I don't know what you're referring to here with the `file(1)` part. 

If anything it's clear, kak on some systems, decides to write a file as an `x-c` while on others, it decides it's an `application/octream`. I wouldn't be surprised  if the preprocessor is to be  blamed for this erratic behavior, but that's besides the point, really.

nonumeros: This is the related https://github.com/mawww/kakoune/issues/4764

nonumeros: And this is the prior pull request that resolved the problem. 
https://github.com/mawww/kakoune/pull/4765

```
hook global BufCreate .+\.(conf|repo|cfg|properties|desktop) %{
```

Obviously that was from before the other revisions.

nonumeros: That's not even touching the fact that `hook global BufCreate .*` or `BufCreate .+`  is just for aesthetics really

Let  `nginx.conf` be  a new file  and `:write` it accordingly with kak. What do we have?

`$ xxd nginx.conf | head -1`

results:

`00000000: 0a                                       .`


Now, let emacs or neovim `write` a new `nginx.conf` file , or different `nginx.conf` file

`xxd` returns what is to be expected

nonumeros: > Thanks but overall I'm not convinced this is worth merging.

It is not. It doesn't need to be convincing :) But I've never seen a more contradictory statement for refusing to merge not only this one on its own, but also this one https://github.com/mawww/kakoune/pull/4765  from earlier that just had the addition of `conf` to 

```
hook global BufCreate .+\.(conf|repo|cfg|properties|desktop) %{
```

In the meantime xxd shows anything written with kak as a

```
00000000: 0a                                       .
```

but with the same token, it `can't ` detect syntax highlighting accordingly for any new `nginx.conf` file. That's beyond me.

krobelus: the 0a you are seeing in the xxd output is https://github.com/mawww/kakoune/issues/2147, there are usable workarounds and proposed fixes

PR #4780, opened 2022-11-16
rc git: remove annoying stderr in non-git repository on show/update-diff

xnuk:


rc/tools/git.kak
```diff
@@ -169,7 +169,7 @@ define-command -params 1.. \
     update_diff() {
         (
             cd_bufdir
-            git --no-pager diff --no-ext-diff -U0 "$kak_buffile" | perl -e '
+            git --no-pager diff --no-ext-diff -U0 "$kak_buffile" 2>/dev/null | perl -e '
```
krobelus, line 172: How come are you calling `git` commands?
If it's a hook, maybe don't install the hook outside of git repositories.
Anyway, this change doesn't negatively impact anything, so why not.

xnuk: > If it's a hook, maybe don't install the hook outside of git repositories.

That's what I'm doing:

```kak
hook global BufWritePost .* %{
  try 'git show-diff'
}
```

I didn't want anything more complex than this just for a git gutter.

krobelus: I wonder if we should have a variant of `try` that inhibits any message that would normally go into the debug buffer

PR #4781, opened 2022-11-17
Preserve indentation type when auto-indenting javascript block comments

Ssamm45:
The current implementation uses '&' to auto-indent comments, which always uses spaces. If we use <a-&> it copies the indentation from the previous line instead of just blindly using spaces.

### DIFF
--- a/rc/filetype/javascript.kak
+++ b/rc/filetype/javascript.kak
@@ -99,7 +99,7 @@ define-command -hidden javascript-insert-on-new-line %<
         # trim trailing whitespace on the previous line
         try %[ execute-keys -draft s\h+$<ret> d ]
         # align the new star with the previous one
-        execute-keys Kx1s^[^*]*(\*)<ret>&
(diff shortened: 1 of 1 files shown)


rc/filetype/javascript.kak
```diff
@@ -99,7 +99,7 @@ define-command -hidden javascript-insert-on-new-line %<
         # trim trailing whitespace on the previous line
         try %[ execute-keys -draft s\h+$<ret> d ]
         # align the new star with the previous one
-        execute-keys Kx1s^[^*]*(\*)<ret>&
+        execute-keys Kx1s^[^*]*(\*)<ret><a-(><a-&>
```
krobelus, line 102: Nice find, LGTM. 
To reproduce it's enough to press `o` on a file

		/*

	         */

I wonder how many of these incorrect uses of & are hidden in rc/filetype/.
I personally have all the auto-insert hooks turned off.

krobelus (approved):

PR #4782, opened 2022-11-18
Add support for Tilix terminal

doppioandante:
I can't believe I have been using tilix for all this time without doing this.

I would like a feedback on the commit that patches the `new` command, I did that because an empty `:new` would pass nothing to `kak -e`, making it exit.

### DIFF
--- a/rc/windowing/new-client.kak
+++ b/rc/windowing/new-client.kak
@@ -3,7 +3,15 @@ new [<commands>]: create a new Kakoune client
 The ''terminal'' alias is being used to determine the user''s preferred terminal emulator
 The optional arguments are passed as commands to the new client' \
 %{
-    terminal kak -c %val{session} -e "%arg{@}"
+
+    evaluate-commands %sh{
+        if [ $# -gt 0 ]; then
+            echo "echo -debug '$@'"
+            echo "terminal kak -c %val{session} -e \"$@\""
+        else
+            echo "terminal kak -c %val{session}"
+        fi
+    }
 }
 
 complete-command -menu new command
--- a/rc/windowing/tilix.kak
+++ b/rc/windowing/tilix.kak
@@ -1,0 +1,69 @@
+
+provide-module tilix %{
+
+# ensure that we're running on kitty
+evaluate-commands %sh{
+    [ -z "${kak_opt_windowing_modules}" ] || [ -n "$TILIX_ID" ] || echo 'fail Tilix not detected'
+}
+
+define-command -hidden -params 2.. tilix-cmd %{
+    evaluate-commands %sh{
+        if [ -z "$TILIX_ID" ]; then
+            echo "fail 'This command is only available in a Tilix session'"
+            exit
+        fi
+        tilix "$@" < /dev/null > /dev/null 2>&1 &
+    }
+}
+
+define-command tilix-terminal -params 0.. -docstring '
+tilix-terminal-vertical [<program> <arguments>]: create a new Tilix window
+The program, if provided, will be executed in the new terminal' \
+%{
+    tilix-cmd -a app-new-window -e %arg{@}
+}
+complete-command tilix-terminal shell
+
+define-command tilix-terminal-session -params 0.. -docstring '
+tilix-terminal-vertical [<program> <arguments>]: create a new Tilix session in the current window
+The program, if provided, will be executed in the new terminal' \
+%{
+    tilix-cmd -a app-new-session -e %arg{@}
+}
+complete-command tilix-terminal shell
+
+define-command tilix-terminal-vertical -params 0.. -docstring '
+tilix-terminal-vertical [<program> <arguments>]: split tilix terminal vertically
+The program, if provided, will be executed in the new terminal' \
+%{
+    tilix-cmd -a session-add-down -e %arg{@}
+}
+complete-command tilix-terminal-vertical shell
+
+define-command tilix-terminal-horizontal -params 0.. -docstring '
+tilix-terminal-vertical [<program> <arguments>]: split tilix terminal horizontally
+The program, if provided, will be executed in the new terminal' \
+%{
+    tilix-cmd -a session-add-right -e %arg{@}
+}
+complete-command tilix-terminal-horizontal shell
+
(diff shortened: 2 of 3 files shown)


arrufat:
Oh, I made one for myself a long time ago, to learn how to interface with Kakoune, but I don't use tilix. Have a look and see if there's something you like (I didn't add the focus stuff):

``` kak
hook -group tilix-hooks global KakBegin .* %sh{
    if [ -n "$TILIX_ID" ] && [ -z "$TMUX" ]; then
        echo "require-module tilix"
    fi
}

provide-module tilix %{

declare-option -docstring %{window type that tilix creates on new and repl calls (window|session|right|down|quake)} str tilix_window_type session

define-command -hidden -params 2.. tilix-terminal-impl %{
    nop %sh{
        window_type="$1"
        shift
        # join the arguments as one strings for the shell execution (see iterm.kak)
        args=$(
            for i in "$@"; do
                if [ "$i" = '' ]; then
                    printf "'' "
                else
                    printf %s "$i" | sed -e "s|'|'\\\\''|g; s|^|'|; s|$|' |"
                fi
            done
        )
        if [[ ${window_type} == "window" || ${window_type} == "session" ]]; then
            tilix --action app-new-${window_type} --working-directory "$PWD" -e "${args}"
        elif [[ ${window_type} == "right" || ${window_type} == "down" ]]; then
            tilix --action session-add-${window_type} --working-directory "$PWD" -e "${args}"
        elif [[ ${window_type} == "quake"  ]]; then
            tilix --quake --working-directory "$PWD" -e "$@"
        else
            tilix --action ${window_type} --working-directory "$PWD" -e "${args}"
        fi
    }
}

define-command tilix-terminal -params 1.. -shell-completion -docstring '
tilix-terminal <program> [<arguments>]: create a new terminal with the default window type
The program passed as argument will be executed in the new terminal
See also: tilix_window_type option' \
%{
    tilix-terminal-impl %opt{tilix_window_type} %arg{@}
}

define-command tilix-terminal-right -params 1.. -command-completion -docstring '
tilix-terminal-right <program> [<arguments>]: create a new terminal in a right split
%{
    tilix-terminal-impl right %arg{@}
}

define-command tilix-terminal-down -params 1.. -command-completion -docstring '
tilix-terminal-down <program> [<arguments>]: create a new terminal in a down split
%{
    tilix-terminal-impl down %arg{@}
}

define-command tilix-terminal-quake -params 1.. -command-completion -docstring '
tilix-terminal-quake <program> [<arguments>]: create a new terminal in a quake window
%{
    tilix-terminal-impl quake %arg{@}
}

define-command tilix-client-new -command-completion -docstring '
tilix-client-right [<arguments>]: create a new kakoune client in a right split
See also: tilix_window_type option' \
%{
    tilix-terminal kak -c %val{session}
}

define-command tilix-client-right -command-completion -docstring '
tilix-client-right [<arguments>]: create a new kakoune client in a right split' \
%{
    tilix-terminal-right kak -c %val{session}
}

define-command tilix-client-down -command-completion -docstring '
tilix-client-down [<arguments>]: create a new kakoune client in a right split' \
%{
    tilix-terminal-down kak -c %val{session}
}

define-command tilix-client-quake -params .. -command-completion -docstring '
# tilix-client-quake [<arguments>]: create a new kakoune client in a tilix quake window' \
%{
    tilix-terminal-quake kak -c %val{session}
}

alias global terminal tilix-terminal
alias global terminal-right tilix-terminal-right
alias global terminal-down tilix-terminal-down
alias global terminal-quake tilix-terminal-quake
alias global new tilix-client-new
alias global new-right tilix-client-right
alias global new-down tilix-client-down
alias global new-quake tilix-client-quake
```

doppioandante:
Thanks @arrufat , the focus is actually broken on my system, but maybe that's due to Wayland.
Your script looks interesting, I'll look at it tomorrow to see what I can scavenge.

arrufat:
The names of the splits are confusing to me:
When I do `:tmux-terminal-vertical`, I expect a vertical split (i.e., side by side). But Kakoune interprets is as the final layout (up and down split).

I think the names I used are clearer: `new-right` and `new-down`.

However, after revisiting this, I like `new-below` more, since it has the same number of letters as `new-right` 😅.

Perhaps we could use that naming style upstream? What do you think @mawww @krobelus?

krobelus:
The horizontal/vertical split distinction is indeed ambiguous and confusing, even more so given that i3 uses the exact opposite meaning.
I don't think we should change the *-terminal-{horizontal,vertical} aliases because they just copy the terminology offered by tmux/screen/iterm.
We can expect that a user is already familiar with tmux etc.
Whether we should introduce new aliases, I don't know if it's worth it. I think it comes down to individual preference.

rc/windowing/new-client.kak
```diff
@@ -3,7 +3,15 @@ new [<commands>]: create a new Kakoune client
...
 The optional arguments are passed as commands to the new client' \
 %{
-    terminal kak -c %val{session} -e "%arg{@}"
+
+    evaluate-commands %sh{
+        if [ $# -gt 0 ]; then
+            echo "echo -debug '$@'"
+            echo "terminal kak -c %val{session} -e \"$@\""
+        else
+            echo "terminal kak -c %val{session}"
```
krobelus, line 12: > I did that because an empty `:new` would pass nothing to `kak -e`, making it exit.

`:new` works for me, this must be a problem in your `terminal` implementation.
Instead of

	tilix -a app-new-window -a kak -c session

we probably need something to the effect of

	tilix -a app-new-window -a "kak -c session"

rc/windowing/tilix.kak
```diff
@@ -0,0 +1,69 @@
+
+provide-module tilix %{
+
+# ensure that we're running on kitty
```
jdugan6240, line 4: I'm assuming this is copied from the kitty support script? This should probably be changed to tilix instead.

doppioandante: Indeed. If you wish to make changes feel free to push on my fork; I wish I had time to improve this and merge arrufat's version, but it seems to be hard atm :(

PR #4785, opened 2022-11-19
Fix hare new line insert indentation after comments

stacyharper:
At the moment, inserting a new line while being in a comment result in a `//<indentation>` instead of `<indentation>//`.

To fix this, we just re-ordre both InsertChar hooks.

before
after

### DIFF
--- a/rc/filetype/hare.kak
+++ b/rc/filetype/hare.kak
@@ -118,7 +118,7 @@ provide-module hare %§
             execute-keys -save-regs '' k x s ^\h*\K//\h* <ret> y
             try %{
                 # paste the comment prefix
-                execute-keys x j x s ^\h* <ret>P
+                execute-keys x j x s ^\h* <ret>p
             }
         } }
         try %{


rc/filetype/hare.kak
```diff
@@ -7,8 +7,8 @@ hook global BufCreate .*[.]ha %{
 hook global WinSetOption filetype=hare %{
     require-module hare
     hook window ModeChange pop:insert:.* -group hare-trim-indent hare-trim-indent
-    hook window InsertChar \n -group hare-indent hare-indent-on-new-line
     hook window InsertChar \n -group hare-insert hare-insert-on-new-line
+    hook window InsertChar \n -group hare-indent hare-indent-on-new-line
```
krobelus, line 11: I remember that the hook order was relevant for some other reason, and that we are not consistent about that.
IIRC for the more complex file types it's convenient to use a certain hook order, for others it doesn't matter.

Probably the hare module would be better of staying closer to established filetypes. For example, `go.kak` does not have this problem despite using the same hook order

`hare-insert-on-new-line` includes a snippet that does

	# remove trailing whitespace on the above line

Other filetype highlighter seem to have the same feature but implemented in a different way.
That function could probably just be (untested)

    define-command -hidden hare-insert-on-new-line %{ evaluate-commands -draft -itersel %{
        # copy // comments prefix and following white spaces
        try %{ execute-keys -draft <semicolon><c-s>kx s ^\h*\K/{2,}\h* <ret> y<c-o>P<esc> }
    } }

krobelus (approved):

stacyharper:
I pushed an alternative fix instead of this re-order one.

Basically, replacing `P` with `p` make the comment symbol + indentation
to be paster AFTER the initial new line indentation.

krobelus:
the initial version has been merged with 5314e2375. Fair enough

PR #4786, opened 2022-11-19
Accept "cd dir/" again instead of using a subdirectory

### LINKED ISSUE #4775: [QUESTION]
### Question

I might be wrong here but I think that 69053d96239e5ef1f97522949d8c712fa5846eab actually does the opposite of what was intended.  And rather than saving a tab tap, I ended up having to use another tab tap. 

Although I found the idea behind it nice,  it's problematic 

Let me post a screenshot. Visually speaking, and since there's highlighting involved with the editor,I think I'm better off by explaining it like this. 

If one were to look at the the first screenshot , I was just trying to change to the directory  `Downloads` which in turn has a subdirectory named `lcdf-typetools` 

Screenshot_2022-11-08_02-52-17

Now, normally speaking, or the way I was taught to change to a directory was by not having to press another key 

and when I said it's a misleading behavior, it is really. 

Once the `enter` key is pressed, I expected it to show me the files inside  the `lcdf-typetools` directory but not the files inside the subdirectory `autom4te.cache`  that happens to be the highlighted subdirectory 

Screenshot_2022-11-08_03-10-19

I reverted the line 2555 (I think) from commands.cc and it behaved as I was expecting it. But I don't know if it would be more rational to just correct it  rather than removing a line.

cc: @krobelus

krobelus:
Commit 69053d962 (Use menu behavior when completing change-directory,
2022-07-19) made ":cd dir/" actually run ":cd dir/first-subdir",
which can be surprising.
This is usually irrelevant because you rarely type the trailing slash.
However it does happen after correcting an error with `<backspace>`
and friends. For for example,

	:cd d<tab>/f<backspace>

results in

	:cd dir/

We should probably fix user expectations here. Do this by adding "dir/"
as valid completion.  This requires us to allow empty candidates in
"RankedMatch" but there's no harm in that. This means we need to
filter out empty completions from shell-script-candidates elsewhere.

Alternative fix: we could revert 69053d962. This would remove the
convenient menu behavior but that wouldn't be a huge deal.

Fixes #4775

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -241,7 +241,8 @@ struct ShellScriptCompleter
                                                       ShellManager::Flags::WaitForStdout,
                                                       shell_context).first;
         CandidateList candidates;
-        for (auto&& candidate : output | split<StringView>('\n'))
+        for (auto&& candidate : output | split<StringView>('\n')
+                                       | filter([](auto s) { return not s.empty(); }))
             candidates.push_back(candidate.str());
 
         return {0_byte, pos_in_token, std::move(candidates), m_flags};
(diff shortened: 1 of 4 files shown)


nonumeros:
This should have been merged not since yesterday, or better yet how the figuratively and  literally meanings would go:: since last year, for crying out loud. 

```
kak 
:cd /
:e cat chattr chgrp chmod chown cp date dd df dmesg echo
```

krobelus:
thanks for the review

PR #3724, opened 2020-09-11
Newline eol

occivink:
Fixes #2147

PR #4791, opened 2022-11-19
Don't add missing newline at EOF

### LINKED ISSUE #2147: kak always writes files with a newline at the end

krobelus:
For simplicity of implementation, each Kakoune buffer has at least one
line, and each line has at least one character, the trailing newline
("\n").

When writing a buffer to disk, we include the trailing newlines,
independent of whether they are present in the underpinning file.

There are some [use-cases] where this is undesirable, like when editing
a MOTD file or `<pre>`-formatted text on a web page.  Additionally,
there are plenty of existing [source code files] that are missing
a trailing newline (probably because Emacs creates such files).

I've been using this workaround to avoid unintentional changes:

	hook -always global BufOpenFile .* %{ evaluate-commands %sh{
		if [ $(tail -c1 "$kak_hook_param" | wc -l) = 0 ]; then {
			echo 'hook -always buffer BufWritePost .* %{
				nop %sh{ truncate --size=-1 "$kak_hook_param" }
			}'
		} fi
	} }

The -always switches are necessary when using [find-apply-changes]
which edits and writes files in a draft context with "-no-hooks"
(probably to suppress auto-format hooks).

Obviate the need for this workaround by making Kakoune preserve
missing newlines in the input.
Special care is taken of empty files: they are written back as empty
(without a newline) but if they are no longer empty we do include
the newline.

This patch can be thought of as "damage control".  I believe it does
not make any real case worse because proper termination of files
should already be done at creation time.  We shouldn't try to fix
the user's mistake here. Other costs are
1. Implementation complexity - but it's pretty contained.
2. We are giving the user the pretense that the file has a newline
   even if that won't be written to disk.

The patch makes us match GNU sed but deviate from Vim (which adds
missing newlines).

PR #3724 suggested making this behavior configurable.  We could do
that with either approach.  However, independent of the approach,
this would add some complexity: when reloading a buffer from disk we'd
want to distinguish between explicitly-set and inferred EOL-at-EOF
option to avoid stomping user configuration.
I also think that the configurability is of limited use.
There is even an EditorConfig option but it does not include the
sensible behavior of preserving input, see the [editorconfig-issue].
Overall, I don't think we should make this configurable unless
there is a strong reason. It should be easy to add workarounds
to manually fix up line endings.

Closes #2147
Closes #3724

[use-cases]: https://github.com/mawww/kakoune/issues/2147#issuecomment-1307723864
[source code files]: https://github.com/mawww/kakoune/issues/2147#issuecomment-691042734
[find-apply-changes]: https://github.com/occivink/kakoune-find/
[editorconfig-issue]: https://github.com/editorconfig/editorconfig/issues/475

### DIFF
--- a/doc/pages/expansions.asciidoc
+++ b/doc/pages/expansions.asciidoc
@@ -403,6 +403,13 @@ The following expansions are supported (with required context _in italics_):
     `%val{cursor_column}`), and _text_ is the content of the insertion or
     deletion (see also `%val{history}`)
 
+*%val{eol_at_eof}*::
+   _in buffer scope_ +
+   a bool indicating whether Kakoune will include the last line's newline
+   character when writing the buffer. This is inferred from whether the file
+   had a trailing newline when it was loaded. The default for new files is
+   `true`.
+
 *%val{user_modes}*::
     unquoted list of user modes.
 
--- a/src/main.cc
+++ b/src/main.cc
@@ -400,7 +400,25 @@ static const EnvVarDesc builtin_env_vars[] = { {
         "uncommitted_modifications", false,
         [](StringView name, const Context& context) -> Vector<String>
         { return undo_group_as_strings(context.buffer().current_undo_group()); }
-    }
+    }, {
+        "eol_at_eof", true,
+        [](StringView name, const Context& context) -> Vector<String>
+        { 
+            if (not context.has_buffer())
+                throw runtime_error{"eol_at_eof option is only available in buffer scope"};
+            const auto& buffer = context.buffer();
+            switch (buffer.eol_at_eof()) {
+                case EolAtEof::Present: return {"true"};
+                case EolAtEof::Missing: return {"false"};
+                case EolAtEof::EmptyFile:
+                    if (buffer.line_count() == 1 and buffer[0] == "\n")
+                        return {"false"};
+                    return {"true"};
+            }
+            kak_assert(false);
+            return {"true"};
+        }
+    },
 };
 
 void register_registers()


Screwtapello:
This patch *does* add a new option, it's just not user-accessible. If you don't want it to be writable, could you expose it as `%val{eolateof}` or similar? That would be helpful for debugging "why does file have/omit a newline", or adding to the status line.

krobelus:
> This patch *does* add a new option, it's just not user-accessible. If you don't want it to be writable, could you expose it as `%val{eolateof}` or similar? That would be helpful for debugging "why does file have/omit a newline", or adding to the status line.

A `noeol` indicator in the modeline sounds nice but might be unnecessary noise since it's not always easily possible to fix the endings.

I like to think that it's not necessary to expose this to the user because they can simply check the on-disk file:

	[ $(tail -c1 "$kak_buffile" | wc -l) = 0 ]

Of course this sweeps two edge cases under the rug:
1. empty files have special treatment
2. if the buffer is out of sync with the on-disk file (i.e. the user chose not to reload), then the internal Kakoune option might be stale

but I think neither matters because they will rarely coincide with an external eol-at-eof change, in both cases the user should be aware.

---

I realized that this patch makes anyone who is using the `truncate` workaround truncate too much.
My hope is that those folks already use Git to catch this error.

Emacs and VSCode make it very easy to create files with missing newlines.
So I think taking the pragmatic road will avoid a lot of pain for people collaborating on Git repos. I've seen a mismerge of a file that conflicted due to an eol-at-eof change.

Screwtapello:
Yeah, most of the time it won't matter since most files have a trailing newline, and most of the rest of the time it won't matter because Kakoune will silently do the right thing. Nevertheless, for the remaining tiny percentage of cases, it makes me very uncomfortable for Kakoune to have secret information that determines how it behaves, but which can't be inspected or modified. I see this as the same kind of thing as the `eolformat` or `BOM` options: leave those options alone and Kakoune will do the right thing almost all the time, but if something breaks somewhere then you can quickly and easily discover whether Kakoune had anything to do with it, and why.

krobelus:
Ok I exposed it, which conveniently gives us an obvious place to document the behavior.
I'm not against making it an option but I can't think of a use case where we want to preserve the current  behavior.
editorconfig does foresee such a use case but I think that's only because traditional editors overwrite each other's styles. I think we  can get by with just keeping the existing style. This way we can avoid the complexity of a configurable option.

Screwtapello:
Thanks! One case I can think of for wanting to choose whether a file has a trailing newline is the ambiguous case of adding text to a zero-byte file - is it a trailing-newline file with no lines, or a no-trailing-newline file with zero characters?

That said, that's probably not a major use-case and it may not be worth the implementation complexity to support it, so the current implementation is probably fine.

krobelus:
Right. When adding text to a zero-byte file we always write it with a newline (same as before) because that's what we think is the correct style. Not sure if anyone explicitly wants the other style (for new files)

krobelus:
> I realized that this patch makes anyone who is using the truncate workaround truncate too much.
My hope is that those folks already use Git to catch this error.

It's hard to gage how bad the breakage would be.
I originally started working on this because I had been using the hook for months but it didn't work for `find-apply-changes`. I only realized halfway through that I just need to use `-always`

krobelus:
As one should have expected, this has some issues.
Specifically, when the file is missing a final newline and `:format` would add one, that added newline will not be written by `:write`.
Seems complicated to get right

PR #4792, opened 2022-11-19
website community: add link to Matrix channel (libera.chat IRC bridge)

krobelus:
Kakoune has many users who may not have an IRC bouncer set up. As a
result, they might not have the best experience using IRC.

Let's lower the barrier of entry by linking to the Matrix IRC bridge,
which has a web client with persistent chat history (though for these
bridges, users can only see history from after they joined).

Our peer projects like Neovim and Helix make heavy use of Matrix. They
use proper Matrix channels which may have more features than IRC
bridges.

The icon is fairly random, I didn't find anything resembling the
Matrix logo.

Related: #4766

### DIFF
--- a/index.html
+++ b/index.html
@@ -348,8 +348,8 @@
                                 bővül új funkciókkal, a közreműködők pull request-jei
                                 által is. A felhasználók feltehetik kérdéseiket,
                                 és megoszthatják véleményüket a közösség nagy
-                                részével a <b>#kakoune</b> csatornán az irc.libera.chat
-                                hálózaton.
+                                részével a <a href="https://web.libera.chat/?channels=#kakoune"><b>#kakoune</b></a>
+                                csatornán az irc.libera.chat hálózaton.
                             </span>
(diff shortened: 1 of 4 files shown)


index.html
```diff
@@ -340,24 +345,28 @@ <h2 tid="menu:features">Features</h2>
...
                                 @ irc.libera.chat.
                             </span>
                             <span tid-hu="feature:activeDevelopmentSupportText" class="d-none">
                                 A projekt aktív fejlesztés alatt van, rendszeresen
                                 bővül új funkciókkal, a közreműködők pull request-jei
                                 által is. A felhasználók feltehetik kérdéseiket,
                                 és megoszthatják véleményüket a közösség nagy
-                                részével a <b>#kakoune</b> csatornán az irc.libera.chat
-                                hálózaton.
+                                részével a <a href="https://app.element.io/#/room/#kakoune:libera.chat">#kakoune:libera.chat (Matrix)</a>
```
omasanori, line 358: Ditto.

omasanori (approved):
Thank you for addressing my comment!

PR #4795, opened 2022-11-23
Wrap command in another sh invocation to ensure long PATH works

2xsaiko:
When the PATH is overly long (>=1373 chars at least here), binaries are no longer able to be looked up in PATH for some reason, giving a command not found error, and resulting in this error from osascript due to the new shell immediately exiting:

    execution error: missing value doesn’t understand the “select” message. (-1708)

Wrapping the env invocation in another sh process works around this issue.


Screwtapello:
`env` is normally used to set environment variables and search for an executable in `$PATH` without launching an entire shell. If you're going to launch a shell anyway, you probably don't need `env`. However, you may need an extra round of quoting. It's not clear to me what the different layers of string interpretation are here, so I can't say how many kinds of quoting should be applied in which order.

Screwtapello:
The iTerm docs are frustratingly vague about whether `create window with default profile command "blah"` winds up running `/bin/sh -c "blah"` (and therefore `blah` needs to be shell-quoted) or uses some other method to break the command string into an array of strings (and therefore `blah` needs some other kind of quoting, or cannot be correctly quoted at all).

Screwtapello:
You might also try running `xargs --show-limit < /dev/null` to see what the size limit for environment variables and command-line arguments is on your system.

rc/windowing/iterm.kak
```diff
@@ -29,7 +29,7 @@ define-command -hidden -params 2.. iterm-terminal-impl %{
         escaped=$(do_esc "$args")
         esc_path=$(do_esc "$PATH")
         esc_tmp=$(do_esc "$TMPDIR")
-        cmd="env PATH='${esc_path}' TMPDIR='${esc_tmp}' $escaped"
+        cmd=$(do_esc "/bin/sh -c \"exec env PATH='${esc_path}' TMPDIR='${esc_tmp}' $escaped\"")
```
krobelus, line 32: is this really about PATH or is it the size of the command line or environment?
I'm not sure what's going on here.

Commit 3ae94d017 suggests that iterm launches the user's default shell, hence the use of `env`. Maybe this problem is specific to the default shell and somehow /bin/sh is not affected?

2xsaiko: If I wrap this with any of /bin/sh -c, /bin/zsh -c, /run/current-system/bin/zsh -c (my interactive shell), the issue does not occur, so doesn't look like it passes it verbatim to any shell.

2xsaiko:
> You might also try running `xargs --show-limit < /dev/null` to see what the size limit for environment variables and command-line arguments is on your system.

This is the current output from xargs:

```
Your environment variables take up 2190 bytes
POSIX upper limit on argument length (this system): 1044338
POSIX smallest allowable upper limit on argument length (all systems): 4096
Maximum length of command we could actually use: 1042148
Size of command buffer we are actually using: 131072
Maximum parallelism (--max-procs must be no greater): 2147483647
```

From testing a bit more, looks like the PATH variable gets cut off. This is a heavily modified applescript based on the one which you would get generated from kakoune:

```
tell application "iTerm"
    tell current session of current window
        tell (split vertically with same profile command "/usr/bin/env PATH='/nix/store/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaavjnkjwahdjsgirqvcdz6ryy58-gnutar-1.34/bin:/nix/store/jzfyg4y7avnpcmqwnysfxq7pc9lv1vnh-gzip-1.12/bin:/nix/store/kxillr150qnj38mxwv0gdbmhh8jnq6gw-bzip2-1.0.8-bin/bin:/nix/store/i5wikc9wpr7dvmgrj6cnbzrbrggvk8s5-gnumake-4.3/bin:/nix/store/kv4p5wk0q5h77fr2f8cbwrdlacffha7p-bash-5.1-p16/bin' '/bin/sh'") to select
    end tell
end tell
```

Running it does produce a shell due to absolute paths used for sh but looking at $PATH, it gets cut off at 1020 chars, so that's where the issue lies (notice the final missing letter off 'bin', it always cuts to 1020 regardless of original length of PATH. Maybe something to do with 'PATH' being 4 chars so that adds up to 1024?):

```
sh-3.2$ sh
sh: sh: command not found
sh-3.2$ echo $PATH
/nix/store/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaavjnkjwahdjsgirqvcdz6ryy58-gnutar-1.34/bin:/nix/store/jzfyg4y7avnpcmqwnysfxq7pc9lv1vnh-gzip-1.12/bin:/nix/store/kxillr150qnj38mxwv0gdbmhh8jnq6gw-bzip2-1.0.8-bin/bin:/nix/store/i5wikc9wpr7dvmgrj6cnbzrbrggvk8s5-gnumake-4.3/bin:/nix/store/kv4p5wk0q5h77fr2f8cbwrdlacffha7p-bash-5.1-p16/bi
sh-3.2$ echo $PATH | /usr/bin/wc -c
    1020
```

> `env` is normally used to set environment variables and search for an executable in `$PATH` without launching an entire shell. If you're going to launch a shell anyway, you probably don't need `env`.

Good point, I can get rid of it. Doesn't look like it needs more quoting, I already added one more layer already.

dblsaiko:
I'm no longer using iTerm2, so do with this as you will. However last I checked it was still necessary.

PR #4807, opened 2022-12-03
Integrate selection undo with jump list

krobelus:
**Current status**: still working on some behavior changes

---

- https://github.com/mawww/kakoune/pull/4814

---

The jump commands (`<c-o>` and `<c-i>`) and the selection undo commands
(`<c-h>` and `<c-k>`) are unaware of each other.  Specifically,
this means that `<c-o>` adds an entry to the selection undo list,
so `<c-h>` can be used to undo it.   As reported on IRC, this is
confusing because both `<c-o>` and `<c-h>` are supposed to move
backwards in time, and not cancel each other out.

Make both variants walk the same selection history.  After using
the coarse-grained jump commands to find an approximate location,
this allows to use selection undo commands for fine-grained selection
restoration.


krobelus:
removed the unwarranted behavior change; the behavior should be good now, only cleanups to go

krobelus:
Hmm this makes any movement invalidate `<c-i>` which breaks workflows.
https://github.com/mawww/kakoune/pull/4816 makes it possible to emulate old `<c-i>` with `<c-o><c-j><c-i>`.
That's probably too complicated.
It's easy to add a special case to make `<c-i>` do this by default (if at end).

krobelus:
> Hmm this makes any movement invalidate `<c-i>` which breaks workflows.

ok it should be easy to avoid this behavior change. We could do it separately later.. but it's probably not worth it even though it's a simplification of jump list semantics

krobelus:
This still has some assertion errors, closing for now.
Maybe it's not worth the complexity.

PR #4810, opened 2022-12-17
Fix bugs with <a-t> when run at the beginning of a buffer. 

### LINKED ISSUE #4471: [BUG][CRASH] SIGSEGV after <a-t><ret>
### Version of Kakoune

v2021.11.08-41-g6029ee98

### Reproducer

Run `kak -n`.

Press `%d<a-t><ret>`.

### Outcome

```
Received SIGSEGV, exiting.
Pid: 641976
Callstack:
kak(_ZN7Kakoune9BacktraceC1Ev+0x18) [0x557620eea598]
kak(_ZN7Kakoune14signal_handlerEi+0x6f) [0x557620fef5df]
/usr/lib/libc.so.6(+0x3cda0) [0x7f13407e6da0]
kak(_ZN7Kakoune10get_columnERKNS_6BufferENS_11ColumnCountENS_11BufferCoordE+0x1f) [0x557620efbc4f]
kak(_ZNK7Kakoune6Window21compute_display_setupERKNS_7ContextE+0x1b6) [0x5576210c7fc6]
kak(_ZN7Kakoune6Window21update_display_bufferERKNS_7ContextE+0x26a) [0x5576210c9c0a]
kak(_ZN7Kakoune6Client10redraw_ifnEv+0x1eb) [0x557620f0537b]
kak(_ZNK7Kakoune13ClientManager14redraw_clientsEv+0x24) [0x557620f09bd4]
kak(_ZN7Kakoune10run_serverENS_10StringViewES0_S0_NS_8OptionalINS_11BufferCoordEEENS_11ServerFlagsENS_6UITypeENS_10DebugFlagsENS_9ArrayViewIKS0_EE+0x9a8) [0x557621002888]
kak(main+0x19f9) [0x557620ee78f9]
/usr/lib/libc.so.6(__libc_start_main+0xd5) [0x7f13407d1b25]
kak(_start+0x2e) [0x557620ee82fe]
```

### Expectations

_No response_

### Additional information

_No response_

potatoalienof13:
\<a-t> previously used to select both the first and second characters if it was run on the first character of a buffer and the cursor was on the character being selected to.  It would crash if there was no second character.  This makes it so that <a-t> refuses to run when its on the first character of a buffer.
closes #4471


potatoalienof13:
This fixes issue #4471.

potatoalienof13:
closes #4471

src/selectors.cc
```diff
@@ -420,6 +420,15 @@ select_to_reverse(const Context& context, const Selection& selection,
                   Codepoint c, int count, bool inclusive)
 {
     auto& buffer = context.buffer();
+
+    // if we are selecting backwards from the beginning of the buffer,
+    // there is nothing more that can be selected.
+    // Unless its inclusive, in which its possible that the current
+    // location of the cursor contains the Codepoint being looked for.
+    if(!inclusive && selection.cursor() == buffer.begin()){
```
krobelus, line 428: couple style issues:
- use `not` instead of `!`
- use `and` instead of `&&`
- add space around the if condition
- whe using braces, put them on a separate line

```suggestion
    if not inclusive and selection.cursor() == buffer.begin())
    {
```

also I'd recommend cleaning up the Git history. In this case you could run

	git reset $(git merge-base origin/master HEAD)

and then create a fresh commit

src/selectors.cc
```diff
@@ -420,6 +420,15 @@ select_to_reverse(const Context& context, const Selection& selection,
...
+        return {};
+    }
```
krobelus, line 430: The inclusive motions also have a weird edge case:
If I run `kak` and type `<a-f>*`, it does not throw an error (but it should).
If I do the equivalent somewhere that's not the beginning of the buffer - for example with `iabc<esc><a-f>*` - I get "no selections remaining", which is expected.
To fix this inconsistency, I think we should apply your logic regardless of whether the motion is inclusive.

```diff
diff --git a/src/selectors.cc b/src/selectors.cc
index 3d9b453d2..60ecf07a1 100644
--- a/src/selectors.cc
+++ b/src/selectors.cc
@@ -421,4 +421,8 @@ select_to_reverse(const Context& context, const Selection& selection,
 {
     auto& buffer = context.buffer();
+
+    if (selection.cursor() == buffer.begin())
+        return {};
+
     Utf8Iterator begin{buffer.iterator_at(selection.cursor()), buffer};
     Utf8Iterator end = begin;
```

potatoalienof13: Ah, thought I had checked that.

src/selectors.cc
```diff
@@ -420,6 +420,13 @@ select_to_reverse(const Context& context, const Selection& selection,
                   Codepoint c, int count, bool inclusive)
 {
     auto& buffer = context.buffer();
+
+    // if we are selecting backwards from the beginning of the buffer,
+    // there is nothing more that can be selected.
+    if (selection.cursor() == buffer.begin()){
+        return {};
+    }
```
krobelus, line 428: the opening brace is still off (should be on a separate line), can you maybe remove the braces since it's just one statement

potatoalienof13:
Yeah, this git history is still really ugly.  My punishment for not switching to a different branch beforehand.

krobelus:
> Yeah, this git history is still really ugly. My punishment for not switching to a different branch beforehand.

you can still change it by resetting the branch.
I don't care about the wrong turns you have taken before reaching the goal, I care about the direct path.

krobelus (approved):

krobelus (changes requested):
braces are still wrong

potatoalienof13:
Okay, everything has to be good now, right? Thanks for your help btw.

krobelus:
good job, thanks

Issue #4472, opened 2021-12-13
[BUG][CRASH] kak -f / execute-keys crash given random sequence of keys


kak-0:
### Version of Kakoune

v2021.11.08-41-g6029ee98 (debug=yes)

### Reproducer

`kak -f` crashes when given some random sequences of keys. For example:

```sh
kak -f 'o<esc><u><a-gt>U' </dev/null # Fatal error: assert failed "m_current_undo_group.empty()" at buffer.cc:310
kak -f 'o<esc><u><A>a<a-;>U' </dev/null # Fatal error: assert failed "m_current_undo_group.empty()" at buffer.cc:310
kak -f '2oK.k<c-n><ret><c-n>' </dev/null # Fatal error: assert failed "from >= 0 and from <= str_len" at string.hh:287
kak -f 'o/<c-n><up><c-n>' </dev/null # SIGSEGV
kak -f 'iab:a<c-p><home><c-p>' </dev/null # SIGSEGV
kak -f 'iab<space>a<c-p><a-;><a-h><c-n>' </dev/null # SIGSEGV
kak -f '+<a-)>iab$a<c-p><ret><c-p>' </dev/null # Fatal error: assert failed "from >= 0 and from <= str_len" at string.hh:287
kak -f 'i<tab><tab><esc><a-u>iw<esc><a-u>+@' </dev/null # Fatal error: assert failed "min >= last_min" at selection.cc:285
```

These key sequences (with the exception of the last one) crash `execute-keys` too, but they don't crash Kakoune when pressed interactively.

### Outcome

Kakoune crashes.

### Additional information

Kakoune must be built with `debug=yes` to enable assertions.

sidkshatriya:
Interesting. Just curious -- were these sequences obtained from fuzzing?

kak-0:
> Interesting. Just curious -- were these sequences obtained from fuzzing?

I just generated random key sequences and passed them to `kak -f`. I'm not familiar with dedicated fuzzing tools, but they will probably be able to find more crashes.

[Here is the script I used](https://gist.github.com/kak-0/59055ce212cd883796abd8f792a7a4da) if you're curious, but running it may be unsafe as random inputs can cause unpredictable behavior.

Screwtapello:
[...]

PR #4812, opened 2022-12-21
Fix m_completions not being reset, causing a crash. 

potatoalienof13:
This fixes the crash when kakoune is run as 
```
kak -f '2oK.k<c-n><ret><c-n>' /dev/null
```

This is part of #4472.  I am not absolutely certain that this is the right way to fix this crash.

### DIFF
--- a/src/insert_completer.cc
+++ b/src/insert_completer.cc
@@ -481,7 +481,7 @@ bool InsertCompleter::has_candidate_selected() const
 
 void InsertCompleter::try_accept()
 {
-    if (m_completions.is_valid() and m_context.has_client() and has_candidate_selected())
+    if (m_completions.is_valid() and has_candidate_selected())
         reset();
 }
 


src/insert_completer.cc
```diff
@@ -481,7 +481,7 @@ bool InsertCompleter::has_candidate_selected() const
 
 void InsertCompleter::try_accept()
 {
-    if (m_completions.is_valid() and m_context.has_client() and has_candidate_selected())
+    if (m_completions.is_valid() and has_candidate_selected())
```
krobelus, line 484: > This commit attempts to fix a crash with -f. Specifically when kakoune is run as
> 
> > ./kak -f '2oK.k<c-n><ret><c-n>' /dev/null
> 
> The crash occurs because when <c-n> is pressed for the second time,
> it attempts to use m_completions from the first press of <c-n>.
> This only happens when kakoune is run with -f, because when this is done
> interactively, there is a client, which means that m_completions gets
> reset.  This removes the check that causes that difference.
> 
> I am not *completely* sure that this is the best way to solve the
> problem, since I am not completely sure why that check was put there
> in the first place.

Looks like I added it in 66078243e (Run InsertCompletionHide hook before
insertions that close completion menu, 2022-05-29).
I can tell why I did it this way: I was trying to make the minimal change that would fix the `InsertCompletionHide` hook. That hook only ever needs to run in the client context, so I made sure to only modify those code paths.

So, no particular reason for excluding client-less contexts.
The fix looks great.

I was wondering why the crash does not reproduce interactively (when removing `InsertCompleter::try_accept()`).
This is probably because in interactive mode, completions are quickly updated by the timer (Insert::m_idle_timer), so we don't need to reset them ever.

you could add the test to `test/regression/`

krobelus (approved):

src/insert_completer.cc
potatoalienof13: I looked at the test infrastructure, but I could not find any way to run to test kakoune with `-f`.

potatoalienof13: I looked at the test infrastructure, but I could not find any way to run to test kakoune with `-f`.

krobelus: > I looked at the test infrastructure, but I could not find any way to run to test kakoune with -f.

I see. We could add a special case to `test/run` but it may not be worth the added complexity.
We generally don't need to test `-f` separately because it's supposed to behave just like interactive mode.
So I think it's fine.

PR #4813, opened 2022-12-22
Allow users to influence sorting of user-supplied completions

### LINKED ISSUE #1709: Remove sorting of user-supplied insert completions

krobelus:
- https://github.com/mawww/kakoune/pull/4790
- https://github.com/mawww/kakoune/pull/4786
- https://github.com/mawww/kakoune/pull/4653

---

This adds an optional "priority" field to completion candidates, which affects
fuzzy match scoring. Example usage is in https://github.com/kak-lsp/kak-lsp/pull/657.
I haven't tested this much so the scoring math might not be perfect.

May close #1709


krobelus:
This feature is quite helpful when working with chatty language servers like rust-analzer for two scenarios:

1. When hovering over a symbol that is missing an import statement, we can use `lsp-code-actions` to bring up a `prompt -menu` to pick a code action.
   Sometimes there are loads of code actions but 80% of the time I want to run the "Add missing imports" one.
   By allowing to specify the order, the import fix will be first hence easier to run.
2. Sometimes there are loads of insert-mode completions, typically after typing `someobject.`
   Some of the completions are not even specific to the object and usually not interesting for discovery.

Both cases affect mostly scenarios where the query is empty.
I wonder if this simple change is good enough

	if (query.empty())
		// use user-supplied ranking
	else
		// use fuzzy ranking

The formula in this PR tries to interpolate between the two cases, for a more natural feeling.

---

The other question is the UI. The extra priority field is ugly because it's not obvious how it's interpreted.

I wonder if we can find a safe and elegant way to allow users to pass their own ranked-match comparator.
It is a function of two completion candidates and a query (or the precomputed fuzzy-match score).
Maybe a mini-DSL. It would need to support table lookups for cases like rust-analyzer where each completion is assigned a priority.

PR #4815, opened 2022-12-22
Fix Buffer::advance out of bounds access.

potatoalienof13:
This commit fixes a bug in Buffer::advance where it would first access m_lines[-1], and then check whether or not that access would have segfaulted.  This commit moves the check to before the segfault would occur.

### DIFF
--- a/src/buffer.cc
+++ b/src/buffer.cc
@@ -575,9 +575,9 @@ BufferCoord Buffer::advance(BufferCoord coord, ByteCount count) const
         count += coord.column;
         while (count < 0)
         {
-            count += m_lines[--line].length();
-            if (line < 0)
+            if (--line < 0)
                 return {0, 0};
+            count += m_lines[line].length();
         }
         return { line, count };
     }


src/buffer.cc
```diff
@@ -575,9 +575,9 @@ BufferCoord Buffer::advance(BufferCoord coord, ByteCount count) const
         count += coord.column;
         while (count < 0)
         {
-            count += m_lines[--line].length();
-            if (line < 0)
+            if (--line < 0)
                 return {0, 0};
+            count += m_lines[line].length();
```
krobelus, line 580: nice find - `m_lines[-1]` is undefined behavior, so we should put the bounds check first.
The symmetry with the above case (with ++ instead of --) is deceptive - that one is fine but this one isn't.

Do you have a command that reproduces the crash?

potatoalienof13: The command that caused this crash was 
```
./kak -f 'o/<c-n><up><c-n>' </dev/null
```
I was going to post it, but I completely forgot to do that.

potatoalienof13: That only works with -f btw.

krobelus (approved):

PR #4816, opened 2022-12-26
Allow to redo old branches from selection history

krobelus:
**Current status**: still testing if the UI is useful in practice

- https://github.com/mawww/kakoune/pull/4807

---

Today, selection undo and jump list commands only travel the latest branch
of the selection history. If I undo a selection change and then make another
change, the undone change cannot be redone - it lives on an old branch of
the selection history.  Let's make it possible to revisit such selections
by adding a shortcut to toggle between redo branches whenever there is more
than one.  Hopefully this is not too hard to use.


krobelus:
the main motivation for this was to make the initial version of https://github.com/mawww/kakoune/pull/4807 nicer to use. It's not really relevant for the current approach.

PR #4824, opened 2023-01-08
Implement bracketed paste

krobelus:
Text pasted into Kakoune's normal mode is interpreted as command
sequence, which is probably never what the user wants. Text
pasted during insert mode will be inserted fine but may trigger
auto-indentation hooks which is likely not what users want.

Bracketed paste is pair of escape codes sent by terminals to
applications so that applications can distinguish between pasted text
and typed text.

Let's use this feature to always insert pasted text verbatim,
suppressing any hooks and keymaps.

We need to make a decision on whether to paste before or after
the selection. I chose "before" because that's what I'm used to.
Maybe we should change this.

Another decision is how to treat linewise pastes, should those be
inserted at line boundaries? This implementation inserts at the cursor
instead which is the most straightforward to implement.

Closes #2465

### DIFF
--- a/src/client.cc
+++ b/src/client.cc
@@ -60,6 +60,9 @@ Client::Client(std::unique_ptr<UserInterface>&& ui,
         else
             m_pending_keys.push_back(key);
     });
+    m_ui->set_on_paste([this](StringView content) {
+        context().input_handler().paste(content);
+    });
 
     m_window->hooks().run_hook(Hook::WinDisplay, m_window->buffer().name(), context());
 
--- a/src/input_handler.cc
+++ b/src/input_handler.cc
@@ -77,13 +77,17 @@ void InputMode::paste(StringView content)
     try
     {
         Buffer& buffer = context().buffer();
+        const bool linewise = not content.empty() and content.back() == '\n';
         ScopedEdition edition{context()};
         ScopedSelectionEdition selection_edition{context()};
-        context().selections().for_each([&buffer, content=std::move(content)]
+        context().selections().for_each([&buffer, content=std::move(content), linewise]
                                         (size_t index, Selection& sel) {
-            BufferRange range = buffer.insert(sel.min(), content);
-            sel.min() = range.begin;
-            sel.max() = range.end > range.begin ? buffer.char_prev(range.end) : range.begin;
+            auto& min = sel.min();
+            auto& max = sel.max();
+            BufferRange range =
+                buffer.insert(paste_pos(buffer, min, max, PasteMode::Insert, linewise), content);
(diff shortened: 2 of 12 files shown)


caksoylar:
This is working great in my testing, thanks @krobelus !

nonumeros:
@caksoylar  great news. 

sorry I didn't test these changes but with d8883d47  (unless I'm doing something off), is this working ??

mawww:
The feature is nice, but I am not really keen on the proposed approach. I think we should expand the UI protocol so that the UI can request an insert with an arbitrary string. Sending this as a sequence of keys seems wrong, and leads to overly complex code.

I suggest adding a `paste(String content)` callback to UserInterface the same way we have a on_key callback. On a regular client paste would be pretty trivial, we just insert the provided string in front of each selection. On a remote client we add a Paste message type that contains the string.

How does that sound ?

krobelus:
yeah I think so too

krobelus:
Updated with the UserInterface paste method.
It also fixes a bug where InsertCompletionHide was not run.

The last commit adds linewise pasting which might be the wrong behavior in some cases, I'm not sure

krobelus:
thanks for the feedback, this version should be less sloppy.
The behavior in some edge cases is accidental, like the behavior when pasting after `i<c-r>`.
Vim does not paste and cancels the `<c-r>`, this one pastes and keeps `<c-r>` alive.

mawww (approved):
Merged it and applied the extra fixes I wanted myself: I did not want to suggest change that would not work, and I needed to do the work to check... Still providing the comments here as I had them written, in case you find them useful.

krobelus:
oh the reason for the awkward logic was for when the pasted data contains a CSI.
I chose to never return them to the caller.
Otherwise, pasting binary data could trigger random keys I think. Probably not a big problem in practice. Some terminals strip nonprintable characters from bracketed paste contents but not all.
Maybe we should add an extra check to never return valid keys while `m_paste_buffer` is active.

caksoylar:
Somewhere between https://github.com/mawww/kakoune/commit/874435e392eb4645d655f96c9903af713789b947 and https://github.com/mawww/kakoune/commit/1990a764e3d2ffa77931068f876cd49f76bd43f9 the behavior of pasting multiple line text changed for me: It inserts CR characters (`0x0d`) rather than newlines in the editor (I assume LF?).

I am aware I am using this in an unorthodox environment which is probably the cause (in Windows 11, WSL through Windows Terminal) but this wasn't the behavior before -- would it be possible to handle this like it used to be? I'd be happy to provide more info if no immediate cause comes to mind.

PR #4827, opened 2023-01-14
Add APL Highlighting

razetime:
Addition of APL highlighting was discussed earlier in IRC. This is a first draft of a highlighter for APL. Testing help, code review and highlight rule suggestions are appreciated.


razetime:
In order to avoid conflicts with other input methods, this does not have symbol completion. `.inputrc` works correctly inside of kakoune.

Issue #4741, opened 2022-09-21
[BUG] rust indentation hates unmatched parentheses in strings.


ewtoombs:
### Version of Kakoune

v2021.11.08

### Reproducer

Open a .rs file.

Write the following:
```
fn f() {
    let x = "string with a ( in it"
}
```

Put the cursor here, where the hash is:
```
fn f() {
    let x = "string with a ( in it"#
}
```

Insert with `i`, then hit enter.

### Outcome

![screenshot_2022-09-21-214328](https://user-images.githubusercontent.com/2224848/191615915-19156417-3fd8-41af-9f9b-44b5a10839da.png)

### Expectations

Maybe, something more like
![screenshot_2022-09-21-214818](https://user-images.githubusercontent.com/2224848/191616170-7aba45f4-75a2-4d7d-af96-733369ced016.png)

ewtoombs:
# workaround
Add this to your kakrc:
```
hook global WinSetOption filetype=rust %{
    remove-hooks window rust-indent
    remove-hooks window rust-trim-indent
    hook window InsertChar \n %{ try %{
        execute-keys -draft -itersel k x s ^\h+ <ret> y gh j P
    }}
}
```

inahga:
The culprit is this line https://github.com/mawww/kakoune/blob/master/rc/filetype/rust.kak#L170. It can be commented out to bring good autoindent behavior back--on Fedora the path is `/usr/share/kak/rc/filetype/rust.kak`.

Looks like its original intent is to facilitate writing argument lists, i.e.
![image](https://user-images.githubusercontent.com/57697428/216800947-a412d180-ab54-4c9e-8ef1-7e0fff961c07.png)
![image](https://user-images.githubusercontent.com/57697428/216800958-dd3a069b-17c7-4ee2-8652-b4abb1be3318.png)

Although IMO that's more of a job for `rustfmt` :shrug:. I can try put up a fix that keeps that behavior intact, although I'm half way tempted to just remove it--it's not behavior shared with the Go, C, or Python filetypes.

inahga:
Hah, the line right after it https://github.com/mawww/kakoune/blob/master/rc/filetype/rust.kak#L173 breaks multiline quotations:
![image](https://user-images.githubusercontent.com/57697428/216802316-75795fad-158c-453e-8d57-b5912c9cd479.png)

Again I'm tempted to delete that behavior and let `rustfmt` be in charge of the finer points of formatting...

[...]

PR #4837, opened 2023-02-06
rc/filetype/rust: remove after-open-with-chars indentation

inahga:
Supports https://github.com/mawww/kakoune/issues/4741, see screenshots in that thread for more info.

This breaks any quotations that have an opening brace in them. Rather than add more regex-spaghetti to handle this case--let it be up to the user to do this kind of advanced formatting, preferably with `rustfmt`.

Let me know if this is too aggressive of a move--trying the simple approach first but let me know if you'd rather keep this functionality.


mawww:
Anybody writing rust having an opinion on this change ?

inahga:
I've pretty much gone with the approach described here https://github.com/mawww/kakoune/issues/4741#issuecomment-1418510637 which is to just turn off autoindent, so I have no need for this PR anymore.

I'll have to leave the fixing of the rust autoident to someone else.

PR #4846, opened 2023-02-16
Add option to set maximum info box width

inahga:
Some plugins (*cough* kak-lsp) and help texts tend to have immensely long content in a single line. This generates info boxes that span the entire terminal width. This is made especially worse on widescreen monitors or at small text size.

This grants user control over how wide these boxes are.

I deliberately avoid pushing this change to `kak-lsp` because it's not the only plugin that this could help--see the `hook` help text for an example of this problem in vanilla Kakoune. I would also suggest that since this is a rendering concern, it be handled by the terminal rendering logic.

Here's some screenshots (with `set option global ui_options terminal_info_max_width=80`)

<details> 
  <summary>menu</summary>
<p>

image

</p>
</details>

<details> 
  <summary>inline</summary>
<p>

image

</p>
</details>

<details> 
  <summary>modal</summary>
<p>

image

</p>
</details>

### DIFF
--- a/doc/pages/options.asciidoc
+++ b/doc/pages/options.asciidoc
@@ -377,6 +377,10 @@ are exclusively available to built-in options.
             reduce terminal output with sequences that could trigger flickering
             if unsynchronized (defaults to *false*)
 
+        *terminal_info_max_width*:::
+            set the maximum allowable width of an info box. set to zero for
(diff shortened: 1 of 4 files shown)


jwhett:
Thank you! 🙏

PR #4848, opened 2023-02-17
rc windowning zellij

QiBaobin:
add zellij support like screen, or tmux

https://github.com/zellij-org/zellij

### DIFF
--- a/rc/windowing/zellij.kak
+++ b/rc/windowing/zellij.kak
@@ -13,7 +13,7 @@ declare-option -hidden -docstring %{zellij run options (please check out "zellij
 
 define-command -hidden -params .. zellij-run %{
     nop %sh{
-        zellij run ${kak_opt_zellij_run_options} -- $SHELL -i -c "$*" < /dev/null > /dev/null 2>&1 &
+        zellij run $kak_opt_zellij_run_options -- "$@"
     }
     set global zellij_run_options ""
     unset buffer zellij_run_options
--- a/rc/windowing/zellij.kak
+++ b/rc/windowing/zellij.kak
@@ -5,24 +5,17 @@ provide-module zellij %{
 
 # ensure we're running under zellij
 evaluate-commands %sh{
-    [ -z "${kak_opt_windowing_modules}" ] || [ -n "$ZELLIJ" ] || echo 'fail zellij not detected'
+    [ -z "${kak_opt_windowing_modules}" ] || [ -n "$ZELLIJ" -a -n "$ZELLIJ_SESSION_NAME" ] || echo 'fail zellij not detected'
 }
 
-declare-option -hidden -docstring %{zellij run options (please check out "zellij run --help" for detail)}\
-    str zellij_run_options ""
-
-define-command -hidden -params .. zellij-run %{
-    nop %sh{
-        zellij run $kak_opt_zellij_run_options -- "$@"
-    }
-    set global zellij_run_options ""
-    unset buffer zellij_run_options
-    unset window zellij_run_options
-}
+define-command -hidden -params 2.. zellij-run %{ nop %sh{
+    zellij_run_options=$1
+    shift
+    zellij --session "$kak_client_env_ZELLIJ_SESSION_NAME" run $zellij_run_options -- "$@"
+}}
 
 define-command -hidden -params 1.. zellij-terminal-impl %{
-    set global zellij_run_options "--close-on-exit"
-    zellij-run %arg{@}
+    zellij-run "--close-on-exit" %arg{@}
 }
 complete-command zellij-terminal-impl shell
 
@@ -31,8 +24,7 @@ zellij-terminal-vertical <program> [<arguments>]: create a new terminal as a zel
 The current pane is split into two, top and bottom
 The program passed as argument will be executed in the new terminal' \
 %{
-    set global zellij_run_options "--direction down"
-    zellij-run %arg{@}
+    zellij-run "--direction down" %arg{@}
 }
 complete-command zellij-terminal-vertical shell
 
@@ -41,8 +33,7 @@ zellij-terminal-horizontal <program> [<arguments>]: create a new terminal as a z
 The current pane is split into two, left and right
 The program passed as argument will be executed in the new terminal' \
 %{
-    set global zellij_run_options "--direction right"
-    zellij-run %arg{@}
+    zellij-run "--direction right" %arg{@}
 }
 complete-command zellij-terminal-horizontal shell
 
--- a/rc/windowing/zellij.kak
+++ b/rc/windowing/zellij.kak
@@ -8,6 +8,9 @@ evaluate-commands %sh{
     [ -z "${kak_opt_windowing_modules}" ] || [ -n "$ZELLIJ" -a -n "$ZELLIJ_SESSION_NAME" ] || echo 'fail zellij not detected'
 }
 
+define-command -params 1.. zellij-action %{ nop %sh{
+    zellij --session "$kak_client_env_ZELLIJ_SESSION_NAME" action "$@"
+}}
 define-command -hidden -params 2.. zellij-run %{ nop %sh{
     zellij_run_options=$1
     shift
--- a/rc/windowing/zellij.kak
+++ b/rc/windowing/zellij.kak
@@ -1,0 +1,74 @@
+# https://zellij.dev/
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+provide-module zellij %{
+
+# ensure we're running under zellij
+evaluate-commands %sh{
+    [ -z "${kak_opt_windowing_modules}" ] || [ -n "$ZELLIJ" ] || echo 'fail zellij not detected'
+}
+
+declare-option -hidden -docstring %{zellij run options (please check out "zellij run --help" for detail)}\
+    str zellij_run_options ""
+
+define-command -hidden -params .. zellij-run %{
+    nop %sh{
+        zellij run ${kak_opt_zellij_run_options} -- $SHELL -i -c "$*" < /dev/null > /dev/null 2>&1 &
+    }
+    set global zellij_run_options ""
+    unset buffer zellij_run_options
+    unset window zellij_run_options
+}
+
+define-command -hidden -params 1.. zellij-terminal-impl %{
+    set global zellij_run_options "--close-on-exit"
+    zellij-run %arg{@}
+}
+complete-command zellij-terminal-impl shell
+
+define-command zellij-terminal-vertical -params 1.. -docstring '
+zellij-terminal-vertical <program> [<arguments>]: create a new terminal as a zellij pane
+The current pane is split into two, top and bottom
+The program passed as argument will be executed in the new terminal' \
+%{
+    set global zellij_run_options "--direction down"
+    zellij-run %arg{@}
+}
+complete-command zellij-terminal-vertical shell
+
+define-command zellij-terminal-horizontal -params 1.. -docstring '
+zellij-terminal-horizontal <program> [<arguments>]: create a new terminal as a zellij pane
+The current pane is split into two, left and right
+The program passed as argument will be executed in the new terminal' \
+%{
+    set global zellij_run_options "--direction right"
+    zellij-run %arg{@}
+}
+complete-command zellij-terminal-horizontal shell
+
+define-command zellij-focus -params ..1 -docstring '
+zellij-focus [<client>]: focus the given client
+If no client is passed then the current one is used' \
+%{ evaluate-commands %sh{
+    if [ $# -eq 1 ]; then
+        printf "evaluate-commands -client '%s' focus" "$1"
+    elif [ -n "${kak_client_env_ZELLIJ}" ]; then
+        output=$(mktemp -d "${TMPDIR:-/tmp}"/kak-zellij.XXXXXXXX)/dump-screen
+        pane_count=0
+        while [ $((pane_count+=1)) -lt 10 ]; do
+             if zellij action dump-screen "$output" && grep "${kak_client}@\[$kak_session\]" "$output" > /dev/null ; then
+                 break;
+             fi
+             zellij action focus-next-pane
+        done
+        rm -r $(dirname $output)
+    fi
+}}
+complete-command -menu zellij-focus client
+
+## The default behaviour for the `new` command is to open an horizontal pane in a zellij session
+alias global focus zellij-focus
+alias global terminal zellij-terminal-impl
+
+}
+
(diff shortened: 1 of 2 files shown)


mawww:
Nice, could you squash those commits together and add a copyright waiver one (see CONTRIBUTING)

QiBaobin:
> Nice, could you squash those commits together and add a copyright waiver one (see CONTRIBUTING)

Done, thanks for the review.

rc/windowing/zellij.kak
```diff
@@ -0,0 +1,74 @@
...
+               ((i+=2))
+           else
+               ((i+=1))
+           fi
+           arg_count=$((i-1))
+       else
+           arg_count=$((i-1))
+           i=$#
+       fi
+   done
```
QiBaobin, line 32: this allow `:terminal` accept options like `-c` (close-on-exit) or `-d down`

rc/windowing/zellij.kak
```diff
@@ -0,0 +1,74 @@
...
+   echo zellij run ${@:1:$arg_count} -- $SHELL -i -c "'"${@:$((1+$arg_count))}"'"
```
QiBaobin, line 33: use a shell to support `alias` and command chains like 'fd -tf | sk`

rc/windowing/zellij.kak
```diff
@@ -0,0 +1,74 @@
...
+zellij-terminal-window-edit <file>: create a new window to edit given file' \
+%{
+    zellij-terminal-impl -c kak -c %val{session} %arg{@}
+}
+complete-command zellij-terminal-window-edit file
+
+define-command zellij-focus -params 0 -docstring '
+zellij-focus: focus a panel' \
+%{
+    zellij action switch-mode pane
```
QiBaobin, line 65: this will enter zellij pane mode so that we can use keys to select panels

rc/windowing/zellij.kak
```diff
@@ -0,0 +1,74 @@
...
+zellij-terminal-horizontal <program> [<arguments>]: create a new terminal as a zellij pane
+The current pane is split into two, left and right
+The program passed as argument will be executed in the new terminal' \
+%{
+    zellij-terminal-impl -d right %arg{@}
+}
+complete-command zellij-terminal-horizontal shell
+
+define-command zellij-terminal-window-edit -params .. -docstring '
+zellij-terminal-window-edit <file>: create a new window to edit given file' \
```
mawww, line 56: This seems equivalent to `new edit <file>` I dont think we need to add a new alias just for that use case.

rc/windowing/zellij.kak
mawww: I dont think we should support that for now, `terminal` is a unified interface for various windowing provider. Additionally this seems to be using bash array accesses, which is not supported by posix shells, so this wont work on non-bash shells.

mawww (changes requested):

rc/windowing/zellij.kak
QiBaobin: Make sense, fixed, thanks!

rc/windowing/zellij.kak
QiBaobin: Good point, fixed now, thanks!

QiBaobin:
@mawww at the last commit, I introduced an option to provide zellij options like `--close-on-exit' as it's very import and we want set it case by case. Sometimes we don't need it as we want the ability to see the output after the completion, and enter to re-run etc; sometimes we need it to close the pane automatically while calling `new`, customized fuzzy finder etc.

Let me know if it doesn't make sense, I can revert it then, thanks!

rc/windowing/zellij.kak
```diff
@@ -0,0 +1,69 @@
+# https://zellij.dev/
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+provide-module zellij %{
+
+# ensure we're running under zellij
+evaluate-commands %sh{
+    [ -z "${client_opt_windowing_modules}" ] || [ -n "$ZELLIJ" ] || echo 'fail zellij not detected'
```
glaeqen, line 8: `[ -z "${client_opt_windowing_modules}" ]` always returns 0. It should be `client_env..` right? Also, if we are checking the client, not server, we probably should check it on any command invocation. Because module will be loaded only once. So having this in a separate, callable command would probably make more sense.

rc/windowing/zellij.kak
```diff
@@ -0,0 +1,69 @@
...
+}
+
+declare-option -hidden -docstring %{zellij run options (
+    -c, --close-on-exit            Close the pane immediately when its command exits
```
glaeqen, line 12: Including the `help` of an external tool that might independently change does not seem like a good idea to me.

rc/windowing/zellij.kak
```diff
@@ -0,0 +1,69 @@
...
+    -f, --floating                 Open the new pane in floating mode
+    -h, --help                     Print help information
+    -n, --name <NAME>              Name of the new pane
+    -s, --start-suspended          Start the command suspended, only running after you first presses ENTER)}\
+    str zellij_run_options ""
+declare-option -hidden -docstring %{zellij run command} str zellij_run_cmd ""
+
+define-command -hidden -params 0 zellij-run %{
+    nop %sh{
+    zellij run ${kak_opt_zellij_run_options} -- $SHELL -i -c "${kak_opt_zellij_run_cmd}" < /dev/null > /dev/null 2>&1 &
```
glaeqen, line 24: Shouldn't you specify a `--session $ZELLIJ_SESSION_NAME` to make sure that this still works when you have multiple zellij sessions running?

rc/windowing/zellij.kak
```diff
@@ -0,0 +1,69 @@
...
+The current pane is split into two, left and right
+The program passed as argument will be executed in the new terminal' \
+%{
+    set global zellij_run_options "--direction right"
+    set global zellij_run_cmd "%arg{@}"
+    zellij-run
+}
+complete-command zellij-terminal-horizontal shell
+
+define-command zellij-focus -params 0 -docstring '
```
glaeqen, line 58: `focus` alias is invoked every once in a while and should not be interactive, right @mawww ?

rc/windowing/zellij.kak
```diff
@@ -0,0 +1,69 @@
...
+zellij-focus: focus a panel' \
+%{ nop %sh{
+    zellij action switch-mode pane
+}}
+
+## The default behaviour for the `new` command is to open an horizontal pane in a zellij session
+alias global focus zellij-focus
+alias global terminal zellij-terminal-impl
+
```
glaeqen, line 67: `zellij-terminal-impl` does not seem to work correctly with `connect.kak` which probably means that arguments are not forwarded correctly. 

Also, `connect.kak` requires `popup` to be defined but considering that we are not providing it by convention across all `windowing` implementations so far (and it's up to a user to setup a hook correctly to bind extra aliases), this is something that can be left for a different discussion

rc/windowing/zellij.kak
glaeqen, line 24: `connect.kak` spawns quite wild shell processes that will mess with the Zellij's pane name rendering. In #4850 I suggested setting it to nothing but that's probably negotiable.

rc/windowing/zellij.kak
QiBaobin: Sorry, I might copy the logic from an old `tmux.kak`, I will change it to `"${kak_opt_windowing_modules}"` as the latest one, does it OK?

rc/windowing/zellij.kak
QiBaobin: I don't think we need it as zellij would generate a `ZELLIJ_SESSION_NAME` env automatically, use it when running command.

QiBaobin: I prefer using the command which is the default so that we can know what that pane does especially when we have multiple jobs.

rc/windowing/zellij.kak
QiBaobin: let me check other focus implementations, will follow the standards, thanks

rc/windowing/zellij.kak
QiBaobin: I didn't find `connect.kak`, could you point me where it is or what's the issue?

QiBaobin: Fount it, it's a plugin, will try it.

rc/windowing/zellij.kak
QiBaobin: Fixed it, thanks for the review!

rc/windowing/zellij.kak
QiBaobin: I gave up this approach, it's very hard to pass arguments and make them in correct quotes.

rc/windowing/zellij.kak
QiBaobin: let me if there is any real use which gives a trouble, thanks

glaeqen:
Looks okay, although I have a few nitpicks

- In the beginning I though that you wanted to use `zellij_run_options` as some kind of memory for the last set of options. But you clear them in `zellij_run`. What's the value of doing that instead of
```
def -params 1.. zellij-run %{
    nop %sh{
        ARG1="$1"
        shift
        zellij run $ARG1 -- "$@"
    }
}
```
?

- I'm a little scared of having `rm -r` in a re-focus routine. If one has to do the ASCII screenshot of a multiplexer and grep for stuff, maybe we should just not implement it? Admittedly, I don't like this behaviour at all but my concern here is mostly regarding complexity of this command as of now.
- I've mentioned in one of the comments that maybe we should check the `$kak_client_env_ZELLIJ_SESSION_NAME` on every command invocation in case someone `kak -c` to the kak session within the zellij session. I think trying to do the `terminal` from a client that is outside of the zellij session should yield an error.
- `zellij` module load should also fail if we won't find `zellij` in PATH IMHO, at least it could be a nice to have thing.
- `zellij action` is even more general than `zellij run` and allows resizing of windows and all kinds of stuff. It'd be nice to have easy access to it from within the kakoune. One can do fun stuff like
```
hook global ModuleLoaded zellij %{
    define-command -hidden zellij-bigger-terminal-popup -params 1.. %{
        set global zellij_run_options "--floating -c"
        zellij-run %arg{@}
        #zellij-action toggle-pane-embed-or-floating
        #zellij-action toggle-fullscreen
        eval %sh{
            for i in {0..8}
            do
                echo zellij-action resize increase
            done
        }
   }
    alias global popup zellij-bigger-terminal-popup
}
```


If you want, take a look at my PR #4850. I did some of that stuff there.

QiBaobin:
> 



> Looks okay, although I have a few nitpicks
> 
>     * In the beginning I though that you wanted to use `zellij_run_options` as some kind of memory for the last set of options. But you clear them in `zellij_run`. What's the value of doing that instead of
> 
> 
> ```
> def -params 1.. zellij-run %{
>     nop %sh{
>         ARG1="$1"
>         shift
>         zellij run $ARG1 -- "$@"
>     }
> }
> ```
> 
> ?
> 
>     * I'm a little scared of having `rm -r` in a re-focus routine. If one has to do the ASCII screenshot of a multiplexer and grep for stuff, maybe we should just not implement it? Admittedly, I don't like this behaviour at all but my concern here is mostly regarding complexity of this command as of now.
> 
>     * I've mentioned in one of the comments that maybe we should check the `$kak_client_env_ZELLIJ_SESSION_NAME` on every command invocation in case someone `kak -c` to the kak session within the zellij session. I think trying to do the `terminal` from a client that is outside of the zellij session should yield an error.
> 
>     * `zellij` module load should also fail if we won't find `zellij` in PATH IMHO, at least it could be a nice to have thing.
> 
>     * `zellij action` is even more general than `zellij run` and allows resizing of windows and all kinds of stuff. It'd be nice to have easy access to it from within the kakoune. One can do fun stuff like
> 
> 
> ```
> hook global ModuleLoaded zellij %{
>     define-command -hidden zellij-bigger-terminal-popup -params 1.. %{
>         set global zellij_run_options "--floating -c"
>         zellij-run %arg{@}
>         #zellij-action toggle-pane-embed-or-floating
>         #zellij-action toggle-fullscreen
>         eval %sh{
>             for i in {0..8}
>             do
>                 echo zellij-action resize increase
>             done
>         }
>    }
>     alias global popup zellij-bigger-terminal-popup
> }
> ```
> 
> If you want, take a look at my PR #4850. I did some of that stuff there.



> Looks okay, although I have a few nitpicks
> 
>     * In the beginning I though that you wanted to use `zellij_run_options` as some kind of memory for the last set of options. But you clear them in `zellij_run`. What's the value of doing that instead of
> 
> 
> ```
> def -params 1.. zellij-run %{
>     nop %sh{
>         ARG1="$1"
>         shift
>         zellij run $ARG1 -- "$@"
>     }
> }
> ```
> 
> ?
> 
>     * I'm a little scared of having `rm -r` in a re-focus routine. If one has to do the ASCII screenshot of a multiplexer and grep for stuff, maybe we should just not implement it? Admittedly, I don't like this behaviour at all but my concern here is mostly regarding complexity of this command as of now.
> 
>     * I've mentioned in one of the comments that maybe we should check the `$kak_client_env_ZELLIJ_SESSION_NAME` on every command invocation in case someone `kak -c` to the kak session within the zellij session. I think trying to do the `terminal` from a client that is outside of the zellij session should yield an error.
> 
>     * `zellij` module load should also fail if we won't find `zellij` in PATH IMHO, at least it could be a nice to have thing.
> 
>     * `zellij action` is even more general than `zellij run` and allows resizing of windows and all kinds of stuff. It'd be nice to have easy access to it from within the kakoune. One can do fun stuff like
> 
> 
> ```
> hook global ModuleLoaded zellij %{
>     define-command -hidden zellij-bigger-terminal-popup -params 1.. %{
>         set global zellij_run_options "--floating -c"
>         zellij-run %arg{@}
>         #zellij-action toggle-pane-embed-or-floating
>         #zellij-action toggle-fullscreen
>         eval %sh{
>             for i in {0..8}
>             do
>                 echo zellij-action resize increase
>             done
>         }
>    }
>     alias global popup zellij-bigger-terminal-popup
> }
> ```
> 
> If you want, take a look at my PR #4850. I did some of that stuff there.

The origin idea to declare an option is that we can control the behavior of the `terminal` command, but you are correct if we know we are using zellij, we can use zellij run directly instead of `terminal`.

I just looked at your PR, all looks great except two things:
1. Since we give the control to caller using the $1, I think we shall move `-n "" --close-on-exit' from zellij-terminal-impl to commands like zellij-terminal-horizontal so that we can define new commands using different options.
2. I think the default `terminal` implementation command shall avoid provide the direction, you can see it would be a lot better when you open more than 3 window using `new` command, the default zellij algorithm does a better job than a fixed direction.

glaeqen:
Looks good to me in general. Seems to work. What do you think about having the `zellij-action` command?

QiBaobin:
> Looks good to me in general. Seems to work. What do you think about having the `zellij-action` command?

Added now

PR #4860, opened 2023-03-02
Fix Svelte detection: use `set buffer filetype` instead of `window`

poweredbypie:
As the title suggests, this fixes the Svelte filetype detector to only set the buffer as a Svelte file.

### DIFF
--- a/rc/filetype/svelte.kak
+++ b/rc/filetype/svelte.kak
@@ -1,5 +1,5 @@
-hook global WinCreate .*\.svelte %[
-    set-option window filetype svelte
+hook global BufCreate .*\.svelte %[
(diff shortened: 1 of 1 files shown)


rc/filetype/svelte.kak
```diff
@@ -1,5 +1,5 @@
 hook global WinCreate .*\.svelte %[
```
mawww, line 1: This should be `BufCreate`

mawww (changes requested):

rc/filetype/svelte.kak
poweredbypie: Fixed in rebase!

PR #4863, opened 2023-03-06
git-format-patch and git-send-email added.

nonumeros:
git-format-patch options are based on gitdiffcore. git-send-email settings may be configured by the user. The git manual documentation provides more information.

Everything else is handled primarily by git and git-send-email respectively.

Of course that any `git diff --no-index ` would provide more than enough for the majority of cases. But if `git diff` is available within kak, it follows that `git-format-patch` should complete it.

The same from above can also be said about `git-send-email`. All of these follow each other. `git-diff` `git-format-patch` `git-send-email` 

I tested it and couldn't see any problems. If the `diff` is created, then `git-format-patch` should format it accordingly e.g., `git format-patch -1 *.patch` followed right after by `git-send-email` 

As long as the latter `.git/config` settings are correct, there's no need in not to use the editor. For example, from what I've read (not the first time), the option `--signoff` is used mainly as a  featured requirement for the kernel and git projects. But it's simply beyond me having a tool like kakoune available and not use it with all the git options. Or even for local projects.

### DIFF
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -39,6 +39,22 @@ hook -group git-show-branch-highlight global WinSetOption filetype=git-show-bran
     hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/git-show-branch}
 }
 
+hook -group git-format-patch global WinSetOption filetype=git-format-patch %{
+    require-module diff
+    add-highlighter window/git-format-patch group
+    add-highlighter window/git-format-patch/ regex '^( ?[*|\\ /_.-])*\h{,3}([a-zA-Z_-]+:) (.*?)$' 2:variable 3:value
+
+    hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/git-format-patch }
+}
+
+hook -group git-send-email global WinSetOption filetype=git-send-email %{
+    require-module diff
+    add-highlighter window/git-send-email group
+    add-highlighter window/git-send-email/ regex '^( ?[*|\\ /_.-])*\h{,3}([a-zA-Z_-]+:) (.*?)$' 2:variable 3:value
+
+    hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/git-send-email}
+}
+
 declare-option -hidden line-specs git_blame_flags
 declare-option -hidden line-specs git_diff_flags
 declare-option -hidden int-list git_hunk_list
@@ -50,17 +66,18 @@ define-command -params 1.. \
         Available commands:
             add
             rm
-            reset
             blame
             commit
             checkout
             diff
+            format-patch
             hide-blame
             hide-diff
             init
             log
             next-hunk
             previous-hunk
+            send-email
             show
             show-branch
             show-diff
@@ -68,11 +85,13 @@ define-command -params 1.. \
             update-diff
     } -shell-script-candidates %{
     if [ $kak_token_to_complete -eq 0 ]; then
-        printf "add\nrm\nreset\nblame\ncommit\ncheckout\ndiff\nhide-blame\nhide-diff\nlog\nnext-hunk\nprev-hunk\nshow\nshow-branch\nshow-diff\ninit\nstatus\nupdate-diff\n"
+        printf "add\nrm\nblame\ncommit\ncheckout\nformat-patch\ndiff\nhide-blame\nhide-diff\nlog\nnext-hunk\nprev-hunk\nsend-email\nshow\nshow-branch\nshow-diff\ninit\nstatus\nupdate-diff\n"
     else
         case "$1" in
             commit) printf -- "--amend\n--no-edit\n--all\n--reset-author\n--fixup\n--squash\n"; git ls-files -m ;;
             add) git ls-files -dmo --exclude-standard ;;
+            send-email) git ls-files -dmo --exclude-standard ;;
+            format-patch) git ls-files -dmo --exclude-standard ;;
             rm) git ls-files -c ;;
         esac
     fi
@@ -92,6 +111,8 @@ define-command -params 1.. \
 
         case "$1" in
            diff) map_diff_goto_source=true; filetype=diff ;;
+           format-patch) map_diff_goto_source=true; filetype=git-format-patch ;;
+           send-email) map_diff_goto_source=true; filetype=git-send-email;;
            show) map_diff_goto_source=true; filetype=git-log ;;
            show-branch) filetype=git-show-branch ;;
            log)  filetype=git-log ;;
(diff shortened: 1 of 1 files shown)


nonumeros:
Reloading seems to be working fine. At least with the revision I'm running. Can't complain.

Most kak users are fully aware of these `git` operations. But I'll mention it either way because I personally haven't seen a problem. 

`git diff --no-index a.txt b.txt` and saving it afterwards with say e.g., `a-b-diff` is more than enough to invoke `git-format-patch -1 a-b-diff` which should in turn create the `000*-1-a-b-diff.patch*` formatted patch needed . But even then re-running the `format-patch` with say a `--signoff`option should rewrite it and reload it without any issues.

`git-send-email` follows relatively easily as long as `git-format-patch` generated/formatted it accordingly.

rc/tools/git.kak
```diff
@@ -39,6 +39,22 @@ hook -group git-show-branch-highlight global WinSetOption filetype=git-show-bran
     hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/git-show-branch}
 }
 
+hook -group git-format-patch global WinSetOption filetype=git-format-patch %{
+    require-module diff
```
Screwtapello, line 43: Why require the `diff` module if you don't use it?

Screwtapello:
I'm a little confused what the purpose of this is.

`git diff` normally writes the diff to stdout, so it's reasonable to me that `git diff` in Kakoune would open the diff result as a new buffer.

`git format-patch` I think normally writes the diff to a series of files, so it's not clear to me what running it inside Kakoune would do. Does it open the files in Kakoune? Just the first one? Does it just display the list of filenames it wrote, which is what `git format-patch` normally writes to stdout?

`git send-email` I haven't used in a very long time, but as far as I know it sends an email without writing anything to stdout. I can't imagine what it would do from inside Kakoune.

nonumeros:
> I'm a little confused what the purpose of this is.
> 
> `git diff` normally writes the diff to stdout, so it's reasonable to me that `git diff` in Kakoune would open the diff result as a new buffer.
> 
> `git format-patch` I think normally writes the diff to a series of files, so it's not clear to me what running it inside Kakoune would do. Does it open the files in Kakoune? 

That's the trace I get by following the file.  Obviously after `git-commit` 

>Just the first one? Does it just display the list of filenames it wrote, which is what `git format-patch` normally writes to stdout?
> 

`git-format-patch` after `git-commit diff> writes to stdout that very same `diff` with a format usually as `000*-diff` with the extension `.patch` amended 

> `git send-email` I haven't used in a very long time, but as far as I know it sends an email without writing anything to stdout. I can't imagine what it would do from inside Kakoune.

It usually, as long as `git-commit` successfully  goes  through sent to stdout with:

```
To whom should the emails be sent (if anyone)? 
Message-ID to be used as In-Reply-To for the first email (if any)? 

From: 
To:
Subject:
Date:
Message-Id:
X-Mailer:
MIME-Version:
Content-Transfer-Encoding: 

    The Cc list above has been expanded by additional
    addresses found in the patch commit message. By default
    send-email prompts before sending whenever this occurs.
    This behavior is controlled by the sendemail.confirm
    configuration setting.

    For additional information, run 'git send-email --help'.
    To retain the current behavior, but squelch this message,
    run 'git config --global sendemail.confirm auto'.

Send this email? ([y]es|[n]o|[e]dit|[q]uit|[a]ll): 
```

and if `[e]` is selected it'd open kakoune afterwards for further editing

rc/tools/git.kak
nonumeros: I didn't compile without it.

nonumeros:
> I'm a little confused what the purpose of this is.

There is no need to implement `git.kak` in the first place without merging all `git` operations into it. For the simple reason that one would have bits of the program thrown  in `rc/tools` while  the remainder is delegated all over. 

Anyone knows that by introducing more features also makes it prone to more errors, but in this case, as long as it's not a bug in the program itself, all `git` features ought to be intertwined  in the editor. No need to have a piece there and the other one not implemented 

>git send-email I haven't used in a very long time, but as far as I know it sends an email without writing anything to stdout. I can't imagine what it would do from inside Kakoune.

You  probably had a mismanaged configuration that couldn't come to terms

nonumeros:
@Screwtapello  Let me give you the benefit of the doubt here. You probably are leaning more towards `git-format-patch` than  a `git-send-email` implementation here, correct? 

When you said, quote `so it's not clear to me what running it inside Kakoune would do`, end quote, what piece of code exactly did it `git-send-email` inject the editor with? :) 

You made my day with that. 

If you were to configure `git-send-email` accordingly with default values as the git documentation recommends, that conclusion from your statement is just unwarranted.  There's no basis for it.

nonumeros:
the editing part from `git-send-email` is, as of now, not handled by the editor itself. Why? 

Just to have  the following? 

```
To whom should the emails be sent (if anyone)? username@hostname
Message-ID to be used as In-Reply-To for the first email (if any)? 

From: <email> 
To: <email> 
Subject: [PATCH] a-b-diff.patch 
Date: Mon,  6 Mar 2023 23:29:37 -0500
Message-Id: <xxxxxxx-useremail@hostname>
X-Mailer: git-send-email 2.36.5
MIME-Version: 1.0
Content-Transfer-Encoding: 8bit

    The Cc list above has been expanded by additional
    addresses found in the patch commit message. By default
    send-email prompts before sending whenever this occurs.
    This behavior is controlled by the sendemail.confirm
    configuration setting.

    For additional information, run 'git send-email --help'.
    To retain the current behavior, but squelch this message,
    run 'git config --global sendemail.confirm auto'.

Send this email? ([y]es|[n]o|[e]dit|[q]uit|[a]ll): e
```
 and then, ending up with `e` which will loop around kakoune regardless. Why?

Most likely kakoune has to `[edit]` the `diff`  file created by `git-format-patch` either way, correct? It makes no sense not to include both `git-send-email` and `git-format-patch` separate from the editor.

nonumeros:
If any addition in `git.kak` is long overdue, I can safely say it would be this very same  `git-format-patch` and `git-send-email`. 

Also, I can almost assure you, that If by any chance, the git devs were to work extensively on `git-diff` now or in the near future, `git-format-patch`, `gitdiffcore` as it's being primarily the accepted method used by the kernel devs, and even by the devs on `git` itself, `git-format-patch` , or for that matter `git-send-email` will not get overlooked/neglected. It would make no sense to do so, when this is the accepted mechanism in place.

nonumeros:
@Screwtapello  I think there may be an issue with the way that `git add` is behaving for the file. Some sort of conflict that prevents `git-format-patch` to behave and hence format the file accordingly and as intended. 

Will have to close until I find the underlying cause. thanks for the review. I appreciate it. :+1:

PR #4865, opened 2023-03-09
git-format-patch,git-restore,git-send-email

nonumeros:
@Screwtapello  I know you'd like this one. The (3) tenets of git. `restore` was never included and this is perhaps as important as anything else within … almost mandatory its implementation, since `git-format-patch` has among among many options the capability of patching up everything under the worktree and beyond 

Never included under kak. but I t should have been. 

`git format-patch -1` should be more than enough in the meantime . The revision is against master and not with `--root` or `--all` 

The latter (sounding redundant now) but the latter ouygth to be used only if you want to end up with patches all over 

When it comes to `git-send-email` and getting the config right  `git config sendemail.<tab>` should be more than enough to get you going. 

I don't know if at one time it was relatively successful under `imap` but I couldn't get it to work. Looking under the options fo `git config sendemaiil.<tab>` didn't see anything about it either. So on that part about `imap` I can't comment there

@Screwtapello  also. 

You were right about being concerned with the `supposed` characters that could get passed on to the editor with the `git-send-email` But that concerning would be valid only with say, Emacs for example in sporadic cases. I can confirm it that it did happened. But it was also something else that may have caused it. 

Normally speaking,, nor even emacs had that problem. Much less kakoune.

Kakoune never had a problem with it. Never. Under emacs it did happened running the sub-process on the shell. But I can't reproduce it either afterwards. It happened once. And that was it. 

What is it that you don't like or that you would recommend to remove from this pull request


nonumeros:
I don't want to remove `git-send-email`. Not with a valid reason. I can't think of one (valid enough)  now under the current revision  And `git-format-patch` is the de facto by git devs. Anyone know this by now. 

And `git-restore` is almost as important as any other git operations out there.

nonumeros:
simple configuration for `git-send-email` 

```
[user]
	name = xxxxxx
	email = xxxxx@xxx.com
[sendemail]
	smtpServer = xxx.server.com
	smtpServerPort = 
	smtpEnncrytion = tls
	smtpUser = username@server.com
	smtpPass = token
[credeential]
	helper = store
```

should be sufficient enough to work 

and if the formatted patch is somehow affected in any shape or form by whatever version of `git send-email` was used, I reckon the devs from git should be notified about it.

Issue #11, opened 2014-01-02
'gf' on absolute path not working


alexherbo2:
[![](http://i.imgur.com/ZudhuSKl.png)](http://i.imgur.com/ZudhuSK.png)

PR #4867, opened 2023-03-12
Fix crash when redrawing after connecting external monitor

krobelus:
When Kakoune's terminal is shown on my laptop monitor and I plug in my
external monitor, the terminal's workspace will move to that external
monitor. When this happens, Kakoune segfaults consistently.

The crash happens in "TerminalUI::Screen::output" where we access a
dangling reference

	    auto output_line = [&](const Line& line) {
	        ColumnCount pending_move = 0;
	        for (auto& [text, skip, face] : line.atoms)
	        {
	/->          if (text.empty() and skip == 0)
	|                continue;
	|
	 cannot access memory referenced by "text"

The stack trace is

	Program received signal SIGSEGV, Segmentation fault.
	0x0000555555a5f1ac in Kakoune::String::Data::is_long (this=0x1a) at /home/johannes/git/kakoune/src/string.hh:180
	(gdb) bt
	#0  0x0000555555a5f1ac in Kakoune::String::Data::is_long (this=0x1a) at /home/johannes/git/kakoune/src/string.hh:180
	#1  0x0000555555a5f1d6 in Kakoune::String::Data::size (this=0x1a) at /home/johannes/git/kakoune/src/string.hh:181
	#2  0x0000555555de34e4 in Kakoune::String::length (this=0x1a) at /home/johannes/git/kakoune/src/string.hh:139
	#3  Kakoune::StringOps<Kakoune::String, char>::empty (this=0x1a) at /home/johannes/git/kakoune/src/string.hh:68
	#4  operator() (__closure=0x7fffffffba50, line=...) at terminal_ui.cc:303
	#5  0x0000555555de45aa in Kakoune::TerminalUI::Screen::output (this=0x5555560a7a60, force=true, synchronized=true, writer=...) at terminal_ui.cc:371
	#6  0x0000555555de548a in Kakoune::TerminalUI::redraw (this=0x5555560a7a40, force=true) at terminal_ui.cc:535
	#7  0x0000555555de579e in Kakoune::TerminalUI::refresh (this=0x5555560a7a40, force=true) at terminal_ui.cc:556
	#8  0x0000555555a9ae71 in Kakoune::Client::redraw_ifn (this=0x55555609f9b0) at client.cc:284
	#9  0x0000555555aab3db in Kakoune::ClientManager::redraw_clients (this=0x7fffffffd060) at client_manager.cc:233
	#10 0x0000555555c86ef8 in Kakoune::run_server (session="", server_init="", client_init="", init_buffer="fish-rust/src/ast.rs", init_coord=..., flags=Kakoune::ServerFlags::None, ui_type=Kakoune::UIType::Terminal,
	    debug_flags=Kakoune::DebugFlags::None, files=ArrayView<Kakoune::StringView> = {...}) at main.cc:908
	#11 0x0000555555c8bb76 in main (argc=2, argv=0x7fffffffe7b8) at main.cc:1263

We check for terminal resize in

	TerminalUI::TerminalUI
	TerminalUI::get_next_key
	TerminalUI::draw
	TerminalUI::suspend

I think neither of those happens when the window is moved.

Looks like TerminalUI::refresh relies on up-to-date dimensions,
so add a resize check here too, fixing the crash.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,9 @@
 .*.kak.*
 src/kak
 src/kak.debug
+src/kak.debug.san_*
 src/kak.opt
+src/kak.opt.san_*
 src/.version*
 src/.*.json
 doc/kak.1.gz
--- a/src/file.cc
+++ b/src/file.cc
@@ -252,13 +252,14 @@ bool regular_file_exists(StringView filename)
            (st.st_mode & S_IFMT) == S_IFREG;
 }
 
+template<bool atomic>
 void write(int fd, StringView data)
 {
     const char* ptr = data.data();
     ssize_t count   = (int)data.length();
 
     int flags = fcntl(fd, F_GETFL, 0);
-    if (EventManager::has_instance())
+    if (not atomic and EventManager::has_instance())
         fcntl(fd, F_SETFL, flags | O_NONBLOCK);
     auto restore_flags = on_scope_end([&] { fcntl(fd, F_SETFL, flags); });
 
@@ -269,12 +270,15 @@ void write(int fd, StringView data)
             ptr += written;
             count -= written;
         }
-        else if (errno == EAGAIN and EventManager::has_instance())
+        else if (errno == EAGAIN and not atomic and EventManager::has_instance())
             EventManager::instance().handle_next_events(EventMode::Urgent, nullptr, false);
         else
             throw file_access_error(format("fd: {}", fd), strerror(errno));
     }
 }
+template void write<true>(int fd, StringView data);
+template void write<false>(int fd, StringView data);
+
 
 void write_to_file(StringView filename, StringView data)
 {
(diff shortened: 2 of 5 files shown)


mawww:
I am having trouble understanding what is going on here. It would mean m_screen.size.line is not in sync with either m_screen.line or m_screen.hashes buffer sizes. Those should only change inside TerminalUI::check_resize and I cannot find how they would get out of sync.

Running check_resize in refresh is not really nice because it will not update the actually display buffer from the server. This is why things work the current way:

1. SIGWINCH is emitted by the terminal emulator
2. Kakoune runs `signal_handler<&resize_pending>` with sets `resize_pending` to 1 and forces fd 0 to be set.
3. On next event loop fd 0 is set so we trigger `m_stdin_watcher` wich calls `get_next_key()`
4. `get_next_key()` calls`check_resize()` which calls `set_resize_pending` which sets `m_resize_pending` to true (this is a bit confusing, and not sure why we re-signal fd 0 there)
5. `get_next_key()` sees that `m_resize_pending` is set and returns a Resize key to the server
6. The server triggers a redraw with the new dimensions

If you manage to reproduce this easily, please try to `watch -l m_screen.size.line` in gdb and see if something else changes it without immediately recreating the lines/hashes buffer.

krobelus:
Figured it out, it's because of ae001a1f9 (Run EventManager whenever writing to a file descriptor would block, 2022-05-10) which means we run the event manager after SIGWINCH, which causes us to process the resize while we are still rendering an earlier one.

PR #4878, opened 2023-04-09
Implement on-abort for menu

natsukagami:
Most stuff is just being copied over from `prompt`.


src/commands.cc
```diff
@@ -2291,6 +2292,8 @@ const CommandDesc menu_cmd = {
                 select_cmds.push_back(parser[i+2]);
         }
 
+        auto on_abort = parser.get_switch("on-abort").value_or("").str();
```
krobelus, line 2295: It makes sense to support this, `prompt` has the same switch.

it doesn't feel great to use the empty string as sentinel value, it would be better to use the empty optional

PR #4892, opened 2023-05-09
Work around clang not inferring default args of template template args

### LINKED ISSUE #4872: Compilation failures on clang 16.0.0
Current master 019fbc54 (2023-03-14) doesn't compile on the most recent clang release (16.0.0). There seem to be two separate incompatibilities.

The first affects both clang 16.x and 15.x. Bisection shows it first appears in commit b2cf74bb on 2022-12-11:

```
c++  -O3 -pedantic -std=c++2a -g -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare -Wno-address -frelaxed-template-template-args -Wno-ambiguous-reversed-operator -MD -MP -MF .main.opt.d -c -o .main.opt.o main.cc
input_handler.cc:1472:16: error: alias template 'ConstArrayView' requires template arguments; argument deduction only allowed for class templates
        insert(ConstArrayView{content});
               ^
./array_view.hh:72:1: note: template is declared here
using ConstArrayView = ArrayView<const T>;
^
input_handler.cc:1518:16: error: alias template 'ConstArrayView' requires template arguments; argument deduction only allowed for class templates
        insert(ConstArrayView{str});
               ^
./array_view.hh:72:1: note: template is declared here
using ConstArrayView = ArrayView<const T>;
^
```

The second has been present since before the last release (v2022.10.31) but only affects compilation with the new clang 16.x, not the previous 15.x:

```
c++  -O3 -pedantic -std=c++2a -g -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare -Wno-address -frelaxed-template-template-args -Wno-ambiguous-reversed-operator -MD -MP -MF .command_manager.opt.d -c -o .command_manager.opt.o command_manager.cc
command_manager.cc:825:104: error: no matching function for call to 'gather'
        auto params = tokens | skip(1) | transform(&Token::content) | filter(std::not_fn(is_switch)) | gather<Vector>();
                                                                                                       ^~~~~~~~~~~~~~
./ranges.hh:633:6: note: candidate template ignored: invalid explicitly-specified argument for template parameter 'Container'
auto gather()
     ^
./ranges.hh:642:6: note: candidate template ignored: invalid explicitly-specified argument for template parameter 'Container'
auto gather()
     ^
1 error generated.
```

and similarly

```
c++  -O3 -pedantic -std=c++2a -g -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare -Wno-address -frelaxed-template-template-args -Wno-ambiguous-reversed-operator -MD -MP -MF .commands.opt.d -c -o .commands.opt.o commands.cc
commands.cc:2130:76: error: no matching function for call to 'gather'
                CommandManager::instance().execute_single_command(parser | gather<Vector>(), context, shell_context);
                                                                           ^~~~~~~~~~~~~~
./ranges.hh:633:6: note: candidate template ignored: invalid explicitly-specified argument for template parameter 'Container'
auto gather()
     ^
./ranges.hh:642:6: note: candidate template ignored: invalid explicitly-specified argument for template parameter 'Container'
auto gather()
     ^
1 error generated.
```

krobelus:
Clang 16 fails to compile calls like `gather<Vector>()`.

	c++  -O3 -pedantic -std=c++2a -g -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare -Wno-address -frelaxed-template-template-args -Wno-ambiguous-reversed-operator -MD -MP -MF .command_manager.opt.d -c -o .command_manager.opt.o command_manager.cc
	command_manager.cc:825:104: error: no matching function for call to 'gather'
	        auto params = tokens | skip(1) | transform(&Token::content) | filter(std::not_fn(is_switch)) | gather<Vector>();
	                                                                                                       ^~~~~~~~~~~~~~
	./ranges.hh:633:6: note: candidate template ignored: invalid explicitly-specified argument for template parameter 'Container'
	auto gather()
	     ^
	./ranges.hh:642:6: note: candidate template ignored: invalid explicitly-specified argument for template parameter 'Container'
	auto gather()
	     ^
	1 error generated.

because Vector has a defaulted template argument but gather's formal
parameter does not.

Upstream bug report: https://github.com/llvm/llvm-project/issues/63281
(not sure why it worked for us with Clang 15).

Work around this by adding gather() overloads whose template template
parameter clang can match against Vector. Add another one for HashSet.

GCC and earlier Clangs reject calls to these overloads as ambiguous, so use conditional
compilation.

Fixes #4872

### DIFF
--- a/src/ranges.hh
+++ b/src/ranges.hh
@@ -8,6 +8,11 @@
 #include <tuple>
 
 #include "constexpr_utils.hh"
+#if __clang_major__ >= 16
+#include "memory.hh"
+#include "meta.hh"
+#include "vector.hh"
+#endif
 
 namespace Kakoune
 {
@@ -638,6 +643,35 @@ auto gather()
     }};
 }
 
+#if __clang_major__ >= 16
+
+template <template <typename T,
+                    MemoryDomain domain = memory_domain(Meta::Type<T>{})>
+          class Container>
+auto gather()
+{
+    return ViewFactory{[](auto&& range) {
+        using std::begin; using std::end;
+        using ValueType = std::remove_cv_t<std::remove_reference_t<decltype(*begin(range))>>;
+        return Container<ValueType>(begin(range), end(range));
+    }};
+}
+
+template <template <typename T,
+                    MemoryDomain domain = MemoryDomain::Undefined,
+                    template<typename, MemoryDomain> class C = Vector>
+          class Container>
+auto gather()
+{
+    return ViewFactory{[](auto&& range) {
+        using std::begin; using std::end;
+        using ValueType = std::remove_cv_t<std::remove_reference_t<decltype(*begin(range))>>;
+        return Container<ValueType>(begin(range), end(range));
+    }};
+}
+
+#else
+
 template<template <typename Element> class Container>
 auto gather()
 {
@@ -647,6 +681,8 @@ auto gather()
         return Container<ValueType>(begin(range), end(range));
     }};
 }
+
+#endif
 
 template<typename ExceptionType, bool exact_size, size_t... Indexes>
 auto elements()


arachsys:
If you're on a glibc-based distro @mawww, the kernel.org slim toolchains at 

  https://mirrors.edge.kernel.org/pub/tools/llvm/

are super-handy for doing a throwaway test without actually building and installing particular versions of llvm yourself.

So for example in a Debian chroot with only gcc and binutils from build-essentials, it's sufficient to do:
```
# curl -LO https://mirrors.edge.kernel.org/pub/tools/llvm/files/llvm-16.0.6-x86_64.tar.xz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 79.7M  100 79.7M    0     0  3367k      0  0:00:24  0:00:24 --:--:-- 3172k
# tar xf llvm-16.0.6-x86_64.tar.xz
# make -C kakoune-2023.08.05 CC=llvm-16.0.6-x86_64/bin/clang++
[...]
```

They run anywhere with not too ancient a glibc. They won't work on musl distros, so ironically aren't as useful for me despite recommending them, but they're handy for reproducing compiler issues quickly in a glibc chroot nonetheless.

There are also multiple gcc+binutils toolchains at

  https://mirrors.edge.kernel.org/pub/tools/crosstool/

which might be useful.

krobelus:
for the record the actual upstream bug report is https://github.com/llvm/llvm-project/issues/63281

The `if clang-version > 16` is supremely ugly but I wasn't sure how to probe for this bug. Maybe there is a way with SFINAE.
Or we fix clang

svmhdvn:
I can confirm that this PR fixes the issue on my gentoo install running the musl+clang profile. My system compiler is clang-17.0.2. This is being tracked already in this gentoo bug https://bugs.gentoo.org/894378.

arachsys:
Just for @svmhdvn's benefit: another option when patching Kakoune for clang as a distro maintainer is to fix the handful of call sites of gather() to more explicitly declare types, so `gather<HashSet>` becomes `gather<HashSet<String>>` and so on.

In my build, I have a [patch that does this](https://github.com/arachsys/packages/blob/master/kakoune/clang.diff) you're welcome to grab if you feel more comfortable with that that an `#if __clang_major__ >= 16` overload of gather().

PR #4894, opened 2023-05-11
Fix debug keys output for shift/ctrl modified mouse events

arachsys:
Although Kakoune responds to modified mouse events, they show up in the debug buffer corrupted. to_string() tests for equality on the mouse event modifiers rather than testing just the relevant bits, so the modified mouse events incorrectly fall through to the normal key handling.

Fix this and restructure to allow mouse events to be modifier-prefixed.

### DIFF
--- a/src/keys.cc
+++ b/src/keys.cc
@@ -184,37 +184,33 @@ Key::MouseButton str_to_button(StringView str)
 String to_string(Key key)
 {
     const auto coord = key.coord() + DisplayCoord{1,1};
-    switch (Key::Modifiers(key.modifiers & ~Key::Modifiers::MouseButtonMask))
-    {
-        case Key::Modifiers::MousePos:
-            return format("<mouse:move:{}.{}>", coord.line, coord.column);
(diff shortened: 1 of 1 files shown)


arachsys:
To reproduce, `:set-option global debug keys` then switch to the `*debug*` buffer. Send some mouse clicks and you'll see them reported correctly. Send some shift- or ctrl- mouse clicks and you'll see invalid/corrupted keyboard keys reported instead.

Khechani (approved):

Khechani (approved):
src/input_handler.cc

arachsys:
> src/input_handler.cc

Yes, this is the place that these mouse events are handled (as opposed to stringified for debugging), and it already handles modified events correctly because it masks out the c-, s- and a- from key.modifiers before dispatching on them - modifying the behaviour slightly for c- clicks to add a new cursor/selection rather than set the main one.

Khechani (approved):

Khechani:
- [ ] []()

PR #4897, opened 2023-05-13
Cancel regex search on Control-C

### LINKED ISSUE #3950: Allow cancelling highlighting on huge buffers
When editing huge buffers, highlighting can slow down the UI, it'd be useful to allow cancelling the process.

krobelus:
On large files (for example "seq 10000000"), regex search is slow.
I often only realize that _after_ starting an (incremental) search,
which cannot be canceled.
Let's allow canceling regex searches with Control-C, similar to Vim
and less.

Apply this to all regex searches except for things like highlighting
(following commit). Also exclude things like hook filtering (users
of regex_match()) to ensure the behavior is predictable.

There are probably other places where we should check for Control-C
to prevent hangs on huge buffers.

Closes #3950

Also includes experimental commits to disable
- word completion in regex prompts
- regex highlighting in a buffer after interrupted by Control-C

### DIFF
--- a/src/regex_impl.hh
+++ b/src/regex_impl.hh
@@ -478,9 +478,10 @@ private:
         m_found_match = false;
         while (true) // Iterate on all codepoints and once at the end
         {
+            idle_func();
+
             if (++current_step == 0)
             {
-                idle_func();
 
                 // We wrapped, avoid potential collision on inst.last_step by resetting them
                 ConstArrayView<CompiledRegex::Instruction> instructions{m_program.instructions};
--- a/src/regex_impl.hh
+++ b/src/regex_impl.hh
@@ -269,7 +269,7 @@ public:
         {
             if (search)
             {
-                to_next_start(start, config, *start_desc);
+                to_next_start(start, config, *start_desc, idle_func);
                 if (start == config.end) // If start_desc is not null, it means we consume at least one char
                     return false;
             }
@@ -512,17 +512,18 @@ private:
             if (search and not m_found_match)
             {
                 if (start_desc and m_threads.next_is_empty())
-                    to_next_start(pos, config, *start_desc);
+                    to_next_start(pos, config, *start_desc, idle_func);
                 m_threads.push_next({first_inst, -1});
             }
             m_threads.swap_next();
         }
     }
 
-    static void to_next_start(Iterator& start, const ExecConfig& config, const StartDesc& start_desc)
+    static void to_next_start(Iterator& start, const ExecConfig& config, const StartDesc& start_desc, auto&& idle_func)
     {
         while (start != config.end)
         {
+            idle_func();
             static_assert(StartDesc::count <= 128, "start desc should be ascii only");
             if constexpr (forward)
             {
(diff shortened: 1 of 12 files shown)


mawww:
This prompted me to cleanup and push what I had in progress for this, see the `cancellation` branch.

I think it is a better base as I really do not want to couple the regex vm with the event handling (I still try to keep the regex implementation a bit separate from the rest of the code so that it can eventually be used out of Kakoune). My implementation used `<c-g>` for this to avoid killing a long `:make` build if Kakoune gets irresponsive due to some regex search in the mean time.

Both implementations are probably a bit brittle as its quite likely there is some code that is not properly exception safe and that will break if cancelled at the wrong moment. Those should be fixed as we discover them.

krobelus:
Great I'll work off that.

I guess we should poll stdin regularly in wall-clock time intervals to avoid spamming pselect.

> My implementation used <c-g> for this to avoid killing a long :make build if Kakoune gets irresponsive due to some regex search in the mean time.

Ok that wasn't on my radar (if my makecmd is not cheap to restart I usually run it in a separate session or terminal because I tend to cancel it accidentally).
Today `<c-c>` is the only key that interrupts UI hangs (when waiting for sh).
`<c-g>` is another obvious choice though we should probably try to overload `<esc>` with this behavior.

mawww:
> Ok that wasn't on my radar (if my makecmd is not cheap to restart I usually run it in a separate session or terminal because I tend to cancel it accidentally). Today `<c-c>` is the only key that interrupts UI hangs (when waiting for sh). `<c-g>` is another obvious choice though we should probably try to overload `<esc>` with this behavior.

Yep, I am not quite decided on if this should be `<c-c>` or a separate key, thats definitely something that can be discussed.

krobelus:
I've pushed something that should be close to the desired behavior.
There's some smaller corrections and cleanup left to do.

src/regex_impl.hh
```diff
@@ -478,9 +478,10 @@ private:
         m_found_match = false;
         while (true) // Iterate on all codepoints and once at the end
         {
+            idle_func();
+
```
mawww, line 482: Why did you change this ? I would expect that running a select syscall after every codepoint during regex matching is going to slow down things quite a bit. Doing it every 64k codepoints seemed regular enough for the cancellation use case, was that not the case ?

src/regex_impl.hh
```diff
@@ -511,17 +512,18 @@ private:
...
-                    to_next_start(pos, config, *start_desc);
+                    to_next_start(pos, config, *start_desc, idle_func);
                 m_threads.push_next({first_inst, -1});
             }
             m_threads.swap_next();
         }
     }
 
-    static void to_next_start(Iterator& start, const ExecConfig& config, const StartDesc& start_desc)
+    static void to_next_start(Iterator& start, const ExecConfig& config, const StartDesc& start_desc, auto&& idle_func)
```
mawww, line 522: Similarly, this should be fast enough even on very big buffer, checking for events on every byte is going to slow the general case quite a lot.

mawww:
Ohh, sorry, did not mean to close this, merged the cancellation branch into master as it is helpful as-it-is and removed that branch but that automatically closed that PR.

src/regex_impl.hh
krobelus: yeah I misread "wrap" as "wrap around buffer"

PR #4899, opened 2023-05-17
Fix redundant move warning with GCC 13.1.1

arrufat:
After upgrading GCC, this warning surfaced.


krobelus (approved):

mawww:
Missed this, already fixed.

PR #4900, opened 2023-05-17
Fix warning about ambiguous reversed operator

arrufat:
Apparently, the second argument is missing a const qualifier, so we need to make the method a friend function and explicitly make it const.

### DIFF
--- a/src/coord.hh
+++ b/src/coord.hh
@@ -42,16 +42,16 @@ struct LineAndColumn
     }
 
     [[gnu::always_inline]]
-    constexpr auto operator<=> (EffectiveType other) const
+    friend constexpr auto operator<=> (const EffectiveType lhs, const EffectiveType rhs)
     {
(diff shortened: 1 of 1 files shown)


src/coord.hh
```diff
@@ -42,16 +42,16 @@ struct LineAndColumn
     }
 
     [[gnu::always_inline]]
-    constexpr auto operator<=> (EffectiveType other) const
+    friend constexpr auto operator<=> (const EffectiveType lhs, const EffectiveType rhs)
```
krobelus, line 45: weird that it needs `const` but not `&`

krobelus (approved):

src/coord.hh
arrufat: I was also expecting that :S

mawww:
missed that as well, already fixed.

PR #4903, opened 2023-05-21
Allow canceling buffer highlighting

krobelus:
(I won't need this myself but maybe it's useful to others)

On huge files [1], initial highlighting can take a while.
The canonical solution is to disable highlighting for such files.
However we can improve the experience if the user forgot to do so

Allow regex highlighters to be interrupted with the interrupt key.
When that happens, mark the affected buffer and never again try to
highlight it with regex highlighters.

This commit is experimental because the behavior is difficult to
predict (it only affects some highlighters) but it could be quite
helpful for users who forgot to disable highlighting.

[1]: https://github.com/mawww/kakoune/issues/4685#issuecomment-1208129806

### DIFF
--- a/src/buffer.hh
+++ b/src/buffer.hh
@@ -217,6 +217,9 @@ public:
     };
     ConstArrayView<Change> changes_since(size_t timestamp) const;
 
+    void set_highlighting_interrupted() { m_highlighting_interrupted = true; }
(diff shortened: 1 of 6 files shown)


krobelus:
I'm not sure about this change because `<c-g>` should be like `<esc>`, never destructive.
OTOH cancelling only a single highlighting pass would be either ineffective or confusingly complex so I'm not sure

PR #4907, opened 2023-05-24
Suppress false positive error when <c-g> is pressed during shell execution

krobelus:
During shell command execution <c-g> does nothing but the error
message in the *debug* buffer suggests otherwise.  Fix this.

### DIFF
--- a/src/shell_manager.cc
+++ b/src/shell_manager.cc
@@ -363,6 +363,9 @@ std::pair<String, int> ShellManager::eval(
         {
             EventManager::instance().handle_next_events(EventMode::Urgent, &orig_mask);
         }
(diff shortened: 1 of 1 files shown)


mawww:
I am not sure what `<c-g>` should do during shell execution, I think it ought to cancel the operation by sending SIGINT to the shell process (and not the full process group, just the underlying shell that hopefully will kill subprocesses) and go back to the main loop. Does that make sense ?

krobelus:
Sounds good.

I'm thinking the `<c-g>` behavior should be added to `<esc>`.
I think emacs makes a distinction because escape can be included in key chains but we don't have that problem I think.
Not sure yet.

PR #4909, opened 2023-05-24
Allow to customize interrupt keys

krobelus:
Both `<c-c>` and `<c-g>` and are not mappable, in any mode.
(Despite `<c-g>` only being useful in normal mode). 

I like to map `<c-g>` in insert and normal mode.

Make this possible by adding a new "-interrupt" switch to map/unmap,
which allows to give arbitrary keys the interrupt behavior.

There is a slight behavior change here: `<c-c>` is *only* mapped in
normal mode, which means that it now acts as escape in insert, prompt
and menu modes (that code has been dead).

`<c-g>` is mapped in normal and prompt (to cancel incsearch).

Example usage:

	unmap -interrupt global normal <c-g>
	map -interrupt global normal <c-c> %{:cancel-subprocess; cancel-regex<ret>}
	map -interrupt global prompt <c-c> %{<a-semicolon>:cancel-subprocess; cancel-regex<ret>}

This commit is very hacky. It's likely not kosher to allow arbitrary
commands during regex execution because buffer modification could
trip up the regex.

Related: we could have a DWIM-style command that cancels regex search
if one is active, else sends SIGINT.


mawww:
Not too keen on that change in general, I am not sure there is enough to gain from remapping `<c-c>` and `<c-g>` to justify adding all of this scaffolding. Those keys are mostly escape hatches when things go wrong, the simpler they behave the better IMHO.

krobelus:
Yeah sorry this implementation is over the top.
I still think it's valuable to be able to remap the keys at runtime.
The fact that `<c-c>` is not mappable surprises users from time to time.
It'd be nice to not need to patch the editor here.
I'd like to experiment with mapping it to `:quit` or copy-to-clipboard.

What if we mimicked `stty's` `intr` setting?

```
declare-option str interrupt <c-c> 
declare-option str cancel <c-g>
```

still a bit ugly because the option only makes sense in the global scope I think

PR #4929, opened 2023-06-20
Add normal commands to extend selections downwards/upwards.

arrufat:
Since we changed the `x` keys behaviors, the `X` and `<a-X>` have been free.

I think it makes sense to make them work mostly like `/` and other commands, where shift extends the selection and alt changes the direction.

So, the implemented behavior is:
`<X>`   : select the current and next line (extend downwards)
`<a-X>` : select the current and previous line (extend upwards)

Overall, I think it's a good change that's consistent with the way other keys (and modifiers) behave in Kakoune.

Note: in this implementation, pressing once `<X>` or `<a-X>` will result in two lines being selected (regardless of the current line being completely selected). This makes the keys more predictable, but it might be surprising at the beginning, if someone expects the old `X` behavior.

### DIFF
--- a/doc/pages/changelog.asciidoc
+++ b/doc/pages/changelog.asciidoc
@@ -5,6 +5,9 @@ released versions.
 
 == Development version
 
+* `<X>` and `<a-X>` to expand line selections downwards and upwards,
+  respectively. 
+
 * History is now stored linearly instead of in a tree
 
 * `<a-u>` and `<a-U>` now undo selection history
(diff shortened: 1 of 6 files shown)


mawww:
I am not too keen on that change, free keys are very scarce and kakoune normal mode is already quite complex, I am not convinced adding a slightly different way of doing `Jx` and `Kx` is worth it.

arrufat:
Ok, got it. Just wanted to make sure this was considered. Thank you :)

PR #4935, opened 2023-06-24
map: fail if key is currently executing

krobelus:
If during execution of a mapping, that same mapping is replaced,
there is undefined behavior because we destroy a mapping that we are
still iterating over.

I have been using this mapping inside my kakrc to re-source the kakrc.

	map global user s %{:source "%val{config}/kakrc"<ret>} -docstring 'source "%val{config}/kakrc"'

Now <space>s happens to not trigger undefined behavior because the
mapping stays the same.

However it triggers an assertion added by
Commit e49c0fb04 (unmap: fail if the mapping is currently executing,
2023-05-14), specifically the destructor of ScopedSetBool that
guards mapping execution.

Fix both both by making sure mappings outlive mapping execution;
garbage-collect them later.

Add a constructor to make it work with Clang 11.

The extra indirection is not nice but we can get rid of it by using
an arena to allocate KeymapInfo's.

We can now support unmap during mapping execution (i.e. remove the
is_executing check).  I kept it because I'm not sure what's the better
behavior. The error means it can prevent accidents but if there's a
sensible use case we should probably support this.


mawww:
I think it would be better to refuse to map a key that is already executing. And then for your use case wrapping the `map global user s` into a try block would prevent failing to reload. I think that would be much simpler implementation and also would avoid confusion, as it is unclear from when a new mapping ought to take effect.

krobelus:
> would avoid confusion, as it is unclear from when a new mapping ought to take effect.

I think it's fairly natural that the new mapping activates immediately while active mapping executions in an outer stack frame still work with the old version.
That being said maybe this is too subtle, so I'm fine with banning it for now. The scenarios where it will matter will be very rare

PR #4941, opened 2023-07-01
lisp highlighter is marking everything as variable

ericv8v9s:
The following line in `lisp.kak`
```
add-highlighter shared/lisp/code/ regex (#?(['`:]|,@?))?\b[a-zA-Z][\w!$%&*+./:<=>?@^_~-]* 0:variable
```
is marking keywords like `nil`, `true`, `false`, `def`s as variables.

Moving the highlighter to before the others seems to fix the issue, though I'm not familiar with kak config enough to know if that's a proper fix.


mawww:
I am not sure highlighting every identifier was providing much value in the first place, but I agree fixing the ordering makes things better. Thanks

PR #4945, opened 2023-07-06
Allow map/unmap during mapping execution

krobelus:
Commits e49c0fb04 (unmap: fail if the mapping is currently executing,
2023-05-14) 42be0057a (map: fail if key is currently executing,
2023-06-24) fixed potential use-after-free issues. By doing so,
it broke configurations that in practice have not triggered any
crashes see
https://github.com/kak-lsp/kak-lsp/issues/689 and
https://github.com/alexherbo2/auto-pairs.kak/issues/60

For example with,

	set -remove global autocomplete insert
	hook global InsertCompletionShow .* %{
	    map window insert <esc> <c-o>
	}
	hook global InsertCompletionHide .* %{
	    unmap window insert <esc> <c-o>
	}

The execution of the <esc> mapping triggers InsertCompletionHide fails
at unmapping. This seems legit and I don't see an obvious alternative
way to write it (InsertIdle would not be correct though it would work
in practice).

Fix the regression by allowing map and unmap again while keeping the
mappings alive until garbage collection. This is similar to how we
treat hooks and others.

Applying map/unmap immediately seems like the most obvious semantics.
Alternatively, we could apply them in between key presses.

To-do: we can probably clean up the KeymapManager a bit.


src/keymap_manager.hh
```diff
@@ -40,9 +40,8 @@ public:
     KeyList get_mapped_keys(KeymapMode mode) const;
 
     auto get_mapping_keys(Key key, KeymapMode mode) {
-        struct Keys : ConstArrayView<Key> { ScopedSetBool executing; };
         auto& mapping = get_mapping(key, mode);
-        return Keys{mapping.keys, mapping.is_executing};
+        return mapping.keys;
```
mawww, line 44: This is returning the keys by value which means we do a copy here, which removes the need for any trash management. That might be a better approach, I am not sure avoiding the allocation and copy of the mapping will have a significant performance impact.

krobelus: Yeah copying is fine. We could also use a reference-counted pointer

PR #4948, opened 2023-07-20
Add filetype for ember js .gjs and .gts files

velrest:
Im trying to add highlighting for [`.gjs` and `.gts`](https://github.com/ember-template-imports/ember-template-imports) files. I copied the `.js` and `.hbs` file and started to hack around a bit.
But im now getting errors that my closing expansion characters are not commands. Could someone help me out on this?

This is just an attempt and by no means complete or anything yet. I just thought i would ask if someone could help me out on this since I've been struggling with this for some times.

Example of an `.gjs` file:
```gjs
import { on } from '@ember/modifier';
import FancyButton from './fancy-button';

function greet() {
  alert("AHOY!")
}

<template>
  <p>Hello, {{@name}}!</p>
  <FancyButton @label="Say hello!" {{on "click" greet}} />
</template>
```

### DIFF
--- a/rc/filetype/gjs.kak
+++ b/rc/filetype/gjs.kak
@@ -1,0 +1,65 @@
+# https://github.com/ember-template-imports/ember-template-imports
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+# Detection
+# ‾‾‾‾‾‾‾‾‾
+
+hook global BufCreate .*[.](gjs) %{
+    set-option buffer filetype gjs
+}
+
+hook global BufCreate .*[.](gts) %{
+    set-option buffer filetype gts
+}
+
+# Initialization
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+hook global WinSetOption filetype=(gjs|gts) %{
+    require-module javascript
+
+    hook window ModeChange pop:insert:.* -group "%val{hook_param_capture_1}-trim-indent" javascript-trim-indent
+    hook window InsertChar .* -group "%val{hook_param_capture_1}-indent" javascript-indent-on-char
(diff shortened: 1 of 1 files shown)


Screwtapello:
Kakoune's scripting language has a number of ways of writing strings. There's `''` and `""` of course, but also things like `%{}` and `%@@`.

In particular, a `%{}` string (or any other ASCII grouping character, like `%()` or `%[]` or `%<>`) keeps going until you hit the matching half of the pair. So `%{abcd}` is the same as `'abcd'` and `%{ab%{}cd}` is the same as `'ab%{}cd'`. A `%@@` string (or any other non-grouping character) keeps going until it hits the next instance of that character.

On line 24, you start a block with `%~`. Since that's a non-grouping character, the block will end at the next `~` on line 42. That's almost certainly not what you intended, so it doesn't surprise me that you're getting confusing error messages.

If at all possible, use ASCII grouping characters in your strings, since they nest nicely. If you can't, for example because you have regexes that use a lot of the same characters, then you'll need to use a different non-grouping character for each level of nesting, to prevent this kind of accidental block-ending-early business. Note that you don't have to use ASCII non-grouping characters, some plugins use symbols like § or emoji to keep things visually distinctive.

velrest:
Ah thanks! I misunderstood the following that all characters , if balances, can be used inside. But that's of course only for grouping characters.
> Nested pairs of the braces used in the expansion are allowed, but they must be balanced.

velrest:
This would be ready for review now.

mawww:
Looks good, I do not think I have a copyright waiver commit from you yet, could you add one on that PR ? See the CONTRIBUTING file.

velrest:
Hey @mawww 
It's been a long time but i added the waiver commit to the PR.
Thanks for leaving this open so long.

velrest:
@mawww Thanks!

PR #4950, opened 2023-07-26
zig: update builtin functions for 0.11.0

arrufat:
Zig 0.11.0 is set to be released on [August 3rd, 2023](https://ziglang.org/news/0.11.0-postponed-again/).

I grabbed the builtin functions from the [official Vim syntax file](https://github.com/ziglang/zig.vim/blob/master/syntax/zig.vim).

I've also updated their type from `function` to `meta`. I think it makes more sense, since these functions are implemented by the compiler.

This is also up for discussion in case people prefer them to be highlighted as normal functions.


arrufat:
Ah, I was thinking maybe @ifreund has an opinion on this.

ifreund:
In any case, I don't have a strong opinion here, it's trivial for me to get the behavior I want for my personal usage.

PR #4968, opened 2023-09-02
Add kak-connect to connect external programs as file browsers

stacyharper:
Kakoune doesn't embark a file browser. It has been designed as client-server, and this make it easy to connect to external programs.

The problem is that there is no easy way to do so out of the box. The users are expected to glue this themselves.

This try to help users to connect external softwares to Kakoune.

:connect is a new command to be used with :terminal or :run to start a new connected terminal, or graphical application.

Those connected software have the corresponding KAKOUNE_SESSION and KAKOUNE_CLIENT environment variables. It also set kak-connect as EDITOR.

kak-connect is a simple POSIX shell script that send commands to the connected Kakoune session, to open the files.

The basic usage of all could be:

:connect terminal nnn
:connect terminal ranger

Which open nnn or ranger in another terminal (external, or in a tmux pane), and to open the files in the Kakoune client.

We can itterate over this:

- Add hooks to open a file browser when the user open a directory, or no file with :edit.
- Make nnn, or other programs detected and configured by default, as we do in windowing terminals. And wrap it with a :file-browser command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -180,6 +180,8 @@ installdirs-debug-yes: installdirs-debug-no
 install: src/kak installdirs install-debug-$(debug) install-gzip-man-$(gzip_man)
 	cp src/kak$(suffix) $(bindir)
 	chmod 0755 $(bindir)/kak
+	cp src/kak-connect$(suffix) $(bindir)
+	chmod 0755 $(bindir)/kak-connect
 
 	ln -sf ../../bin/kak $(libexecdir)/kak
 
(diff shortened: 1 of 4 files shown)


robem:
I've been using https://github.com/kkga/kks for a while and it works great.
You can use `kks-connect` to open `nnn` and select a file to run in the same client.
```sh
kak -e "kks-connect terminal nnn -e"
```
If you are looking to build this functionality into `kakoune` itself then it might be worth looking at `kks`.

---
Edit: While playing more with `nnn` in `kakoune`. I ended up ditching `kks-connect` for this:
```kak
map global user e ':tmux-repl-vertical "echo eval -client %val{client} \""edit $(nnn -o -p -)\"" | kak -p %val{session}"<ret>' -docstring 'NNN'
```

stacyharper:
I've just tested kks, and it is of interset. The scripts cover a broader set of feature. Both relate in their minimal use -case, which is helping the user to connect another environment with the EDITOR.

What I'm trying to upstream still fit better imo, and because is POSIX compatible, should work immediately with extra dependency.

Delapouite:
As @alexherbo2 did a lot of stuffs around connect back in the day, maybe he's interested by reviewing this PR.

PR #4972, opened 2023-09-06
Add options for changing make.kak error patterns

lf94:
This makes it easier to add your own matchers for whatever build tools you're using with Makefiles.

I haven't tested the highlighting since I don't use it .

### DIFF
--- a/rc/tools/make.kak
+++ b/rc/tools/make.kak
@@ -1,7 +1,15 @@
 declare-option -docstring "shell command run to build the project" \
     str makecmd make
-declare-option -docstring "pattern that describes lines containing information about errors in the output of the `makecmd` command" \
-    str make_error_pattern " (?:fatal )?error:"
+declare-option -docstring "pattern to find the next error" \
+    str make_error_pattern "^(?:\w:)?[^:\n]+:\d+:(?:\d+:)? (?:fatal )?error:"
+declare-option -docstring "pattern for highlighting errors" \
+    str make_error_highlight_pattern "(?:\w:)?[^:\n]+):(\d+):(?:(\d+):)"
+declare-option -docstring "pattern to capture information for navigating errors" \
+    str make_open_error_pattern "((?:\w:)?[^:]+):(\d+):(?:(\d+):)?([^\n]+)\z"
+declare-option -docstring "to determine if the build system entered a directory when building" \
+    str make_dir_pattern "Entering directory"
+declare-option -docstring "similar to make_open_error_pattern but for nested build files" \
+    str make_open_dir_error_pattern "Entering directory [`']([^']+)'.*\n([^:/][^:]*):(\d+):(?:(\d+):)?([^\n]+)\z"
 
 declare-option -docstring "name of the client in which utilities display information" \
     str toolsclient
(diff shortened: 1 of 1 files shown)


lf94:
(I had to force-push to rearrange the commits for the copyright waiver.)

rc/tools/make.kak
```diff
@@ -53,13 +61,13 @@ define-command -hidden make-open-error -params 4 %{
 define-command -hidden make-jump %{
     evaluate-commands %{
         try %{
-            execute-keys gl<a-?> "Entering directory" <ret><a-:>
+            execute-keys gl<a-?> %opt{make_dir_pattern} <ret><a-:>
             # Try to parse the error into capture groups, failing on absolute paths
-            execute-keys s "Entering directory [`']([^']+)'.*\n([^:/][^:]*):(\d+):(?:(\d+):)?([^\n]+)\z" <ret>l
+            execute-keys s %opt{make_open_dir_error_pattern} <ret>l
```
krobelus, line 66: not sure if this is worth the breaking change.
I personally override the make-* commands for projects where I need different behavior

lf94: What would be the breaking change? The string has just moved to an option?...

lf94: > I personally override the make-* commands for projects where I need different behavior

I've done similar but it's nice to have consistent behaviors for make.kak and just needing to have to change the patterns... 

Again if there's a breaking change I'd love to figure out how to not make it breaking change :D

krobelus: sorry not on this line but `make_error_pattern` has a new behavior

rc/tools/make.kak
```diff
@@ -25,7 +33,7 @@ define-command -params .. \
 }}
 
 add-highlighter shared/make group
-add-highlighter shared/make/ regex "^((?:\w:)?[^:\n]+):(\d+):(?:(\d+):)?\h+(?:((?:fatal )?error)|(warning)|(note)|(required from(?: here)?))?.*?$" 1:cyan 2:green 3:green 4:red 5:yellow 6:blue 7:yellow
+add-highlighter shared/make/ regex "^%opt{make_error_highlight_pattern}?\h+(?:((?:fatal )?error)|(warning)|(note)|(required from(?: here)?))?.*?$" 1:cyan 2:green 3:green 4:red 5:yellow 6:blue 7:yellow
```
mawww, line 36: This will not work, you'd need to have a hook that changes this highlighter whenever that option changes, and it will be made trickier as that option might have different values in different windows which means it cannot really be shared. `dynregex` might help but I have been wanting to get rid of it for a while so I would not encourage using it. Not sure what the solution would be here

rc/tools/make.kak
```diff
@@ -53,13 +61,13 @@ define-command -hidden make-open-error -params 4 %{
 define-command -hidden make-jump %{
     evaluate-commands %{
         try %{
-            execute-keys gl<a-?> "Entering directory" <ret><a-:>
+            execute-keys gl<a-?> %opt{make_dir_pattern} <ret><a-:>
```
mawww, line 64: Putting this in an execute key is dangerous as and `<foo>` in the regex might get misinterpretted as a key spec. The correct way is to `set-register slash %opt{make_dir_pattern}` then `execute-keys gl<a-?><ret><a-:>`.

mawww (commented):
I am not too sure about this change, it seems the introduced options have many constraints that are not obvious (the order of captures for example), and I cannot see how to solve the highlighting issue.

rc/tools/make.kak
lf94: I'll fix it up!

rc/tools/make.kak
lf94: Roger that o7. I'll remove this change and leave it as it was.

rc/tools/make.kak
lf94: Will do

lf94:
> I am not too sure about this change, it seems the introduced options have many constraints that are not obvious (the order of captures for example), and I cannot see how to solve the highlighting issue.

This is a good point. Let me come back to this with a new proposal in the next week or two. :) With this initial review I know better on what to focus on. The initial focus was just "let people change the patterns". The next focus will be on being more explicit about the pieces of the patterns. Thank you for being quick on the review!

krobelus:
should be implemented now with https://github.com/mawww/kakoune/pull/5007
In future we can also update highlighting

lf94:
Nice! Good enough for my needs :)

PR #4974, opened 2023-09-07
Support setting diff characters in git diffs.

phaazon:
This commit adds `diff_add_char`, `diff_mod_char`, `diff_del_char` and `diff_top_char` as `str` options, with typical defaults.

This commit also replaces the hard coded +, _, ≃, etc. hardcoded characters in `git update-diff` to use the options from above.

### DIFF
--- a/doc/pages/options.asciidoc
+++ b/doc/pages/options.asciidoc
@@ -388,22 +388,6 @@ are exclusively available to built-in options.
     relating to a Kakoune version greater than this value will be displayed. Versions
     are written as a single number: Like `20180413` for version `2018.04.13`
 
-*diff_add_char* `str`::
-	  _default_ ▏ +
-	  Character used to denote added lines in diffs (e.g. git gutter).
-
-*diff_mod_char* `str`::
-	  _default_ ▏ +
-	  Character used to denote modified lines in diffs (e.g. git gutter).
-
-*diff_del_char* `str`::
-	  _default_ _ +
-	  Character used to denote deleted lines in diffs (e.g. git gutter).
-
-*diff_top_char* `str`::
-	  _default_ ‾ +
-	  Character used to denote the first line deleted in diffs (e.g. git gutter).
-
 == Current values
 
 The current value for an option can be viewed using
--- a/doc/pages/options.asciidoc
+++ b/doc/pages/options.asciidoc
@@ -388,6 +388,22 @@ are exclusively available to built-in options.
     relating to a Kakoune version greater than this value will be displayed. Versions
     are written as a single number: Like `20180413` for version `2018.04.13`
 
+*diff_add_char* `str`::
+	  _default_ ▏ +
+	  Character used to denote added lines in diffs (e.g. git gutter).
+
+*diff_mod_char* `str`::
+	  _default_ ▏ +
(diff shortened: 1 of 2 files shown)


doc/pages/options.asciidoc
```diff
@@ -388,6 +388,22 @@ are exclusively available to built-in options.
     relating to a Kakoune version greater than this value will be displayed. Versions
     are written as a single number: Like `20180413` for version `2018.04.13`
 
+*diff_add_char* `str`::
+	  _default_ ▏ +
+	  Character used to denote added lines in diffs (e.g. git gutter).
```
krobelus, line 393: doesn't seem unreasonable to add.
One alternative to these options is to add parameters to the `show-diff` and `update-diff` commands

phaazon: Yes, I thought about that too. Which direction would we prefer to go regarding this kind of things? On one side, having them as options seems nice in terms of sharing, but on the other, we shouldn’t try to make kakoune an IDE. I was honestly surprised that the `git` wrapper commands existed in the first place!

mawww: AFAIR we do not document non-core options here, those are support script specific so they should not be in `pages/options.asciidoc`

rc/tools/git.kak
```diff
@@ -1,6 +1,18 @@
...
+    str diff_add_char "▏"
+
+declare-option -docstring "diff modified character" \
+    str diff_mod_char "▏"
+
+declare-option -docstring "diff deleted character" \
+    str diff_del_char "▁"
+
+declare-option -docstring "diff top deleted character" \
+    str diff_top_char "‾"
```
mawww, line 14: Not too sure about the names, maybe `git_diff_` prefix would be more appropriate.

mawww (commented):
I am not sure how much value there is in making those configurable, I am happy to change their value to what is proposed as default here but its unclear adding options to control those four characters is worth it. Are there more opinions on this ?

CharlesGueunet:
@mawww You asked for feedback so I will answer here as I love the idea. On my fork of Kakoune, there are only two changes, and [this is one of them](https://gitea.gueu.net/charles.gueunet/kakoune/commit/fbeb00192954f758b7eac158be11b4e796a30139). The version written here is clearly more flexible.

phaazon:
> I am not sure how much value there is in making those configurable, I am happy to change their value to what is proposed as default here but its unclear adding options to control those four characters is worth it. Are there more opinions on this ?

I do think that _most people_ should be happy with those default values, but I guess some people might also like the `+` and `-` ones, as it looks like regular diff; hence why I made it configurable.

I’m updating the PR with your comments about naming and I’m removing the docs.

phaazon:
I preferred adding new commits to keep the history in case we want to backtrack but if this is eligible for a merge, I’ll squash everything into a single commit if you prefer.

mawww:
I dont think I have a copyright waiver commit from you, could you add one on this PR please ?

phaazon:
> I dont think I have a copyright waiver commit from you, could you add one on this PR please ?

Here you go [9e4dbdb](https://github.com/mawww/kakoune/pull/4974/commits/9e4dbdbe6c017f0dc823b030931db06ed7ba7632).

PR #4977, opened 2023-09-11
Add <C-e> and <C-y> to scroll one line at a time

stacyharper:
Recently I noticed changes with the behavior of my keybinds:

```
map global normal <C-e> vj
map global normal <C-y> vk
```

This is because the cursor movement behavior changed with vj and vk. Scrolling window was previously blocked by the cursor. Now the cursor and selection stays in place.

I think the view keyboard behaviors are consistent, and coherent with the current state. But this isn't exactly what we want from `<C-e>` and `<C-y>`.

Example, when you start to scroll with `<C-e>` (your cursor stay in place), then you press `<C-d>` (your cursor is moved back to your view).

Example, when you start to scroll with `<C-e>` (your cursor stay in place), then you press `j` or `k` (your view is pushed back to initial position).

To solve this, I think we need news and differents default keybinds `<C-e>` and `<C-y>`.

### DIFF
--- a/src/context.hh
+++ b/src/context.hh
@@ -25,6 +25,7 @@ class KeymapManager;
 class HookManager;
 
 enum Direction { Backward = -1, Forward = 1 };
+enum Distance { FullScreen = 0, HalfScreen = 1, Line = 2 };
 
 struct JumpList
 {
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -452,6 +452,8 @@ See <<Appending>> below for instructions on extending (appending to) the current
  * `pagedown, <c-f>`: scroll one page down
  * `<c-u>`: scroll half a page up
  * `<c-d>`: scroll half a page down
+ * `<c-y>`: scroll one line up
+ * `<c-e>`: scroll one line down
 
  * `)`: rotate selections (the main selection becomes the next one)
  * `(`: rotate selections backwards
--- a/doc/pages/keys.asciidoc
+++ b/doc/pages/keys.asciidoc
@@ -193,6 +193,12 @@ the Shift modifier and moving will extend each selection instead.
 *<c-d>*::
     scroll half a page down
 
+*<c-y>*::
+    scroll one line up
+
+*<c-e>*::
+    scroll one line down
+
 *;*, *<semicolon>*::
     reduce selections to their cursor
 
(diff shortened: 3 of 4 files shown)


src/context.hh
```diff
@@ -20,6 +20,7 @@ class KeymapManager;
 class AliasRegistry;
 
 enum Direction { Backward = -1, Forward = 1 };
+enum Velocity { FullScreen = 0, HalfScreen = 1, Line = 2 };
```
evanrelf, line 23: "Velocity" is defined as "the speed of something in a given direction". I think "distance" would be a better name here.

stacyharper: Fair enough, I renamed this. Thanks!

stacyharper:
I just rebased this. Could this be reviewed again?

mawww:
The implementation looks fine, but I am wondering if we need to dedicate yet another binding to this. Would it be okay to rely on  passing a line count to another command (say `1<c-b>` to move one line instead of a full screen)

stacyharper:
imo `3<c-b>` would make more senses as "3 pages back". I'm really not convinced

mawww:
I do agree with it would be more consistent, but we already have a binding for scrolling page by page, then a binding for scrolling half page by half page, and now we want one for scrolling line by line. I think this is too many mappings for this and would like to get some consolidation. Especially as this is on top of a whole set of other mappings in view mode.

Another direction would be to say that this is a view command and we should have a way to decide how the cursor should behave. An idea that comes to mind is to have say `<a-v>` that enters view mode with an extra flag to move cursors if necessary.

stacyharper:
> I do agree with it would be more consistent, but we already have a binding for scrolling page by page, then a binding for scrolling half page by half page, and now we want one for scrolling line by line. I think this is too many mappings for this and would like to get some consolidation. Especially as this is on top of a whole set of other mappings in view mode.

I don't share your opinion on if it is too many mapping or not. To me there is 3 distinct way to scroll the buffer, and they are very common, and present on other softwares. Plus we managed to refactorise the code, to avoid creating too many functions for them.

Do you fear about missing available shortcuts?

> Another direction would be to say that this is a view command and we should have a way to decide how the cursor should behave. An idea that comes to mind is to have say <a-v> that enters view mode with an extra flag to move cursors if necessary.

I'm not sure to fully understand what you propose here. You mean to add a dedicated mode, that offer different ways to scroll? And the users could map a shortcut to this if they want to? (`<c-e>` in my case `<c-v>j`).

But anyway, imo we still need an internal command to scroll one line, that respect the behavior I've mentioned in the body here.

PR #4978, opened 2023-09-12
rc/tools/git: enable <ret> to diff-jump

stacyharper:
As we require diff here, it make using git log -p very convenient to go to the source files.

### DIFF
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -95,7 +95,7 @@ define-command -params 1.. \
            diff) map_diff_goto_source=true; filetype=diff ;;
            show) map_diff_goto_source=true; filetype=git-log ;;
            show-branch) filetype=git-show-branch ;;
-           log)  filetype=git-log ;;
+           log)  map_diff_goto_source=true; filetype=git-log ;;
            status)  filetype=git-status ;;
            *) return 1 ;;
         esac


rc/tools/git.kak
```diff
@@ -10,6 +10,8 @@ hook -group git-log-highlight global WinSetOption filetype=git-log %{
     add-highlighter window/git-log/ ref diff # highlight potential diffs from the -p option
 
     hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/git-log }
+
+    map buffer normal <ret> :diff-jump<ret>
```
krobelus, line 14: how about

```diff
diff --git a/rc/tools/git.kak b/rc/tools/git.kak
index b1ec8c11c..b326ae6ff 100644
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -94,7 +94,7 @@ define-command -params 1.. \
            diff) map_diff_goto_source=true; filetype=diff ;;
            show) map_diff_goto_source=true; filetype=git-log ;;
            show-branch) filetype=git-show-branch ;;
-           log)  filetype=git-log ;;
+           log)  map_diff_goto_source=true; filetype=git-log ;;
            status)  filetype=git-status ;;
            *) return 1 ;;
         esac
```

then it will use the right root directory

stacyharper: That's better! My bad. Thanks

PR #4984, opened 2023-09-18
rc/tools/git add cat-file

QiBaobin:
Support check out current file content in another branch/commit, like: 
`git cat-file blob ohter-branch:<c-r>%`

### DIFF
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -52,6 +52,7 @@ define-command -params 1.. \
             rm
             reset
             blame
+            cat-file
             commit
             checkout
             diff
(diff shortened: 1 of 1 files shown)


mawww:
`cat-file` does not strike me as something that is very regularly used, could you at least describe the use case you have for it ?

QiBaobin:
I don't use it a lot either, my use case is that I want checkout a file content in another commit, it's similar with using Gread in vim-fugitive.

krobelus:
I use `git show origin/master:path/to/file` for that

QiBaobin:
Good to know, thanks! Let me close this then.

Issue #4982, opened 2023-09-18
[QUESTION] How to set the foreground color of a face to match the terminal background


lobre:
### Question

I understand that using `default` in a face will result in using the terminal default color.

```
Default
    default colors
        The default foreground and background colors. If the value of default is used here, then the colors used will be your terminal's defaults.
```

So doing this will use the terminal foreground default color:

```
set-face global LineNumbersWrapped default,black
```

While doing this will use the terminal background default color:

```
set-face global LineNumbersWrapped black,default
```

However, can I set my face's **foreground** color to match the terminal's default **background**?

My use case is that I want to hide line numbers for lines that are soft-wrapped, but the background of my face for `LineNumbers` is `default`.

Using `reverse` does not do what I want to do.

Related issue:

https://github.com/mawww/kakoune/issues/4004

Screwtapello:
Unfortunately, this is not possible. There's only one "default" colour, and the terminal decides whether to draw the default foreground or the default background colour based on where it's used, and the state of the `reverse` flag. If you want to perfectly hide wrapped line numbers, you'll need to set a specific background colour for the non-wrapped numbers, and use the same foreground and background colour for the wrapped numbers.

Alternatively, if you don't mind the colours not being *perfectly* hidden, you can just say "my default terminal background is close to black, so I'll just make the wrapped numbers `black,default`.

lobre:
Thanks for confirming my guess saying that it is not currently possible.

Coming back to my precise desired behaviour, which is to **hide the line number on soft-wrapped lines**, do you think it could be something that could be added to Kakoune?

It is about visual preferences, but I prefer not to have them displayed so that I can clearly distinguish wrapped lines from regular ones.

I guess that some people would prefer to change the face to have another way to identify them (for example with another colour). But having an option to hide them completely makes sense to me.

I am not familiar with the code base and with c++ in general, but after a few minutes, I am able to quickly hack a dirty solution to have them hidden.

Here is a modification of the following function.

https://github.com/mawww/kakoune/blob/01fd6552af6b909f533d5cd8bf3a5385d4e35215/src/highlighters.cc#L1108

```
    void do_highlight(HighlightContext context, DisplayBuffer& display_buffer, BufferRange) override
    {
        ...

        char format[16];
        format_to(format, "%{}d", digit_count);

        /* Create a buffer to contain the correct amount of empty spaces */
        char empty[16];
        format_to(empty, "%{}s", digit_count);

        ...

[...]

PR #4991, opened 2023-09-28
Hide line number on wrapped lines

lobre:
I started discussing this topic on that issue: https://github.com/mawww/kakoune/issues/4982#issuecomment-1723428016

This is a draft, so it does not implement the whole behavior, and it is really rough. This is more to discuss the idea, and the implementation can be reviewed then.

This is about trying to hide line numbers on lines that are wrapped. As of today, line numbers are duplicated when lines are wrapped. This duplication can be desired for some users. But for some others, they prefer having them hidden, to clearly distinguish a regular line from a wrapped one.

So the idea of this PR is to draw empty spaces in replacement of the line number for wrapped lines. And this could be activated using an option on the `number-lines` highlighter.

What do you think?

See the result:

wrap_lines_double

wrap_lines_hidden

### DIFF
--- a/src/highlighters.cc
+++ b/src/highlighters.cc
@@ -1118,6 +1118,8 @@ private:
 
         char format[16];
         format_to(format, "%{}d", digit_count);
+        char empty[16];
+        format_to(empty, "%{}s", digit_count);
         const int main_line = (int)context.context.selections().main().cursor().line + 1;
         int last_line = -1;
         for (auto& line : display_buffer.lines())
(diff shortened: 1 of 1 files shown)


ftonneau:
I wonder whether this proposal if necessary, because hiding line numbers can be easily done within any color scheme with `set-face global LineNumbersWrapped ...`. To hide wrapped line numbers, just set their foreground color equal to the background color for `LineNumbers`.

lobre:
If you check the comment I mentioned in my message, you can see that this workaround does not work if you are using the default color for  `LineNumbers` unfortunately.

ftonneau:
You are correct -- however, another workaround in the case of `LineNumbers` with the default background is to color the wrapped line numbers white (in the case of a light theme) or black (in the case of a dark theme). The wrapped line numbers will be virtually invisible even though their color is technically different from that of of the background. (I know because I tried!) In any case, the issue can always be dealt with by tuning a color scheme.

rhelmot:
I would like to see if this can be continued, since none of these workarounds work if your default terminal background color is meant to have transparency. I would be willing to complete this PR if I could get the blessing that such a feature would actually be accepted.

PR #4997, opened 2023-10-15
Add a contrib/gendocs.py script for generating documentation web site

igor-ramazanov:
Hey, I think Kakoune is a great editor!

I struggled with a searchable and mobile-phone friendly documentation.

Here're a Bash and Scala scripts to use the existing `**/*.asciidoc`
for generating the static website.

The code is far from ideal, but should be a good start.

Would be good to improve the script and automatically publish the
documentation using Github Actions and Pages.

Hosting the draft on my personal Github pages.

Here's how it looks like: https://igor-ramazanov.github.io/.


mawww:
Hey, this is great, I am just a bit concerned about the use of scala, I would have preferred a language that almost always available on posix platforms (such as sh/perl/python), but this is contrib/ so I guess it is fine.

igor-ramazanov:
Hi, yes, I would like to remove Scala and use something more standard too.

Tried first with `awk`, but it was difficult for me as I am not that profficient in it.

Let me check if I can replace Scala with `sh/sed/grep/awk`!

krobelus:
Maybe writing the entire script in Python is better than mixing languages. But your call, whatever works

igor-ramazanov:
@mawww @krobelus Hi, I've rewritten the script to Python 3.

Now it does not depend on `fd`, `scala-cli`, `sed` and `awk`, uses only Python 3 and https://antora.org.

contrib/gendocs.py
```diff
@@ -0,0 +1,297 @@
...
+
+
+# Antora fails if the repo contains broken symbolic links.
+shutil.rmtree("libexec", ignore_errors=True)
+
+# Canonical Antora paths.
+# See: https://docs.antora.org/antora/latest/standard-directories.
+#      https://docs.antora.org/antora/latest/root-module-directory.
+os.makedirs("modules/ROOT/images", exist_ok=True)
+os.makedirs("modules/ROOT/pages", exist_ok=True)
```
krobelus, line 65: FWIW instead of doing an in-tree build that requires renaming of some files,
I'd prefer to creating a separate directory (`html_doc`?),
copy all asciidoc files there and put all build artifacts there.
Not renaming files  avoids confusing editors etc.

contrib/gendocs.py
```diff
@@ -0,0 +1,297 @@
...
+        fragment = dropwhile(lambda c: c != "#", prefix)
+        fragment = fragment if fragment else None
+        fragmentless = takewhile(lambda c: c != "#", prefix)
+        segments = fragmentless.split("/")
+        path, file = (
+            ("/".join(segments[:-1]), segments[-1])
+            if "/" in fragmentless
+            else (None, fragmentless)
+        )
+        return Link(path, file, fragment, title)
```
krobelus, line 128: Actually GitHub is decent at rendering asciidoc, they also support the cross references.
Though the dedicated site definitely looks nicer and is more discoverable.
Maybe we can add a link to it on the website.
It does have a large overlap with github so I'm not sure it will gain a lot of momentum.
But it's nice to have a replacement we can custommize as we add more docs

contrib/gendocs.py
```diff
@@ -0,0 +1,297 @@
...
+    url: https://gitlab.com/antora/antora-ui-default/-/jobs/artifacts/HEAD/raw/build/ui-bundle.zip?job=bundle-stable
+"""
+
+with open("antora-playbook.yml", "w") as f:
+    f.write(antora_playbook_content)
+
+# Finally, generate the documentation,
+# results will be saved to the output directory
+# as specified in the `antora-playbook.yml`.
+subprocess.run(["antora", "generate", "antora-playbook.yml"])
```
krobelus, line 295: In addition to broken symlink, antora hangs indefinitely if there is a fifo in $PWD.
Also on a clean secondary worktree, it failed with

    [23:39:47.976] FATAL (antora): Local content source must be a git repository: /home/johannes/git/kakoune2 (url: ./)

so it doesn't recognize that .git is a gitlink.
I'm not sure why it even interacts with Git.

contrib/gendocs.py
igor-ramazanov: @krobelus Hey, long time. I've finally did that, now it keeps everything in the `doc_gen` dir without modifying the worktree.

contrib/gendocs.py
igor-ramazanov: @krobelus I'm not sure why it's been behaving like that,
but after keeping the work in the dedicated `doc_gen/`
I didn't have to remove the symlink,
so maybe this issue is gone now?

Could you verify it again, please?

PR #5000, opened 2023-10-17
Add ctrl left/right word navigation

JasonBrownDeveloper:


PJungkamp:
Why would you do this?

```kak
map global prompt <c-right> <a-f>
map global prompt <c-left> <a-b>
```

If you want to change the default shipped with kakoune you should probably add a description with your argument and some clarification!

I needed to check whether control+arrow navigation is already supported in prompt mode, it really isn't there.
I'm in general pretty indifferent towards this change and don't really see the need.

JasonBrownDeveloper:
I'll check if your map covers what I want, I didn't realize prompt was a valid mode to target.  Ctrl + arrow is a standard key pretty much everywhere.  It's jarring when it doesn't work here.

krobelus:
I think this change makes sense. Pompt mode emulates readline which also accepts this (as does pretty much every text field).

> I'm in general pretty indifferent towards this change and don't really see the need.

I think it makes sense to either support none of `<left>` and `<c-left>`, or both. Same for a buch of others like `<c-a>`/`<home>` etc.
I personally don't use arrow keys where I have readline bindings but I can see why people would want to - for consistency with the rest of the OS.

An interesting question is whether to support it in insert mode.
When I first started using Kakoune I was really thrown off by the lack of `<c-w>` in insert mode (which is supported by Vim).
I think I got used to it somehow, I don't think I ever added a mapping.

PR #5003, opened 2023-10-22
rc/tools/git: make git commands work from anywhere

glaeqen:
Previously, commands like `git log` or `git show` assumed kakoune workspace path (`cwd`) when invoking the underlying `git` command. This made them fail if eg. `cwd` happened to be a superdirectory of any git repository. This behaviour was also somewhat inconsistent with eg. `git (show|hide)-blame` which happens to work on the file basis and not on the `cwd`.

Now, all git commands assume `cwd` only if a git repository is reachable from there. Otherwise they operate on the repository corresponding to a recently opened file.

This change makes it more convenient to work on multiple repositories within the same kakoune session when `cwd` is not within a repository itself.

### DIFF
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -59,7 +59,7 @@ declare-option -hidden \
         Path of the top-level directory of the working tree currently
         used in `git` operations from within the `kakoune`
     } \
-    str git_work_tree_top_level
+    str git_work_tree
 
 define-command -params 1.. \
     -docstring %{
@@ -112,28 +112,32 @@ define-command -params 1.. \
             update-diff \
         ;
     else
-        # Try cd-ing into the git work tree if exists to make autocompletion slightly more robust
-        cd "$kak_opt_git_work_tree_top_level" >/dev/null 2>&1
+        # Without maintaining the state we cannot know the work tree.
+        # Autocompletion will work only for the default case when git work tree is at `cwd`
+        # Piping to /dev/null to silent noisy git errors
         case "$1" in
-            commit) printf -- "--amend\n--no-edit\n--all\n--reset-author\n--fixup\n--squash\n"; git ls-files -m ;;
-            add) git ls-files -dmo --exclude-standard ;;
+            commit) printf -- "--amend\n--no-edit\n--all\n--reset-author\n--fixup\n--squash\n"; git ls-files -m 2>/dev/null ;;
+            add) git ls-files -dmo --exclude-standard 2>/dev/null ;;
             apply) printf -- "--reverse\n--cached\n--index\n" ;;
-            grep|edit) git ls-files -c --recurse-submodules ;;
+            grep|edit) git ls-files -c --recurse-submodules 2>/dev/null ;;
         esac
     fi
   } \
   git %{
-  set-option global git_work_tree_top_level %sh{
+  set-option buffer git_work_tree %sh{
     # - try to get the toplevel for the `cwd`
     # - if failed, try to find a toplevel for a given buffile
     # - if failed, test the current toplevel variable and if it is a functional git repository
     # - if failed, fallback to `cwd` ()
-    git rev-parse --show-toplevel 2>/dev/null || \
-    ( cd "${kak_buffile%/*}" && git rev-parse --show-toplevel ) 2>/dev/null || \
-    ( cd "${kak_opt_git_work_tree_top_level}" && git rev-parse --show-toplevel ) 2>/dev/null || \
+    git rev-parse --show-toplevel 2>/dev/null ||
+    git -C "${kak_buffile%/*}" rev-parse --show-toplevel 2>/dev/null ||
+    git -C "${kak_opt_git_work_tree}" rev-parse --show-toplevel 2>/dev/null ||
     pwd
   }
   evaluate-commands %sh{
+    export GIT_WORK_TREE="$kak_opt_git_work_tree"
+    export GIT_DIR="$kak_opt_git_work_tree/.git"
+
     show_git_cmd_output() {
         local filetype
         local map_diff_goto_source
@@ -159,6 +163,7 @@ define-command -params 1.. \
         printf %s "evaluate-commands -try-client '$kak_opt_docsclient' %{
                   edit! -fifo ${output} *git*
                   set-option buffer filetype '${filetype}'
+                  set-option buffer git_work_tree '${kak_opt_git_work_tree}'
                   hook -always -once buffer BufCloseFifo .* %{ nop %sh{ rm -r $(dirname ${output}) } }
                   ${map_diff_goto_source}
               }"
@@ -207,9 +212,12 @@ define-command -params 1.. \
         if git "${@}" > /dev/null 2>&1; then
           printf %s "echo -markup '{Information}git $1 succeeded'"
         else
-          printf 'echo -debug Note: file paths are assumed to start in the `git_work_tree_top_level`\n' "$1"
-          printf 'echo -debug Note: mileage might vary when `git_work_tree_top_level` != kak workspace path\n' "$1"
-          printf 'fail git %s failed\n' "$1"
+          local extra_warning
+          if [ "$kak_opt_git_work_tree" != "$(pwd)" ]
+          then
+            extra_warning=" $(printf '(autocompletion unavailable, relative paths resolved against `%s`)' "$kak_opt_git_work_tree")"
+          fi
+          printf 'fail "failed to run `git%s`%s"' "$(printf ' %s' "$@")" "$extra_warning"
         fi
     }
 
@@ -353,7 +361,7 @@ define-command -params 1.. \
         msgfile="$(git rev-parse --absolute-git-dir)/COMMIT_EDITMSG"
         printf %s "edit '$msgfile'
               hook buffer BufWritePost '.*\Q$msgfile\E' %{ evaluate-commands %sh{
-                  cd "$kak_opt_git_work_tree_top_level"
+                  cd "$kak_opt_git_work_tree"
                   if git commit -F '$msgfile' --cleanup=strip $* > /dev/null; then
                      printf %s 'evaluate-commands -client $kak_client echo -markup %{{Information}Commit succeeded}; delete-buffer'
                   else
@@ -362,10 +370,6 @@ define-command -params 1.. \
               } }"
     }
 
-    cd "$kak_opt_git_work_tree_top_level" || {
-        printf 'fail Unable to set the work tree for git: %s\n' "$kak_opt_git_work_tree_top_level"
-        exit 1
-    }
     case "$1" in
         apply)
             shift
@@ -437,5 +441,5 @@ define-command -params 1.. \
 define-command git-diff-goto-source \
     -docstring 'Navigate to source by pressing the enter key in hunks when git diff is displayed. Works within :git diff and :git show' %{
     require-module diff
-    diff-jump %opt{git_work_tree_top_level}
+    diff-jump %opt{git_work_tree}
 }
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -54,6 +54,11 @@ hook -group git-show-branch-highlight global WinSetOption filetype=git-show-bran
 declare-option -hidden line-specs git_blame_flags
 declare-option -hidden line-specs git_diff_flags
 declare-option -hidden int-list git_hunk_list
+declare-option -hidden \
+    -docstring %{
+        Path of the top-level directory of the working tree relevant for the given file
+    } \
+    str git_work_tree_top_level
 
 define-command -params 1.. \
     -docstring %{
@@ -106,6 +111,8 @@ define-command -params 1.. \
             update-diff \
         ;
     else
+        # Try cd-ing into the git work tree if exists
+        cd "$kak_opt_git_work_tree_top_level" &> /dev/null
         case "$1" in
             commit) printf -- "--amend\n--no-edit\n--all\n--reset-author\n--fixup\n--squash\n"; git ls-files -m ;;
             add) git ls-files -dmo --exclude-standard ;;
@@ -114,15 +121,16 @@ define-command -params 1.. \
         esac
     fi
   } \
-  git %{ evaluate-commands %sh{
-    cd_bufdir() {
-        dirname_buffer="${kak_buffile%/*}"
-        cd "${dirname_buffer}" 2>/dev/null || {
-            printf 'fail Unable to change the current working directory to: %s\n' "${dirname_buffer}"
-            exit 1
-        }
-    }
-
+  git %{
+  set-option global git_work_tree_top_level %sh{
+    # - try to find a toplevel for a given buffile
+    # - if failed, test the current toplevel variable and if it is a functional git repository
+    # - if failed, fallback to `pwd`
+    ( cd "${kak_buffile%/*}" && git rev-parse --show-toplevel ) 2>/dev/null || \
+    ( cd "${kak_opt_git_work_tree_top_level}" && git rev-parse --show-toplevel ) 2>/dev/null || \
+    pwd
+  }
+  evaluate-commands %sh{
     show_git_cmd_output() {
         local filetype
         local map_diff_goto_source
@@ -155,7 +163,6 @@ define-command -params 1.. \
 
     run_git_blame() {
         (
-            cd_bufdir
             printf %s "evaluate-commands -client '$kak_client' %{
                       try %{ add-highlighter window/git-blame flag-lines Information git_blame_flags }
                       set-option buffer=$kak_bufname git_blame_flags '$kak_timestamp'
@@ -197,13 +204,14 @@ define-command -params 1.. \
         if git "${@}" > /dev/null 2>&1; then
           printf %s "echo -markup '{Information}git $1 succeeded'"
         else
+          printf 'echo -debug Note: file paths are assumed to start in the `git_work_tree_top_level`\n' "$1"
+          printf 'echo -debug Note: mileage might vary when `git_work_tree_top_level` != kak workspace path\n' "$1"
           printf 'fail git %s failed\n' "$1"
         fi
     }
 
     update_diff() {
         (
-            cd_bufdir
             git --no-pager diff --no-ext-diff -U0 "$kak_buffile" | perl -e '
             use utf8;
             $flags = $ENV{"kak_timestamp"};
@@ -337,11 +345,12 @@ define-command -params 1.. \
 			$@
 		EOF
 
-        # fails, and generate COMMIT_EDITMSG
+        # Run and fail in order to generate a `COMMIT_EDITMSG` file
         GIT_EDITOR='' EDITOR='' git commit "$@" > /dev/null 2>&1
-        msgfile="$(git rev-parse --git-dir)/COMMIT_EDITMSG"
+        msgfile="$(git rev-parse --absolute-git-dir)/COMMIT_EDITMSG"
         printf %s "edit '$msgfile'
               hook buffer BufWritePost '.*\Q$msgfile\E' %{ evaluate-commands %sh{
+                  cd "$kak_opt_git_work_tree_top_level"
                   if git commit -F '$msgfile' --cleanup=strip $* > /dev/null; then
                      printf %s 'evaluate-commands -client $kak_client echo -markup %{{Information}Commit succeeded}; delete-buffer'
                   else
@@ -350,6 +359,10 @@ define-command -params 1.. \
               } }"
     }
 
+    cd "$kak_opt_git_work_tree_top_level" || {
+        printf 'fail Unable to set the work tree for git: %s\n' "$kak_opt_git_work_tree_top_level"
+        exit 1
+    }
     case "$1" in
         apply)
             shift
@@ -421,5 +434,5 @@ define-command -params 1.. \
 define-command git-diff-goto-source \
     -docstring 'Navigate to source by pressing the enter key in hunks when git diff is displayed. Works within :git diff and :git show' %{
     require-module diff
-    diff-jump %sh{ git rev-parse --show-toplevel }
+    diff-jump %opt{git_work_tree_top_level}
 }
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -56,7 +56,8 @@ declare-option -hidden line-specs git_diff_flags
 declare-option -hidden int-list git_hunk_list
 declare-option -hidden \
     -docstring %{
-        Path of the top-level directory of the working tree relevant for the given file
+        Path of the top-level directory of the working tree currently
+        used in `git` operations from within the `kakoune`
     } \
     str git_work_tree_top_level
 
@@ -111,8 +112,8 @@ define-command -params 1.. \
             update-diff \
         ;
     else
-        # Try cd-ing into the git work tree if exists
-        cd "$kak_opt_git_work_tree_top_level" &> /dev/null
+        # Try cd-ing into the git work tree if exists to make autocompletion slightly more robust
+        cd "$kak_opt_git_work_tree_top_level" >/dev/null 2>&1
         case "$1" in
             commit) printf -- "--amend\n--no-edit\n--all\n--reset-author\n--fixup\n--squash\n"; git ls-files -m ;;
             add) git ls-files -dmo --exclude-standard ;;
@@ -123,9 +124,11 @@ define-command -params 1.. \
   } \
   git %{
   set-option global git_work_tree_top_level %sh{
-    # - try to find a toplevel for a given buffile
+    # - try to get the toplevel for the `cwd`
+    # - if failed, try to find a toplevel for a given buffile
     # - if failed, test the current toplevel variable and if it is a functional git repository
-    # - if failed, fallback to `pwd`
+    # - if failed, fallback to `cwd` ()
+    git rev-parse --show-toplevel 2>/dev/null || \
     ( cd "${kak_buffile%/*}" && git rev-parse --show-toplevel ) 2>/dev/null || \
     ( cd "${kak_opt_git_work_tree_top_level}" && git rev-parse --show-toplevel ) 2>/dev/null || \
     pwd


rc/tools/git.kak
```diff
@@ -71,22 +76,25 @@ define-command -params 1.. \
...
-        }
-    }
-
+  git %{
+  set-option global git_work_tree_top_level %sh{
+    # - try to find a toplevel for a given buffile
+    # - if failed, test the current toplevel variable and if it is a functional git repository
+    # - if failed, fallback to `pwd`
+    ( cd "${kak_buffile%/*}" && git rev-parse --show-toplevel ) 2>/dev/null || \
+    ( cd "${kak_opt_git_work_tree_top_level}" && git rev-parse --show-toplevel ) 2>/dev/null || \
```
krobelus, line 94: I sometimes do

	kak src/main.cc
	:edit ../some-other-git-worktree/file.c

but I still expect `:git diff` to run in cwd instead of `some-other-git-worktree`.

I'm not yet sure which use case is more important.
I appreciate your DWIM approach but I'd like to think about the UI a bit more.

One quick workaround for editing multiple repositories is to run (perhaps from a mapping or even a `WinDisplay` hook?): `cd %sh{git -C "${kak_buffile%/*}" rev-parse --show-toplevel}`.
Or would that disrupt your workflow?

If we implemented `setenv`, one could do

	setenv GIT_WORKTREE /path/to/some-git-worktree

which works for all git commands

rc/tools/git.kak
```diff
@@ -71,22 +76,25 @@ define-command -params 1.. \
     if [ $kak_token_to_complete -eq 0 ]; then
         printf "add\nrm\nreset\nblame\ncommit\ncheckout\ndiff\nhide-blame\nhide-diff\nlog\nnext-hunk\nprev-hunk\nshow\nshow-branch\nshow-diff\ninit\nstatus\nupdate-diff\ngrep\n"
     else
+        # Try cd-ing into the git work tree if exists
+        cd "$kak_opt_git_work_tree_top_level" &> /dev/null
```
krobelus, line 80: `&>` is a bashism unfortunately

rc/tools/git.kak
glaeqen: Thanks for a quick response :)

> but I still expect `:git diff` to run in cwd instead of `some-other-git-worktree`.

I was expecting that I will break someone's workflow :P

> One quick workaround for editing multiple repositories is to run (perhaps from a mapping or even a `WinDisplay` hook?): `cd %sh{git -C "${kak_buffile%/*}" rev-parse --show-toplevel}`.
Or would that disrupt your workflow?

I usually keep a long running kakoune session with LSP configured by the kakscript located at the workspace path that gets loaded automatically and which configures multiple projects at the same time
```
set global lsp_config %{
    [language.rust.settings.rust-analyzer]
    linkedProjects = [
        "<project-1>/Cargo.toml",
        "<project-2>/Cargo.toml",
    ]
# ...
}
```
plus I usually have an `.ignore` file there as well that gets picked up by ripgrep and fd-find.

`cd`-ing everytime I want to `:git ..` is probably not great as it will confuse LSP and restoring the previous workspace path is probably quite cumbersome. I like being able to `:fzf-find` or `:fzf-grep` from anywhere, I very rarely do `:e` and similar.

My idea assumes some level of statefulness so random fileless buffers, like *grep*, *git* and such will inherit/pick up the worktree that was used on the very last "normal" file. Maybe with the envvars being controllable it would be as easy as setting `GIT_DIR` and `GIT_WORK_TREE` in a hook with the same logic as in the `git_work_tree_top_level` setter now. At least it sounds like a viable path.

rc/tools/git.kak
glaeqen: Fixed (hopefully)

rc/tools/git.kak
glaeqen: @krobelus I think I restored previous behaviour with the latest commit. Now, if the `git` repository is reachable from within the `cwd`, it will have unconditional precedence. So your workflow is preserved. However, if there's no repository from `cwd`, we try to be more clever and look at the file. I left the `pwd` fallback to preserve errors coming from `git` the way they worked so far. Do you think it makes sense now? :)

rc/tools/git.kak
```diff
@@ -42,6 +42,12 @@ hook -group git-show-branch-highlight global WinSetOption filetype=git-show-bran
 declare-option -hidden line-specs git_blame_flags
 declare-option -hidden line-specs git_diff_flags
 declare-option -hidden int-list git_hunk_list
+declare-option -hidden \
```
glaeqen, line 45: @krobelus Maybe this option should not even be hidden. It's quite useful for scripting.

Very nice command that I have in my private config (which could possibly be a useful addition to the build-in kakscripts) is `git show` of the commit that the given line blames.

```
define-command git-show-blamed -params .. \
    -docstring %{
        git-show-blamed [<arguments>]: show the commit that is blamed for the current line
        All arguments are forwarded to git blame
    } \
%{
    git show %sh{
        # Change directory to where the relevant repo is supposed to be
        cd $kak_opt_git_work_tree_top_level
        git blame \
            "$kak_buffile" \
            -L"$kak_cursor_line,$kak_cursor_line" \
            --porcelain $@ \
            | head -1 | cut -d' ' -f1
    }
}
```

krobelus: the command looks nice. Should quote `'$@"` to be safe.
I don't think it requires your option; we  can cd to the buffile directory instead.
I personally use `tig` because it supports recursive blame.
Actually it seems doable to implement recursive-blame support in your command.

Related question is whether we should have a third-party plugin for Git. Maybe that would enable experiments with various interfaces.
I imagine that after a `git-blame`, `<ret>` should do what your command does. I guess this is something that we could try to add to the built-in scripts.. `<ret>` is currently only used by *grep* and `git-diff-goto-source`.. the latter would collide with recursive blame.. not sure how to fix that. Could take a look at  similar plugins

rc/tools/git.kak
krobelus: Interesting workflow. Is the linkedProjects settings equivalent to defining a cargo workspace?

I need to think about which behavior is better overall.
Right now I personally prefer  that `git status` and `git diff` run in cwd instead of the buffer's git worktree. 

Falling back to the buffer's git worktree seems benign either way. Of course it will fail if `$HOME` is a git worktree, which is unfortunate

I think Emacs with Projectile will behave like your initial patch (didn't check though).

I'll try both options. The behavior change wouldn't be a tragic loss. I think it also depends on what other rc scripts expect.
It is possible that they can benefit from a project root. Personally I haven't dared to use commands like :git for projects below cwd.

rc/tools/git.kak
```diff
@@ -42,6 +42,12 @@ hook -group git-show-branch-highlight global WinSetOption filetype=git-show-bran
...
+    -docstring %{
+        Path of the top-level directory of the working tree currently
+        used in `git` operations from within the `kakoune`
+    } \
+    str git_work_tree_top_level
```
krobelus, line 50: why not `git_work_tree`?

rc/tools/git.kak
```diff
@@ -71,22 +77,27 @@ define-command -params 1.. \
     if [ $kak_token_to_complete -eq 0 ]; then
         printf "add\nrm\nreset\nblame\ncommit\ncheckout\ndiff\nhide-blame\nhide-diff\nlog\nnext-hunk\nprev-hunk\nshow\nshow-branch\nshow-diff\ninit\nstatus\nupdate-diff\ngrep\n"
     else
+        # Try cd-ing into the git work tree if exists to make autocompletion slightly more robust
+        cd "$kak_opt_git_work_tree_top_level" >/dev/null 2>&1
```
krobelus, line 81: I expect completion to use exactly the same cwd as the command itself. It
is surprising that we use the value from a previous :git invocation here.

rc/tools/git.kak
```diff
@@ -71,22 +77,27 @@ define-command -params 1.. \
...
-    cd_bufdir() {
-        dirname_buffer="${kak_buffile%/*}"
-        cd "${dirname_buffer}" 2>/dev/null || {
-            printf 'fail Unable to change the current working directory to: %s\n' "${dirname_buffer}"
-            exit 1
-        }
-    }
-
+  git %{
+  set-option global git_work_tree_top_level %sh{
```
krobelus, line 90: It doesn't seem right to set this in the global scope (it's a source of wrong behavior in completion and `git-diff-goto-source`).
Ideally we get rid of this option, it appears though only `git-diff-goto-source` needs it.
I see two potential fixes:
1. set the option only at buffer scope for buffers where we bind `git-diff-goto-source`
2. bypass  `git-diff-goto-source` and encode the git directory where we map it

		"unmap buffer normal <ret> :diff-jump $PWD<ret>"

I prefer 2 because then we don't need a new option.

rc/tools/git.kak
```diff
@@ -71,22 +77,27 @@ define-command -params 1.. \
...
+    # - try to get the toplevel for the `cwd`
+    # - if failed, try to find a toplevel for a given buffile
+    # - if failed, test the current toplevel variable and if it is a functional git repository
+    # - if failed, fallback to `cwd` ()
+    git rev-parse --show-toplevel 2>/dev/null || \
```
krobelus, line 95: no need for the trailing backslash, `||` already makes the shell read further

rc/tools/git.kak
```diff
@@ -71,22 +77,27 @@ define-command -params 1.. \
...
+    ( cd "${kak_buffile%/*}" && git rev-parse --show-toplevel ) 2>/dev/null || \
```
krobelus, line 96: Technically we don't want to try this for scratch buffers.. but I don't know how to distinguish scratch buffers here. I guess it's okay to ignore this for now

rc/tools/git.kak
```diff
@@ -160,13 +170,14 @@ define-command -params 1.. \
         if git "${@}" > /dev/null 2>&1; then
           printf %s "echo -markup '{Information}git $1 succeeded'"
         else
+          printf 'echo -debug Note: file paths are assumed to start in the `git_work_tree_top_level`\n' "$1"
+          printf 'echo -debug Note: mileage might vary when `git_work_tree_top_level` != kak workspace path\n' "$1"
```
krobelus, line 174: If `git_work_tree_top_level` does not deviate from `git rev-parse --show-toplevel`, then we don't need to print this note.

Tangentially related, perhaps we should print the full command that failed, like:

	printf "fail '%s'\n" "$(printf 'failed to run git -C %s%s' "$kak_opt_git_work_tree_top_level" "$(printf ' %s' "$@")" | sed "s/'/''/g")"

rc/tools/git.kak
krobelus: ok I hit this recently; I think I'm coming around on this one.. I think running in the buffer's worktree is the better default behavior.
However if `$PWD` is a child directory of the buffer's worktree, we should not `cd` up because that could invalidate pathspecs and other things.

So I think the logic should be

```sh
if ! bufworktree=$(git -C "${kak_buffile%/*}" rev-parse --show-toplevel); then
	exit # use $PWD
fi
if [ "${PWD#"$bufworktree"}" != "$PWD" ]; then
	# $PWD is a child of a Git worktree
	exit # use $PWD
fi
cd "$bufworktree"
```

If we really want to add a state variable on top of that, I'd like to hear the motivation.
(we can add support for `:git -Cpath/to/worktree` and GIT_DIR/GIT_WORK_TREE env vars if needed)

If you're too busy, I'm happy to take over this PR, LMK

rc/tools/git.kak
krobelus: oh I see, `*git*` does not have an implied git directory that's why the option is useful here.
Given that, I'm happy to champion a `git_work_tree` option that overrides $PWD* as long as this option is only set in the buffer scope

* (maybe also overrides the buffer directory but that doesn't really matter for scratch buffers)

mawww:
Correct me if I am wrong, but say I am in project A directory, open a file in project B (say by jumping to definition), then lookup some man page, if from that `*man printf*` buffer I try to do a git operation, it will try to do it in project B ?

I am a bit concerned about maintaining state like that because it sounds like you could get hard to predict behaviours.

krobelus:
Yes  we should not introduce global state. Fortunately we don't need to, we can always recompute this.
We might want one in buffer scope (for `*git*` buffers) but that can also be done in a separate PR.

rc/tools/git.kak
glaeqen: Sorry, I disappeared for a longer while. All kinds of errands happening. I really appreciate the input. 

> If you're too busy, I'm happy to take over this PR, LMK

Sure, if you feel like playing with it, that'd be great. I'm not that great of a shell expert anyhow but I'd like to participate and give feedback.

> logic should be

> state variable

A lot of discussion happened across the comments but I understand the global option being evil part. I liked one aspect of my solution that maintained the "information about the latest reasonable git work tree path". Which could be used anywhere for any kind of scripting. It was especially nice for scratch buffers like *git* so one was not losing the information about "what was the relevant git work tree" a moment ago.

rc/tools/git.kak
glaeqen: > I personally use tig because it supports recursive blame.

I use `gitui` usually, that one does not seem to have it. Being able to recursively traverse the blame history would be really neat indeed although..

> Actually it seems doable to implement recursive-blame support in your command.

> the latter would collide with recursive blame

Yeah, not sure how that would work either 🤔 That sounds like a slightly more advance assignment. One could checkout a single file from the commit pointed by the blame into the scratch buffer and apply `run_git_blame` with the `git blame` output from that commit perspective. And that should work recursively until I get to the first commit that given line was introduce at.

rc/tools/git.kak
glaeqen: Git naming is a little confusing and at least at the time I got to the conclusion that `git_work_tree` does not mean exactly what I mean by that option. Because work tree is a work tree and a top-level of it is a top-level of it, and so on. In the end it is ok for me to have `git_work_tree` 👍🏻

rc/tools/git.kak
glaeqen: That was a consequence of my stateful variable. But I see this comment was older than others so you realized that by then.

rc/tools/git.kak
glaeqen: Not sure if I follow the 2 variant. What is on your mind exactly? If you remember after the month 🙈

glaeqen: I see the appeal of the variant 2. but at the same time having the option is really handy for any external scripting that would like to be "reasonable git-work-tree" aware.

rc/tools/git.kak
glaeqen: Fixed, added an extra warning for the "out-of-work-tree" that autocompletion is unavailable and which directories relative paths are resolved against

rc/tools/git.kak
glaeqen: I don't know how to do it as well. I switched to `-C` but the problem remains if some user names their directories like '*git*' or just happens to have scratch buffers named like some directories..? There's probably a bunch of unpleasant corner-cases that might happen.

Speaking of wildcards, I noticed that parameterless `git add/rm` defaults to current buffile which then goes straight to the `git` command which seems to trigger wildcard expansion of the shell in case of scratch buffers like `*git*`.

rc/tools/git.kak
glaeqen: Fixed

glaeqen:
I have made `git_work_tree` to work in a buffer scope and I manually propagate the option to the `*git*` buffers. Unfortunately even more git commands were integrated in the meantime that rely on heavily on autocompletion. Maybe that can be improved also.

glaeqen:
Admittedly it requires quite a lot of boilerplate to make the autocompletion patch-up the paths so they make sense for commands like `add|rm|grep|edit`. This coupling between the assumed worktree and `cwd` is quite strong and leaks everywhere. I guess it works for most but makes it a little awkward with the autocompletion usecases. What do you think @krobelus ?

rc/tools/git.kak
```diff
@@ -106,22 +112,31 @@ define-command -params 1.. \
...
-  git %{ evaluate-commands %sh{
-    cd_bufdir() {
-        dirname_buffer="${kak_buffile%/*}"
-        cd "${dirname_buffer}" 2>/dev/null || {
-            printf 'fail Unable to change the current working directory to: %s\n' "${dirname_buffer}"
-            exit 1
-        }
-    }
+  git %{
+  set-option buffer git_work_tree %sh{
```
krobelus, line 127: I don't think we should set the option on the buffer where the git command is invoked, unless there is a motivation I have missed?

rc/tools/git.kak
```diff
@@ -337,11 +356,12 @@ define-command -params 1.. \
...
 
-        # fails, and generate COMMIT_EDITMSG
+        # Run and fail in order to generate a `COMMIT_EDITMSG` file
         GIT_EDITOR='' EDITOR='' git commit "$@" > /dev/null 2>&1
-        msgfile="$(git rev-parse --git-dir)/COMMIT_EDITMSG"
+        msgfile="$(git rev-parse --absolute-git-dir)/COMMIT_EDITMSG"
         printf %s "edit '$msgfile'
               hook buffer BufWritePost '.*\Q$msgfile\E' %{ evaluate-commands %sh{
+                  cd "$kak_opt_git_work_tree"
                   if git commit -F '$msgfile' --cleanup=strip $* > /dev/null; then
```
krobelus, line 365: I made it so we don't change directory here because that might mess with `$*`

krobelus:
I think this is a good change but we should fix some things (mostly some places where we fail to pass on the worktree).
My proposal is below.

> I guess it works for most but makes it a little awkward with the autocompletion usecases.

Autocompletion should behave exactly as the eventual command.
So, same cwd, same GIT_WORK_TREE and GIT_DIR, then we don't fail educated expectations.
I haven't tested much but it seems to work.

```patch
From 2dd513f46e5f429b062b1ad35f6c9dbac7c02453 Mon Sep 17 00:00:00 2001
From: Johannes Altmanninger <aclopte@gmail.com>
Date: Wed, 13 Dec 2023 09:29:43 +0100
Subject: [PATCH 1/4] rc tools git: quote $kak_buffile to prevent shell
 expansion

TIL `git add` and `git rm` default to the current file.  The filename
undergoes glob expansion, for example if it's a scratch buffer like
`*git*`. Fix this.
---
 rc/tools/git.kak | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

index 2f19593e4..29de038d2 100644
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -391,7 +391,7 @@ define-command -params 1.. \
         add|rm)
             cmd="$1"
             shift
-            run_git_cmd $cmd "${@:-${kak_buffile}}"
+            run_git_cmd $cmd "${@:-"${kak_buffile}"}"
             ;;
         reset|checkout)
             run_git_cmd "$@"
-- 
2.43.0


From d772274ab8f582974956492def23709ee18dafe8 Mon Sep 17 00:00:00 2001
From: Johannes Altmanninger <aclopte@gmail.com>
Date: Wed, 13 Dec 2023 06:45:16 +0100
Subject: [PATCH 2/4] rc tools git: print full command that failed

---
 rc/tools/git.kak | 3 ++-
 1 file changed, 2 insertions(+), 1 deletion(-)

index 29de038d2..978ec6bdc 100644
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -197,7 +197,8 @@ define-command -params 1.. \
         if git "${@}" > /dev/null 2>&1; then
           printf %s "echo -markup '{Information}git $1 succeeded'"
         else
-          printf 'fail git %s failed\n' "$1"
+          printf "fail '%s'\n" \
+            "$(printf 'failed to run git %s' "$(printf ' %s' "$@")" | sed "s/'/''/g")"
         fi
     }
 
-- 
2.43.0


From 06f14ba5a71ea128139cb2715bae4a6d22592c35 Mon Sep 17 00:00:00 2001
From: Johannes Altmanninger <aclopte@gmail.com>
Date: Wed, 13 Dec 2023 06:45:16 +0100
Subject: [PATCH 3/4] rc tools git: if $PWD is outside the buffer's worktree,
 fall back to that worktree

This fixes the scenario where th user is editing $PWD/some-project/...
and runs a Git command; neither $PWD nor any of its parents is a Git
worktree, only $PWD/some-project/ is. Let's use it in this case.

---

As an aside: there is another case: when both $PWD and
$PWD/../some-project/current-buffer are both in Git worktrees, then
it's not clear which one to prefer.  We historically prefer the cwd.
In future we can add a way to toggle this behavior if anyone needs it.
---
 rc/tools/git.kak | 51 +++++++++++++++++++++++++-----------------------
 1 file changed, 27 insertions(+), 24 deletions(-)

index 978ec6bdc..b073da70c 100644
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -106,6 +106,16 @@ define-command -params 1.. \
             update-diff \
         ;
     else
+        if bufworktree=$(git -C "${kak_buffile%/*}" rev-parse --show-toplevel) &&
+                # PWD is not inside the buffer's worktree
+                [ "${PWD#"${bufworktree}/"}" = $PWD ]; then
+            git_work_tree=${bufworktree}
+        else
+            git_work_tree=$(git rev-parse --show-toplevel)
+        fi
+        export GIT_WORK_TREE=${GIT_WORK_TREE:-git_work_tree}
+        export GIT_DIR=${GIT_DIR:-GIT_WORK_TREE}/.git
+
         case "$1" in
@@ -115,13 +125,15 @@ define-command -params 1.. \
     fi
   } \
   git %{ evaluate-commands %sh{
-    cd_bufdir() {
-        dirname_buffer="${kak_buffile%/*}"
-        cd "${dirname_buffer}" 2>/dev/null || {
-            printf 'fail Unable to change the current working directory to: %s\n' "${dirname_buffer}"
-            exit 1
-        }
-    }
+    if bufworktree=$(git -C "${kak_buffile%/*}" rev-parse --show-toplevel) &&
+            # PWD is not inside the buffer's worktree
+            [ "${PWD#"${bufworktree}/"}" = $PWD ]; then
+        git_work_tree=${bufworktree}
+    else
+        git_work_tree=$(git rev-parse --show-toplevel)
+    fi
+    export GIT_WORK_TREE=${GIT_WORK_TREE:-git_work_tree}
+    export GIT_DIR=${GIT_DIR:-GIT_WORK_TREE}/.git
 
     show_git_cmd_output() {
         local filetype
@@ -142,8 +154,8 @@ define-command -params 1.. \
         # We need to unmap in case an existing buffer changes type,
         # for example if the user runs "git show" and "git status".
         map_diff_goto_source=$([ -n "${map_diff_goto_source}" ] \
-          && printf %s "map buffer normal <ret> :git-diff-goto-source<ret> -docstring 'Jump to source from git diff'" \
-          || printf %s "unmap buffer normal <ret> :git-diff-goto-source<ret>")
+          && printf %s "map buffer normal <ret> ':require-module diff; diff-jump ''$git_work_tree''<ret>' -docstring 'Jump to source from git diff'" \
+          || printf %s "unmap buffer normal <ret> ':require-module diff; diff-jump ''$git_work_tree''<ret>'")
 
                   edit! -fifo ${output} *git*
@@ -155,7 +167,6 @@ define-command -params 1.. \
 
     run_git_blame() {
         (
-            cd_bufdir
             printf %s "evaluate-commands -client '$kak_client' %{
                       try %{ add-highlighter window/git-blame flag-lines Information git_blame_flags }
                       set-option buffer=$kak_bufname git_blame_flags '$kak_timestamp'
@@ -198,13 +209,12 @@ define-command -params 1.. \
           printf %s "echo -markup '{Information}git $1 succeeded'"
         else
           printf "fail '%s'\n" \
-            "$(printf 'failed to run git %s' "$(printf ' %s' "$@")" | sed "s/'/''/g")"
+            "$(printf 'failed to run git --work-tree %s --git-dir %s%s' "${git_work_tree}" "${GIT_DIR}" "$(printf ' %s' "$@")" | sed "s/'/''/g")"
         fi
     }
 
     update_diff() {
         (
-            cd_bufdir
             git --no-pager diff --no-ext-diff -U0 "$kak_buffile" | perl -e '
             use utf8;
             $flags = $ENV{"kak_timestamp"};
@@ -338,12 +348,12 @@ define-command -params 1.. \
 			$@
 		EOF
 
-        # fails, and generate COMMIT_EDITMSG
+        # Run and fail in order to generate a `COMMIT_EDITMSG` file
         GIT_EDITOR='' EDITOR='' git commit "$@" > /dev/null 2>&1
-        msgfile="$(git rev-parse --git-dir)/COMMIT_EDITMSG"
+        msgfile="$GIT_DIR/COMMIT_EDITMSG"
         printf %s "edit '$msgfile'
               hook buffer BufWritePost '.*\Q$msgfile\E' %{ evaluate-commands %sh{
+                  if git --work-tree '${git_work_tree}' --git-dir '${GIT_DIR}' commit -F '$msgfile' --cleanup=strip $* > /dev/null; then
                      printf %s 'evaluate-commands -client $kak_client echo -markup %{{Information}Commit succeeded}; delete-buffer'
                   else
                      printf 'evaluate-commands -client %s fail Commit failed\n' "$kak_client"
@@ -356,7 +366,7 @@ define-command -params 1.. \
             shift
             enquoted="$(printf '"%s" ' "$@")"
             echo "require-module patch"
-            echo "patch git apply $enquoted"
+            echo "patch git --work-tree '${git_work_tree}' --git-dir '${GIT_DIR}' apply $enquoted"
             ;;
         show|show-branch|log|diff|status)
             show_git_cmd_output "$@"
@@ -401,7 +411,7 @@ define-command -params 1.. \
             shift
             enquoted="$(printf '"%s" ' "$@")"
             printf %s "try %{
-                set-option current grepcmd 'git grep -n --column'
+                set-option current grepcmd 'git --work-tree ${git_work_tree} --git-dir ${GIT_DIR} grep -n --column'
                 grep $enquoted
                 set-option current grepcmd '$kak_opt_grepcmd'
             }"
@@ -417,10 +427,3 @@ define-command -params 1.. \
             ;;
     esac
 }}
-
-# Works within :git diff and :git show
-define-command git-diff-goto-source \
-    -docstring 'Navigate to source by pressing the enter key in hunks when git diff is displayed. Works within :git diff and :git show' %{
-    require-module diff
-    diff-jump %sh{ git rev-parse --show-toplevel }
-}
-- 
2.43.0


From 56a5654776ea65147766348f6f3c33a092b1831c Mon Sep 17 00:00:00 2001
From: Johannes Altmanninger <aclopte@gmail.com>
Date: Wed, 13 Dec 2023 06:45:16 +0100
Subject: [PATCH 4/4] rc tools git: record the originating worktree in *git*
 buffers, for use in subsequent git commands

Continuing the use case of $PWD being outside a Git worktree: make
:git commands from *git* buffer use the original worktree used to
create the *git* buffer in the first place.

This new option can also be set by the user to select a Git worktree
for scratch buffers or buffers whose underlying file is outside any
Git worktree.  For buffers that are already located within a Git
worktree, this option is ignored because I don't think anyone ever
wants to override those.
---
 rc/tools/git.kak | 19 +++++++++++++++----
 1 file changed, 15 insertions(+), 4 deletions(-)

index b073da70c..020fc97c0 100644
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -54,6 +54,12 @@ hook -group git-show-branch-highlight global WinSetOption filetype=git-show-bran
 declare-option -hidden line-specs git_blame_flags
 declare-option -hidden line-specs git_diff_flags
 declare-option -hidden int-list git_hunk_list
+declare-option -hidden \
+    -docstring %{
+        Path of the top-level directory of the working tree currently
+        used in `git` operations from within the `kakoune`
+    } \
+    str git_work_tree
 
 define-command -params 1.. \
     -docstring %{
@@ -110,8 +116,10 @@ define-command -params 1.. \
                 # PWD is not inside the buffer's worktree
                 [ "${PWD#"${bufworktree}/"}" = $PWD ]; then
             git_work_tree=${bufworktree}
+        elif git_work_tree=$(git rev-parse --show-toplevel); then
+            :
         else
-            git_work_tree=$(git rev-parse --show-toplevel)
+            git_work_tree=${kak_opt_git_work_tree}
         fi
         export GIT_WORK_TREE=${GIT_WORK_TREE:-git_work_tree}
         export GIT_DIR=${GIT_DIR:-GIT_WORK_TREE}/.git
@@ -129,8 +137,10 @@ define-command -params 1.. \
             # PWD is not inside the buffer's worktree
             [ "${PWD#"${bufworktree}/"}" = $PWD ]; then
         git_work_tree=${bufworktree}
+    elif git_work_tree=$(git rev-parse --show-toplevel); then
+        :
     else
-        git_work_tree=$(git rev-parse --show-toplevel)
+        git_work_tree=${kak_opt_git_work_tree}
     fi
     export GIT_WORK_TREE=${GIT_WORK_TREE:-git_work_tree}
     export GIT_DIR=${GIT_DIR:-GIT_WORK_TREE}/.git
@@ -154,12 +164,13 @@ define-command -params 1.. \
         # We need to unmap in case an existing buffer changes type,
         # for example if the user runs "git show" and "git status".
         map_diff_goto_source=$([ -n "${map_diff_goto_source}" ] \
-          && printf %s "map buffer normal <ret> ':require-module diff; diff-jump ''$git_work_tree''<ret>' -docstring 'Jump to source from git diff'" \
-          || printf %s "unmap buffer normal <ret> ':require-module diff; diff-jump ''$git_work_tree''<ret>'")
+          && printf %s "map buffer normal <ret> ':require-module diff; diff-jump %opt{git_work_tree}<ret>' -docstring 'Jump to source from git diff'" \
+          || printf %s "unmap buffer normal <ret> ':require-module diff; diff-jump %opt{git_work_tree}<ret>'")
 
                   edit! -fifo ${output} *git*
                   set-option buffer filetype '${filetype}'
+                  set-option buffer git_work_tree '${git_work_tree}'
                   hook -always -once buffer BufCloseFifo .* %{ nop %sh{ rm -r $(dirname ${output}) } }
                   ${map_diff_goto_source}
               }"
-- 
2.43.0
```

krobelus:
>  Your option usage seems to work in a opposite way than I intended it. Here
>  it's write-only and git kak module can use it as an extra fallback. I
>  wanted to have an read-access to it for scripting. For example be able to
>  run gitui straight from the kakoune (tig alternative) for a "given buffile
>  repository". Or spawn a terminal on a root of a "given buffile repository".

it's not write-only, see:

> This new option can also be set by the user to select a Git worktree for
> scratch buffers or buffers whose underlying file is outside any Git worktree.
> For buffers that are already located within a Git worktree, this option
> is ignored because I don't think anyone ever wants to override those.

Is there a scenario where `$PWD` is already in a Git worktree and you want to pick another one?
Then we can consider changing the precedence but I wouldn't do it without a motivating case.

krobelus:
> cd-ing everytime I want to :git .. is probably not great as it will confuse LSP and restoring the previous workspace path is probably quite cumbersome. I like being able to :fzf-find or :fzf-grep from anywhere, I very rarely do :e and similar.

I haven't tested this workflow but it shouldn't confuse LSP*. What I meant is something like the `autochdir` Vim option where the editor always runs `cd $(dirname $kak_buffile)`. So never a need to restore the pwd. And `fzf` could have a root dir option.
Just an idea. There are probably more things affected by $PWD so it depends, like what do you want `make` or `cargo b` to do.

* kak-lsp will spawn a new language server for every project but you can set `KAK_LSP_FORCE_PROJECT_ROOT` on the "main" project.

glaeqen:
I don't think my `cwd` will be under any worktree usually, only different repositories in some subdirectories under `cwd`. So that's fine I think.

Maybe to clarify what I'd like with an example:

Patch on top of your patch:
```patch
index b2544692..09043d46 100644
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -87,6 +87,7 @@ define-command -params 1.. \
             show-diff
             status
             update-diff
+            work-tree
     } -shell-script-candidates %{
     if [ $kak_token_to_complete -eq 0 ]; then
         printf %s\\n \
@@ -110,6 +111,7 @@ define-command -params 1.. \
             show-diff \
             status \
             update-diff \
+            work-tree \
         ;
     else
         if bufworktree=$(git -C "${kak_buffile%/*}" rev-parse --show-toplevel 2>/dev/null) &&
@@ -432,6 +434,9 @@ define-command -params 1.. \
             enquoted="$(printf '"%s" ' "$@")"
             printf %s "edit -existing -- $enquoted"
             ;;
+        work-tree)
+            printf "set-register a %s" "$git_work_tree"
+            ;;
         *)
             printf "fail unknown git command '%s'\n" "$1"
             exit
```

and the command spawning external process (I'm using `connect.kak` to spawn popups and etc. but to keep the example as vanilla as possible I'll just use `terminal` command) as so:

```
define-command spawn-terminal-at-repo %{
    eval -draft -save-regs a %{
        git work-tree
        eval %sh{ printf "terminal sh -c 'cd \"%s\"; exec $SHELL'" "$kak_reg_a" }
    }
}
```

I'm not really sure what's the right way of passing information around but I'm just trying to make a point about the usecase. I might just not see the obvious way to use this option to achieve the same thing. It does not seem to be set by anything _in a general case_ so unless I happen to be in a `*git*` buffer that has this option set, I wouldn't have access to this information. Sorry if I'm missing something obvious.

glaeqen:
As for LSP I'm not sure exactly how you imagine around the project but I'm very much in favour of a static `cwd`. I just like to have access to multiple projects at the same time. I used the `KAK_LSP_FORCE_PROJECT_ROOT` flag before I discovered that `rust-analyzer` can handle multple unrelated project via `linked-projects` property which helps a lot. I'm not sure if  `cwd` that follows `dirname of buffile` is a good idea in general, all the tooling for fuzzy search can't rely on `cwd` anymore, jump reference for diff buffers would change with every buffer switch and so on.

krobelus:
> I'm not really sure what's the right way of passing information

Yeah it's not obvious how the UI should look like.
We could set `git_work_tree` for every buffer using a BufCreate hook.
Or only for buffers where you run a git command.

The first seems less surprising.
I don't have an opinion yet on whether this is worth adding.
For now you can duplicate the `git_work_tree` selection logic in your `terminal` invocation (or use the bespoke BufCreate hook).

glaeqen:
I might duplicate the logic for now as you say. It irks me how difficult it is to nicely structure things in kakscript. I don't know if there's any sensible way to eg. share code between `-shell-script-candidates` and the actual command body :/ 

I think it makes sense for you to take over the PR, just override my commits with your patch maybe? Thanks a lot for your help!

glaeqen:
Or create a separate PR and we can close this one in favour of the new one

krobelus:
> I don't know if there's any sensible way to eg. share code between -shell-script-candidates and the actual command body 

we can put the shared code in a hidden option. But it's not so important.

Moved the changes to https://github.com/mawww/kakoune/pull/5062

Also, in case we want to overrule `$PWD` there is https://github.com/mawww/kakoune/pull/5063
with that you can do

```
hook global BufCreate .* %{
	evaluate-commands %sh{
		if work_tree=$(git -C "${kak_hook_param%/*}" rev-parse --show-toplevel); then
			echo "set-option -add buffer env GIT_WORK_TREE=$work_tree"
		fi
	}
}
```

to always prefer the buffer worktree over PWD.

glaeqen:
Closing in favour of #5062

PR #5007, opened 2023-10-26
rc/make.kak support override the whole error line pattern

QiBaobin:
Can't use current make_error_pattern to support gradle errors like below:

```
e: file:///Users/xxx/work/prj/File.kt:32:6 Parameters must have type annotation
```

### DIFF
--- a/rc/tools/make.kak
+++ b/rc/tools/make.kak
@@ -1,7 +1,7 @@
 declare-option -docstring "shell command run to build the project" \
     str makecmd make
 declare-option -docstring "pattern that describes lines containing information about errors in the output of the `makecmd` command" \
-    str make_error_pattern " (?:fatal )?error:"
+    str make_error_pattern "^(?:\w:)?[^:\n]+:\d+:(?:\d+:)? (?:fatal )?error:"
 
 declare-option -docstring "name of the client in which utilities display information" \
     str toolsclient
@@ -69,7 +69,7 @@ define-command -hidden make-jump %{
 define-command make-next-error -docstring 'Jump to the next make error' %{
     evaluate-commands -try-client %opt{jumpclient} %{
         buffer '*make*'
-        execute-keys "%opt{make_current_error_line}ggl" "/^(?:\w:)?[^:\n]+:\d+:(?:\d+:)?%opt{make_error_pattern}<ret>"
+        execute-keys "%opt{make_current_error_line}ggl" "/%opt{make_error_pattern}<ret>"
         make-jump
     }
     try %{
@@ -83,7 +83,7 @@ define-command make-next-error -docstring 'Jump to the next make error' %{
 define-command make-previous-error -docstring 'Jump to the previous make error' %{
     evaluate-commands -try-client %opt{jumpclient} %{
         buffer '*make*'
-        execute-keys "%opt{make_current_error_line}g" "<a-/>^(?:\w:)?[^:\n]+:\d+:(?:\d+:)?%opt{make_error_pattern}<ret>"
+        execute-keys "%opt{make_current_error_line}g" "<a-/>%opt{make_error_pattern}<ret>"
         make-jump
     }
     try %{
--- a/rc/tools/make.kak
+++ b/rc/tools/make.kak
@@ -67,9 +67,10 @@ define-command -hidden make-jump %{
 }
 
 define-command make-next-error -docstring 'Jump to the next make error' %{
-    evaluate-commands -try-client %opt{jumpclient} %{
+    evaluate-commands -try-client %opt{jumpclient} -save-regs / %{
         buffer '*make*'
-        execute-keys "%opt{make_current_error_line}ggl" "/%opt{make_error_pattern}<ret>"
+        set-register / %opt{make_error_pattern}
+        execute-keys "%opt{make_current_error_line}ggl" "/<ret>"
         make-jump
     }
     try %{
@@ -81,9 +82,10 @@ define-command make-next-error -docstring 'Jump to the next make error' %{
 }
 
 define-command make-previous-error -docstring 'Jump to the previous make error' %{
-    evaluate-commands -try-client %opt{jumpclient} %{
+    evaluate-commands -try-client %opt{jumpclient} -save-regs / %{
         buffer '*make*'
-        execute-keys "%opt{make_current_error_line}g" "<a-/>%opt{make_error_pattern}<ret>"
+        set-register / %opt{make_error_pattern}
+        execute-keys "%opt{make_current_error_line}g" "<a-/><ret>"
         make-jump
     }
     try %{
--- a/rc/tools/make.kak
+++ b/rc/tools/make.kak
@@ -1,7 +1,9 @@
 declare-option -docstring "shell command run to build the project" \
     str makecmd make
+declare-option -docstring "pattern that find lines containing information about errors in the output of the `makecmd` command" \
+    str make_error_pattern " (?:fatal )?error:"
 declare-option -docstring "pattern that describes lines containing information about errors in the output of the `makecmd` command" \
-    str make_error_pattern "^(?:\w:)?[^:\n]+:\d+:(?:\d+:)? (?:fatal )?error:"
+    str make_error_line_pattern ""
 
 declare-option -docstring "name of the client in which utilities display information" \
     str toolsclient
@@ -51,7 +53,8 @@ define-command -hidden make-open-error -params 4 %{
 }
 
 define-command -hidden make-jump %{
-    evaluate-commands %{
+    evaluate-commands -save-regs / %{
+        set-register / %sh{ if [ -z "$kak_opt_make_error_line_pattern" ] ; then printf %s "^(?:\w:)?[^:\n]+:\d+:(?:\d+:)?$kak_opt_make_error_pattern" ; else printf %s "$kak_opt_make_error_line_pattern"; fi }
         try %{
             execute-keys gl<a-?> "Entering directory" <ret><a-:>
             # Try to parse the error into capture groups, failing on absolute paths
@@ -59,7 +62,7 @@ define-command -hidden make-jump %{
             set-option buffer make_current_error_line %val{cursor_line}
             make-open-error "%reg{1}/%reg{2}" "%reg{3}" "%reg{4}" "%reg{5}"
         } catch %{
-            execute-keys <a-h><a-l> s "((?:\w:)?[^:]+):(\d+):(?:(\d+):)?([^\n]+)\z" <ret>l
+            execute-keys <a-h><a-l> s "%reg{/}([^\n]+)?\z" <ret>l
             set-option buffer make_current_error_line %val{cursor_line}
             make-open-error "%reg{1}" "%reg{2}" "%reg{3}" "%reg{4}"
         }
@@ -69,7 +72,7 @@ define-command -hidden make-jump %{
 define-command make-next-error -docstring 'Jump to the next make error' %{
     evaluate-commands -try-client %opt{jumpclient} -save-regs / %{
         buffer '*make*'
-        set-register / %opt{make_error_pattern}
+        set-register / %sh{ if [ -z "$kak_opt_make_error_line_pattern" ] ; then printf %s "^(?:\w:)?[^:\n]+:\d+:(?:\d+:)?$kak_opt_make_error_pattern" ; else printf %s "$kak_opt_make_error_line_pattern"; fi }
         execute-keys "%opt{make_current_error_line}ggl" "/<ret>"
         make-jump
     }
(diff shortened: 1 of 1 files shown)


rc/tools/make.kak
```diff
@@ -69,7 +69,7 @@ define-command -hidden make-jump %{
 define-command make-next-error -docstring 'Jump to the next make error' %{
     evaluate-commands -try-client %opt{jumpclient} %{
         buffer '*make*'
-        execute-keys "%opt{make_current_error_line}ggl" "/^(?:\w:)?[^:\n]+:\d+:(?:\d+:)?%opt{make_error_pattern}<ret>"
+        execute-keys "%opt{make_current_error_line}ggl" "/%opt{make_error_pattern}<ret>"
```
occivink, line 72: `"/%opt{make_error_pattern}<ret>"` is prone to injection of special characters (e.g. if the pattern contains `<ret>`). To avoid it, you can split it in two lines
```
set-register / %opt{make_error_pattern}
execute-keys "%opt{make_current_error_line}ggl" "/<ret>"
```
As well as adding `-save-regs '/'` to the enclosing `evaluate-commands` call.

It's not as much of a problem for `make_current_error_line` since that's by definition an integer, which cannot contain any special character.

QiBaobin: I am not sure if we shall separate the fix into another PR, what do you think?

mawww: We might as well do it as part of this one

QiBaobin: @occivink @mawww , it's done, please take a another look.

krobelus: this has some of the same problems as https://github.com/mawww/kakoune/pull/4972 (breaks compat, highlighting)
Also you didn't update `make-jump`.
If we want compatibility we could add a new option (`make_location_pattern`) and deprecate the old one.

QiBaobin: @krobelus , use a new option, please take another look, thanks.

rc/tools/make.kak
```diff
@@ -51,25 +53,27 @@ define-command -hidden make-open-error -params 4 %{
...
+        set-register / %sh{ if [ -z "$kak_opt_make_error_line_pattern" ] ; then printf %s "^(?:\w:)?[^:\n]+:\d+:(?:\d+:)?$kak_opt_make_error_pattern" ; else printf %s "$kak_opt_make_error_line_pattern"; fi }
         try %{
             execute-keys gl<a-?> "Entering directory" <ret><a-:>
             # Try to parse the error into capture groups, failing on absolute paths
             execute-keys s "Entering directory [`']([^']+)'.*\n([^:/][^:]*):(\d+):(?:(\d+):)?([^\n]+)\z" <ret>l
             set-option buffer make_current_error_line %val{cursor_line}
             make-open-error "%reg{1}/%reg{2}" "%reg{3}" "%reg{4}" "%reg{5}"
         } catch %{
-            execute-keys <a-h><a-l> s "((?:\w:)?[^:]+):(\d+):(?:(\d+):)?([^\n]+)\z" <ret>l
+            execute-keys <a-h><a-l> s "%reg{/}([^\n]+)?\z" <ret>l
```
krobelus, line 65: I think we should add the `([^\n]+)?\z` to the `set-register /` command above so we can use `s<ret>l` to avoid problems if the regex contains `<`.

rc/tools/make.kak
```diff
@@ -51,25 +53,27 @@ define-command -hidden make-open-error -params 4 %{
 }
 
 define-command -hidden make-jump %{
-    evaluate-commands %{
+    evaluate-commands -save-regs / %{
+        set-register / %sh{ if [ -z "$kak_opt_make_error_line_pattern" ] ; then printf %s "^(?:\w:)?[^:\n]+:\d+:(?:\d+:)?$kak_opt_make_error_pattern" ; else printf %s "$kak_opt_make_error_line_pattern"; fi }
```
krobelus, line 57: this is only needed in the `catch` block, move it there to reduce scope?

QiBaobin: done

rc/tools/make.kak
QiBaobin: done

rc/tools/make.kak
```diff
@@ -1,7 +1,9 @@
 declare-option -docstring "shell command run to build the project" \
     str makecmd make
-declare-option -docstring "pattern that describes lines containing information about errors in the output of the `makecmd` command" \
+declare-option -docstring "pattern that find lines containing information about errors in the output of the `makecmd` command" \
     str make_error_pattern " (?:fatal )?error:"
+declare-option -docstring "pattern that describes lines containing information about errors in the output of the `makecmd` command" \
+    str make_error_line_pattern ""
```
krobelus, line 6: I guess we should use `regex` as type instead of `str`. Then we get early regex validation

Also we should add to the docstring something like

```
Capture groups must be:
    1: filename
    2: line number
    3: optional column
```

rc/tools/make.kak
```diff
@@ -51,25 +53,27 @@ define-command -hidden make-open-error -params 4 %{
...
         try %{
             execute-keys gl<a-?> "Entering directory" <ret><a-:>
             # Try to parse the error into capture groups, failing on absolute paths
             execute-keys s "Entering directory [`']([^']+)'.*\n([^:/][^:]*):(\d+):(?:(\d+):)?([^\n]+)\z" <ret>l
             set-option buffer make_current_error_line %val{cursor_line}
             make-open-error "%reg{1}/%reg{2}" "%reg{3}" "%reg{4}" "%reg{5}"
         } catch %{
-            execute-keys <a-h><a-l> s "((?:\w:)?[^:]+):(\d+):(?:(\d+):)?([^\n]+)\z" <ret>l
+            set-register / %sh{ if [ -z "$kak_opt_make_error_line_pattern" ] ; then printf %s "^(?:\w:)?[^:\n]+:\d+:(?:\d+:)?${kak_opt_make_error_pattern}([^\n]+)?\z" ; else printf %s "${kak_opt_make_error_line_pattern}([^\n]+)?\z"; fi }
+            execute-keys <a-h><a-l> s<ret>l
```
krobelus, line 65: How about this small change that would give nice autocompletion on `:set g make_error_line_pattern `
(Also we can hide the legacy option to reduce confusion).
Untested:

```diff
diff --git a/rc/tools/make.kak b/rc/tools/make.kak
index ab51090b1..39007e0fb 100644
--- a/rc/tools/make.kak
+++ b/rc/tools/make.kak
@@ -1,9 +1,9 @@
 declare-option -docstring "shell command run to build the project" \
     str makecmd make
-declare-option -docstring "pattern that find lines containing information about errors in the output of the `makecmd` command" \
+declare-option -hidden -docstring "pattern that find lines containing information about errors in the output of the `makecmd` command" \
     str make_error_pattern " (?:fatal )?error:"
 declare-option -docstring "pattern that describes lines containing information about errors in the output of the `makecmd` command" \
-    str make_error_line_pattern ""
+    str make_error_line_pattern "^(?:\w:)?[^:\n]+:\d+:(?:\d+:)? (?:fatal )?error:"
 
 declare-option -docstring "name of the client in which utilities display information" \
     str toolsclient
@@ -61,7 +61,7 @@ define-command -hidden make-jump %{
             set-option buffer make_current_error_line %val{cursor_line}
             make-open-error "%reg{1}/%reg{2}" "%reg{3}" "%reg{4}" "%reg{5}"
         } catch %{
-            set-register / %sh{ if [ -z "$kak_opt_make_error_line_pattern" ] ; then printf %s "^(?:\w:)?[^:\n]+:\d+:(?:\d+:)?${kak_opt_make_error_pattern}([^\n]+)?\z" ; else printf %s "${kak_opt_make_error_line_pattern}([^\n]+)?\z"; fi }
+            set-register / %sh{ if [ "$kak_opt_make_error_pattern" != " (?:fatal )?error:" ] ; then printf %s "^(?:\w:)?[^:\n]+:\d+:(?:\d+:)?${kak_opt_make_error_pattern}([^\n]+)?\z" ; else printf %s "${kak_opt_make_error_line_pattern}([^\n]+)?\z"; fi }
             execute-keys <a-h><a-l> s<ret>l
             set-option buffer make_current_error_line %val{cursor_line}
             make-open-error "%reg{1}" "%reg{2}" "%reg{3}" "%reg{4}"
@@ -72,7 +72,7 @@ define-command -hidden make-jump %{
 define-command make-next-error -docstring 'Jump to the next make error' %{
     evaluate-commands -try-client %opt{jumpclient} -save-regs / %{
         buffer '*make*'
-        set-register / %sh{ if [ -z "$kak_opt_make_error_line_pattern" ] ; then printf %s "^(?:\w:)?[^:\n]+:\d+:(?:\d+:)?$kak_opt_make_error_pattern" ; else printf %s "$kak_opt_make_error_line_pattern"; fi }
+        set-register / %sh{ if [ "$kak_opt_make_error_pattern" != " (?:fatal )?error:" ] ; then printf %s "^(?:\w:)?[^:\n]+:\d+:(?:\d+:)?$kak_opt_make_error_pattern" ; else printf %s "$kak_opt_make_error_line_pattern"; fi }
         execute-keys "%opt{make_current_error_line}ggl" "/<ret>"
         make-jump
     }
@@ -87,7 +87,7 @@ define-command make-next-error -docstring 'Jump to the next make error' %{
 define-command make-previous-error -docstring 'Jump to the previous make error' %{
     evaluate-commands -try-client %opt{jumpclient} -save-regs / %{
         buffer '*make*'
-        set-register / %sh{ if [ -z "$kak_opt_make_error_line_pattern" ] ; then printf %s "^(?:\w:)?[^:\n]+:\d+:(?:\d+:)?$kak_opt_make_error_pattern" ; else printf %s "$kak_opt_make_error_line_pattern"; fi }
+        set-register / %sh{ if [ "$kak_opt_make_error_pattern" != " (?:fatal )?error:" ] ; then printf %s "^(?:\w:)?[^:\n]+:\d+:(?:\d+:)?$kak_opt_make_error_pattern" ; else printf %s "$kak_opt_make_error_line_pattern"; fi }
         execute-keys "%opt{make_current_error_line}g" "<a-/><ret>"
         make-jump
     }
```

krobelus:
FWIW, kak-lsp provides a similar solution (`lsp_location_format`). I'm still not sure if it was a good approach (it's not very flexible) although it does work in practice for Python/Go stack traces.
Back when I had to deal with different error patterns, I wrote a script that transforms all locations into `file:line:column:` format and used that as `makecmd`.
I think it's a common use case to want to configure this.
In more complex situations they should copy the commands and maybe override them.

One difference to lsp_location_format is that the proposed `make_error_line_pattern` assumes that the diagnostic message (capture group 4) is the entire rest of the line. This might not always be true though I don't think we care, it's only used for a transient  message in the status line.

Overall this seems like a reasonable change. I think we can fix highlighting with a WinSetOption hook if we want that

rc/tools/make.kak
QiBaobin: Good point, changed. One reason that I use and test kak_opt_make_error_line_pattern if empty is that it's very easy to revert it back to default by set it to empty, but I think it matters not much.

rc/tools/make.kak
QiBaobin: done, thanks

rc/tools/make.kak
krobelus: yeah I think we should encourage users to use the new pattern.
Until I read the implementation I had found `make_error_pattern " (?:fatal )?error:"` to be a bit mysterious.
I think the regex that matches from the line start is easier to understand

QiBaobin: That makes a lot sense, thanks!

rc/tools/make.kak
```diff
@@ -1,7 +1,10 @@
 declare-option -docstring "shell command run to build the project" \
     str makecmd make
-declare-option -docstring "pattern that describes lines containing information about errors in the output of the `makecmd` command" \
+declare-option -hidden -docstring "pattern that find lines containing information about errors in the output of the `makecmd` command" \
     str make_error_pattern " (?:fatal )?error:"
+declare-option -docstring "pattern that describes lines containing information about errors in the output of the `makecmd` command. Capture groups must be: 1: filename 2: line number 3: optional column 4: optional error description" \
+    regex make_error_line_pattern "^(?:\w:)?([^:\n]+):(\d+):(?:(\d+):)? (?:fatal )?error:([^\n]+)?"
```
mawww, line 6: I think we could probably just change make_error pattern here, I figure you did it this way to avoid breaking user configuration but I am not sure this option is used much. I also wonder if we could use named captures here to support different ordering of capture groups, this would require some extra C++ side feature to be usable though,

QiBaobin: changed the code to reuse the old option now. 
Regarding named capture group, I leave it unchanged because I don't know how to refer named capture group, do we support it in the kak file right now?

mawww: No, this would need some extra work to expose capture name -> capture id from the regex. It would be a nice feature but I am not quite clear on the design yet.

PR #5011, opened 2023-11-02
rc/filetype/zig: remove align builtin and add f80 type

arrufat:
In commit https://github.com/mawww/kakoune/commit/ca36d0d2282dd304effe6326393751827c269210 the align function was added as a builtin accidentally.

Also, f80 type was added in Zig 0.10.0: https://ziglang.org/download/0.10.0/release-notes.html#f80


arrufat:
I was wondering whether I should keep this up to date with the latest syntax changes.

As far as I know, since the language is not stable, most people are using the development builds.

PR #4928, opened 2023-06-20
Fixup documentation on history navigation commands

omasanori:
This fixes documentation to align with 5901d2e06b7c43fd0e6156a49761b81b747ea908

PR #5012, opened 2023-11-03
mention <a-u> and <a-U> (selection undo/redo) in readme

tomsaleeba:
Follow up from https://github.com/mawww/kakoune/pull/4928, because there's another spot that mentions key bindings.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -539,8 +539,10 @@ Changes
 
  * `u`: undo last change
  * `<c-k>`: move backward in history
+ * `<a-u>`: undo selection changes. Introduced in v2023.07.29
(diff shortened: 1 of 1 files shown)


README.asciidoc
```diff
@@ -539,8 +539,10 @@ Changes
 
  * `u`: undo last change
  * `<c-k>`: move backward in history
+ * `<a-u>`: undo selection changes. Introduced in v2023.07.29
```
mawww, line 542: Those "Introduced in ..." are unnecessary, can you remove them please.

tomsaleeba: No problem, done 👍.

PR #5013, opened 2023-11-03
Fix "%val{selections_desc}" being joined by nul instead of space

krobelus:


mawww:
Wow, thats an embarassing bug, thanks for the fix.

PR #5015, opened 2023-11-03
rc/tools/git: edit a git indexed file with git edit

lobre:
A quick addition to allow the user to rapidly open a file known in the git index. This can be handy instead of having to rely on a simple "edit", especially for huge projects with a lot of "compiled artefacts" such as files in `node_modules/`.

Also, I chose to include `--recurse-submodules` as default for completion for `grep` and `edit` because `ls-files` is part of commands that don't respect the global configuration `submodule.recurse` (see https://git-scm.com/docs/git-config#Documentation/git-config.txt-submodulerecurse). And I think it makes sense to let the user grep or open a file in a submodule as well.

### DIFF
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -69,12 +69,13 @@ define-command -params 1.. \
             grep
     } -shell-script-candidates %{
     if [ $kak_token_to_complete -eq 0 ]; then
-        printf "add\nrm\nreset\nblame\ncommit\ncheckout\ndiff\nhide-blame\nhide-diff\nlog\nnext-hunk\nprev-hunk\nshow\nshow-branch\nshow-diff\ninit\nstatus\nupdate-diff\ngrep\n"
+        printf "add\nrm\nreset\nblame\ncommit\ncheckout\ndiff\nhide-blame\nhide-diff\nlog\nnext-hunk\nprev-hunk\nshow\nshow-branch\nshow-diff\ninit\nstatus\nupdate-diff\ngrep\nedit\n"
     else
         case "$1" in
             commit) printf -- "--amend\n--no-edit\n--all\n--reset-author\n--fixup\n--squash\n"; git ls-files -m ;;
             add) git ls-files -dmo --exclude-standard ;;
-            rm|grep) git ls-files -c ;;
+            rm) git ls-files -c ;;
+            grep|edit) git ls-files -c --recurse-submodules ;;
         esac
     fi
   } \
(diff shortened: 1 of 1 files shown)


rc/tools/git.kak
```diff
@@ -357,6 +358,10 @@ define-command -params 1.. \
                 set-option current grepcmd '$kak_opt_grepcmd'
             }"
             ;;
+        edit)
+            shift
+            echo "edit $@"
```
krobelus, line 363: good change, this allows my to drop a few lines from my config.
I'd make it `edit -existing -- $enquoted` (defining enquoted like above).
The `-existing` is debatable.

rc/tools/git.kak
```diff
@@ -69,12 +69,13 @@ define-command -params 1.. \
...
     if [ $kak_token_to_complete -eq 0 ]; then
-        printf "add\nrm\nreset\nblame\ncommit\ncheckout\ndiff\nhide-blame\nhide-diff\nlog\nnext-hunk\nprev-hunk\nshow\nshow-branch\nshow-diff\ninit\nstatus\nupdate-diff\ngrep\n"
+        printf "add\nrm\nreset\nblame\ncommit\ncheckout\ndiff\nhide-blame\nhide-diff\nlog\nnext-hunk\nprev-hunk\nshow\nshow-branch\nshow-diff\ninit\nstatus\nupdate-diff\ngrep\nedit\n"
     else
         case "$1" in
             commit) printf -- "--amend\n--no-edit\n--all\n--reset-author\n--fixup\n--squash\n"; git ls-files -m ;;
             add) git ls-files -dmo --exclude-standard ;;
-            rm|grep) git ls-files -c ;;
+            rm) git ls-files -c ;;
+            grep|edit) git ls-files -c --recurse-submodules ;;
```
krobelus, line 78: In my config I also have an alternative command that completes from top-level instead of cwd (`git ls-files "$(git rev-parse --show-toplevel)"`) that is useful occasionally. Let's worry about that later, not yet sure about the interface (separate command / flag / GIT_WORK_TREE option)

rc/tools/git.kak
lobre: I have modified it according to your comment. I think `-existing` makes sense, so I added it. Effectively, when searching for a specific git file, it is most of the time to quickly open a file that I know exists around, but I don't want to type the whole path.

For a new file, I guess the user would have to revert to a regular `edit` instead of `git edit`. By the way, this can be quickly done by going up in the command history, doing a `Ctrl-a` and `Alt-d` to remove the `git` part of the command and creating the file that was just searched.

krobelus: yes `-existing` is safer for deleted files that haven't yet been deleted in the Git index.
There is an argument that edit is still useful there -- because you could restore the file later and trigger autoreload -- but that requires more presence of mind from the user to not be confused. Probably better to fail early with `-existing`

rc/tools/git.kak
krobelus: yeah using `$PWD` is the correct default, not `git rev-parse --show-toplevel`.
Same for other git commands (this allows relative pathspecs for example).

krobelus (approved):

PR #5018, opened 2023-11-06
build: switch to POSIX make

svmhdvn:
This change rewrites the main Makefile to use strictly POSIX `make` supported features. The main changes are:
* Use macro expansion as a means for portable conditional statements across all `make` implementations
* Use the POSIX accepted (and widely implemented across all `make` implementations) `!=` shell command syntax
* Use POSIX compatible flags on shell commands

Tested on linux, freebsd, and macos using system provided `make`, CXX=g++, and CXX=clang++.


svmhdvn:
Not sure why CI is failing on macos. I have tested independently with a friend's help on macos and it builds successfully. Also tested locally on both linux and freebsd, works fine.

Makefile
```diff
@@ -1,7 +1,199 @@
...
+
+LDFLAGS-sanitize-undefined = -lubsan
+suffix-sanitize-undefined = .san_u
+
+LDFLAGS-static-yes = -static -pthread
+
+version != cat .version 2>/dev/null || git describe --tags HEAD 2>/dev/null || echo unknown
+
+sources != find src -type f -name '*.cc' | sed -e '/\.version\.cc/d'
+deps != touch src/main.d && find src -type f -name '*.d'
```
mawww, line 35: Why do we touch main.d here, would not that always trigger recompilation of main.cc ?

Makefile
```diff
@@ -1,7 +1,199 @@
...
+
+CPPFLAGS-os-OpenBSD = -DKAK_BIN_PATH="$(bindir)/kak" -I/usr/local/include
+LDFLAGS-os-OpenBSD = -L/usr/local/lib
+mandir-os-OpenBSD = $(DESTDIR)$(PREFIX)/man/man1
+
+LDFLAGS-os-SunOS = -lsocket -rdynamic
+
+is_windows != uname | grep -o '_NT'
+CPPFLAGS-is_windows-_NT = -D_XOPEN_SOURCE=700
+LIBS-is_windows-_NT = -ldbghelp
```
mawww, line 65: Is it possible to merge this into the `...-os-...` support? possibly by sed-ing uname when setting os ?

Makefile
```diff
@@ -1,7 +1,199 @@
...
+
+suffix = $(suffix-debug-$(debug))$(suffix-sanitize-$(sanitize))
+
+all: src/kak
+
+src/kak: src/kak$(suffix)
+	ln -sf kak$(suffix) $@
+
+src/kak$(suffix): src/.version.o $(objects)
+	$(CXX) $(KAK_LDFLAGS) $(LDFLAGS) $(KAK_CXXFLAGS) $(CXXFLAGS) $(objects) src/.version.o $(KAK_LIBS) $(LIBS) -o $@
```
mawww, line 103: maybe LDFLAGS/CXXFLAGS should be added to KAK_LDFLAGS/KAK_CXXFLAGS above and not appear here ?

mawww (commented):
This is a great change, getting rid of the last non-posix tool we require to build kakoune (well, excepted a C++20 compiler...). I will need to merge that locally to use it for a while, making sure it does not impact day to day workflow, and there are a few extra cleanups that would be nice to do.

svmhdvn:
Pushed v2, this further cleans up the use of non-POSIX tools like `install` for simple POSIX equivalents. I may need to do one more iteration on this work, but it seems to be working for me locally.

Could you also figure out how to fix the CI pipeline so that these checks aren't failing based on monthly compute credits? Is there another provider you could use?

mawww:
Any plans regarding the few inline comments ?

svmhdvn:
Just pushed a version with a couple fixes that I missed. Are there any outstanding inline comments now? I think I've resolved them all.

Makefile
mawww: This one ?

svmhdvn:
Done.

PR #5020, opened 2023-11-10
rc windowing: allow to configure windowing system and window placement in new/terminal commands

### LINKED ISSUE #3943: The built-in multiplexing commands lack flexibility

krobelus:
Today I can control "terminal" and "new" by changing the terminal
alias but I always need to choose a concrete implementation, like
"tmux-terminal-horizontal", even when there is otherwise no need to
mention tmux in my config.

Allow to configure windowing system and window placement independently
by introducing dedicated options.

This allows to create mappings that work in any windowing system like

	map global user c   %{:with-option windowing_placement window new<ret>}
	map global user '"' %{:with-option windowing_placement vertical new<ret>}
	map global user '%' %{:with-option windowing_placement horizontal new<ret>}

For windowing systems that don't support all placements, you can wrap
the above in try/catch to fall back on the "window" variant which is
defined for all windowing systems.

When using multiple (nested) windowing systems, you might want to
add mappings like

	map global user t %{:with-option windowing_module tmux new<ret>}
	map global user T %{:with-option windowing_module wayland new<ret>}

---

This changes the default "terminal" alias for some modules. In
particular, instead of delegating to

    iterm-terminal-vertical
    screen-terminal-vertical
    tmux-terminal-horizontal
    wezterm-terminal-vertical

it will now by default delegate to the respective "-window" variant.
We could maintain backwards compatiblity here by setting the
"windowing_placement" option accordingly, but the new behavior seems
more logical?

Also, this removes the "terminal-tab" alias which was only defined
by the kitty module.

We could try to keep the alias approach and implement a "with-alias"
command, however that approach can only capture both dimensions
(windowing system and placement) if we add tons of commands
like "terminal-horizontal" (with implied windowing system) and
"tmux-terminal" (with implied placement).

Side thought: we could also get rid of the "focus" alias and instead
define

    define-command focus %{
        "%opt{windowing_module}-focus"
    }

Closes #3943, #4425

### DIFF
--- a/rc/windowing/detection.kak
+++ b/rc/windowing/detection.kak
@@ -25,14 +25,62 @@ both automatic module loading and environment detection, enabling complete
 manual control of the module loading." \
 str-list windowing_modules 'tmux' 'screen' 'zellij' 'kitty' 'iterm' 'sway' 'wayland' 'x11' 'wezterm'
 
-hook -group windowing global KakBegin .* %{
+declare-option -docstring %{
+    windowing module to use in the 'terminal' command
+} str windowing_module
 
+declare-option -docstring %{
+    where to create new windows in the 'terminal' command.
+
+    Possible values:
+    - "window" (default) - new window
+    - "horizontal" - horizontal split (left/right)
+    - "vertical" - vertical split (top/bottom)
+    - "tab" - new tab besides current window
+} str windowing_placement window
+
+define-command terminal -params 1.. -docstring %{
+    terminal <program> [<arguments>]: create a new terminal using the preferred windowing environment and placement
+
+    This executes "%opt{windowing_module}-terminal-%opt{windowing_placement}" with the given arguments.
+    If the windowing module is 'wayland', 'sway' or 'x11', then the 'termcmd' option is used as terminal program.
+
+    Example usage:
+
+        terminal sh
+        with-option windowing_placement horizontal terminal sh
+
+    See also the 'new' command.
+} %{
+    "%opt{windowing_module}-terminal-%opt{windowing_placement}" %arg{@}
+}
+complete-command terminal shell
+
+# TODO Move this?
+define-command with-option -params 3.. -docstring %{
+    with-option <option_name> <new_value> <command> [<arguments>]: evaluate a command with a modified option
+} %{
+    evaluate-commands -save-regs s %{
+        evaluate-commands set-register s %exp{%%opt{%arg{1}}}
+        set-option current %arg{1} %arg{2}
+        evaluate-commands %sh{
+            shift 2
+            for arg
+            do
+                printf "'%s' " "$(printf %s "$arg" | sed "s/'/''/g")"
+            done
+        }
+        set-option current %arg{1} %reg{s}
+    }
+}
+
+hook -group windowing global KakBegin .* %{
     evaluate-commands %sh{
         set -- ${kak_opt_windowing_modules}
         if [ $# -gt 0 ]; then
             echo 'try %{ '
             while [ $# -ge 1 ]; do
-                echo "require-module ${1} } catch %{ "
+                echo "require-module ${1}; set-option global windowing_module ${1} } catch %{ "
                 shift
             done
             echo "echo -debug 'no windowing module detected' }"
--- a/rc/windowing/iterm.kak
+++ b/rc/windowing/iterm.kak
@@ -120,6 +120,5 @@ If no client is passed then the current one is used' \
 complete-command -menu iterm-focus client
 
 alias global focus iterm-focus
-alias global terminal iterm-terminal-vertical
 
 }
--- a/rc/windowing/kitty.kak
+++ b/rc/windowing/kitty.kak
@@ -10,8 +10,8 @@ evaluate-commands %sh{
 
 declare-option -docstring %{window type that kitty creates on new and repl calls (window|os-window)} str kitty_window_type window
 
-define-command kitty-terminal -params 1.. -docstring '
-kitty-terminal <program> [<arguments>]: create a new terminal as a kitty window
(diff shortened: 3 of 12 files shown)


lobre:
Great PR.

The code is much cleaner this way. Though I understand those mappings are easy to make and that we don't want to add too many default mappings, but would it make sense to at least add corresponding commands?

I know I can easily define something like:

```
define-command new-window -params .. %{ with-option windowing_placement window new "%arg{@}" }
define-command new-horizontal -params .. %{ with-option windowing_placement horizontal new "%arg{@}" }
define-command new-vertical -params .. %{ with-option windowing_placement vertical new "%arg{@}" }
```

But I feel that those are basic commands that could be default ones. In vim for example, it is really easy to create new splits by doing `:split` or `:vsplit`. Many people use them out of the box. They are handy enough.

Here with the current implementation, one still has to define its version of mappings or commands to benefit from easy splits (or type the full thing which is tedious).

And my second point is about REPL. I don't have a lot of knowledge of how they work and what is their purpose. To me, they allow to open a new terminal that is "connected" and controllable from kakoune itself. But now that we have a clean way to do new terminal windows (and new kakoune splits), should we also clean the way REPLs are implemented to reuse this system?

krobelus:
those sound tempting; but let's try to plan ahead: we might want to add

- `new-horizontal`
- `repl-horizontal`
- `tmux-terminal`
- `tmux-repl`

or future commands

- `diff-horizontal`
- `git-blame-horizontal`
- `gdb-horizontal`
- `fzf-horizontal`

more concretely, consider my

```kak
define-command -override tig -params .. %{
	terminal env "EDITOR=kak -c %val{session}" tig %arg{@}
}
define-command -override tig-blame -docstring "Run tig blame on the current line" %{
	tig -C %sh{git rev-parse --show-toplevel} blame -C "+%val{cursor_line}" -- %sh{
		dir="$(git rev-parse --show-toplevel)"
		printf %s "${kak_buffile##$dir/}"
	}
}
map global git b %{:tig-blame<ret>} -docstring "blame - Show what revision and author last modified each line of the current file"
```

I only ever invoke `new` / `tig-blame` / `ranger` etc. via mappings. Adding dedicated commands would mean 3x as many mappings.

I'm all for making simple use cases (`new`) more discoverable but before we add abbreviations I'd like to figure out a good UI that I works OOTB with anything that uses the `terminal` command
Maybe

```kak
map global view V '<a-semicolon>:set-option current windowing_placement vertical<ret>' -docstring 'place future windows in vertical split'
map global view B '<a-semicolon>:set-option current windowing_placement horizontal<ret>' -docstring 'place future windows in horizontal split'
map global view w '<a-semicolon>:set-option current windowing_placement window<ret>' -docstring 'place future windows in separate window'
map global view T '<a-semicolon>:set-option current windowing_placement tab<ret>' -docstring 'place future windows in a new tab'
```

or mappings that set the placement only for the next `terminal` command, after which it falls back to the global default. 

The keys are not ideal, maybe we should use a separate user mode instead of `view`.

I'm not against adding `new-horizontal` etc. but it should not cloud the user's view on what's possible.

> should we also clean the way REPLs are implemented to reuse this system?

I completely forgot about repl, I can fix that

lobre:
> but let's try to plan ahead

Agreed.

> Adding dedicated commands would mean 3x as many mappings.

To avoid complicating too much, maybe just 4 aliases such as below could be enough?

```
define-command as-new-window -params .. %{ with-option windowing_placement window "%arg{@}" }
define-command as-horizontal-window -params .. %{ with-option windowing_placement horizontal "%arg{@}" }
define-command as-vertical-window -params .. %{ with-option windowing_placement vertical "%arg{@}" }
define-command as-tab -params .. %{ with-option windowing_placement tab "%arg{@}" }

complete-command as-new-window command
complete-command as-horizontal-window command
complete-command as-vertical-window command
complete-command as-tab command
```

I feel this is handy in the sense that it is "typable" as a user command (not too long, no arguments, and easy to autocomplete). With this solution, I can define my own mappings, but I still have the option of just typing `:as-tab new`. Or this could be to call a custom command such as `:as-vertical-window mygitblame`.

Also, as a side note, I see that there is no `complete-command` defined for the current `with-option` command. Does it make sense to add one? It would require a custom one, as the first argument should be an option while the second one is a command.

krobelus:
while that wouldn't work for my mappings, it might make sense to have these abbreviations.

If we expect that 90% of `as-horizontal-window` calls will get `new` as command, then I don't think `as-horizontal-window` would pull its weight and we should prefer `new-horizontal`.
For other uses, we can type `set g wp horizontal`, actually we could allow abbreviating that then it's just `set g wp h`

lobre:
I am still thinking about all of this. I don't know if those mappings should be added as is. 

But overall, I think concepts in Kakoune should first materialize through commands before being adopted to mappings. That is the case for this `with-option`, but at first sight, they may be cumbersome to type manually. So users will preferably have to define mappings.

After 2 weeks of usage of my above commands, I may need to remove them and test without. See if manually calling  `set g wp horizontal` (using abbreviations) is convenient enough, but I am not convinced yet.

Overall, commands are more complicated to type in Kakoune than vim (despite abbreviations) as they are more verbose and comprise more keywords. However, the solution might not be to encourage the definition of shorter versions of commands just for the sake of convenience. There might be other solutions with a better design.

I would like to hear more about what others think about this.

krobelus:
Sure, you can define the `with-option` command if you want to use it. See `git log -Gwith-option -p`.
I don't expect it's a good idea to ship it by default because orthogonality of `set-option` probably weighs stronger than an abbreviation which is expected to be mainly used non-interactively.

PR #5030, opened 2023-11-16
filetype/hg-commit: Fix broken filetype detection and improve highlighting

herrhotzenplotz:
I typically use Mercurial with hg-git for my work. A few months ago I added my own hg-commit file type into my kakoune config not realising that there is one included in kakoune ootb. However, it is broken in that it doesn't properly detect mercurial VCS files, which is why I didn't discover it.

This patch fixes the file type detection and also adds a few nice highlighters that I had added to my own mecurial file type.

Also see the commit messages for details on my changes.

Please let me know if there's anything I should change.

### DIFF
--- a/rc/filetype/mercurial.kak
+++ b/rc/filetype/mercurial.kak
@@ -4,7 +4,7 @@
 # Detection
 # ‾‾‾‾‾‾‾‾‾
 
-hook global BufCreate .*hg-editor-\w+\.txt$ %{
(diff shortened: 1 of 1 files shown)


herrhotzenplotz:
Some CI checks are broken by themselves - not caused by me.

Also, I don't know why Github shows the commits in the wrong order, it is correct [here](https://github.com/herrhotzenplotz/kakoune/commits/feat/hg-highlight-keywords).

krobelus (approved):

PR #5033, opened 2023-11-17
ranked match: prefer when each path component matches within a path component

krobelus:
This implements a variant of the approach described in
https://nathancraddock.com/blog/2023/a-different-approach-to-fuzzy-finding/#making-filepath-matches-more-accurate


src/ranked_match.hh
```diff
@@ -52,6 +53,7 @@ private:
     StringView m_candidate{};
     bool m_matches = false;
     Flags m_flags = Flags::None;
+    Vector<RankedMatch, MemoryDomain::RankedMatch> m_path_component_matches;
```
krobelus, line 56: If these semantics are okay we could get rid of this allocation by computing the path component matches lazily in `RankedMatch::operator<`. I think this could work fine because we use a heap to get the top 100 or so completions, so we never need to sort everything; the number of calls to `RankedMatch::operator<` is linear.
I'd need to check.

src/ranked_match.cc
```diff
@@ -196,6 +229,13 @@ bool RankedMatch::operator<(const RankedMatch& other) const
     if (diff != Flags::None)
         return (int)(m_flags & diff) > (int)(other.m_flags & diff);
 
+    if (m_path_component_matches.empty() ^ other.m_path_component_matches.empty())
+        return m_path_component_matches.size() > other.m_path_component_matches.size();
+    kak_assert(m_path_component_matches.size() == other.m_path_component_matches.size());
+    for (auto [lhs, rhs] : zip(m_path_component_matches, other.m_path_component_matches))
+        if (*lhs < *rhs)
+            return true;
```
krobelus, line 237: probably if we use the same recursive-matching for the basename (even if the query contains no slash) then we can make @raiguard's test work:

`kak_assert(preferred("luaremote", "src/script/LuaRemote.cpp", "tests/TestLuaRemote.cpp"));`

because `luaremote` is a prefix (or is it?) of the basename so the RankedMatch will win

PR #5035, opened 2023-11-18
ranked match: prefer input order over alphabetical order for user-specified completions

### LINKED ISSUE #1709: Remove sorting of user-supplied insert completions

krobelus:
When using either of

	set-option g completers option=my_option
	prompt -shell-script-candidates ...

While the search text is empty, the completions will be sorted
alphabetically.
This is bad because it means the most important entries are not listed
first, making them harder to select or even spot.

Let's compare input order before sorting alphabetically.

In theory there is a more elegant solution: sort candidates (except if they're user input) before passing them to RankedMatch, and only use stable sort. However that doesn't work because we use a heap which doesn't support stable sort.

Closes #1709, #4813

### DIFF
--- a/rc/tools/ctags.kak
+++ b/rc/tools/ctags.kak
@@ -25,7 +25,7 @@ define-command -params ..1 \
                 cut -f 1 "$tags" | grep -v '^!' | uniq > "$namecache"
             fi
             cat "$namecache"
-        done} \
+        done | sort } \
     -docstring %{
         ctags-search [<symbol>]: jump to a symbol's definition
         If no symbol is passed then the current selection is used as symbol name
--- a/rc/tools/doc.kak
+++ b/rc/tools/doc.kak
@@ -189,7 +189,7 @@ complete-command doc shell-script-candidates %{
                     /^\[\[[^\]]+\]\]/ { sub(/^\[\[/, ""); sub(/\]\].*/, ""); print }
                 ' < $page | tr '[A-Z ]' '[a-z-]'
             fi;;
-    esac
+    esac | sort
 }
 
 alias global help doc
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -62,27 +62,49 @@ define-command -params 1.. \
         Available commands:
             add
             apply (alias for "patch git apply")
-            rm
-            reset
             blame
-            commit
             checkout
+            commit
             diff
+            edit
+            grep
             hide-blame
             hide-diff
             init
             log
             next-hunk
             prev-hunk
+            reset
+            rm
             show
             show-branch
             show-diff
             status
             update-diff
-            grep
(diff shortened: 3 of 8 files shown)


src/insert_completer.hh
```diff
@@ -53,12 +53,13 @@ struct InsertCompletion
...
     struct Candidate
     {
+        size_t input_order;
         String completion;
         String on_select;
         DisplayLine menu_entry;
 
         bool operator==(const Candidate& other) const { return completion == other.completion; }
-        auto operator<=>(const Candidate& other) const { return completion <=> other.completion; }
+        auto operator<=>(const Candidate& other) const { return std::tie(completion, input_order) <=> std::tie(other.completion, other.input_order); }
```
mawww, line 62: This is a bit weird, equality and ordering are not consistent with each other.

krobelus: true. Happily we don't need this change at all because the comparison is only used for deduplicating full-line completions. For those I  don't think we care about preserving the input order.
And if we do care, we could use a stable sort for `<c-x>l` completions.

Same for buffer completions, I don't have a strong opinion on whether they should respect input order.
I'm leaning towards no but if we do want it it'd be

```c++
index fb0b790dd..097895c98 100644
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -185,8 +185,8 @@ static Completions complete_buffer_name(const Context& context, CompletionFlags
         if (RankedMatch match{bufname, query})
             matches.emplace_back(match, buffer.get());
     }
-    std::sort(filename_matches.begin(), filename_matches.end());
-    std::sort(matches.begin(), matches.end());
+    std::stable_sort(filename_matches.begin(), filename_matches.end());
+    std::stable_sort(matches.begin(), matches.end());
 
     CandidateList res;
     for (auto& match : filename_matches)
```

rc/tools/ctags.kak
```diff
@@ -25,7 +25,7 @@ define-command -params ..1 \
                 cut -f 1 "$tags" | grep -v '^!' | uniq > "$namecache"
             fi
             cat "$namecache"
-        done} \
+        done | sort } \
```
mawww, line 28: Piping into sort means we wont get partial results in shell-script-candidates, which is a shame in the cases such as doc or man where we could get candidates progressively out of the find command. I wonder if sorting really is better than random order.

PR #5036, opened 2023-11-18
Fix use after move in HashMap::insert

krobelus:
Apparently GCC builds worked fine but Clang builds started failing the
"(hash == hash_value(item_key(item)))" assertion.


mawww:
BTW, I believe the reason this worked on gcc and not clang is because C++ does not enforce the ordering of parameter evaluation (since C++17 its guaranteed there is no interleaving), so I think clang was evaluating left-to-right and gcc right-to-left.

PR #5038, opened 2023-11-20
rc tools menu: replace menu builtin with a prompt-based implementation

krobelus:
prompt has fuzzy filtering which is more discoverable than the menu
mode's regex filtering (because that one needs / to trigger it).
There are no important differences left, so replace the menu builtin
with a prompt-based command.

prompt does not support markup in the completion menu, so drop that
feature for now.


ftonneau:
A comment about a possible downside of the new menu command. I have been refactoring and improving digraphs.kak, a plugin of mine that searches for Vim-like digraphs based on their two-character codes or short descriptions. Looking for LETTER, for example, will return 792 digraph candidates. With menu as a builtin, the menu used to populate instantaneously. With the new implementation, however, I see the "waiting for shell command to finish ..." message for about 3 seconds. 

Nothing life-threatening, of course, but annoying to those familiar with the older menu implementation. Plugins written by other authors (I have no specifics in mind) may also involve menus with hundreds of entries, which will require seconds to build.

krobelus:
yeah that's why I made https://github.com/mawww/kakoune/pull/5060 and now https://lists.sr.ht/~mawww/kakoune/patches/51802

Issue #4918, opened 2023-06-04
[QUESTION] remove entry from assistant


bagelwaffle:
### Question

I am one of those weird non-QWERTY users. I rebinded hjkl to UNEI (IJKL)

```
map global view h ''
map global view j ''
map global view k ''
map global view l ''
```
The above block is me unbinding HJKL from the "view mode". However, in doing so, I make a bunch of entries in the assistant (see the picture below)
![image](https://github.com/mawww/kakoune/assets/134188288/781ae409-6987-4d24-a810-8c7273c13e39)
Is there a way to hide the entries?

krobelus:
this isn't possible yet, these keys are currently hardcoded. I'm not sure what's the best way to fix this

bagelwaffle:
@krobelus, what do you think about something like `-hide-docstring`, would there be a chance of that being implemented if a feature request was made?

Thanks!

arachsys:
One could even just skip displaying any bindings that map to empty strings?

(Perhaps require an empty mapping *and* an empty docstring for that, to exclude weird cases like a mapping whose contents change depending on hooks, and might sometimes be empty, but which should always be listed with its description?)

arachsys:
For example, something like
```diff
diff --git a/src/normal.cc b/src/normal.cc
index d86ea3e..4fdd2fa 100644
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -190,8 +190,13 @@ String build_autoinfo_for_mapping(const Context& context, KeymapMode mode,
     }

     for (auto& key : keymaps.get_mapped_keys(mode))
+    {
+        if (keymaps.get_mapping_keys(key, mode).empty())
+            if (keymaps.get_mapping_docstring(key, mode).empty())
+                continue;
         descs.emplace_back(to_string(key),
                            keymaps.get_mapping_docstring(key, mode));
+    }

     auto max_len = 0_col;
     for (auto& desc : descs)
```
which I think probably does what you're after? (Have I missed any potential downsides of suppressing these empty mappings? We're not dropping them from `:debug mappings` output nor autocomplete on the `:unmap` command, just the info display.)

bagelwaffle:
I can not think of any downsides to this, since if the binding is empty and the docstring is also, it makes perfect sense to hide the binding from the assistant.

PR #5041, opened 2023-11-24
Hide empty and undocumented mappings from autoinfo

arachsys:
Users who rebind default keys and unmap the originals by binding them to empty strings with empty docstrings end up with empty lines in the autoinfo. For example, https://github.com/mawww/kakoune/issues/4918.

Hide completely null bindings which have both an empty mapping and an empty docstring in the autoinfo, as an easy mechanism for these users to eliminate the UI noise.

Fixes: https://github.com/mawww/kakoune/issues/4918

### DIFF
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -193,6 +193,8 @@ String build_autoinfo_for_mapping(const Context& context, KeymapMode mode,
     for (auto& key : keymaps.get_mapped_keys(mode))
     {
         const String& docstring = keymaps.get_mapping_docstring(key, mode);
+        if (keymaps.get_mapping_keys(key, mode).empty() and docstring.empty())
+            continue;
         if (auto it = find_if(descs, [&](auto& elem) { return elem.second == docstring; });
             it != descs.end())
             it->first += ',' + to_string(key);


krobelus (approved):

krobelus:
You can add a line like `Fixes #4918` to the commit message to close the issue

arachsys:
> You can add a line like `Fixes #4918` to the commit message to close the issue

Thanks, useful to know! I'm still a bit unused to GitHub's web UI compared to `git-send-email`, so often miss the intended way to do these things.

krobelus:
you can also use `git-send-email` I think; the `Fixes #xyz` in the commit message will work the same way

arachsys:
> you can also use `git-send-email` I think; the `Fixes #xyz` in the commit message will work the same way

I think I have to do web-based pull requests to send patches for Kakoune; there isn't a mailing list equivalent as far as I know? Or is there automatically a way of emailing a PR to any GitHub repository which I also haven't found yet? I've seen I can reply to PR discussion threads by email which is nice.

krobelus:
You can send patches by email to the maintainer directly.
I've also made a mailing list here https://lists.sr.ht/~krobelus/kakoune

arachsys:
> You can send patches by email to the maintainer directly.

Ah, I didn't want to create extra work for @mawww as I know he's super-busy, happy to send patches via GitHub given that's his preferred route. I was really just wondering if there was some automatic GitHub email interface I'd missed.

> I've also made a mailing list here https://lists.sr.ht/~krobelus/kakoune

Ah excellent, a Kakoune mailing list - I will definitely subscribe. I've often had things which aren't quite issues and aren't quite definite patches I'm proposing that I've not quite worked out the right place to send. :)

PR #5047, opened 2023-12-04
Perl files are not always executable

lobre:
The way kakoune is packaged on NixOS makes rc files not executable.

So in this example, `rc/tools/patch-range.pl` was not executable and it led to errors when trying to use `:patch`.

Solution here is to avoid relying on the shebang and directly call the script with perl.

Not sure exactly about all the usecases of `:patch` as I did not take long to test them. @krobelus , maybe you can have a quick look and see if it is still ok for you with this change?

Thanks

### DIFF
--- a/rc/tools/patch.kak
+++ b/rc/tools/patch.kak
@@ -54,7 +54,7 @@ define-command patch -params .. -docstring %{
                 # a shell where it expands to nothing.
                 eval set -- $kak_quoted_reg_a
 
-                "${kak_reg_f%/*}/patch-range.pl" $min_line $max_line "$@" ||
+                perl "${kak_reg_f%/*}/patch-range.pl" $min_line $max_line "$@" ||
                     echo >$kak_command_fifo "set-register e fail 'patch: failed to apply selections, see *debug* buffer'"
             }
             execute-keys |<ret>


krobelus:
sounds like a Nix problem. But sure, I guess we can do this

lobre:
Hmm, might be interesting to wonder where is the best place to do this before moving on.

Do you think that is the role of the packager to make sure files are executable when the package requires a script? Or should they trust git and preserve the permissions?

Also, as Nix should track precise dependencies in the sources, it should normally patch so that a precise version of perl is used. So I am not sure if it is easier for them to patch the shebang or the calling script. I will dig this.

krobelus:
> Do you think that is the role of the packager to make sure files are executable when the package requires a script? Or should they trust git and preserve the permissions?

I think in this case (e.g., a release tarball) the executable bit is not a security feature, only a safety feature.
So I think it'd be obvious to preserve permissions, the intent is clear.

That being said I haven't had time to research this; I think we should consult somthing like the Debian packaging standards. But in doubt we should take this, that's fine.

> Also, as Nix should track precise dependencies in the sources, it should normally patch so that a precise version of perl is used.
> So I am not sure if it is easier for them to patch the shebang or the calling script. I will dig this.

This problem is self-imposed by Nix. They go against common convention so that means a lot of effort for them (for perhaps little gain?). It forces users to go through the package manager in order to override binaries.

lobre:
> This problem is self-imposed by Nix.

Yes of course. I was not saying that something had to be done here on the Kakoune side. It was just a remark to say that anyway, Nix will have to patch whatever the solution. If we rely on the shebang, they should patch the shebang to inject a precise perl version. Otherwise, they will have to patch the kak file to replace that "perl" with their precise version.

Now about the executable bit, as you say, I am not sure what is the best solution. To me, this is the first time that we have a "proper shebang" script in `rc/`.

```
$ grep -r '#!/' rc/
rc/tools/patch-range.pl:#!/usr/bin/env perl
```

So we are talking about scripts that accompany/complement the main software. In a way, speaking about the intent, those scripts are not provided to the end user as being directly usable. But they are used through Kakoune. They are not meant to be called directly. In that regard, should they be executable at the system level?

I have no clue about the right solution here. Ideally, we could try to find another software that also has side scripts and see how it is packaged. But I don't have any at the top of my head.

lobre:
Just to complement my last answer, I see that kak-lsp uses some perl also in its `lsp.kak` script. But instead of having it in a separate file, it is inlined in the kak script.

https://github.com/kak-lsp/kak-lsp/blob/9cf04ffa3b83d99e208e8ceeaccb92d3d901b36e/rc/lsp.kak#L2436C17-L2436C18

So as it is not in an external file, there is not the same problem with the shebang, so it does not really help here. But I just wanted to reference it so that it illustrates what I was saying above with Nix. We can see that Nix packages kak-lsp by patching the kak file to replace the perl call with a precise one.

https://github.com/NixOS/nixpkgs/pull/217658/files

Especially this part:

```
substituteInPlace rc/lsp.kak \
      --subst-var-by perlPath ${lib.getBin perl}
```

For the initial issue, I am still searching for an application that would use external scripts so that I can dig its package in different distros. Then I can see if those scripts are used explicitly, or if they have shebang + executable bit.

If you have ideas for apps, don't hesitate.

PR #5055, opened 2023-12-13
Avoid semantically significant comments in kak files

arachsys:
Kakoune's balanced strings require that delimiter characters nested inside them are also paired, so for example in `%{ }`, each nested `{` must occur before a corresponding `}` to balance it out.

In general this will automatically be the case for code in common scripting languages, but sometimes regular expressions used for syntax highlighting do end up containing an unbalanced bracket of one type or another.

This problem is easily solved because there is a free choice of balanced delimiter characters. However, it can also be worked around by adding a comment which itself contains an unbalanced delimiter character, to 'balance out' the unpaired one in the regular expression.

These unbalanced comments are not ideal for readability as the semantic role they perform is easy for a casual reader to overlook. A good example is

```
    catch %{
        # indent after lines with an unclosed { or (
        try %< execute-keys -draft [c[({],[)}] <ret> <a-k> \A[({][^\n]*\n[^\n]*\n?\z <ret> j<a- gt> >
        # indent after a switch's case/default statements
        try %[ execute-keys -draft kx <a-k> ^\h*(case|default).*:$ <ret> j<a-gt> ]
        # deindent closing brace(s) when after cursor
        try %[ execute-keys -draft x <a-k> ^\h*[})] <ret> gh / [})] <ret> m <a-S> 1<a-&> ]
    }
```

in `rc/filetype/go/kak`. Here, it is not instantly obvious that the comment containing an unmatched `{` is required for correctness. If you change the comment, delete it or rearrange the contents of the catch block, `go.kak` will fail to load, and if you cut-and-paste this code as the basis for a new filetype, it is a loaded gun pointing at your feet.

Luckily, a careful audit of the standard Kakoune library turned up only three such instances, in `go.kak`, `hare.kak` and `markdown.kak`.

The examples in `go.kak` and `hare.kak` are easily made robust by replacing a `%{ }` with `%< >` or `%[ ]` respectively. The example in `markdown.kak` is least-intrusively fixed by rewriting the affected regular expression slightly so it has balanced `{,}` anyway.

### DIFF
--- a/rc/filetype/go.kak
+++ b/rc/filetype/go.kak
@@ -105,20 +105,20 @@ define-command -hidden go-indent-on-new-line %~
         try %{ execute-keys -draft <semicolon>K<a-&> }
         # cleanup trailing white spaces on the previous line
         try %{ execute-keys -draft kx s \h+$ <ret>d }
-        try %{
+        try %<
             try %{ # line comment
                 execute-keys -draft kx s ^\h*// <ret>
             } catch %{ # block comment
                 execute-keys -draft <a-?> /\* <ret> <a-K>\*/<ret>
             }
-        } catch %{
+        > catch %<
             # indent after lines with an unclosed { or (
             try %< execute-keys -draft [c[({],[)}] <ret> <a-k> \A[({][^\n]*\n[^\n]*\n?\z <ret> j<a-gt> >
             # indent after a switch's case/default statements
             try %[ execute-keys -draft kx <a-k> ^\h*(case|default).*:$ <ret> j<a-gt> ]
             # deindent closing brace(s) when after cursor
             try %[ execute-keys -draft x <a-k> ^\h*[})] <ret> gh / [})] <ret> m <a-S> 1<a-&> ]
-        }
+        >
     =
 ~
 
--- a/rc/filetype/markdown.kak
+++ b/rc/filetype/markdown.kak
@@ -93,7 +93,7 @@ add-highlighter shared/markdown/inline/text/ regex "\H( {2,})$" 1:+r@meta
 
 define-command markdown-load-languages -params 1 %{
     evaluate-commands -draft %{ try %{
-        execute-keys "%arg{1}s```\h*\{?[.=]?\K\w+<ret>" # }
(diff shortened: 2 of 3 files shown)


rc/filetype/go.kak
```diff
@@ -105,20 +105,20 @@ define-command -hidden go-indent-on-new-line %~
...
-        } catch %{
+        > catch %<
             # indent after lines with an unclosed { or (
             try %< execute-keys -draft [c[({],[)}] <ret> <a-k> \A[({][^\n]*\n[^\n]*\n?\z <ret> j<a-gt> >
             # indent after a switch's case/default statements
             try %[ execute-keys -draft kx <a-k> ^\h*(case|default).*:$ <ret> j<a-gt> ]
             # deindent closing brace(s) when after cursor
             try %[ execute-keys -draft x <a-k> ^\h*[})] <ret> gh / [})] <ret> m <a-S> 1<a-&> ]
-        }
+        >
```
krobelus, line 121: ok this fixes an evil pattern

rc/filetype/markdown.kak
```diff
@@ -93,7 +93,7 @@ add-highlighter shared/markdown/inline/text/ regex "\H( {2,})$" 1:+r@meta
 
 define-command markdown-load-languages -params 1 %{
     evaluate-commands -draft %{ try %{
-        execute-keys "%arg{1}s```\h*\{?[.=]?\K\w+<ret>" # }
+        execute-keys "%arg{1}1s```\h*\{?[.=]?(\w+)\}?<ret>"
```
krobelus, line 96: this one is fine I think because it's obvious that the comment exists to balance the braces

arachsys: Getting rid of all three of them might make it possible to write an automated test that warns about the evil pattern at the cost of also fixing this one that's ugly but not so unclear.

(My audit of /lib/kak was very nearly automated and I'm hoping I can turn it into a fully automated check suitable to add to the test suite. I'm not quite there yet, and you do have to make some 'sanity' assumptions about the contents of kak files, but none that all the existing files don't already obey.)

PR #5062, opened 2023-12-16
rc tools git: if $PWD is not inside a Git worktree, fall back to the buffer's worktree

krobelus:
When editing $PWD/some-project/foo and neither $PWD nor any of its
parents is a Git worktree (only $PWD/some-project/ is), then git
commands will fail. Let's use the buffer's worktree in this case.

When when running a command from a *git* buffer we face the same problem.
Fix this by using the worktree that was used to create the *git* buffer.

Closes #5003

### DIFF
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -106,6 +106,14 @@ define-command -params 1.. \
             update-diff \
         ;
     else
+        if git_work_tree=$(
+            git rev-parse --show-toplevel 2>/dev/null ||
+            git -C "${kak_buffile%/*}" rev-parse --show-toplevel 2>/dev/null
+        ); then
+            export GIT_WORK_TREE=${GIT_WORK_TREE:-"$git_work_tree"}
+            export GIT_DIR=${GIT_DIR:-"$GIT_WORK_TREE/.git"}
+        fi
+
         case "$1" in
             commit) printf -- "--amend\n--no-edit\n--all\n--reset-author\n--fixup\n--squash\n"; git ls-files -m ;;
             add) git ls-files -dmo --exclude-standard ;;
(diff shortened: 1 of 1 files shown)


krobelus:
@glaeqen after some offline discussion I find that this is maybe a bit too much magic.
While it's not breaking any use case that worked before, the behavior will be somewhat hard to understand, which might rob users of the opportunity to learn how to do things properly.

I think a command like `git diff foo/bar` should treat `foo/bar` as pathspec that's relative to `$PWD`, irrespective of buffile.

I believe it's very easy to figure out non-magic solutions. There's two issues

1. for your use case (`$PWD` is a parent of several git worktrees), we can implement support for `git -C dir`.
Then you can alias/map your git commands to start with `git -C %sh{your-logic-here} `.

2. certain commands in `*git*` buffers always want to use the worktree that
   was used to create the `*git*` buffer (but not all git commands as this
   PR implies).  We should look at all the git commands and find the ones that
   want this behavior. So far I think this includes `git-diff-goto-source`
   and `blame-jump`.  FWIW a simple `blame` outside $PWD already works today
   (because we pass absolute paths).

I plan to work on 2 and maybe also 1 if that turns out to be useful in practice.

krobelus:
actually instead of `git -C` you should define something like (untested)

```kak
define-command with-directory -params 2.. %{
	evaluate-commands -save-regs d %{
		set-register d %sh{pwd}
		try %{
			cd %arg{1}
			evaluate-commands %sh{
				shift
				for arg
				do
					printf " '%s'" "$(printf %s "$arg" | sed "s/'/''/g")"
				done
			}
		} catch %{
			cd %reg{d}
			fail %val{error}
		}
		cd %reg{d}
	}
}

# with-directory ../some-lib git diff
```

PR #5063, opened 2023-12-16
Option to override environment variables in child processes

### LINKED ISSUE #4482: Ability to change session and client environment variables
I was wondering if we could change the session and client environment variables once Kakoune has started.

Having a way to access and change them could be nice to

- set and update Kakoune environment without having to restart Kakoune in a new shell,
- and fine-tune tools to Kakoune client-server model.

Kakoune already does some path manipulation (#4433) by prefixing the currently running `kak` binary in `PATH`, and it could be a good way to expose shell scripts in Kakoune plugins (#4425).

The proposal is to add

- `set-environment [session|client] <name> value`,
- `unset-environment [session|client] <name>`,
- and `%env{name}` expansion.

Here is a plugin example with the following files:

```
bin/:e -> :edit
bin/:edit
bin/:fzf
rc/connect.kak
```

`rc/connect.kak`

``` kak
# Exposes bin in PATH.
set-environment session PATH "%sh{dirname ""$kak_source""}/../bin:%env{PATH}"

# Exposes the session and client environment variables,
# so :edit command has the necessary information to open files in a client/session.
#
# Note: needs a hook to handle session rename.
set-environment session kak_session %val{session}
hook global ClientCreate '.+' %{
  set-environment client kak_client %val{client}
}
```

`bin/:edit`

``` sh
#!/bin/sh

export file
for file do
  echo evaluate-commands -client %env{kak_client} -- edit -- %env{file} |
  kak -p "$kak_session"
  # Note: We use the %env{name} expansion, so we don’t have to “kak quote” values—passing values is straightforward.
  # We expose the `file` environment variable to the `kak` binary, and `kak_client` is already exported.
  #
  # Having `kak -p session_name -args -- 1 2 3` could be nice to pass values in %arg{@} expansion.
done
```

`bin/:fzf`

``` sh
find "$@" -type f | fzf | xargs :edit --
```

In Kakoune, we can now do:

``` kak
set-environment client FZF_DEFAULT_OPTS '--height=40% --layout=reverse --border'
terminal :fzf
```

Assuming we have a Kakoune command to spawn a program, it still convenient for interactive usage without writing a wrapper for every windowing system.

``` kak
spawn alacritty -e :fzf
spawn tmux split :fzf path/to/dir
```

Since we have `:edit` and `:fzf` in `PATH`, we can open a terminal from Kakoune and use the commands in the shell.

``` kak
spawn alacritty
```

then in the terminal:

``` sh
:e docs/*.txt
:fzf
```

krobelus:
Today I have to restart Kakoune in order to set environment variables
in shell expansions globally. For example KAK_LSP_FORCE_PROJECT_ROOT.

When running ":git" commands, there is an ambiguity on whether to
use the $PWD or the buffer's worktree.  We use $PWD but it should be
possible to override this behavior.  An obvious way to do this is to
override Git environment variables:

	set-option -add global env GIT_WORK_TREE=/path/to/my/repo GIT_DIR=/path/to/my/repo.git

(another in-flight patch adds the obvious default to GIT_DIR so we don't need to set that).

---

There are some minor issues left
- If a user sets PATH, this will stomp the one we setenv()'d.
- Today both key and value require escaping for = and \. This is not
  intuitive. Since neither environment variables nor ui_options ever
  need a = in the key name, we can get rid of the escaping.

Alternative approach: The str-to-str-map can be somewhat clunky. We
could instead export any option named like env_FOO. Not yet sure
which approach is better.

Closes #4482
That issues asks for a separate client-specific env as well but we
don't have a client scope so I'm not sure.

### DIFF
--- a/src/shell_manager.cc
+++ b/src/shell_manager.cc
@@ -146,6 +146,8 @@ Vector<String> generate_env(StringView cmdline, const Context& context, GetValue
     static const Regex re(R"(\bkak_(quoted_)?(\w+)\b)");
 
     Vector<String> env;
+    for (const auto& [key, value] : context.options()["env"].get<HashMap<String, String, MemoryDomain::Options>>())
+        env.push_back(format("{}={}", key, value));
     for (auto&& match : RegexIterator{cmdline.begin(), cmdline.end(), re})
     {
         StringView name{match[2].first, match[2].second};
--- a/doc/pages/options.asciidoc
+++ b/doc/pages/options.asciidoc
@@ -189,6 +189,9 @@ are exclusively available to built-in options.
     key is provided it removes that entry regardless of the associated
     value. +
 
+    Any `=` or `\` characters that in `key` and `value` must be escaped as
+    `\=` or `\\`.
+
     Only `str-to-str-map` options can be created with `declare-option`.
 
 == Builtin options
@@ -339,6 +342,9 @@ are exclusively available to built-in options.
             in face Information.
 
     The default value is '%val{bufname} %val{cursor_line}:%val{cursor_char_column} {{context_info}} {{mode_info}} - %val{client}@[%val{session}]'
+
+*env* `str-to-str-map`::
+    a list of environment variable overrides to be passed to external processes.
 
 *ui_options* `str-to-str-map`::
     a list of `key=value` pairs that are forwarded to the user
--- a/src/main.cc
+++ b/src/main.cc
@@ -574,6 +574,9 @@ void register_options()
     reg.declare_option<int, check_timeout>(
         "fs_check_timeout", "timeout, in milliseconds, between file system buffer modification checks",
         500);
+    reg.declare_option("env",
+                       "a list of environment variable overrides to be passed to external processes",
+                       HashMap<String, String, MemoryDomain::Options>{});
     reg.declare_option("ui_options",
                        "space separated list of <key>=<value> options that are "
                        "passed to and interpreted by the user interface\n"


src/shell_manager.cc
```diff
@@ -146,6 +146,8 @@ Vector<String> generate_env(StringView cmdline, const Context& context, GetValue
     static const Regex re(R"(\bkak_(quoted_)?(\w+)\b)");
 
     Vector<String> env;
+    for (const auto& [key, value] : context.options()["env"].get<HashMap<String, String, MemoryDomain::Options>>())
+        env.push_back(format("{}={}", key, value));
```
krobelus, line 150: This is an ergonomic abstraction; notably it's incomplete in that it does not allow to unset environment variables, and neither does it allow erase the entire environment (though I don't think there is a use case for that).
A more flexible (but probably not better) approach would be to allow overriding `KAKOUNE_POSIX_SHELL` dynamically, then we can set it to `env ... "$@"`..

PJungkamp:
I'd like to integrate kakoune with `direnv`. Since `direnv` relies on changing environment variables, I'd need to be able to propagate the changed environment to a running kakoune session...

Is there anything happening around here?

krobelus:
Nothing wrong with using this patch (I do use it occasionally).

One alternative approach is to simply add `setenv` and `unsetenv` commands, which is more obvious than the above `str-to-str-map`. I believe that would suffice for the direnv use case. Probably this should change places where we do `getenv("XDG_RUNTIME_DIR")` to instead cache the value at startup, to prevent weird states.

The advantage of the `str-to-str-map` is that it's easy to undo an override but for the use cases I had so far, `unsetenv` works too.

> I'd like to integrate kakoune with direnv.

This means that the `:cd` command might apply a new `envrc`.
Sounds fine. I guess usually Kakoune doesn't accumulate much state, in that case restarting should work.

PR #5081, opened 2024-01-13
rc tools ctags/man: rationalize sorting of completion candidates

krobelus:
As of recently, shell script candidate completions are computed in
the background, and displayed incrementally.
Sorting completions means we can't show partial results.

We do this for "ctags-search"; but if there is only one ctags file
there is already a sensible order (which is slightly different than
what GNU sort does). So let's preserve the order in that case.
The number of completions is probably too high for an order to be useful.

Similarly, for "man", sorting is probably not very helpful because
there are so many results.

See https://github.com/mawww/kakoune/pull/5035#discussion_r1413015934


mawww:
Merged, however I wonder if we actually need to sort anything, respecting the order in which tag files are configured might make sense as well for sorting completions.

krobelus:
Yeah, keeping the order of ctags files sounds fine too. I wouldn't know since I never used multiple ctags files.

PR #5084, opened 2024-01-16
rc tools git: command for easy recursive blaming

krobelus:
(EDIT This description is mildly outdated, refer to the commit messages)

I frequently want to find out why a piece of code was introduced.
We have ":git blame" to annotate each line with the most recent commit.
However often a line has been modified by several commits.

Introduce ":git show-blamed" which shows the commit that added the
line at cursor. Crucially, it works also in Git diff buffers, so it can
be used recursively to find the full history of a line.

Previously I used Tig's "b" shortcut. In diff views, I would navigate
to the old ("-"-prefixed) version of the line, press "b" and repeat.
In most cases, it's a boring task to navigate to the old version
because that one is usually located very close to the new version.
In Tig I use "/" to search for it.

Automate this common case by making ":git show-blamed" jump to
the blamed line directly. This means the initial diff view might
not include the commit message etc. Compensate this by showing the
commit's date+author+subject in the status line.

Future work:

- The existing ":git blame" should also learn to work inside
  ":git show" buffers, much like Tig/Magit/Fugitive etc..
- To go back to the previously-blamed commit we need to have had the
  foresight to use "rename-buffer". Perhaps we want to add some kind
  of buffer stack (like Magit does for example).

### DIFF
--- a/doc/pages/changelog.asciidoc
+++ b/doc/pages/changelog.asciidoc
@@ -18,7 +18,11 @@ released versions.
 
 * View mode commands and mouse scrolling no longer change selections when those go off-screen.
 
-* New commands `git apply`, `git edit`, `git grep`
+* New commands `git apply`, `git edit`, `git grep` and `git show-blamed`.
+
+* When git blame annotations are shown, press `<ret>` to jump to blamed commit.
+
+* `git blame` now also works in git diff buffers.
 
 == Kakoune 2023.08.08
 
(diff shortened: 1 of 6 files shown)


mawww:
That sounds great, I very often select the commit id from blame with the mouse to `git show` it, not having to use the mouse for that would be really nice.

krobelus:
It's already useful but since blame can be slow on large repos, I'll probably make it reuse `:git blame` annotations (`flag-lines` tells us if the line is still valid).
We don't even need to expose/provide `git show-blamed` because filtering down to a single line does not find that line faster than a full incremental blame.

krobelus:
mostly happy with the UI now. I'll test it some more

PR #5087, opened 2024-01-19
ci: move linux and macos to Github Actions

svmhdvn:
This change replaces Cirrus CI tasks for linux and macos with Github Actions equivalents.

Github actions doesn't seem to have a "compute credits" or similar restriction for public instances. Since kakoune is mainly hosted on github, it's also really well integrated with the pull request flow.

See https://github.com/svmhdvn/kakoune/actions/runs/7578698028 for a successful run.

### DIFF
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -1,0 +1,44 @@
+name: Makefile CI
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+jobs:
+  build:
+    strategy:
+      matrix:
+        os: [ubuntu, macos]
+        toolchain: [gcc, llvm]
+        include:
+          - os: ubuntu
+            toolchain: gcc
+            install_cmd: |
+              sudo apt update
+              sudo apt install g++
+            CXX: g++
+          - os: ubuntu
+            toolchain: llvm
+            install_cmd: |
+              sudo apt update
+              sudo apt install clang
+            CXX: clang++
+          - os: macos
+            toolchain: gcc
+            install_cmd: |
+              brew update
+              brew install gcc
+            CXX: g++
+          - os: macos
+            toolchain: llvm
+            install_cmd: |
+              brew update
+              brew install llvm
+            CXX: clang++
+    runs-on: ${{ matrix.os }}-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install dependencies
+      run: ${{ matrix.install_cmd }}
+    - name: Build and test
+      run: make CXX=${{ matrix.CXX }} -j$(nproc 2>/dev/null || sysctl -n hw.ncpu) test
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,3 +1,4 @@
+# TODO move to Github Actions after https://github.com/actions/runner/issues/385
 freebsd_task:
     freebsd_instance:
         image_family: freebsd-14-2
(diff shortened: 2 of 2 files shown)


krobelus:
Perhaps the problem with Cirrus was that a hung test (which has been fixed since) used up all available time.
This change makes it so we no longer specify the compiler version explicitly, and moves from `g++-10` to `g++-11`. Both seem fixable if desired.

svmhdvn:
> Perhaps the problem with Cirrus was that a hung test (which has been fixed since) used up all available time.

Sure, but I still find it more valuable to be free from these limits in the first place, and to reduce further dependencies on other services outside of the (implied constant) Github.

> Both seem fixable if desired.

Yes exactly, I can easily change the toolchain versions (or add more to the matrix) if desired by kakoune's maintainers.

krobelus:
Looks like Cirrus works again.
In the meantime we also added [Sourcehut CI](https://builds.sr.ht/~mawww/kakoune) which covers everything except macOS.
The status quo seems acceptable but let's see what the future brings

.github/workflows/makefile.yml
```diff
@@ -0,0 +1,44 @@
...
+              brew update
+              brew install llvm
+            CXX: clang++
+    runs-on: ${{ matrix.os }}-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install dependencies
+      run: ${{ matrix.install_cmd }}
+    - name: Build and test
+      run: make CXX=${{ matrix.CXX }} -j4 test
```
krobelus, line 44: IIRC, runners will only have one (virtual) CPU.
so maybe we should use `-j$(nproc 2>/dev/null || sysctl -n hw.ncpu)`.
Not a big deal

krobelus:
> I'd like to revisit this as the same issue is occurring. Any comments on this, either an approval or rejection?

Thanks for flagging this. I'd be happy to see this land.
I wonder if the compute-credit problem also happens for Cirrus' FreeBSD images.
Sourcehut CI is nicer to work with, but this gives us macOS coverage, which is a nice addition.

Can we use the same container images as for cirrus? i.e. clang 13 and gcc 10?
I'd guess it would be easier to accept this PR if it is limited only changing the CI provider.
The status quo means we need to bump compiler versions explicitly, which seems potentially useful

svmhdvn:
> Sourcehut CI is nicer to work with, but this gives us macOS coverage, which is a nice addition.

But people contribute to kakoune on Github, and thus will mostly pay attention to (and will notice) the results of their PR CI runs on Github. This might be better for catching developer errors more quickly.

> I'd guess it would be easier to accept this PR if it is limited only changing the CI provider. The status quo means we need to bump compiler versions explicitly, which seems potentially useful.

Why is this an important consideration? Since kakoune uses the latest c++ standard, IMO it would be beneficial to always be tracking the latest compilers, which my PR reflects. Additionally, developers are most likely to be running those compilers as well, since it uses out-of-the-box package manager versions.

.github/workflows/makefile.yml
svmhdvn: Just for reference, it seems most of them have 4 CPUs: https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners#standard-github-hosted-runners-for-public-repositories

However, I like your approach, I made these changes now.

mawww:
I try to keep Kakoune compiling on whatever debian stable compiler is, so ideally that would be the version that is tested. This is already tested by the sourcehut build (which uses debian images) so we are covered on that front. It would be nice to see breakage on github pull requests before merging, but to be honest I do not often check the CI for PRs, and AFAIR its pretty rare that we break the debian stable build.

Issue #50, opened 2014-02-04
Should Kakoune have a parser?


alexherbo2:
Use case:
- [PEG](http://en.wikipedia.org/wiki/Parsing_expression_grammar) to handle syntax highlighting

Example:

``` peg
; JSON grammar
; see http://json.org

JSON ← S? ( Object / Array / String / True / False / Null / Number ) S?

Object ← "{"
             ( String ":" JSON ( "," String ":" JSON )*
             / S? )
         "}"

Array ← "["
            ( JSON ( "," JSON )*
            / S? )
        "]"

String ← S? ["] ( [^ " \ U+0000-U+001F ] / Escape )* ["] S?

Escape ← [\] ( [ " / \ b f n r t ] / UnicodeEscape )

UnicodeEscape ← "u" [0-9A-Fa-f]{4}

True ← "true"

False ← "false"

Null ← "null"

Number ← Minus? IntegralPart FractionalPart? ExponentPart?

Minus ← "-"

IntegralPart ← "0" / [1-9] [0-9]*

FractionalPart ← "." [0-9]+

ExponentPart ← ( "e" / "E" ) ( "+" / "-" )? [0-9]+

S ← [ U+0009 U+000A U+000D U+0020 ]+
```

haarts:
I know this is an ancient issue. But recently I found this project: http://tree-sitter.github.io/tree-sitter/ (created by the Github organisation). It seems to address exactly the issues raised in https://github.com/mawww/kakoune/issues/50#issuecomment-34294369 by @mawww. From the website:
-    General enough to parse any programming language
- Fast enough to parse on every keystroke in a text editor
- Robust enough to provide useful results even in the presence of syntax errors
- Dependency-free so that the runtime library (which is written in pure C) can be embedded in any application

It has been around for a while now speaking to it's maturity.

Delapouite:
Related ongoing PR about tree-sitter: https://github.com/mawww/kakoune/pull/3160 
Also https://github.com/ul/kak-tree for selectors

haarts:
That's pretty spectacular. I hadn't even thought about selections! Syntax highlighting was my annoyance. It's just too damn hard to write regexes (think code in string interpolations, ugh).

PR #5099, opened 2024-02-06
Add tree-sitter based highlighter

### LINKED ISSUE #3056: Regions are matched greedily, not recursively
Consider the same context as in #3055, adding this time matching nested scopes i.e. `$(… $(…)…)`.

This issue probably also depends on #3055.

### Steps 

Same as in #3055.

### Outcome 

If #3055 were fixed, only the opening and closing delimiters of the region would be matched.

### Expected 

Regions are detected as being nested, and so all opening and closing delimiters are highlighted.

It's my understanding that when regions are nested, only the one with the largest scope (which contains the nested scopes) will be highlighted, which makes relying on `\A` and `\z` moot.

PJungkamp:
# Description

This introduces a `tree-sitter` highlighter that maps tree-sitter captures to kakoune faces.
This allows kakoune to highlight some recursive grammars that the regions based highlighting is not able to parse (e.g. Nix, Shell, Python).

- The default mapping from query captures to kakounes faces is based on [nvim-treesitter](https://neovim.io/doc/user/treesitter#treesitter-highlight).
- A subset of language injections are supported, these can be added as tree-sitter-injection sub-highlighters.

# Example - Nix

This is an example from a Nix codebase i recently visited.

The problem here is that a Nix string can contain interpolated nix expression, which in turn can have nested strings with the same delimiter (e.g `"outer string ${ let var = "inner string"; in var }"`).

- Regions based highlighting:
Screenshot from 2024-02-06 14-08-50
---

- tree-sitter based highlighting:
Screenshot from 2024-02-06 14-08-16

# Building

This highlighter is optional and can be excluded by passing `tree_sitter=no` to `make`.

# Related Issues

- Maybe fixes #3056

# TODO

- [ ] proper tree edits with cached trees (currently reparses the whole file instead of adjusting trees incrementally)
- [ ] lazily compile grammars (if precompiled grammars are part of the the configuration tree a config couldn't be shared across architectures)
- [ ] decide on a file path for grammars and queries (currently `%val{runtime}/grammars`)
- [ ] support more injections (especially `combined` injections this would allow us to e.g. highlight the embedded Bash in the example above)

### DIFF
--- a/rc/filetype/markdown.kak
+++ b/rc/filetype/markdown.kak
@@ -41,70 +41,85 @@ provide-module markdown %{
 # Highlighters
 # ‾‾‾‾‾‾‾‾‾‾‾‾
 
-add-highlighter shared/markdown regions
-add-highlighter shared/markdown/inline default-region regions
-add-highlighter shared/markdown/inline/text default-region group
-
-add-highlighter shared/markdown/listblock region ^\h*[-*]\s ^(?=\S) regions
-add-highlighter shared/markdown/listblock/g default-region group
-add-highlighter shared/markdown/listblock/g/ ref markdown/inline
-add-highlighter shared/markdown/listblock/g/marker regex ^\h*([-*])\s 1:bullet
-
-add-highlighter shared/markdown/codeblock region -match-capture \
-    ^(\h*)```\h* \
-    ^(\h*)```\h*$ \
-    regions
-add-highlighter shared/markdown/codeblock/ default-region fill meta
-add-highlighter shared/markdown/listblock/codeblock region -match-capture \
-    ^(\h*)```\h* \
-    ^(\h*)```\h*$ \
-    regions
-add-highlighter shared/markdown/listblock/codeblock/ default-region fill meta
-add-highlighter shared/markdown/codeline region "^( {4}|\t)" "$" fill meta
-
-# https://spec.commonmark.org/0.29/#link-destination
-add-highlighter shared/markdown/angle_bracket_url region (?<=<)([a-z]+://|(mailto|magnet|xmpp):) (?!\\).(?=>)|\n fill link
-add-highlighter shared/markdown/inline/url region -recurse \( (\b[a-z]+://|(mailto|magnet|xmpp):) (?!\\).(?=\))|\s fill link
-add-highlighter shared/markdown/listblock/angle_bracket_url region (?<=<)(\b[a-z]+://|(mailto|magnet|xmpp):) (?!\\).(?=>)|\n fill link
-
 try %{
-    require-module html
-    add-highlighter shared/markdown/inline/tag region (?i)</?[a-z][a-z0-9-]*\s*([a-z_:]|(?=>)) > ref html/tag
+    require-module tree-sitter
+    tree-sitter-load-highlighter shared/markdown markdown %opt{tree_sitter_default_faces}
+
+    define-command markdown-load-languages -params 1 %{
+        evaluate-commands -draft %{ try %{
+            execute-keys "%arg{1}s```\h*\{?[.=]?\K\w+<ret>" # }
+            evaluate-commands -itersel %{ try %{
+                require-module %val{selection}
+                add-highlighter "shared/markdown/%val{selection}" tree-sitter-injection ref %val{selection}
+            }}
+        }}
+    }
+} catch %{
+    add-highlighter shared/markdown regions
+    add-highlighter shared/markdown/inline default-region regions
+    add-highlighter shared/markdown/inline/text default-region group
+
+    add-highlighter shared/markdown/listblock region ^\h*[-*]\s ^(?=\S) regions
+    add-highlighter shared/markdown/listblock/g default-region group
+    add-highlighter shared/markdown/listblock/g/ ref markdown/inline
+    add-highlighter shared/markdown/listblock/g/marker regex ^\h*([-*])\s 1:bullet
+
+    add-highlighter shared/markdown/codeblock region -match-capture \
+        ^(\h*)```\h* \
+        ^(\h*)```\h*$ \
+        regions
+    add-highlighter shared/markdown/codeblock/ default-region fill meta
+    add-highlighter shared/markdown/listblock/codeblock region -match-capture \
+        ^(\h*)```\h* \
+        ^(\h*)```\h*$ \
+        regions
+    add-highlighter shared/markdown/listblock/codeblock/ default-region fill meta
+    add-highlighter shared/markdown/codeline region "^( {4}|\t)" "$" fill meta
+
+    # https://spec.commonmark.org/0.29/#link-destination
+    add-highlighter shared/markdown/angle_bracket_url region (?<=<)([a-z]+://|(mailto|magnet|xmpp):) (?!\\).(?=>)|\n fill link
+    add-highlighter shared/markdown/inline/url region -recurse \( (\b[a-z]+://|(mailto|magnet|xmpp):) (?!\\).(?=\))|\s fill link
+    add-highlighter shared/markdown/listblock/angle_bracket_url region (?<=<)(\b[a-z]+://|(mailto|magnet|xmpp):) (?!\\).(?=>)|\n fill link
+
+    try %{
+        require-module html
+        add-highlighter shared/markdown/inline/tag region (?i)</?[a-z][a-z0-9-]*\s*([a-z_:]|(?=>)) > ref html/tag
+    }
+
+    add-highlighter shared/markdown/inline/code region -match-capture (`+) (`+) fill mono
+
+    # Setext-style header
+    add-highlighter shared/markdown/inline/text/ regex (\A|^\n)[^\n]+\n={2,}\h*\n\h*$ 0:title
+    add-highlighter shared/markdown/inline/text/ regex (\A|^\n)[^\n]+\n-{2,}\h*\n\h*$ 0:header
+
+    # Atx-style header
+    add-highlighter shared/markdown/inline/text/ regex ^#[^\n]* 0:header
+
+    add-highlighter shared/markdown/inline/text/ regex (?<!\*)(\*([^\s*]|([^\s*](\n?[^\n*])*[^\s*]))\*)(?!\*) 1:+i
+    add-highlighter shared/markdown/inline/text/ regex (?<!_)(_([^\s_]|([^\s_](\n?[^\n_])*[^\s_]))_)(?!_) 1:+i
+    add-highlighter shared/markdown/inline/text/ regex (?<!\*)(\*\*([^\s*]|([^\s*](\n?[^\n*])*[^\s*]))\*\*)(?!\*) 1:+b
+    add-highlighter shared/markdown/inline/text/ regex (?<!_)(__([^\s_]|([^\s_](\n?[^\n_])*[^\s_]))__)(?!_) 1:+b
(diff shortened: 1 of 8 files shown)


tototest99:
Hello,
are you aware of [kak-tree-sitter](https://github.com/phaazon/kak-tree-sitter) and [merge initiative with redondant LSP features](https://github.com/kak-lsp/kak-lsp/issues/712) ?

phaazon:
Hello,

I think it’s an interesting approach, but in the same time, it bothers me a bit. Not because I’m the author of `kak-tree-sitter` and I spent a lot of time working on it. But because of the fact that I love the philosophy behind Kakoune, and I think you’re bending it. The reason for that is that Kakoune ships with zero dependencies. What makes it use external tools is basically just calling-out to external programs via `%sh{}` or `kak -p`. You would argue that you have `:git` commands, but look closer: those are just Kakoune commands wrapping around the `git`, `awk`, `perl` etc. commands.

However, I do think that what I made with KTS is super complex and that Kakoune needs more toolings to make it easier to integrate (and I think @krobelus would also benefit from that for `kak-lsp`). For instance, in KTS, I have to start a daemonized server to handle a parsed representation of buffers for a set of sessions. That means that Kakoune must stream its buffers’ content to KTS (which can be expensive). I optimized that by writing directly to FIFO opened by KTS (i.e. that means that there is no shell creation to do so, it’s just a basic, low-level `write` to the FIFO). Even with that, I think we could do better, because you need to do that for all « integrations » (KTS, `kak-lsp`, `kak-whatnot`, etc.).

mawww:
This is indeed an impressive PR, but I do not intend to merge it for a few reason.

First I do not want to have multiple, competing, built-in highlighting to maintain, and I do not want to solely rely on tree-sitter for highlighting. Having both will likely lead to one bitrotting with time.

Second (and most importantly), as noted by @phaazon, this goes against Kakoune's design principles. Introducing a dependency directly in core for a functionality that could be implemented externally. I do agree that there are some limitations at the moment, I have not looked at kak-tree-sitter but I suspect it is a far more complex codebase than what you did in that PR, I hope we can find a way to simplify how external plugins that rely on buffer content work.

PJungkamp:
### Motivation

I just spend some time with `typst` which is a language that suffers extraordinarily bad from the limitations of the regions highlighter. I tried to write kakoune highlighters that worked both for the markdown and code regions in `typst` but I ran into so many unfixable highlighting errors that I looked into other types of highlighting.

### `kak-tree-sitter`

The most promising approach that I saw for accurate highlighting is probably `tree-sitter`. I did check out `kak-tree-sitter`. I'm using it and love it! But it feels very much less responsive than what I'd expect from kakoune, especially when using a power-saving governor and platform profile on my laptop...

The inherent problem there is that it can't use the `ts_tree_edit` API efficiently to make reparsing of large buffers cheap instead reparsing the entire buffer sent over a pipe and it has to use a `kak -p` process to actually report the highlighted ranges back to kakoune.

I do have some ideas for the second problem, e.g. a `kak -P <session>` flag that in contrast to `kak -p <session>` keeps running and allows multiple commands to be passed to kakoune.

The first problem though lead me to write the code in this PR. I checked out `kak-tree-sitter`'s code and the tree-sitter project itself. This draft is just a POC of integrating `tree-sitter` with kakoune's internal structures. My main goal is to see where I'd have to introduce IPC interfaces to expose the necessary information to drive a more efficient tree-sitter plugin *without* introducing the tree-sitter dependency.

I wouldn't want to have this `tree-sitter` highlighter PR merged into master either. It does not fit kakounes goals. You can't even compile statically because of the dynamic loading of parsers.

### Some thoughts

My goal is to supply the [`ts_tree_edit`](https://tree-sitter.github.io/tree-sitter/using-parsers#editing) API with the changes in a buffer. The code in this PR can't do that, it reparses the entire buffer.

The `tree-sitter` library takes all positions on edits as *both* point (row & column) based and byte offset based coordinates. The kakoune `struct Buffer` is optimized for line-based editing and does only provide point based views of changes. See [`Buffer::changes_since`](https://github.com/mawww/kakoune/blob/4101e1814469d192b9fad9b75c86f764e743f8ac/src/buffer.hh#L222). I don't quite see a way to add efficient annotations of byte offsets to `struct Buffer`.

---

TLDR: I opened this PR because I was tinkering around and wanted to see who's interested and active on this topic.

phaazon:
About the partial updates / edit in place, I have [this still open](https://github.com/phaazon/kak-tree-sitter/issues/26) about that topic. It’s not something I have started working on because I want to stabilize the performance and features already (and I think it’s more important to have semantic text-objects first before going full optimizations), but clearly yes, it can have a negative impact on _“how fast you see highlighting”_. Also, the speed at which `kak -p` applies and blocks the editor is probably something that could be worked on completely independently of KTS or anything else.

Song-Tianxiang:
can we have something like vim's [text-properties](https://vimhelp.org/textprop.txt.html)

DarkArc:
So, I wanted to share some thoughts as a relatively new kakoune user that finds the kakoune design _very_ interesting.

> I think it’s an interesting approach, but in the same time, it bothers me a bit. Not because I’m the author of `kak-tree-sitter` and I spent a lot of time working on it. But because of the fact that I love the philosophy behind Kakoune, and I think you’re bending it. The reason for that is that Kakoune ships with zero dependencies. 

I'm not sure this is entirely fair. tree-sitter as it's used here is _not_ an external binary or tool. It is a third-party dependency, it is beyond the standard C & C++ libraries, as well as the POSIX APIs ... but it's a third-party dependency that's _very well_ regarded for solving a _very hard_ problem that kakoune already tries to solve as part of its core offering.

This seems like less of breaking the kakoune ideology (as I understand it) and more outsourcing a hard problem to the broader community (which arguably makes kakoune easier to maintain in the long term). The tree-sitter dependency isn't particularly onerous either as it itself is dependent only on the standard C library. So, it's not a _language_ or _platform_ level standard library/API but it's increasingly _the_ standard library for this domain.

This also seems like an area where pragmatism may be necessary. There _are_ *great* command line tools for finding files, listing files, source control, etc. However, that sort of tool just does not exist at an edit level.

Also as someone that works in a code with large C++ source files, I would be concerned about the performance implications of marshaling syntax highlighting annotations in a way that doesn't have a lot of overhead or have the potential for a litany of escape-sequence related bugs with arbitrary input.

A daemon is potentially viable (and that seems to be the community's taken approach), however a tree-sitter daemon is (while impressive!) _particularly exotic_ (AFAIK). This project is also written in Rust, which is a significant increase in the complexity (in terms of the size of the dependency chain) required to get this functionality vs directly integrating tree-sitter.

I've tried to look more into exactly how the kak-tree-sitter project works while writing this up, but unfortunately the project's website seems to be having issues that make browsing the source very difficult (so I've worked off the old GitHub sources). 

> First I do not want to have multiple, competing, built-in highlighting to maintain, and I do not want to solely rely on tree-sitter for highlighting. Having both will likely lead to one bitrotting with time.

I found this old comment when investigating the state of tree-sitter as it relates to kakoune: https://github.com/mawww/kakoune/issues/50#issuecomment-34294369

It _really_ seems like tree-sitter may be the answer to that research problem; I think it would be unfortunate to not integrate tree-sitter since the CST parser and parse rules are handled by a third party and the existing functionality provided via regex highlighters is upgraded/outclassed.

My thought on this would be (even if not via this specific PR's implementation):
- take tree-sitter as an optional (but recommended) dependency as the new standard syntax highlighting implementation
- drop the builtin regex highlighters registration rules from the mainline
  - make them a "community maintained best effort" for folks that can't use tree-sitter (... which really should be no-one given the low technical and legal requirements of tree-sitter)
  - folks that need these can just copy and paste the rules into their personal configuration files with no effort as the highlighting engine will still be there (those regex highlighters are still super useful for personal highlighting rules -- so I wouldn't expect the "core" code to really have much if any bitrot here)

That to me, seems like it would provide a nice compromise between ideology and pragmatism. Kakoune would still say _far far_ away from the "kitchen sink" that's emacs, but would pick up an easy to use, mature, and highlighly competent syntax highlighting engine.

Thanks for reading; I hope these are some helpful thoughts.

arrufat:
My two unsolicited two cents:

I honestly, don't see the point in tree-sitter as more and more LSP servers start adding semantic highlighting.
Whenever I try editors that use tree-sitter, they always get the syntax highlighting wrong in common places, albeit less often than regexes.

A classic example that I think it's unsolvable is, for example, in Zig you can do:

```zig
const debug = @import("std").debug;
const print = @import("std").debug.print;
```

Where `debug` is a struct/namespace and `print` is a function`. Only kakoune-lsp with zls gets this right, there's no way a regex can know that, and AFAIK, the same goes for tree-sitter. Maybe the main advantage is that we could all reuse the same tree-sitter definitions instead of each editor defining their own?

However, as more and more LSPs are adding semantic highlighting, what's the point of getting wrong highlighting, no matter how fast you go? I am surprised this is not brought up more often, but maybe I am missing something here.

DarkArc:
@arrufat 

Thanks for the read! So just to respond to this... I think language servers are really interesting, but they're ultimately different. Using a language server is the ideal because it understands the semantics of the language and can do "the best" that can be done.

However, language servers have _a lot_ of overhead associated with them (both technical and human) and there's wide variance in the quality of various language server projects. They can be hard to correctly configure in projects with abnormal build graphs.

Personally, my attempts to use clangd (attempted in emacs, Zed, and Helix) on the code base I work on professionally has been _very_ frustrating. When I worked on clang _itself_ I did get it working but the clangd overhead was just absurd (I forget specifics as it's been a few years) since clang is _such_ a large code base.

The CSTs used by TreeSitter are pretty much as close as you can get without having a true compiler running for semantic-aware highlighting (and as someone that works on compilers professionally ... those can be expensive). So, I think from a sort of "technology" stand point tree-sitter is "the" solution for something that can actually be reasonable integrated into the binary.

It's a good baseline balance of crowdsourced language parsing, performance, competency, and relatively low dependency overhead. It's a big step up from just using regex.

arrufat:
@DarkArc Yes, I entirely agree with you on clangd, whenever I work on large C++ projects, all my CPUs go to 100% when I modify heavily templated code. ZLS is really fast and lightweight, though. Also, go-pls and js/ts language servers work fine with semantic highlighting with no noticeable overhead. It's a shame Python doesn't have an open-source LSP with semantic highlighting, though.
In those cases, tree-sitter might be the better approach.

hadronized:
I agree with most of the point you made @arrufat, but also agree with @DarkArc. LSP can get in the way on some projects (for instance, at work, `rust-analyzer` takes _minutes_ to index correctly and give me anything useful), while `tree-sitter`, because it works at the buffer level, is instant.

It’s not unlikely that this problem gets fixed in the LSP protocol ~soon — at least from an engineering perspective, it makes sense to me to make `tree-sitter` a subpart of LSP, and let the trees live in the LSP’s state (so that you can have a tree parsed very quickly for an open buffer, and semantic highlighting applied above it once ready).

Take note that this last point is a pain point in Kakoune currently (highlighters do not have priorities), so we cannot make `kakoune-lsp`’s semantic highlighters and `kak-tree-sitter` highlighting work in harmony — I usually disable semantic tokens in LSP due to that.

If you think more about it, at some point, because LSPs are strictly more powerful than `tree-sitter`, it’s not unlikely that we do not need `tree-sitter` at some point. I’m just a bit concerned about scratch buffers / one-shot files that do not « belong to any project » and thus I’m not sure what an LSP will do / how it handles that.

PR #5102, opened 2024-02-17
Add terminal_info_inline_borders option

raiguard:
**Before:**
image

**After:**
image

### DIFF
--- a/src/terminal_ui.hh
+++ b/src/terminal_ui.hh
@@ -176,6 +176,7 @@ private:
     ColumnCount m_status_pos = 0;
     ColumnCount m_status_cursor_pos = 0;
     ColumnCount m_info_max_width = 0;
+    bool m_info_inline_borders = false;
 };
 
 }
--- a/doc/pages/options.asciidoc
+++ b/doc/pages/options.asciidoc
@@ -392,6 +392,10 @@ are exclusively available to built-in options.
             if *yes* or *true*, use native terminal cursor visibility control
             instead of Kakoune's cursor management (defaults to *false*)
 
+        *terminal_info_inline_borders*:::
+            if *yes* or *true*, borders will be drawn around info boxes with the
+            *inline* style.
+
 [[startup-info]]
 *startup_info_version* `int`::
     _default_ 0 +
--- a/src/main.cc
+++ b/src/main.cc
@@ -533,7 +533,8 @@ void register_options()
                        "    terminal_padding_char          codepoint\n"
                        "    terminal_padding_fill          bool\n"
                        "    terminal_cursor_native         bool\n"
-                       "    terminal_info_max_width        int\n",
+                       "    terminal_info_max_width        int\n"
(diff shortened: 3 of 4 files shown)


src/terminal_ui.hh
```diff
@@ -175,6 +175,7 @@ private:
 
     ColumnCount m_status_len = 0;
     ColumnCount m_info_max_width = 0;
+    bool m_info_inline_borders = false;
```
krobelus, line 178: I wonder if this should be the default except maybe for tiny info boxes that only span one line.

Also isn't it a bit weird that this is an option? I can imagine wanting two info boxes with different styles in the same window. For example `lsp-signature-help` is just one line so it shouldn't have borders while `lsp-hover` does want borders. So shouldn't it be a flag to `info`? Even more so because info already has a border by default. Only `info -style` doesn't

raiguard: The idea behind the option was to avoid changing the default behavior, because I assumed the default behavior was the way it was for a reason. As for single line boxes, I think it should still draw borders then. The whole point is that with my color scheme it is really difficult to distinguish what's in an info box and not because it doesn't use a super high contrast background.

mawww: This is a tricky problem, the reason i did not make the inline info bordered was to make it possible to align info with the anchor position. For example `ctags-enable-autoinfo` will display the info box aligned on the `(` opening the function call. Additionally, borders on small, inline info boxes hide a lot more of the buffer.

I initially thought a switch on the `info` command made most sense, but info box borders are a UI only concept, so adding this to the info command would move that knowledge in the UI agnostic part of Kakoune. I think we should find a switch/info style name that hints the UI it can use borders. That said, it seems the root of the issue is @raiguard colorscheme not distinguishing clearly info boxes background, maybe changing that would be a better fix.

mawww: Wouldn't it work to use a high enough color difference between Default and Information backgrounds so that its clear you are looking at an info box while still keeping Information background dark enough so that highlighted code looks okay ?

raiguard: Perhaps. But the reality is that having the border will _always_ make it easier to read. Instead of having characters directly next to each other only separated by color, it has a very clear and easy to parse separation.

raiguard: image
With `ctags-enable-autoinfo`

mawww: Yep, this autoinfo was trying to align its opening `(` to the one in the file (it is not anchored on the cursor), but cannot anymore because it would need to predict if a border is applied and how many characters it adds. Once you remove the ability to align it begs the question of why we have parenthesis in that info box in the first place...

On the other hand, I was envisioning the info boxes to be allowed to use a different font / font size on GUI implementations, which would also introduce the same issue breaking info alignement to buffer text.

I am still not fully sold on that change because it feels like allowing scripts to align info to buffer content is a nice feature, and I think the readability concern can be solved with the colorscheme.

Issue #2742, opened 2019-02-17
map global insert <c-c> does not work


vbauerster:
Following map does not work in insert mode:
```
map global insert '<c-c>' '<a-;>'
```
however this one works:
```
map global insert '<c-y>' '<a-;>'
```

andreyorst:
Basically, any key that can't be inserted via <kbd>Ctrl</kbd>+<kbd>v</kbd> <kbd>Key</kbd> can not be mapped.

vbauerster:
Is it limitation of ncurses? Though it won't help, but in neovim I was able to map `<c-c>` in insert mode.

andreyorst:
There were some discussions about handling inputs, you may find interesting these particular ones:
https://github.com/mawww/kakoune/issues/2554
https://github.com/mawww/kakoune/issues/2209

c-hild:
> Basically, any key that can't be inserted via Ctrl+v Key can not be mapped.

I don't understand the meaning of this, since for me `<c-v> <c-c>` yields `^C`.

andreyorst:
> > Basically, any key that can't be inserted via Ctrl+v Key can not be mapped.
> 
> I don't understand the meaning of this, since for me `<c-v> <c-c>` yields `^C`.

Where? In terminal - sure is, in Kakoune insert mode it is not. Well, at least for me.

PR #5107, opened 2024-02-26
Enable remapping of Ctrl-C and Ctrl-G

djpohly:
The interrupt <c-c> and cancel <c-g> keys [cannot be remapped](https://github.com/mawww/kakoune/issues/2742) because they are processed by the Kakoune client before reaching the input manager. (As noted [in a comment](https://github.com/mawww/kakoune/pull/4909#issuecomment-1571133026), processing them early is intentional.)

This PR adds two options, `interrupt_key` and `cancel_key`, which can be used to customize which keys are used for these purposes, freeing the defaults up to be mapped to another action using the `map` command.  This approach ends up requiring fewer changes (+27/-3) than the earlier work in #4909, and it keeps the processing of these keys early and uncomplicated.

(Note: I'm experienced with C but not C++, so feel free to educate me if I've written anything unidiomatic.)

### DIFF
--- a/doc/pages/keys.asciidoc
+++ b/doc/pages/keys.asciidoc
@@ -837,7 +837,9 @@ The following keys are recognized by this mode to help with editing
 These keys are used to cancel long-running operations, either inside
 Kakoune or outside it. Because they are intended as a safety mechanism
 when something goes wrong, these keys are handled very early on in
-Kakoune's input processing, and therefore cannot be remapped in any mode.
+Kakoune's input processing, so they can only be remapped by changing
+the `interrupt_key` and `cancel_key` options (see
+<<options#builtin-options,`:doc options builtin-options`>>).
 
 *<c-c>*::
     Stop any external processes. If you ever see Kakoune display a message
--- a/doc/pages/options.asciidoc
+++ b/doc/pages/options.asciidoc
@@ -317,6 +317,15 @@ are exclusively available to built-in options.
 *debug* `flags(hooks|shell|profile|keys|commands)`::
     dump various debug information in the '\*debug*' buffer
 
+*interrupt_key* `key`::
+    _default_ <c-c> +
+    key used to interrupt any running external processes
+
+*cancel_key* `key`::
+    _default_ <c-g> +
+    key used to cancel long-running Kakoune operations and clear the input
+    buffer
+
 *idle_timeout* `int`::
     _default_ 50 +
     timeout, in milliseconds, with no user input that will trigger the
--- a/doc/pages/options.asciidoc
+++ b/doc/pages/options.asciidoc
@@ -79,6 +79,11 @@ are exclusively available to built-in options.
     as a string but the set commands will complain if the entered text
     is not a valid regex
 
+*key*::
+    a single keypress using the same syntax as `map` (see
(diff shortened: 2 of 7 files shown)


doc/pages/options.asciidoc
```diff
@@ -79,6 +79,11 @@ are exclusively available to built-in options.
     as a string but the set commands will complain if the entered text
     is not a valid regex
 
+*key*::
+    a single keypress using the same syntax as `map` (see
+    <<mapping#mappable-keys,`:doc mapping mappable-keys`>>).  If
+    multiple keys are entered, only the first will be used.
```
krobelus, line 85: > If multiple keys are entered, only the first will be used.

I think it's better to throw  a `runtime_error{"too many keys"}`, then we can delete this line from docs.

src/client.cc
```diff
@@ -47,13 +47,14 @@ Client::Client(std::unique_ptr<UserInterface>&& ui,
     m_ui->set_ui_options(m_window->options()["ui_options"].get<UserInterface::Options>());
     m_ui->set_on_key([this](Key key) {
         kak_assert(key != Key::Invalid);
-        if (key == ctrl('c'))
+        auto opts = context().options();
```
krobelus, line 50: should be `auto&` else we copy the `OptionManager` every time.

src/client.cc
```diff
@@ -47,13 +47,14 @@ Client::Client(std::unique_ptr<UserInterface>&& ui,
...
+        if (key == opts["interrupt_key"].get<Key>())
```
krobelus, line 51: so this means that if a user really wants to, they can map it to different
keys in different modes with `ModeChange` hooks.
I wonder if we want to make this more ergonomic, I don't have a strong
opionion on this, this seems fine.
I guess `<c-g>` is only ever useful in normal or prompt mode, so it likely doesn't matter.

doc/pages/keys.asciidoc
```diff
@@ -837,7 +837,9 @@ The following keys are recognized by this mode to help with editing
 These keys are used to cancel long-running operations, either inside
 Kakoune or outside it. Because they are intended as a safety mechanism
 when something goes wrong, these keys are handled very early on in
-Kakoune's input processing, and therefore cannot be remapped in any mode.
+Kakoune's input processing, so they can only be remapped by changing
```
krobelus, line 840: So this allows us to maybe make commands like `map global normal <c-c> ...`
throw an error if `cancel_key=<c-c>`, and point to that option.

Not sure if we want to do this; it would be a bit more user friendly on one
hand but also enforce an ordering of map/set-option commands in kakrc which
would be weird.

doc/pages/options.asciidoc
djpohly: I changed it to require exactly 1, following the example of codepoint.  However, I just realized that does prevent determined users from disabling it altogether if they want.  Should `set global interrupt_key ''` be allowed, with a caveat of "be it upon your own head"?

doc/pages/keys.asciidoc
djpohly: Yeah, this sounds like it would introduce a little more complexity on both the user's and Kakoune's side.  I do wonder if maybe the pointer about "warning, these two behave differently" belongs near the top of `:doc mapping` rather than at the very bottom of `:doc keys`?

PR #5109, opened 2024-03-01
Change `kakrc` source path to the correct path in Makefile

natsukagami:
It was failing my builds :)


natsukagami:
Nvm, #5108 fixed it

PR #5111, opened 2024-03-04
Makefile: allow PREFIX to be overridden. Example: `PREFIX=~/.local make install`

sidkshatriya:
Makefile: allow PREFIX to be overridden. Example: `PREFIX=~/.local make install`

It was always possible to install kakoune with custom prefix. This seems to be a regression due to recent changes to the Makefile. This commit should hopefully restore old behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ LDFLAGS-static-yes = -static -pthread
 
 version != cat .version 2>/dev/null || git describe --tags HEAD 2>/dev/null || echo unknown
 
-PREFIX = /usr/local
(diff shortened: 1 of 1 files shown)


sidkshatriya:
Thanks for the tip ! useful ! I leave it to others to decide if this PR should be closed / merged.

mawww:
Unfortunately `VAR ?= value` does not seem to be part of POSIX make, which is what the recent changes to the Makefile were about. So yeah, PREFIX should be passed after the make command.

PR #5123, opened 2024-03-20
Update README to match new `make install` command

jbrains:
It appears that the command to specify the `PREFIX` variable to the makefile has changed.

I had to run `make PREFIX=$HOME/.local install` in order to override the hardcoded `PREFIX` value in the makefile. I infer that the instructions need to be changed for those who don't know. (TIL. ;) )


mawww:
Thanks !

PR #5125, opened 2024-03-22
Language Support - Arturo

g3rwy:
Pretty basic support for [arturo](https://arturo-lang.io/) programming language syntax.
I also got a plugin with the same file already uploaded on [github](https://github.com/g3rwy/kakoune-arturo) if someone will want to use it before PR gets merged. After it gets merged i will archive the repo and change the script to nothing so anyone using plug.kak with updated Kakoune wont have a conflict.

### DIFF
--- a/rc/filetype/arturo.kak
+++ b/rc/filetype/arturo.kak
@@ -1,0 +1,103 @@
+# https://arturo-lang.io/documentation/language
+#
+
+# Detection
+# ‾‾‾‾‾‾‾‾‾
+
+hook global BufCreate .*\.art? %{
+    set-option buffer filetype arturo
+}
+
+# Initialization
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+hook global WinSetOption filetype=arturo %{
+    require-module arturo
+    set-option window static_words %opt{arturo_static_words}
+
+    hook window InsertChar \n -group arturo-insert arturo-insert-on-new-line
+
+    hook -once -always window WinSetOption filetype=.* %{ remove-hooks window arturo.+ }
+}
+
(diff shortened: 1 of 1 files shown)


mawww:
I am a bit conflicted about this change, I am not sure I want to have every language under the sun in the main repository, Arturo seems like a very niche language and I wonder if it makes sense to merge this.

g3rwy:
Well kakoune still includes language support for Ragel language which for me seems a bit niche and last development build of it was 3 years ago. Probably there are more, but i also can understand if you want to keep the editor not be loaded with these. :)
I will leave it as is but sure you can close it, the plugin for it is on my repo too so it will be available.

mawww:
Indeed I have not been very consistent with this, it would be good to define some kind of policy so that I dont have to make that decision with whatever the feeling of the moment is, but I have not taken the time to do that yet. As I said I am conflicted on this,  but if I were to merge this PR it would need a copyright waiver commit from you (see CONTRIBUTING) as I do not think I have one already.

eduardvercaemer:
@mawww 

Inevitably for things that won't make it into the main repository: Is there a way of centralizing contributions like this ? So things that may be helpful for someone can be searched for in a single place and easily added to an individual installation ?

I'm interested in contributing ADA language support and I was wondering if that is something that would be merged here or somewhere else (:

g3rwy:
@eduardvercaemer Well you know, like i mentioned i had a repository made for my plugin. Which can easily get added to kakoune's plugins list, so if you make support for Ada, you could make a repository for it and add it to [Kakoune's plugins list](https://kakoune.org/plugins.html)

mawww:
@eduardvercaemer Ada is definitiely a language that deserves to be in core, its a mainstream language in my book.

PR #5131, opened 2024-03-26
Expose %val{register} and %val{count} to the modeline

topisani:
this opens up for a lot more customization of the modeline, as all information in {{mode_info}} can now be obtained individually.

See for instance https://github.com/Hjagu09/yummy.kak for a usecase.

### DIFF
--- a/src/client.cc
+++ b/src/client.cc
@@ -5,6 +5,7 @@
 #include "buffer_manager.hh"
 #include "buffer_utils.hh"
 #include "file.hh"
+#include "normal.hh"
 #include "remote.hh"
 #include "option.hh"
 #include "option_types.hh"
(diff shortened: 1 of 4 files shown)


src/client.cc
```diff
@@ -169,7 +170,12 @@ DisplayLine Client::generate_mode_line() const
         HashMap<String, DisplayLine> atoms{{ "mode_info", context().client().input_handler().mode_line() },
                                            { "context_info", {generate_context_info(context()),
                                                               context().faces()["Information"]}}};
-        auto expanded = expand(modelinefmt, context(), ShellContext{},
+        const NormalParams* normal_params = context().client().input_handler().get_normal_params();
+        ShellContext shell_context = { {}, {
+        {"register", normal_params ? String(normal_params->reg) : String("") }, 
+        {"count", normal_params ? String(to_string(normal_params->count)) : ""}}};
```
mawww, line 176: I would prefer to not have $kak_register and $kak_count defined if normal_params is nullptr, so more `auto shell_context = normal_params ? ShellContext{ ... } : ShellContext{};`

src/input_handler.cc
```diff
@@ -54,6 +54,8 @@ class InputMode : public RefCountable
         auto coord = context().window().display_position(cursor).value_or(DisplayCoord{});
         return {CursorMode::Buffer, coord};
     }
+    
+    virtual const NormalParams* get_normal_params() const { return nullptr; }
```
mawww, line 58: This feels quite hackish.

mawww (commented):
Started to comment on this, but (as often) I was not sure what approach I would prefer, which led me to basically implement this in another way I am a bit more happy about, left the comments I had in case they are useful.

src/client.cc
topisani: The reason why I wanted to avoid this, is that it means you can only use it through a shell invocation, and not directly in the modeline, as it will give an invalid expansion error. Having the expansion always available, but empty, feels more user friendly to me than having a modelinefmt that errors in some modes

mawww: Good point, fixed that.

PR #5132, opened 2024-03-27
Add -indent option to show-whitespace highlighter

### LINKED ISSUE #2323: Indentation guides
Hello, I'm continuing exploring Kakoune, and I'd like to ask about it's ncurses TUI capabilities.

Is it possible to have indentation guides, like in modern GUI editors like Sublime, VSCode, Atom, Notepad++, and others? 

In short, Indentation guides is a GUI editors feature, which is highly ignored by most of TUI editor developers, since it is hard to have thin vertical lines, because each will occupy cell, and terminals are not support drawing vertical lines as a cell parameter.

I use Kakoune with `add-highlighter global show_whitespaces -tab "▏" -lf " " -nbsp "⋅" -spc " "`, and used Vim (can't avoid comparison here, since Vim has similar problem since created) for many years with similar setings, and never actually had problems with cell occupation, since these cells are always contain whitespaces:
image
(Kakoune | Neovim | VSCode)

You can see that VSCode has one line colored more bright, it shows what area my cursor currently in. That's one more useful feature despite being able to see indentation guides. With this tight code placement it isn't clear how Indentation guides actually can help, let's look at another example, which also contains empty lines between lines of code:
image
(Kakoune | Neovim | VSCode)

While it's not a problem to see what area cursor is, it is little more ambiguous where does `while` scope starts and end. VSCode shows it with highlighted indentation guide, since cursor is inside of it. (And thankfully this file is formatted *properly* which isn't a common practice AFAICS, since I've seen too much unformatted code for like past 6 years. And when it comes to unformatted code, Indentation guides sometimes are time-savers.)
Deep indented code (which isn't a great practice, I know, but not all code we work with is great) will also benefit from indentation guides, since it is easy to find where scope of current block ends. Yes, I know about `m` motion, however moving to the scope start and then to the end, is slower than just fast look with your eyes over the indentation guide.

The problem with last screenshot (and the main problem with this method), is that not only indentations guides are not drawn at empty lines (since it is just whitespaces with custom highlighting), but that if file is indented with spaces, or mixed spaces, and tabs (like there) you will not see any guides at all, because spaces are not tabs and therefore no vertical bars with `show_whitespaces`. Emulating indentation guides with Kakoune's `show_whitespaces` or Vim's `listchars` isn't great by itself, since I'd like to set tabs to `→` for example, so I could see when indentation is done with spaces, while still being able to see guides. Using code formatters isn't an option, since I often fix other people's code, and wouldn't like to mess with them after committing their code to the repo fully reformatted, which will not please them, since they have [their own strong view of proper formatting](https://user-images.githubusercontent.com/19470159/44397616-a2f59800-a549-11e8-9198-46ce15f5bb18.png) (we don't have style guides, and tons of legacy code, that was written in spaghetti style, but that's another story).

Since ncurses can draw pretty everything we want, can't we have indentation guides either as vertical bars of changed background color, so it won't occupy cell, or as `▏` (or any other pipechar) since indentation guides never actually drawed on non-whitespace area, even in GUI editors.

I've already opened similar issue in neovim: https://github.com/neovim/neovim/issues/7544 and proposed using a folds api, for providing information about indentation guides, since this is really the same thing. AFAICS there's no folds in Kakoune yet https://github.com/mawww/kakoune/issues/453, and knowing that most GUI editors I've mentioned draw indentation guides based on indentwith setting, I think that continious, scope aware guides are possible without fold api.

topisani:
image

A couple of semi-opinionated choices were made in this implementation:

 1. The guide is hidden in the first column.
 2. The indent guides are highlighted using a new `WhitespaceIndent` face.
 3. Nothing is done to continue the guide through empty lines. I believe this to be the correct approach, at least as long as it is kept as a part of the show-whitespaces highlighter. However some people's oppinion may differ, and if so, that could be implemented.
 4. The guides default to on, like the other show-whitespace options. Default character is "│".
 5. Spaces between the indent guides are currently highlighted as other spaces. Other reasonable options would be no replacement, -tabpad, or a similar -indentpad.
 6. Guides are disabled by passing `-indent ""`.
 7. Indent guides are separate from tab highlighting.

Additionally, we could consider adding a separate face for the "current" indent level as many editors do, but this is a bit harder in kakoune because of multiple selections.

Closes #2323

### DIFF
--- a/doc/pages/highlighters.asciidoc
+++ b/doc/pages/highlighters.asciidoc
@@ -55,6 +55,10 @@ highlighter is replaced with the new one.
     *-tabpad* <separator>:::
         a one character long separator that will be appended to tabulations to honor the *tabstop* option
 
+    *-indent* <separator>:::
+        a one character long separator that will replace the first space in indentation according to the *indentwidth* option
+        This will use the `WhitespaceIndent` face.
+
     *-only-trailing*:::
         only highlight whitespaces at the end of the line
 
--- a/src/face_registry.cc
+++ b/src/face_registry.cc
@@ -204,6 +204,7 @@ FaceRegistry::FaceRegistry()
         { "MatchingChar", {Face{ Color::Default, Color::Default, Attribute::Bold }} },
         { "BufferPadding", {Face{ Color::Blue, Color::Default }} },
         { "Whitespace", {Face{ Color::Default, Color::Default, Attribute::FinalFg }} },
+        { "WhitespaceIndent", {Face{}, "Whitespace"} },
       }
 {}
 
--- a/src/highlighters.cc
+++ b/src/highlighters.cc
@@ -954,7 +954,7 @@ struct TabulationHighlighter : Highlighter
 };
 
 const HighlighterDesc show_whitespace_desc = {
-    "Parameters: [-tab <separator>] [-tabpad <separator>] [-lf <separator>] [-spc <separator>] [-nbsp <separator>]\n"
+    "Parameters: [-tab <separator>] [-tabpad <separator>] [-lf <separator>] [-spc <separator>] [-nbsp <separator>] [-indent <separator>]\n"
     "Display whitespaces using symbols",
     { {
         { "tab",    { ArgCompleter{}, "replace tabulations with the given character" } },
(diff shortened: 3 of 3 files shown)


Screwtapello:
In a file where indentation is done with tabs, does the tab-highlighting or ident-highlighting win? What happens if `tabstop` and `indentwidth` are set to different values?

topisani:
> In a file where indentation is done with tabs, does the tab-highlighting or ident-highlighting win? What happens if `tabstop` and `indentwidth` are set to different values?

Indhent guides currently only highlight leading spaces, according to indentwidth. So tab highlighting and indent highlighting never interact. If you want similar indent guides for tabs, they were always, and are still, possible using the tab highlighter. We could consider using this logic for both preceding tabs and spaces though, but as I see it, this is the most useful design, to still easily be able to differentiate tabs and spaces

mawww:
Nice, my only comment is whether we need a separate `WhitespaceIndent` face, I suspect it not really necessary. I'll merge it as-is for now but might remove that face later on if I cant find a good justification for it.

PR #5133, opened 2024-03-27
Session/Client Renamed hooks

topisani:
These are needed to not break systems like kakoune-lsp when the server or client are renamed. Other usecases exist, like using client names to implement functionality such as `jumpclient`


krobelus:
will be great to finally fix session-renaming in kakoune-lsp.
It's somewhat at odds with allowing to use the same `kak-lsp` server across multiple kak sessions but I don't think anyone uses that.

The `ClientRenamed` hook begs the question whether option like `toolsclient` should have a special `client-name` type that is automatically updated on rename. Perhaps even set to empty when the associated client is deleted (though I don't think we need that because `eval -try-client` already has the right fallback in this case).

Issue #5147, opened 2024-04-18
[BUG] Python filetype does not allow paragraphs in block comments.


sjjf:
### Version of Kakoune

v2023.08.05-290-ge9bd7083

### Reproducer

Enter the following in a filetype=python buffer:

```python
# this is a paragraph
#
```

Go to the end of the last line, and hit enter.

### Outcome

The preceding empty comment line is deleted, giving:

```python
# this is a paragraph

```

### Expectations

The preceding empty comment should be retained, to allow for creating a paragraph break in the block comment. This would give:

```python
# this is a paragraph
#

```

### Additional information

As per pep8, block comments in Python should support paragraphs, with paragraph separation being provided by an empty comment line (see https://peps.python.org/pep-0008/#block-comments).

I'm guessing the intent of the current implementation is to allow exiting a block comment without having to explicitly backspace to clear the comment prefix, which I think is perfectly reasonable. The issue is that the current implementation means you have to jump through extra manual hoops in order to get paragraphs in your block comments.

The block comments section of pep8 isn't explicit about what to do with more than one empty comment line in block comments, the exact text is:

> Paragraphs inside a block comment are separated by a line containing a single #.

I think it's reasonable to interpret this as meaning that a single empty comment needs to be supported, but that more than one isn't - that would mean you could adjust the `python-insert-on-new-line` command to treat two consecutive empty comments the way it currently treats one (exiting the block comment), while leaving a single empty comment as-is to allow a paragraph break. I'm getting a pull request implementing this change ready.

sjjf:
Thanks for merging my proposed fix.

Issue #860, opened 2016-10-10
Incorrect commenting of new line


bhajneet:
```
# comment
if haha
```

* alt+j to join on comment line
* set selection to before if
* insert enter/return

new hashtag added after if and before "haha".

```
# comment
if#  haha
```

mawww:
In what filetype does that happens ?

bhajneet:
Py (Python)

PR #5148, opened 2024-04-18
Support paragraph breaks in python block comments.

sjjf:
The current python filetype module treats a single empty comment line (typically created by hitting enter twice while in a block comment) as the end of the block comment, deleting the empty comment line and ending comment prefix copying. This runs contrary to PEP8, which explicitly allows for paragraphs in block comments, with an empty comment as the paragraph separator (see https://peps.python.org/pep-0008/#block-comments for the actual text).

This change implements support for using a single empty comment as a paragraph separator, with two consecutive empty comments being treated as the end of the block comment; both empty comment lines are deleted and comment prefix copying is ended.

This would close issue #5147 

This change basically means that you can hit enter twice while you're in a block comment, and you'll get an empty comment and the cursor at the start of a new commented line - if you enter text the block comment will continue as normal, if you hit enter again at this point the block comment is terminated, the empty comments are deleted, and you go back to regular code. This follows the basic logic of the current implementation, while allowing for paragraph breaks in block comments.

This behaviour arguably isn't ideal, particularly with the two empty comments disappearing suddenly - that rather goes against the principle of least surprise. But given the way block comments work it's hard to see a better approach - something like Ctrl-enter, perhaps? I don't know how that would work with a modal editor like Kakoune, though.

Possible issues with the whole idea aside, I'm not confident this is the best possible implementation - it was a lot fiddlier than I anticipated to make it work the way I intended, and it doesn't feel very elegant. There may be cases that aren't handled properly, though I've tried to make it reasonably robust. Any suggestions for improving the code would be appreciated.

I've added tests in the `regression/5147-python-block-comment-paragraphs` directory, if there's a more appropriate place for them please let me know and I'll relocate them. Additional test cases or a better way to implement them are welcome.

I've already submitted a copyright waiver for previous contributions, I assume that's still considered valid.


sjjf:
Fixed regression on issue #860 - I forgot to run the full test suite before pushing the changes, only the tests I'd added . . .

PR #5172, opened 2024-05-10
Add builds.sr.ht alpine latest configuration

stacyharper:
This might help detecting issues with alpine builds


mawww:
The issue with this change is that we already have the maximum build configs in .build (4) which means if we add one then the builds that are run will be randomized, but we have the readme.yml build that we would like to run on every update and I am not sure how to get sr.ht to always run that one with more than 4 builds.

PR #5180, opened 2024-05-19
doc: clarify that C and <a-C> act column-wise 

dontlaugh:


dontlaugh:
This would have helped me, but feel free to close if it doesn't fit your style.

arrufat:
Not a native English speaker, but `column-wise` sounds more natural to me.

arachsys:
Native speaker: I agree with @arrufat that "column-wise" is clear. (Also "mnemonic" doesn't have a 'u' in it and probably isn't the right word. "Clarify that C and <a-C> act column-wise" might describe the content of the patch better?)

PR #5185, opened 2024-06-06
Add EnterDirectory hook

PJungkamp:
We currently don't have any way to detect a change to the current working directory.

I like to base my kakoune session name on my working directory and I'd like for it to be able to change when I use `change-directory` in kakoune... This doesn't currently seem to be possible.

I've started considering this after the recently added `SessionRename` and `ClientRename` hooks that allowed more complex plugins to tolerate name changes.


src/commands.cc
```diff
@@ -2600,13 +2600,14 @@ const CommandDesc change_directory_cmd = {
...
         }),
-    [](const ParametersParser& parser, Context&, const ShellContext&)
+    [](const ParametersParser& parser, Context& ctx, const ShellContext&)
     {
         StringView target = parser.positional_count() == 1 ? StringView{parser[0]} : "~";
         if (chdir(parse_filename(target).c_str()) != 0)
             throw runtime_error(format("unable to change to directory: '{}'", target));
         for (auto& buffer : BufferManager::instance())
             buffer->update_display_name();
+        ctx.hooks().run_hook(Hook::DirectoryChanged, "", ctx);
```
mawww, line 2610: I think we should pass the result of `parse_filename(target)` as the filter string, so that hooks can be added to trigger on specific directories.

mawww (changes requested):

mawww:
`EnterDirectory <absolute path>` sounds good and yes, it should be emitted on startup as well (either after or just before `KakBegin`). I dont think we need a separate `ChangeDirectory` hook, just re-exec `EnterDirectory` whenever we change the current directory.

PJungkamp:
I reworked it to `EnterDirectory` with the new `real_path` as the hook parameter.

---

Here's a little snippet to use the current working directory as the session name or, if it's part of a git tree, use the git directories name. I don't know whether I'll use it in my actual configuration, as the `rename-session` is executed *after* `kakrc` has been sourced. Plugins relying on a stable session name will break (most plugins that spawn background processes, e.g. kak-lsp and kak-tree-sitter).

```kak
hook global EnterDirectory .*/(?<name>[^/]*)/? %{ evaluate-commands %sh{
    if toplevel="$(git rev-parse --show-toplevel 2> /dev/null)" && [ "$toplevel" != / ]; then
        name="$(basename "$toplevel")"
    else
        name="$kak_hook_param_capture_name"
    fi

    if [ -n "$name" ] && [ "$name" != "$kak_session" ]; then
        printf "%s\n" "rename-session %/$name/"
    fi
}}
```

mawww:
Thanks, the last missing piece is a copyright waiver commit from you as I do not think I have one from you yet. See the CONTRIBUTING file.

PJungkamp:
> Thanks, the last missing piece is a copyright waiver commit from you as I do not think I have one from you yet. See the CONTRIBUTING file.

Done!

PR #5186, opened 2024-06-08
Increase SSO from 22 to 23 chars.

Icantjuddle:
This change gives an extra character of SSO space.

The core of the change is to make the short-string indicator byte be used for remaining space instead of current size when in short string mode. When there is 0 remaining space, the indicator byte will be 0 which is also a null terminator. Notably this optimization was popularized by Andrei Alexandrescu in fbstring.

### DIFF
--- a/src/string.hh
+++ b/src/string.hh
@@ -1,14 +1,13 @@
 #ifndef string_hh_INCLUDED
 #define string_hh_INCLUDED
 
-#include <bit>
-#include <cstring>
+#include <climits>
+#include <cstddef>
 #include "memory.hh"
 #include "hash.hh"
 #include "units.hh"
 #include "utf8.hh"
 
-#include <climits>
 
 namespace Kakoune
 {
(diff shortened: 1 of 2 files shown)


src/string.hh
```diff
@@ -195,18 +197,37 @@ public:
         struct Long
         {
             static constexpr size_t max_capacity =
-                (size_t)1 << 8 * (sizeof(size_t) - 1);
+                ((size_t)1 << (CHAR_BIT * (sizeof(size_t) - 1))) - 1;
 
             char* ptr;
             size_t size;
-            size_t capacity;
+            unsigned char m_capacity[sizeof(size_t) - 1];
```
mawww, line 204: Wouldn't a bitfield work here ? `size_t m_capacity : CHAR_BIT * (sizeof(size_t) -1);`

Icantjuddle: Perhaps. I'm not the most familiar with c++'s bitfield semantics, I was not sure if semantics for assigning to an unsigned bitfield from an out-of-range type was well defined. From https://en.cppreference.com/w/cpp/language/bit_field#Notes it seemed that for signed bitfields it is not (nothing there on unsigned). 

When I checked on godbolt it seems that the `memcpy`s are compiled down to bit ops which I imagine is the same for a bitfield implementation.

Happy to switch to bitfield if it would work here.

mawww: I think using a bitfield would remove the need for methods to update/read m_capacity which would be nicer, and I'd expect essentially the same codegen. That would be my preference if it works fine.

PR #5190, opened 2024-06-17
editorconfig: allow to disable automatic edition

stacyharper:


mawww:
I am not sure this option is necessary, you can achieve the same thing by disabling the hooks themselves with `set global disabled_hooks autowrap|editorconfig-hooks` which does not seem much worse than an editorconfig specific option.

stacyharper:
I think I could go leaving the `trim_trailing_whitespace`, because it is a less common option, and because the BufWritePre hook is less intrusive imo.

But enabling auto-wrapping is a problem to me (maybe because I think this module is poorly designed atm, and behave badly most of the time). This should be up to the user to enable this with filetypes they want.

`set global disabled_hooks autowrap` is a valid work-around, but then the user have to configure each filetypes themselves, which go against the goal of editorconfig.

(In a general manner, I would like for Kakoune to not edit files by itself. But I understand your opinion on this might diverge. Autowrapping, whitespace cleanup, or any automatic edition, can cause lot of vcs conflicts on projects we work for.)

PR #5191, opened 2024-06-17
autowrap: automatically set autowrap column highlighter

stacyharper:
Also the user now can change the value, without duplicating the column.

~~is based on https://github.com/mawww/kakoune/pull/5190~~


stacyharper:
Maybe this is a good idea to change the default value to 0?

stacyharper:
I rebased this one, to not depend on the other editorconfig git history.

mawww:
I wonder about this change, I am not sure most users would want enabling autowrap to automatically add a column highlighter, I wonder if those hooks should be wrapped into a command a user could explicitely call to opt-into this behaviour.

stacyharper:
We currently do this with editorconfig, maybe we should dedicate an option for this instead.

stacyharper:
closed in favor of: https://github.com/mawww/kakoune/pull/5526

PR #5193, opened 2024-06-19
introduce jump_pattern option

QiBaobin:
Sometime, we don't have line and column information in context, like fd output, log etc. 
Let's extract an option `jump_pattern` to let other use cases can override it.

### DIFF
--- a/rc/tools/jump.kak
+++ b/rc/tools/jump.kak
@@ -2,6 +2,8 @@ declare-option -docstring "name of the client in which all source code jumps wil
     str jumpclient
 declare-option -docstring "name of the client in which utilities display information" \
     str toolsclient
+declare-option -docstring "the pattern for the jump information in the line, such as file:line:col" \
+    str jump_pattern "^([^:\n]+)(?::(\d+))?(?::(\d+))?"
 
 provide-module jump %{
 
@@ -11,10 +13,7 @@ define-command -hidden jump %{
     evaluate-commands -save-regs a %{ # use evaluate-commands to ensure jumps are collapsed
         try %{
             evaluate-commands -draft %{
-                # file
-                # file:line
-                # file:line:col
-                execute-keys ',xs^([^:\n]+)(?::(\d+))?(?::(\d+))?<ret>'
+                execute-keys ",xs%opt{jump_pattern}<ret>"
                 set-register a %reg{1} %reg{2} %reg{3}
             }
             set-option buffer jump_current_line %val{cursor_line}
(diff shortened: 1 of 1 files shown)


rc/tools/jump.kak
```diff
@@ -11,7 +11,10 @@ define-command -hidden jump %{
     evaluate-commands -save-regs a %{ # use evaluate-commands to ensure jumps are collapsed
         try %{
             evaluate-commands -draft %{
-                execute-keys ',xs^([^:\n]+):(\d+):(\d+)?<ret>'
+                # file
+                # file:line
+                # file:line:col
```
krobelus, line 16: to be clear, `file:line:col` and `file:line` are already supported.
Whether we want to support `file`, maybe, I'm not sure.
Do you have a use case in mind?
Note that it should be possible to create a buffer-local `jump` alias to your preferred implementation.
The LSP plugin uses this technique for example.
Also, if you want to support `file`, you probably want to do the same to `jump-select-next` and `jump-select-previous` (also those can be customized by means of aliases).

QiBaobin: We only support `file:line:` currently, it doesn't recognize the pattern without last `:`.

One of my use case is that finding some files using `fd` or `rg -l`, look at them and open some of them. I don't know if there is a better workflow.
```kakrc
define-command -params 1.. -docstring "find files" fd %{
    require-module fifo
    evaluate-commands -try-client %opt{toolsclient} %{
        fifo -name "*fd*" fd -tf %arg{@}
        require-module jump
        hook buffer NormalKey <ret> jump
    }
}
```

QiBaobin: @krobelus , I reverted the pattern and introduced an option instead, could you take another look?

krobelus: > One of my use case is that finding some files using fd or rg -l,
> look at them and open some of them

Interesting. I personally don't use jump for that.
Instead I use this command:.

```kak
define-command -override my-goto-file -docstring "jump to file:line:col at cursor" %{
	evaluate-commands -save-regs ^ %{
		try %{
			execute-keys -save-regs '' Z
			# Try to also match filenames containing ':' here.
			execute-keys %{<semicolon><a-a><a-w>1s(?:cargo:warning=)?([^\s]+)(?::(\d+))(?::(\d+)\b(?![.]))<ret>)<space><esc>,<esc>}
		} catch %{
			execute-keys -save-regs '' z
			execute-keys %{<semicolon><a-a><a-w>1s(?:cargo:warning=)?([^:\s]+)(?::(\d+))?(?::(\d+)\b(?![.]))?<ret>)<space><esc>,<esc>}
		}
	}
	execute-keys -with-hooks "gf%reg{2}g<a-h>%reg{3}lh"
}
```

which might even be a preferable solution to most users because users shouldn't need to know that `jump.kak` exists, but they should know commands like `gf`.

It's interesting that `jump-next` and `jump-previous` can benefit from this too; maybe there is benefit to providing this option.
However overall I'm not so sure if it's worth it (would anyone use jump-next/jump-previous in this scenario?).
I think I'd rather encourage power users to create their buffer-local aliases for jump (though you could argue that that's a hack). I guess I'm biased here. I'm not sure..

QiBaobin: You make  a good point, let's use `gf` instead, thanks for the review.

PR #5199, opened 2024-07-03
add default terminal implementation

QiBaobin:
it seems that most of terminal support `-e` option, let use it as a fallback implementation so that we can use `terminal` and `new` on other undefined terminal like alacritty, Terminal.app.

### DIFF
--- a/rc/windowing/detection.kak
+++ b/rc/windowing/detection.kak
@@ -23,7 +23,7 @@ declare-option -docstring \
 "Ordered list of windowing modules to try and load. An empty list disables
 both automatic module loading and environment detection, enabling complete
 manual control of the module loading." \
-str-list windowing_modules 'tmux' 'screen' 'zellij' 'kitty' 'iterm' 'appleterminal' 'sway' 'wayland' 'x11' 'wezterm'
+str-list windowing_modules 'tmux' 'screen' 'zellij' 'kitty' 'iterm' 'appleterminal' 'sway' 'wayland' 'x11' 'wezterm' 'envterm'
 
 declare-option -docstring %{
     windowing module to use in the 'terminal' command
--- a/rc/windowing/envterm.kak
+++ b/rc/windowing/envterm.kak
@@ -1,0 +1,16 @@
+provide-module envterm %{
+
+# ensure that we're running on iTerm
+evaluate-commands %sh{
+    [ -z "${kak_opt_windowing_modules}" ] || [ -z "$kak_env_TERM" ] || echo 'fail $TERM not found'
+}
+
+define-command envterm-terminal-window -params 1.. -docstring '
+envterm-terminal-window <program> [<arguments>]: create a new terminal as an $TERM window
+The program passed as argument will be executed in the new terminal'\
+%{
+    nop %sh{ "$TERM" -e "$@" > /dev/null 2>&1 & }
+}
+
+}
+
--- a/rc/windowing/envterm.kak
+++ b/rc/windowing/envterm.kak
@@ -2,7 +2,7 @@ provide-module envterm %{
 
 # ensure that we're running on iTerm
 evaluate-commands %sh{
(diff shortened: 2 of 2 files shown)


mawww:
I am not sure how much `$TERM` can be used to find the actual binary of the current terminal, the TERM env var is used for terminfo resolution and common values that come to mind will not resolve correctly (say `xterm-256color` which would not find `xterm` or `rxvt-unicode` which would not match `urxvt`)

QiBaobin:
> I am not sure how much `$TERM` can be used to find the actual binary of the current terminal, the TERM env var is used for terminfo resolution and common values that come to mind will not resolve correctly (say `xterm-256color` which would not find `xterm` or `rxvt-unicode` which would not match `urxvt`)

That's a good point, I made another change to ensure we have '$TERM' exists using `which`, could you take another look?

Screwtapello:
Do alacritty or Terminal.app even set $TERM to something that could be used to launch them?

My understanding is that most terminals set $TERM to "xterm" or "xterm-256color" for compatibility reasons, regardless of what the terminal's actual name is.

QiBaobin:
> Do alacritty or Terminal.app even set $TERM to something that could be used to launch them?
> 
> My understanding is that most terminals set $TERM to "xterm" or "xterm-256color" for compatibility reasons, regardless of what the terminal's actual name is.

I tested, alacritty works as it set TERM to 'alacritty', and some others also works based on http://jdebp.uk./Softwares/nosh/guide/TERM.html , of course there are still a lot of others don't work.

In this patch, I added this to the last to only give it a try if none of others doesn't work.

QiBaobin:
@mawww , @Screwtapello , do you think using `ps` to get parent of parent of the kak client is a better idea?

```
ps -p %val{client_pid} -o args,ppid # to get the shell pid
ps -p shell_pid -o args # to get the terminal path
```

QiBaobin:
don't figure out a good way, let's close it for now

QiBaobin:
here is the logic from neofetch, it looks like complex.

```
get_term() {
    # If function was run, stop here.
    ((term_run == 1)) && return

    # Workaround for macOS systems that
    # don't support the block below.
    case $TERM_PROGRAM in
        "iTerm.app")    term="iTerm2" ;;
        "Terminal.app") term="Apple Terminal" ;;
        "Hyper")        term="HyperTerm" ;;
        *)              term="${TERM_PROGRAM/\.app}" ;;
    esac

    # Most likely TosWin2 on FreeMiNT - quick check
    [[ "$TERM" == "tw52" || "$TERM" == "tw100" ]] && term="TosWin2"
    [[ "$SSH_CONNECTION" ]] && term="$SSH_TTY"
    [[ "$WT_SESSION" ]]     && term="Windows Terminal"

    # Check $PPID for terminal emulator.
    while [[ -z "$term" ]]; do
        parent="$(get_ppid "$parent")"
        [[ -z "$parent" ]] && break
        name="$(get_process_name "$parent")"

        case ${name// } in
            "${SHELL/*\/}"|*"sh"|"screen"|"su"*|"newgrp") ;;

            "login"*|*"Login"*|"init"|"(init)")
                term="$(tty)"
            ;;

            "ruby"|"1"|"tmux"*|"systemd"|"sshd"*|"python"*|\
            "USER"*"PID"*|"kdeinit"*|"launchd"*|"bwrap")
                break
            ;;

            "gnome-terminal-") term="gnome-terminal" ;;
            "urxvtd")          term="urxvt" ;;
            *"nvim")           term="Neovim Terminal" ;;
            *"NeoVimServer"*)  term="VimR Terminal" ;;

            *)
                # Fix issues with long process names on Linux.
                [[ $os == Linux ]] && term=$(realpath "/proc/$parent/exe")

                term="${name##*/}"

                # Fix wrapper names in Nix.
                [[ $term == .*-wrapped ]] && {
                   term="${term#.}"
                   term="${term%-wrapped}"
                }
            ;;
        esac
    done

    # Log that the function was run.
    term_run=1
}
```

PR #5205, opened 2024-07-21
 Take into account scrolloff when moving viewport

### LINKED ISSUE #5178: [BUG] Viewport motion no longer takes scrolloff into account
### Version of Kakoune

v2024.05.09

### Reproducer

- Set a scrolloff in kakrc, i.e.: `set-option global scrolloff 5,5`
- Open a document that's longer than the viewport
- Scroll down further than the scroll offset (say 10 lines)
- Press `vt`

### Outcome

The viewport will place the current line at the absolute top of the viewport, without taking the scroll offset into account (which it used to do before this version). So that means that if you press `j` after the above steps, the view will suddenly jump to apply the scroll offset.

### Expectations

For the viewport motions (`vt`, `vb`, etc.) to take the scroll offset into account, as it used to do before the latest release.

### Additional information

_No response_

JustTaken:
`vt` and `vb` respect custom scroll offset.
fixes #5178

### DIFF
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -383,6 +383,9 @@ void view_commands(Context& context, NormalParams params)
 
         const BufferCoord cursor = context.selections().main().cursor();
         Window& window = context.window();
+        const DisplayCoord max_offset{(window.dimensions().line - 1)/2, (window.dimensions().column - 1)/2};
+        const DisplayCoord scrolloff =
+            std::min(context.options()["scrolloff"].get<DisplayCoord>(), max_offset);
         switch (*cp)
         {
         case 'v':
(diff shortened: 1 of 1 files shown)


src/normal.cc
```diff
@@ -383,6 +383,9 @@ void view_commands(Context& context, NormalParams params)
 
         const BufferCoord cursor = context.selections().main().cursor();
         Window& window = context.window();
+        const DisplayCoord max_offset{(window.dimensions().line - 1)/2, (window.dimensions().column - 1)/2};
+        const DisplayCoord scrolloff =
```
mawww, line 387: I dont think this is correct, you want the min of lines and the min of columns, but here `min({1_line, 3_col}, {2_line, 1_col})` is `{1_line, 3_col}`, not {1_line, 1_col}`

JustTaken: You are absolutely right. Perhaps i should modify the bindings `v<` and `v>` to account column offsets as well

Issue #4856, opened 2023-02-26
[REQUEST] Smart case regex modifier


evanrelf:
### Feature

We have `(?i)` to start case-insensitive matching, `(?I)` (the default) to start case-sensitive matching, but Kakoune doesn't Vim's `smartcase` behavior (case-insensitive until you add a capital letter). It would be nice if there was another modifier to enable `smartcase` behavior.

`ripgrep` also supports a `--smart-case` flag, which I think describes the behavior nicely:

```
-S, --smart-case
   Searches case insensitively if the pattern is all lowercase. Search case sensitively
   otherwise.

   A pattern is considered all lowercase if both of the following rules hold:

   First, the pattern contains at least one literal character. For example, a\w contains a
   literal (a) but just \w does not.

   Second, of the literals in the pattern, none of them are considered to be uppercase
   according to Unicode. For example, foo\pL has no uppercase literals but Foo\pL does.

   This overrides the -s/--case-sensitive and -i/--ignore-case flags.
```

Related issues:

- https://github.com/mawww/kakoune/issues/1958 (was mentioned but not really discussed)
- https://github.com/mawww/kakoune/issues/84 (discussed `*`, not `/` `?` `<a-/>` `<a-?>`)

### Usecase

Right now I have the following mappings in my config:

```kakounescript
map global "normal" "/" "/(?i)"
map global "normal" "?" "?(?i)"
map global "normal" "<a-/>" "<a-/>(?i)"
map global "normal" "<a-?>" "<a-?>(?i)"
```

I'd prefer if my search mappings were case-aware, changing behavior based on the contents of my regex, rather than needing a separate set of mappings for each behavior.

PR #5208, opened 2024-07-28
Add smartcase to the regex engine.

jamiedsmith95:
Implementation of smart case for regex, using (?c) as a modifier flag.
I believe completes feature request: https://github.com/mawww/kakoune/issues/4856

### DIFF
--- a/src/regex_impl.cc
+++ b/src/regex_impl.cc
@@ -1,5 +1,4 @@
 #include "regex_impl.hh"
-
 #include "exception.hh"
 #include "string.hh"
 #include "unicode.hh"
@@ -9,7 +8,6 @@
 #include "string_utils.hh"
 #include "vector.hh"
 #include "utils.hh"
-
 #include <cstdio>
 #include <cstring>
 #include <limits>
@@ -153,6 +151,7 @@ private:
         None              = 0,
         IgnoreCase        = 1 << 0,
         DotMatchesNewLine = 1 << 1,
+        SmartCase         = 1 << 2,
     };
     friend constexpr bool with_bit_ops(Meta::Type<Flags>) { return true; }
 
@@ -212,6 +211,7 @@ private:
                     case 'I': m_flags &= ~Flags::IgnoreCase; break;
                     case 's': m_flags |= Flags::DotMatchesNewLine; break;
                     case 'S': m_flags &= ~Flags::DotMatchesNewLine; break;
+                    case 'c': m_flags |= Flags::SmartCase; m_flags |= Flags::IgnoreCase; break;
                     case ')':
                         m_pos = Iterator{it, m_regex};
                         return true;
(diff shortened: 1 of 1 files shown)


src/regex_impl.cc
```diff
@@ -212,6 +211,7 @@ struct RegexParser
                     case 'I': m_flags &= ~Flags::IgnoreCase; break;
                     case 's': m_flags |= Flags::DotMatchesNewLine; break;
                     case 'S': m_flags &= ~Flags::DotMatchesNewLine; break;
+                    case 'c': m_flags |= Flags::SmartCase; m_flags |= Flags::IgnoreCase; break;
```
mawww, line 214: `m_flags |= Flags::SmartCase | Flags::IgnoreCase;` ?

src/regex_impl.cc
```diff
@@ -328,6 +328,14 @@ struct RegexParser
                 if (contains(StringView{"^$.*+?[]{}"}, cp) or (cp >= 0xF0000 and cp <= 0xFFFFF))
                     parse_error(format("unexpected '{}'", cp));
                 ++m_pos;
+                if (is_upper(cp) && (m_flags & Flags::SmartCase)) {
+                  m_flags &= ~Flags::IgnoreCase;
```
mawww, line 332: Why do we clear IgnoreCase here ? Doesn't that mean we have a different behaviour for the first upper case character ?

mawww (changes requested):
A few comments on the implementation, would be nice to add a few test cases and add this to the regex documentation.

src/regex_impl.cc
```diff
@@ -9,7 +8,6 @@
 #include "string_utils.hh"
 #include "vector.hh"
 #include "utils.hh"
-
```
mawww, line 12: Can we leave those empty lines ?

src/regex_impl.cc
jamiedsmith95: Shouldn't do, the IgnoreCase isn't checked until the add_node call just after this check.

jamiedsmith95:
> A few comments on the implementation, would be nice to add a few test cases and add this to the regex documentation.

I have addressed your comments. I have also modified it to also make an uppercase char alter existing character classes (otherwise the behaviour of character classes would depend on whether they were before or after the uppercase). A couple of tests and a line in the modifier part of the docs. The only thing to emphasise is that character classes will depend on smart case but will not themselves trigger it. e.g. [A-Z] will not count as an uppercase.

Supreeeme:
@jamiedsmith95 Are you still interested in getting this merged, or would you mind if I took it over?

src/regex_impl.cc
```diff
@@ -334,8 +334,9 @@ struct RegexParser
                     m_flags &= ~Flags::IgnoreCase;
                     for (ParsedRegex::Node &node : m_parsed_regex.nodes)
                         node.ignore_case = false;
-                  
-
+                    for (CharacterClass &char_class : m_parsed_regex.character_classes)
```
d3m3t3r, line 337: This is insufficient as these character classes might have already been built as case-insensitive (see character_class() method using to to_lower() on ranges).  E.g. "(?c)[A-Z]+foO"  wouldn't match "BARfoO" because '[A-Z]' was already created as '[a-z]'.

src/regex_impl.cc
```diff
@@ -334,8 +334,9 @@ struct RegexParser
                     m_flags &= ~Flags::IgnoreCase;
```
d3m3t3r, line 334: I believe SmartCase flag should be cleared too as it's not necessary to execute this block for every subsequent uppercase codepoint again and again.

PR #5209, opened 2024-08-04
add scroll coordinates in scroll key

enricozb:
`on-key` + a scoll event now emits `<scroll:amount:line.column>`, instead of `<scroll:amount>`. That is, the position of the mouse is included in the scroll key. This matches up with mouse press and release events `<mouse:button:action:line.column>`.

- adds scroll amount in the upper 16-bits of `Key.modifiers`, allowing `key` to store the coordinates.

- additionally, this PR fixes an issue where `ui_options terminal_status_on_top=yes` is set, we run `on-key %{ echo %val{key} }` and clicking on the status line, we get `<mouse:press:left:65536.1>` (which is not the correct line location). This PR now returns `<mouse:press:left:0.1>`, which I think is at least a bit more correct.

### DIFF
--- a/src/input_handler.cc
+++ b/src/input_handler.cc
@@ -127,7 +127,10 @@ struct MouseHandler
 
         Buffer& buffer = context.buffer();
         BufferCoord cursor;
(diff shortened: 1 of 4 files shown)


enricozb:
It looks like the checks failed due to timeouts, not sure how to get them to re-run.

PR #5210, opened 2024-08-04
Add support for double underline

arrufat:
Nice to have, personally, I want to use it with this snippet, from @alexherbo2 

```kak
declare-option -hidden range-specs hex_color_code_ranges
add-highlighter shared/hex_color_code ranges hex_color_code_ranges
define-command update_hex_color_code_ranges %{
    set-option window hex_color_code_ranges %val{timestamp}
    evaluate-commands -draft %{
        execute-keys 'gtGbx' # select the visible part of the buffer only.
        evaluate-commands -draft -verbatim try %{
            execute-keys 's(#|0[xX]|rgb:)([0-9A-Fa-f]{6})<ret>'
            evaluate-commands -itersel %{
                set-option -add window hex_color_code_ranges "%val{selection_desc}|,,rgb:%reg{2}+U"
            }
        }
    }
}
hook -always global NormalIdle '' update_hex_color_code_ranges
hook -always global InsertIdle '' update_hex_color_code_ranges
hook -always global PromptIdle '' update_hex_color_code_ranges
```

Which will highlight colors by double-underlining them with the appropriate color.
I use curly underlines for warnings and errors, and a single underline to highlight searches.

A sample on Foot:
image


arrufat:
If you don't like U, we could use w, instead.

PR #5215, opened 2024-08-12
Echo json

web-sst:
Add -quoting json option to echo. This converts each argument to a json string and joins on newline. It supports this, for example:
```
echo -quoting json -to-shell-script 'jq -n "[inputs] ..."' %val{selections}
```
A script can obtain potentially large sets of strings, like selections, as input rather than environment.

### DIFF
--- a/doc/pages/commands.asciidoc
+++ b/doc/pages/commands.asciidoc
@@ -275,8 +275,11 @@ of the file onto the filesystem
             embedded quotes.
 
         - *shell*::::
-            also wrap each arguments in single quotes and escape
+            also wrap each argument in single quotes and escape
             embedded quotes in a shell compatible way.
+
+        - *json*::::
+            convert each argument to a json string and join on newline.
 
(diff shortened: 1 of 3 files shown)


web-sst:
A coincident merge makes this PR conflict with src/string_utils.hh. In addition, json.hh now includes string_utils.hh, complicating my changes. I think the way to fix is to break out a quoting.hh from string_utils.hh, but I'd appreciate guidance from someone more familiar with the code.

mawww:
I am not sure how much value json quoting has in the first place as we still only emit a list of strings. I would like to see a use case where using the existing mechanisms is much more cumbersome before adding this.

You can, for example, trivially get the list of strings in "$@" with `echo -quoting shell -to-shell-script 'set -- $(cat); ...'`

web-sst:
I just opened an updated PR (#5217) that is compatible with current master. I'll respond over there.

Issue #3783, opened 2020-10-07
Add `echo -quoting json`


alexherbo2:

danr:
Would you use this with echo's `-file` switch? Regardless I would see more use for a shell env var `$kak_json_selections` similar to `$kak_quoted_selections`: it would json-escape instead of shell-escape.

PR #5217, opened 2024-08-14
Echo quoting json

web-sst:
This PR implements a new quoting option, `echo -quoting json` as suggested in #3783. This option formats each argument as a json string and prints them joined by newlines. A simple use case is:
```
echo -quoting json -to-shell-script "jq -n '[inputs]' | ..."
```
This way large and list-valued outputs are easy to handle in external scripts without having to pass them in environment.

The implementation is not very complex, but does involve some minor refactoring, `splitting quoting.hh` and `option_strings.hh` out from `string_utils.hh`. While this could have been done with a split of `quoting.hh` alone, it seemed to me less awkward to split them both. I followed the apparent conventions as I understand them, but would appreciate a review from a real C++ programmer, especially of the `#include` updates .

Also  `string.hh` needed `#include <algorithm>` to compile on FreeBSD 14.1, clang 18.1.5.


web-sst:
@mawww wrote:

> I am not sure how much value json quoting has in the first place as we still only emit a list of strings. I would like to see a use case where using the existing mechanisms is much more cumbersome before adding this.
> 
> You can, for example, trivially get the list of strings in "$@" with echo -quoting shell -to-shell-script 'set -- $(cat); ...'

Following this suggestion I can do this:
```
echo -quoting shell -to-shell-script 'eval set -- $(cat); jq -n "\$ARGS.positional" --args "$@" ...'
```
rather than
```
echo -quoting json -to-shell-script 'jq "[inputs]" ...'
```
I prefer the second version because it avoids extra parsing in shell, avoids having to put the potentially large content into command-line arguments, and is much clearer. The first approach was not obvious to me, and needed the `eval` to function as expected, whereas the second approach does seem obvious. There are perhaps two notions of simplicity here: minimizing the number of ways to access data versus allowing the most direct access to data. I look at the addition of json quoting as the mere composition of features already in kakoune in service of direct access.

PR #5231, opened 2024-09-08
Highlight ASCII control characters unambiguously

### LINKED ISSUE #2936: Display Control Characters as ^A, ^M
This is a feature request:

When using kakoune, ascii control characters display as a generic non-displayable unicode character:

kakoune-control-characters

using another editor I get this which is more useful for me:

nvim-control-characters

I'd like it if kakoune could render the control characters like that so I can distinguish between them. (I think there may be fonts out there that can render them, but none of the ones I've tried work - Droid Sans, Liberation Sans, Source Code Pro.)

Thanks for considering my request

alois31:
Previously, all ASCII control characters would be shown as the replacement character. This leads to a visual loss of information and ambiguities, particularly when opening files that contain lots of control characters (or binary files). Use the common circumflex notation, known from software such as `cat -v` or vim, instead.

Fixes #2936

### DIFF
--- a/src/highlighters.cc
+++ b/src/highlighters.cc
@@ -1327,7 +1327,12 @@ void expand_unprintable(HighlightContext context, DisplayBuffer& display_buffer,
                     if (ByteCount pos(next - line_data); pos < atom_it->end().column)
                         atom_it = line.split(atom_it, {begin.line, pos});
 
-                    atom_it->replace("�");
+                    if (cp < ' ')
+                        atom_it->replace(format("^{}", (char)(cp + '@')));
+                    else if (cp == 127)
+                        atom_it->replace("^?");
+                    else
+                        atom_it->replace("�");
                     atom_it->face = error;
                     break;
                 }


mawww:
I am afraid this will make vertical movements a bit broken as we use the buffer line column width so we are not relying on highlighting at all for this. If we really want to change this I'd prefer to use the unicode control pictures which should not shift anything vertically.

alois31:
Thank you for catching the breakage. Not sure if the Unicode pictures are much better than the current state, they are almost completely unreadable with many fonts in normal sizes.

arachsys:
Random thought: would a more readable alternative to a replacement character or a tiny unicode control picture be a specially highlighted letter, e.g. ^a (optionally?) becomes an 'a' in a special face instead of �?

Delapouite:
Talking about `cat -v`, [bat](https://github.com/sharkdp/batl) which aims to be a "modern" alternative with bells and whistles adds a way to decide between [Caret Notation](https://en.wikipedia.org/wiki/Caret_notation) and [Control Pictures](https://en.wikipedia.org/wiki/Control_Pictures) (`␜, ␝, ␞, ␟`…).

Here's the relevant extract of its [man](https://man.archlinux.org/man/bat.1): 

```
-A, --show-all

    Show non-printable characters like space, tab or newline. Use '--tabs' to control the width of the tab-placeholders.

--nonprintable-notation <notation>

    Specify how to display non-printable characters when using --show-all.

    Possible values:

caret
    Use character sequences like ^G, ^J, ^@, .. to identify non-printable characters
unicode
    Use special Unicode code points to identify non-printable characters
```

I'm not sure how such an option would translate in Kakoune world.

> Not sure if the Unicode pictures are much better than the current state, they are almost completely unreadable with many fonts in normal sizes.

I agree that these chars are usually quite small, but it would nonetheless carry more information that a soulless � and when such (rare) situation happens, well you can always temporarily zoom the font size in your terminal.

Issue #4436, opened 2021-11-13
[REQUEST] Spell-previous command


Elither:
### Feature

A command like spell-next but for previous instead.

### Usecase

We already have a next, I'm not sure why we don't have a previous?

yoshiyoshyosh:
[this commit](https://github.com/mawww/kakoune/commit/1a2bf04f1bd60cd50367e644517c2e282910c04e) to add `spell-prev` was merged in version 2025.06.03

Delapouite:
Yes `spell-prev` is now available, this issue can be closed.

PR #5233, opened 2024-09-09
rework spell-next and add a spell-prev command

yoshiyoshyosh:
addresses #4436 since I was finding the lack of this command annoying as well

we can take advantage of the fact that the field separator in awk is a regular expression in POSIX to condense the awk command by splitting records on space and fields on any one of `[.,|]`

additionally, add a `spell-prev` command by simply keeping track of the previous spell template and choosing when the awk iteration stops via a `rev` variable

even though it's posix, I tested with gawk, nawk, and busybox just to be safe. all worked

I tried to look around and emulate the same coding style of other shell commands but if something's off or can be done better let me know

### DIFF
--- a/rc/tools/spell.kak
+++ b/rc/tools/spell.kak
@@ -87,57 +87,61 @@ define-command spell-clear %{
     unset-option buffer spell_regions
 }
 
-define-command spell-next %{ evaluate-commands %sh{
-    anchor_line="${kak_selection_desc%%.*}"
-    anchor_col="${kak_selection_desc%%,*}"
-    anchor_col="${anchor_col##*.}"
-
-    start_first="${kak_opt_spell_regions%%|*}"
-    start_first="${start_first#* }"
+define-command spell-rel-jump -hidden -params 0..1 %{ evaluate-commands %sh{
+    spell_first="${kak_opt_spell_regions%%|*}"
+    spell_first="${spell_first#* }"
+    spell_last="${kak_opt_spell_regions##* }"
+    spell_last="${spell_last%|*}"
 
     # Make sure properly formatted selection descriptions are in `%opt{spell_regions}`
-    if ! printf %s "${start_first}" | grep -qE '^[0-9]+\.[0-9]+,[0-9]+\.[0-9]+$'; then
+    if ! printf %s "${spell_first}" | grep -qE '^[0-9]+\.[0-9]+,[0-9]+\.[0-9]+$'; then
         exit
     fi
 
-    printf %s "${kak_opt_spell_regions#* }" | awk -v start_first="${start_first}" \
-                                                  -v anchor_line="${anchor_line}" \
-                                                  -v anchor_col="${anchor_col}" '
-        BEGIN {
-            anchor_line = int(anchor_line)
-            anchor_col = int(anchor_col)
-        }
+    get_prev=0
+    if [ "$1" = "-rev" ]; then
+        get_prev=1
+    elif [ -n "$1" ]; then
+        echo "fail -- Unrecognised parameter $(kakquote "$1")"
+    fi
 
-        {
-            for (i = 1; i <= NF; i++) {
-                sel = $i
-                sub(/\|.+$/, "", sel)
-
-                start_line = sel
-                sub(/\..+$/, "", start_line)
-                start_line = int(start_line)
-
-                start_col = sel
-                sub(/,.+$/, "", start_col)
-                sub(/^.+\./, "", start_col)
-                start_col = int(start_col)
-
-                if (start_line < anchor_line \
-                    || (start_line == anchor_line && start_col <= anchor_col))
-                    continue
-
-                target_sel = sel
-                break
+    printf %s "${kak_opt_spell_regions#* }" | \
+        awk -v spell_first="${spell_first}" -v spell_last="${spell_last}" \
+            -v get_prev="${get_prev}" -F '[.,|]' -v RS=" " '
+            BEGIN {
+                split(ENVIRON["kak_selection_desc"], sel)
+                cursor_row = sel[3]
+                cursor_col = sel[4]
+                # reverse order so cursor can be at beginning for spell-prev
+                split(spell_last, tmp)
+                spell_last = tmp[3] "." tmp[4] "," tmp[1] "." tmp[2]
             }
-        }
 
-        END {
-            if (!target_sel)
-                target_sel = start_first
+            {
+                # $1 - $4 is candidate row.col,row.col
+                check_row = $1 > cursor_row
+                check_col = $1 == cursor_row && (get_prev ? $2 >= cursor_col : $2 > cursor_col)
 
-            printf "select %s\n", target_sel
-        }'
+                if (check_row || check_col) {
+                    next_spell = $1 "." $2 "," $3 "." $4
+                    exit
+                }
+                prev_spell = $3 "." $4 "," $1 "." $2
+            }
+
+            END {
+                target_spell = get_prev ? prev_spell : next_spell
+                if (!target_spell)
+                    target_spell = get_prev ? spell_last : spell_first
+
+                printf "select %s\n", target_spell
+            }
+        '
 } }
+
+define-command spell-next %{ spell-rel-jump }
+
+define-command spell-prev %{ spell-rel-jump -rev }
 
 define-command \
     -docstring "Suggest replacement words for the current selection, against the last language used by the spell-check command" \


rc/tools/spell.kak
```diff
@@ -87,57 +87,60 @@ define-command spell-clear %{
...
-        BEGIN {
-            anchor_line = int(anchor_line)
-            anchor_col = int(anchor_col)
-        }
-
-        {
-            for (i = 1; i <= NF; i++) {
-                sel = $i
-                sub(/\|.+$/, "", sel)
+    if [ "$1" = "-rev" ]; then
```
krobelus, line 105: The FS approach seems good.

Here we should initialize `rev`, for the (unlikely) case where we inherit it from the environment

rc/tools/spell.kak
```diff
@@ -87,57 +87,60 @@ define-command spell-clear %{
...
+    elif [ -n "$1" ]; then
+        echo "fail -- Unrecognised parameter $(kakquote "$1")"
+    fi
 
-                start_line = sel
-                sub(/\..+$/, "", start_line)
-                start_line = int(start_line)
+    printf %s "${kak_opt_spell_regions#* }" | \
+        awk -v spell_first="${spell_first}" -v spell_last="${spell_last}" \
+            -v anchor_line="${anchor_line}" -v anchor_col="${anchor_col}" \
```
krobelus, line 113: (unrelated:) I don't really understand why we use the anchor's coordinates and not the cursor's,
or the minimum/maximum of cursor/anchor.
The current behavior is that given

	  mispelled    tpo
	A------------C

spell-next will jump to "mispelled", which is weird.

rc/tools/spell.kak
```diff
@@ -87,57 +87,60 @@ define-command spell-clear %{
...
+            }
 
-                start_col = sel
-                sub(/,.+$/, "", start_col)
-                sub(/^.+\./, "", start_col)
-                start_col = int(start_col)
+            {
+                # rev still has anchor at start, so we need to check end of spell
+                col_check = (!rev && $2 > anchor_col) || (rev && $4 >= anchor_col)
+                if ($1 >= anchor_line && ($1 != anchor_line || col_check)) {
```
krobelus, line 123: I don't understand why we need a special case for the spell-prev case. Shoulnd't we just look at the starting position of the word (for simplicity)?

rc/tools/spell.kak
yoshiyoshyosh: good point--I find that weird as well. definitely wouldn't be hard to implement

rc/tools/spell.kak
yoshiyoshyosh: the comment here... kinda sucks, my bad. the intention is that when using `spell-prev`, the "next word" we should search for should "start" at the end of the word, since the search direction is reversed
imagine the following case where the anchor and cursor are reversed when selecting a word:

```
word misspelled1 word misspelled2 word
                      C---------A
```

if we compare the anchor to the start of the word when using `spell-prev`, it'll find `misspelled2` again, since the start of the word is behind the anchor. if we compared the anchor to the end of the word, it'll find `misspelled1` as intended.

the same situation when using `spell-next` is:

```
word misspelled1 word misspelled2 word
     A---------C
```

since it compares the anchor to the start of the word, it'll find `misspelled2` next, as intended

this kind of relates to the above notion of maybe reworking what is actually compared when searching for the next misspelled word.

rc/tools/spell.kak
yoshiyoshyosh: hmm. I think the best solution to this would be to, for spell-next, actually check the end of the word with the end of the current selection, while for spell-prev, check the start of the word with the start of the current selection, opposite of intuition.

This would cause the following situations, which to me, feels the most intuitive when searching for the next misspelled word:

When using `:spell-next`:
```
word misspelled1 word misspelled2 word
   X-------X
     X---------X
```
```
word misspelled1 word misspelled2 word
         X---------X
                      X---------X
```

When using `:spell-prev`:
```
word misspelled1 word misspelled2 word
         X---------X
     X---------X
```
```
word misspelled1 word misspelled2 word
                  X--------X
     X---------X
```

yoshiyoshyosh:
here's the paradigm I implemented:

for `spell-next`:

```
word misspelled1 word misspelled2 word
  X
     X---------X
```
```
word misspelled1 word misspelled2 word
  X-----X
     X---------X
```
```
word misspelled1 word misspelled2 word
       X-----X
                      X---------X
```
```
word misspelled1 word misspelled2 word
           X------X
                      X---------X
```
```
word misspelled1 word misspelled2 word
  X---------------X
                      X---------X
```


for `spell-prev`:

```
word misspelled1 word misspelled2 word
                                   X
                      X---------X
```
```
word misspelled1 word misspelled2 word
                             X-----X
                      X---------X
```
```
word misspelled1 word misspelled2 word
                        X-----X
                      X---------X
```
```
word misspelled1 word misspelled2 word
                  X------X
     X---------X
```
```
word misspelled1 word misspelled2 word
                  X-----------------X
     X---------X
```

to me, it makes more intuitive sense for `:spell-prev` to select the "current" word when the selection is contained completely inside a word, while `:spell-next` goes to select the next word. would like your thoughts though

krobelus:
Thanks for the test cases; this looks okay though
I think I'd rather simplify this to only compare the cursor against the selection start. That should be more predictable.
Other movements commands like `w` and `f` behave the same way.
Maybe I've missed a case where the proposed/existing behavior is better; I haven't used spell much.

If we stick with the proposed behavior, I'd like to make the
code easier to read by adding some variable names; here's my attempt:


```diff
diff --git a/rc/tools/spell.kak b/rc/tools/spell.kak
index 04c8518d2..93e384f5a 100644
--- a/rc/tools/spell.kak
+++ b/rc/tools/spell.kak
@@ -111,26 +111,30 @@ define-command spell-rel-jump -hidden -params 0..1 %{ evaluate-commands %sh{
             BEGIN {
                 # set selection start and end independent of orientation
                 split(ENVIRON["kak_selection_desc"], sel, /[.,|]/)
-                abc = sel[1] < sel[3] || (sel[1] == sel[3] && sel[2] <= sel[4])
-                sel_start[1] = sel[3 - abc * 2]
-                sel_start[2] = sel[4 - abc * 2]
-                sel_end[1] = sel[1 + abc * 2]
-                sel_end[2] = sel[2 + abc * 2]
-            }
-
-            !reversed {
-                row_check = $1 > sel_end[1]
-                col_check = $1 == sel_end[1] && \
-                    $4 > sel_end[2] && $2 > sel_start[2]
-            }
-            reversed {
-                row_check = $1 > sel_start[1]
-                col_check = $1 == sel_start[1] && \
-                    $4 >= sel_start[2] && $2 >= sel_start[2]
+                anchor_before_cursor = sel[1] < sel[3] || (sel[1] == sel[3] && sel[2] <= sel[4])
+                sel_start_line = sel[3 - anchor_before_cursor * 2]
+                sel_start_column = sel[4 - anchor_before_cursor * 2]
+                sel_end_line = sel[1 + anchor_before_cursor * 2]
+                sel_end_column = sel[2 + anchor_before_cursor * 2]
             }
 
             {
-                if (row_check || col_check) {
+                word_start_line = $1
+                word_start_column = $2
+                word_end_line = $3
+                word_end_column = $4
+
+                word_start_after_sel_start = word_start_line > sel_start_line ||
+                    (word_start_line == sel_start_line && word_start_column > sel_start_column)
+                word_end_after_sel_end = word_end_line > sel_end_line ||
+                    (word_end_line == sel_end_line && word_end_column > sel_end_column)
+                word_start_before_sel_start = word_start_line < sel_start_line ||
+                    (word_start_line == sel_start_line && word_start_column < sel_start_column)
+
+                if (reversed && !word_start_before_sel_start) {
+                    exit;
+                }
+                if (!reversed && word_start_after_sel_start && word_end_after_sel_end) {
                     next_spell = $1 "." $2 "," $3 "." $4
                     exit
                 }
```

yoshiyoshyosh:
totally didn't forget about this PR. I would never.

anyway, after rummaging on it for a bit, yeah checking from the cursor makes the most sense--not any of my weird "absolute selection bounds" shenanigans I was doing. it's consistent with the movement keys, which will feel the most intuitive to those used to kakoune imo (it makes the most sense to me, at least)

regarding the variable names... I feel like your names introduce a ton of visual noise, which makes it really hard to actually read the conditionals that are going on. I know mine are quite terse, but with the knowledge that it's always split to be `start_row.start_col,end_row.end_col`, I'm not sure it matters that much? At least for me, keeping the fields `$1` thru `$4` as-is is easier to read. I did add a comment if that helps

krobelus (approved):

PR #5234, opened 2024-09-11
Highlight targets in Makefiles

arrufat:


rc/filetype/makefile.kak
```diff
@@ -34,6 +34,7 @@ add-highlighter shared/makefile/content default-region group
 add-highlighter shared/makefile/comment region (?<!\\)(?:\\\\)*(?:^|\h)\K# '$' fill comment
 add-highlighter shared/makefile/evaluate-commands region -recurse \( (?<!\$)(?:\$\$)*\K\$\( \) fill value
 
+add-highlighter shared/makefile/content/ regex ^([a-zA-Z0-9_.-]+): 1:function
```
krobelus, line 37: I think they are already highlighted by this line:

`add-highlighter shared/makefile/content/ regex ^\S.*?(::|:|!)\s 0:variable`

The trailing `\s` is awkward but since we always have a trailing newline, it's correct

arrufat: Right, I hadn't noticed because it's highlighted as a variable, which in my colorscheme is the default for normal text, too. If I try gruvbox, then everything is fine. I will just override it locally in my config then.

PR #5239, opened 2024-09-22
Allow setting $kak_bin to test kak binaries in alternate locations

dontlaugh:
The test/run script will check for a $kak_bin environment variable and, if it's set, treat it as an absolute path to a kak binary. Otherwise the default path src/kak is used.

### DIFF
--- a/test/run
+++ b/test/run
@@ -20,6 +20,9 @@ main() {
       '
 
   root=$PWD
+  if [ -z $kak_bin ]; then
+    kak_bin=$root/../src/kak
+  fi
   tmpdir="${TMPDIR:-/tmp}"
   work=$(mktemp -d $tmpdir/kak-tests.XXXXXXXX)
   session="kak-tests"
@@ -57,7 +60,7 @@ main() {
       if [ -f env ]; then
         . ./env
       fi
-      exec $DEBUGGER $root/../src/kak out -n -s "$session" -ui json -e "$kak_commands" >ui-out <ui-in
+      exec $DEBUGGER $kak_bin out -n -s "$session" -ui json -e "$kak_commands" >ui-out <ui-in
     ) &
     kakpid=$!
 
@@ -71,7 +74,7 @@ main() {
       ui_out '{ "jsonrpc": "2.0", "method": "set_ui_options", "params": [{}] }'
     fi
 
-    finished_commands |$root/../src/kak -p "$session" 2>/dev/null
+    finished_commands |$kak_bin -p "$session" 2>/dev/null
     cat <&4 >/dev/null
 
     wait $kakpid


test/run
```diff
@@ -20,6 +20,9 @@ main() {
       '
 
   root=$PWD
+  if [ -z $kak_bin ]; then
```
krobelus, line 23: should quote the variable expansion here to avoid weird errors when it contains spaces.
Same below, or do I do want to allow `kak_bin='strace -o foo.strace kak'` like this?
Seems like what DEBUGGER is for.

test/run
```diff
@@ -20,6 +20,9 @@ main() {
...
+    kak_bin=$root/../src/kak
```
krobelus, line 24: interesting, I've never needed that; it seems potentially useful (can you maybe elaborate?) but hard to get right

Note that `$kak_bin` must have the `kak` basename (so no `kak.debug` or similar because of our libexec trick
Our makefile does `ln -sf ../../bin/kak $(libexecdir)/kak` and on startup, `kak` prepends libexecdir to PATH to be able to invoke copies that are compatible with this version.

So if your usage doesn't make sure that the libexec version points to `$kak_bin`, things are subtly broken.
I'm not sure if this is easy to use safely

test/run
dontlaugh: quoting seems appropriate for both usages of `kak_bin`, since I can't think of a use case where splitting the variable's value on spaces would be useful. It will always be a single path, and not a potential list of args.

test/run
dontlaugh: The TL;DR use-case is that it would be convenient (for me) to run tests with binaries that aren't at the path **./src/kak**. This came up for me because I'm using a different build system (see below), but it also seems like it could be useful to run tests against different binaries built from different branches, different worktrees, etc. while still using a single revision of the test code.

<details>
<summary>Short digression: Why I have an alternate Kakoune build</summary>

Regarding the alternate build system. I've long maintained a Kakoune fork just so that I can pull in the master branch, merge locally, build, make install, and run with the latest code.

Over the weekend I switched my fork's build to Meson, because - while Kakoune has (seriously) the most beautiful Makefile I've ever seen - I wanted to build my fork with Meson because it's just easier for me. I am hoping to keep my hacks out of the src/ folder and the upstream Makefile

* [Repo](https://git.sr.ht/~coleman/kakoune/tree/hacking/item/meson.build)
* [Build](https://builds.sr.ht/~coleman/job/1334252)

Meson builds things out-of-tree, so the binary ends up in `buildDir/kak`

```
% ls -l buildDir
.rw-rw-r-- 31k coleman 21 Sep 21:33 build.ninja
.rw-rw-r-- 33k coleman 21 Sep 21:33 compile_commands.json
.rwxrwxr-x 39M coleman 21 Sep 21:33 kak
drwxrwxr-x   - coleman 21 Sep 21:33 kak.p
drwxrwxr-x   - coleman 21 Sep 21:33 meson-info
drwxrwxr-x   - coleman 21 Sep 21:33 meson-logs
drwxrwxr-x   - coleman 21 Sep 21:33 meson-private
drwxrwxr-x   - coleman 21 Sep 21:33 src
```

And I ran my tests like

```
kak_bin=/home/coleman/Code/kakoune/buildDir/kak ./run
```

</details>

If this isn't something you'd like to be configurable, I can close this. I won't be offended. But to me it seems like something the test runner ought to be able to do.

mawww: I am wondering about this change, on one hand it is pretty small and I can see the utility of specifying another binary, but as @krobelus mentionned this is brittle due to our expectation that the binary is called `kak`.

From your description of what you´d like this for, wouldn't it be even easier to just create a `src/kak` symlink that points to `buildDir/kak`?

krobelus: to reiterate, the thing that prevents your approach from working is this hardcoded link:

```
$ file libexec/kak/kak
libexec/kak/kak: symbolic link to ../../src/kak
```

which is checked into Git.

It would be possible to put the makefile in charge of creating this symlink
(which is currently dangling on a fresh clone though that's fine).
Not sure if that's worth it. I guess it's possible that one would want to run a test using `/bin/kak` but I guess haven't needed it.

>  I am hoping to keep my hacks out of the src/ folder and the upstream Makefile

> As long as upstream keeps all the actual source code under src/ we'll
> be able to merge in changes without much trouble.

I really don't understand why; copying `src/main.cc` seems strange to me.
Merge conflicts are our friends, learn to love them.

dontlaugh:
> From your description of what you´d like this for, wouldn't it be even easier to just create a src/kak symlink that points to buildDir/kak?

This would work great.

I am inclined to close this PR, since I have workarounds and it's not that big of a deal. And my job here is done, since you seem to understand my use case (testing $this binary vs $that binary). If you are working on the test framework in the future, keep this sort of configurability in mind :)

PR #5246, opened 2024-10-06
rc make: support multiline error pattern

ygrek:
apparently this is what is needed to support multiline error message pattern, e.g. for ocaml : 
```
File "admin.ml", line 109, characters 26-30:
109 |  		Hashtbl.replace status host st
       		                       ^^^^
Error: Unbound value host
```

NB jump_current_line highlights only the last line, technically should change to jump_current_selection, not sure if it is ok (compatibilty?)


ygrek:
ftr
```
  set-option global make_error_pattern %{File "([^"]+?)", line (\d+), characters (\d+)-\d+:.*?^Error: ([^\n]+)$}
```

rc/tools/make.kak
```diff
@@ -60,7 +60,7 @@ define-command -hidden make-jump %{
                 set-register a "%reg{1}/%reg{2}" "%reg{3}" "%reg{4}" "%reg{5}"
             } catch %{
                 set-register / %opt{make_error_pattern}
-                execute-keys <a-h><a-l> s<ret>l
+                execute-keys n
```
krobelus, line 63: sounds reasonable. I didn't test this but I think this should be something like `<a-h><semicolon>n` to make sure `<ret>` keeps working if the cursor is in the middle of an error message.

ygrek:
@krobelus i misunderstood what it was doing, now tested with both original and my error pattern, everything seems to work properly, also left comments

ygrek:
:pray:

PR #5254, opened 2024-11-12
grep.kak: make it possible to search for patterns starting with '-'

occivink:
Previously if you were to call `grep` while just selecting something like `-f`, it would interpret this as a flag and fail.


losnappas:
Won't this make it impossible to pass flags to the grep program? I like to use e.g. `rg -u` often.

occivink:
It shouldn't, the only affected case is when you do `:grep` with no arguments

PR #5266, opened 2024-12-23
create *stdin* buffer even when daemon

enricozb:
When running kakoune with a non-tty stdin (eg piping into kakoune) an `*stdin*` buffer is created. Currently, this only happens when kakoune is not a daemon (`-d` is not passed). This PR creates an `*stdin*` buffer even if kakoune is in daemon mode.

This will now create an `*stdin*` buffer:
```sh
echo hello | kak -d -s my-session
```
that can be accessed by connecting clients:
```sh
kak -c my-session
```


Screwtapello:
I suppose you could do this as:

```
kak -d -s my-session
echo hello | kak -c my-session -e "quit"
```

...but this has the disadvantage that `kak -c` still tries to initialise the terminal, sending a bunch of escape sequences to stdout. I guess if this is being run from a system service where stdout is not a tty, this might not be possible at all?

PR #5270, opened 2025-01-08
For OpenLineBelow, add newline to the end of the line of the current …

JasonBrownDeveloper:
For OpenLineBelow, add newline to the end of the line of the current selection, instead of the line below the selection, for a more consistent history

Given the following buffer:

> Line Zero\n$
> $

Pressing the ```o``` key on (zero indexed) line one, results in the following ```history```:

- 1561858 1 0 1561866 2 +0.0|L +0.1|i +0.2|n +0.3|e +0.4|  +0.5|Z +0.6|e +0.7|r +0.8|o +0.9|\n
1 1561872 - +2.0|\n
+2.0|Line Two

After the patch, pressing the ```o``` key, the history would look like the following, which would match if ```i``` and ```<ret>``` had been pressed on line one.

- 1561858 1 0 1561866 2 +0.0|L +0.1|i +0.2|n +0.3|e +0.4|  +0.5|Z +0.6|e +0.7|r +0.8|o +0.9|\n
1 1561872 - +1.0|\n
+2.0|Line Two

### DIFF
--- a/src/buffer.cc
+++ b/src/buffer.cc
@@ -408,7 +408,7 @@ void Buffer::check_invariant() const
 #endif
 }
 
-BufferRange Buffer::do_insert(BufferCoord pos, StringView content)
+BufferRange Buffer::do_insert(BufferCoord pos, StringView content, bool draft)
 {
     kak_assert(is_valid(pos));
 
@@ -452,7 +452,7 @@ BufferRange Buffer::do_insert(BufferCoord pos, StringView content)
     const auto end = at_end ? line_count()
                             : BufferCoord{ last_line, m_lines[last_line].length() - suffix.length() };
 
-    m_changes.push_back({ Change::Insert, pos, end });
+    m_changes.push_back({ draft ? Change::Draft : Change::Insert, pos, end });
     return {pos, end};
 }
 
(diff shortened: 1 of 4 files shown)


mawww:
I think this is a good change but I dont quite understand what the last commit is for.

JasonBrownDeveloper:
Some of the test suite fails with the simple OpenLineBelow change.  I don't remember the specifics, but the selection isn't updated properly, either it's not being extended to include the new blank line or it is and it shouldn't.  But it causes some plugins that do smart auto tab for braces to fail.

Unfortunately I don't have a lot of time to spend on this, but I am still poking at it when I can.  I'm working on other plugins that will need this eventually.

mawww:
Looked deeper into this, I dont think we can actually do this change as it impacts selections on the end of line, as shown by `regression/0-open-below-should-not-move-cursor-on-eol` failing. Moving cursor on eol impacts lots of indentation scripts because the user cursor now gets moved by scripts using `o` to add a line below it.

To be clearer, if you are in this situation:

```
          v cursor on eol
void foo() 
```

And you insert a `{`, an auto indent script may do `o}<esc>` to auto insert the closing brace on next line. But with this change, this ends up happening **in front** of the users cursor, which means subsequent edits by the user will take place past the auto-inserted `}`.

So it looks like this change is not really doable without significantly breaking the existing scripts, and it seem the existing behaviour is much better for that.

PR #5288, opened 2025-02-03
Change the no-op assert macros to silence a clang-tidy warning

dontlaugh:
When KAK_DEBUG is false (release mode), clang tidy emits a 
warning when we pass a pointer to kak_assert. It's mad about 
passing a pointer to `sizeof`.

We can avoid this warning by using `alignof(decltype(...))`. 
It will still type-check the expression.

### DIFF
--- a/src/assert.hh
+++ b/src/assert.hh
@@ -30,7 +30,7 @@ void on_assert_failed(const char* message);
                          "\" at " __FILE__ ":" TOSTRING(__LINE__)); \
     } catch (exception_type &err) {}
 #else
-    #define kak_assert(...) do { (void)sizeof(__VA_ARGS__); } while(false)
+    #define kak_assert(...) do { (void)alignof(decltype(__VA_ARGS__)); } while(false)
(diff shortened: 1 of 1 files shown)


src/assert.hh
```diff
@@ -30,7 +30,7 @@ void on_assert_failed(const char* message);
                          "\" at " __FILE__ ":" TOSTRING(__LINE__)); \
     } catch (exception_type &err) {}
 #else
-    #define kak_assert(...) do { (void)sizeof(__VA_ARGS__); } while(false)
+    #define kak_assert(...) do { (void)alignof(decltype(__VA_ARGS__)); } while(false)
     #define kak_expect_throw(_, ...) do { (void)sizeof(__VA_ARGS__); } while(false)
```
mawww, line 34: We probably should update this one as well

mawww (changes requested):

src/assert.hh
dontlaugh: You're right. Missed that one.

PR #5292, opened 2025-02-18
rc git: Use wider characters for added/modified lines

Supreeeme:
I find the current default characters very difficult to see, to the point where I didn't know if `git show-diff` was actually working.

### DIFF
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -2,10 +2,10 @@ declare-option -docstring "name of the client in which documentation is to be di
     str docsclient
 
 declare-option -docstring "git diff added character" \
-    str git_diff_add_char "▏"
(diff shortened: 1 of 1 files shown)


rc/tools/git.kak
```diff
@@ -2,10 +2,10 @@ declare-option -docstring "name of the client in which documentation is to be di
     str docsclient
 
 declare-option -docstring "git diff added character" \
-    str git_diff_add_char "▏"
+    str git_diff_add_char "▊"
```
krobelus, line 5: Yeah this is super hard to read, I can't even make out the green color. Good change.
Looks like the person who originally added this doesn't use it anymore..

I guess it's not as clear as `+` but that one is probably too easily confused with actual content.

PR #5302, opened 2025-03-26
Document built in command switch behavior

Supreeeme:


doc/pages/commands.asciidoc
```diff
@@ -4,6 +4,12 @@ Some commands take an exclamation mark (*!*), which can be used to force
 the execution of the command (i.e. to quit a modified buffer, the
 command *q!* has to be used). Aliases are mentioned below each command.
 
+For all commands that accept switches, any parameter beginning with a
+a single dash (*-*) will be interpreted to be a switch up until the first
+parameter that doesn't start with a single dash. Any unknown switches will
+cause an error. To force the remaining parameters to not be interpreted as 
+switches, you can pass two dashes (*--*) as a parameter.
```
mawww, line 11: I am not sure this properly describes how switches work, it is command dependent if switches have to be first argument or not for example, and show switches take parameters while other do not. I think this should be reworded to reflect that.

Supreeeme: Sorry for the delay. I reworded it a bit.

PR #5304, opened 2025-04-02
Add a Scroll Bar

enricozb:
A scroll bar that also displays selection lines in the buffer.

This PR adds:
- `terminal_scroll_bar` to `ui_options`, which controls whether the scroll bar should show.
- `ScrollBarGutter` and `ScrollBarHandle` faces describing how to style the gutter and handle. The foreground of these faces will be the color of the characters used to display selection lines.

## Screenshots

Here is a recording with the default styling and config
[asciicast](https://asciinema.org/a/1yqgd30qQ1G4OkWvOx4FfT29y)

Here is a recording with my personal styling and config:
[asciicast](https://asciinema.org/a/IZGDvrHV2ntDaaKrWCnqoCKFN)

I've tested this with both non-remote and remote clients.

## Issues

I am not the biggest fan of the implementation. The computation of `Vector<LineCount> selection_lines` happens on every draw call. This is needed to serialize these locations for `RemoteUI` and `JsonUI`. This is also computed regardless if the scrollbar is enabled, which is unfortunate.

I'm not sure what the appetite for adding something like this is, especially with my implementation. I saw there was a highlighter implementation (https://github.com/mawww/kakoune/pull/2597) but did not honestly consider it. If we feel like this is best done as a highlighter I can investigate that approach as well.


Delapouite:
Thanks! I appreciate that it does more than a "regular" scroll bar by adding visual markers for selections.

Of course some other text-editors/IDE offer much more with other kinds of markers such as git added/changed/removed lines, or info/warning/error diagnostics. But then you end up with a "3 lanes" scroll bars with a lot of infos which can be confusing. "There's a red mark at the beginning of the buffer : is it red because of git removed lined or red because it's an LSP error?"

So here, I think you strike the right balance by only focusing on one of the core principle of Kakoune which is its multi-selections capabilities.

enricozb:
I've improved the calculation of the size of the scroll bar handle. It works well with the `wrap` highlighter. It will decrease in size if the number of lines displayed to the user is small (due to a wrapped line taking up a lot of vertical space).

There is still a bug where when scrolling horizontally (which is not possible when the `wrap` highlighter is added), the scroll bar position is totally incorrect. This is because the display buffer `range` changes when scrolling horizontally. Still trying to fix this.

raiguard:
@enricozb Could you update this to latest master? The merge conflicts have piled up a bit. :heart:

raiguard:
Also, @mawww is this something you are willing to consider merging? It is a great usability improvement.

enricozb:
It's been a while since I've looked at this code, but I've merged master in and fixed the conflicts. It looks like a `DisplayCoord cursor_pos` argument was added to a few functions, i've also marked it as `const` as it wasn't being modified.

enricozb:
Just writing here, I think there's a small bug where (for some reason) the last (rightmost) char of the modeline is clipped, until we perform a resize.

image

Here's a screenshot where the upper client has been resized after opening, but the bottom has not yet.

raiguard:
I can confirm that I've seen this, but I don't have a reproduction.

enricozb:
This consistently happens to me on every client creation, and is fixed after a resize.

It's likely to do with some computation i wrote that isn't refreshed on changes to `terminal_scroll_bar` in `ui_options`.

enricozb:
Fixed the incorrect dimension calculation, no longer requires a resize. Also updated to master once again.

Yus314:
I'm the author of Kasane, an alternative Kakoune frontend using `kak -ui json` (same project behind #5455 and #5458).

The viewport `range` and `buffer_line_count` exposed in this PR would directly resolve a gap in the JSON UI protocol — currently there's no way for an alternative frontend to know which buffer lines are in the viewport, so document-position-aware features can't be implemented correctly.

Since these values are already available in `Client::redraw_ifn()`, adding them to the `draw` message would be a zero-cost protocol change (same pattern as `widget_columns` in #5455). @enricozb, would it make sense to split out these protocol fields as a standalone first step that your scrollbar rendering can then build on?

Yus314:
I've submitted #5471 which extracts the viewport information subset from this PR as an independent protocol extension — adding `first_buffer_line`, `last_buffer_line`, and `buffer_line_count` to the JSON UI `draw` message.

This is the approach I described in my [earlier comment](https://github.com/mawww/kakoune/pull/5304#issuecomment-2705870508): splitting the protocol change from the scroll bar UI, so alternative frontends can benefit from viewport data regardless of whether a built-in scroll bar is merged.

PR #5306, opened 2025-04-09
Fixed latex highlighter escaping math mode

ericv8v9s:
Previously, using any macro in math mode causes the whole thing to fail to be recognized as math because the macro would be captured by the "cs" region.
For example, this segment would be highlighted incorrectly with `$x \in A$` not highlighted as math and the word `Finally` after it highlighted as math.

```
First $A$ ...
Then $x \in A$ ...
Finally $B$ ...
```
screenshot_2025-04-09T03:31:24


ericv8v9s:
This does mean macros previously highlighted by cs region no longer get their highlights in math but that's something I can live with...

PR #5308, opened 2025-04-11
rc filetype rust: hl return as keyword

vbauerster:
That line makes return keyword to highlight same color as macro plus bold which makes it look like a macro. I'm not sure if it was done on purpose but it's wrong. Return is the same keyword like any other so why make it look apart?


vbauerster:
cc @dead10ck

PR #5313, opened 2025-04-15
Disable keymaps when playing macro

### LINKED ISSUE #5309: [BUG] Playing a macro resolves keyboard mappings recursively, which is incorrect
### Version of Kakoune

Kakoune v2024.05.18-92260722 (commit 922607221b8e5b2683babb9ebeba9372e97b836c)

### Reproducer

1. Open kakoune without any configuration, like with `kak -n`.

2. Insert any text so that the buffer isn't empty, and place the cursor/main selection at the end of the text, for example:
```
hello worl[d]
```
where `[d]` is a `d` with the cursor on it.

3. Run the following commands:
    1. `:map global normal h i`
    2. `:map global normal f h`
    
    Now we can move to the left by pressing `f` and enter insert mode by pressing `h` (sample of my real keyboard mappings)

4. Press `Q` to start recording a macro.

5. Move to the left 3 times by pressing `fff`.

6. Stop recording the macro with `Q`.

7. Put the cursor at the end of the line again like in step 2.

8. Press `q`.

 

### Outcome

The buffer now contains
```
hello worlhhd
```
despite not having inserted any text during the recording of the macro. It seems like it recursively resolved the mapping from `h` to `i` in normal mode, and then just inserted `h` as there are no mappings for it in insert mode.

### Expectations

The buffer should still contain the original text and the cursor should be on the `o`
```
hello w[o]rld
```

### Additional information

- OS: NixOS
- Running kakoune from commit 922607221b8e5b2683babb9ebeba9372e97b836c (I update to master once in a while)
- This bug has existed for a very long time, and only now did I decide to reproduce it in a minimal example.

SeerLite:
Macros are already recorded with keymaps resolved, so there's no need to resolve them again.

Fixes #5309.

### DIFF
--- a/src/normal.cc
+++ b/src/normal.cc
@@ -1610,6 +1610,7 @@ void replay_macro(Context& context, NormalParams params)
     auto keys = parse_keys(reg_val[0]);
     ScopedEdition edition(context);
     ScopedSelectionEdition selection_edition{context};
+    ScopedSetBool disable_keymaps(context.keymaps_disabled());
     do
     {
         for (auto& key : keys)


mawww:
It would be great to also have a regression test added there so that we do not see this issue pop up again.

PR #5317, opened 2025-04-19
Add short key names

sh4r1k7:
Minor addition, this was driving me nuts.

I like that we have `ret, esc, lt, gt`, but then being forced to use `space, backspace, pageup` instead of `spc, bspc, pgup, pgdn` is mental torture and makes for unnecessarily more verbose scripts.


Delapouite:
For me the "mental torture" has always been to endure weird and arbitrary abbreviations that our industry mindshare has inherited from the 70's and ended considering normal in a Stockholm syndrome-like experience.

A few random examples: `pwd` vs `passwd`, `glob` (global what?), `int`, for `interrupt`, or `integer`, `if`config (for interface where we already `if` for conditionals, `str` for both string or struct…

My point is, there's no need to keep going this trend and favor ambiguity just to save a few characters.

sh4r1k7:
Your point is moot, not only because my PR doesn't take anything away from you (you're free to continue writing `<backspace>` as many times as it takes to fix the industry), but also because you're as wrong about my motivations as you are about the definitions of "weird", "arbitrary", and "ambiguity" - `str` for `struct` is the only example that I might agree is ambiguous, but it's certainly not weird or arbitrary (and I've never seen it, nor has google, so it's not something "considered normal" either).

Like it or not, `kak` - I mean k-a-k-o-u-n-e (sorry for the ambiguity) - already uses abbreviations for half the key names and they're not going anywhere (nor is anyone confused by them), so I'm simply making the other half consistent.

I sure hope consistency is still something we can all agree on, although it is inherited all the way back from the 60s so...

sh4r1k7:
> I love shorthands
> I agree with @Delapouite 😬

one of those isn't true.

evnoj:
Just chiming in to say that the current incongruity doesn't gel with me. I still sometimes type `<spc>` because I'm conditioned to use shorthand by the very common `<ret>` and `<esc>`.

PR #5348, opened 2025-06-24
feat(terminal): support native terminal cursor rendering

Yukaii:
Make the cursor trail animation work in the Kitty terminal. The Ghostty terminal is also adding support for cursor shaders.

### Demo

```
set-option -add global ui_options terminal_cursor_native=true
```

https://github.com/user-attachments/assets/de84821f-b77a-4cbc-852e-a4b9592292c6


src/highlighters.cc
```diff
@@ -1290,11 +1291,20 @@ void highlight_selections(HighlightContext context, DisplayBuffer& display_buffe
         highlight_range(display_buffer, begin, end, false,
                         apply_face(sel_faces[primary ? 0 : 1]));
     }
+    // Check if we should skip drawing the primary cursor (for terminal cursor native mode)
+    const bool skip_primary_cursor = TerminalUI::has_instance() &&
+                                     TerminalUI::instance().is_cursor_native();
```
mawww: I wonder about this, I dont like highlighters knowing anything about TerminalUI, I think we could just let users override the PrimaryCursor face in their config to default,default to get the same outcome (and keep the door open for more effects based of primary cursor face).

Yukaii: fixed!

Yukaii: oops not fixed. I'll look into other impelmentation later 😂

src/terminal_ui.cc
```diff
@@ -538,10 +538,30 @@ void TerminalUI::redraw(bool force)
...
         format_with(writer, "\033[{};{}H", (int)c.line + 1, (int)c.column + 1);
     };
+
+    DisplayCoord target_pos;
     if (m_cursor.mode == CursorMode::Prompt)
-        set_cursor_pos({m_status_on_top ? 0 : m_dimensions.line, m_cursor.coord.column});
+        target_pos = {m_status_on_top ? 0 : m_dimensions.line, m_cursor.coord.column};
     else
-        set_cursor_pos(m_cursor.coord + content_line_offset());
+        target_pos = m_cursor.coord + content_line_offset();
```
mawww: Given that both branches start with `set_cursor_pos(target_pos)` , is there any reason to change this code in the first place ?

src/terminal_ui.cc
```diff
@@ -538,10 +538,30 @@ void TerminalUI::redraw(bool force)
...
+
+    if (m_terminal_cursor_native)
+    {
+        // Always position cursor in native mode to ensure it's correct after screen updates
+        set_cursor_pos(target_pos);
+
+        // Only send show cursor command when cursor changes or on force refresh
+        if (m_cursor.mode != m_prev_cursor.mode || m_cursor.coord != m_prev_cursor.coord || force)
+        {
+            format_with(writer, "\033[?25h"); // ensure cursor is visible
```
mawww: Why do we need to make the cursor visible here ? Cant we react directly in set_ui_options ? We already do this whenever m_terminal_synchronised changes.

src/terminal_ui.cc
```diff
@@ -1500,7 +1524,13 @@ void TerminalUI::restore_terminal()
 {
     write(STDOUT_FILENO,
         "\033>"
-        "\033[?25h"
+    );
+
+    // Only restore cursor visibility if it was hidden (non-native mode)
+    if (not instance().m_terminal_cursor_native)
+        write(STDOUT_FILENO, "\033[?25h");
```
mawww: I am not sure it is that important to avoid doing a redundant `"\033[?25h"` when we have not hidden the cursor, so I dont think we really need this to change

src/terminal_ui.hh
```diff
@@ -61,6 +61,8 @@ public:
     static void restore_terminal();
 
     void suspend();
+    
+    bool is_cursor_native() const { return m_terminal_cursor_native; }
```
mawww: I believe this is unused now

src/terminal_ui.hh
```diff
@@ -134,6 +136,8 @@ private:
         CursorMode mode;
         DisplayCoord coord;
     } m_cursor;
+    
+    struct Cursor m_prev_cursor;
```
mawww: If we move the cursor show/hide logic to set_ui_options I think we should not need this member variable anymore

Yukaii:
Have removed all the unnecessary changes. Thanks for the quick review!

src/terminal_ui.cc
```diff
@@ -1575,6 +1575,9 @@ void TerminalUI::set_ui_options(const Options& options)
     m_padding_char = find("terminal_padding_char").map([](StringView s) { return s.column_length() < 1 ? ' ' : s[0_char]; }).value_or(Codepoint{'~'});
     m_padding_fill = find("terminal_padding_fill").map(to_bool).value_or(false);
 
+    // Emit cursor show sequence
+    write(STDOUT_FILENO, "\033[?25h");
```
mawww: I did not mean to always enable it, I think we still want to have the terminal_native_cursor option

Yukaii: Ok! I've reverted that removal commit.

src/terminal_ui.cc
```diff
@@ -1574,9 +1574,14 @@ void TerminalUI::set_ui_options(const Options& options)
...
     m_padding_fill = find("terminal_padding_fill").map(to_bool).value_or(false);
-
-    // Emit cursor show sequence
-    write(STDOUT_FILENO, "\033[?25h");
+    
+    bool new_cursor_native = find("terminal_cursor_native").map(to_bool).value_or(false);
+    if (new_cursor_native != m_terminal_cursor_native)
+    {
+        m_terminal_cursor_native = new_cursor_native;
+        // Emit cursor visibility command when the option changes
```
mawww: that comment does not add much value, it seems to just describe what the code obviously does.

mawww (changes requested):
Could you remove the comment and update the doc/pages/options.asciidoc file as well to document this new ui_option ?

src/terminal_ui.hh
```diff
@@ -167,6 +167,7 @@ private:
 
     Codepoint m_padding_char = '~';
     bool m_padding_fill = false;
+    bool m_terminal_cursor_native = false;
```
krobelus: can probably remove the "terminal_" part from the variable name

src/terminal_ui.cc
```diff
@@ -1574,6 +1574,13 @@ void TerminalUI::set_ui_options(const Options& options)
 
     m_padding_char = find("terminal_padding_char").map([](StringView s) { return s.column_length() < 1 ? ' ' : s[0_char]; }).value_or(Codepoint{'~'});
     m_padding_fill = find("terminal_padding_fill").map(to_bool).value_or(false);
+    
+    bool new_cursor_native = find("terminal_cursor_native").map(to_bool).value_or(false);
+    if (new_cursor_native != m_terminal_cursor_native)
+    {
+        m_terminal_cursor_native = new_cursor_native;
+        write(STDOUT_FILENO, m_terminal_cursor_native ? "\033[?25h" : "\033[?25l");
```
krobelus: I guess this means that the `PrimaryCursor` face will be overridden by the color used by the terminal emulator,
and that cursor shapes (blinking etc) will be visible.

On a related note, I like that Neovim changes the cursor to a vertical line while in insert mode; Maybe Kakoune can achieve this even with its software cursor.

mawww: PrimaryCursor will still be applied, then on top of it the native cursor effect from the terminal, I suspect most users will want to override PrimaryCursor to default,default.

src/terminal_ui.hh
Yukaii: got it 👍🏼

PR #5353, opened 2025-07-04
Introduce `history_since_<id>` value expansion

### LINKED ISSUE #5345: [REQUEST] `HistoryCommit` hook for incremental updates of buffers in external tools.
### Feature

I'd like to see a `HistoryCommit` hook that's emitted whenever a new `HistoryNode` is appended to a buffer's history. The hook's parameter should be some combination of the new `history_id`, the time since epoch, and the parent `history_id`. The actual modifications can be read from the `uncommitted_modifications` value expansion. I doubt that embedding the modifications in the hook's parameter is feasible.

### Usecase

I want to have an external process that can efficiently track some ranges in kakoune's buffer. My main motivation is maintaining an optimized mapping from line-column coordinates to byte-offsets into the buffer to satisfy the `ts_tree_edit` API for more efficient highlighting in tree-sitter.

pjungkamp:
# Description

The `history_since_<id>` value expansion allows incremental parsing of a buffer's history.

Resolves #5345

# Example

```kak
declare-option int my_last_history_id
define-command my-process-history ...

# process the initial buffer history
my-process-history %val{bufname} 0 %val{history}
set-option buffer my_last_history_id 0

# only process new history changes on idle
hook buffer NormalIdle %{
  evaluate-commands %exp{
    my-process-history \
      %%val{bufname} \
      %%opt{my_last_history_id} \
      %%val{history_since_%opt{my_last_history_id}}
  }
  set-option buffer my_last_history_id %val{history_id}
}
```

### DIFF
--- a/doc/pages/expansions.asciidoc
+++ b/doc/pages/expansions.asciidoc
@@ -305,6 +305,11 @@ The following expansions are supported (with required context _in italics_):
     history), and each _modification_ is presented as in
     `%val{uncommitted_modifications}`.
 
+*%val{history_since_id}*::
+    _in buffer, window scope_ +
+    a partial history of the buffer in the same format as `%val{history}`
+    starting after entry _id_
+
 *%val{history_id}*::
     _in buffer, window scope_ +
     history id of the current buffer, an integer value which refers to a
--- a/src/buffer_utils.cc
+++ b/src/buffer_utils.cc
@@ -392,7 +392,7 @@ static String modification_as_string(const Buffer::Modification& modification)
                   modification.content->strview());
 }
 
-Vector<String> history_as_strings(const Vector<Buffer::HistoryNode>& history)
+Vector<String> history_as_strings(ConstArrayView<Buffer::HistoryNode> history)
 {
     Vector<String> res;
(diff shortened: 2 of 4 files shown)


pjungkamp:
I think there are some missing template deduction guides for `ConstArrayView` that mirror those of `ArrayView`.

```cpp
template <typename Container>
ConstArrayView(const Container& c) -> ConstArrayView<decltype(*c.data())>;
```

I was trying to get a `ConstArrayView` for the buffer history by calling `ConstArrayView(context.buffer().history())`, but this actually calls the single element constructor for `ConstArrayView` instead of the container constructor.

mawww:
This looks big enough to warrant a copyright waiver, and I dont think I have one from you, can you add a copyright waiver commit to this PR (see the CONTRIBUTING file)

pjungkamp:
You've already got one! (From my `EnterDirectory` hook PR) 

https://github.com/mawww/kakoune/commit/07000adb9bf659a879b8c272426e5c5201a7b0c1

mawww:
Indeed, sorry about that.

PR #5360, opened 2025-07-23
rc/windowing/wayland.kak: add kgx support

thacuber2a03:
...[this](https://gitlab.gnome.org/GNOME/console) thing.
why did they switch consoles like that all of a sudden, and why does it have even *less* features, what the hell????

### DIFF
--- a/rc/windowing/wayland.kak
+++ b/rc/windowing/wayland.kak
@@ -18,6 +18,7 @@ A shell command is appended to the one set in this option at runtime} \
                    'termite        -e      ' \
                    'wterm          -e sh -c' \
                    'gnome-terminal -e      ' \
+                   'kgx            -e      ' \
                    'xfce4-terminal -e      ' \
                    'konsole        -e      '; do
         terminal=${termcmd%% *}


Delapouite:
It looks like there also exists a third one called `Ptyxis`? https://discourse.gnome.org/t/why-two-terminals/2279

rc/windowing/wayland.kak
```diff
@@ -18,6 +18,7 @@ A shell command is appended to the one set in this option at runtime} \
                    'termite        -e      ' \
                    'wterm          -e sh -c' \
                    'gnome-terminal -e      ' \
+                   'kgx            -e      ' \
```
krobelus, line 21: kgx has a [long-standing issue](https://gitlab.gnome.org/GNOME/console/-/issues/135) that means that if we do

     kgx -e kak -c my-session

and close the client with `:quit`, the terminal window will stick around ("Read-only, process exited").
`ptyxis` doesn't seem to have that issue, so maybe we should prefer that.
(though adding kgx is fine I guess)

It is kinda minor (you could argue that using the OS's "close window" shortcut over `:q` is mostly safe) but still, weird

thacuber2a03: I was aware of that issue, but I also saw it as minor and thought it had some workaround... but *3 years? damn*

anyways, about `ptyxis`, I guess that would be a better idea since Fedora packs with it ~~and it has more features than `kgx` does~~, but still, having `kgx` here wouldn't hurt anyone, it would still be convenient (I suppose)

I could always add `ptyxis` as well, mostly for the same convenience reason as w/`kgx`

krobelus:
On Wed, Jul 23, 2025 at 12:01:16AM -0700, ThaCuber wrote:
> @thacuber2a03 commented on this pull request.
> 
> 
> 
> > @@ -18,6 +18,7 @@ A shell command is appended to the one set in this option at runtime} \
>                     'termite        -e      ' \
>                     'wterm          -e sh -c' \
>                     'gnome-terminal -e      ' \
> +                   'kgx            -e      ' \
> 
> I was aware of that issue, but I also saw it as minor, and I thought it had some workaround, but *3 years?* that's unheard of

yeah we haven't found a workaround
except for making whatever program is running
in the terminal the login shell..
but not via chsh but via dconf.. which makes the workaround evenn worse
https://github.com/fish-shell/fish-shell/issues/11509

> 
> anyways, about ptyxis, I guess that would be a better idea since Fedora packs with it ~~and it has more features than kgx does~~, but still, having kgx here wouldn't hurt anyone, it would still be convenient (I suppose)

yeah kgx should probably be added because it seems to be the new default for gnome (I wonder why)

thacuber2a03:
@krobelus your comment got really badly formatted

Issue #4841, opened 2023-02-12
[QUESTION] `ModuleLoaded` behaviour


glaeqen:
Is it sane for `ModuleLoaded` hook not to be triggered if it was defined after `require-module` call?

```
provide-module a %{
    hook global ModuleLoaded b %{}
}

provide-module b %{}

require-module b

require-module a # hook will not ever be triggered!
```

As a result, order of lazy-evaluation matters. Is there any rationale behind it? I was expecting the hook to be instantly triggered as soon as I `require-module a`.

mawww:
This is an interesting corner case, I think it would make sense for ModuleLoaded hooks to immediately execute if the module has already been loaded. Any opinions ?

krobelus:
If we didn't expose the `windowing_modules` option this would allow to remove the `KakBegin` hook from windowing/detection.kak.
The suggested behavior makes sense. It's useful for "mixin" style modules like `repl/x11.kak` which adds functionality on top of `x11.kak`.

It would also help kak-lsp get rid of this workaround:

```kak
try %{
    add-highlighter shared/kakrc/code/lsp_keywords regex \[(language\.[a-z_]+\.settings(?:\.[^\]\n]*)?)\] 1:title
} catch %{
    hook global -once ModuleLoaded kak %{
        add-highlighter shared/kakrc/code/lsp_keywords regex \[(language\.[a-z_]+\.settings(?:\.[^\]\n]*)?)\] 1:title
    }
}
```

Issue #4375, opened 2021-09-24
[REQUEST] A better way to prevent standard library plugins from loading


Screwtapello:
### Feature

Kakoune ships with a broad standard library, but Kakoune's maintainers are not experts in every tool and language with a stdlib plugin. As more people use Kakoune, it becomes more likely that the best plugin for integrating with tool X/editing language Y/communicating with service Z will be a plugin outside the standard library, and people will want to use that instead of whatever ships with Kakoune.

This is already possible with Kakoune - if you create a `%val{config}/autoload` directory, Kakoune will stop loading the standard library, and you can symlink in just the parts you want. However, there are a *lot* of plugins that ship with Kakoune, and Kakoune reserves the right to add more, rename them, reorganise the plugin directory structure, and so forth. This makes a symlink tree very easy to break, and very tedious to repair.

One alternative approach:

- at startup, make a list of relative paths to `.kak` files that exist inside `%val{config}`, and a list of relative paths to `.kak` files that exist inside `%val{runtime}`
- join the two lists together, to make three categories: files that exist only in `%val{config}`, files that exist only in `%val{runtime}`, and files that exist in both places.
- load all the files that exist only in `%val{config}`
- load all the files that exist only in `%val{runtime}`
- for files that exist in both, load the `%val{config}` version.

Then, if a user wants to replace a single file from the standard library, they create just that file in `%val{config}` and all other files will continue to load from `%val{runtime}`. I think this could reasonably be implemented with `join(1)`.

An additional wrinkle: currently people expect Kakoune to not automatically load any standard library files when `%val{config}/autoload` exists, so they have all the standard library files symlinked into the autoload directory (i.e. present, but with a different file-path than exists in `%val{runtime}`. If we just adopt the above scheme, everybody's going to get a bunch of "duplicate definition" errors. Instead, I propose:

- if `%val{config}/autoload` exists, only load it (for backwards compatibility)
- otherwise, use the above algorithm to compare `%val{config}/rc` and `%val{runtime}/rc`

Then, we tell people:

- if you want to load a third-party plugin in addition to the standard library, install it in `%val{config}/rc`
- if you want to prevent a specific standard library plugin from loading, `touch` a file with the appropriate name in `%val{config}/rc`
- if you want to skip all standard library plugins and only load your own plugins, create `%val{config}/autoload` and put the plugins in there.

alexherbo2:
Personally, I like my kakrc to be declarative.

[kakrc]:

``` kak
[...]

PR #5365, opened 2025-07-29
Wrap filetype detection hooks in module to allow order of evaluation specification with ModuleLoaded hook

pylasnier:
I propose wrapping filetype detection config code in a dedicated module, for now I've named `detect-<filetype>`, which gets immediately required after defining it, so as to preserve the normal autoload behaviour for the filetype rcs.

What this change does is enable people to manage conflicts between their own plugin or kakrc code, without having to modify or delete default rcs from their autoload, by forcing their code to run after the filetype detection codes using the `ModuleLoaded` hook. This specifies an order of execution, and is consistent regardless of the order of loading of files, as per the behaviour described in #4841.

For example, the SML filetype detection regex may be partially overridden by making a new `BufCreate` hook, which runs *after* the default SML detection:

``` kak
hook global ModuleLoaded detect-sml %{

hook global BufCreate .*Script\.sml %{
    set-option buffer filetype holscript
}

}
```

The changes I have included for now only address the issues from [this discussion](https://discuss.kakoune.com/t/how-do-you-easily-disable-the-default-init-file-for-a-filetype/2607), but I propose similar changes may be added to all filetype rcs, or possibly any eagerly-evaluated autoload code which might be helpful to specify order of evaluation for.

This change could help solve the following issues:
- #4375
- [https://discuss.kakoune.com/t/how-do-you-easily-disable-the-default-init-file-for-a-filetype/2607](https://discuss.kakoune.com/t/how-do-you-easily-disable-the-default-init-file-for-a-filetype/2607)
- [https://discuss.kakoune.com/t/how-can-i-make-pepe-janet-kak-properly-override-built-in-janet-kak/1889](https://discuss.kakoune.com/t/how-can-i-make-pepe-janet-kak-properly-override-built-in-janet-kak/1889)

### DIFF
--- a/rc/filetype/exherbo.kak
+++ b/rc/filetype/exherbo.kak
@@ -1,5 +1,7 @@
 provide-module detect-exherbo %{
 
+# exheres
+
 ## Repository metadata files
 hook global BufCreate .*/metadata/mirrors\.conf         %{ set-option buffer filetype paludis-mirrors-conf }
 hook global BufCreate .*/metadata/licence_groups.conf   %{ set-option buffer filetype exheres-0-licence-groups }
@@ -46,6 +48,45 @@ hook -group exheres-0-licence-groups-highlight global WinSetOption filetype=exhe
     hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/exheres-0-licence-groups }
 }
 
+
+# paludis
+
+hook global WinSetOption filetype=paludis-(key-value|options|mirrors|specs)-conf %{
+    require-module paludis
+}
+
+hook -group paludis-options-conf-highlight global WinSetOption filetype=paludis-options-conf %{
+    add-highlighter window/paludis-options-conf ref paludis-options-conf
+    hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/paludis-options-conf }
+}
+
+hook -group paludis-key-value-conf-highlight global WinSetOption filetype=paludis-key-value-conf %{
+    add-highlighter window/paludis-key-value-conf ref paludis-key-value-conf
+    hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/paludis-key-value-conf }
+}
+
+hook -group paludis-mirrors-conf-highlight global WinSetOption filetype=paludis-mirrors-conf %{
+    add-highlighter window/paludis-mirrors-conf ref paludis-mirrors-conf
+    hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/paludis-mirrors-conf }
+}
+
+hook -group paludis-specs-conf-highlight global WinSetOption filetype=paludis-specs-conf %{
+    add-highlighter window/paludis-specs-conf ref paludis-specs-conf
+    hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/paludis-specs-conf }
+}
+
+
+# glep42
+
+hook global WinSetOption filetype=glep42 %{
+    require-module glep42
+}
+
+hook -group glep42-highlight global WinSetOption filetype=glep42 %{
+    add-highlighter window/glep42 ref glep42
+    hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/glep42 }
+}
+
 }
 
 require-module detect-exherbo
@@ -74,29 +115,6 @@ add-highlighter shared/exheres-0-licence-groups/ regex ^(\S+) 0:type
 add-highlighter shared/exheres-0-licence-groups/ regex ^#.*?$ 0:comment
 }
 
-hook global WinSetOption filetype=paludis-(key-value|options|mirrors|specs)-conf %{
-    require-module paludis
-}
-
-hook -group paludis-options-conf-highlight global WinSetOption filetype=paludis-options-conf %{
-    add-highlighter window/paludis-options-conf ref paludis-options-conf
-    hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/paludis-options-conf }
-}
-
-hook -group paludis-key-value-conf-highlight global WinSetOption filetype=paludis-key-value-conf %{
-    add-highlighter window/paludis-key-value-conf ref paludis-key-value-conf
-    hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/paludis-key-value-conf }
-}
-
-hook -group paludis-mirrors-conf-highlight global WinSetOption filetype=paludis-mirrors-conf %{
-    add-highlighter window/paludis-mirrors-conf ref paludis-mirrors-conf
-    hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/paludis-mirrors-conf }
-}
-
-hook -group paludis-specs-conf-highlight global WinSetOption filetype=paludis-specs-conf %{
-    add-highlighter window/paludis-specs-conf ref paludis-specs-conf
-    hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/paludis-specs-conf }
-}
 provide-module paludis %{
 ## Paludis configurations
 ### options.conf
@@ -124,15 +142,6 @@ add-highlighter shared/paludis-specs-conf group
 add-highlighter shared/paludis-specs-conf/ regex [\s\t]+(\S+(?:[\s\t]+))*$ 0:attribute
 add-highlighter shared/paludis-specs-conf/ regex ^(\S+/\S+) 0:type
 add-highlighter shared/paludis-specs-conf/ regex ^#.*?$ 0:comment
-}
-
-hook global WinSetOption filetype=glep42 %{
-    require-module glep42
-}
-
-hook -group glep42-highlight global WinSetOption filetype=glep42 %{
-    add-highlighter window/glep42 ref glep42
-    hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/glep42 }
 }
 
 provide-module glep42 %{
--- a/rc/filetype/perf.kak
+++ b/rc/filetype/perf.kak
@@ -1,3 +1,24 @@
+provide-module detect-perf %{
+
+hook -group perf-report-highlight global WinSetOption filetype=perf-report %{
+    require-module perf-report
+    add-highlighter window/perf-report ref perf-report
+    hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/perf-report }
+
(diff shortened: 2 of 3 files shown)


krobelus:
I haven't had time to look at this in detail but 
this might be a workaround that goes in the wrong direction.

In general I'd either fix the stdlib rc file, or use a different filetype
that doesn't clash.  The stdlib files are supposed to work for almost everyone.

If your plugin disables it, that might break other use cases and plugins.

> I’m working on an extension for a proof assistant called HOL4, which uses
> script files written in SML, typically with the file format nameScript.sml

So the issue is that for your case, "*.sml" files should not have the `sml` filetype
but the `holscript` one?
Can we maybe detect these cases and build them into `sml.kak`?

Of course the user should always be in control;
it's probably possible to achieve your goal today 
by being careful about hook ordering:
if your `BufCreate` hook is added after the stdlib's,
then the final filetype will be `holscript` (
and the transient `sml` filetype shouldn't matter, right?)

pylasnier:
> So the issue is that for your case, "*.sml" files should not have the sml filetype
> but the holscript one?
> Can we maybe detect these cases and build them into sml.kak?

I would rather keep the HOL4-specific stuff separate because it is not commonly used compared to SML, and I would rather not surprise a kak user who happened to decide to call their file `somethingScript.sml`. I would prefer it to be an opt-in plugin, especially because full functionality of the plugin depends on a HOL4 installation.

> If your plugin disables it, that might break other use cases and plugins.

With this version of config, the HOL4 plugin doesn't disable SML detection, but merely takes precedence over it by setting the regex hook after. It should not break other use-cases, so long as the user does not name their file `somethingScript.sml` (which, by having the plugin, they have opted-in to).

As far as I can tell, this proposal of wrapping with a module should not impact default behaviour in any way. It simply sets a checkpoint which plugin or user code can wait for to allow their code to run after, which should be a helpful feature. This is the best solution I could come up with that uses existing syntax.

> Of course the user should always be in control;
> it's probably possible to achieve your goal today
> by being careful about hook ordering:
> if your BufCreate hook is added after the stdlib's,
> then the final filetype will be holscript (
> and the transient sml filetype shouldn't matter, right?)

This is exactly what my example does, it just uses the module wrapping and `ModuleLoaded` hook to achieve that, rather than modifying stdlib.

The point is that it is difficult to be careful about hook ordering. Many of these issues arise from the fact that stdlib is autoloaded *after* people's custom hooks and overrides them. I expect an idiomatic solution would be the allow the user to specify an order of execution, in spirit of the lazy, module-based evaluation that does not require file-level changes.

pylasnier:
I have added similar `detect-filetype` modules to all filetypes, so `ModuleLoaded` hooks can be used to specify order of evaluation of detection and initialisation hooks for any filetype.

I really think a solution like this would be a useful addition to the rcs. Even if it's not this particular solution and its something else, I think it would be a step in the right direction to have some sort of feature which allows plugins to resolve conflicts with the stdlib without requiring the user to modify their default autoload. Perhaps some change to hooks.

rc/filetype/apl.kak
```diff
@@ -1,3 +1,4 @@
+provide-module detect-apl %@
```
krobelus: I wonder if we should change module names. With this PR, we have.

```kak
provide-module detect-apl %{
	hook global WinSetOption filetype=apl %{
	    require-module apl
	}
	...
}
provide-module apl %{
	add-highlighter shared/apl regions
	...
}
```

It's possible that we want to include the "filetype" path component,
to prevent future clashes.
We could do this in a backwards-compatible way
by also providing an empty module with the old name.

```kak
provide-module filetype-apl-detection %{
	hook global WinSetOption filetype=apl %{
	    require-module filetype-apl
	}
	...
}
provide-module filetype-apl %{
	provide-module apl %{} # historical
	add-highlighter shared/apl regions
	...
}
```

Not yet sure if this is worth it, but it seems like now is a good time to think about this.
There's no existing discussion in https://github.com/mawww/kakoune/pull/2782.

I wonder what are the odds that our modules (like `jump`, `make`, `tmux`, `screen`) will ever collide with filetype modules of the same name..

rc/filetype/apl.kak
```diff
@@ -26,6 +27,10 @@ hook -group apl-highlight global WinSetOption filetype=apl %{
     hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/apl }
 }
 
+@
```
krobelus: (mega nit, feel free to ignore) I think in this case we can use `%< >` to help the `m` command

rc/filetype/exherbo.kak
```diff
@@ -44,6 +46,10 @@ hook -group exheres-0-licence-groups-highlight global WinSetOption filetype=exhe
     hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/exheres-0-licence-groups }
 }
 
+}
+
+require-module detect-exherbo
+
 provide-module exheres %{
```
krobelus, line 94: huh, so this is an odd module where the filename (and thus the "detect-exherbo" hook)
don't match the actual filetype modules (exheres, paludis etc.).
I guess if we really qualify module names, this would imply
module names like `filetype-exherbo-exheres` and `filetype-exherbo-paludis`.
Moving all into a single module would probably work too, but I guess it's nice to allow multiple modules per file.

rc/filetype/sml.kak
```diff
@@ -22,6 +24,10 @@ hook -group sml-highlight global WinSetOption filetype=sml %{
     hook -once -always window WinSetOption filetype=.* %{ remove-highlighter window/sml }
 }
 
+}
+
+require-module detect-sml
```
krobelus, line 29: Thanks for this, this seems like the ideal solution to define custom
filetype detection hooks *after* the ones defined by the stdlib,
which is usually the right thing to do (as opposed to completely removing the stdlib one).
I agree that hook groups wouldn't help here.

Your [last comment](https://discuss.kakoune.com/t/how-do-you-easily-disable-the-default-init-file-for-a-filetype/2607/11) on Discourse was also helpful.

I guess this means that all scripts in "rc/filetype/*.kak" put all executable code into the "detect-$filetype" and "$filetype" modules?

Two possible future directions:
- for people who want to completely override a "$filetype" module, we
  can add `provide-module -override` as suggested in https://github.com/mawww/kakoune/issues/4375#issuecomment-927396805
- for people who want to (also) override the filetype detection hooks,
  that is, either of these:
  - `hook global BufCreate .*\.(sml|fun|sig)`
  - `hook global WinSetOption filetype=sml`
  - `hook -group sml-highlight global WinSetOption filetype=sml`
  we could
  - either add hook groups to the first two, so they can use the familiar `remove-hooks` and `set-option global disabled_hooks`
  - or
    - move the "require-module detect-$foo" hooks into a "hook -once global KakBegin .*".
    - then they can use `provide-module -override detect-sml %{}`

rc/filetype/apl.kak
pylasnier: Yep angle brackets work, I seem to have missed that on my first pass

rc/filetype/exherbo.kak
pylasnier: Oops, it indeed seems like I missed some hooks scattered around in files, I think I have cleaned everything up now.

I have opted to go with one detection module per file,
1. for consistency; some files (for example [git.kak](https://github.com/mawww/kakoune/blob/27d5949d63505f5462d94225562a008c7de53c8d/rc/filetype/git.kak)) have I think too many filetypes to want to split apart, so if those stay grouped then others should stay grouped
2. I believe users are unlikely to need these modules below the granularity of a file, since they exist only to specify an order of execution. They would simply want to state 'run the hooks from this file first', and I expect further specificity of hook ordering is redundant.
3. IMO slightly easier discovery, it's just the name of the kak file

rc/filetype/apl.kak
pylasnier: I would be happy to use different names for the detection modules if you'd like. I would probably need to check but I'm not entirely sure if your suggested filetype module name change, though, wouldn't be breaking? The `apl` module would be kept hidden until `filetype-apl` is required?

Perhaps we could open an issue on the topic, I would personally consider it a separate change.

rc/filetype/sml.kak
pylasnier: > I guess this means that all scripts in "rc/filetype/*.kak" put all executable code into the "detect-$filetype" and "$filetype" modules?

Almost everything, there are some hidden commands (like indentation commands) which I thought were probably redundant to wrap up. All hooks are in modules now.

mawww:
Hi, coming late to this discussion, I am still unclear on how this is supposed to work, as far as I understand the only difference this introduces is that we will get a `ModuleLoaded` hook triggered when sourcing those filetype support files.

`ModuleLoaded` triggers immediately when we load those autoload files, but we do have special code so that adding a `ModuleLoaded` hook will trigger immediately if that module has already been loaded, which means that all `ModuleLoaded` hooks taking advantage of this in user configuration are equivalent to just running the commands they contain directly.

in other words, AFAIU, the only difference between:

```
hook global ModuleLoaded detect-sml %{
hook global BufCreate .*Script\.sml %{ set-option buffer filetype holscript }
}
```

and 

```
hook global BufCreate .*Script\.sml %{  set-option buffer filetype holscript }
```

In the user kakrc, is that the former will only add the hook if we have sourced sml.kak while the latter will always add that hook.

Is my understanding correct ? Is that what you are trying to achieve with this change ?

pylasnier:
The issue is that, as is commonly the case, `hook global BufCreate .*Script\.sml %{  set-option buffer filetype holscript }` gets run *before* the stdlib rcs from the autoload, so the hook gets shadowed. This change allows users to force their code to run *after* any particular stdlib filetype rc, so that their hooks can't get shadowed.

mawww:
Hum, do you mean then that this hook is inside another autoload file ? Because if it is in the user kakrc it should run after.

The existing workaround for this is to use the `KakBegin` hook which should run after all autoload and user kakrc have been sourced. It is not a very elegant solution but I am not sure the problem you are trying to solve here is common enough to warrant the extra confusion introduced by the "immediately loaded module" pattern.

pylasnier:
Oh I see, that could work to solve my issue. I hadn't come across that workaround before.

This change could still offer solutions that the `KakBegin` hook workaround can't, for example if the order of evaluation issue comes up twice in a chain. And I would argue that, because this change doesn't change existing behaviour, it might still be valuable for the sake of including more stdlib code into patterns of lazy evaluation. But if you think it would be confusing that's reasonable, since it's a new pattern.

Thank you for the workaround regardless!

PR #5368, opened 2025-08-02
add -ruler switch to column highlighter

enricozb:
the `-ruler <character>` switch on the column highlighter replaces the column (for empty and whitespace cells) with `<character>`. this is useful for drawing rulers. because rulers typically appear over whitespace regions in other editors, we also do that here.

additionally, the color of the ruler should only be applied to the ruler, so when -ruler `<character>` is provided, we do not apply the `<face>` parameter to cells which are non-empty, aka where the ruler isn't inserted.

However, in order to ensure the ruler has the same color even in regions that are already highlighted, we must use a `+f` attribute on the face provided to the highlighter. Perhaps that attribute should be automatically added when passing the `-ruler` switch, I'm not sure.

For example, after
```
add-highlighter buffer/ column -ruler ┊ 80 rgb:444444+f
```

image

I'm also open to making this a separate `ruler` highlighter instead of hijacking the column highligher. Please let me know if there's any interest in this change, thank you in advance.

### DIFF
--- a/src/highlighters.cc
+++ b/src/highlighters.cc
@@ -19,6 +19,7 @@
 #include "utf8.hh"
 #include "utf8_iterator.hh"
 #include "window.hh"
+#include "unicode.hh"
 
 #include <cstdio>
 #include <limits>
@@ -519,16 +520,26 @@ UniquePtr<Highlighter> create_line_highlighter(HighlighterParameters params, Hig
 }
 
 const HighlighterDesc column_desc = {
-    "Parameters: <value string> <face>\n"
-    "Highlight the column given by evaluating <value string> with <face>",
-    {}
+    "Parameters: [-ruler <character>] <column> <face>\n"
+    "Highlight the column <column> with <face>",
+    {
+        { { "ruler", { ArgCompleter{}, "replace empty or whitespace cells with the given character. When provided, <face> is not applied to non-empty cells" } } },
+        ParameterDesc::Flags::SwitchesOnlyAtStart, 2, 2
+    },
 };
 UniquePtr<Highlighter> create_column_highlighter(HighlighterParameters params, Highlighter*)
 {
-    if (params.size() != 2)
-        throw runtime_error("wrong parameter count");
+    ParametersParser parser{params, column_desc.params};
 
-    auto func = [col_expr=params[0], facespec=parse_face(params[1])]
+    auto positionals = parser.positionals_from(0);
+    auto ruler = parser.get_switch("ruler");
+    if (ruler.value_or(" ").char_length() > 1)
+        throw runtime_error("-ruler expects a single character");
+
+    auto ruler_provided = static_cast<bool>(ruler);
+    auto col_char = ruler.value_or(" ").str();
+
+    auto func = [col_expr=positionals[0], facespec=parse_face(positionals[1]), ruler_provided, col_char]
                 (HighlightContext context, DisplayBuffer& display_buffer, BufferRange)
     {
         ColumnCount column = -1;
@@ -563,7 +574,12 @@ UniquePtr<Highlighter> create_column_highlighter(HighlighterParameters params, H
                         atom_it = ++line.split(atom_it, remaining_col);
                     if (atom_it->length() > 1)
                         atom_it = line.split(atom_it, 1_col);
-                    atom_it->face = merge_faces(atom_it->face, face);
+                    if (!ruler_provided)
+                        atom_it->face = merge_faces(atom_it->face, face);
+                    if (ruler_provided && is_blank(atom_it->content()[0])) {
+                        atom_it->replace(col_char);
+                        atom_it->face = merge_faces(atom_it->face, face);
+                    }
                     found = true;
                     break;
                 }
@@ -574,7 +590,7 @@ UniquePtr<Highlighter> create_column_highlighter(HighlighterParameters params, H
 
             if (remaining_col > 0)
                 line.push_back({String{' ', remaining_col}, Face{}});
-            line.push_back({" ", face});
+            line.push_back({col_char, face});
         }
     };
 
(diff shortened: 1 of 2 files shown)


src/highlighters.cc
```diff
@@ -519,16 +520,26 @@ UniquePtr<Highlighter> create_line_highlighter(HighlighterParameters params, Hig
...
 
 const HighlighterDesc column_desc = {
-    "Parameters: <value string> <face>\n"
-    "Highlight the column given by evaluating <value string> with <face>",
-    {}
+    "Parameters: [-ruler <character>] <column> <face>\n"
+    "Highlight the column <column> with <face>",
+    {
+        { { "ruler", { ArgCompleter{}, "replace empty or whitespace cells with the given character. When provided, <face> is not applied to non-empty cells" } } },
+        ParameterDesc::Flags::SwitchesOnlyAtStart, 2, 2
```
mawww: Is there any reason here for SwitchesOnlyAtStart ? It seems to me we could accept switches anywhere for this command.

src/highlighters.cc
```diff
@@ -519,16 +520,26 @@ UniquePtr<Highlighter> create_line_highlighter(HighlighterParameters params, Hig
...
-        throw runtime_error("wrong parameter count");
+    ParametersParser parser{params, column_desc.params};
 
-    auto func = [col_expr=params[0], facespec=parse_face(params[1])]
+    auto positionals = parser.positionals_from(0);
+    auto ruler = parser.get_switch("ruler");
+    if (ruler.value_or(" ").char_length() > 1)
+        throw runtime_error("-ruler expects a single character");
+
+    auto ruler_provided = static_cast<bool>(ruler);
```
mawww: I'd prefer `ruler.has_value()` here, and I think we could rename "ruler_provided" with what this bool is used for: highlight_non_blank, so essentially to `bool highlight_non_blank = not ruler.has_value()`

src/highlighters.cc
```diff
@@ -519,16 +520,26 @@ UniquePtr<Highlighter> create_line_highlighter(HighlighterParameters params, Hig
...
+    auto col_char = ruler.value_or(" ").str();
```
mawww, line 538: Maybe do that before the char_length testing and test `col_char.char_length()` ?

src/highlighters.cc
```diff
@@ -563,7 +574,12 @@ UniquePtr<Highlighter> create_column_highlighter(HighlighterParameters params, H
                         atom_it = ++line.split(atom_it, remaining_col);
                     if (atom_it->length() > 1)
                         atom_it = line.split(atom_it, 1_col);
-                    atom_it->face = merge_faces(atom_it->face, face);
+                    if (!ruler_provided)
+                        atom_it->face = merge_faces(atom_it->face, face);
+                    if (ruler_provided && is_blank(atom_it->content()[0])) {
```
mawww: Kakoune codebase uses the alternate spelling for boolean operator, `and` here rather than `&&`, and curly brace on its own line. Moreover you probably can make this an `else if` to avoid testing `ruler_provided` again.

mawww (changes requested):
While I wonder if there is strong demand for this, the implementation is simple enough so that I am happy to merge it provided a few (mostly code style) tweaks are done.

src/highlighters.cc
enricozb: Hmm, `has_value` doesn't seem defined for `Optional`, only for `HashMap`.

enricozb:
Thanks for taking a look, much appreciated. I've addressed the comments, except for the one on `has_value`, as `Optional` doesn't look like it has a `has_value`. Let me know what you'd like done there. Thanks!

src/highlighters.cc
krobelus: `std::optional::has_value()` is C++ canon so it would be fine to add,
The reason why we don't have it yet is: when we want to call `explicit operator bool`, we basically always use C-style casts (`(bool)someexpr`) instead of `static_cast`. Maybe keep things consistent and make a sweeping change later if we really want that

enricozb: @krobelus so your suggestion here is to just do `(bool)ruler`, right?

krobelus: yes, or if we actually want the negation, `not ruler` should work since contextual conversion will happily call explicit operators

enricozb: done :)

enricozb:
I switched to `(bool)` instead of `static_cast` following the discussion with @krobelus

src/highlighters.cc
```diff
@@ -563,7 +574,13 @@ UniquePtr<Highlighter> create_column_highlighter(HighlighterParameters params, H
                         atom_it = ++line.split(atom_it, remaining_col);
                     if (atom_it->length() > 1)
                         atom_it = line.split(atom_it, 1_col);
-                    atom_it->face = merge_faces(atom_it->face, face);
+                    if (!highlight_non_blank)
```
mawww: I think the negation is incorrect here, but it works because of line 537 being wrong as well.

src/highlighters.cc
```diff
@@ -519,16 +520,26 @@ UniquePtr<Highlighter> create_line_highlighter(HighlighterParameters params, Hig
...
 {
-    if (params.size() != 2)
-        throw runtime_error("wrong parameter count");
+    ParametersParser parser{params, column_desc.params};
+
+    auto positionals = parser.positionals_from(0);
 
-    auto func = [col_expr=params[0], facespec=parse_face(params[1])]
+    auto ruler = parser.get_switch("ruler");
+    bool highlight_non_blank = (bool)ruler;
```
mawww: I think this should be initialized to `not (bool)ruler`, as my understand is that we highlight non-blank characters if we do **not** have a ruler.

enricozb:
Apologies, i misunderstood the intention behind the suggested naming `highlight_non_blank`; fixed this.

src/highlighters.cc
enricozb: Yep you're correct, fixed this.

PR #5370, opened 2025-08-09
Clarify command parsing with #

Kratacoa:
Not sure if this is the best way of specifying that, so corrections are welcome.

### DIFF
--- a/doc/pages/command-parsing.asciidoc
+++ b/doc/pages/command-parsing.asciidoc
@@ -9,6 +9,14 @@ prompt, are parsed according to the following rules:
 
 - Words (command names and parameters) are delimited by whitespaces.
 
+- <<expansions#, Expansions>> are parsed and replaced before the execution of
+  the commands.
+
+== Comments parsing
+
+Commands are commented out after `#` until end of line (not `;`).
+`#` takes effect _after_ the expansions are parsed.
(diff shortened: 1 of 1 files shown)


doc/pages/command-parsing.asciidoc
```diff
@@ -9,6 +9,14 @@ prompt, are parsed according to the following rules:
 
 - Words (command names and parameters) are delimited by whitespaces.
 
+- <<expansions#, Expansions>> are parsed and replaced before the execution of
+  the commands.
+
+== Comments parsing
+
+Commands are commented out after `#` until end of line (not `;`).
+`#` takes effect _after_ the expansions are parsed.
```
mawww, line 18: That sounds incorrect to me, `decl str foo 'foo # bar'; echo %opt{foo}` does echo `foo # bar`, or maybe you meant something different with "after the expansions are parsed"

Kratacoa: I meant that it's only executed, if it is, only after the expansions are parsed. But reading this again and after your observation, it's clearly poorly worded.

How about:
`The parsing of strings and expansions happens before `#` is evaluated as commenting out the rest of the line, hence its meaning might change according to command-parsing.`

PR #5371, opened 2025-08-11
Document blocking behavior of %sh expansions

Supreeeme:


Screwtapello:
Kakoune also waits for the shell to exit, no?

It seems a good idea to mention that Kakoune also waits for stdout and stderr. It might also be a good idea to mention that this is relevant if you try to launch a long-running process in the background. Something like:

> Kakoune waits, blocking user input, for the script to complete. If you want to launch a long-running process in the background, note that Kakoune *also* waits for the script's stdout and stderr to be closed, so your long-running process should also have its stdout and stderr redirected.

Supreeeme:
Good point, I reworded it to mention that.

doc/pages/expansions.asciidoc
```diff
@@ -101,6 +101,11 @@ Expansions with the type `sh` are executed as shell-scripts, and whatever
 the script prints to standard output replaces the expansion. For example,
 the command `echo %sh{date}` will echo the output of the `date` command.
 
+Kakoune will block user input until the script is complete. If you want to
```
ficd0, line 104: May be worth mentioning that Kakoune blocks _completely_, not just user input

doc/pages/expansions.asciidoc
```diff
@@ -101,6 +101,11 @@ Expansions with the type `sh` are executed as shell-scripts, and whatever
...
+run a process in the background, note that Kakoune will also wait for the
+script's stdout and stderr to be closed. You can redirect stdout and stderr
+of your long running background process to avoid blocking.
```
ficd0, line 107: Could be good to show an example of the POSIX pattern for this:

```sh
daemon-command >/dev/null 2>&1 </dev/null &
```

PR #5373, opened 2025-08-16
rc/filetype/wren.kak: fix invalid escape interpretation in raw strings

thacuber2a03:
I must've had a brainfart while writing this but raw strings interpret escapes when the website reads against it, it's literally the second sentence in the section about raw strings:

> **Raw strings do not process escapes and do not apply any interpolation.**
> \- https://wren.io/values.html#raw-strings

sorry for the small PR :/


thacuber2a03:
this is hilarious though
wren.io's highlighter failing at processing raw strings

Issue #4392, opened 2021-10-18
[REQUEST] Support JSON5 in `json.kak`


lenormf:
### Feature

[JSON5](https://spec.json5.org/) comes with several improvements over regular JSON to make data (mode) human-readable.

HTH.

### Usecase

This format is used in various other projects, such as [vue-i18n](https://vue-i18n.intlify.dev/guide/advanced/sfc.html#json5).

ficd0:
json5 was added in #5413 so this issue can be closed :)

PR #5381, opened 2025-09-05
Json5

slatian:
Add filetype support for JSON5
    
This can do syntax highlighting with some basic error detection.
 
Supports rust-like line breaking behavior inside comments (half the logic is taken from the rust config) and JSON like (taken or imported from the JSON config) line breaking behavior outside of comments.

Related #4392 (though I didn't implement it as part of `json.kak` to avoid highlighting syntax that isn't supported by most json parsers)

### DIFF
--- a/rc/filetype/json5.kak
+++ b/rc/filetype/json5.kak
@@ -1,0 +1,126 @@
+# https://json5.org
+# -----------------
+
+# Detection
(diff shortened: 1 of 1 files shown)


ficd0:
json5 was added in #5413 so this can be closed

PR #5384, opened 2025-09-11
Don’t inject `/opt/local` and `/usr/local` library paths

emilazy:
As no third‐party libraries are used, there’s no reason to add these.

Even if third‐party libraries were being used, it probably wouldn’t be a good idea; `/opt/local` is used by MacPorts, but there’s no reason to automatically inject it specifically and not e.g. `/opt/homebrew`, which is considerably more popular. It also causes build reproducibility issues when building with hermetic systems like Nix.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,16 +49,9 @@ mandir = $(DESTDIR)$(PREFIX)/share/man/man1
 os = $(shell uname | sed 's/.*_NT.*/Windows/')
 os != uname | sed 's/.*_NT.*/Windows/'
 
-CPPFLAGS-os-Darwin = -I/opt/local/include
-LDFLAGS-os-Darwin = -L/opt/local/lib
-
-CPPFLAGS-os-FreeBSD = -I/usr/local/include
-LDFLAGS-os-FreeBSD = -L/usr/local/lib
-
 LIBS-os-Haiku = -lnetwork -lbe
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -49,9 +49,6 @@ mandir = $(DESTDIR)$(PREFIX)/share/man/man1
 os = $(shell uname | sed 's/.*_NT.*/Windows/')
 os != uname | sed 's/.*_NT.*/Windows/'
 
-CPPFLAGS-os-Darwin = -I/opt/local/include
-LDFLAGS-os-Darwin = -L/opt/local/lib
-
```
krobelus, line 54: Probably the injection of /usr/local on FreeBSD/OpenBSD
can be updated the same way.

These were added back when we used third party libraries.
Now that we only use the C++ stdlib, there is probably no reasonable
scenario where we still need this.

Even if the compiler was installed to /usr/local/bin/cc,
it should still be able to find its associated standard library
(though I haven't tested that).

(In case it's not obvious to anyone, a possible workaround for this
would be `make CPPFLAGS-os-Darwin= LDFLAGS-os-Darwin=`)

krobelus (approved):

Makefile
emilazy: I’ve dropped those as well.

And yeah, it’s easy to work around for us, but thought it’d be best to upstream this instead :) Injecting this stuff into the user’s build environment is really the job of the package manager IMO, even if there were libraries to find there.

mawww:
Thanks !

PR #5387, opened 2025-09-17
Add value expansion for macro recording state

evnoj:
I wanted to display whether a macro is being recorded in the status line, but I wanted it to be shorter than the `[recording (@)]` displayed by the `{{context_info}}` markup string that can be used in `modelinefmt`. I also did not want the other information present in `{{context_info}}`. Setting an option when pressing Q/q doesn't seem to work, because there is no "client" scope - so it either takes effect across clients, or doesn't keep when switching buffers while recording. Exposing this via a value expansion was very simple, as it is already tracked and just needed to be exposed. This allows me to display it in the status line.

Let me know if you have any concerns, and I'll see what can be done to address them.

### DIFF
--- a/doc/pages/expansions.asciidoc
+++ b/doc/pages/expansions.asciidoc
@@ -334,6 +334,9 @@ The following expansions are supported (with required context _in italics_):
     an inner selection, `to_begin` if the user wants to select to the
     beginning, and `to_end` if the user wants to select to the end
 
+*%val{recording_register}*::
+    the register being recorded to if currently recording a macro, empty when not recording
+
 *%val{register}*::
     _in `map` command <keys> parameter and `<a-;>` from the object menu_ +
     current register when the mapping was triggered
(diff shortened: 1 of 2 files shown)


doc/pages/expansions.asciidoc
```diff
@@ -324,6 +324,9 @@ The following expansions are supported (with required context _in italics_):
     _in `hook` command <command> parameter_ +
     the complete parameter string of the executing hook
 
+*%val{is_recording}*::
```
mawww: I think it would be better to expose 'recording_register' that would be empty when not recording and contain the recording register when recording.

evnoj: I like that idea, I'll see what I can do

evnoj:
Here is an implementation that creates a `recording_register` value, following your idea. I'm not positive if I'm following the right pattern to turn a `char` into a `String`, but this seemed like the most direct way.

If we arrive at an implementation that you'd like to merge, I can squash before the merge.

mawww:
That looks good, please squash this into just two commits (the copyright waiver + actual change)

evnoj:
Awesome! Squashed. Let me know if anything else is needed here.

PR #5388, opened 2025-09-18
Add Niri windowing module

ficd0:
[Niri](https://yalter.github.io/niri/) is a Wayland compositor that's increasing in popularity. I've been using this windowing module with it for about a month now with no issues. So in that sense, the implementation is complete.

**However, this PR is marked as a Draft because I would like @mawww's input on how we should handle windowing modules that don't conform to the "tiling" paradigm.** Niri is a _scrolling_ WM, so there is no concept of vertical or horizontal "splits" like in Sway or i3. A workspace can have any number of "columns", where a column can be one or more windows stacked on top of each other, of arbitrary width.

Thus, command names like `niri-terminal-horizontal` or `niri-terminal-vertical` don't map cleanly onto any action you can actually do in Niri. Therefore, I've intentionally avoided defining such aliases.

The actions I've implemented are "`consume`" and "`consume-right`". Consuming a window means to pull it into a neighboring column. For example:

```
CONSUMING B INTO A
BEFORE
+-----+ +-----+
|     | |     |
|  A  | |  B  |
|     | |     |
+-----+ +-----+
AFTER
+-----+        
|  A  |        
+-----+        
|  B  |        
+-----+        
```

So, `niri-terminal-consume` will spawn a new terminal and consume it into the column with the calling Kakoune client in it.

```
niri-terminal-consume
BEFORE
+-----+
|     |
| KAK |
|     |
+-----+
AFTER
+-----+
| KAK |
+-----+
|TERM |
+-----+
```

The `niri-terminal-consume-right` command consumes the new terminal window _into the column **to the right of** the Kakoune client_.

```
niri-terminal-consume-right
BEFORE
+-----+ +-----+
|     | |OTHER|
| KAK | |WINDW|
|     | |     |
+-----+ +-----+
AFTER
+-----+ +-----+
|     | |OTHER|
| KAK | +-----+
|     | |TERM |
+-----+ +-----+
```

These are the actions that I personally find most convenient to use with Kakoune.

## Comments

The actual implementation using Niri's IPC may seem pretty complicated on the surface. I've included a _lot_ of comments to explain what's going on, which can hopefully guide future contributors. I've also noted myself as the author (with an email address), in case future contributors have some questions for me about it. The only reason I did this was because, well, the code is not nearly as straightforward as the other windowing modules, so I want to make myself available.

@mawww please let me know if the comments in the module are appropriate. I'm happy to take my own contact out and reduce the overall number of comments if you prefer.

### DIFF
--- a/rc/windowing/niri.kak
+++ b/rc/windowing/niri.kak
@@ -2,8 +2,52 @@
 # ------------------------------
 
 # Kakoune Niri windowing module
-# original author: Daniel <daniel AT ficd DOT sh>
-# (contact with questions or problems)
+
+# Niri is a scrolling window manager;
+# vertical/horizontal splits aren't really a native concept.
+# A workspace consists of columns of arbitrary width, where each
+# column can have one or more window in it. Windows in a column
+# are stacked on top of each other. To "consume" a window into a
+# column is to pull it BELOW the other windows in the column.
+# For example, consider:
+# +-----+ +-----+                         +-----+
+# |     | |     |                         |  A  |
+# |  A  | |  B  |  --consume B into A-->  +-----+
+# |     | |     |                         |  B  |
+# +-----+ +-----+                         +-----+
+
+# We provide three ways to open a new window.
+
+# normal: simply spawn a new window to the right.
+#   This is the default for new windows in Niri.
+# Alias: horizontal
+# +-----+ +-----+
+# |     | |     |
+# | KAK | | NEW |
+# |     | |     |
+# +-----+ +-----+
+
+# consume: the new window is consumed into the current column.
+# Alias: vertical
+# +-----+
+# | KAK |
+# +-----+
+# | NEW |
+# +-----+
+
+# consume-right: spawn new window and consume into column to the right
+#   Note: FOO is an already open window to the right.
+# Alias: none
+# +-----+ +-----+
+# |     | | FOO |
+# | KAK | +-----+
+# |     | | NEW |
+# +-----+ +-----+
+
+# In terms of closest analogues to the vertical & horizontal split paradigm,
+# we alias "vertical" to "consume" and we alias "horizontal" to "normal".
+# There isn't a clean alias for "consume-right", but it's a convenient command
+# that users can invoke directly.
 
 provide-module niri %~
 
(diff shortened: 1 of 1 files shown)


rc/windowing/niri.kak
```diff
@@ -0,0 +1,169 @@
...
+        	ready=false
+        	# read the event stream
+        	while IFS= read -r line; do
+        		case "$line" in
+        		# this line means stream is initialized
+        		*"OverviewOpenedOrClosed"*)
+        			if [ "$ready" = false ]; then
+        				ready=true
+        				# spawn terminal with command
+        				printf 'wayland-terminal-window %s' "$quoted_args" >"$kak_command_fifo"
```
Screwtapello: If we have a `niri-terminal-window` alias we should probably use it. Even if it defaults to `wayland-terminal-window`, the user might have changed it behind our backs.

Screwtapello:
As somebody who hasn't used a scrolling window manager much (except for the GNOME one until I got fed up by the crashes), it seems to me that if I called `niri-terminal-vertical` and it turned out to be an alias for `niri-terminal-consume`, I probably wouldn't be disappointed or even surprised. If there's a way to open a new window *between* the current column and the one on the right (i.e. start a new, full-height (i.e. spawn a new window without it being consumed into any existing column) that would seem like a good candidate for `niri-terminal-horizontal`.

It would be nice if we could find somebody who regularly uses Kakoune's specific split commands, and get them to try a scrolling window manager to see if those mappings suit their workflow.

rc/windowing/niri.kak
ficd0: Good catch, I'll change this

ficd0:
> If there's a way to open a new window _between_ the current column and the one on the right (i.e. start a new, full-height (i.e. spawn a new window without it being consumed into any existing column) that would seem like a good candidate for `niri-terminal-horizontal`.

Funny enough, that's just what happens when you open a window normally. So, I guess `niri-terminal-horizontal` could be an alias for `niri-terminal-window`. `vertical` for `consume` makes sense, too. And `consume-right` is (in my opinion) still a useful action to have, but it just won't be aliased to anything else.

Screwtapello:
What are the `niri-new-*` commands for? My understanding is that `new` delegates to `terminal` which delegates to `%opt{windowing_module}-terminal-%opt{windowing_placement}`, which should be one of the `niri-terminal-*` commands, and I don't expect anything to call `niri-new-*` directly.

(although the standard values for `%opt{windowing_placement}` are "window", "horizontal", "vertical", etc. there's no reason a user couldn't temporarily set them to "consume" or "consume-right" if that's the specific behaviour they wanted.)

I'm also concerned that `kak -e "%arg{@}"` doesn't properly quote each argument, but it seems the standard `new` command does exactly the same thing. Ostrich algorithm, indeed. :)

Screwtapello (approved):

PR #5389, opened 2025-09-22
Remove max path

drewfer:
This commit changes the relevant internal path buffers to use dynamic array sizing and allows Kakoune to build on GNU HURD.

### DIFF
--- a/src/file.cc
+++ b/src/file.cc
@@ -63,24 +63,33 @@ std::pair<StringView, StringView> split_path(StringView path)
     return { {path.begin(), slash+1}, {slash+1, path.end()} };
 }
 
+struct MallocDeleter
+{
+    void operator()(char *ptr) const
+    {
+        free(ptr);
+    }
+};
+
 String real_path(StringView filename)
 {
     if (filename.empty())
         return {};
 
-    char buffer[PATH_MAX+1];
-
     StringView existing = filename;
     StringView non_existing{};
 
+    using unique_cstr = std::unique_ptr<char, MallocDeleter>;
+
     while (true)
     {
-        if (char* res = realpath(existing.zstr(), buffer))
+        unique_cstr res(realpath(existing.zstr(), nullptr));
+        if (res)
         {
             if (non_existing.empty())
-                return res;
+                return String(res.get());
 
-            StringView dir = res;
+            StringView dir = res.get();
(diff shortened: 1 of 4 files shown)


src/file.cc
```diff
@@ -295,22 +304,25 @@ void write_to_file(StringView filename, StringView data)
...
-        format_to(buffer, ".{}.kak.XXXXXX", file);
+        buffer = format(".{}.kak.XXXXXX", file);
     else
-        format_to(buffer, "{}/.{}.kak.XXXXXX", dir, file);
+        buffer = format("{}/.{}.kak.XXXXXX", dir, file);
 
-    return mkstemp(buffer);
+    char* tmplt = strndup(buffer.c_str(), (size_t)buffer.length());
+    int fp = mkstemp(tmplt);
+    buffer = tmplt;
```
mawww, line 319: Aren´t we leaking tmplt here ? I am not sure why we need the `strndup` call in the first place, you can just use `mkstemp(buffer.data())` which will be a pointer to the buffer string buffer and can be mutated safely as long as the string length does not change.

src/file.cc
```diff
@@ -63,24 +63,33 @@ std::pair<StringView, StringView> split_path(StringView path)
...
 {
     if (filename.empty())
         return {};
 
-    char buffer[PATH_MAX+1];
-
     StringView existing = filename;
     StringView non_existing{};
 
+    using unique_cstr = std::unique_ptr<char, MallocDeleter>;
```
mawww, line 82: Kakoune moved away to a custom `UniquePtr` implementation for compilation speed reasons, it does not support custom deleters as we did not have actual uses, but in this specific case a `OnScopeEnd free{[&] { free(res); };` would probably do the trick.

src/file.hh
```diff
@@ -46,8 +46,8 @@ template<bool force_blocking = false>
 void write(int fd, StringView data);
 void write_to_file(StringView filename, StringView data);
 int create_file(const char* filename);
+int open_temp_file(StringView filename, String &tmpFile);
```
mawww, line 49: we do not use camel case for parameters in kakoune, it should be `tmp_file`

mawww (commented):
While I did enjoy avoiding those allocations, if this breaks build on Hurd I think it makes sense to have more portable and robust code.

PR #5392, opened 2025-09-26
Improve autowrap paragraph formatting

Chris-F5:
To demonstrate my problem with the existing `autowrap_format_paragraph` option, try the following, bring a cursor to the end of one of the middle lines and type some text until it wraps. You will notice that a new line is created without running the `autowrap_fmtcmd` on the remainder of the text. The behaviour I expect is similar to what happens if you were to start typing in the middle of one of the lines.

```
echo "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Proin eu convallis
neque. Phasellus varius efficitur lacus ut sodales. Aenean pellentesque ut dui
ut dapibus. Phasellus eros nisl, accumsan nec malesuada at, luctus aliquet
diam. Mauris ac velit quis felis pellentesque placerat. Vivamus condimentum
eget massa eget tempus. Duis luctus finibus tellus, quis lacinia metus
vestibulum a. Donec viverra ullamcorper dui in cursus. Donec sapien est,
aliquam vel velit nec, ultricies gravida ante. Etiam bibendum dui rhoncus,
pharetra augue id, venenatis lectus. Vestibulum pretium finibus nisi, quis
molestie est accumsan a. Sed id purus id ipsum consequat volutpat eu at
mi. Sed euismod tincidunt eros. Donec ac turpis eu eros condimentum laoreet
non ut neque. Quisque vel congue enim, id iaculis lectus. Quisque pretium,
metus id egestas auctor, ex elit sodales elit, a luctus turpis orci nec velit.
" | kak -e 'autowrap-enable; set-option window autowrap_format_paragraph yes'
```

I do not see why the try catch was used in the initial implementation.

### DIFF
--- a/rc/tools/autowrap.kak
+++ b/rc/tools/autowrap.kak
@@ -16,33 +16,33 @@ define-command -hidden autowrap-cursor %{ evaluate-commands -save-regs '/"|^@m' 
         execute-keys -draft "x<a-k>^[^\n]{%opt{autowrap_column},}[^\n]<ret>"
 
         try %{
-            reg m "%val{selections_desc}"
 
             ## if we're adding characters past the limit, just wrap them around
             execute-keys -draft "<a-h><a-k>.{%opt{autowrap_column}}\h*[^\s]*<ret>1s(\h+)[^\h]*\z<ret>c<ret>"
-        } catch %{
-            ## if we're adding characters in the middle of a sentence, use
-            ## the `fmtcmd` command to wrap the entire paragraph
-            evaluate-commands %sh{
-                if [ "${kak_opt_autowrap_format_paragraph}" = true ] \
-                    && [ -n "${kak_opt_autowrap_fmtcmd}" ]; then
-                    format_cmd=$(printf %s "${kak_opt_autowrap_fmtcmd}" \
-                                 | sed "s/%c/${kak_opt_autowrap_column}/g")
-                    printf %s "
-                        evaluate-commands -draft %{
-                            execute-keys '<a-]>px<a-j>|${format_cmd}<ret>'
-                            try %{ execute-keys s\h+$<ret> d }
-                        }
-                        select '${kak_main_reg_m}'
-                    "
-                fi
-            }
+        }
+
+        reg m "%val{selections_desc}"
+        ## if we're adding characters in the middle of a sentence, use
+        ## the `fmtcmd` command to wrap the entire paragraph
+        evaluate-commands %sh{
+            if [ "${kak_opt_autowrap_format_paragraph}" = true ] \
+                && [ -n "${kak_opt_autowrap_fmtcmd}" ]; then
+                format_cmd=$(printf %s "${kak_opt_autowrap_fmtcmd}" \
+                             | sed "s/%c/${kak_opt_autowrap_column}/g")
+                printf %s "
+                    evaluate-commands -draft %{
+                        execute-keys '<a-]>px<a-j>|${format_cmd}<ret>'
+                        try %{ execute-keys s\h+$<ret> d }
+                    }
+                    select '${kak_main_reg_m}'
+                "
+            fi
         }
     }
 } }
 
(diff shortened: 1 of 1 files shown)


rc/tools/autowrap.kak
```diff
@@ -16,33 +16,33 @@ define-command -hidden autowrap-cursor %{ evaluate-commands -save-regs '/"|^@m'
...
-                        }
-                        select '${kak_main_reg_m}'
-                    "
-                fi
-            }
+        }
+
+        reg m "%val{selections_desc}"
+        ## if we're adding characters in the middle of a sentence, use
+        ## the `fmtcmd` command to wrap the entire paragraph
```
mawww: This comment seems out of date now, AFAIU this change runs the format command regardless of if we are adding character at the end or in the middle of the sentence as we do not only run in the case where the previous execute-keys failed anymore.

mawww (commented):
I do not quite remember why autowrap does not run the command if we are at the end of the line, I suppose it is for performance reasons but that seems like a weak argument for me as this needs to be fast enough to be unnoticeable when editing the middle of a paragraph anyway.

I may have missed something, but isn´t the try block now redundant with whatever the format command will do afterwards ? If we move towards always running the format command we may remove the whole manual wrapping logic as well.

Anybody familiar with this script wants to chime in ?

Chris-F5:
The try block is necessary to move the cursor onto the next line. If, when inserting characters past the autowrap limit, we only ran the format command then the cursor would remain on the first line when the text you just inserted moves to the next line.

We might be able to simplify the logic a bit with something like: 1. record the column number you are on (call this n) 2. run the formatting command. 3. move n characters into the block of text you just ran the format command on. This way the cursor would (hopefully) be in the right place even when going onto a newline. I'll have a play with this idea and see if I can make it work nicely.

Chris-F5:
The fundamental problem with using autoformat commands such as `fold` and `par` is that they do not know about kakoune selections. If the text under a selection moves to a newline then `fold` cant tell kak 'please move the selection over here to the newline'.

I've been trying out the following. When the user inserts a character we 1. replace it with a rare unicode symbol (∅ in this case) 2. run `par` on the paragraph 3. move the cursor back over ∅ 4. replace ∅ with the originally inserted character. This way we are seeking the kakoune cursor positions through the auto formatting command by using ∅ to represent the cursor position. By doing this we can start using more complex formatting commands such as those which know how to handle comments at the start of lines.

```
define-command -hidden autowrap-cursor %{
    execute-keys -draft "hr∅"
    # we would probably want to replace `par -w80` with %opt{autowrap_fmtcmd}
    execute-keys "<a-;><a-i>p<a-;>|par -w80<ret><a-;>s∅<ret><a-;>l"
    execute-keys -draft "hr%val{hook_param}"
}
```

The obvious downside of this is that it will break if you actually want to use this unicode character in your paragraph. Also it means when autowrap is enabled your selection will be reduced to a single character wide when you insert text. But apart from that I have found it to be quite robust.

Any thoughts on this idea? I will try using this for a while and see how it goes.

[asciicast](https://asciinema.org/a/AkOTurH4Rpb1HkbfmEVCChK0q)

mawww:
> The try block is necessary to move the cursor onto the next line. If, when inserting characters past the autowrap limit, we only ran the format command then the cursor would remain on the first line when the text you just inserted moves to the next line.

Hum, but that try block just contains an `execute-keys -draft ...` which would not actually impact the catch block as it ran in a draft context and hence did not touch the selections the catch block will work with.

> The fundamental problem with using autoformat commands such as fold and par is that they do not know about kakoune selections. If the text under a selection moves to a newline then fold cant tell kak 'please move the selection over here to the newline'.

We could work-around this by extending the current piping diff logic to work on characters instead of lines which should detect that we inserted newlines and correctly preserve selections. This might however prove quite expensive, but may be something to explore.

Chris-F5:
Thats a great idea, I think extending the piping diff logic is the way to go if turns out to be fast enough.
For my future reference: relevant call subgraph is `pipe->apply_diff in src/normal.cc`.

PR #5402, opened 2025-10-16
Fix manpath detection in tools/man.kak

silversquirl:
The `manpath` binary is the preferred way to find the path for man pages with GNU man. This is relevant when `PATH` contains extra bin directories, but `MANPATH` is unset - `manpath` will find the corresponding `share/man` directories for each entry in `PATH`.

### DIFF
--- a/rc/tools/man.kak
+++ b/rc/tools/man.kak
@@ -63,8 +63,11 @@ define-command -hidden -params ..3 man-impl %{ evaluate-commands %sh{
 
 define-command -params ..1 \
     -shell-script-candidates %{
-        find /usr/share/man/ $(printf %s "${MANPATH}" |
-            sed 's/:/ /') -name '*.[1-8]*' |
+        : "${MANPATH:="$(manpath)"}"
+        : "${MANPATH:=/usr/share/man}"
+        set -f
+        IFS=:
+        find $MANPATH -name '*.[1-8]*' |
             sed 's,^.*/\(.*\)\.\([1-8][a-zA-Z]*\).*$,\1(\2),'
     } \
     -docstring %{


rc/tools/man.kak
```diff
@@ -63,8 +63,11 @@ define-command -hidden -params ..3 man-impl %{ evaluate-commands %sh{
 
 define-command -params ..1 \
     -shell-script-candidates %{
-        find /usr/share/man/ $(printf %s "${MANPATH}" |
-            sed 's/:/ /') -name '*.[1-8]*' |
+        : "${MANPATH:="$(manpath)"}"
```
krobelus, line 66: I think this breaks the case where `MANPATH` is set but contains an empty component (which should be substituted with the "system default path").

```
$ env | grep MANPATH
MANPATH=/home/johannes/.local/share/man:
$ manpath
manpath: warning: $MANPATH set, appending /etc/man_db.conf
/home/johannes/.local/share/man:/home/johannes/.local/share/man:/usr/local/man:/usr/local/share/man:/usr/share/man:/usr/lib/jvm/default/man
```

So I think we want to call `manpath` always,
or reimplement it by looking in `$PATH`

Either way, it looks like this patch would allow me to drop this explicitly-set `$MANPATH` which I just realized is already implied by `$PATH` (except in Kakoune's `:man`).

BTW, a completely different approach would be to use `apropos ^` to compute these completions,
which should give the same results (it's shipped with mandoc man-db).
That's what [fish shell uses](https://github.com/fish-shell/fish-shell/blob/15f9d6d2793c3ea82e5fdd9f8da0df1a4e7f84bb/share/functions/__fish_complete_man.fish); but it has [gross workarounds for macOS](https://github.com/fish-shell/fish-shell/blob/15f9d6d2793c3ea82e5fdd9f8da0df1a4e7f84bb/share/functions/__fish_apropos.fish).
We could try to get rid of those.

rc/tools/man.kak
```diff
@@ -63,8 +63,11 @@ define-command -hidden -params ..3 man-impl %{ evaluate-commands %sh{
...
+        : "${MANPATH:=/usr/share/man}"
+        set -f
+        IFS=:
+        find $MANPATH -name '*.[1-8]*' |
```
krobelus, line 70: > manpath will find the corresponding share/man directories for each entry in PATH.

where is this documented?
I can see the `PATH` behavior mentioned in [FreeBSD man](https://manpage.me/?q=manpath)
but not in man-db man

PR #5403, opened 2025-10-16
Make the :jump-next and :jump-previous commands remember the last used jump buffer

silversquirl:
Useful to allow bindings that work for any plugin that uses a jump buffer :)

### DIFF
--- a/rc/tools/jump.kak
+++ b/rc/tools/jump.kak
@@ -5,6 +5,7 @@ declare-option -docstring "name of the client in which utilities display informa
 
 provide-module jump %{
 
+declare-option -hidden str jump_current_buffer ''
 declare-option -hidden int jump_current_line 0
 
 define-command -hidden jump %{
@@ -14,6 +15,7 @@ define-command -hidden jump %{
                 execute-keys ',xs^([^:\n]+):(\d+):(\d+)?<ret>'
                 set-register a %reg{1} %reg{2} %reg{3}
             }
+            set-option global jump_current_buffer %val{bufname}
             set-option buffer jump_current_line %val{cursor_line}
             evaluate-commands -try-client %opt{jumpclient} -verbatim -- edit -existing -- %reg{a}
             try %{ focus %opt{jumpclient} }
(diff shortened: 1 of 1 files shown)


rc/tools/jump.kak
```diff
@@ -14,24 +15,27 @@ define-command -hidden jump %{
...
             evaluate-commands -try-client %opt{jumpclient} -verbatim -- edit -existing -- %reg{a}
             try %{ focus %opt{jumpclient} }
         }
     }
 }
 
-define-command jump-next -params 1.. -docstring %{
+define-command -hidden jump-buffer -params 1.. %{ buffer %arg{1} }
+
+define-command jump-next -params ..1 -docstring %{
```
krobelus: unfortunately this breaks `jump-next -matching \*grep.*`
which is what I use to address a similar problem:

```kakrc
declare-option str my_jump_buffers \*(grep|find|callers|callees|diagnostics|goto|symbols|gdb-backtrace)\b.*\*

map global normal <c-n> %{:jump-next -matching %opt{my_jump_buffers}<ret>} -docstring 'next location'
map global normal <c-p> %{:jump-previous -matching %opt{my_jump_buffers}<ret>} -docstring 'previous location'
```

but I guess your workflow is different in that you do `:grep` then press `<ret>` and then want to use `:jump-{next,previous}` to keep going in the same buffer.
I guess that probably makes sense, I just never had the need to

silversquirl: ohhhh, *that's* why it was `1..`, that makes sense. I didn't clock that you could use `-matching` since that's not documented in `jump-next`/`jump-prev`'s docstrings, but it's obvious in hindsight!

rc/tools/jump.kak
```diff
@@ -14,24 +15,25 @@ define-command -hidden jump %{
...
-        jump
+    evaluate-commands -try-client %opt{jumpclient} -save-regs / %sh{
+        args='%arg{@}'
+        [ "$#" -eq 0 ] && args='%opt{jump_current_buffer}'
+        printf %s\\n "buffer $args" jump-select-next jump
     }
     try %{
         evaluate-commands -client %opt{toolsclient} %{
-            buffer %arg{@}
+            buffer %opt{jump_current_buffer}
```
krobelus, line 36: as written, this breaks users who override `jump` in buffer scope,
for example the kakoune-lsp plugin 

creates some scratch buffers with
`alias buffer jump "%val{hook_param_capture_1}-jump"`

where the rhs can be one of

```
lsp-goto-jump
lsp-document-symbol-jump
lsp-diagnostics-jump
```

which are not aware of `%opt{jump_current_buffer}`

So it seems if we want this, we should set `jump_current_buffer` outside `jump` because that's intended to be overridden (as are `jump-select-next` and `jump-select-previous`).

Overall I'm not sure if this PR is worth it. It introduces extra implicit state.
Consider that other commands like `:buffer` have no memory either; but we do have generic command history which should cover some of your need.
I'm not against DWIM but it can be a lot of work to get right.

Does the `jump-next -matching ...` approach not work for the same use case?
We should definitely document that better.

mawww:
This use case is already catered for but I realize this is not really documented:

`jump-next` passes its arguments through to the `buffer` command, which means the `-matching` switch can be used. This means you can do `jump-next -matching \*(grep|make)(-.*)\*` and this will jump using the first buffer in the buffer list that matches the regex, with new buffers being added at the start. This means you can use the buffer list ordering (using `arrange-buffers` command) to control which buffer should be used for your jump command.

In other words, with the suggested previous command, the latest buffer that matches the regex will be used for jumping, so if you do a `:grep` then the new grep result buffer will be used, then if you do a `:make` it will jump using the make result buffer, then if you close the make result buffer the command will use the grep buffer again. And if you need to manually control which buffer should be used, the `arrange-buffer` command makes it possible to move existing buffer to the front of the buffer list.

PR #5432, opened 2026-01-06
Skip call to wcwidth for ascii in codepoint_width

bscuron:
`perf record kak src/commands.cc`, page down to bottom of file

wcwidth before:
- ~4-5%

wcwidth after:
- Does not appear in perf report

All tests continue to pass with `make test`

### DIFF
--- a/src/unicode.hh
+++ b/src/unicode.hh
@@ -101,7 +101,7 @@ inline bool is_identifier(Codepoint c) noexcept
 
 inline ColumnCount codepoint_width(Codepoint c) noexcept
 {
-    if (c == '\n')
+    if (c < 0x80) [[likely]]
         return 1;
     const auto width = wcwidth((wchar_t)c);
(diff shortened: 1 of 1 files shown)


src/unicode.hh
```diff
@@ -101,7 +101,7 @@ inline bool is_identifier(Codepoint c) noexcept
 
 inline ColumnCount codepoint_width(Codepoint c) noexcept
 {
-    if (c == '\n')
+    if (c < 0x80) [[likely]]
```
krobelus, line 104: FWIW I'm not sure if we ever call this on control characters,
but this looks correct at least for terminal UI which renders them as control pictures, see `fix_atom_text`

mawww: Did a quick test, seems like wcwidth returns -1 for all control characters except '\0' which would make this function return 1, so this change should only impact the return value for '\0' which should likely not return 0 in the first place.

Issue #5435, opened 2026-01-12
[BUG] git clone gives error on macOS (filename case insensitivity)


sidkshatriya:
### Version of Kakoune

master

### Reproducer

The macOS filesystem is case insensitive. Filenames that collide in a case insensitive filesystem should not be checked into kakoune.

```bash
$ git clone https://github.com/mawww/kakoune.git
Cloning into 'kakoune'...
remote: Enumerating objects: 60018, done.
remote: Counting objects: 100% (1816/1816), done.
remote: Compressing objects: 100% (735/735), done.
remote: Total 60018 (delta 1341), reused 1115 (delta 1063), pack-reused 58202 (from 3
Receiving objects: 100% (60018/60018), 42.74 MiB | 11.77 MiB/s, done.
Resolving deltas: 100% (41828/41828), done.
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'test/normal/object/nested/around/WORD/cmd'
  'test/normal/object/nested/around/word/cmd'
  'test/normal/object/nested/around/WORD/in'
  'test/normal/object/nested/around/word/in'
  'test/normal/object/nested/around/WORD/kak_quoted_selections'
  'test/normal/object/nested/around/word/kak_quoted_selections'
  'test/normal/object/nested/inner/WORD/cmd'
  'test/normal/object/nested/inner/word/cmd'
  'test/normal/object/nested/inner/WORD/in'
  'test/normal/object/nested/inner/word/in'
  'test/normal/object/nested/inner/WORD/kak_quoted_selections'
  'test/normal/object/nested/inner/word/kak_quoted_selections'
```

### Outcome

git complains

### Expectations

git should not complain

PR #5436, opened 2026-01-14
Bug/5435 case fs macos

paaloeye:
Closes https://github.com/mawww/kakoune/issues/5435


bscuron:
@paaloeye I opened a pr that just renames the tests (#5438). Feel free to do the same and I can close mine.

PR #5444, opened 2026-01-30
Add is_word fast path for Codepoint{-1}

bscuron:


mawww:
Can you expand on when this makes a difference ? -1 codepoints should be rare, as they are what we use whenever we fail to decode utf8.

bscuron:
> Can you expand on when this makes a difference ? -1 codepoints should be rare, as they are what we use whenever we fail to decode utf8.

I believe it only matters in src/ranked_match when reading the previous utf codepoint for completion ordering.

RankedMatch::operator<
```c++
        // compare codepoints
        it1 = utf8::character_start(it1, last1);
        it2 = utf8::character_start(it2, last2);
        const auto itsave1 = it1, itsave2 = it2;
        const auto cp1 = utf8::read_codepoint(it1, end1);
        const auto cp2 = utf8::read_codepoint(it2, end2);
        if (cp1 != cp2)
        {
            const auto cplast1 = utf8::prev_codepoint(itsave1, begin1);
            const auto cplast2 = utf8::prev_codepoint(itsave2, begin2);
            const bool is_wb1 = is_word_boundary(cplast1, cp1);
            const bool is_wb2 = is_word_boundary(cplast2, cp2);
            if (is_wb1 != is_wb2)
                return is_wb1;

            const bool low1 = is_lower(cp1);
            const bool low2 = is_lower(cp2);
            if (low1 != low2)
                return low1;

            return order(cp1) < order(cp2);
        }
```

Maybe it makes more sense to add the check for invalid policy here if you think it is worth adding at all.

PR #5446, opened 2026-02-16
normal: make <c-o> always jump backward when snapshotting current selection

### LINKED ISSUE #5066: [BUG] <c-o> creates jump points instead of jumping backwards
### Version of Kakoune

83fb65aef5729167b47c6ec22a57e5bd6ac5a9ae

### Reproducer

1. Start an empty kakoune with `kak -n`.
2. `<c-s>` save the beginning of the file to the jump list (so first `*` is selected and saved).
3. `w` to select the next word.
4. `<c-s>` to save again (it will save with the full `*** ` selected).
5. `w` to to next word again (just to move away from the previous selection, any direction would do).

At that point, I am supposed to have in jumplist:

- `#0`: first `*` at the beginning of the file.
- `#1`: first word (`*** `).

Now, doing `<c-o>` goes back to `#1` and selects again `*** `. Until then, everything is fine. However, from now the bug will start to appear.

So just try to move in any direction (such as `j` to go to the second line), and press `<c-o>` again, and it will create a new entry in the jumplist instead of going back to `#1` again.

And you can keep reproducing. Move again, do `<c-o>`, new entry again, and so on.

### Outcome

It adds new jump points instead of going back to the last one.

### Expectations

Go back to `#1`(which is the latest), and if I do again `<c-o>` from `#1`, go back to `#0`.

### Additional information

_No response_

lobre:
`JumpList::backward` used to sometimes consume `<c-o>` without moving: when the current selection had diverged from the jump at `m_current`, we first pushed the current selection, then subtracted count.

With `count=1`, this landed on the freshly pushed entry itself, so `<c-o>` behaved like "save checkpoint" instead of "jump backward".

This was especially visible after going back into the middle of history, moving, then pressing `<c-o>` again: the first `<c-o>` could be a no-op from the user point of view.

Fix by treating the implicit push as an extra internal step: if we push current before jumping, move back by `count + 1`. This keeps the "preserve current position for `<c-i>`" behavior, while making `<c-o>` consistently perform an actual backward jump.

Add regression test: `test/normal/jump/backward-dirty-middle`

Fixes #5066

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -133,39 +133,24 @@ const SelectionList& JumpList::forward(Context& context, int count)
 
 const SelectionList& JumpList::backward(Context& context, int count)
 {
-    if ((int)m_current - count < 0)
+    const SelectionList& current = context.selections();
+    const bool should_push_current =
+        m_current == m_jumps.size() or
+        (m_current != m_jumps.size() and m_jumps[m_current] != current);
+    if (should_push_current)
+        push(current);
+
+    const int steps = count + (should_push_current ? 1 : 0);
+    if ((int)m_current - steps < 0)
         throw runtime_error("no previous jump");
 
-    const SelectionList& current = context.selections();
-    if (m_current != m_jumps.size() and
-        m_jumps[m_current] != current)
-    {
-        push(current);
-        m_current -= count;
-        SelectionList& res = m_jumps[m_current];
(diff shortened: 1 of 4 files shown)


mawww:
Thanks, this looks like a nice cleanup of that function on top of fixing the bug.

PR #5448, opened 2026-02-27
Fix Kakoune logo URL in README

3m4r5:
Updated logo URL to point to the new image location.


README.asciidoc
```diff
@@ -1,6 +1,6 @@
 = image:{logo}[K,30,30,link="{website}",title="Kakoune logo by p0nce"] Kakoune image:{cirrus-img}[link="{cirrus-url}"] image:{srht-img}[link="{srht-url}"] image:{irc-img}[link="{irc-url}"]
 ifdef::env-github,env-browser[:outfilesuffix: .asciidoc]
-:logo: https://rawgit.com/mawww/kakoune/master/doc/kakoune_logo.svg
+:logo: https://kakoune.org/img/kakoune_logo_full.png
```
mawww: Why migrate from the svg to the png ?

3m4r5: the original svg doesn't work anymore, when I tried to link to a working svg (https://raw.githubusercontent.com/mawww/kakoune/refs/heads/master/doc/kakoune_logo.svg) it looked off on my mobile device, when I tried the png it looked good, I think it might be a webkit issue.

should I add another commit or close this and open another pr with the svg instead?

mawww: In that case we may as well use the kakoune_logo_32.png image, and update the image size to 32,32 to avoid rescaling

3m4r5: done

PR #5449, opened 2026-02-28
rc git: Add more completions (diff/show/log)

KoviRobi:
In particular `git diff --cached` is something I use frequently, but also added some others that might be useful from the editor.

### DIFF
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -133,11 +133,28 @@ define-command -params 1.. \
             update-diff \
         ;
     else
+        # Quote POSIX extended regular expressions, see SPEC_CHAR/QUOTED_CHAR in
+        # https://pubs.opengroup.org/onlinepubs/009696899/basedefs/xbd_chap09.html#tag_09_05_01
+        sed_quote_ext_re() {
+            # SED uses BREs by default, and matching in a group the ERE SPEC_CHAR
+            sed -e 's/[\^\.\[\$\\()|\*+?{]/\\&/g' "$@"
+            # to balance kakoune %<lbrace>: }
+        }
+        suggest_refs() {
+            # Ten most recent refs
+            git for-each-ref --format='%(refname:short)' --sort=-comitterdate --count=10
+            # Commits, using name to match instead of hash (in the absence of having subject
+            # as a comment for completion).
+            git log -100 --format='%s' | sed_quote_ext_re -e 's|.*|@^{/&}|g'
+        }
         case "$1" in
             commit) printf -- "--amend\n--no-edit\n--all\n--reset-author\n--fixup\n--squash\n"; git ls-files -m ;;
             add) git ls-files -dmo --exclude-standard ;;
             apply) printf -- "--reverse\n--cached\n--index\n--3way\n" ;;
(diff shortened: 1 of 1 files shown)


rc/tools/git.kak
```diff
@@ -133,11 +133,21 @@ define-command -params 1.. \
             update-diff \
         ;
     else
+        suggest_refs() {
+            # Ten most recent refs
+            git for-each-ref --format='%(refname:short)' --sort=-comitterdate | head -n10
+            # Commits, using name to match instead of hash (in the absence of having subject
+            # as a comment for completion).
+            git log -100 --format='@^{/%s} # %s'
```
krobelus: `@^{/%s}` is pretty neat. If we take this, we might want to regex-escape the `%s` in future

rc/tools/git.kak
```diff
@@ -133,11 +133,21 @@ define-command -params 1.. \
             update-diff \
         ;
     else
+        suggest_refs() {
+            # Ten most recent refs
+            git for-each-ref --format='%(refname:short)' --sort=-comitterdate | head -n10
```
krobelus: `git for-each-ref` has a `--count` option which we could use here I guess

rc/tools/git.kak
```diff
@@ -133,11 +133,21 @@ define-command -params 1.. \
...
+            git log -100 --format='@^{/%s} # %s'
+        }
         case "$1" in
             commit) printf -- "--amend\n--no-edit\n--all\n--reset-author\n--fixup\n--squash\n"; git ls-files -m ;;
             add) git ls-files -dmo --exclude-standard ;;
             apply) printf -- "--reverse\n--cached\n--index\n--3way\n" ;;
             grep|edit) git ls-files -c --recurse-submodules ;;
+            diff) printf -- "--cached\n----merge-base\n--stat\n--word-diff\n"; suggest_refs ;;
+            show) printf -- "--patch\n--stat\n--remerge-diff\n"; suggest_refs ;;
+            log) printf -- "--patch\n--stat\n--remerge-diff\n--walk-reflogs\n--grep"; suggest_refs ;;
```
krobelus: missing final `\n` after "grep"

rc/tools/git.kak
KoviRobi: Oops for the left-in `# %s` which doesn't work but was just a test.

Figuring out which (ir)regular expression git uses is a bit of a dive, but I think I found it, `get_oid_oneline` calls `regcomp(&regex, prefix, REG_EXTENDED)`, and only does PCRE for `git grep`. Who knows if that works on Windows, maybe I'll test it if I remember by the time I get to a WIndows computer.

Added quoting, it seems to work for e.g this very commit which uses parentheses.

KoviRobi: Also yes `@^{/%s}` is neat but it doesn't work for e.g. reflog (which often has the same commit messages due to rebasing) -- if we could have comments for completion (e.g. `\0` to the next `\n` maybe or something similar) that would be less of a workaround

mawww:
That `@^{...}` trick is indeed pretty nice, maybe a bit too smart for my taste but I decided to merge it and see later if it bothers me too much. Thanks !

PR #5450, opened 2026-03-01
Fix broken Reddit URL

paaloeye:


paaloeye:
Reddit looks somewhat still active and, hence, the fix. 
@mawww let me know if you would rather have it gone.

PR #5453, opened 2026-03-04
fix(client): refresh only if m_ui_pending is not empty

paaloeye:
Fixes regression introduced in 551b20a where moving `generate_mode_line` after `update_display_buffer` removed the early return check. This caused `refresh` to be called even when no UI updates were needed, breaking tests that expected specific message sequences.

Fixes: 551b20a34aeab6e1736da42d9ae5b0356678a7f3


paaloeye:
@mawww I believe this one fixes **5405-error-face-on-empty-prompt** regression test. I'm glad it is in place! Cheers!

PR #5454, opened 2026-03-07
fix(ci): prevent git tools deadlock in ci

paaloeye:
- drain fifo in git blame test to prevent deadlock on Darwin
- added 10-minute timeout to prevent indefinite hanging in CI builds

That fix doesn't eliminate the deadlock completely, but it makes it less probable. From what I gathered, the reason it is only a Darwin problem, is the difference between fifo kernel buffer size (8kb vs 64kb) on Darwin and Linux, as well as heavy usage of fifos for bidirectional communication in git.kak.

### DIFF
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -1,11 +1,12 @@
 name: Makefile CI
 on:
   push:
-    branches: [ "master" ]
+    branches: [ "master", "main" ]
   pull_request:
-    branches: [ "master" ]
+    branches: [ "master", "main" ]
 jobs:
   build:
+    timeout-minutes: 10
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/makefile.yml
```diff
@@ -1,11 +1,12 @@
 name: Makefile CI
 on:
   push:
-    branches: [ "master" ]
+    branches: [ "master", "main" ]
   pull_request:
-    branches: [ "master" ]
+    branches: [ "master", "main" ]
```
mawww, line 6: There is no "main" branch, why add it here ?

paaloeye: It's handy for CI flakes testing if your fork called "main" instead of "master".
I'm happy to delete it tho.

PR #5455, opened 2026-03-10
Expose widget_columns in json-ui draw message

Yus314:
Spawned from discussion in #4737.

This adds `widget_columns` to the json-ui `draw` call, allowing
front-ends to determine how many columns at the start of each line
are used by flag-lines highlighters (e.g. `number-lines`).

## Design choice

In the review of #4737, @mawww initially preferred passing
`widget_columns` as an explicit `draw()` parameter rather than
embedding it in `DisplayBuffer`. After @mspielberg pointed out that
`DisplaySetup` goes out of scope at the end of
`update_display_buffer()`, @mawww accepted exposing it as a
`DisplayBuffer` field for pragmatic reasons.

Since that discussion, the codebase has evolved:

- `Window::last_display_setup()` is now a public accessor, so the
  scoping concern no longer applies.
- `set_cursor` was removed and `cursor_pos` was moved into the
  `draw()` signature (3dd6f30d), establishing a precedent for
  passing display metadata as explicit `draw()` parameters.

Given these changes, this implementation passes `widget_columns` as
an explicit parameter to `draw()` — consistent with @mawww's
original design preference and the `cursor_pos` precedent. If you
would prefer the `DisplayBuffer` field approach instead, I am happy
to adjust.

## Changes

- Add `ColumnCount widget_columns` parameter to
  `UserInterface::draw()` and all implementations
- Pass `widget_columns` from `Window::last_display_setup()` in
  `Client::redraw_ifn()`
- Update remote protocol serialization
- Update `doc/json_ui.asciidoc`
- Update all test expectations


mawww:
The change looks reasonable, however I wonder if the use case still exists as @mspielberg was not active #4737 since 2022, do you have that use case yourself ?

Yus314:
Yes, I do. I'm building [Kasane](https://github.com/Yus314/kasane),
an alternative Kakoune front-end in Rust using `kak -ui json`, and
have been tracking #4737 for this.

Kasane has a color-preview plugin that detects color codes in the
buffer and shows interactive overlays. It currently has to use a
[search-based workaround](https://github.com/Yus314/kasane/blob/e5e0a0caa4851c16d9f56cb4e9dc026fadd59543/kasane-core/src/plugins/color_preview.rs#L485-L496)
to avoid display-vs-buffer offset issues, because the draw atoms
may contain leading widget columns that don't correspond to actual
buffer content. Mouse coordinate mapping has the same problem —
the front-end cannot reliably tell where the buffer content starts
within each line.

Since Kakoune already computes this value internally, exposing it
in the draw call would make these cases work correctly instead of
by heuristic. The explicit parameter approach here matches the
`cursor_pos` precedent nicely.

Issue #2019, opened 2018-04-29
Limitations of the JSON UI


Screwtapello:
I've spent a little time recently trying to build [a GTK-based front-end](https://gitlab.com/Screwtapello/kakoune-gtk) on top of Kakoune's JSON UI feature. It's not nearly complete, but I've already bumped into a few limitations of the JSON interface Kakoune exposes, and I figured I'd note them down here.

## No mouse event support

The JSON interface has a `keys()` message that tells Kakoune that keys were pressed, but it only supports keys that can be used with the `map` command—there's no way to tell Kakoune about mouse events, including scroll events and focus events.

## No clipboard integration support

One of the big selling points of the original Macintosh, and widely available in every GUI since, is the universal clipboard shared between all applications. It's always been difficult to make terminal applications interact with a GUI clipboard, so clipboard integration is a big potential benefit of a GUI-native Kakoune client.

Sadly, the JSON UI protocol does not provide any clipboard-related events in either direction, so a GUI-based Kakoune client is exactly as limited as the terminal UI.

## No mechanism for UI-specific events

Let's say my graphical UI wants to offer a "font" option, so people can switch what font is used. The obvious choice would be to add it to `ui_options` (that's what it's for!) but if the user changes that option, manually or via `kakrc`, there's no way for the UI to be notified and update the screen.

Another example: only the custom UI knows how to launch another instance of the custom UI (consider a front-end that does something like `ssh remotehost kak -ui json`), so in order for commands like `:new` to be possible, there needs to be a way for the server to tell the UI to open a new window.

## Atoms should include their character-cell width

The `draw()` message sends a terminal-sized block of text in the form of a sequence of `Atom` objects, where the character-cell width of all the Atoms on a line should add up to the total width. However, "character-cell width" is calculated from the data tables released by the Unicode consortium and different applications might have different versions of that data (and even if they use the same data, they still have to choose whether "ambiguous width" characters are 1 or 2 cells, see #1941), so when it comes to drawing Atoms to the screen, it doesn't so much matter what *I* think the width should be, or the Unicode database in *my* front-end thinks, I'll only get a sensible display if I follow what *Kakoune* thinks the width should be, and the only way for me to know for sure what Kakoune thinks the width should be is for Kakoune to include that data in each Atom.

## JSON UI protocol fundamentally tied to terminal UI

[...]

PR #5458, opened 2026-03-15
Add style parameter to draw_status in json-ui

### LINKED ISSUE #5428: [REQUEST] No way to distinguish status line context (command/search/info) in JSON UI
### Feature

Following up issue 2 in #5400 

I'm creating a Kakoune frontend using the JSON UI. As noted in the former issue, I want to achieve to following in my frontend:
- Command prompts (:) appear centered
- Status messages (:echo) appear bottom-left
- Search prompts (/) appear top-right

However, what the status line is used for (prompt/status messages/search) is not communicated to the frontend.

There are some ways to infer what the status line is used for, but they are not perfect. One way is to check if the prompt is empty or not, which will differentiate status messages from command and search prompts. However, it does not differentiate between command and search prompts. You could try to use the menu style as an indicator, but this does not work when the menu is hidden (it doesn't seem reliable to use the menu style from when the menu was previously shown either). You could also try scraping the prompt for words like "search" or "select", but this is obviously not a good solution.

I still like my proposed solution in #5400 of adding a status_style parameter to draw_status, like menu_style. I'm open for other suggestions of course.

### Usecase

_No response_

Yus314:
Closes #5428

Adds a `StatusStyle` enum and a 6th parameter `style` (string) to
the `draw_status` JSON UI message, following the `MenuStyle` /
`InfoStyle` pattern.

Values: `"command"` (`:` prompt), `"search"` (`/`/`?` prompt),
`"prompt"` (general prompt), `""` (normal status).

## Use case

I'm building [Kasane](https://github.com/Yus314/kasane), a Kakoune
frontend in Rust using `kak -ui json`. Currently, Kasane can only
detect that a prompt is active (via `cursor_pos >= 0`), but cannot
distinguish what kind of prompt it is. Cursor style inference relies
on string matching against the mode line, which breaks when users
customize `modelinefmt`.

With this parameter, frontends can reliably provide context-aware
rendering — e.g. different placement or styling for command vs search
prompts, search-specific progressive highlighting, or distinguishing
interactive prompts from echo messages.

This is the same class of limitation reported in #2019 by multiple
frontend developers (kakoune-gtk, kakoune-qt).

## Design

- Adds `PromptFlags::Command` to explicitly tag command prompts,
  avoiding fragile string matching on the prompt text
- `StatusStyle` enum in `user_interface.hh`, matching the existing
  `MenuStyle` / `InfoStyle` convention
- Follows the same approach as #5455 (`widget_columns`): add the
  parameter to the virtual method and all implementations, update
  JSON and remote serialization, documentation, and tests

Backward compatible — existing frontends that don't parse the 6th
parameter continue to work.


src/json_ui.cc
```diff
@@ -109,6 +109,18 @@ String to_json(InfoStyle style)
     return "";
 }
 
+String to_json(StatusStyle style)
+{
+    switch (style)
+    {
+        case StatusStyle::Status: return R"("")";
```
mawww: Why use an empty string here instead of "status" ?

Yus314: Good point, I'll change it to `"status"` for consistency with `MenuStyle` / `InfoStyle`.

PR #5459, opened 2026-03-15
git.kak: improve "git commit" wrapper

phillipwood:
Improve the wrapper for `git commit` so that it runs `git commit` in the background with `GIT_EDITOR` set to a script that calls back into the kakoune session that started it. This stops kakoune prompting the user to edit the commit message when the commit cannot be created due to a failing `pre-commit` hook and avoids having to second guess which options ask the user to edit the message.

### DIFF
--- a/rc/tools/blocking-editor-in-client
+++ b/rc/tools/blocking-editor-in-client
@@ -1,0 +1,50 @@
+die() {
+    printf "%s\n" "$*" >&2
+    exit 1
+}
+
+# Nested levels of kakoune single quote quoting
+kak_quote() {
+    level=${2:-1}
+    Q="'"
+    while [ ${level} -gt 1 ]; do
+        Q="${Q}${Q}"
+        level=$((${level} - 1))
+    done
+
+    printf "%s\n" "$1" | sed "s/'/${Q}${Q}/g; 1s/^/${Q}/; \$s/\$/${Q}/"
+}
+
+[ $# -eq 4 ] ||
+    die "usage: blocking-editor-in-client <fifo-dir> <kak-session> <kak-client> <filename>"
+
+fifo_dir="$1"
+kak_session="$2"
+kak_client="$3"
+filename="$4"
+response_fifo="${fifo_dir}/response_fifo" &&
+printf %s "
+    try '
+        eval -client ${kak_client} ''
+            edit $(kak_quote "${filename}" 3)
+            hook buffer BufWritePost .* ''''
+                echo -to-file $(kak_quote "${response_fifo}" 4) -end-of-line edited
+                remove-hooks buffer giteditor
+                delete-buffer
+            ''''
+            hook -group giteditor buffer BufClose .* ''''
+                echo -to-file $(kak_quote "${response_fifo}" 4) -end-of-line cancelled
+            ''''
+        ''
+    ' catch '
+        echo -to-file $(kak_quote "${response_fifo}" 2) -end-of-line failed
+    '" | kak -p "${kak_session}" ||
+        die "could not contact kak session ${kak_session}"
+# wait for the user to edit the commit message
+read status <"${response_fifo}"
+if [ "${status}" = cancelled ]; then
+    >"${fifo_dir}/cancelled"
+    die "editing cancelled"
+elif [ "${status}" != edited ]; then
+    die "could not open file '${filename}'"
+fi
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -542,29 +542,42 @@ define-command -params 1.. \
     }
 
     commit() {
-        # Handle case where message needs not to be edited
-        if grep -E -q -e "-m|-F|-C|--message=.*|--file=.*|--reuse-message=.*|--no-edit|--fixup.*|--squash.*"; then
-            if git commit "$@" > /dev/null 2>&1; then
-                echo 'echo -markup "{Information}Commit succeeded"'
+        if ! fifo_dir="$(mktemp -d "${TMPDIR:-/tmp}/kak-git-commit-XXXXXX")" ||
+           ! mkfifo "${fifo_dir}/response_fifo"
+        then
+            echo "fail %{failed to run git commit (see *debug* buffer)}"
+            [ -n "${fifo_dir}" ] && rmdir "${fifo_dir}"
+            exit 1
+        fi
+        {
+            trap 'rm -r "${fifo_dir}"' EXIT
+            export GIT_EDITOR="$(git rev-parse --sq-quote \
+                "${KAKOUNE_POSIX_SHELL:-/bin/sh}" \
+                "${kak_runtime}/rc/tools/blocking-editor-in-client" \
+                "${fifo_dir}" "${kak_session}" "${kak_client}")"
+            failed=false
+            if err="$(git commit "$@" 2>&1)"; then
+                cmd="eval -try-client ${kak_client} %{
+                    echo -markup '{Information}Commit succeeded'
+                }"
+            elif [ -f "${fifo_dir}/cancelled" ]; then
+                cmd="eval -try-client ${kak_client} %{
+                    echo -markup '{Information}Commit cancelled'
+                }"
             else
-                echo 'fail Commit failed'
+                failed=true
+                cmd="eval -try-client ${kak_client} %{
+                    try %{
+                        edit! -fifo $(kakquote "${fifo_dir}/response_fifo") '*git-commit*'
+                    }
+                    echo -markup '{Error}Commit failed'
+                }"
             fi
-            exit
-        fi <<-EOF
-			$@
-		EOF
-
-        # fails, and generate COMMIT_EDITMSG
-        GIT_EDITOR='' EDITOR='' git commit "$@" > /dev/null 2>&1
-        msgfile="$(git rev-parse --git-dir)/COMMIT_EDITMSG"
-        printf %s "edit '$msgfile'
-              hook buffer BufWritePost '.*\Q$msgfile\E' %{ evaluate-commands %sh{
-                  if git commit -F '$msgfile' --cleanup=strip $* > /dev/null; then
-                     printf %s 'evaluate-commands -client $kak_client echo -markup %{{Information}Commit succeeded}; delete-buffer'
-                  else
-                     printf 'evaluate-commands -client %s fail Commit failed\n' "$kak_client"
-                  fi
-              } }"
+            printf %s "${cmd}" | kak -p "${kak_session}"
+            if [ ${failed} = true ]; then
+                printf %s "$err" >"${fifo_dir}/response_fifo"
+            fi
+        } 2>/dev/null 1>&2 </dev/null &
     }
 
     blame_jump() {


rc/tools/git.kak
```diff
@@ -542,29 +542,93 @@ define-command -params 1.. \
     }
 
     commit() {
-        # Handle case where message needs not to be edited
-        if grep -E -q -e "-m|-F|-C|--message=.*|--file=.*|--reuse-message=.*|--no-edit|--fixup.*|--squash.*"; then
-            if git commit "$@" > /dev/null 2>&1; then
-                echo 'echo -markup "{Information}Commit succeeded"'
+        if ! fifo_dir="$(mktemp -d "${TMPDIR:-/tmp}/kak-git-XXXXXX")" ||
```
krobelus: > This commit runs "git commit" in the background with an editor that
calls back into the kakoune client that started it

yeah, I've been using that (more robust) approach for a while for some Git and jj commands.
I extracted most of the logic into an standalone script, called `rc/tools/blocking-editor-in-client`; used in [jj.kak](https://github.com/krobelus/kakoune/commit/jj~). Another (earlier) implementation used in some 3rd party git commands lives [here](https://github.com/krobelus/kakoune-boost/blob/77eaaacd2a38882dd34e84ce2646ac7d6c8db6e4/git.kak#L99-L145).

Would be great if we can join forces and share some code (which entails extracting the parts that are not specific to Git).

I'll try find time to review later. The reason why I didn't suggest this improvement to `:git commit` is that the failure scenario in my implementation was a bit worse; if either fifo reader/writer dies, there's some blockage which may leave behind Git's `index.lock`. But I'd need to recheck if that's actually a problem in practice.

phillipwood: @krobelus thanks for the links, it would be good to have some shared code for setting `GIT_EDITOR`. The editor script here writes some files that are read after `git commit` has finished to help figure out if the commit failed because of an empty file, hook failure or something else so the caller would need to pass in a directory name to where those files should be written.

With regard to the fifo reader or writer dying it is certainly a concern but as far as I can see git does not lock the index while the commit message is being edited (`GIT_EDITOR='echo "$(git rev-parse --git-dir)"/*'` does not print any lock files). If kak crashes while the commit message is being edited then `git commit` will never exit which isn't great, I don't know if we could have some kind of watchdog that periodically pings kakoune and kills git if it does not get a response. On the other hand I've been using kakoune for nearly a year and it has only crashed on me once so maybe it is not really a problem in practice. I've been using this code for several months without noticing any problems. It avoids reading from the fifo if `kak -p` fails and if the kakoune code that opens the commit message file and sets up the BufPostWrite and BufClose hooks fails because the client has disappeared or it cannot open the file it still writes to the fifo to unblock the reader.

BTW congratulations to you and the other fish contributors on the rust rewrite - from my point of view as a user it has been a very smooth transition.

rc/tools/git.kak
```diff
@@ -541,30 +541,186 @@ define-command -params 1.. \
         fi
     }
 
-    commit() {
```
krobelus, line 544: looks like solid work, I gotta run now but I think I'd be okay with this more-or-less as-is.
Not yet sure about the second commit.

rc/tools/git.kak
```diff
@@ -541,30 +541,186 @@ define-command -params 1.. \
...
-            if git commit "$@" > /dev/null 2>&1; then
-                echo 'echo -markup "{Information}Commit succeeded"'
+    classify_new_commit() {
+        {
+            git log --no-walk=unsorted --format='%H:%T:%P:%s' \
+                HEAD HEAD@{1} -- ||
+            git log --no-walk=unsorted --format='%H:%T:%P:%s' \
+                HEAD --
+        } 2>/dev/null | (
+            IFS=:
```
krobelus: maybe unset IFS after we're done using it? Would feel cleaner, no?

rc/tools/git.kak
```diff
@@ -541,30 +541,186 @@ define-command -params 1.. \
...
+
+    empty_commit_message_is_allowed() {
+        ret=1
+        for arg; do
+            case $arg in
+                --) break;;
+                # Allow a prefix match but don't be confused by "--allow-empty"
+                --allow-empty-*)
+                    case --allow-empty-message in
+                        "$arg"*) ret=0;;
```
krobelus: this logic is pretty wild haha but I guess it's the best way

rc/tools/git.kak
```diff
@@ -541,30 +541,186 @@ define-command -params 1.. \
...
+    commit() {
+        if ! fifo_dir="$(mktemp -d "${TMPDIR:-/tmp}/kak-git-XXXXXX")" ||
+           ! mkfifo "${fifo_dir}/response_fifo"
+        then
+            echo "fail %{failed to run git commit (see *debug* buffer)}"
+            [ -n "${fifo_dir}" ] && rmdir "${fifo_dir}"
+            exit 1
+        fi
+        {
+            trap 'rm -r "${fifo_dir}"' EXIT INT
```
krobelus: If I do

	dash -c "trap 'echo CLEANUP' EXIT INT; sleep 1"

and run ctrl-c, it will print CLEANUP twice.
Such a double cleanup is wrong because (in theory) `$fifo_dir` may have been created by another process after the first `rm`.
Also, I don't think there's a realistic scenario where we receive INT here.
So maybe remove that, or make it idempotent.

rc/tools/git.kak
```diff
@@ -541,30 +541,186 @@ define-command -params 1.. \
...
+            GIT_EDITOR='
+                die() {
```
krobelus: I think we should extract this inline script into a separate file,
both to reduce quote nesting, and also to make it more reusable.
That shouldn't have user-visible effects.
The invocation can look something like

	GIT_EDITOR="${KAKOUNE_POSIX_SHELL:-/bin/sh} ${kak_runtime}/rc/tools/blocking-editor-in-client ${fifo_dir} ${kak_session} ${kak_client}"

(Rather than passing environment variables, I prefer passing positional arguments,
because they have a more narrow and more explicit scope.
I guess kak_session etc are already exported.. though if we were to spawn a potentially long-running process with children, then we should unexport)

rc/tools/git.kak
```diff
@@ -541,30 +541,186 @@ define-command -params 1.. \
...
+                    printf "%s\n" "$*" >&2
+                    exit 1
+                }
 
-        # fails, and generate COMMIT_EDITMSG
-        GIT_EDITOR='' EDITOR='' git commit "$@" > /dev/null 2>&1
-        msgfile="$(git rev-parse --git-dir)/COMMIT_EDITMSG"
-        printf %s "edit '$msgfile'
-              hook buffer BufWritePost '.*\Q$msgfile\E' %{ evaluate-commands %sh{
```
krobelus, line 561: One thing that has always bugged me about this hook approach is that `:wa` will trigger it too.
Of course that's unrelated and with `jj` it's not really a big deal because `COMMIT_EDITMSG` type buffers can be written multiple times.. but if a user accidentally has multiple for the same commit, chaotic results will ensue.

rc/tools/git.kak
```diff
@@ -541,30 +541,186 @@ define-command -params 1.. \
...
+                }
+
+                editor() {
+                    response_fifo="${fifo_dir}/response_fifo" &&
+                    # The quoting here is slightly awkward but it is works when
+                    # the path of the file being edited contains single quotes
+                    # or unbalanced braces.
+                    printf %s "
+                        try ${Q1}
+                            eval -client $(kak_quote "${kak_client}" 2) ${Q2}
```
krobelus: quoting `kak_client` is like going the extra mile; nothing wrong with it; technically it's not necessary since client names must be valid identifiers

rc/tools/git.kak
```diff
@@ -541,30 +541,186 @@ define-command -params 1.. \
...
+                        ${Q1} catch ${Q1}
+                            echo -to-file $(kak_quote "${response_fifo}" 2) -end-of-line failed
+                        ${Q1}" | kak -p "${kak_session}" ||
+                            die "could not contact kak session ${kak_session}"
+                    # create a sed expression to print the first uncommented,
+                    # non-empty line before any scissors line.
+                    comment_re="$(get_comment_re)"
+                    re="/^${comment_re} -\{8,\} >8 -\{8,\}\$/q
+                        /^${comment_re}.*/d
+                        /[^ 	]/{p;q;}"
```
krobelus: it's unfortunate that we need to know about some of the COMMIT_EDITMSG format intricacies here.
Could we instead have `git commit` return a special failure exit status (e.g. 2) when the message was empty?

Alternatively, given that the output is like

	$ git commit --allow-empty -m ''
	Aborting commit due to empty commit message.

and we already show that in a `*git-commit*` buffer, do we even need the second commit?

rc/tools/git.kak
```diff
@@ -541,30 +541,186 @@ define-command -params 1.. \
...
+                        >"${fifo_dir}/empty-message"
+                    fi
+                }; editor'
+            export fifo_dir GIT_EDITOR
+            failed=0
+            if err="$(git commit "$@" 2>&1)"; then
+                cmd="eval -try-client '${kak_client}' '
+                    echo -markup ''{Information}{\}$(classify_new_commit | sed "s/'/''''/g")''
+                '"
+            elif [ -f "${fifo_dir}/cancelled" ]; then
```
krobelus, line 563: > That pitfall is avoided

Imperative mood would have been clearer.

rc/tools/git.kak
```diff
@@ -541,30 +541,186 @@ define-command -params 1.. \
...
+                failed=1
+                cmd="eval -try-client '${kak_client}' %{
+                    try %{
+                        edit! -fifo $(kakquote "${fifo_dir}/response_fifo") '*git-commit*'
+                    }
+                    echo -markup '{Error}Commit failed'
+                }"
+            fi
+            printf %s "${cmd}" | kak -p "${kak_session}"
+            if [ ${failed} = 1 ]; then
```
krobelus: the inconsistent use of true/false and 0/1 is strange

rc/tools/git.kak
krobelus: right, this seems solid enough for git commands that don't need to lock the index

> On the other hand I've been using kakoune for nearly a year

in case you haven't, try `:git apply` for staging/reverting and also recursive `:git blame` / `:git blame-jump` (the latter we should rename, maybe to `:git blame-show-diff` or `:git blame-show-in-diff` ?). I'd also like to rename `:git show-blame` and `:git hide-blame`.

rc/tools/git.kak
mawww: Yeah, would be good to make this more consistent

mawww:
I this this needs a copyright waiver commit as well, I dont think I have one from you yet.

Not too sure about the two last commits. Regarding the first one, I am not convinced we should deviate from git's behaviour, if the commit message is empty and git fails to commit, showing this as an error to the user sounds reasonable to me, and we need quite a bit of extra code to deviate from this. For the second one, it is convenient to show the commit message, but similarly I find the amount of extra code required to achieve this a bit hard to justify.

rc/tools/git.kak
phillipwood: I didn't bother because it is set in a subshell, I'll change it save and restore `IFS` either side of the `read`s.

rc/tools/git.kak
phillipwood: Oh, thanks for checking that - I'll remove `INT`

rc/tools/git.kak
phillipwood: Having it a separate file is a good idea. The positional arguments will be parsed by the shell that git executes to run the editor so I'll quote them to avoid any nasty surprises. It's unlikely that `${kak_runtime}` or `${fifo_dir}` contain any special characters but they're under the control of the user so we should be careful.

rc/tools/git.kak
phillipwood: I guess we could have a separate command to say "I've finished editing the commit message - please save and close it". Magit's commit message mode has that. I seems a bit tangential to the changes here though.

rc/tools/git.kak
phillipwood: `:git apply` is one of my favorite things in kakoune. I've found the blame support handy as well.

rc/tools/git.kak
phillipwood: Ah, I wasn't sure what the rules were for the client name - I'll remove the quoting as there is no point forking `sed` if we don't need to.

rc/tools/git.kak
phillipwood: I could look at getting `git commit` to be more helpful with its exit codes, though in general `git` commands do not have reliable exit codes beyond zero for success and non-zero for failure. I'd like to treat an empty message separately from other errors as the user choosing to cancel the commit feels quite different to something going wrong that prevents git from creating a commit.

rc/tools/git.kak
phillipwood: Sure

rc/tools/git.kak
phillipwood: I'll change to true/false here

rc/tools/git.kak
phillipwood: Yes, the logic feels a bit back-to-front but I couldn't think of a better way to do a prefix match.

rc/tools/git.kak
krobelus: For me personally, this error scenario is not super important - when I want to cancel the commit step, I run `:delete-buffer` rather than saving an empty buffer. Seems like a more intuitive way

phillipwood:
> I this this needs a copyright waiver commit as well, I dont think I have one from you yet.

I'll add that

> Not too sure about the two last commits. Regarding the first one, I am not convinced we should deviate from git's behaviour, if the commit message is empty and git fails to commit, showing this as an error to the user sounds reasonable to me, and we need quite a bit of extra code to deviate from this.

git fails because the user asked it to cancel the commit. That feels quite different to me to git failing for some other reason. If I cancel the commit it is a pain to have to close the `*git-commit*` buffer, rather than kakoune just printing a status message. I see deleting the buffer and writing an empty buffer as two different ways to do the same thing so it makes sense to me to handle them in the same way. git tells the user that an empty message will cancel the commit so it seems strange to treat it as an error. If you really don't like it I can drop this commit.

>  For the second one, it is convenient to show the commit message, but similarly I find the amount of extra code required to achieve this a bit hard to justify.

I added that commit after creating some bad fixup commits by a mistake and I've found it really useful to be able to check that I've fixed up the commit I meant to. I've also found it useful with `:git commit --amend --no-edit` to check I've amended the commit I intended to.
I'm quite reluctant to drop this one as I've found it so useful.

rc/tools/git.kak
phillipwood: I agree it's less intuitive to save an empty buffer, but that is the advice git gives in the commit message template, so if the user does not know they can cancel the commit with `:delete-buffer` they'll likely create an empty file.

mawww:
I am not too sure about that second commit, git returns a non-zero value and emits an error message on stderr, I think we should not special case this. I think it would be better to drop that commit, and have `git commit` (or other git commands for that matter) show what the git error message to the user instead of asking them to look into the `*debug*` buffer.

For the third commit, I like the feature, but a 40 line shell function seems like a hefty price to pay for it, I wonder if we can simply echo the first line of git's output. I view git.kak as a thin wrapper to make interacting with git easy from Kakoune, but I am a bit worried about it becoming an opinionated interface to git on top of Kakoune, which should likely be a separate plugin. Ideally, I would have Kakoune just echo the first output line of git, and have git improve its output to distinguish between amend/rewords/empty/root commits.

phillipwood:
I've added a copyright waiver and dropped the last two patches. Here are the changes in the remaining patch 
 - Added copyright waiver
 - Split out editor script into a separate file that takes positional arguments rather than using environment variables.
 - Removed unnecessary quoting around `${kak_client}`
 - Use true/false instead of 0/1
 - Only cleanup on `EXIT`

rc/tools/blocking-editor-in-client
```diff
@@ -0,0 +1,50 @@
...
+            ''''
+        ''
+    ' catch '
+        echo -to-file $(kak_quote "${response_fifo}" 2) -end-of-line failed
+    '" | kak -p "${kak_session}" ||
+        die "could not contact kak session ${kak_session}"
+# wait for the user to edit the commit message
+read status <"${response_fifo}"
+if [ "${status}" = cancelled ]; then
+    >"${fifo_dir}/cancelled"
```
mawww, line 46: Hum, dont you need a `:` in front of the redirection ?

rc/tools/blocking-editor-in-client
```diff
@@ -0,0 +1,50 @@
+die() {
```
mawww, line 1: A shebang line would be good here

rc/tools/blocking-editor-in-client
```diff
@@ -0,0 +1,50 @@
...
+            ''''
+            hook -group giteditor buffer BufClose .* ''''
+                echo -to-file $(kak_quote "${response_fifo}" 4) -end-of-line cancelled
+            ''''
+        ''
+    ' catch '
+        echo -to-file $(kak_quote "${response_fifo}" 2) -end-of-line failed
+    '" | kak -p "${kak_session}" ||
+        die "could not contact kak session ${kak_session}"
+# wait for the user to edit the commit message
```
mawww, line 43: Seems this script is expected to be a more general utility so it is weird to refer to "commits" in this comment.

rc/tools/blocking-editor-in-client
```diff
@@ -0,0 +1,50 @@
...
+kak_client="$3"
+filename="$4"
+response_fifo="${fifo_dir}/response_fifo" &&
+printf %s "
+    try '
+        eval -client ${kak_client} ''
+            edit $(kak_quote "${filename}" 3)
+            hook buffer BufWritePost .* ''''
+                echo -to-file $(kak_quote "${response_fifo}" 4) -end-of-line edited
+                remove-hooks buffer giteditor
```
mawww, line 32: the reference to git in the hook name is surprising for a general utility script.

rc/tools/blocking-editor-in-client
```diff
@@ -0,0 +1,50 @@
...
+}
+
+# Nested levels of kakoune single quote quoting
+kak_quote() {
+    level=${2:-1}
+    Q="'"
+    while [ ${level} -gt 1 ]; do
+        Q="${Q}${Q}"
+        level=$((${level} - 1))
+    done
```
mawww, line 13: How about:
```sh
Q="'"
for _ in $(seq ${2:0}); do
    Q="${Q}${Q}"
done
```
instead ?

rc/tools/blocking-editor-in-client
phillipwood: As far as I know it's not required - git's test suite uses this style and is portable across different POSIX shell implementations.

rc/tools/blocking-editor-in-client
phillipwood: The file isn't executable as @krobelus suggested running it with `${KAKOUNE_POSIX_SHELL:-/bin/sh}`. I can make it executable if you want but there doesn't seem to be much point in adding a shebang without that - or do you want the shebang so that it triggers the correct syntax highlighting?

rc/tools/blocking-editor-in-client
phillipwood: I forgot to update this when I split it out into a separate file - I'll update it.

rc/tools/blocking-editor-in-client
phillipwood: I forgot to update this when I split it out into a separate file - I'll update it.

rc/tools/blocking-editor-in-client
phillipwood: It adds another process which I though was discouraged after seeing the recommendation to avoid `$(basename "${var}")` in docs/writing_scripts.asciidoc

rc/tools/blocking-editor-in-client
krobelus: TIL; that looks like good style then. Maybe search for `\s: [<>]` to remove other occurrences of `: `.

rc/tools/blocking-editor-in-client
krobelus: maybe pass the quote directly, like `$(kak_quote "${response_fifo}" "''''''''")`

rc/tools/blocking-editor-in-client
krobelus: FWIW my approach to run the interpreter explicitly is for consistency with a362c8a145 (Perl files are not always executable, 2023-12-04).
Our `rc/*/*.pl` are still executable though, so that's an inconsistency we should get rid of.

It might be good to explicitly declare the language, and shebang seems the best way to do that.
`*.sh` might be mistaken as Bash script.
But no strong opinion here, I guess POSIX is the default

rc/tools/blocking-editor-in-client
phillipwood: Nice idea, I like it. One of the benefits of splitting out the editor script into its own file is that we can write `'` directly rather than messing about with `${SQ}`.

rc/tools/blocking-editor-in-client
phillipwood: I think passing the file to the interpreter is the only way to support `${KAKOUNE_POSIX_SHELL}`. I'll add a shebang for the syntax highlighting and look at opening a separate PR to remove the executable bit from the perl files.

krobelus: > I think passing the file to the interpreter is the only way to support ${KAKOUNE_POSIX_SHELL}

I wonder if we really need to support it, given that we already don't tell Git to use it.
Something like

    GIT_EDITOR='echo $(touch /tmp/pwned)' git commit --allow-empty

already invokes its own idea of POSIX shell, so given that we do

> export GIT_EDITOR="$(git rev-parse --sq-quote \
>     "${KAKOUNE_POSIX_SHELL:-/bin/sh}" \
>     "${kak_runtime}/rc/tools/blocking-editor-in-client" \
>     "${fifo_dir}" "${kak_session}" "${kak_client}")"

I'm not sure if really care to invoke another shell.
Could we change this to `. "${kak_runtime}/rc/tools/blocking-editor-in-client" $args...` ?
I haven't tried.
But again, either way is fine by me, appreciate it.

rc/tools/git.kak
phillipwood: > right, this seems solid enough for git commands that don't need to lock the index

Unfortunately I've just found that although `git commit` does not lock the index `git commit my-file` does so I wonder if we'd be better doing something like the diff below. It keeps the current approach of running `git commit` twice but avoids guessing when git would edit the message, detects if `git` fails due to the pre-commit hook failing and works with options like `--fixup=(amend|reword):<commit>` that are incompatible with `-F`

```diff
diff --git a/rc/tools/git.kak b/rc/tools/git.kak
index 656dd73f..49aed45b 100644
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -171,7 +171,8 @@ define-command -params 1.. \
         }
     }
     kakquote() {
-        printf "%s" "$1" | sed "s/'/''/g; 1s/^/'/; \$s/\$/'/"
+        Q=${2:-"'"}
+        printf "%s" "$1" | sed "s/'/${Q}${Q}/g; 1s/^/${Q}/; \$s/\$/${Q}/"
     }
 
     show_git_cmd_output() {
@@ -542,29 +543,55 @@ define-command -params 1.. \
     }
 
     commit() {
-        # Handle case where message needs not to be edited
-        if grep -E -q -e "-m|-F|-C|--message=.*|--file=.*|--reuse-message=.*|--no-edit|--fixup.*|--squash.*"; then
-            if git commit "$@" > /dev/null 2>&1; then
-                echo 'echo -markup "{Information}Commit succeeded"'
-            else
-                echo 'fail Commit failed'
-            fi
-            exit
-        fi <<-EOF
-			$@
-		EOF
-
-        # fails, and generate COMMIT_EDITMSG
-        GIT_EDITOR='' EDITOR='' git commit "$@" > /dev/null 2>&1
-        msgfile="$(git rev-parse --git-dir)/COMMIT_EDITMSG"
-        printf %s "edit '$msgfile'
-              hook buffer BufWritePost '.*\Q$msgfile\E' %{ evaluate-commands %sh{
-                  if git commit -F '$msgfile' --cleanup=strip $* > /dev/null; then
-                     printf %s 'evaluate-commands -client $kak_client echo -markup %{{Information}Commit succeeded}; delete-buffer'
-                  else
-                     printf 'evaluate-commands -client %s fail Commit failed\n' "$kak_client"
-                  fi
-              } }"
+        err="$(GIT_EDITOR='f() {
+            printf %s "$1" >"${kak_response_fifo%/*}/msgfile-name"
+            exit 1
+        }; f' git commit "$@" 2>&1)"
+        if [ $? -eq 0 ]; then
+            echo 'echo -markup %{{Information}Commit succeeded}'
+        elif [ -f "${kak_response_fifo%/*}/msgfile-name" ];   then
+            msgfile="$(cat "${kak_response_fifo%/*}/msgfile-name")"
+            sh_quoted_msgfile="$(git rev-parse --sq-quote "${msgfile}")"
+            printf %s "
+                edit $(kakquote "${msgfile}")
+                hook -group gitcommit buffer BufWritePost .* 'evaluate-commands %sh$(kakquote "
+                    kakquote() {
+                        Q=\${2:-\"'\"}
+                        printf \"%s\" \"\$1\" | sed \"s/'/\${Q}\${Q}/g; 1s/^/\${Q}/; \\\$s/\\\$/\${Q}/\"
+                    }
+                    echo 'remove-hooks buffer gitcommit'
+                    mv ${sh_quoted_msgfile} ${sh_quoted_msgfile}.edited >/dev/null
+                    err=\"\$(GIT_EDITOR='f() {
+                        mv ${sh_quoted_msgfile}.edited ${sh_quoted_msgfile}
+                    }; f' git commit $(git rev-parse --sq-quote "$@") 2>&1)\"
+                    if [ \$? -eq 0 ]; then
+                        printf %s 'evaluate-commands -client ${kak_client} %{
+                            delete-buffer
+                            echo -markup %{{Information}Commit succeeded}
+                        }'
+                    else
+                        printf %s \"evaluate-commands -client ${kak_client} '
+                            echo -debug %{git commit failed}
+                            echo -debug %{git output <<<}
+                            echo -debug \$(kakquote \"\${err}\" \"''\")
+                            echo -debug %{>>>}
+                            echo -markup %{{Error}Commit failed (see *debug*)}
+                        '\"
+                    fi
+                            " "''")'
+                hook -group gitcommit buffer BufClose .* %{
+                    evaluate-commands -client ${kak_client} %{
+                        echo -markup %{{Information}Commit cancelled}
+                    }
+                }
+            "
+        else
+            printf %s "echo -debug %{git commit failed}
+                       echo -debug %{git output <<<}
+                       echo -debug $(kakquote "${err}")
+                       echo -debug %{>>>}
+                       fail %{Commit failed (see *debug*)}"
+        fi
     }
 
     blame_jump() {
```

PR #5471, opened 2026-04-09
Add viewport info to JSON UI draw message

Yus314:
## Summary

Append `first_buffer_line`, `last_buffer_line`, and `buffer_line_count` parameters to the `draw` JSON-RPC method, enabling alternative frontends to implement scroll bars, minimap indicators, and other viewport-aware UI elements.

This is the viewport information subset extracted from the scroll bar PR (#5304), proposed as an independent protocol extension that benefits all JSON UI consumers regardless of whether a built-in scroll bar is added.

## Motivation

Alternative frontends (e.g. [Kasane](https://github.com/Yus314/kasane)) currently cannot determine the buffer position corresponding to the displayed viewport. The information needed — which buffer lines are visible and how many total lines exist — is available inside Kakoune but not exposed through the JSON-RPC protocol. Without these values, features like scroll bars, minimap position indicators, and viewport-relative overlays require fragile heuristics that break with word-wrap, replace-ranges, and folding.

## Implementation

**New parameters** (appended to existing `draw` positional params):

| Parameter | Type | Source |
|---|---|---|
| `first_buffer_line` | `int` (0-based) | `window.last_display_setup().first_line` |
| `last_buffer_line` | `int` (0-based) | `display_buffer.lines().back().range().begin.line` |
| `buffer_line_count` | `int` | `context.buffer().line_count()` |

**Data source rationale:**
- `first_buffer_line` uses `last_display_setup().first_line` rather than `display_buffer.range().begin.line` because the latter is computed before highlighter passes and becomes stale after wrap/trim/erase operations.
- `last_buffer_line` uses the last display line's range begin, which is updated by `compute_range()` during `trim_from` and reflects the actual last visible buffer line after all highlighter transforms.

**Files changed:**
- `src/user_interface.hh` — Added 3 `LineCount` params to `draw()` virtual interface
- `src/client.cc` — Extract and pass viewport values from correct sources
- `src/json_ui.cc/hh` — Added `to_json(LineCount)` overload; updated draw signature and `rpc_call`
- `src/terminal_ui.cc/hh` — Updated signature (values unused, unnamed params)
- `src/remote.cc` — Updated signature; `send_message`/template dispatch handles serialization automatically
- `src/main.cc` — DummyUI draw override updated
- `doc/json_ui.asciidoc` — Protocol documentation updated
- 74 test script files — Updated draw message assertions with viewport values

**Backward compatibility:** Existing JSON UI consumers that don't parse trailing positional parameters are unaffected. This follows the same "append to end" pattern established by `widget_columns` (#5455) and `draw_status` style (#5458).

## Test plan

- [x] `make` builds with zero new warnings
- [x] All 74 test scripts with `draw` messages updated with correct viewport values
- [x] `make test` passes (656 tests, 0 failures)
- [x] Manual verification: `kak -ui json` emits correct viewport values with word-wrap, replace-ranges, and large buffers

---

*Discussed in [#5304 (comment)](https://github.com/mawww/kakoune/pull/5304#issuecomment-2705870508)*


lobre:
I'm also building an alternative Kakoune GUI, and hit exactly the gap this PR addresses. No reliable way to size a scrollbar thumb or otherwise reflect viewport position without `buffer_line_count`/viewport range data. Would be great to see this land!

PR #5474, opened 2026-04-19
Add support for the Lean programming language

Chris-F5:
This PR adds support for the [Lean Programming Language](https://lean-lang.org/).

Lean editors must support easy insertion of unicode characters. As far as I know, all other lean editor plugins use the same mechanism for this: you type as escape code into the buffer e.g. `\omega` and it gets expanded to `ω`. There is some complexity in ensuring that `\o` does not get expanded to `∘` while you are mid way through typing `\omega`, but that `\o` does get expanded if you put a space after it `\o `.

I have been using these modifications to write Lean for a few months and its working fine for me.

- `lean_abbreviations.json` contains all the abbreviations. I took this from the [VS code plugin](https://github.com/leanprover/vscode-lean4/blob/master/lean4-unicode-input/src/abbreviations.json).
- `lean_abbreviations.pl` reads the json file and produces regexes to help kakoune do the expansion when the user types.
- `lean.kak` detects lean files, invokes `lean_abbreviaions.pl` and adds some basic auto indent.

Lean users should know that [I have added support](https://github.com/kakoune-lsp/kakoune-lsp/pull/860) for the lean 'goal view' to kak-lsp. This is another feature essential for writing Lean proofs.

kaklean


mawww:
I dont think we can just import lean_abbreviation.json from VSCode and put it in the public domain under the Unlicense.

The fact that this filetype requires its own perl script and json file makes me wonder if this should be published as a separate plugin instead of inside Kakoune. I dont have a strong opinion yet but in any case I dont think I can merge this due to the licensing issue.

Chris-F5:
Ah, I did not think about licencing. Just had a quick look and could not find any `lean_abbreviations.json` in the public domain. I guess in that case a separate plugin is the way to go.  I'll make a repo for a kak-lean plugin and link it here.

Chris-F5:
I have created a repo for this to exist as a separate plugin [here](https://github.com/Chris-F5/kakoune-lean).

PR #5476, opened 2026-04-21
corrected shell generated groovy highlighters and static words

allenyade:
when using groovy for the first time with kakoune, i noticed the static words were a bit odd, and println which is highlighted as a builtin in some default rc filetypes (like go) should be highlighted asweell, so i decided to redo the generated highlighters in a, in my opinion, more scalable way, using the same join() function defined in plenty of other highlighters (e.g.:c-family.kak)

before fix:
image

after fix:
image


mawww:
Hello, I think this change is big enough to warrant a copyright waiver commit, and I dont think I have one from you, can you add one to this PR please ? (see the CONTRIBUTING file)

allenyade:
oh i didn't notice the CONTRIBUTING file, here! is that correct?

PR #5479, opened 2026-04-23
share/kakrc: Preserve current colorscheme as an opt

possumvibes:
This change defines a new option `current_colorscheme` and updates the `colorscheme` command to set current_colorscheme on successfully loading a colorscheme. It will allow programatic management of colorschemes based on the last-loaded colorscheme is. If a user sets their colorscheme by `source`-ing the file directly instead of using `colorscheme`, the opt will not be updated, making this opt specifically a convenience for users of the `colorscheme` command.

I have not seen a noticeable increase to startup time with this change when running kakoune with all standard lib in `val%{runtime}` and nothing in `val%{config}`. I tested this by running `time /path/to/development/kak/bin -n -e 'quit!'` repeatedly before and after the change.

### DIFF
--- a/share/kak/kakrc
+++ b/share/kak/kakrc
@@ -1,3 +1,7 @@
+declare-option -hidden \
+    -docstring "Exposes the most recent successfully set colorscheme name." \
+    str colorscheme_current '' 
+
 def -params 1 -docstring "colorscheme <name>: enable named colorscheme" \
     -shell-script-candidates %{
     find -L "${kak_runtime}/colors" "${kak_config}/colors" -type f -name '*\.kak' \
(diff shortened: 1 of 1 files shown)


ficd0 (commented):
I wonder if `colorscheme_current` would be a more appropriate name for the option, since it follows the convention of prefixing the option name with the command/module it belongs to. 

I would also add a docstring, they are still helpful even when the option is hidden. On which note, I'm on the fence about whether or not the option should be hidden at all. 

Non-hidden would allow for better discoverability, but it may also mislead users into thinking directly setting the option achieves something.

possumvibes:
> I wonder if colorscheme_current would be a more appropriate name for the option, since it follows the convention of prefixing the option name with the command/module it belongs to.

> I would also add a docstring, they are still helpful even when the option is hidden.

Good call, I'll make both those changes!

> I'm on the fence about whether or not the option should be hidden at all.
> Non-hidden would allow for better discoverability, but it may also mislead users into thinking directly setting the option achieves something.

I chose hidden for precisely this reason, since colorschemes are just sourced files and there's no true colorscheme "setting". Definitely open to arguments for making it non-hidden, but since it's effectively a "read-only" option, I'm really not sure if there's a benefit. It may well also be the case that this patch is too much of a stop-gap over how colorschemes actually work and how we'd like to be able to consider changing colorschemes in logic, and I would definitely understand if mawww didn't think this made sense to merge.

Issue #3572, opened 2020-06-29
Allow custom mappings in the "combine selections" menu


Screwtapello:
### Feature 

The `:map`command takes a "mode" argument that restricts the availability of the mapping. As well as the obvious modes (insert, normal) and some UI-based modes (prompt, menu, user) there's specific modes for the various normal-mode commands that are prefixes: goto, view, and object. A user can add a mapping (with a docstring) to those modes, and the mapping will show up alongside the built-in/official mappings, so plugins can extend Kakoune in a way that feels completely normal and familiar.

However, there's one normal-mode prefix that does *not* accept custom mappings: the "combine selections" menu that appears after you press `<a-z>` or `<a-Z>`.

The specific list of changes requested by this issue are:

  - the `:map` command should accept a new "mode" ("combine_selections"? "combsel"?) that represents the "combine selections" menu
  - mappings in the "combine_selections" mode should have access to the `%val{count}` and `%val{register}` expansions (like mappings in "object" do)
  - mappings will probably need an additional `%val{combine_direction}` expansion to distinguish between `<a-z>` (from register) and `<a-Z>` (to register)

### Usecase 

People coming to Kakoune for the first time frequently expect the "combine selections" menu to do mathematical-style set operations (union, intersection, left-subtraction, right-subtraction) on the set of selected characters:

  - https://github.com/mawww/kakoune/issues/3560
  - https://github.com/mawww/kakoune/issues/3504
  - https://discuss.kakoune.com/t/selection-combining-operator-u-union-is-little-bit-misleading/724
  - https://discuss.kakoune.com/t/invert-selection/1037
  - https://github.com/mawww/kakoune/issues/2850

However, the existing items in the combine selections menu don't work like that. There's good reasons *why* they don't work like that, but experimenting with different kinds of selection operations is difficult because they must be implemented in C++.

If custom mappings could be added to the "combine selections" menu, people could implement different selection operations themselves. If any of those experiments turned out to be particularly useful, they could be ported to C++ later.

Delapouite:
Hello.

In this PR https://github.com/mawww/kakoune/pull/5485, I implemented the first of your actions point, because it's already useful in itself : 

> the :map command should accept a new "mode" ("combine_selections"? "combsel"?) that represents the "combine selections" menu

I went for the short `combine` as the various C++ functions are already called in the codebase.

We still need to tackle the remaining points about the propagation of `%val`s in later PRs.

mawww:
[...]

PR #5485, opened 2026-05-01
feat: allow mapping keys to `combine` mode (`<a-z>`, `<a-Z>`)

Delapouite:
Hello

As described in https://github.com/mawww/kakoune/issues/3572 by @Screwtapello , along the years lots of people have experimented with new ways to combine selections beyond the operations currently available like union or intersection.

For example adding ways to make substraction, complement or (symmetric) difference.

Recently, it's getting steam again with attemps of using this set combinations to build code-folding systems.
https://discuss.kakoune.com/t/a-toy-code-folding-plugin/2864

This commit is quite conservative it what it offers for now.
It only lets users map new keys to a proper `combine` mode, as a way to better group custom commands around this topic.

The introduced changes follow the same pattern than `KeymapMode::Goto`, `KeymapMode::View` or `KeymapMode::Object`.

The diff looks unfortunately a big long but it's mostly just cosmetic re-indenting to be consistent with the aforementioned `KeymapModes`
The "real" meat of this commit is basically just `KeymapMode::None` → `KeymapMode::Combine` for `on_next_key_with_autoinfo` and `build_autoinfo_for_mapping`

Thanks

### DIFF
--- a/src/commands.cc
+++ b/src/commands.cc
@@ -1591,14 +1591,15 @@ const CommandDesc echo_cmd = {
 
 KeymapMode parse_keymap_mode(StringView str, const KeymapManager::UserModeList& user_modes)
 {
-    if (prefix_match("normal", str)) return KeymapMode::Normal;
-    if (prefix_match("insert", str)) return KeymapMode::Insert;
-    if (prefix_match("menu", str))   return KeymapMode::Menu;
-    if (prefix_match("prompt", str)) return KeymapMode::Prompt;
-    if (prefix_match("goto", str))   return KeymapMode::Goto;
-    if (prefix_match("view", str))   return KeymapMode::View;
-    if (prefix_match("user", str))   return KeymapMode::User;
-    if (prefix_match("object", str)) return KeymapMode::Object;
+    if (prefix_match("normal", str))  return KeymapMode::Normal;
+    if (prefix_match("insert", str))  return KeymapMode::Insert;
+    if (prefix_match("menu", str))    return KeymapMode::Menu;
+    if (prefix_match("prompt", str))  return KeymapMode::Prompt;
+    if (prefix_match("goto", str))    return KeymapMode::Goto;
+    if (prefix_match("view", str))    return KeymapMode::View;
+    if (prefix_match("user", str))    return KeymapMode::User;
+    if (prefix_match("object", str))  return KeymapMode::Object;
+    if (prefix_match("combine", str)) return KeymapMode::Combine;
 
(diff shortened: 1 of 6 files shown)


src/commands.cc
```diff
@@ -1591,14 +1591,15 @@ const CommandDesc echo_cmd = {
...
 {
-    if (prefix_match("normal", str)) return KeymapMode::Normal;
-    if (prefix_match("insert", str)) return KeymapMode::Insert;
-    if (prefix_match("menu", str))   return KeymapMode::Menu;
-    if (prefix_match("prompt", str)) return KeymapMode::Prompt;
-    if (prefix_match("goto", str))   return KeymapMode::Goto;
-    if (prefix_match("view", str))   return KeymapMode::View;
-    if (prefix_match("user", str))   return KeymapMode::User;
-    if (prefix_match("object", str)) return KeymapMode::Object;
+    if (prefix_match("normal", str))  return KeymapMode::Normal;
```
Delapouite, line 1594: this hunk is mostly 1 char to the right realigning

src/normal.cc
```diff
@@ -2074,41 +2074,42 @@ void combine_selections(Context& context, SelectionList list, Func func, StringV
     if (&context.buffer() != &list.buffer())
         throw runtime_error{"cannot combine selections from different buffers"};
 
-    on_next_key_with_autoinfo(context, "combine-selections", KeymapMode::None,
+    on_next_key_with_autoinfo(context, "combine-selections", KeymapMode::Combine,
                              [func, list](Key key, Context& context) mutable {
```
Delapouite, line 2078: this reindenting to the left follow the same style in this `normal.cc` file, like the `goto_commands` for example.

src/normal.cc
```diff
@@ -2074,41 +2074,42 @@ void combine_selections(Context& context, SelectionList list, Func func, StringV
...
+            if (list.size() != sels.size())
+                throw runtime_error{format("the two selection lists don't have the same number of elements ({} vs {})",
+                                           list.size(), sels.size())};
+            for (int i = 0; i < list.size(); ++i)
+                combine_selection(sels.buffer(), list[i], sels[i], op);
+            list.set_main_index(sels.main_index());
+        }
+        func(context, std::move(list));
+    }, title.str(),
+    build_autoinfo_for_mapping(context, KeymapMode::Combine,
```
Delapouite, line 2105: here, the code now uses `build_autoinfo_for_mapping` instead of raw strings

Issue #5484, opened 2026-04-28
[REQUEST] `%env` expansion for environment variables


ficd0:
### Feature

A new kind of expansion, `env`, which expands to the content of an environment variable in the current session. It would be of form `%env{NAME}`, where `NAME` is the environment variable. For example, `%env{EDITOR}` or `%env|XDG_CONFIG_HOME|`, etc.

### Usecase

There are times we may want to access the content of an environment variable for a simple command without spawning an entire subshell. For example I have this command in my config:

```kak
define-command todo %{
	edit %sh{
		echo "$HOME/.todo/todo.txt"
	}
}
```

It would be much nicer if we could do it like this:

```kak
define-command todo %{
	edit "%env{HOME}/.todo/todo.txt"
}
```

Delapouite:
A few other issues related to envars : 
- https://github.com/mawww/kakoune/issues/2330
- https://github.com/mawww/kakoune/pull/2348
- https://github.com/mawww/kakoune/issues/4482
- https://github.com/mawww/kakoune/pull/4431

PR #5486, opened 2026-05-02
Implement `%env` expansion

netfri25:
as requested in #5484


krobelus:
thanks but I'm not sure if there's a good use case for this, since we already have `%sh{...}`

ficd0:
> thanks but I'm not sure if there's a good use case for this, since we already have `%sh{...}`

This could be used to avoid spawning a subshell just to read an env variable. Although as mawww said in https://github.com/mawww/kakoune/issues/2330#issuecomment-720270352, that may not be enough of a motivating use case.

PR #5493, opened 2026-05-17
Add -back switch to arrange-buffers command

Blezzing:
Hi,

This is my first pull request on kakoune and github in general, so please correct me if i did something wrong.

I like using buffer-next and buffer-previous to move through buffers where the sequence in the buffer list is defined by when they where last accessed. This implies rearranging the buffer list whenever a buffer is accessed to place specific buffers at the back, rather than the front, of the buffer list, which is not something the arrange-buffer command support without modification.

This pull request adds a switch to the arrange-buffers command to control which end of the list specified buffers are placed at, preserving the current behavior as default.

To visualize this, the below configuration a workflow where the buffer-next and buffer-previous move through the history of accessed buffers, with "previous" mapping to "previously in time", consistent between opening of new buffers, and changing to an already opened buffer.
```
hook global WinDisplay .* %{ arrange-buffers -back %val{hook_param} }
map global goto p "<esc>\:buffer-previous<ret>" -docstring "next buffer"
map global goto n "<esc>\:buffer-next<ret>" -docstring "previous buffer"
```

### DIFF
--- a/src/buffer_manager.cc
+++ b/src/buffer_manager.cc
@@ -120,10 +120,10 @@ void BufferManager::clear_buffer_trash()
     m_buffer_trash.clear();
 }
 
-void BufferManager::arrange_buffers(ConstArrayView<String> first_ones)
+void BufferManager::arrange_buffers(ConstArrayView<String> buffers, bool to_back)
 {
     Vector<size_t> indices;
-    for (const auto& name : first_ones)
+    for (const auto& name : buffers)
     {
         auto filename = real_path(parse_filename(name));
         auto it = find_if(m_buffers, [&](auto& buf) { return buf->display_name() == name or (buf->flags() & Buffer::Flags::File and buf->filename() == filename); });
@@ -143,6 +143,9 @@ void BufferManager::arrange_buffers(ConstArrayView<String> first_ones)
         if (buf)
             res.push_back(std::move(buf));
     }
+    if (to_back)
+        std::rotate(res.begin(), res.begin() + indices.size(), res.end());
+
     m_buffers = std::move(res);
 }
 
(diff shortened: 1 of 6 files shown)


src/buffer_manager.cc
```diff
@@ -120,7 +120,7 @@ void BufferManager::clear_buffer_trash()
     m_buffer_trash.clear();
 }
 
-void BufferManager::arrange_buffers(ConstArrayView<String> first_ones)
```
mawww, line 123: I'd rather have this change look like that:
```diff
-void BufferManager::arrange_buffers(ConstArrayView<String> first_ones)
+void BufferManager::arrange_buffers(ConstArrayView<String> buffers, bool to_back)
 {
     Vector<size_t> indices;
-    for (const auto& name : first_ones)
+    for (const auto& name : buffers)
     {
         auto filename = real_path(parse_filename(name));
         auto it = find_if(m_buffers, [&](auto& buf) { return buf->display_name() == name or (buf->flags() & Buffer::Flags::File and buf->filename() == filename); });
@@ -143,6 +143,8 @@ void BufferManager::arrange_buffers(ConstArrayView<String> first_ones)
         if (buf)
             res.push_back(std::move(buf));
     }
+    if (to_back)
+        std::rotate(res.begin(), res.begin() + indices.size(), res.end());
     m_buffers = std::move(res);
 }
 ```

mawww (commented):
I think the feature sounds reasonable, can you try to suggested alternative implementation and add some notes in both src/main.cc and doc/pages/changelog.asciidoc not document the change to users ?

Blezzing:
I have made the following amends:
- added the missing documentation and changelog entries in similar style as existing entries.
- renamed the parameters to BufferManager::arrange_buffers according to your example.
- updated logic to be the previous behavior with the optional rotate as suggested (neat idea).

Please let me know if i missed anything.

PR #5496, opened 2026-05-21
tests: disable autorestore test if pathchk is not there (no coreutils)

stacyharper:
Alternatively, we could run pathchk only if present. As this is an extra minor check over the backup file path.

### DIFF
--- a/test/tools/autorestore/autorestore/enabled
+++ b/test/tools/autorestore/autorestore/enabled
@@ -1,0 +1,2 @@
+#!/bin/sh
+command -v pathchk >/dev/null


mawww:
Which OS misses pathchk ? I wrote that check this way because pathchk is part of posix and I assumed it would be available everywhere.

stacyharper:
In Alpine Linux we need the extra dependency coreutils (GNU) to have pathchk. It is curiously not part of busybox

mawww:
Thanks for the feedback, used an alternate way of doing this check that does not rely on `pathchk`

Delapouite:
For future reference: https://github.com/vda-linux/busybox_mirror/blob/master/docs/posix_conformance.txt

> POSIX Tools not supported:
  asa, at, batch, bc, c99, command, compress, csplit, ex, fc, file,
  gencat, getconf, iconv, join, link, locale, localedef, lp, m4,
  mailx, newgrp, nl, pathchk, pax, pr, qalter, qdel, qhold, qmove,
  qmsg, qrerun, qrls, qselect, qsig, qstat, qsub, tabs, talk, tput,
  unlink, uucp, uustat, uux

PR #5499, opened 2026-05-25
Implement paste method for JSON UI

falbru:
Implements the `paste(String content)` method for JSON UI. The callback was already implemented, but there was no way to call it through the JSON interface.


mawww:
Thanks !

PR #5502, opened 2026-05-25
feat(rc): add AppleScript syntax highlighting

paaloeye:
- add AppleScript filetype detection and syntax highlighting
- highlight keywords, operators, strings, comments, and values
- add basic indent rules for tell, on, if/then, repeat, and try blocks
- add comment string support via comment.kak
- add highlight and indent regression tests


mawww:
I wonder about ll those tests included in this, I am worried adding highlighting tests like this is mainly going to add extra maintenance churn without add much value, something like `highlight/applescript/keywords` seems quite trivial for example. I think highlighting tests should focus on things that are tricky to get right.

paaloeye:
Fair, I'll re-do the tests or delete them altogether.

PR #5503, opened 2026-05-25
feat(rc): add Nushell syntax highlighting

paaloeye:
Based on:
- https://raw.githubusercontent.com/shikijs/textmate-grammars-themes/refs/heads/main/packages/tm-grammars/grammars/nushell.json
- https://github.com/nushell/vscode-nushell-lang/blob/main/syntaxes/nushell.tmLanguage.json

### DIFF
--- a/rc/filetype/nushell.kak
+++ b/rc/filetype/nushell.kak
@@ -1,0 +1,168 @@
+# https://www.nushell.sh
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+# Based on:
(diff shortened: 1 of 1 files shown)


mawww:
What do you mean by "based on" ? I assume that means reading those files as reference to write this one, which I would assume would not make this derivative work.

paaloeye:
> What do you mean by "based on" ? I assume that means reading those files as reference to write this one, which I would assume would not make this derivative work.

Sorry for being MIA. Correct. Not sure about being derivative work, but I've seen a bunch of other rc files referencing TM grammars. 

I hope it helps!

Issue #5505, opened 2026-05-29
[BUG][CRASH] segfault (null deref) - #5483 / 930034c4 missed edge case


zetier-dm:
### Version of Kakoune

v2026.05.21 (36ad52e)

### Reproducer

```bash
#!/bin/sh

KAK="${KAK:-kak}"
COLS="${COLS:-80}"

$KAK -version

ulimit -c 0 2>/dev/null
work=$(mktemp -d)
trap 'rm -rf "$work"' EXIT

# A buffer whose line 2 is longer than the wrap column.
cat > "$work/file" <<'EOF'
short first line
    let x = compute_a_really_long_value_that_will_exceed_the_wrap_column_width_for_sure(1, 2, 3);
EOF

# rc: wrap + a whole-line, MULTI-LINE replace-ranges atom on the long line 2.
cat > "$work/rc.kak" <<'EOF'
add-highlighter global/ wrap
declare-option range-specs my_ranges
set-option buffer my_ranges %val{timestamp}
add-highlighter buffer/anno replace-ranges my_ranges
set-option -add buffer my_ranges "2.1,2.97|{red}NOTE multiline annotation note that is itself quite long and exceeds the wrap column
    let x = compute_a_really_long_value_that_will_exceed_the_wrap_column_width_for_sure(1, 2, 3);"
execute-keys gg
EOF

printf '{ "jsonrpc": "2.0", "method": "resize", "params": [ 40, %s ] }\n' "$COLS" \
  | "$KAK" -n -ui json -e "source $work/rc.kak" "$work/file" >/dev/null 2>&1 && true

exit 0
```

### Outcome

```
$./repro.sh
Kakoune v2026.05.21
Segmentation fault (core dumped)
```

```
(gdb) bt
#0  0x00005555555ad8a0 in Kakoune::DisplayAtom::type (this=0x0) at src/display_buffer.hh:77
#1  0x00005555556e395e in Kakoune::WrapHighlighter::next_split_pos (this=0x555555a8f500, pos=..., line_end=non-dereferenceable iterator for std::vector, wrap_column=79, prefix_len=0) at src/highlighters.cc:702
#2  0x00005555556e3040 in Kakoune::WrapHighlighter::do_highlight (this=0x555555a8f500, context=..., display_buffer=...) at src/highlighters.cc:643
[...]

PR #5506, opened 2026-05-29
Fix 5505 segfault

zetier-dm:
Fixes NULL-deref in WrapHighlighter::next_split_pos as described in #5505


mawww:
Hello, I did merge your commit as it does provide the reproducing test, however the fix was not working (the test was still failing) so I followed up with a proper fix (1e70389).

PR #5507, opened 2026-06-02
Document behaviour of `\z` zero-width-assertion.

Screwtapello:
I wanted to select up to the end of the buffer, so I did `?\z<ret>` and was very surprised to get "no matches found", since the buffer very definitely has an end. I thought this was an inconsistency, since I could search for other zero-width assertions like `^` or `$`, but it turns out to actually be consistent: searching for a ZWA on its own will select the character after the point where the ZWA matches, and (by definition) there is no character after the end of the buffer.

I have updated the documentation with the information I wanted to find.

If this information is incorrect or misleading, I'm willing to fix it.

### DIFF
--- a/doc/pages/regex.asciidoc
+++ b/doc/pages/regex.asciidoc
@@ -122,7 +122,8 @@ as possible.
 == Zero width assertions
 
 Assertions do not consume any character, but they will prevent the regex
-from matching if not fulfilled.
+from matching if not fulfilled. If you search for one of these on its own,
+Kakoune will select the character *after* the point where it matches.
 
 * `^` matches at the start of a line; that is, just after a newline
       character, or at the subject's beginning (unless it is specified
@@ -137,7 +138,8 @@ from matching if not fulfilled.
        character and the current character are word characters, or both
        are not.
 * `\A` matches at the subject string's beginning.
-* `\z` matches at the subject string's end.
+* `\z` matches at the subject string's end. Use `.\z` to match the last
+       character in the subject string.
 * `\K` matches anything, and resets the start position of capture group
        0 to the current position.
 
--- a/doc/pages/regex.asciidoc
+++ b/doc/pages/regex.asciidoc
@@ -122,33 +122,37 @@ as possible.
 == Zero width assertions
 
 Assertions do not consume any character, but they will prevent the regex
-from matching if not fulfilled. If you search for one of these on its own,
-Kakoune will select the character *after* the point where it matches.
+from matching if their specific property is not true at the current
+position.
 
-* `^` matches at the start of a line; that is, just after a newline
-      character, or at the subject's beginning (unless it is specified
-      that the subject's beginning is not a start of line).
-* `$` matches at the end of a line; that is, just before a newline, or
-      at the subject end (unless it is specified that the subject's end
-      is not an end of line).
-* `\b` matches at a word boundary; which is to say that between the
-       previous character and the current character, one is a word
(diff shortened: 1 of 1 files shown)


doc/pages/regex.asciidoc
```diff
@@ -122,7 +122,8 @@ as possible.
 == Zero width assertions
 
 Assertions do not consume any character, but they will prevent the regex
-from matching if not fulfilled.
+from matching if not fulfilled. If you search for one of these on its own,
+Kakoune will select the character *after* the point where it matches.
```
mawww: I am not too keen on this wording as it suggests the point regex matches on are in-between characters, which then makes basic atom (such as a plain letter) work by checking 'the character *after* the point' as well. While this mental model can work, I think in the context of Kakoune, which avoids 'in-between-character' in its editing model, we would be better off keeping the onto-a-character mental model.

I think the main point of confusion here is what the "end" of the subject string is. Similarly to say C++ iterators, the end of the subject string is one-past-the-last, which explains why `\z` would not select anything (because Kakoune does not permit selecting one-past-the-last).

What do you think ?

Screwtapello: (sorry I left this for a while, I had some real-life responsibilities to take care of)

The name of the section is "zero-width assertions", and a regex like `^.` matches one character, not two, so it makes sense to me that zero-width assertions like `^` match between characters rather than on them. My mental model of regular expressions comes from using them in other contexts (languages like JavaScript and Python, and tools like `grep`) as well as Kakoune, so I'm already thinking about them as "in-between-character" spans, and I imagine that Kakoune takes the matching spans and "rounds" them to the nearest whole character to suit its editing model.

Specifically for Kakoune's own documentation, it does make sense to describe the behaviour as it exists in Kakoune. I'll think about how to reword things in that way.

mawww: I am not sure that mental model is correct even for those contexts, if you look at the code of the regex vm that handles the `\z` assertion, you will see (at regex_vm.hh:458):
```cpp
                case CompiledRegex::SubjectAssertion:
                    if (pos != (param.subject_begin ? config.subject_begin : config.subject_end))
                        return failed();
                    break;
```
That is, for `\z` assertions checks if `pos == config.subject_end`, there is no concept of the character after the current point, the regex engine processes every codepoints in the subject strings, then runs an extra iteration at one-past-the-last codepoint (where the current codepoint value is -1).

A quick look at pcre2_dfa_match.c:811 suggests at least pcre2 relies on the same logic
```c
      case OP_EOD:
      if (ptr >= end_subject)
        {
        if ((mb->moptions & PCRE2_PARTIAL_HARD) != 0)
          could_continue = TRUE;
        else { ADD_ACTIVE(state_offset + 1, 0); }
        }
      break;
```
So `\z` matches if our location is >= end_subject, which suggests the same model: current position is one-past-the-last character.

My thinking may be quite tainted by my knowledge of the implementation details, so I wonder if most users do think of regex matching using the in-between-character mental model. That seems like an unnecessary complication to me, but if it was the case I'd be more inclined to merge this, any opinions from other users on this ?

Screwtapello: I've pushed a new commit that rewords the "zero-width assertions" section to more explicitly describe each assertion in terms the current position being "at" a character, rather than "after" or "before" or "between". What do you think?

mawww: That new wording looks great, thanks.

Issue #3560, opened 2020-06-21
<a-z>i within original selection should produce result


vbauerster:
I wish `<a-z>i` would allow me to intersect two selection lists with not equal number of elements (selections), if one selection contains others. It's hard to explain with words so here is illustration:

[![alt-z.png](https://i.postimg.cc/W3dxbvTD/alt-z.png)](https://postimg.cc/0MqVch6x)

So selection is saved first with `Z`, then we have two more selections with help of `<a-S>` (both are within original), next we want to intersect in order to get result as depictured above on third line.

At the moment `<a-z>i` complains like `the two selection lists don't have the same number of elements (1 vs 2)`.

Also imo following (same number of selections, but the second one is within the first one) should also work:

[![one-within.png](https://i.postimg.cc/hvdvkvjp/one-within.png)](https://postimg.cc/bSqqbz1t)

mreppen:
This is the same feature I requested here: https://github.com/mawww/kakoune/issues/3504
I am currently running with my own patch and love this feature. My patch is only a proof of concept. Note that the order, i.e., which selection is removed from which, is reversed.

PS.
I don't think this should be called intersection. Seeing selections as sets, it is a difference/subtraction.

vbauerster:
You're right this is not intersection.

If we assume that selected (by anchor or cursor) cell is 1 and not selected is 0, then current intersection behaviour is AND table, survives 1 and 1:
[![intersection.png](https://i.postimg.cc/ZYVpNX7c/intersection.png)](https://postimg.cc/K431Spyk)

Difference/subtraction is like negated AND table, we negate the result of AND operation:
[![diff.png](https://i.postimg.cc/MKf08d5m/diff.png)](https://postimg.cc/8jDfRdq7)


The tricky part is that current implementation doesn't allow to combine single selection with multiple or vise versa. But I reckon for difference/subtraction operation it should.

mreppen:
@vbauerster I might be misunderstanding you, but no, it is not "negated AND". What you describe in your second example is the symmetric difference https://en.wikipedia.org/wiki/Symmetric_difference
That's actually why I called it subtraction: to avoid confusion with symmetric difference.

From your example, I'm quite convinced it's subtraction. Please have a look at the linked issue and I think you will agree.

In that issue, we also discuss paired vs non-paired operations, which likely relates to what you call single/multiple. Please check it out and join the discussion.

[...]

Issue #2850, opened 2019-04-11
Add a difference 'd' operator to marks


alexherbo2:
https://github.com/alexherbo2/split-object.kak/issues/2#issuecomment-482069331

Delapouite:
![image](https://user-images.githubusercontent.com/39315/65507095-6e32dd80-decd-11e9-8be1-66bc8e547d16.png)

As shown on this illustration we may need 2 "difference operators" to decide which parts we want to keep. (Selections from the register or the ones in the buffer)

And even a third operator to express the symmetric difference. 
![image](https://user-images.githubusercontent.com/39315/65507238-c79b0c80-decd-11e9-9af0-68eac381499a.png) 
https://en.wikipedia.org/wiki/Symmetric_difference

PR #5512, opened 2026-06-06
feat: add difference operators to combine mode (`a-z`, `a-Z`)

Delapouite:
Hello

## TL;DR:

Along the years, the "selections difference" operator has been requested multiple times on GitHub, on the forum, in IRC, in the vis editor… showing the strong interest of people for this feature. 

Non exhaustive lists:

- https://github.com/mawww/kakoune/issues/2850
- https://github.com/mawww/kakoune/issues/3560
- https://discuss.kakoune.com/t/selection-combining-operator-u-union-is-little-bit-misleading/724
- https://discuss.kakoune.com/t/selection-sets-do-mathematical-set-operations-on-selections/1728
- …

Lots of well-known Kakoune users have made attempts to write such feature in user-land during this period.
Various languages were used: @danr did it in Python, @Screwtapello in pure shell, @gustavo-hms in Lua and @occivink in Perl, again demonstrating that people enjoys these set operations.

Some users also demonstrated that set-operations are useful building blocks to implement more powerful [code folding features](https://discuss.kakoune.com/t/a-toy-code-folding-plugin/2864).

So here's my attempt to finally get in in the core itself!

I'd like to thank warmly all the people mentioned above. This contribution is definitely made on the "shoulders of these giants", especially the [implementation made by Occivink](https://github.com/occivink/kakoune-set-operations/blob/master/set-operations.kak) which served as the base for the translation C++ version in this commit. The challenge was indeed to port the 500 LOC perl version into a 250 LOC version C++ fitting into `normal.cc`.

---

Below are more explanations about design choices.

## Current situation

The current combine mode is a mixed bag of operations.
The majority of them are "pairwise", `a` (for append) being the only exception.
It's not always intuitive to understand what each operation does, especially the `u` and `i` which have surprising result when selections in a pair does not overlap. Also the "difference" operation(s) are missing.

## Goals

The first idea is to disambiguate the two groups of operations (regular vs pairwise) without introducing yet another key in normal mode. Here I suggest that by default, the combinations are made on full selections list. To get back to the "old" pairwise behavior, the `<a-z>` and `<a-Z>` needs to be prefixed with a count of `2`: `2<a-z>` and `2<a-Z>`.

The second thing to understand when we ask for a "difference operator" (also called "complement" in the litterature), is that there's not 1 but 3 "difference operator". So with union and intersection we have a grand total of 5 operations as depicted below:

image
image

In the images above, the sets are called `A` and `B`, but for our understanding it's better to calls them `B` for selections currently in buffer and `R` for selections currently in the register.

When we say "difference" we sometimes want to keep the part of the selection list that is only in the buffer (`B \ R`), sometimes we wand to keep the part only in the register (`R \ B`) and sometimes both (`B Δ R`).

The pairwise operators are kept because they are already used in multiple scripts in the `rc` directory.

That's why we obtain the following mappings:

| | Before | After |
| - | ------- | ------ |
| union | `<a-z>a` | `<a-z>u` |
| intersection | | `<a-z>i` |
| buffer difference | | `<a-z>b` |
| register difference | | `<a-z>r` |
| symmetric difference | | `<a-z>s` |
| pairwise union (with gap) | `<a-z>u` | `2<a-z>u` |
| pairwise intersection (with gap) | `<a-z>i` | `2<a-z>i` |
| pairwise left most cursor | `<a-z><` | `2<a-z><` |
| pairwise right most cursor | `<a-z>>` | `2<a-z>>`  |
| pairwise shortest | `<a-z>-` | `2<a-z>-` |
| pairwise longest | `<a-z>+` | `2<a-z>+` |

In the code, I extracted the current "append" logic into a proper `selections_union` function. This is useful to implement the symmetric difference because it is the union of the buffer and register differences : `A Δ B = (A \ B) ∪ (B \ A)`

The C++ code is probably not the most elegant nor the most optimized performance wise but it can be refined, so @mawww don't hesitate to clean it / readjust it afterwards your Cpp-fu is obviously way better than mine.

For the testing scenario, I also reused the work done by occivink : https://github.com/occivink/kakoune-set-operations/blob/master/test.kak_ because the way he wrote tests are legible in one file instead of splitting in hard multiple files which is harder to follow.

## Future considerations

We may go deeper in optional future pull-requests. For instance, we could add the `<`, `>`, `-` and `+` operators in the non-pairwise mode. Also we may also want to introduce pairwise differences `2<a-z>b`, `2<a-z>r` and `2<a-z>s`. But I prefer keeping this PR not too long for now.

Thanks!  Don't hesitate to provide your feedback.

### DIFF
--- a/contrib/TRAMPOLINE
+++ b/contrib/TRAMPOLINE
@@ -42,7 +42,7 @@ using the built-in `:doc` command.
        .---,        described in this document are "editing primitives" that
        |esc|        have to be hit in normal mode, which is the default mode
        `---'        when you start the editor. To enter insert mode, hit the
-                    `i` key, and to leave it and return to normal mode, hit the 
+                    `i` key, and to leave it and return to normal mode, hit the
                     escape key.
 
 
@@ -322,7 +322,7 @@ using the built-in `:doc` command.
    | " |_. .---,    temporary search, one could make this primitive save the
    `---'  `| _ |    pattern to a different register, to preserve the default one,
            `---'    e.g., `"m*` to save the pattern to the `m` register, or even
-                    `"_*` to save the pattern to a "null" register, which does 
+                    `"_*` to save the pattern to a "null" register, which does
                     not store anything written to it.
 
                     ==[ CAPTURE GROUPS
@@ -373,11 +373,11 @@ using the built-in `:doc` command.
         .---,       expressions. To allow that, the save mark (`Z`) and append
         | Z |       mark (`<a-z>`) come in handy, as they respectively save
         `---'       the current selection to the mark register (`^`), and
-                    show a menu that allows appending the current selection
-.---, .---,         to the mark register upon hitting the `a` key. That way,
+                    show a menu that allows combining the current selection
+.---, .---,         to the mark register upon hitting the `u` key. That way,
 |alt|+| z |_.       it becomes possible to chain and save (append) several
 `---' `---'  `.---, selections made using completely different methods
-              | a | (select, split etc) without being forced to preserve
+              | u | (select, split etc) without being forced to preserve
               `---' them at all times.
         .---,
         | z |       Restoring a mark saved to the mark register using those
--- a/doc/pages/changelog.asciidoc
+++ b/doc/pages/changelog.asciidoc
@@ -5,6 +5,9 @@ To control the startup info message, see <<options#startup-info,`:doc options st
 This changelog contains major and/or breaking changes to Kakoune between
 released versions.
 
+* Combine mode, `<a-z>`, `<a-Z>`, now has distinct regular and pairwise operations.
+  Regular combinations now include: difference in both directions and symmetric difference.
+
 == Kakoune 2026.05.21
 
 * Support the `\N` escape sequence in regex (like in PCRE, matches `[^\n]`).
@@ -131,7 +134,7 @@ released versions.
   (See <<expansions#command-and-response-fifo,`:doc expansions command-and-response-fifo`>>)
 
 * Shell expansions only trim the last trailing newline instead of all of
-  them to make is possible to losslessly pass text through `%sh{...}`. 
+  them to make is possible to losslessly pass text through `%sh{...}`.
 
 * `set-option -remove` support for subtracting/removing from option values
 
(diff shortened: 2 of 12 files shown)


occivink:
eheh nice stuff, I had a good time working on the plugin: lots of tricky edge cases involved, feels a bit like puzzle-solving. The functionality is pretty niche, but it's pretty handy to have when needed. Now I feel like wrapping it up a bit, so thanks for the burst of motivation. That said, a core implementation would definitely be much more robust.

I remember finding this article later on: https://zayenz.se/blog/post/how-to-check-for-overlapping-intervals
which I found fairly relevant, and would probably allow simplifying some of my implementation. Maybe some of it would apply for yours as well

gustavo-hms:
Hello, Delapouite!

Thanks for this great work, and sorry for taking so long to comment.

For the sake of completeness and documentation, let me write here the arguments I gave in the forum. This is my small contribution for the topic. I hope it can complement the rationale given by you in the description of this PR and help guiding further developments on marks.

I see a parallel between marks and the following keys:

- `?`
- `<a-?>`
- `N`
- `<a-N>`
- `s`
- `S`
- `<a-k>`
- `<a-K>`

The idea is that, whereas marks are used to combine current buffer selections with *previous* buffer selections, the aforementioned keys are used to combine current buffer selections with *potencial selections described by regular expressions but not yet materialised in the buffer*. In other words: marks work by combining buffer selections with *past selections*, and those keys work by combining buffer selections with *future selections*. This way, there is an equivalence between many of those keys and operations on marks:

- `N` and `<a-N>` map to the `append list` operation on marks;
- `?` and `<a-?>` map to `union`;
- `s` maps to `intersection`;

 Once we realise this parallel, what follows is that:

1. Since those keys come built-in, the operations they provide are considered important.
2. Whereas there are equivalents for `N`, `<a-N>`, `?`, `<a-?>`, and `s`, marks operations lack equivalents for the other keys (see bellow);
3. If those operations are considered important so that there are keys for them, they should also be present as marks operations.

The missing operations then are:

- Difference, the equivalent of the `S` key. That's what is covered by this PR;
- Keep only those marks that fully contains the selections (or vice-versa). It's the equivalent of `<a-k>`;
- Keep only those marks that *does not* fully contain the selections (or vice-versa). It's the equivalent of `<a-K>`.

I can attest that those three operations are indeed useful, and I've implement all of them in Lua for different use cases.

And again: the fact that they are implemented as keys is a testament of their importance.

mawww:
Assigning myself so that I dont forget, its high time I review this properly.

PR #5513, opened 2026-06-12
feat: Make git.kak next-hunk and prev-hunk select all hunk lines

pri-yan-shu:
This plays nicely when manually staging a file, goto a hunk and stage it all with git apply --cached.

Not too sure about using str vs str-list for git_hunk_list.

### DIFF
--- a/rc/tools/git.kak
+++ b/rc/tools/git.kak
@@ -78,7 +78,7 @@ declare-option -hidden line-specs git_blame_index
 declare-option -hidden str git_blame
 declare-option -hidden str git_blob
 declare-option -hidden line-specs git_diff_flags
-declare-option -hidden int-list git_hunk_list
+declare-option -hidden str-list git_hunk_list
 
 define-command -params 1.. \
     -docstring %{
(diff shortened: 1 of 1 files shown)


mawww:
Hello, I think this should select the full last line of the hunk, instead of only the first character. Not sure if the line length is easily available to correctly fill the git_hunk_list option, but a workaround would be to `exec x` after the `select ...` command.

pri-yan-shu:
Is this ok?

krobelus:
FWIW `:git apply --cached` operates on the entire hunk if the selection doesn't contain a newline

mawww:
> FWIW `:git apply --cached` operates on the entire hunk if the selection doesn't contain a newline

AFAIU `git next-hunk` only works on `git show-diff` gutter, while `git apply` works on git diff buffers.

krobelus:
ah right, I was confused because I never used those. I've been meaning to suggest renaming the gutter commands..
`git apply` actually works on git-gutter lines as well (even without `git show-diff`) but without the whole-hunk selection heuristic I mentioned

PR #5515, opened 2026-06-13
feat(rc): add Raku syntax highlighting

exlee:
Raku syntax highlighting was missing and I made multiple rounds of GPT-5.5 generation.

I tested it and it works well with non-trivial syntax without braking. I cannot voach for full completeness though.

Since it's not me-made I don't claim any authorship over it (though just in case I made a PR for waiver too).

### DIFF
--- a/rc/filetype/raku.kak
+++ b/rc/filetype/raku.kak
@@ -1,0 +1,85 @@
+# https://raku.org/
+
+# Detection
+
+hook global BufCreate .*\.(raku|rakumod|rakutest|rakudoc|p6|pm6|t6|pod6)$ %{
+    set-option buffer filetype raku
+}
+
+# Initialization
+
+hook global WinSetOption filetype=raku %{
+    require-module raku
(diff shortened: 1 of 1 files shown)


mawww:
Hello, unfortunately, if you cannot claim you made this, the copyright is unclear and I don't think we can put it in the public domain with your copyright waiver. Moreover, the expectations for PRs is that the person submitting it understands the change thoroughly and takes responsibility for it, regardless of which tools were used to generate it.

exlee:
As far as I know based on current legal situation public domain is the default for generated content, thus I cannot claim copyright for it. 

I'm leaving this to your judgment. 

Regarding understanding - I verified that it works and reviewed it thoroughly. In the end it is a syntax highlighting with regexp so there's not a lot to understand.

I decided to make a PR because I'd expect those looking for Raku highlighters to look here for issues/discussions. Even if PR is rejected I believe it will be of use to Raku enthusiasts as they might find it through search box.

PR #5518, opened 2026-06-20
javascript.kak: Add function highlighter

kmafeni04:


mawww:
I dont think we want to hightlight every identifier followed by a parenthesis, at the moment we dont really do that in most highlighters as it makes the highlighting very colorful, but I guess you could argue that its just a matter of making the function face match back to default.

kmafeni04:
> I dont think we want to hightlight every identifier followed by a parenthesis, at the moment we dont really do that in most highlighters as it makes the highlighting very colorful, but I guess you could argue that its just a matter of making the function face match back to default.

I did that because of what I saw in the lua highlighter: https://github.com/mawww/kakoune/blob/336aed9a5cb13fc3d15af2dc30cdda3f3bccee94/rc/filetype/lua.kak#L51

kmafeni04:
Decided to close it and clear out my repo until further clarifiication, I can always make another PR

PR #5520, opened 2026-07-03
format.kak: Make sure Kakoune provides the envvars formatcmd needs

Screwtapello:
When Kakoune executes a shell command, it scans the text for tokens like `kak_*` and if it finds any it recognises it adds them to the environment for the command (as described in `:doc expansions shell-expansions`). For example, if you select some text and type:

    |fold -w $kak_opt_autowrap_column<ret>

...then Kakoune sees `kak_opt_autowrap_column`, and exports the value of the `autowrap_column` option into the environment of the shell, which expands the variable reference then runs the command.

The `:format` command amounts to something like:

    |eval "$kak_opt_formatcmd"

...which means that Kakoune never sees "kak_opt_autowrap_column", so it never provides the value, so the shell winds up with a malformed command like "fold -w" which just causes an error.

To avoid this problem, as well as passing "$kak_opt_formatcmd" into the shell command, we also *append* it to the shell command where Kakoune can see it, guarded by an `exit` statement to prevent it from being processed.

### DIFF
--- a/rc/tools/format.kak
+++ b/rc/tools/format.kak
@@ -16,7 +16,7 @@ define-command format-selections -docstring "Format the selections individually"
     }
     evaluate-commands -draft -no-hooks -save-regs 'e|' %{
         set-register e nop
-        set-register '|' %{
+        set-register '|' %exp{
             format_in="$(mktemp "${TMPDIR:-/tmp}"/kak-formatter.XXXXXX)"
             format_out="$(mktemp "${TMPDIR:-/tmp}"/kak-formatter.XXXXXX)"
 
@@ -29,6 +29,15 @@ define-command format-selections -docstring "Format the selections individually"
                 cat "$format_in"
             fi
             rm -f "$format_in" "$format_out"
+
+            # We want the format command to be literally present in this
+            # command, so that Kakoune will supply any any $kak_* variables
+            # it mentions, but we don't want to evaluate the command here
+            # so we'll tell the shell to exit before it gets to it.
+            # Just commenting it out would not be sufficient, since it might
+            # include newlines.
+            exit
+            %opt{formatcmd}
         }
         execute-keys '|<ret>'
         %reg{e}


rc/tools/format.kak
mawww, line 24: Sorry if there an obvious reason, but cannot we expand formatcmd directly here ?

Screwtapello: The shell has special rules for expanding inside double quotes, so things work as expected even if the thing you're expanding includes double quotes:

```console
$ q='"'
$ printf '%s\n' "a" "b" # actual double quotes delimit args
a
b
$ printf '%s\n' "a${q} ${q}b" # quotes resulting from expansion don't
a" "b
```

If Kakoune did the expansion instead of the shell, it wouldn't get the protection of the shell's "expand inside double quotes" logic. In particular, if `formatcmd` included a double quote and then a newline, the `eval` would be executed up until the newline, then whatever follows the newline would be executed as a separate command with the redirections.

As an alternative, it might be possible to re-implement the format plugin on top of Kakoune's `|` command, which would solve this problem, but would make it impossible to collect errors reported by the formatter as the current version of the plugin does. Perhaps there's a better way to get the best of both worlds, but I haven't thought of it yet.

mawww: How about if we made `formatcmd` a `str-list` and use `eval %val{quoted_opt_formatcmd}` ?

Screwtapello: If `%val{quoted_opt_formatcmd}` did what it looks like it should do, then that might work. However:

```
Error: 1:1: 'format-selections': 7:5: 'evaluate-commands': no such variable: quoted_opt_formatcmd
```

The other thing I'm worried about is that `formatcmd` is not always a simple command-line. For example, in the Kakoune wiki's "Format" page, the example configuration for [HTML](https://github.com/mawww/kakoune/wiki/Format#html-tidy) defines a shell function, and [Ruby](https://github.com/mawww/kakoune/wiki/Format#rubocop) uses a pipeline through `sed`. I'm not sure if that would still work, but it would be easy enough to test.

Alternatively, we could just take the compatibility hit and say that anything more than a simple command needs to be wrapped in a shell-script.

Issue #752, opened 2016-07-28
w, b and e selection inconsistencies


kurkale6ka:
the cursor will be represented by `[]`

`one tw[o] three`:
- _problem_: w results in only selecting the following white space
- _expected_: select o + space (same behaviour as for `one t[w]o three`)

`one[ ] two three`:
- _problem_: w results in only selecting white spaces
- _expected_: select two + following spaces (same behaviour as for `one [ ]two three`)

`one two[ ] three`:
- _problem_: b results in only selecting two
- _expected_: select space + two (same behaviour as for `one two [ ]three`)

`one [ ]two three`:
- _problem_: e results in only selecting two
- _expected_: select space + two (same behaviour as for `one[ ] two three`)

mawww:
Hello, all that is actually the expected behaviour.

For the first case, we have the cursor on the last char of the word, so `w` selects the next word (use `W` if you really want to extend to it).

For the second case, we consider the consecutive spaces as a word, in `w`, `b`, `e` sense.

The other cases are are behaving like that for the same reasons.

kurkale6ka:
Thanks for explaining.
It's still confusing that white spaces would be considered as words.
It might be worth explaining in the README what a word is in the context of w, b, e.

PR #5521, opened 2026-07-06
Fix w

### LINKED ISSUE #3132: <a-w> does not work as expected for one letter long Words at the beginning of a line.
### Steps 
Put the following text in a buffer:
```
c ccc ccc ccc
cc ccc ccc ccc
```
Move to the second line and do <a-h> to get your grey cursor to the beginning. If you do one <a-w>, the text "cc " at the beginning of the line will be selected. That's what is supposed to happen.
Now, try the same thing on the first line.

### Outcome 
The first "c" is not included in the resulting selection.

### Expected 
The first "c" should be in the selection.

lrworth:
`w` is intentionally inconsistent, in that it selects from the cursor up to the start of the next word _unless_ this would have done nothing, in which case it first advances the cursor. This is so that repeatedly pressing `w` progresses through the text instead of getting stuck at the first word.

The way it is implemented checks if the characters under the cursor and the next one are in different "categories", which is a little too general because it means that if you're on the last character of a word (or if your word is only 1 character long!) then it moves the cursor off your word. e.g.

```
[a] cat
```
`w`
```
a[ ]cat
```

whereas
```
[t]wo cats
```
`w`
```
[two ]cats
```

which seems wrong and is confusing for many people.

Fixes https://github.com/mawww/kakoune/issues/3132


paaloeye:
https://github.com/mawww/kakoune/issues/752#issuecomment-242909227

src/selectors.cc
```diff
@@ -55,8 +55,11 @@ select_to_next_word(const Context& context, const Selection& selection)
     Utf8Iterator begin{buffer.iterator_at(selection.cursor()), buffer};
     if (begin+1 == buffer.end())
         return {};
-    if (categorize<word_type>(*begin, extra_word_chars) !=
-        categorize<word_type>(*(begin+1), extra_word_chars))
+    const auto cursor_category = categorize<word_type>(*begin, extra_word_chars);
+    const auto next_category = categorize<word_type>(*(begin+1), extra_word_chars);
+    if (not ((cursor_category == CharCategories::Punctuation || cursor_category == CharCategories::Word) &&
```
paaloeye, line 60: I'm not a fan of this

paaloeye (commented):
Kakoune user for about a year; my muscle memory strongly disagrees with that PR.

lrworth:
Thanks for the feedback @paaloeye! Can you please give examples of what situations tripped you up when you tried out this PR? It is definitely worth adding unit tests for those situations, and probably worth adding nuance to the documentation.

paaloeye:
Sorry for the confusion. 

I haven't tried the PR, nor have/had I desire doing so. Why? I'm quite happy with current status quo. 

PR description states `w` is intentionally inconsistent. Given how well **Kakoune** is designed (orthogonal and pragmatic), what have been reported/to-be-fixed is either **wrong** or a design choice.

Regardless the above, based on [Hyrum's Law](https://www.hyrumslaw.com) current behaviour is the public API with user's muscle memory and based on [Occam's razor](https://en.wikipedia.org/wiki/Occam%27s_razor), the right solution is _no solution_.

I could be wrong; @mawww is the ultimate judge here.
Thank you for contributing tho!

mawww:
To be honest I do not really remember why things work this way, however seeing my comment from 2016 in #752 it seems I was quite convinced this approach was correct, and the reasons were likely much more clear in my mind at the time. As @paaloeye mentionned, I think for such a basic movement as `w` I think the ship has sailed, changing it now would be big inconvenience to most users' muscle memory.

lrworth:
It's hard to develop muscle memory when the behaviour of a command depends on the length of the word your cursor is on. Currently we have to read the word under the cursor to know if a cluster like `wd` "delete the word" will actually work.

This change does break `ew` "select the space after this word" which is arguably not as common.

Well anyway, this seems to do the job if we can't change the built-in behaviour:

```kak
map global normal w ';W'
```

PR #5524, opened 2026-07-18
Support windowing in Ghostty

lrworth:
Ghostty [as of 1.3.0](https://ghostty.org/docs/features/applescript#overview) supports AppleScript on macOS. This can be used to implement `ghostty-terminal-%opt{windowing_placement}`, to create splits, tabs, and windows. Milestone for the unreleased 1.4.0 includes https://github.com/ghostty-org/ghostty/issues/11592 which can be used to support the `ghostty-focus` command. I have tested this on the nightly build.

I've been using this for a few days now and it seems to work how it should.

I haven't considered Linux at all for this PR, except that the `ghostty` module should fail gracefully on that platform. I haven't tested on Linux. A Linux Ghostty integration would look quite different, assuming it is possible at all.

### DIFF
--- a/rc/windowing/ghostty.kak
+++ b/rc/windowing/ghostty.kak
@@ -1,0 +1,131 @@
+# https://ghostty.org
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+# Ghostty has some bugs that affect this module:
+# - Commands that return fast are treated as failing. `sleep 0.1s` fails but `sleep 0.5s` is fine.
+#   - As at 2026-07-16 this error is less frightening, but it still happens.
+# - 'wait after command false' does not work: it still waits for a keypress.
+
+provide-module ghostty %{
+
+# ensure that we're running in Ghostty
+evaluate-commands %sh{
+    [ -z "${kak_opt_windowing_modules}" ] || [ "$TERM" = "xterm-ghostty" ] || echo 'fail Ghostty not detected'
+    [ -x "$(which osascript)" ] || echo 'fail Ghostty detected; could not find osascript (macOS only)'
+}
+define-command -params 2.. -docstring '
+ghostty-terminal-impl <direction> <program> [<arguments>]
+direction is tab|window|up|down|left|right' ghostty-terminal-impl %{
+    nop %sh{
+        direction="$1"
+        shift
+        # join the arguments as one string for the shell execution (see x11.kak)
+        args=$(
+            for i in "$@"; do
+                printf "'%s' " "$(printf %s "$i" | sed "s|'|'\\\\''|g")"
+            done
+        )
+
+        # go through another round of escaping for osascript (see iterm.kak)
+        # \ -> \\
+        # " -> \"
+        do_esc() {
+            printf %s "$*" | sed -e 's|\\|\\\\|g; s|"|\\"|g'
+        }
+
+        escaped=$(do_esc "$args")
+        esc_path=$(do_esc "$PATH")
+        esc_tmp=$(do_esc "$TMPDIR")
+        cmd="env PATH='${esc_path}' TMPDIR='${esc_tmp}' $escaped"
+        if [ "$direction" = 'tab' ]; then
+            osascript                                                             \
+            -e "tell application \"Ghostty\""                                     \
+            -e "    set cfg to new surface configuration"                         \
+            -e "    set command of cfg to \"${cmd}\""                             \
+            -e "    set wait after command of cfg to false"                       \
+            -e "    set newtab to new tab in front window with configuration cfg" \
+            -e "    select tab newtab"                                            \
+            -e "end tell" >/dev/null
+        elif [ "$direction" = 'window' ]; then
+            osascript                                                \
+            -e "tell application \"Ghostty\""                        \
+            -e "    set cfg to new surface configuration"            \
+            -e "    set command of cfg to \"${cmd}\""                \
+            -e "    set wait after command of cfg to false"          \
+            -e "    set newwin to new window with configuration cfg" \
+            -e "    activate window newwin"                          \
+            -e "end tell" >/dev/null
+        else
+            osascript                                                                         \
+            -e "tell application \"Ghostty\""                                                 \
+            -e "    set cfg to new surface configuration"                                     \
+            -e "    set command of cfg to \"${cmd}\""                                         \
+            -e "    set wait after command of cfg to false"                                   \
+            -e "    set term to focused terminal of selected tab of front window"             \
+            -e "    set newsplit to split term direction ${direction} with configuration cfg" \
+            -e "    focus newsplit"                                                           \
+            -e "end tell" >/dev/null
+        fi
+      }
+}
+define-command ghostty-terminal-vertical -override -params 1.. -docstring '
+ghostty-terminal-vertical <program> [<arguments>]: create a new terminal as an ghostty pane
+The current pane is split into two, left and right
+The program passed as argument will be executed in the new terminal'\
+%{
+    ghostty-terminal-impl 'right' %arg{@}
+}
+complete-command ghostty-terminal-vertical shell
+
+define-command ghostty-terminal-horizontal -override -params 1.. -docstring '
+ghostty-terminal-horizontal <program> [<arguments>]: create a new terminal as an ghostty pane
+The current pane is split into two, top and bottom
+The program passed as argument will be executed in the new terminal'\
+%{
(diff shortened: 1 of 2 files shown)


rc/windowing/ghostty.kak
```diff
@@ -0,0 +1,129 @@
+# https://ghostty.org
+# ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+
+# Bug: Ghostty (up to at least nightly build 2026-07-16) ignores our `set wait after command
+# of cfg to false` instruction and requires a keypress to close a terminal.
```
lrworth, line 5: https://github.com/ghostty-org/ghostty/discussions/13167. No issue has been created yet.

rc/windowing/ghostty.kak
```diff
@@ -0,0 +1,129 @@
...
+%{
+    ghostty-terminal-impl 'window' %arg{@}
+}
+complete-command ghostty-terminal-window shell
+
+define-command ghostty-focus -override -params ..1 -docstring '
+ghostty-focus [<client>]: focus the given client
+If no client is passed then the current one is used' \
+%{ evaluate-commands %sh{
+  set -x
```
lrworth: Delete

benjaminwil:
For what it's worth, I use Kakoune with Ghostty on Linux via the Wayland windowing module.

Because there is no equivalent functionality like AppleScript integration for Linux, and there is no CLI options for opening new splits (only new windows), my integration is very simple. And I don't hate it. Just:

```
set-option global termcmd "ghostty -e sh -c" 
```

So, I can open new windows, but not new tabs, splits, and so on.

To make this pull request more compatible with Ghostty on Linux as it works today, maybe there is an opportunity to guard in the `ghostty-terminal-impl` function so that we always default to the `window` direction on Linux, as that would be the only one that would work.

rc/windowing/ghostty.kak
```diff
@@ -0,0 +1,128 @@
...
+# ensure that we're running in Ghostty
+evaluate-commands %sh{
+    [ -z "${kak_opt_windowing_modules}" ] || [ "$TERM" = "xterm-ghostty" ] || echo 'fail Ghostty not detected'
+    [ -x "$(which osascript)" ] || echo 'fail Ghostty detected; could not find osascript (macOS only)'
+}
+define-command -params 2.. -docstring '
+ghostty-terminal-impl <direction> <program> [<arguments>]
+direction is tab|window|up|down|left|right' ghostty-terminal-impl %{
+    nop %sh{
+        direction="$1"
```
benjaminwil, line 18: A suggestion for Linux compatibility.

If `uname` is `Darwin` we can use `$1` as the direction,  but if `uname` is `Linux` then we can just eagerly set the direction to `window` since no other shell integration would work at this time.

rc/windowing/ghostty.kak
```diff
@@ -0,0 +1,128 @@
...
+            -e "end tell" >/dev/null
+        elif [ "$direction" = 'window' ]; then
+            osascript                                                \
+            -e "tell application \"Ghostty\""                        \
+            -e "    set cfg to new surface configuration"            \
+            -e "    set command of cfg to \"${cmd}\""                \
+            -e "    set wait after command of cfg to false"          \
+            -e "    set newwin to new window with configuration cfg" \
+            -e "    activate window newwin"                          \
+            -e "end tell" >/dev/null
```
benjaminwil, line 55: For Linux compatibility, we should be able to basically slot in the existing `wayland-terminal-window` here, which is already working for me on Linux.

I just pulled up the version of the Wayland windowing module I'm currently using in case the reference is useful:
```
define-command wayland-terminal-window -params 1.. -docstring '
wayland-terminal-window <program> [<arguments>]: create a new terminal as a Wayland window
%{
    evaluate-commands -save-regs 'a' %{
        set-register a %arg{@}
        evaluate-commands %sh{
            if [ -z "${kak_opt_termcmd}" ]; then
                echo "fail 'termcmd option is not set'"
                exit
            fi
            termcmd=$kak_opt_termcmd
            args=$kak_quoted_reg_a
            unset kak_opt_termcmd kak_quoted_reg_a
            setsid ${termcmd} "$args" < /dev/null > /dev/null 2>&1 &
        }
    }
}
complete-command wayland-terminal-window shell
```

It would need to be changed if it's going to be inside this `elif`, but yeah. Just want you to know that this is working for me already.

rc/windowing/ghostty.kak
```diff
@@ -0,0 +1,128 @@
...
+    if [ $# -eq 1 ]; then
+        printf "evaluate-commands -client '%s' focus" "$1"
+    else
+        # Works on Ghostty commit 73534c4680a809398b396c94ac7f12fcccb7963d (2026-07-16).
+        # Expected to be officially supported beginning with Ghostty 1.4.0.
+        THIS_TTY=$(ps -p "$kak_client_pid" -o tty=)
+        osascript                                                           \
+        -e 'tell application "Ghostty"'                                     \
+        -e '    repeat with term in terminals'                              \
+        -e '        if tty of term as text = "/dev/'"${THIS_TTY% }"'" then' \
```
lrworth, line 116: [alexherbo2 suggests](https://discuss.kakoune.com/t/ghostty-support/2952/6?u=lrworth) using the `pid` property instead. I recall trying this and not being able to match the PIDs. Either I was wrong, or I should add a comment explaining why we can't use the seemingly-obvious solution.

rc/windowing/ghostty.kak
```diff
@@ -0,0 +1,128 @@
...
+# of cfg to false` instruction and requires a keypress to close a terminal.
+
+provide-module ghostty %{
+
+# ensure that we're running in Ghostty
+evaluate-commands %sh{
+    [ -z "${kak_opt_windowing_modules}" ] || [ "$TERM" = "xterm-ghostty" ] || echo 'fail Ghostty not detected'
+    [ -x "$(which osascript)" ] || echo 'fail Ghostty detected; could not find osascript (macOS only)'
+}
+define-command -params 2.. -docstring '
```
lrworth, line 14: [alexherbo2 says](https://discuss.kakoune.com/t/ghostty-support/2952/3?u=lrworth) that
```kak
eval -client kak_client_in_given_ghostty_terminal ghostty-terminal-vertical kak -c %val{session}
```
(where `kak_client_in_given_ghostty_terminal` should be substituted with a client name) should do the right thing, i.e. open a terminal to the right of the specified client, but it does not. This is probably impossible in Ghostty 1.3.1 for the same reason I gave for `ghostty-focus` below, but we should make this work in nightly.

I think we should also include in this module a comment listing commands and their expected outcome to use for regression testing.

lrworth:
> To make this pull request more compatible with Ghostty on Linux as it works today, maybe there is an opportunity to guard in the `ghostty-terminal-impl` function so that we always default to the `window` direction on Linux, as that would be the only one that would work.

@benjaminwil I don't have a linux machine to test with but please let me know if you have something more concrete I should include in this PR. [The docs](https://ghostty.org/docs/linux/systemd#scripting-new-windows) say that `ghostty +new-window` ought to be faster than just `ghostty` because it tries to reuse an existing process, so I suppose it would be good to use that option in `termcmd`. I'm not sure that alone justifies adding Linux support to this PR, especially since as you say the Wayland integration already works and is going to be preferable to most people.

rc/windowing/ghostty.kak
lrworth: `pid of term` doesn't give the PID of Kakoune in windows opened with `:new`. I've added an explanation above the implementation.

rc/windowing/ghostty.kak
lrworth: I've made this command work, but I now have a question: when opening a new window/tab/split should that client be focused automatically or should focus remain on the client you opened it from?

alexherbo2: I think we shouldn’t do anything special.

benjaminwil:
Hey @lrworth, I'll leave it to you and the `rc/` maintainers to decide what the best thing to do is. But I went ahead and made a commit that demonstrates what would be required to make this module compatible with Linux. Feel free to steal it or cherry-pick it if you'd like:

https://github.com/benjaminwil/kakoune/commit/a135a68c723be9762d93db33621bd05a83343a1c

I have tested that this is working as expected. :+1: 

I would personally rather use this `ghostty.kak` over the current `wayland.kak`. I think that, eventually, Ghostty will natively support the programmatic opening of splits and tabs. When that day comes, it won't be much work to tweak this module to support that (for both Linux and macOS).

PR #5528, opened 2026-07-29
Support running kakoune without a controlling /dev/tty terminal

exlee:
This fixes issue when trying to run kakoune without controlling terminal (e.g. trying to run from within Claude Code in full-terminal mode).

Without the fix following crash happens:

```
Received SIGSEGV, exiting.
Callstack:
0   kak  Kakoune::Backtrace::Backtrace()
1   kak  Kakoune::(anonymous namespace)::signal_handler(int)
2   libsystem_platform.dylib  _sigtramp
3   kak  void Kakoune::select_coord<(Kakoune::SelectMode)1>(Kakoune::Context&, Kakoune::BufferCoord) + 76
4   kak  Kakoune::goto_commands<(Kakoune::SelectMode)1>(...)::'lambda'(Key, Context&)::operator()
5   kak  Kakoune::InputModes::NextKey::on_key(Kakoune::Key)
6   kak  Kakoune::InputHandler::handle_key(Kakoune::Key, bool)
7…  kak  execute-keys → context_wrap → CommandManager::execute → evaluate-commands
```

Fix resolves the issue, all tests are green.

## Reproduction

Run Claude Code, run any prompt, press Left Arrow (that backgrounds the process and shows the list or thread), select the thread and press Ctrl-g (open Editor).

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -628,6 +628,19 @@ UniquePtr<UserInterface> create_local_ui(UIType ui_type)
     return ui;
 }
 
+// Give the terminal ui a fd 0 to read keys from, never the piped in one, which
+// is already consumed elsewhere.
+void replace_stdin_with_tty()
+{
+    int tty = open("/dev/tty", O_RDONLY);
+    if (tty < 0)
+        tty = open("/dev/null", O_RDONLY);
+    if (tty < 0)
+        return;
+    dup2(tty, 0);
+    close(tty);
+}
+
 int run_client(StringView session, StringView name, StringView client_init,
                Optional<BufferCoord> init_coord, UIType ui_type,
                bool suspend)
@@ -643,9 +656,7 @@ int run_client(StringView session, StringView name, StringView client_init,
         {
             // move stdin to another fd, and restore tty as stdin
             stdin_fd = dup(0);
-            int tty = open("/dev/tty", O_RDONLY);
-            dup2(tty, 0);
-            close(tty);
+            replace_stdin_with_tty();
         }
 
         EventManager event_manager;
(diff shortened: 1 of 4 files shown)


Screwtapello:
Why would you want to use Kakoune without a controlling terminal?

If you want to execute Kakoune's command language, you can use `kak -f`. 

I guess Kakoune probably shouldn't crash, just exit with a non-zero exit code.

exlee:
> Why would you want to use Kakoune without a controlling terminal?

The case here is using Kakoune as $EDITOR within applications that are running in alternate-screen terminal. I checked bunch of other TUI/CLI editor (nano, neovim, vim, helix) and only Kakoune failed to launch in this environment.

Discovered by accident when I was trying to start Kakoune from within Claude Code running in full TUI mode.

src/main.cc
```diff
@@ -628,6 +628,19 @@ UniquePtr<UserInterface> create_local_ui(UIType ui_type)
     return ui;
 }
 
+// Give the terminal ui a fd 0 to read keys from, never the piped in one, which
+// is already consumed elsewhere.
+void replace_stdin_with_tty()
+{
+    int tty = open("/dev/tty", O_RDONLY);
+    if (tty < 0)
+        tty = open("/dev/null", O_RDONLY);
```
mawww, line 637: So we cannot get any keyboard input in that case ? From what I understand we would just end-up with Kakoune showing up but no ability to directly interact with it except through `kak -p`. Am I missing something ?

exlee: Correct, no keys in that case, but this branch only runs when stdin is piped (not a tty), so there was never a key source without `/dev/tty`. The old code left `fd 0` racing against the `*stdin*` fifo reader on the same pipe. `/dev/null` turns that into a clean EOF/hangup instead.

mawww: Then we should probably error out in this case, which I would expect to happen due to the tty check in TerminalUI constructor (terminal_ui.cc:457). How come this does not fire ? Are we in a situation where stdout is a tty but /dev/tty does not exist ?

Do you have a use case for the half-failing behaviour being proposed ? (where Kakoune quits gracefully but does not provide any interactivity)

exlee: I don't have a use case, just a thought that it might be better than erroring out. Should I change that?

PR #5530, opened 2026-08-09
Fix the string unit tests on 32-bit architectures

ppentchev:
Hi,

Thanks a lot for writing and maintaining Kakoune!

What do you think about the attached trivial patch that fixes the unit tests on systems where sizeof(char *) == 4, such as Linux on i386?

Thanks in advance for looking at this, and keep up the great work!

G'luck,
Peter


ppentchev:
FTR, I don't think the macOS CI failure has anything to do with my change, but of course ICBW.

Issue #4043, opened 2021-02-02
Scrollable info box？


clad26:
### Feature 
Scrollable info box

### Usecase 
I use julialang with Kak-lsp, which uses kakoune's info box to display docs on lsp-hover
The info box fills up the screen and is cut off when I limit the max lines. 
Any way I can get a scroll bar to scroll along the info box? If not, can I get it implemented as a feature? Thank you

krobelus:
probably better to display docs in a dedicated buffer, like `:doc` - this gives more powerful navigation than a scrollbar

clad26:
@krobelus Any image/gif you can show for how :doc works？

clad26:
@mawww,
@krobelus's idea is better
Since I'm assuming that doc uses the scratch buffer, is there any way to anchor it somehow?

If not, should I change my feature request to add anchor to the docs?
Thank you

mawww:
One issue here is that info-boxes are passed as-is to the UI and Kakoune server has minimal control on how they are rendered: The terminal UI does all the cropping logic to try to make it fit. The reason for that approach is to make it possible for eventual graphical UI to display info boxes as they see fit, but it means that any scrolling support in the info box will need to take place at the client UI level.

This is definitely doable, but it would be a bit weird, as those keys would not go through the regular input stack (no modes, not mappings...).

eko234:
I don't know how registers work very well, but maybe adding the content info boxes to a determined register would make it easy to
write hooks and bindings to handle large info boxes,

agmitron:
> This is definitely doable, but it would be a bit weird, as those keys would not go through the regular input stack (no modes, not mappings...).

Can it be implemented like in Helix? It's very simple and has only scroll, no motions inside the floating-window like Neovim. So the only keymap is needed is to scroll up/down/left/right

agmitron:
https://github.com/mawww/kakoune/pull/5544

Hi there, I created a simple implementation for this. Would be glad for your feedback. Let me know if I haven't covered any important scenarios. 

> This is definitely doable, but it would be a bit weird, as those keys would not go through the regular input stack (no modes, not mappings...).

@mawww I have also tried to do it consistently, so it goes through the server, but the server doesn't store any state, it just echoes the event back to the client.

krobelus:
if the motivation is kakoune-lsp's `lsp-hover`, consider using `lsp-hover-buffer` instead

agmitron:
[...]

PR #5544, opened 2026-08-30
#4043: scrollable info-box

agmitron:
Simple scroll implementation for info-boxes. The server just echoes the event, then the scroll is handled on the client side. The state of the scroll is also stored on the client, the server knows nothing about scroll's cursors. 

https://github.com/user-attachments/assets/d688380d-c740-4d2e-86c0-39d5abc93487

Disclaimer: I have used LLM assistance to write the code.


agmitron:
Closed in order to a simpler and less intrusive client-only solution: https://github.com/mawww/kakoune/pull/5547

